PR #284, opened 2019-11-06
Using the Component Story Format in Storybook

### LINKED ISSUE #262: RFC: Should we use Component Story Format
The pending release of Storybook 5.3 will surface a new API by which we can write the stories that we surface in our packages. The Storybook team is using the tagline "Simple, portable, future-proof component examples" in that the new format is no longer tied directly to the Storybook APIs, which allows the code to be used across numerous other contexts. Immediately, I could see us leveraging this change in our unit tests and benchmarks, but clearly the idea is that more and more tools could leverage this format: design tools (hello XD?), build tools, possibly markdown based documentation, etc. Please read more on the format here: https://medium.com/storybookjs/component-story-format-66f4c32366df

I'd like to capture people's thoughts on the capabilities that this opens for the SWC project to see if this makes sense for us to adopt. To do so, I'm working on a single package example of what this change would look like for comparison. I'll share said PR in a comment, please try to keep comments about meta-conversation of "should we use this format?/how can we leverage this format for good?" in this issue as best we can and leverage the PR comments to focus on the actual technical implementation of such a transition.

Westbrook:
## Description
The upcoming release of Storybook 5.3 will surface the ability to use the Component Story Format in our stories, which can then be subsequently leveraged in other contexts (e.g. our tests, etc.) to reduce repeated code across our packages. 

`@open-wc/demoing-storybook@^1.3.7` supports this by default, so I've applied both in this RFC.

## Related Issue
fixes #262

## Motivation and Context
There is a good amount of duplicate markup that we could eliminate from just our current workflow with this move. Beyond that feel free to join the conversation on how else we might be able to leverage this conversion in #262.

## How Has This Been Tested?
It is a refactor of unit tests and demos, all of which still work!

## Types of changes
- [x] Refactor!

## Checklist:


packages/dropdown/stories/dropdown.stories.ts
```diff
@@ -9,96 +9,52 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
...
-                        Save Selection
-                    </sp-menu-item>
-                    <sp-menu-item disabled>
-                        Make Work Path
-                    </sp-menu-item>
-                </sp-menu>
-            </sp-dropdown>
-        `;
-    })
-    .add('Initial Value', () => {
```
cuberoot: Was this test redundant? Seems to have been dropped.

Westbrook: Add this story (and a visual regress in associate with it) back to the repo. Thanks for the second set of eyes!

PR #302, opened 2019-11-20
fix: normalize "event" and "error" argument names

### LINKED ISSUE #153: We should standarize on basic variable names
### Expected Behaviour

Our code should be consistent for variable names. Currently we use some various short forms for `event` and `error`. I propose that we use the full names everywhere as that is a simple rule to remember and is the easiest to read. I am open to discussion though. We at least need to be consistent.

### Actual Behaviour

We use all kinds of variable names (e.g. evt, e, etc)

Westbrook:
## Description
- update existing code to meet the argument naming standard
- create and apply an eslint plugin to prevent this from happening with understood names in the future.

## Related Issue
fixes #153

## Motivation and Context
Code should be more predictable

## How Has This Been Tested?
- wrote some non-compliant code and then tried to commit it, eslist said "NO"

image

## Types of changes

## Checklist:

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,6 +1,6 @@
 {
     "root": true,
-    "plugins": ["notice"],
+    "plugins": ["notice", "spectrum-web-components"],
     "env": {
         "browser": true,
         "node": true,
@@ -12,6 +12,10 @@
         "sourceType": "module"
     },
     "rules": {
+        "spectrum-web-components/prevent-argument-names": [
+            "error",
+            ["e", "ev", "evt", "err"]
+        ],
         "notice/notice": [
             "error",
             {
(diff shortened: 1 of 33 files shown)


utils/eslint/index.js
```diff
@@ -0,0 +1,29 @@
+module.exports = {
```
Westbrook: Just learning that there is a `linters` folder in the project 😮 

This packages should likely live there, right?

Later, we should look into moving all three, as well as the example project, into `/packages/` so that Greenkeeper and Lerna can manage them better...

cuberoot: Yah, looks like there's a pattern for custom linters, so it would be good to follow it. I will defer to you on the packages thing. I'm still ramping on lerna.

PR #306, opened 2019-11-22
[#303] Implement the 'sp-toast' pattern.

### LINKED ISSUE #303: Implement `sp-toast` pattern
### Expected Behaviour
The `sp-toast` pattern should be available for use in the library.

### Actual Behaviour
https://opensource.adobe.com/spectrum-css/components/toast/

Westbrook:
## Description
- add the "sp-toast" pattern
- add the "sp-clear-button" pattern to support
- refactor "sp-search" to also use "sp-clear-button"

## Related Issue
fixes #303

## Motivation and Context
Users should be informed of things.

## How Has This Been Tested?
- unit tests
- snapshot tests

image

## Types of changes

## Checklist:


__snapshots__/Toast.md
```diff
@@ -0,0 +1,136 @@
...
+    <sp-clear-button
+        data-js-focus-visible=""
+        label="Close"
+        tabindex="0"
+        variant="overBackground"
+    ></sp-clear-button>
+</div>
+```
+
+#### `loads - [variant="error"]`
```
davidyxu: should we bother exposing the error/warning variants on the storybook page? it looks like the spectrum-css advises using the negative variant instead.
https://opensource.adobe.com/spectrum-css/components/toast/#Error

packages/toast/src/toast.ts
```diff
@@ -0,0 +1,207 @@
...
+export class Toast extends LitElement {
+    public static get styles(): CSSResultArray {
+        return [toastStyles];
+    }
+
+    @property({ type: Boolean, reflect: true })
+    public open = false;
+
+    @property({ type: Boolean, reflect: true })
+    public disappearing = false;
```
davidyxu: not sure if this is the responsibility of swc to enforce, but looking at the toast design on spectrum.adobe.com, anything that is "actionable" is should not be able to be disappearing:
https://spectrum.adobe.com/page/toast/#Priority-queue

packages/toast/src/toast.ts
```diff
@@ -0,0 +1,207 @@
...
+
+    /**
+     * The variant applies specific styling when set to `negative`, `positive`, `info`, `error`, or `warning`.
+     * `variant` attribute is removed when not matching one of the above.
+     *
+     * @param {String} variant
+     */
+    @property({ type: String })
+    public set variant(variant: string) {
```
davidyxu: could we export a ToastVariant type for typescript users here?

packages/toast/README.md
```diff
@@ -0,0 +1,61 @@
+## Overview
+
+`sp-toast`s display brief, temporary notifications. They are noticeable but do not disrupt the user experience and do not require an action to be taken.
```
davidyxu: could we add an example of disappearing as well

packages/toast/src/toast.ts
```diff
@@ -0,0 +1,207 @@
...
+                label=${label}
+                name=${name}
+                size="s"
+            ></sp-icon>
+        `;
+    }
+
+    private countdownStart = 0;
+    private nextCount = -1;
+    private countdownLength = 6000;
```
davidyxu: should we make this countdown length a property? seems like something that users of this component might want to control

__snapshots__/Toast.md
adixon-adobe: I could see allowing `error` as a pseudonym for `negative`. I'm guessing most coders are gonna *want* to call the red version an error.

I'd also vote for keeping warning since it'll help with our adoption (but we should plan to list it as deprecated).

packages/button/src/clear-button.ts
```diff
@@ -0,0 +1,42 @@
...
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import { CSSResultArray, TemplateResult, html, property } from 'lit-element';
+import { ButtonBase } from './button-base.js';
+import buttonStyles from './clear-button.css.js';
+import crossMediumStyles from '@spectrum-web-components/icon/lib/spectrum-icon-cross-medium.css.js';
+
+export class ClearButton extends ButtonBase {
```
adixon-adobe: is there a reason you added this rather than making an `over-background` variant of `<sp-button>`?

packages/toast/src/toast.ts
adixon-adobe: agreed, especially if for some reason you have long text. The guideline says "at least 5 seconds" so it might be good to enforce a minimum if so.

packages/toast/src/toast.ts
adixon-adobe: I'd vote for `auto-dismiss` rather than `disappearing` for the naming here.

packages/button/src/clear-button.ts
Westbrook: `<sp-button variant="overBackground">` already exists, but with different delivery. This is specifically the "Clear Button" sub-pattern and it's `variant="overBackground"` alterations, which the CSS outline as being unique: https://opensource.adobe.com/spectrum-css/components/clearbutton/

I struggle with how our library should support these sub-variants. Should they be new elements, should we pack the old elements with more attributes/etc., this seems like the right way being this ties content and style delivery into one place. Thoughts?

packages/toast/src/toast.ts
Westbrook: @davidyxu It's for this reason that I added this as an alternative instead of the default. For that, I am working from https://github.com/adobe/react-spectrum-v3/blob/master/specs/accessibility/Toast.mdx as supplied by @jnurthen  as the source of the timing decisions here. I should have documented it some way, but I meant to deliver a baseline, and then extend functionality later to manage text with more than 120 words. Maybe it makes sense to surface that math as an exported function? A la:
```
export const howLongToWait = (words: number | string) => number; // equalling 1 for every 120 words
```

In that case, @adixon-adobe rather than relying on `disappearing` or `auto-dismiss` we can match more closely to https://github.com/adobe/react-spectrum-v3/blob/master/packages/%40react-types/toast/src/index.d.ts#L9 and use `timeout` to signify that a toast is self-dismissing?

packages/toast/src/toast.ts
Westbrook: Would someone use that differently that simply changing this to?
```suggestion
    public set variant(variant: '' | 'negative' | 'positive' | 'info' | 'error' | 'warning') {
```
With better understanding of this, I can addresss other instances of this pattern in a separate PR.

__snapshots__/Toast.md
davidyxu: sounds reasonable to leave it until spectrum-css removes the variants 👍 and just to clarify, do we need to leave it in stories for testing coverage reasons?

packages/toast/src/toast.ts
davidyxu: i'm open to either approach, was thinking that an exported toast variant type could be useful for clients to define for model interfaces that drive the toast view, although i think you could pull out the typing either way with some typescript trickery.

__snapshots__/Toast.md
Westbrook: It doesn't need to remain in the stories, no. I'll remove there.

__snapshots__/Toast.md
```diff
@@ -0,0 +1,136 @@
...
+        variant="overBackground"
+    ></sp-clear-button>
+</div>
+```
+
+#### `loads - [variant="negative"]`
+
+```html
+<sp-icons-medium style="display: none;"></sp-icons-medium>
+<sp-icon class="type" label="Negative" name="ui:AlertSmall" size="s"></sp-icon>
```
jnurthen:
```suggestion
<sp-icon class="type" label="Error" name="ui:AlertSmall" size="s"></sp-icon>
```

__snapshots__/Toast.md
```diff
@@ -0,0 +1,136 @@
...
+</div>
+```
+
+#### `loads - [variant="positive"]`
+
+```html
+<sp-icons-medium style="display: none;"></sp-icons-medium>
+<sp-icon
+    class="type"
+    label="Positive"
```
jnurthen:
```suggestion
    label="Success"
```

packages/toast/src/toast.ts
```diff
@@ -0,0 +1,215 @@
...
+        let name = '';
+        switch (variant) {
+            case 'info':
+                label = 'Information';
+                name = 'ui:InfoSmall';
+                break;
+            case 'negative':
+            case 'error': // deprecated
+            case 'warning': // deprecated
+                label = 'Negative';
```
jnurthen:
```suggestion
                label = 'Error';
```

packages/toast/src/toast.ts
```diff
@@ -0,0 +1,215 @@
...
+                name = 'ui:AlertSmall';
+                break;
+            case 'positive':
+            case 'success': // deprecated
+                label = 'Positive';
```
jnurthen:
```suggestion
                label = 'Success';
```

packages/button/src/clear-button.ts
adixon-adobe: Oh -- took me a while to realize that the "Clear Button" is really just a *specific* icon button with the x. (I was thinking Clear had to do with visibility...). It seems a little weird to me to create a custom button for this rather than just creating the icon button inline with the right icon. Just an observation though - no need to hold anything up here.

packages/toast/README.md
Westbrook: @adixon-adobe @davidyxu Can the two of you please take a look at the [a11y doc](https://github.com/adobe/react-spectrum-v3/blob/master/packages/%40react-types/toast/src/index.d.ts#L9) and [this comment](https://github.com/adobe/spectrum-web-components/pull/306/files#r350175966) to help clarify how we'd like the API to come together before I add this example. Thanks!

packages/toast/src/toast.ts
davidyxu: The timeout option used in react-spectrum makes the most sense to me, where the lack of a timeout signifies that it isn't self-dismissing. I guess the downside to this approach is that there wouldn't be an easy way to provide a sensible default timeout setting?

Westbrook: I think _maybe_ we could still require the timeout to be at least 6000 milliseconds, as it’s the baseline a11y requirement. But, that’s a great point!

davidyxu: I think in an ideal world, we could raise console warnings when we go below a recommended threshold when in dev mode, but these get compiled out when building for prod. I'm not aware of any clean way of setting this generically in SWC.

davidyxu:
in the interest of actually getting this ticket in, how does everyone feel about creating an issue to continue these discussions around the ideal toast api for SWC and merge in the current api with disappearing?

Westbrook:
@davidyxu I think that's a great idea. I was 3/4 of the way done with the shift to `timeout` when you posted, so please let me know if you're still comfortable with moving forward in that way with this slightly updated code. To support the ongoing API review, I've create the following issue: https://github.com/adobe/spectrum-web-components/issues/351

PR #307, opened 2019-11-22
chore: add version linting to ensure out deps stay in sync

Westbrook:
## Description
A quick tool, as inspired by https://github.com/open-wc/open-wc/blob/master/scripts/lint-versions.js to ensure out dependancies stay in order.

## Types of changes

## Checklist:
- [ ] My code follows the code style of this project.
- [ ] I have read the **CONTRIBUTING** document.


scripts/lint-versions.js
```diff
@@ -0,0 +1,103 @@
...
+        newVersions,
+    };
+}
+
+let currentVersions = readPackageJsonDeps('./package.json');
+let endReturn = 0;
+
+// find all versions in the monorepo
+getDirectories('./packages').forEach((subPackage) => {
+    const filePath = `./packages/${subPackage}/package.json`;
```
davidyxu: not sure if this is still an issue, but we might want to use `path.join` instead for os compatibility (specifically windows). @benjamind could you chime in?

scripts/lint-versions.js
```diff
@@ -0,0 +1,103 @@
...
+        return result;
+    }
+    return {};
+}
+
+function readPackageJsonNameVersion(filePath) {
+    if (existsSync(filePath)) {
+        const jsonData = JSON.parse(readFileSync(filePath, 'utf-8'));
+        const result = {};
+        result[jsonData.name] = `^${jsonData.version}`;
```
davidyxu: sounds like this sets the expectation that the ^ matcher should be used consistently between packages, should we document this explicitly somewhere?

scripts/lint-versions.js
benjamind: I don't tend to use the windows filesystem for my web development anymore, all under WSL now and I think this is probably the recommended approach. However, probably should just use one of the path module functions, I think you can just do `path.normalize('./packages/${subPackage}/package.json')` and it will convert all the paths to the platform specific path at runtime.

Westbrook: Updated.

scripts/lint-versions.js
Westbrook: Interesting. I think it's outlining we only confirm that versions match when the versions are not fixed or `~`. I think if you use either of those, you need to make a good reasoning for it in a PR, normally. Maybe that's over simplifying?

Happy to expand on the documentation here. What sort of messaging would make either of these realities clear?

davidyxu: I think using ^ is a reasonable expectation, perhaps adding this script to a pre-commit hook would be sufficient to raise attention to this? Otherwise this seems like an easy thing to miss (in that unless a developer is aware of this script existing and knows what it does, they wouldn't know that ~ or fixed versions are not allowed).

Westbrook: I added some documentation to that end. Do you think it would also be good to have it run on commit? I've not really run into this issue in this repo, so I'm not sure how to best apply it here.

scripts/lint-versions.js
```diff
@@ -0,0 +1,104 @@
...
+    const filePath = path.normalize(`./packages/${subPackage}/package.json`);
+    currentVersions = {
+        ...currentVersions,
+        ...readPackageJsonNameVersion(filePath),
+    };
+});
+
+// lint all versions in packages
+getDirectories('./packages').forEach((subPackage) => {
+    const filePath = `./packages/${subPackage}/package.json`;
```
davidyxu: could we also use `path.normalize` down here?

scripts/lint-versions.js
davidyxu: 👍 i think what you have is fine as is, we can always add a hook later if it comes up frequently

scripts/lint-versions.js
Westbrook: Check!

PR #312, opened 2019-11-26
docs: add a search field to the documentation

cuberoot:
## Description

Continuing to make our documentation more like other sites in the spectrum web ring by adding in a search field.

## Motivation and Context

We have been asked to align our documentation site with others in the spectrum web ring

## How Has This Been Tested?

Manual interaction with the documentation site.

Screen Shot 2019-11-26 at 2 41 03 PM

## Types of changes

## Checklist:


cuberoot:
Closing this. I am going to try using some of the overlay machinery instead...

cuberoot:
I pushed a better version that uses the `overlay-root`

documentation/scripts/buildSearchIndex.js
```diff
@@ -0,0 +1,75 @@
...
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+const lunr = require('lunr');
+const globby = require('globby');
+const path = require('path');
+const fs = require('fs-extra');
```
cuberoot: I am impressed with how easy it is to build a canned search index using `lunr`. This is the same library that the spectrum docs page uses (by chance)

documentation/src/components/side-nav-search.ts
```diff
@@ -0,0 +1,234 @@
...
+    private popover?: HTMLElement | null;
+
+    private handleSearchInput(event: InputEvent) {
+        if (event.target) {
+            const searchField = event.target as Search;
+            this.updateSearchResults(searchField.value);
+        }
+    }
+
+    private openPopover() {
```
cuberoot: We should nut out an API for doing programmatically triggered overlays (as opposed to using `overlay-trigger`).

cuberoot: See #321

documentation/src/components/side-nav-search.ts
```diff
@@ -0,0 +1,234 @@
...
+        } else {
+            return html``;
+        }
+    }
+
+    render(): TemplateResult {
+        return html`
+            <div id="search-container">
+                <div id="search-field">
+                    <sp-search
```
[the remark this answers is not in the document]
cuberoot: I filed #314 for this. Should search fields ever use the builtin autocomplete? I presume not...

documentation/src/components/side-nav-search.ts
```diff
@@ -0,0 +1,234 @@
...
+import sideNavSearchMenuStyles from './side-nav-search.css';
+import lunr from 'lunr';
+import { Search } from '../../../packages/search';
+import {
+    OverlayOpenDetail,
+    OverlayCloseDetail,
+} from '../../../packages/overlay-root';
+import { AppRouter } from '../router';
+
+const searchIndex = require('../../components/searchIndex.json');
```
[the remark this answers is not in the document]
cuberoot: Agreed. I will refactor. I know a bunch about Webpack dynamic loading.

documentation/src/components/side-nav-search.ts
```diff
@@ -0,0 +1,234 @@
...
+                    ${this.results.map(
+                        (category) => html`
+                            <sp-menu-group>
+                                <span slot="header">${category.name}</span>
+                                ${category.results.map(
+                                    (result) => html`
+                                        <sp-menu-item
+                                            @click=${() => {
+                                                AppRouter.go(result.url);
+                                                this.closePopover();
```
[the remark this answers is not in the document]
cuberoot: Almost. If you click on the page you are already on, the overlay doesn't close. So if you are looking at the button docs, and click on button in the search results, nothing would happen without this close call.

documentation/src/components/side-nav-search.ts
```diff
@@ -0,0 +1,234 @@
...
+            result.push({ name, results, maxScore });
+        }
+        result.sort((a, b) => {
+            if (a.maxScore < b.maxScore) return 1;
+            if (a.maxScore > b.maxScore) return -1;
+            return 0;
+        });
+        this.results = result;
+
+        this.openPopover();
```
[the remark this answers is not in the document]
cuberoot: I forgot this. I had intended to add it.

documentation/src/components/side-nav-search.ts
```diff
@@ -0,0 +1,234 @@
...
+                    <sp-search
+                        @input=${this.handleSearchInput}
+                        @change=${this.handleSearchInput}
+                    ></sp-search>
+                </div>
+                <sp-popover id="search-results-menu" open>
+                    <style>
+                        #search-results-menu {
+                            width: 368px;
+                            min-height: 220px;
```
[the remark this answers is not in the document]
cuberoot: I might be getting carried away. However, I think it is shaking out some good stuff as we try to write real site code...

documentation/src/components/side-nav-search.ts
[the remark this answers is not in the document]
Westbrook: That's a little ticky. We've not implemented an extended pattern like this in Crisp and likely won't, being our search is literally image (so everything search), so having the autocomplete feels like a feature. We will _probably_ build an alternate UI, at some point, for recent searches, but it's likely not a near-future concept. We've customized other concepts as "organisms" in our application from the "atoms" outlined herein, so if we decided not to have optional, it wouldn't block our usage, either way.

documentation/src/components/side-nav-search.ts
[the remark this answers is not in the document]
Westbrook: 🍿 ...looking forward to learning some then. Me and Webpack have sharp words whenever we work together.

documentation/src/components/side-nav-search.ts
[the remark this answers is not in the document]
Westbrook: Is that because we're gobbling up the click event somewhere in the menu? It makes sense as to why the code is there but feels really unexpected.

documentation/src/components/side-nav-search.ts
```diff
@@ -0,0 +1,234 @@
...
+        return html`
+            <div id="search-container">
+                <div id="search-field">
+                    <sp-search
+                        @input=${this.handleSearchInput}
+                        @change=${this.handleSearchInput}
+                    ></sp-search>
+                </div>
+                <sp-popover id="search-results-menu" open>
+                    <style>
```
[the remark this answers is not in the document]
cuberoot: Well no. When part of the shadow DOM gets reparented to the `active-overlay` it loses the context of being in the component. I could create a sub-component to solve that problem, but I didn't want to add more files :-)

documentation/src/components/side-nav-search.ts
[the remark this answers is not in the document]
cuberoot: We allow overlays to have arbitrary UI content, so we only close the overlay automatically for click events that happen outside of the overlay content.

cuberoot:
@Westbrook I refreshed this change. I made the following changes:
- added message for when there are no results
- close the overlay when there are less than 3 results
- used the `@query` decorator to simplify code (learned something new!)
- delay loading the search index until `firstUpdate`
- turned off the built-in autocomplete

The one thing I see as missing is keyboard control of the search menu. I don't have time for that right at the moment.

documentation/src/components/side-nav.ts
```diff
@@ -73,6 +75,7 @@ class SideNav extends LitElement {
                     </div>
                 </div>
                 <div id="navigation">
+                    <docs-search></docs-search>
```
Westbrook: Please make this the last element in `id="nav-header"` so that the delivery more closely matches https://spectrum.adobe.com/ where the search bar is "fixed".

documentation/src/components/side-nav.ts
```diff
@@ -15,9 +15,11 @@ import {
     CSSResultArray,
     property,
     PropertyValues,
+    TemplateResult,
 } from 'lit-element';
 import { ComponentDocs } from '../../components';
```
Westbrook: This uses different data than `import('../../components/searchIndex.json')` so the initial navigation listings are different than the search listings...

documentation/src/components/side-nav-search.ts
```diff
@@ -0,0 +1,212 @@
...
+            if (a.maxScore > b.maxScore) return -1;
+            return 0;
+        });
+        this.results = result;
+
+        this.openPopover();
+    }
+
+    firstUpdated(): void {
+        import('../../components/searchIndex.json').then((searchIndex) => {
```
Westbrook: I think because this is `import`ed instead of `fetch`ed it is pushed into the initial bundle. It's kind of large, and would be nice to defer. However, it also has the nicer component name listing, so it may also be nice to use more broadly... see related note in https://github.com/adobe/spectrum-web-components/pull/312/files#diff-e9f9b198c8c59ad0fe92c58f32b4f33b

documentation/src/components/side-nav-search.ts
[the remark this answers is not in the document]
Westbrook: I'd still like to remove this min-height...

documentation/src/components/side-nav.css
```diff
@@ -14,7 +14,7 @@ governing permissions and limitations under the License.
     display: flex;
     flex-direction: column;
     max-height: 100vh;
-    overflow: auto;
+    overflow: visible;
```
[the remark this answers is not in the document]
Westbrook: Can remove?

documentation/src/components/side-nav-search.ts
cuberoot: @Westbrook  Actually, using `import()` in webpack results in a separate dynamically loaded js file. So that is not a problem here. I can see about using this data more broadly though.

Screen Shot 2020-02-06 at 9 26 34 AM

Westbrook: Oh, it parses it into a JS file, got it. I was like "where's the .json file" 🙈

cuberoot: Ok, I reworked things so that we always work off of the search index, even for side bar.

cuberoot:
removed `overflow: visible` and `min-height: 220px` as requested

documentation/src/components/side-nav.ts
cuberoot: Done

Westbrook:
Can we get the search results to be cased the same way as the menu?

image

Also, we now have PR preview!?!?! https://deploy-preview-312--stupefied-franklin-07fec4.netlify.com/

cuberoot:
Yes...yes you can. Search results now match in case. Good catch.

packages/overlay/src/active-overlay.css
```diff
@@ -35,7 +35,7 @@ governing permissions and limitations under the License.
 }
 
 :host {
-    z-index: 2;
+    z-index: 1000;
```
cuberoot: @Westbrook This fixes the z issue. I was not sure how to set this for overlays. It looks like spectrum-css uses 100 for side navs, so we have to be higher than that...

Westbrook: I think this is great. I'm gonna make a ticket to review this at a later date as it should probably be powered by CSS custom properties, but should be backed by some strong agreement of what it actually empowers... https://github.com/adobe/spectrum-web-components/issues/485

benjamind: Isn't this sort of problem why we made it possible to send overlays to specific places through the event system?

i.e. you would have an 'overlay container' div at the document root level, this would have a z-index defined for the overlay plane, and we'd have all our overlay content transferred to this root when popping up?

Likewise I could see us setting up several different containers for different z-index destined content, i.e. tooltips are always at the very top most plane, dialogs are a plane below etc.

benjamind: Ah actually taking a look at our current overlay-stack behavior we don't currently allow this.

I think it'd be really great if the `sp-overlay-query` event could carry with it a `overlay-type` property and allow handlers of the event to set a `overlay-root` property on the event payload. This would then allow us to carry a 'tooltip' value for the `overlay-type` property for tooltip overlays, and for consumers to add a new root for the overlays that are 'tooltips' that can then have specific z-indexing. This allows us to explicitly control at the document level both where overlays are destined to be hoisted to, and the z-index ordering at the document level. This would allow more natural use of z-index ordering I think as well?

Westbrook: Agree that now that we have popper available we should do a review of the broader functionality like this that we want to power from this new foundation. I've added #487 with a quote of the above to discuss as I'm not sure it's specifically related to this PR.

cuberoot:
Fixed a build issue that entered in when I rebased.

cuberoot:
@Westbrook I think this is ready to go if it still passes muster

PR #336, opened 2019-12-02
docs: optimize loading process and build

Westbrook:
## Description
Speculative performance tuning for the docs site.

## Motivation and Context
Smaller, faster, better!

## Types of changes

## Checklist:

### DIFF
--- a/documentation/components/index.ts
+++ b/documentation/components/index.ts
@@ -1,7 +1,11 @@
 import { toHtmlTemplateString } from '../src/utils/templates';
 import { TemplateResult } from 'lit-element';
 
-const componentDocs = require.context('../../packages', true, /\.md$/);
+const componentDocs = require.context(
+    '../../packages',
+    true,
+    /(?<!CHANGELOG)\.md$/
+);
 
 export const ComponentDocs = new Map<string, TemplateResult>();
 
--- a/documentation/src/components/component.ts
+++ b/documentation/src/components/component.ts
@@ -9,14 +9,14 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
 OF ANY KIND, either express or implied. See the License for the specific language
 governing permissions and limitations under the License.
 */
-import { html, CSSResultArray } from 'lit-element';
-import './code-example';
-import { ComponentDocs } from '../../components';
-import { ComponentApiDocs } from '../../api-docs';
+import { html, CSSResultArray, TemplateResult } from 'lit-element';
 import { LayoutElement } from './layout';
 import componentStyles from './markdown.css';
 import { AppRouter } from '../router';
 import { TabList } from '../../../packages/tab-list';
+
+let ComponentApiDocs: Map<string, TemplateResult>;
+let ComponentDocs: Map<string, TemplateResult>;
 
 enum TabValue {
     Api = 'api',
@@ -32,6 +32,9 @@ class ComponentElement extends LayoutElement {
         };
         pathname: string;
     };
+
+    private apiDocsLoaded = false;
+    private docsLoaded = false;
 
     public static get styles(): CSSResultArray {
         return [super.styles, componentStyles];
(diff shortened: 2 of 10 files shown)


documentation/src/components/component.ts
```diff
@@ -9,15 +9,15 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
 OF ANY KIND, either express or implied. See the License for the specific language
 governing permissions and limitations under the License.
 */
-import { html, CSSResultArray } from 'lit-element';
-import './code-example';
-import { ComponentDocs } from '../../components';
-import { ComponentApiDocs } from '../../api-docs';
```
Westbrook: Move data loading off the critical path.

documentation/src/components/component.ts
```diff
@@ -62,6 +65,27 @@ class ComponentElement extends LayoutElement {
         });
     }
 
+    loadDocs() {
```
Westbrook: Lazily load the documentation. `ComponentApiDocs` will theoretically already be available, but having it here means when we move to the updated side nav process that we should get that off the initial load as well!

documentation/src/components/component.ts
```diff
@@ -100,5 +124,10 @@ class ComponentElement extends LayoutElement {
         }
         return result || super.renderContent();
     }
+
+    connectedCallback() {
```
Westbrook: Move the code example element off of the critical path.

documentation/src/components/side-nav.ts
```diff
@@ -16,7 +16,7 @@ import {
     property,
     PropertyValues,
 } from 'lit-element';
-import { ComponentDocs } from '../../components';
+import { ComponentApiDocs } from '../../api-docs';
```
Westbrook: This is the smaller of the two files, and has the data in the same structure so it will serve as a drop-in replacement here.

documentation/src/main.ts
```diff
@@ -10,13 +10,10 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import loadPolyfills from '@open-wc/polyfills-loader';
 import polyfill from 'element-closest';
 
-loadPolyfills().then(() => {
```
Westbrook: OpenWC no longer suggests the use of this technique https://www.npmjs.com/package/@open-wc/polyfills-loader

documentation/webpack.config.js
```diff
@@ -147,5 +150,5 @@ module.exports = merge(openWcConfig, {
             },
         ],
     },
-    plugins: [new WebpackBar()],
+    plugins: [new WebpackBar(), new BundleAnalyzerPlugin()],
```
Westbrook: Added this, but need to confirm the way to conditionally include so it doesn't block on the CI...

package.json
```diff
@@ -185,7 +185,8 @@
         "posthtml-transform": "^1.0.6",
         "prettier": "^1.16.4",
         "pretty-quick": "^1.10.0",
-        "prismjs": "^1.15.0",
+        "prism-themes": "^1.3.0",
+        "prismjs": "^1.17.1",
```
Westbrook: Need to remove all of these Prism related changes. Was attempting to get something more accessible in the light range, but none of their standard themes pass AA a11y standards even.

documentation/components/index.ts
```diff
@@ -1,7 +1,11 @@
 import { toHtmlTemplateString } from '../src/utils/templates';
 import { TemplateResult } from 'lit-element';
 
-const componentDocs = require.context('../../packages', true, /\.md$/);
+const componentDocs = require.context(
+    '../../packages',
+    true,
+    /(?<!CHANGELOG)\.md$/
```
Westbrook: Prevent the changelogs from getting loaded, we don't use them anyway.

documentation/src/components/layout.ts
```diff
@@ -95,4 +95,9 @@ export class LayoutElement extends RouteComponent {
             </sp-theme>
         `;
     }
+
+    connectedCallback() {
```
Westbrook: Realized I had added this to ALL of the extensions of this base class, so refactored.

PR #360, opened 2019-12-12
fix: apply Focuable styles in class extensions

### LINKED ISSUE #318: Focusable components gets tab selection focus if you click in the right spot
### Expected Behaviour
Clicking on a checkbox (or other focusable component) selects it, but doesn't show as having tab focus

### Actual Behaviour
You can get a blue selected outline if you click in the right place.

#### Steps to Reproduce
You can repro pretty easily by clicking in the upper-right corner of the checkbox on the docs. You might need to click a few times to get it to happen.

#### Screenshot

selected-focus

Westbrook:
## Description 
Use `super.styles` in all `Focusable` extensions that don't already have it available. This allows for `click` events on non-`focusElement` covered areas to occur without errantly focusing the element root.

## Related Issue
fixes #318

## Motivation and Context
@cdvu1 did a great job fixing this in the elements her team is actively using, but as I was about to release those updates I realized that the change would bump the version of the element included herein anyway, so I wanted to ensure that they include these updates, too.

## How Has This Been Tested?
Manually in storybook

## Types of changes

## Checklist:

### DIFF
--- a/packages/link/src/link.ts
+++ b/packages/link/src/link.ts
@@ -30,7 +30,7 @@ import linkStyles from './link.css.js';
  */
 export class Link extends Focusable {
     public static get styles(): CSSResultArray {
-        return [linkStyles];
+        return [...super.styles, linkStyles];
     }
 
     @query('#anchor')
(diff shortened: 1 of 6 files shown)


Westbrook:
@cdvu1 I know I told you we could skip these, but I decided that we should get them out of the way and prevent any confusion when making a release with these updates. Thanks in advance for taking a look!

PR #365, opened 2019-12-13
Tooltip/Overlay Updates

Westbrook:
## Description
- update `active-overlay` to position itself after the RAF _always_ so that it is measured after styles are applied and that it's not placed twice where it can cancel the mouse over event and hide errantly.

## Motivation and Context
Delivery UI fidelity and hover interactions as expected.

## How Has This Been Tested?
Manually in Storybook, see below...

### After
image
### Before
image

## Types of changes

## Checklist:


Westbrook:
## Future Work
In the future, we should look into doing the work to update the "carat" placement when the lack of hard word breaks causes the `active-overlay` element to change its position/width, but I think this is a good 3/4 fix. Future work can be tracked in #366

Issue #364, opened 2019-12-13
Radio Button Group breaks when radio value is an integer


bengfarrell:
### Expected Behaviour
No JS error

#### Sample Code that illustrates the problem

```
<sp-radio-group selected="${Model.input.outputStride}" name="outputstride">
          <sp-radio value="8">8</sp-radio>
          <sp-radio value="16">16</sp-radio>
</sp-radio-group>
```

#### Logs taken while reproducing problem

radio-group.ts:46 Uncaught DOMException: Failed to execute 'querySelector' on 'Element': 'sp-radio[value=16]' is not a valid selector.
    at HTMLElement.set selected [as selected] (https://localhost:8080/node_modules/@spectrum-web-components/radio-group/lib/radio-group.js:34:20)

The error boils down to the following code, where sp-radio[value=xxxx] is query selected.

```
    public set selected(value: string) {
        const radio = value
            ? (this.querySelector(`sp-radio[value=${value}]`) as Radio)
            : undefined;
```

In my local environment, wrapping in quotes seems to fix the issue, but I don't know if there are 
larger implications to doing so. I'm about to extend to fix it on my end so I can continue on with my work:

`? (this.querySelector(`sp-radio[value="${value}]"`) as Radio)`

Westbrook:
@bengfarrell nice catch, that's a tricky situation! I don't see any way that would break anything from the outside. It also adds some added stability in the case that someone sets a value of a string with spaces in it `radio value`, etc.

Would you be able to/willing to submit a PR with that change?

bengfarrell:
Sure thing, done!

Westbrook:
fixed by #378

PR #375, opened 2019-12-14
Fix for Radio Button Group breaks when radio value is an integer Issue #364

bengfarrell:
https://github.com/adobe/spectrum-web-components/issues/364

I've tested this fix by extending the Radio-Group component in my project, and don't see anything bad. However, in this pull request, I'm not testing because I don't have the local environment set up

## Description

Simply wrapped the offending selector in quotes.

## Related Issue

Radio Button Group breaks when radio value is an integer Issue #364
https://github.com/adobe/spectrum-web-components/issues/364

## Motivation and Context

Radio Button Groups could not contain radio buttons with a value of type integer previously. The Radio Group, when set selected would throw an error due to query selection

## How Has This Been Tested?

Have not tested this fix at all, other than extending Radio Button Group on my project to patch this issue temporarily

## Types of changes

## Checklist:

### DIFF
--- a/packages/radio-group/src/radio-group.ts
+++ b/packages/radio-group/src/radio-group.ts
@@ -43,7 +43,7 @@ export class RadioGroup extends LitElement {
 
     public set selected(value: string) {
         const radio = value
-            ? (this.querySelector(`sp-radio[value=${value}]`) as Radio)
+            ? (this.querySelector(`sp-radio[value="${value}"]`) as Radio)
             : undefined;
 
         this.deselectChecked();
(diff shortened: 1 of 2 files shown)


Westbrook:
@bengfarrell Thanks for putting this together! If you could just do one more thing here, I look forward to approving this!

If you could update the code found https://github.com/adobe/spectrum-web-components/blob/master/packages/radio-group/test/radio-group.test.ts#L55-L59 to include number based `value` and `selected` attributes, and then pair that with updates to https://github.com/adobe/spectrum-web-components/blob/master/packages/radio-group/test/radio-group.test.ts#L198 that match those values so that we are protected from future regressions in this area, I'd be much obliged.

bengfarrell:
OK, updated as per @Westbrook. As I state in my commit, unfortunately, I'd have to spend a fair bit of time figuring out why things aren't building for me, and why my tests seem to be running silently (not breaking when I intend them to), so I haven't validated my new tests. However, I added a new test with a new block of HTML. The new HTML simply includes radio buttons with integer values, and the group selects one of them. Without my change, this test would produce a runtime error -  so I assume it's good, but I am flying blind here

PR #377, opened 2019-12-16
docs: install instructions

Westbrook:
## Description
Expand on the README.md documentation to outline installation instructions.

## Motivation and Context
Ongoing prep for making public.

image

## Types of changes
- [x] Documentation

## Checklist:

### DIFF
--- a/packages/action-menu/README.md
+++ b/packages/action-menu/README.md
@@ -2,6 +2,16 @@
 
 An `<sp-action-menu />` is simply an action button with a Popover. Use an `<sp-menu>` element to outline the items that will be made available to the user when interacting with the `sp-action-menu` element.
 
+### Installation
+
+```
+npm install @spectrum-web-components/action-menu
+
+# or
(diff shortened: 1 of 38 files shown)


Westbrook:
I'll need to do a once over to make sure that all these are correct. I named them based on the folder names, but I'm not 100% sure that all our packages are exported under the same names at the folders they are in. Whether that's something we should fix in this PR or not is a different issue.

andrewhatran:
Swept through the `spectrum-web-component` package names in artifactory and looks like the names are alignment 👍 

I believe that the package names align with the folder names for all components? can't find any exceptions so far

Westbrook:
@andrewhatran Thanks for double-checking this with me. I agree that we're in a good place on this! We'll be editing some of the UI for code delivery as part of our review with the Spectrum team for release, but I think we should move this forward to support users now. Let me know if you have any changes you'd like to see.

Issue #71, opened 2019-06-19
Css variables should be displayed in the documentation of all components


andrewhatran:
### Expected Behavior
The documentation of our components should also display the css-variables used in that component's styling. This is so users of the components can easily restyle the component.

web-component-analyzer comes with css-variable extraction built in.

### Actual Behavior
css-variables are currently not presented in the documentation

Westbrook:
Requirements for a final solution:
- dynamically generated from the Spectrum CSS (and/or Spectrum DNA) input
- doesn't negatively effect readability or usability of the component TS files
- attempts to "resolve" the CSS Custom Properties to a human readable value

I've created https://github.com/adobe/spectrum-css/issues/459 to see if the Spectrum CSS team is interested in more in depth support of this sort of offering at their level as well.

PR #379, opened 2019-12-16
Work on CSS Vars Docs

### LINKED ISSUE #316: 'docs:start' doesn't analyze the components like 'docs:build'
When running the docs for the first time, the API tabs should be available across all components. `yarn docs:build` does this analysis, but `yarn docs:start` does not. Both commands should perform similarly.

Westbrook:
## Description
- Updates web-component-analyzer to `@1.0.0`
- Outlines what it _could_ look like to add CSS Custom Property docs in JSDocs
- Wants input because beyond the included elements in the PR the amount of custom properties gets CRAZY pant.
- I'm gonna see if the analyzer will still work if we move the JSDocs block to the end of the file, or other files, or whatnot. If people have ideas on tooling this issue away, I'm all ears.
- also, most of this points to Spectrum CSS custom properties, and not to actual values, would we expect to see a glossary of those values somewhere in our docs, or possibly theirs in response to this addition?

## Related Issue
Fixes #361 
Refs #71
Fixes #316 

## Motivation and Context
Better documentation!

## How Has This Been Tested?
I've see it in the docs...

image
image

## Types of changes
- [x] Docs

## Checklist:


tasks/documentation.js
```diff
@@ -25,23 +25,6 @@ const srcPath = path.join(projectDir, 'src');
 
 const BASE_URL = 'https://opensource.adobe.com/spectrum-web-components/';
 
-const extractComponentDocumentation = () => {
-    return exec(
-        `node "${path.join(
-            projectDir,
-            'documentation/scripts/extractComponentAPIDocs.js'
```
Westbrook: The work that was being done in `documentation/scripts/extractComponentAPIDocs.js` is no longer compatible with the most recent version of `web-component-analyzer`, so I took it out of the workflow and replaced it with the component based JSON interpolation above.

documentation/src/components/component.ts
```diff
@@ -14,15 +14,79 @@ import { LayoutElement } from './layout';
...
+    if (a.name > b.name) {
+        return 1;
+    }
+    if (b.name > a.name) {
+        return -1;
+    }
+    return 0;
+}
+
+function buildTable(
```
Westbrook: This replaces the pre-processing of the API Documentation that had previously been done. This approach also saves us ~15KB of payload, which is nice.

documentation/src/components/markdown.css
```diff
@@ -38,3 +38,7 @@ hr.spectrum-Rule {
 article {
     margin-bottom: 3em;
 }
+
+.spectrum-Table {
```
Westbrook: Unifies the table delivery.

documentation/src/components/side-nav.css
```diff
@@ -14,11 +14,14 @@ governing permissions and limitations under the License.
     background-color: var(--spectrum-global-color-gray-100);
 }
 
+aside {
```
Westbrook: Discovered that the responsive change I made prevented the navigation content from scrolling...

package.json
```diff
@@ -43,9 +43,10 @@
         "storybook": "run-p watch build:watch storybook:start",
         "storybook:start": "start-storybook -p 6006",
         "storybook:build": "build-storybook -c .storybook -o ./documentation/dist/storybook",
+        "docs:analyze": "wca analyze 'packages/*/src/**/*.ts' --format json --outFile documentation/custom-elements.json",
         "docs:build": "yarn build && yarn docs:ci",
-        "docs:ci": "gulp docsCompile && yarn storybook:build",
-        "docs:start": "gulp docsWatchCompile",
+        "docs:ci": "yarn docs:analyze && gulp docsCompile && yarn storybook:build",
```
Westbrook: This is one time, so right now you have to manually run analyze for updates...

wca doesn't support a `live` mode, but we could work to add this via scripting on our end if people felt strongly about it.

benjamind:
In general this PR looks great. With regard to how we'll support and maintain custom css properties documentation going forward, would we not be better off with a postcss plugin?

After all we're running all our CSS files through PostCSS, so could we not use something like this: https://github.com/postcss/postcss-custom-properties to export the variables we find to a JSON file, then include this into our custom elements documentation somehow?

Given the huge array of components and variables we're ultimately looking at, and that the docs for the variables are not in the same file where the variables are declared right now, I think putting the time in on an automated solution might be worthwhile?

adixon-adobe:
Definitely agree that automating this is gonna be nice. 

My imperfect understanding is that most/all of these *should* get driven from spectrum-dna. In addition to #71 I'd find this useful if I was auditing values between DNA/CSS/WC. So to that end, if you can output both the Spectrum CSS name & value, I think it'll be more useful--at least from your screenshot, it looks like you're just outputting the name. Regardless, this is a great step towards that.

Westbrook:
💯 we can't do this long term without doing it dynamically somehow! This initial look is to show how CSS Custom Properties look in documentation by default and how they need to look in the codebase to generate those docs. My first look at it is that I'm not 100% sure it's worth the trouble if the JSDoc for the properties HAS to live at the top of the JS file. Even here with ~12 properties listed, this starts to push the JS down a good bit, with the (checks notes) 312 properties outlined in `sp-button` that location becomes not acceptable.

In that way, what are our thoughts on moving forward with this PR and then making an issue on finding the right dynamic solution?

I like that plugin that Ben shared it should be super useful, if nothing else to string pipe the CSS Custom Props and JS docs into a single pass on WCA. I think getting a Phase 1 where some version of that happens is a good follow up, and then separately we can look into chasing the CSS Custom Props to static values. I'm a little worried that the chain of custody on some of those could be _very_ long, everything would be at least 3 steps by default (component props -> spectrum prop -> value), but there will be some with more...

In the interim, I'll do some more testing of JSDocs location in a file/directory, etc and report back.

Westbrook:
Trying to capture the longer term feedback in these review [here](https://github.com/adobe/spectrum-web-components/issues/71#issuecomment-570703623) so that it can be separate from the idea of moving this specific PR forward, if possible. To support that, I'd also be amenable to removing some of the added documentation in order to catch up master with this newo version of `web-component-analyzer` and the build process for the docs site in advance of the CSS documentation work settling a bit.

documentation/src/components/component.ts
```diff
@@ -118,11 +175,112 @@ class ComponentElement extends LayoutElement {
                         : html``}
                     ${this.tab === TabValue.Examples
                         ? ComponentDocs.get(this.location.params.component)
-                        : ComponentApiDocs.get(this.location.params.component)}
+                        : this.renderDocs()}
```
benjamind: Since we're finding the docs in the call above, could we not pass them into the renderDocs function directly and save the additional find call below?

benjamind:
I think we should probably just bite the bullet here and go straight to automation, since we're now generating the API docs table at runtime through the `custom-elements.json`, can we note just generate a `css-variables.json` using the postcss plugin, and then generate the docs for the css variables in a similar way?

Westbrook:
In the context of the docs site, that would certainly "get it done", and being we bundle everything really heavily in the build it wouldn't create an additional request or anything. What it would require is a second set of scripting to do a pass through the CSS files and extended templating to render that to the page. Do we think it would be better to take two passes through the CSS (the standard pass to generate the `*.css.js` files and one to generate the `css-variables.json` later) or should we do the work to have the happen at the same time?

Also, while this will work explicitly for the Docs Site, we will need to get this information into the `custom-elements.json` pipeline at some point as that's the build artifact that powers the editor/tools benefits of having this documented for implementing users. See https://github.com/adobe/spectrum-web-components/blob/master/package.json#L58 where the same tooling creates individual `custom-elements.json` files for each published package. It seems like the two problems can be solved together if we focus on making the process automated separate of the PR, but I'm OK with either way.

benjamind:
I'd be happy with a second pass that runs after the `custom-elements-json` step which re-parses the spectrum-css, extracts the css variables and appends them it into the generated `custom-elements.json` for each package.

This way any css variables defined in the component itself will be parsed and added to the json via the lit-analyzer step, while any variables inherited from spectrum-css will be added by our processor?

benjamind:
Quickly threw this branch together which extracts from the css the used variables:

https://github.com/adobe/spectrum-web-components/tree/extract-css-vars

This will output a `.json` alongside every processed css file that contains the used variables, e.g. for `avatar.css`:

```json
{
    "--spectrum-avatar-small-width": "var(--spectrum-global-dimension-size-200)",
    "--spectrum-avatar-small-height": "var(--spectrum-global-dimension-size-200)",
    "--spectrum-avatar-small-border-radius": "50%",
    "--spectrum-avatar-small-opacity": "1",
    "--spectrum-avatar-small-opacity-disabled": ".3"
}
```

This seems like a good starting point. Next steps would be to associate these json files with the actual css used for a specific component, not exactly sure where we can do that. From there it shouldn't be too hard to add the variables to the `custom-elements.json`.

Westbrook:
@benjamind I think you've got a great idea there. Closing this PR so that we can address updating WCA separately of addressing the way we process CSS Custom Properties is the right way to go. I've moved the WCA update to #419 which should be pretty quick to get out of the way as everyone seems good on the actual update changes that were previously here. I'll check out your experimental branch and see what it looks like to run it along with the newer version of the docs. Thanks!

PR #380, opened 2019-12-16
feat(sidenav): add keyboard accessibility

### LINKED ISSUE #281: `sp-sidenav` doesn't implement focus context from Spectrum CSS
### Expected Behaviour
Component is implemented based on the full Spectrum CSS import.

### Actual Behaviour
The following classes aren't processed:
- `.spectrum-SideNav-itemIcon`

The `:focus` context isn't managed correctly in the CSS/`render()` method.

- visit the stories for the component in Storybook
- attempt to focus on any of the `sp-sidenav` implementations.

### LINKED ISSUE #329: keyboard focus issues on website
### Expected Behaviour
Keyboard focus should move to all actionable components and be visible when it gets there

### Actual Behaviour
Focus jumps directly from the "spectrum web components" link at top left to the first interactive item in the page. It skips the sidenav and the tabs completely

Go to https://opensource.adobe.com/spectrum-web-components/ and press tab a few times

#### Steps to Reproduce
Go to https://opensource.adobe.com/spectrum-web-components/ and press tab a few times

#### Platform and Version
Chrome  78.0.3904.97
Mac 10.15.1

Westbrook:
## Description 
Expand on a11y support for the sidenav pattern.

- I changed the default value for `expanded` to `false` which changes the API a little but prevents of edge case bugs long term. Let me know if we think this might not be the right thing...

## Related Issue
fixes #329
fixes #281

## Motivation and Context
Everything should be accessible.

## How Has This Been Tested?
- unit tests
- manually

## Types of changes

## Checklist:


packages/sidenav/src/sidenav-item.ts
```diff
@@ -38,7 +39,7 @@ export class SideNavItem extends LitElement {
     public disabled = false;
 
     @property({ type: Boolean, reflect: true })
-    public expanded = true;
+    public expanded = false;
```
Westbrook: This can create some issues when defaulted to `true` if anyone reading this knows of users to whom I might need point this out, let me know. It's a "feature" change, which will amount to a "breaking" change being we're still pre-1.0.0, so I don't expect there to be issues.

packages/sidenav/src/sidenav-item.ts
```diff
@@ -82,7 +84,10 @@ export class SideNavItem extends LitElement {
         this.selected = event.target === this;
     }
 
-    protected handleClick(): void {
+    protected handleClick(event?: Event): void {
+        if (!this.href && event) {
```
Westbrook: Allows the use of native `Enter` events on the link, without actual linking.

packages/sidenav/src/sidenav-item.ts
```diff
@@ -106,10 +111,19 @@ export class SideNavItem extends LitElement {
         this.handleClick();
     }
 
+    public get focusElement(): HTMLElement {
+        /* istanbul ignore if */
+        if (!this.shadowRoot) {
+            return this;
+        }
+        return this.shadowRoot.querySelector('#itemLink') as HTMLElement;
```
Westbrook: We should probably do the work to update all instances of this pattern to `@query` decorators in the future, for that tiny little perf bump. 🌟

packages/sidenav/src/sidenav.ts
```diff
@@ -36,6 +39,88 @@ export class SideNav extends LitElement {
...
+        this.startListeningToKeyboard();
+        if (event.target !== this) {
+            return;
+        }
+        const selected = this.querySelector('[selected]') as SideNavItem;
+        if (selected && !this.isDisabledChild(selected)) {
+            selected.focus();
+            return;
+        }
+        const items = [...this.querySelectorAll('sp-sidenav-item')];
```
Westbrook: I do this a couple of times across the element, not sure if it's worth holding the PR for centralizing on a lifecycle point to cache this value...

packages/sidenav/src/sidenav.ts
```diff
@@ -36,6 +39,88 @@ export class SideNav extends LitElement {
...
+        const focused = items.indexOf(document.activeElement as SideNavItem);
+        let next = focused;
+        next = (items.length + next + direction) % items.length;
+        while (this.isDisabledChild(items[next])) {
+            next = (items.length + next + direction) % items.length;
+        }
+        items[next].focus();
+    }
+
+    private isDisabledChild(child: SideNavItem): boolean {
```
Westbrook: A child is disabled if `this.disabled`, a parent is `parent.disabled`, or a parent is `!parent.expanded`.

packages/sidenav/src/sidenav.ts
```diff
@@ -45,6 +130,11 @@ export class SideNav extends LitElement {
             </nav>
         `;
     }
+
+    protected firstUpdated(changes: PropertyValues): void {
+        super.firstUpdated(changes);
+        this.tabIndex = 0;
```
Westbrook: I didn't make this element an extension of `Focusable` at this time. It sort of is, but sort of isn't, and think reviewing this decision later makes sense.

packages/sidenav/src/sidenav.ts
benjamind: Can we not just use the `queryAll` decorator?

https://lit-element.polymer-project.org/api/modules/_lib_decorators_.html#queryall

packages/sidenav/src/sidenav.ts
benjamind: Why do we search up through the dom to determine disabled state instead of propagate disabled state down?

Westbrook: This maybe less clear from the outside because `disabled` isn't the only thing that's being checked here. It's also checking is a parent is `expanded`, that way maybe this should be `isAccessibleChild` or `isFocusableChild`?  Because `parent` can be a number of different element types and they all exist in a light DOM series none of them manages their children at current. We could acquire the slotted elements, though we might actually need a deep mutation observer for the reality that there can be intermediary `sp-sidenav-heading` elements, and work with altering the state of items imperatively, but that seemed like a lot of prep work that might not be needed, whereas this process this reality lazily. Along with that, if parents contextually managed the `disabled` state of an item, then we'd also have to manage a cache of the "expected" state to return to, which seemed like a lot of extraneous bookkeeping. Let me know if you have other thoughts that could shift the weight of one side of the balance.

packages/sidenav/src/sidenav.ts
Westbrook: That's what I mean. I didn't realize there were so many places that used this data while deving. I'll get into adding this!

Westbrook: Actually, that's not what I meant being the `queryAll` decorator only works for the `renderRoot` and not for the light DOM. Want me to get the similar work for the light DOM element management into this PR?

benjamind: Ah good point, fair point, I'd leave it in this case then.

andrewhatran:
This feels really great 🎉

I was wondering if we wanted to afford both `arrow key` navigation as well as `tab` navigation to move focus between the `sidenav-items`

Was looking at the [React Implementation](https://react-spectrum.corp.adobe.com/components/SideNav) which supports both

Westbrook:
@andrewhatran That's interesting. RS is offering a `manageTabIndex: true` option for the difference between the two.

@jnurthen do you have thoughts on this? Most docs I've seen outline that an interface like this should have only the floating focus element, and not `tabindex=0` on EVERYTHING. Yay, aria docs, they're _great_.

jnurthen:
@Westbrook it depends a lot on usage what you want so I support the react-spectrum implementation. 
Tab vs Arrow navigation is a trade-off of discoverability vs efficiency so when to use one or the other depends a lot on how often a visitor is expected to use the site/application AND how many items there are in the sidenav

jnurthen:
@Westbrook we don't have a sidenav in the ARIA Authoring Practices so am interested to know which docs you are referring to

Westbrook:
@jnurthen I'm using an intersection of tab, menu, and listbox (which are kind of like the parents of the sidenav pattern, structurally). Of all of those, only w3.org/TR/wai-aria-practices-1.1/examples/disclosure/disclosure-navigation.html allows you to tab through lists of "options" like this, and I can't really see any reasoning behind it. From that standpoint it would seem that _all_ (or almost all) navigation contexts featuring a list of links/options/etc should have only arrow key interactions. Making arrow key interactions not the default seems like setting users up for failure in their accessibility goals.

jnurthen:
Interaction mechanisms depend most on the role(s) you add as that is communicated to an screen reader user. tabs, menus and listboxes are single tab-stop components so this would be what you find for those.
However, a sidenav is none of those and the interaction mechanism you choose depends on the roles you add.

- In react-spectrum when it is a single-level it defaults to manageTabIndex=false and we present a list of links to the user where the user can tab through each one. 
- A Multi-level sidenav is a tree which I believe only supports manageTabIndex=true and it is a single tab-stop as expected.

Now - I agree that manageTabindex=true with a single-level sidenav is potentially problematic. IMO it would be better if react-spectrum presented this as a different role (which one I'm not sure without some discussion with my colleagues). I'll log this and hopefully we can tackle it in v3.

Westbrook:
@andrewhatran now, we have both!

PR #383, opened 2019-12-16
feat(slider): adds getAriaValueText property, fixes #381

### LINKED ISSUE #381: sp-slider should provide text value customization
### Expected Behaviour
We should be able to specify how the text value in the slider is visually represented, separately from the value of the slider. In React-Spectrum this is implement as a `getAriaTextValue` function property.

### Actual Behaviour
Cannot specify how to present the value to the user, so percentage sliders have to be 0-100 valued, and have no % suffix.

Any slider that wants to display non-floating point output.

#### Steps to Reproduce
Use a slider.

#### Platform and Version
All

benjamind:
## Description

Adds a new property `getAriaValueText` which accepts a function converting a value number argument into a string to be displayed.

## Related Issue

#381 

## Motivation and Context

Need to be able to control how the value of the slider is presented, both in aria presentation and in our visual presentation above the slider.

## How Has This Been Tested?

* Unit testing has been updated to test that the aria text value is updated correctly.

image

## Types of changes

## Checklist:

### DIFF
--- a/__snapshots__/Slider.md
+++ b/__snapshots__/Slider.md
@@ -5,12 +5,9 @@
 ```html
 <div id="labelContainer">
     <label for="input" id="label"></label>
-    <div
-        aria-labelledby="label"
-        aria-readonly="true"
-        id="value"
-        role="textbox"
-    ></div>
+    <div aria-labelledby="label" aria-readonly="true" id="value" role="textbox">
+        10
+    </div>
 </div>
 <div id="controls">
     <div
@@ -25,7 +22,8 @@
             aria-label=""
             aria-valuemax="20"
             aria-valuemin="0"
-            aria-valuetext=""
+            aria-valuenow="10"
+            aria-valuetext="10"
             id="input"
             max="20"
             min="0"
@@ -48,12 +46,9 @@
 ```html
 <div id="labelContainer">
     <label for="input" id="label"></label>
-    <div
-        aria-labelledby="label"
-        aria-readonly="true"
-        id="value"
-        role="textbox"
-    ></div>
+    <div aria-labelledby="label" aria-readonly="true" id="value" role="textbox">
+        10
+    </div>
 </div>
 <div id="controls">
     <div class="track"></div>
@@ -63,7 +58,8 @@
             aria-label=""
             aria-valuemax="20"
             aria-valuemin="0"
-            aria-valuetext=""
+            aria-valuenow="10"
+            aria-valuetext="10"
             id="input"
             max="20"
             min="0"
@@ -1104,6 +1100,7 @@
             aria-label=""
             aria-valuemax="100"
             aria-valuemin="-100"
+            aria-valuenow="0"
             aria-valuetext="0"
             id="input"
             max="100"
--- a/__snapshots__/Slider.md
+++ b/__snapshots__/Slider.md
@@ -5,9 +5,12 @@
 ```html
 <div id="labelContainer">
     <label for="input" id="label"></label>
-    <div aria-labelledby="label" aria-readonly="true" id="value" role="textbox">
-        10
-    </div>
+    <div
+        aria-labelledby="label"
+        aria-readonly="true"
+        id="value"
+        role="textbox"
+    ></div>
 </div>
 <div id="controls">
     <div
@@ -22,7 +25,7 @@
             aria-label=""
             aria-valuemax="20"
             aria-valuemin="0"
-            aria-valuetext="10"
+            aria-valuetext=""
             id="input"
             max="20"
             min="0"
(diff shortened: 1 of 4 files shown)


packages/slider/src/slider.ts
```diff
@@ -74,6 +75,12 @@ export class Slider extends Focusable {
     /* Ensure that a '' value for `variant` removes the attribute instead of a blank value */
     private _variant = '';
 
+    @property({ attribute: false })
+    public getAriaValueText: (value: number) => string = (value) => `${value}`;
+
+    @property({ reflect: true, attribute: 'aria-valuetext' })
```
Westbrook: I'm fairly certain that this doesn't "reflect" to the parent. You referenced the React version, so I was looking there and `aria-valuetext` is only applied to the `input` there.

__snapshots__/Slider.md
```diff
@@ -5,9 +5,12 @@
...
     <label for="input" id="label"></label>
-    <div aria-labelledby="label" aria-readonly="true" id="value" role="textbox">
-        10
-    </div>
+    <div
+        aria-labelledby="label"
+        aria-readonly="true"
+        id="value"
+        role="textbox"
+    ></div>
```
Westbrook: This points out that is the slider uses the default `10` for a value that the `ariaValueText` prop won't be set at that point. I think being we don't really want `ariaValueText` to reflect that we should make it a programatic getter: ```
get ariaValueText() {
  return this.getAriaValueText(this.value);
}
``` Which would also ensure that the value is always correct without having to be a part of the value setter.

File: packages/slider/src/slider.ts [hunk 3]
```diff
@@ -74,6 +75,12 @@ export class Slider extends Focusable {
     /* Ensure that a '' value for `variant` removes the attribute instead of a blank value */
     private _variant = '';
 
+    @property({ attribute: false })
+    public getAriaValueText: (value: number) => string = (value) => `${value}`;
```
Westbrook: TS covers us here, I guess, but do we want to do anything about this getting set badly? A fallback to default or something?

packages/slider/src/slider.ts
benjamind: The reflect on this property does get applied to the host. 
image

So I've not actually been following the aria stuff too much yet wrt WCs. Do we need the aria attributes just on the active focused input, or also on the slider itself?

Westbrook: It's likely that @jnurthen has some added insight here as I've found even "clear" documentation leaves things to be desired. However, reviewing the MDN page for this attribute, there are a couple of things that look likely to me:

1. We shouldn't append it to any element unless we are actually representing a non-simple number: 
> Authors SHOULD only set the `aria-valuetext` attribute when the rendered value cannot be accurately represented as a number.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-valuetext_attribute

2. We shouldn't be applying `aria-valuetext` multiple times for one "widget': https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_slider_role#Examples

3. We should be applying `aria-valuetext` on the same element as `aria-valuemin`, `aria-valuemax`, and `aria-valuenow`. 

Maybe James can speak to a more comprehensive internal doc for slider based accessibility as similar docs for the Toast, etc patterns found [here](https://github.com/adobe/react-spectrum-v3/tree/master/specs/accessibility) have been very useful.

benjamind: Does the `Authors SHOULD only set the aria-valuetext attribute when the rendered value cannot be accurately represented as a number.` actually exclude adding it even if the rendered value is a number? It would certainly keep the code paths simpler to just always render `aria-valuetext` and just put the number in it.

packages/slider/src/slider.ts
benjamind: Added a basic guard in the new getter version of `ariaValueText` which if `getAriaValueText` is falsy will use the default string. I don't think we need to handle all cases I think, just enough to allow someone to clear their changes is good?

__snapshots__/Slider.md
benjamind: Updated in b49eddf

packages/slider/src/slider.ts
benjamind: Updated in b49eddf, also added `aria-valuenow` to the input for good measure.

jnurthen: @Westbrook I agree with your comments above. Unfortunately, we don't have a spec for slider for v3 yet. The best I can do is point you to the Accessibility section of the v2 documentation at https://react-spectrum.corp.adobe.com/components/Slider

packages/slider/src/slider.ts
```diff
@@ -74,6 +74,17 @@ export class Slider extends Focusable {
...
+
+    @property({ attribute: false })
```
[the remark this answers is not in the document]
benjamind: Yeah the decorator just adds an accessor to the property such that it triggers re-render on change.

PR #391, opened 2019-12-19
fix: factor theme to use a single DOM node

### LINKED ISSUE #154: Can't override `display` property on `sp-theme`
### Expected Behaviour
`<sp-theme style="display: flex;">` actually results in a flex layout

### Actual Behaviour
The nested `<sp-theme-dark>` or `<sp-theme-light>` applies `display: block`, so the `display: flex` doesn't apply to the child contents you specify.

cuberoot:
## Description

Currently themes are implemented as a heirarchy of nodes to get the CSS custom property scoping correct. This change crunches that down into a single node.

The PR also renames the `themes` package to `theme` to match the naming convention in place for the rest of the components.

This PR takes over for #388 

## Related Issue

fixes #154 

## Motivation and Context

This PR removed extra DOM nodes from the shadow DOM that are un-style-able by the user.

## How Has This Been Tested?

I have tested in the docs app and Storybook, as well as using the tests

## Types of changes

## Checklist:


cuberoot:
@Westbrook The above build failures seem to be related to changes to the search field, requiring an updated snapshot. The are not related to my changes. The tests pass locally for me. Any thoughts?

Westbrook:
Yes, I need to go and do the real work of adding to timing to all of the places 🤦‍♂ 🙈

I had been doing it piecemeal as failures came up, see #390

Westbrook:
Tests and CI ran through! Some visual regression questions in there, but those were to be expected.

packages/theme/stories/theme.stories.ts
```diff
@@ -162,65 +162,4 @@ storiesOf('Theme', module)
                 </div>
             </sp-theme>
         `;
-    })
-    .add('Nested color', () => {
```
Westbrook: Can you talk about why you chose to remove this story?

__snapshots__/Medium.md
```diff
@@ -0,0 +1,7 @@
+# `Medium`
```
Westbrook: These elements moving to `<slot></slot>` only shadow DOM, I wonder if we'd be better off not having the snapshot tests for them?

packages/theme/README.md
```diff
@@ -2,8 +2,8 @@
 
 An **sp-theme** sets the rendering theme for all child components.
 The Spectrum design system supports four color themes and two different
-scales. `spectrum-web-components` currently supports two of the four
-color themes (dark and light) one one of the scales (medium).
+scales. `spectrum-web-components` currently supports three of the four
+color themes (dark, light and lightest) and one one of the scales (medium).
```
Westbrook: 👍

packages/theme/stories/theme.stories.ts
cuberoot: The whole story is there to test nesting an `<sp-theme-lightest>` which doesn't exist anymore. There's already another story for nesting an `<sp-theme>`

cuberoot: I will put a reversed version in... I am not sure that there is a better alternative to using the stories.

cuberoot:
The original issue was around `display: flex` (see #154). I added a story to test that `display: flex` works on an `sp-theme`. The visual tests should now cover that. We don't have to say anything because it now works with themes the same way it works on any other node.

I don't know much about how lerna does package releases. I'm included to just let it be `1.1` unless you really want to practice on it. I'm hoping we don't have to rename packages after v1.0.

packages/theme/package.json
```diff
@@ -1,17 +1,17 @@
...
+        "directory": "packages/theme"
     },
     "bugs": {
         "url": "https://github.com/adobe/spectrum-web-components/issues"
     },
-    "homepage": "https://adobe.github.io/spectrum-web-components/components/themes",
+    "homepage": "https://adobe.github.io/spectrum-web-components/components/theme",
     "keywords": [
         "spectrum css",
         "web components",
```
Westbrook: Agree that letting this next version be 0.1.1 makes sense. To get there, set line 23 to "0.1.0", and then when we do `yarn lerna-publish` it'll manage the `fix` alteration outlined in the commits on its own. After that's there, I'm excited to approve/merge this!

cuberoot: Done. Version modified!

PR #402, opened 2020-01-02
Remove requirement of `<overlay-root>` element.

### LINKED ISSUE #151: Eliminate need for overlay-root
### Expected Behaviour

We should be able to use overlays (e.g. dropdown menus and popovers) without having to create a `overlay-root`

### Actual Behaviour

Right now, we have to manually create an `overlay-root` element to host our popovers

Any use of `overlay-trigger` or any component that uses it.

#### Steps to Reproduce

Run the `overlay-trigger` samples in Storybook

Westbrook:
## Description
- Move support of overlay content at the `<body>` tag, as opposed to the `overlay-root` element.
- Acquire style info from nearest ancestor `sp-theme` element
- Add some tests and demos
- Merge the `overlay-root` and `overlay-trigger` packages to just `overlay`

## Related Issue
Fixes #151

## Motivation and Context
Elements should generally be "self-contained" and the requirement of an `overlay-root` is confusing and cumbersome. 

Currently, the root is now being set to `body` but default. However, we could easily add configuration to the `OverlayRoot` constructor in the future to customize the DOM element on which this occurs.

## How Has This Been Tested?
- passes old test
- passes some new tests

## Types of changes

## Checklist:


Westbrook:
@cuberoot I'm working through some repo level issues with the screenshots (#403) and the test starting (no solid leads yet) but otherwise, I like how this has come together. It's so much easier with your updates to `sp-theme`!

Please take a look.

karma.conf.js
```diff
@@ -40,8 +40,8 @@ module.exports = (config) => {
             coverageIstanbulReporter: {
                 thresholds: {
                     global: {
-                        statements: 95,
-                        branches: 86,
+                        statements: 96,
+                        branches: 89,
```
Westbrook: Mainly from if/else ignores, but it is nice to see this number slowly trend upwards...

packages/overlay/package.json
```diff
@@ -1,24 +1,24 @@
...
-    "homepage": "https://adobe.github.io/spectrum-web-components/components/overlay-root",
+    "homepage": "https://adobe.github.io/spectrum-web-components/components/overlay",
     "keywords": [
         "spectrum css",
         "web components",
         "lit-element",
         "lit-html"
     ],
-    "version": "0.1.7",
+    "version": "0.0.1",
```
Westbrook: The `feat(overlay): ...` commit will bump this to `0.1.0` when we release.

packages/overlay/src/active-overlay.ts
```diff
@@ -133,6 +141,12 @@ export class ActiveOverlay extends LitElement {
 
     @property({ reflect: true })
     public placement: Placement = 'bottom';
+    @property({ attribute: false })
+    public color: Color = DefaultColor;
+    @property({ attribute: false })
+    public size: Size = DefaultSize;
+    @property({ type: Boolean, attribute: false })
+    public hasTheme = false;
```
Westbrook: The `hasTheme` property allows me to be more strict about the typing of `size` and `color` and prevents the need to recast those when putting them into the `<sp-theme>` element later. I could be convinced that it's better to lean more heavily on the typing, however.

packages/overlay/src/active-overlay.ts
```diff
@@ -289,10 +313,20 @@ export class ActiveOverlay extends LitElement {
         this.schedulePositionUpdate();
     }
 
-    public render(): TemplateResult {
+    public renderTheme(content: TemplateResult): TemplateResult {
+        import('@spectrum-web-components/theme');
```
Westbrook: Don't force non-`sp-theme` users to download this code if they aren't actually using it.

packages/overlay/src/overlay-root.ts
```diff
@@ -0,0 +1,69 @@
...
+        for (const overlay of activeOverlays) {
+            if (overlay.parentElement !== document.body) {
+                overlay.setAttribute('slot', 'overlays');
+                document.body.append(overlay);
+            }
+        }
+    };
+
+    public listen(): void {
+        document.body.addEventListener(
```
Westbrook: Always listen on `document` but that could be customized later if we need subwindow centric overlays.

packages/overlay/src/overlay-stack.ts
```diff
@@ -46,39 +46,24 @@ export class OverlayStack {
...
-
     private addEventListeners(): void {
-        if (!this.document) return;
         this.document.addEventListener('click', this.handleMouseCapture, true);
         this.document.addEventListener('click', this.handleMouse);
         this.document.addEventListener('keyup', this.handleKeyUp);
         window.addEventListener('resize', this.handleResize);
     }
 
-    private removeEventListeners(): void {
```
Westbrook: There's no longer a custom element with `disconnectedCallback` methods to connect this functionality to...

packages/overlay/src/overlay-trigger.ts
```diff
@@ -55,25 +61,42 @@ export class OverlayTrigger extends LitElement {
...
 
+        /* istanbul ignore if */
         if (!overlayElement) {
             return;
         }
+        if (!overlayRoot) {
+            overlayRoot = new OverlayRoot();
+        }
+
+        const delayAttribute = overlayElement.getAttribute('delay');
```
Westbrook: Code reorder for simplicity.

packages/overlay/src/overlay-trigger.ts
```diff
@@ -55,25 +61,42 @@ export class OverlayTrigger extends LitElement {
...
+        const delay = delayAttribute ? parseFloat(delayAttribute) : 0;
+
+        const queryThemeDetail: ThemeData = {
```
Westbrook: Manage the nearest ancestor theming...allows us to treat that ancestor like the `overlay-root` (sort of). It's not 100% the same functionality, but I think it's more understandable to say "the closest `sp-theme` manages your styles" rather than "the closest `sp-theme` to the closest `overlay-root` manages your styles".

packages/overlay/src/overlay-trigger.ts
```diff
@@ -55,25 +61,42 @@ export class OverlayTrigger extends LitElement {
...
+            color: undefined,
+            size: undefined,
+        };
+        const queryThemeEvent = new CustomEvent<ThemeData>('query-theme', {
+            bubbles: true,
+            composed: true,
+            detail: queryThemeDetail,
+            cancelable: true,
```
Westbrook: I initially set this as `cancelable` so that it could merge with the `overlayOpenEvent` below and make a single traversal, but I ended up not wanting to mess with the data structure there...I could see merging them again, or moving to a `stopPropagation()` approach instead...thoughts?

packages/overlay/src/overlay-trigger.ts
```diff
@@ -109,21 +133,25 @@ export class OverlayTrigger extends LitElement {
         );
 
         this.dispatchEvent(overlayCloseEvent);
+        // overlayRoot.onOverlayClose(overlayCloseEvent);
```
Westbrook: Woops, will review whether this should come back or be removed.

packages/overlay/src/overlay-trigger.ts
```diff
@@ -153,24 +181,28 @@ export class OverlayTrigger extends LitElement {
     }
 
     private onClickSlotChange(event: Event): void {
-        if (event.target) {
-            const slot = event.target as HTMLSlotElement;
-            const content = this.extractSlotContent(slot);
+        /* istanbul ignore if */
+        if (!event.target) {
```
Westbrook: Code re-org for simplicity.

packages/overlay/test/overlay-trigger.test.ts
```diff
@@ -381,4 +370,48 @@ describe('Overlays', () => {
 
         expect(isVisible(hoverContent)).to.be.false;
     });
+    it('acquires a `color` and `size` from `sp-theme`', async () => {
+        const el = await fixture<HTMLDivElement>(html`
+            <sp-theme color="dark">
+                <sp-theme color="light">
```
Westbrook: Make sure the nearest neighbor has precedence on theming.

adixon-adobe:
Overall this looks good to me. Did a little testing and all looks good. I'll want @cuberoot to take a look. I don't think the switch to <body> will affect us (aside from fixing the breaking change in some existing code).

Westbrook:
Cool! He's back next week?

That hope is for anyone who's already using the overlay system to only have to remove `<overlay-root>` elements and update `import` statements. If it appears that more than that would be needed, reviewing that in context will be helpful.

adixon-adobe:
Yes -- he'll be back next week. I don't expect us to have any more work than what you mentioned (should be really trivial).

packages/overlay/src/overlay-trigger.ts
cuberoot: I cannot remember why that was needed.

Westbrook:
@cuberoot updated. PTAL!

cuberoot:
@Westbrook You have some conflicts. I'll approve once those are sorted.

Issue #314, opened 2019-11-26
We should be able to turn off autocomplete for fields


cuberoot:
### Expected Behaviour

We should be able to pass the [autocomplete](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete) attribute to text fields to control completion.

### Actual Behaviour

We cannot create a custom completion because the browser's default autocompletion kicks in

<img width="230" alt="Screen Shot 2019-11-26 at 2 57 28 PM" src="https://user-images.githubusercontent.com/154097/69593151-35de8580-105d-11ea-9a85-0d2cbfd9dbd1.png">

#### Steps to Reproduce

Run the documentation site from #312 and run various searches in Chrome.

Westbrook:
@cuberoot if there were other additions you wanted to make here, feel free to reopen, but it looks like we're getting what we were looking for here in the docs site, at least.

PR #408, opened 2020-01-06
feat: pass through autocomplete attribute to text inputs

cuberoot:
## Description

For text fields, allow the passing through of the [`autocomplete`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete) attribute

## Related Issue

#314 

## Motivation and Context

I need this feature to stop the search field from autocompleting when I am implementing my own autocomplete.

## How Has This Been Tested?

I tested this in my branch where I am adding a search field to the documentation.

## Types of changes

## Checklist:

### DIFF
--- a/packages/textfield/src/textfield.ts
+++ b/packages/textfield/src/textfield.ts
@@ -76,6 +76,11 @@ export class Textfield extends Focusable {
     @property({ type: Boolean, reflect: true })
     public required = false;
 
+    @property({ type: String, reflect: true })
+    public autocomplete?:
+        | HTMLInputElement['autocomplete']
+        | HTMLTextAreaElement['autocomplete'];
+
     public get focusElement(): HTMLInputElement | HTMLTextAreaElement {
         return this.inputElement;
     }
@@ -124,6 +129,7 @@ export class Textfield extends Focusable {
                           <div id="sizer">${this.value}</div>
                       `
                     : nothing}
+                <!-- @ts-ignore -->
                 <textarea
                     aria-label=${this.label || this.placeholder}
                     id="input"
@@ -134,11 +140,13 @@ export class Textfield extends Focusable {
                     @input=${this.onInput}
                     ?disabled=${this.disabled}
                     ?required=${this.required}
+                    autocomplete=${ifDefined(this.autocomplete)}
                 ></textarea>
                 ${this.renderStateIcons()}
             `;
         }
         return html`
+            <!-- @ts-ignore -->
             <input
                 aria-label=${this.label || this.placeholder}
                 id="input"
@@ -149,6 +157,7 @@ export class Textfield extends Focusable {
                 @input=${this.onInput}
                 ?disabled=${this.disabled}
                 ?required=${this.required}
+                autocomplete=${ifDefined(this.autocomplete)}
             />
             ${this.renderStateIcons()}
         `;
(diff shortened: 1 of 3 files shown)


packages/toast/src/toast.ts
```diff
@@ -55,7 +55,6 @@ export class Toast extends LitElement {
     @property({ type: Number })
     public set timeout(timeout: number | null) {
         const hasTimeout = typeof timeout !== null && (timeout as number) > 0;
-        debugger;
```
cuberoot: Found this when debugging. Slipping it into this PR

packages/textfield/src/textfield.ts
```diff
@@ -33,6 +33,70 @@ import { nothing } from 'lit-html';
...
+    | 'tel-extension'
+    | 'tel-local'
+    | 'tel-local-prefix'
+    | 'tel-local-suffix'
+    | 'tel-national'
+    | 'transaction-amount'
+    | 'transaction-currency'
+    | 'url'
+    | 'username'
+    | 'work';
```
cuberoot: This is kind of big. Prettier insists on one item per line 😛

cuberoot:
@Westbrook I got a test failure on the first run related to overlays. Is this related to your recent work: https://1826-158297197-gh.circle-artifacts.com/0/home/circleci/project/test/visual/screenshots-current/overlay--edges-diff.png

packages/textfield/src/textfield.ts
jnurthen: I don't think this is complete. autocomplete can include multiple tokens - for example `autocomplete="section-blue shipping street-address"` is a valid autocomplete value (taken from HTML spec)

cuberoot: Ah, good catch. That wasn't clear from the MDN docs. I will resort to using a plain string. I was working off of the Typescript typings, which may have made the same mistake.

cuberoot:
@jnurthen I fixed the issue with the `autocomplete` attribute. We now just take a string. VSCode complains, but it compiles and passes lint 😄

cuberoot:
I spoke too soon.  The CircleCI Build is failing because it thinks that the `autocomplete` attribute on the `input` must be exactly one of the list of possible values. The spec is quite clear that is not the case though. I will see if I can pound the compiler into compliance.

benjamind:
@cuberoot interesting failure, since the official dom.d.ts seems to not specify the possible values for an input element.

https://github.com/microsoft/TypeScript/blob/master/lib/lib.dom.d.ts#L7276

I wonder if its related to this issue: https://github.com/runem/lit-analyzer/issues/43

You may be better off just disabling ts checking on that line with a comment?

cuberoot:
@benjamind It took a bit of work to figure out how to [disable the error in JSX code](https://github.com/runem/lit-analyzer/issues/1). Waiting to see if that does the trick.

benjamind:
Other approach would be to cast to any and then ignore the es-lint warning?

packages/textfield/src/textfield.ts
```diff
@@ -76,6 +76,9 @@ export class Textfield extends Focusable {
     @property({ type: Boolean, reflect: true })
     public required = false;
 
+    @property({ type: String, reflect: true })
+    public autocomplete?: string;
```
benjamind: For future proofing you could probably type this as:

```
public autocomplete?: HTMLInputElement['autocomplete'] | HTMLTextAreaElement['autocomplete']
```

It appears TS's `dom.d.ts` currently actually has autocomplete as a string, its just `lit-analyzer` which is getting its types from the IDL somehow. I would hope sometime in the future TS provides better types for these.

cuberoot: @benjamind Ok, I am trying out your version. That is a clever trick.

Issue #157, opened 2019-08-23
Investigating tying `sp-dropdown` and `sp-action-menu` into overlay system


Westbrook:
### Expected Behaviour
Content should be available on top of everything!

### Actual Behaviour
Some content can flow over the dropdown/action-menu content.

### Reproduce Scenario (including but not limited to)
Put content after the dropdown or action-menu with a very high z-index...

### Possibilities
This may be manageable by adding a fallback to the `overlay` system that hoists overlay content to the `<body/>` when and `overlay-root` is not available. Further research should go into whether `popover`, which powers these elements, should deliver this feature by default.

Westbrook:
Right now, the element API for this looks like:
```
            <sp-dropdown>
                Select a Country with a very long label, too long in fact
                <sp-menu slot="options" role="listbox">
                    <sp-menu-item>
                        Deselect
                    </sp-menu-item>
                    <sp-menu-item>
                        Select Inverse
                    </sp-menu-item>
                    <sp-menu-item>
                        Feather...
                    </sp-menu-item>
                    <sp-menu-item>
                        Select and Mask...
                    </sp-menu-item>
                    <sp-menu-divider></sp-menu-divider>
                    <sp-menu-item>
                        Save Selection
                    </sp-menu-item>
                    <sp-menu-item disabled>
                        Make Work Path
                    </sp-menu-item>
                </sp-menu>
            </sp-dropdown>
```
In that way, we're not requiring a `items` property at the `<sp-dropdown>` level, because we are slotting the `<sp-menu>` content into that part of the interface. Working with `overlay-trigger` prevents the ability for this to continue as it does not support forwarding content from inside of a slot to `overlay-root` element. 

## Options:

1. Extend `overlay-trigger` to flatten slotted content when forwarding content to `overlay-root`. In this was we'd need to not just grad the root element of the overlay and pass it along, but we'd need to walk the entire tree looking for `<slot>` elements, then query their `assignedNodes()` (note the newest version of Safari has some issues with `assignedElements()`) and send those elements into the overlay recursively until the entire overlay is populated. I think where this creates the most issue is when you close the overlay, as you will need to have noted what content came from where all along the process of opening the overlay so that you can replace it all as part of the `close` process. This could be really interesting and expressive, but likely fairly buggy and possibly really slow. It also opens us to extended issues around CSS Custom Property application that are featured https://github.com/adobe/spectrum-web-components/issues/151#issuecomment-523295170.

[...]

PR #420, opened 2020-01-08
[RFC] feat(testing-library-dom): add esm mirror of testing-library-dom

Westbrook:
# Explainer
I really like the ideas behind the testing philosophy that [`@testing-library/dom`](https://testing-library.com/docs/dom-testing-library/intro) espouses. Beyond its philosophy it also sets the table for a more uniform testing approach by focusing on the content of an element and how that content effects the accessibility of the element. However, the library is distributed with a number of CJS dependencies that prevent it from working with Karma, not to mention the issues of introspecting Shadow DOM. I'm not fully decided myself as to whether the cost outlined in this PR are worth the benefits of being able to use Testing Library generally and hoped to share my thoughts and findings to empower gather ideas in this are from the rest of the team.

## Description
Add an ES Module mirror of `@testing-library/dom` to the project with a _little_ bit of Shadow DOM hacking so that we can leverage the a11y/content centric selectors of the library in our code. I was hoping to be able to avoid majorly monkey patching the actual code here, but one is needed:

```
querySelectorAll: 'querySelectorAllWithShadowDOM',
```

Even with this one change, it feels like we could easily keep up with changes upstream. That being so, I think we'd be in an OK place to take on the publishing responsibilities outline [here](https://github.com/testing-library/dom-testing-library/issues/413#issuecomment-579614867)

The switch from `querySelectorAll` to `querySelectorAllWithShadowDOM` is also trouble. I wanted to adjust `HTMLElement.prototype.querySelectorAll` directly, but it alters element functionality, too. This allows the query to act more like the render does and avoid not slotted light DOM content while being able to pierce shadow roots. Being it's for tests, any performance issues seem like they'd be less devastating.

### Usage
Generally, the issue that `@testing-library/dom` attempts to correct for in the testing lifecycle is the tendency to rely too heavily on implementation details to ensure coverage.

One place where this arises in the use of snapshot testing, while the DOM is one form of output for an element it is subsequently reprocessed at its intersection with CSS in a way that can find it changing more often than a human can process the diffs in a snapshot test. Here I've shown how ensuring that actual content (and not just a specific DOM element) is available can replace this sort of test. Notice the change in the `sp-actionbar` test to no longer rely on snapshot testing and instead query elements by content rather than by structure or selector.

Even when not relying on snapshot testing, it is easy to fall into the trap of relying on DOM implementation testing. See the changes to `sp-sidenav` that updates a recently created test to rely on presence of content as opposed to the details of a template. The value of this may not be immediately clear for this element in particular, but imaging a world where the expanded content needed to be displayed in an overlay, in that way the `queryByText(el, 'Section 1')` calls would only need the `el` to be updated to `document` in order to use the same over structure of test, greatly reducing refactoring costs. This would be very clearly felt via #157 which will require just that sort of change.

## Motivation and Context
While testing a unit of shadow DOM is pretty straight forward the steps required to move towards integrations or e2e testing can be less clear, this will hopefully surface some possibilities there.

## How Has This Been Tested?
It's test?

## Types of changes

## Checklist:


benjamind:
I'm definitely in favor of our tests moving away from snapshot and towards testing semantics and exposed interface surface (content) rather than internal shadowdom details.

One thought came up when looking at the hacks being made to the `@testing-library/dom`. Rather than monkeypatching the use of `querySelectorAll`, could we not create a `Proxy` over the top level element that contains the elements under test, that would do the same work, but only modifying the top level elements `querySelectorAll` behavior, and would mean we did not modify the behavior of the elements under test?

benjamind:
Also, looks like jsdom is going to very soon support custom elements: https://github.com/jsdom/jsdom/pull/2548

Would this eliminate the need for some of this hackery?

Westbrook:
I'll take a look at what's going on with JSDom, however my understanding is that the project is targeting being spec compliant, which would leave it in the same issue of needing to patch the query code to see through the shadow DOM.

As for the `Proxy` approach, would you be able to share a little more of how you think that would work. My first thought is that the approach pushes the techniques possible into the corner of unit testing only, particularly the point about it "contain[ing] the elements under test". This seems like it would both lower the flexibility of test, while raising the manual requirement to make a test work, which wouldn't be optimal. However, I don't know a lot about `Proxies`, so I'm sure there's much that I just can't see there.

benjamind:
Diving into the actual implementation of `getByLabelText` and similar it appears that a proxy solution wouldn't work. I had hoped that they merely do top level queries, without re-querying back down the stack, but it appears `getByLabelText` will first query the container element for labels, and then re-queries from these labels again to find textareas and other items within them.

My idea was to just have a simple helper which you could apply to a container element to create a Proxy over the HTMLElement instance of the container, and intercept `querySelectorAll` on that instance only to do the custom ShadowDOM piercing. I suppose it could self re-insert the proxy down the tree as it retrieves elements, but it'll start getting messy fast.

Would definitely be better to just get this into the core library I think.

Westbrook:
At least a little bit of interest seems to have peeked up on that end: https://github.com/testing-library/dom-testing-library/issues/413 We'll see what they think of the approach so far 🤞

projects/testing-library-dom/rollup.config.js
```diff
@@ -0,0 +1,48 @@
+import resolve from 'rollup-plugin-node-resolve';
+import commonjs from 'rollup-plugin-commonjs';
+import alias from '@rollup/plugin-alias';
+import replace from '@rollup/plugin-replace';
+
+export default [
```
Westbrook: This gets skinnier and skinnier! Now, knowing https://github.com/testing-library/dom-testing-library/issues/413#issuecomment-579834401 I'd almost say that instead of processing the dependency that we replace it with a warning that says you need to be using this library in an environment that supports `MutationObserver` to make the code to manage here even smaller.

projects/testing-library__dom/index.js
```diff
@@ -0,0 +1,54 @@
...
+};
+
+ShadowRoot.prototype.querySelectorAllWithShadowDOM = function(query) {
+    const children = [...(this.children || [])];
+    let results = children.filter((el) => el.matches(query));
+    children.map((child) => {
+        results = results.concat([
+            ...(child.querySelectorAllWithShadowDOM(query) || []),
+        ]);
+    });
```
sseg, line 50: This section appears to be missing a check of the property before invoking the function, e.g.
```suggestion
    children.map((child) => {
        if (child.querySelectorAllWithShadowDOM) {
            results = results.concat([
                ...(child.querySelectorAllWithShadowDOM(query) || []),
            ]);    
        }
    });
```

I've hit this error in my own tests, while adding the check seems to introduce no adverse effects.

Westbrook: Definitely makes sense!

You wouldn't happen to have any extra information about the type of `child` that it's been triggering this issue for you on? Want to make sure we're applying the method to all of the required prototypes along with making a check/gate like this for backup safety.

sseg: The node that triggers this is an instance of [`SVGSVGElement`](https://developer.mozilla.org/en-US/docs/Web/API/SVGSVGElement).

Issue #321, opened 2019-11-27
Feature Request: A way to manually control overlays


cuberoot:
### Expected Behaviour

We should have a way to manually open or control overlays.

### Actual Behaviour

Right now the only easy way to open an overlay is to use an `overlay-trigger`. However, not all use cases of overlays are triggered by `click` or `hover` events. 

#### Sample Code that illustrates the problem

Right now you have to write code like this to manually open an overlay: https://github.com/adobe/spectrum-web-components/blob/83e4151dc7d1be7617c48ec08fa39e4acbf0f9a1/documentation/src/components/side-nav-search.ts#L74-L115

Westbrook:
This gets dangerously close to pairing #402 with a "Remove the need for `overlay-trigger`" issue/PR pair. Should we push in that direction and make the overlay functionality some sort of mixin or module? The alternative would be to actually use an `overlay-trigger` element in this context and surface a manual API for `OverlayOpen`, possibly an attribute/method pair. Which do you think would solve this issue specifically, while also preparing the library for the most success long term?

cuberoot:
My gut feeling is that the first option is the way to go. I don't know if we want to eliminate the `overlay-trigger` as it is a way to use our overlay system without having to write JS code. You can open overlays from vanilla HTML.

However, if we want to use overlays in our own code and allow programatic control, a mixin might be the thing. We could make `overlay-trigger` use the mixin...

PR #422, opened 2020-01-09
feat: add simpler API for controlling overlays

cuberoot:
## Description

The current implementation uses events to open and close overlays. This would be hard for users to adopt and is not easy to use in our other components that need overlay behaviour.

This change refactors how our overlays are opened and closed to address issue #321.

## Related Issue

#321 

## Motivation and Context

This change puts us in a better place for incorporating overlay functionality into other components (such as #157).

## How Has This Been Tested?

I have tested this in Storybook and have added more tests

## Types of changes

## Checklist:


packages/overlay/src/overlay-stack.ts
```diff
@@ -22,19 +22,11 @@ function hasModifier(event: MouseEvent): boolean {
 }
 
 export class OverlayStack {
-    public overlays: ActiveOverlay[] = [];
```
cuberoot: We can just use the DOM as our data structure. Follow the DRY principle and get rid of this.

packages/overlay/src/overlay.ts
```diff
@@ -10,22 +10,117 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+    delay?: number;
+    placement?: Placement;
+    offset?: number;
+};
 
-import { ThemeData } from '@spectrum-web-components/theme';
+/**
+ * This class allows access to the overlay system which allows a client to
+ * position an element in the overlay positioned relative to another node.
+ */
```
cuberoot: This class can be used from other components

packages/overlay/src/overlay-stack.ts
```diff
@@ -54,47 +83,37 @@ export class OverlayStack {
...
                 // Don't show a hover popover if the click popover is already active
                 return;
             }
 
-            const activeOverlay = ActiveOverlay.create(event, this.root);
-            this.overlays.push(activeOverlay);
-
-            this.onChange(this.overlays);
+            const activeOverlay = ActiveOverlay.create(details);
+            document.body.appendChild(activeOverlay);
```
Westbrook: With the structural changes herein, if I wanted to make the "overlay root" element flexible, where would you see adding that? Having this as a `static` on `Overlay` makes me feel like it's a little fixed, but I don't use a lot of statics in my code to be fully aware of the possibilities they afford.

The issue this supports is something that is currently in Crisp where we don't actually use `sp-theme` to set our styles, as previously there was no way to use it with only _some_ of the styles. The answer may simply be for me to add the `large` styles to the `sp-theme` package, but an alternative would be to allow for our `crisp-app` element to be set for use in place of `document.body` here by sort sort of configuration.

packages/overlay/src/overlay-stack.ts
```diff
@@ -43,7 +35,44 @@ export class OverlayStack {
...
+            }
+        }
+    }
+
+    private *overlays(): Generator<ActiveOverlay, void, undefined> {
+        for (const item of document.body.children) {
+            if (item instanceof ActiveOverlay) {
+                yield item;
+            }
+        }
```
benjamind: I'm really inclined to keep the member overlays array here, just to avoid this loop. Right now, every time you do a popup, of any kind (main case i'm thinking of is a tooltip), you end up enumerating every single child of the body tag. Now true, in a PWA style application this might be quite few, but in the general use case you could potentially have hundreds of elements to iterate and call `instanceof` on here. Since we know the list of overlays, and are managing that list ourselves, we should probably just cache it for speed here.

packages/overlay/src/overlay-stack.ts
benjamind: I too am a little concerned about not being able to control where we reparent overlays to.

The case I'm thinking of is a modal thats contained within an area of an application, in this case I want its z-index to be above the content in the containment area, but below content in other areas outside of the container. This seems natural to me to create an overlay root thats specifically for use by this modal, with the main document.body root being used for tooltips etc.

Is this a case this change supports? I think we may have even lost this capability before this change?

Westbrook: This was a feature taken away when removing `overlay-root`, yes. I don't personally have a preference between needing it now vs the ability to easily refactor this functionality in later, which is why it was omitted previously when this wasn't a `static`.

@benjamind could you expand on:
> The case I'm thinking of is a modal thats contained within an area of an application, in this case I want its z-index to be above the content in the containment area, but below content in other areas outside of the container.

Is this a case of a UI with multiple modals open at the same time, or something else? Would be a good use case to have clearly defined in the docs as a "possible" or "not possible" depending on how this comes together.

packages/overlay/src/overlay-stack.ts
cuberoot: Ok, I can accept that reasoning.

cuberoot:
@Westbrook I was aware of the need for a declarative API too. I was going to leave that for now, but I will go back and have a think about it now.

packages/overlay/src/overlay-stack.ts
benjamind: The case I was thinking of was something like this:

image

Where we have an area where we want a contained modal, this has a mask area behind the modal and a dialog with tooltips etc inside. This is one stack of overlays.

The application however supports dropdowns and tooltips that can potentially overlap this stack, but do so from outside of the contained area.

I've not delved too deeply into how our overlay stack works at this time, so I'm not sure how we support this today, especially with regards to resize of the masked area etc.

cuberoot: So, our overlays stack globally in the order in which they are opened. I could see the dialog being an overlay itself. If you opened an overlay in the dialog (tooltip on hover, click popover, etc) it would open at a higher level than the dialog. If you opened a dropdown from the header, while the dialog was open (ignoring that it is supposed to be modal), the dropdown would appear over the dialog as you would expect.

A problem with explicitly restricting the overlay area to a subset of the view (e.g. the mask) is that we may restrict the positioning of something like a dropdown more than we need to. I don't know if that is something that we want?

So, I guess I am saying that I think we would get reasonable results in the scenario you have outlined without multiple levels of overlay root. I am not yet convinced that the added complexity is warranted. But, I could be convinced 😁

benjamind: Yeah I think I agree that our current code covers the case for the z-stacking issue.

The main concern I have is I think about responding to resize. Since the modal mask and the dialog would be centered within their container, but they are re-parented to `document.body` (or our overlay root), then I'm not quite sure how they would manage their resizing without having to have very explicit code to manage sizing the modal and dialog against their original container?

benjamind:
As discussed offline (on Slack), I think I'd like to see the following changes to this PR:

* Use an event in the `Overlay.open` function to request a root node to target the overlay to, similar logic to what we use for theme discovery
  * this will allow the implementation of custom overlay-roots besides `document.body`
  * Event should be cancellable, and only if not prevented should the overlay behavior kick in
  * Handlers of this event can then modify the parameters of the overlay open event detail and prevent default should they wish

* Use a singleton (module level) WeakMap to keep an OverlayStack per root element
  * this can be done within the `Overlay.open` api to keep all the functionality in the same place, and will allow multiple stacking contexts to behave correctly as z-index stacking contexts do. 

* Since we have module level singletons, consider removing `Overlay` class entirely and instead using an exported function `createOverlay` which returns an object with a `close` method to close the overlay, e..g:

```typescript
const overlay = createOverlay(button, 'click', outerPopover, { delay: 0, placement: 'top', offset: 10,});
// sometime later
overlay.close();
```

We can defer `sp-tooltip` rework to another later ticket I think.

cuberoot:
@benjamind Do you want to extract the future work you want done into an issue and then we can close this?

benjamind:
I'm honestly happy to close this, what we have today serves purpose and we can create new tickets for future work if we decide it needs revisiting then?

cuberoot:
Ok, shutting this one down!

PR #426, opened 2020-01-13
WIP: Try using popper.js for positioning overlays (don't merge)

cuberoot:
## Description

This is an experiment to try out using popper.js. I put it into a PR for discussion

## Motivation and Context

Replaces some poorly understood custom code with a well tested open source alternative

## Types of changes

## Checklist:


cuberoot:
So, this shows a lot of promise. However, I have hit one annoying issue. The spectrum tooltip has the css `width: auto` which doesn't work on an absolutely positioned element.

Screen Shot 2020-01-13 at 6 14 39 PM

I think we got away with this in the old version by computing the position on the overlaid element and then applying the transform to the parent `active-overlay` element instead.

packages/overlay/src/global.d.ts
```diff
@@ -0,0 +1,14 @@
+declare module '@popperjs/core/lib/popper-lite' {
```
cuberoot: The new version of popper comes with Typescript definitions, but they are only included in the fully built version of the library. When you use the uncompiled `lib` version, you have to import them manually.

packages/overlay/src/popper.ts
```diff
@@ -0,0 +1,48 @@
...
+interface WindowNodeWorkaround {
+    process: {
+        env?: {
+            NODE_ENV?: string;
+        };
+    };
+}
+const windowNodeWorkaround = window as WindowNodeWorkaround;
+
+// Popper code references process.env
```
cuberoot: This is really annoying. [This bit of code](https://github.com/popperjs/popper.js/blob/master/src/index.js#L102) in the popper source gets built into this in the released npm:

        if (process.env.NODE_ENV !== "production") {

The following hack works around that.

packages/overlay/src/global.d.ts
cuberoot: I logged https://github.com/popperjs/popper.js/issues/934 for this issue

packages/overlay/src/popper.ts
cuberoot: I logged this issue: https://github.com/popperjs/popper.js/issues/933

benjamind: Yeah this is because we're not actually bundling our code anywhere, not exactly sure why they have this in their production code tbh, they should probably just remove.

We're running our code through `es-dev-server` though so the simplest fix for us is to do this in `es-dev-server.config.json`:

```
  babelConfig: {
    plugins: ['transform-node-env-inline'],
  },
```

Andd add the babel plugin to the dev dependencies. I think we should probably have `@popperjs/core` as a peer dependency in the package if we do go ahead with this though so it'll probably want to be fixed at source.

PR #428, opened 2020-01-13
fix(shared): prevent focusing focusable root on second click

### LINKED ISSUE #318: Focusable components gets tab selection focus if you click in the right spot

Westbrook:
## Description
The fix herein roughly follows the suggestion in https://github.com/adobe/spectrum-web-components/issues/318#issuecomment-566861628 except it uses `mousedown` rather than `click` because on further investigation the `focus` state starts that early in the interaction 😱 

It feels like a bit of a hack, which I guess it is being the `Focusable` base class is a workaround for not having `delegates-focus` functionality x-browser. If it feels like too much of a hack, I'm happy to review further.

The associated `tab-list` update isn't explicitly related, but I ran into it while confirming this fix. As such, I can break it out if anyone is worried about cleanliness or what not.

## Related Issue
fixes #318 maybe?

## Motivation and Context
Focusing the things is the path to them being accessible.

## How Has This Been Tested?
Manually. Focus and the shadow DOM is hard to test... 😢 

## Types of changes

## Checklist:

### DIFF
--- a/packages/shared/src/focusable.ts
+++ b/packages/shared/src/focusable.ts
@@ -90,6 +90,9 @@ export class Focusable extends FocusVisiblePolyfillMixin(LitElement) {
     }
 
     protected manageFocusIn(): void {
+        this.addEventListener('mousedown', (event) => {
+            event.preventDefault();
+        });
         this.addEventListener('focusin', (event) => {
             if (event.composedPath()[0] === this) {
                 this.handleFocus();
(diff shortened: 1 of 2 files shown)


Westbrook:
@cdvu1 and @adixon-adobe if the one of the two of you were able to take a look at this branch and see if I'm actually fixing #318 this time, or just think I have, again... I'd be much obliged.

PR #433, opened 2020-01-14
build: update karma conf for stability (hopefully)

davidyxu:
## Description

Slightly modifying @Westbrook 's improvements the karma conf to increase testing stability

## Motivation and Context

The existing karma configuration seems to fix the flaky tests, but it might also be covering up other issues.

If our tests are taking close to a minute before responding at all, then that's a performance issue in our testing framework that we should resolve.
if our tests are still periodically flaky, we're potentially adding several minutes to our build times without realizing it using the retry setting.

Increasing timeout and retry count might still be necessary to fully resolve issues, but I'd like to see if these settings are sufficient by themselves first.

## How Has This Been Tested?

I created two circleCI configurations, testing various settings and logging out CPU usage on the build instance.

[concurrency of 1 and timeout of 20000](https://circleci.com/gh/adobe/workflows/spectrum-web-components/tree/dxu%2Fkarma-experiments)
[concurrency of Infinity and timeout of 20000](https://circleci.com/gh/adobe/workflows/spectrum-web-components/tree/dxu%2Fkarma-experiments-with-concurrency)

I found that with concurrency on, the resource usage spiked and browsers didn't spin up the testing page within the timeout duration.
Also, running the browsers in sequence instead of in parallel doesn't appear to significantly change the time required to run the tests (at least given our current test suite).

## Types of changes

## Checklist:

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -37,8 +37,13 @@ module.exports = (config) => {
                     },
                 },
             },
-            browserDisconnectTolerance: 2,
-            browserNoActivityTimeout: 60000,
+            // our ci uses a 2 core 4 gb ram instance to run tests
+            // higher concurrency reduces reliability of browser timeout
+            // without a significant effect on testing speed
+            concurrency: 1,
+            // this timeout is primarily significant during browser startup
+            // after browser startup we shouldn't be coming close to this limit
+            browserNoActivityTimeout: 30000, // default value from karma
             coverageIstanbulReporter: {
                 thresholds: {
                     global: {


karma.conf.js
```diff
@@ -37,8 +37,13 @@ module.exports = (config) => {
...
             },
-            browserDisconnectTolerance: 2,
-            browserNoActivityTimeout: 60000,
+            // our ci uses a 2 core 4 gb ram instance to run tests
+            // higher concurrency reduces reliability of browser timeout
+            // without a significant effect on testing speed
+            concurrency: 1,
+            // this timeout is primarily significant during browser startup
+            // after browser startup we shouldn't be coming close to this limit
+            browserNoActivityTimeout: 30000,
```
Westbrook: This is the default value, if we want to keep it here for documentation concerns, please notate that.

davidyxu: I'd prefer to keep this value here for multiple reasons. Aside from documentation , I don't want this value changing under us based on upstream changes (i.e. `karma` or `@open-wc/testing-karma` updates).

Westbrook: 👍 All great reason. Just want to save the next person a half cycle of brain, thanks!

PR #434, opened 2020-01-15
Expand theme support, display options in Storybook

### LINKED ISSUE #432: Add support for `size=large` in `sp-theme`
Currently only `size=medium` is supported, add piping for `large` to be available as well. Review delivery techniques for more "specialized" imports when an implementor doesn't want _EVERYTHING_ on their page.

Westbrook:
## Description
- Adds support for `scale=large` in `sp-theme`
- Add support for `theme` knobs system wide in Storybook

## Related Issue
fixes #432 

## Motivation and Context
We're currently using a local instance of the `large` theme in our application, but everyone should be able to use it, too!

## How Has This Been Tested?
Manually.

### Medium
image

### Large
image

## Types of changes

## Checklist:

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -36,7 +36,7 @@ async function run() {
         const scale = select('Scale', scaleOptions, defaultScale, 'Theme');
         defaultScale = scale;
         return html`
-            <sp-theme id="root-theme" color=${color} size=${scale}>
+            <sp-theme id="root-theme" color=${color} scale=${scale}>
                 ${story()}
             </sp-theme>
         `;
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -6,6 +6,7 @@ import {
     withKnobs,
     withWebComponentsKnobs,
     html,
+    select,
 } from '@open-wc/demoing-storybook';
 import '@spectrum-web-components/theme';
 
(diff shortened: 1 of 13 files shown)


.storybook/preview.js
```diff
@@ -18,13 +19,28 @@ async function run() {
...
+        Dark: 'dark',
+    };
+    let defaultColor = colorOptions.Light;
+    const scaleOptions = {
+        Medium: 'medium',
+        Large: 'large',
+    };
+    let defaultScale = scaleOptions.Medium;
+    addDecorator((story) => {
+        const color = select('Color', colorOptions, defaultColor, 'Theme');
```
Westbrook: This immediately surfaced an issue in the Action Button pattern, so I've created #435 to track reviewing the patterns across the various theme settings. Whatever is broken was broken before, so I'd like to be able to merge this without that work, but can be convinced.

Once we get the styles up to snuff, we can also add separate passes on the Visual Regressions to ensure we get keep those styles stable across future changes. Having more than one list would also mark a good time to alter the test collection pattern there as the array of strings is pretty fragile and our recent move to CSF should give us the chance to leverage that approach there. I've created #436 to track that work.

packages/theme/src/scale-large.ts
```diff
@@ -0,0 +1,16 @@
...
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import largeStyles from './scale-large.css.js';
+import { Theme } from './theme';
+
+Theme.registerThemeFragment('large', 'size', largeStyles);
```
Westbrook: I've noticed there's a naming discrepancy between `size` and `scale` across this code... should we try to centralize on what's going on in Spectrum: e.g. `scale`?

packages/theme/src/scale-medium.css
```diff
@@ -11,7 +11,3 @@ governing permissions and limitations under the License.
 */
 
 @import '@spectrum-web-components/styles/scale-medium.css';
-
```
Westbrook: This looks like extra code. Particularly, when using the `sp-theme` element, all of these sheets go after https://github.com/adobe/spectrum-web-components/blob/master/packages/theme/src/theme.css#L16 which already has this rule. I can be convinced that I'm being overly worried here.

packages/theme/src/scale-large.ts
Westbrook: @cuberoot do you have any thoughts on this? It'd be an API breaking change, so we'd want to get to as soon as possible if we did.

Westbrook:
@msdewey agree on the `scale` point, and have made that update. 

@cuberoot want to call out the breaking API change here, so you can have any last words/a heads up if this is being leveraged in LRW.

packages/overlay/src/overlay-trigger.ts
```diff
@@ -75,7 +75,7 @@ export class OverlayTrigger extends LitElement {
 
         const queryThemeDetail: ThemeData = {
             color: undefined,
-            size: undefined,
+            scale: undefined,
```
cuberoot: All good. We'll fix up our

PR #440, opened 2020-01-17
fix(action-menu): spectrum adherence update

Westbrook:
## Description
Responding to design feedback from Spectrum team.

Ensure the `action-button` inside of `action-menu` doesn't apply `field-button` styles.

## Motivation and Context
Spectrum specification adherence.

## How Has This Been Tested?
Manually.

image

## Types of changes
- [x] Visual Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/packages/action-menu/src/action-menu.css
+++ b/packages/action-menu/src/action-menu.css
@@ -15,8 +15,15 @@ governing permissions and limitations under the License.
 }
 
 .icon {
+    /* Sizing for .icon is needed because we are using a raw <svg> due to https://github.com/adobe/spectrum-web-components/issues/155 */
     width: 18px;
     height: 18px;
+
+    /**
+     * Because .icon is acting as default content for its slot, the `::slotted([slot="icon"])` styles do not apply.
+     * In the future it may be necessary to add a default content selector to the style processing code.
+     */
+    flex-shrink: 0;
 }
 
 #popover {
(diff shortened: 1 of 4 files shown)


packages/action-menu/src/action-menu.css
```diff
@@ -17,6 +17,7 @@ governing permissions and limitations under the License.
 .icon {
     width: 18px;
     height: 18px;
+    flex-shrink: 0;
```
adixon-adobe: are these styles/values we should eventually pick up from spectrum-css or dna?

Westbrook: It could be that our processing for `::slotted()` selectors does need some extension. This is needed in our local styles because for `sp-action-menu` we are manually applying the default `icon` to the `ActionButton` that is eventually the base class for this element. Styles for `::slotted([slot=icon])` already apply this rule, but in that default slot content doesn't match this selector (in that it's not actually slotted). Should I add a tracking issue to  look into adding an options `defaultSelector` entry to our `slot` rules?

adixon-adobe: Can you just add a code comment in this case then so someone reading the code here understands why the override is needed? If this ends up being more than an occasional one-off pattern it'd be worth adding an issue.

Westbrook: Done and done. I also added a note on the sizing while I was there. The fact that is was there lead me initially not to delve too deep on this and don't want others to fall into the same trap.

adixon-adobe:
Awesome -- merging up.

PR #441, opened 2020-01-17
test: protect visual regression cycle from flakiness

Westbrook:
## Description
Sometimes a visual regression test just doesn't pass through no fault of its own...that doesn't mean the whole test series should fail. This adds a single retry for tests that fail. This means that the time to green _might_ be test cycle + ~1 test, but not test cycle x 2, anymore. Actually, failures will still fail the test cycle just at the test cycle + ~1 test length.

## Motivation and Context
CI should be more reliable. And, you can already see it in action at: https://app.circleci.com/jobs/github/adobe/spectrum-web-components/2228/parallel-runs/0/steps/0-105 via the `banner--default.png` test.

This may be a sign of a secondary issue, however with aliasing and load time playing into how the URLs for our tests display at screenshot time I think we're in a pretty good place to start with and I'm pretty excited to have this added stability.

## How Has This Been Tested?
Locally.

A full failure will look like the first recycle sign, and a retry for flakiness will look like the second one:
image

## Types of changes
- [x] testing

## Checklist:

### DIFF
--- a/test/visual/visual.js
+++ b/test/visual/visual.js
@@ -82,7 +82,18 @@ module.exports = {
             await page.screenshot({
                 path: `${currentDir}/${type}/${test}.png`,
             });
-            return compareScreenshots(test);
+            let result = await compareScreenshots(test);
+            // Retry the test once for expected flakiness.
+            if (!result.passes) {
+                console.log(`♻️ ${test}.png => retrying...`);
+                await page.screenshot({
+                    path: `${currentDir}/${type}/${test}.png`,
+                });
+                result = await compareScreenshots(test);
+            }
+            // Always run the test at the end, so that you can recieve the failure/diff file
+            // or deliver the passing comparison to the console.
+            return result.test();
         }
 
         function compareScreenshots(view) {
@@ -138,24 +149,30 @@ module.exports = {
                         `${currentDir}/${type}/${view}.png`
                     );
                     const fileSizeInBytes = stats.size;
-                    console.log(
-                        `📸 ${view}.png => ${fileSizeInBytes} bytes, ${percentDiff}% different`
-                    );
 
-                    if (numDiffPixels > 10) {
-                        diff.pack().pipe(
-                            fs.createWriteStream(
-                                `${currentDir}/${view}-diff.png`
-                            )
-                        );
-                    }
-                    // A number of comparisons come out to 10 inexplicable pixes EVERY time,
-                    // so let them have their stupid 10 pixels!
-                    expect(
-                        numDiffPixels,
-                        'number of different pixels'
-                    ).be.lessThan(11);
-                    resolve();
+                    resolve({
+                        // Return the diff file creation and file comparisin expect so that the "test"
+                        // can be run only when it "passes" or is the second try.
+                        test: () => {
+                            console.log(
+                                `📸 ${view}.png => ${fileSizeInBytes} bytes, ${percentDiff}% different`
+                            );
+                            if (numDiffPixels > 10) {
+                                diff.pack().pipe(
+                                    fs.createWriteStream(
+                                        `${currentDir}/${view}-diff.png`
+                                    )
+                                );
+                            }
+                            // A number of comparisons come out to 10 inexplicable pixes EVERY time,
+                            // so let them have their stupid 10 pixels!
+                            expect(
+                                numDiffPixels,
+                                'number of different pixels'
+                            ).be.lessThan(11);
+                        },
+                        passes: numDiffPixels < 11,
+                    });
                 }
             });
         }


benjamind:
Is this not just hiding an issue? I'd honestly rather fix the issue that's causing this inconsistency than workaround it with a retry.

If we at least have root cause on why we need to retry I'd be happier, but we should probably investigate that first before using a workaround to ignore it.

test/visual/visual.js
```diff
@@ -138,24 +149,30 @@ module.exports = {
...
-                        'number of different pixels'
-                    ).be.lessThan(11);
-                    resolve();
+                    resolve({
+                        // Return the diff file creation and file comparisin expect so that the "test"
+                        // can be run only when it "passes" or is the second try.
+                        test: () => {
+                            console.log(
+                                `📸 ${view}.png => ${fileSizeInBytes} bytes, ${percentDiff}% different`
+                            );
```
davidyxu: it would be great if we could get this logged out on failures as well (and persisted as an artifact somewhere so we can review where these blips are coming from), so that we can review the reliability of our visual testing over time.

test/visual/visual.js
```diff
@@ -138,24 +149,30 @@ module.exports = {
...
+                                );
+                            }
+                            // A number of comparisons come out to 10 inexplicable pixes EVERY time,
+                            // so let them have their stupid 10 pixels!
+                            expect(
+                                numDiffPixels,
+                                'number of different pixels'
+                            ).be.lessThan(11);
+                        },
+                        passes: numDiffPixels < 11,
```
davidyxu: could we refactor the number out to a shared const so we don't have to update the value at 3 locations if the threshold is updated?

test/visual/visual.js
```diff
@@ -138,24 +149,30 @@ module.exports = {
...
+                                        `${currentDir}/${view}-diff.png`
+                                    )
+                                );
+                            }
+                            // A number of comparisons come out to 10 inexplicable pixes EVERY time,
+                            // so let them have their stupid 10 pixels!
+                            expect(
+                                numDiffPixels,
+                                'number of different pixels'
+                            ).be.lessThan(11);
```
davidyxu: wasn't aware that we had this threshold here before, do we have any idea what this 10 pixel discrepancy is coming from?

Westbrook:
Generally, this issue arises from what looks like standard font-alias differences which would lead me away from calling this "hiding an issue" as much as admitting reality. Here a list of diffs from failures due to it in the last couple of days:
https://2253-158297197-gh.circle-artifacts.com/0/home/circleci/project/test/visual/screenshots-current/banner--banner-types-diff.png
https://2253-158297197-gh.circle-artifacts.com/0/home/circleci/project/test/visual/screenshots-current/banner--default-diff.png
https://2253-158297197-gh.circle-artifacts.com/0/home/circleci/project/test/visual/screenshots-current/sidenav--default-diff.png
https://2251-158297197-gh.circle-artifacts.com/0/home/circleci/project/test/visual/screenshots-current/actionbar--default-diff.png
https://2251-158297197-gh.circle-artifacts.com/0/home/circleci/project/test/visual/screenshots-current/dropdown--default-diff.png
https://2249-158297197-gh.circle-artifacts.com/0/home/circleci/project/test/visual/screenshots-current/actionbar--default-diff.png
https://2249-158297197-gh.circle-artifacts.com/0/home/circleci/project/test/visual/screenshots-current/dropdown--default-diff.png
https://2249-158297197-gh.circle-artifacts.com/0/home/circleci/project/test/visual/screenshots-current/dropzone--default-diff.png
https://2249-158297197-gh.circle-artifacts.com/0/home/circleci/project/test/visual/screenshots-current/illustratedmessage--default-diff.png
https://2249-158297197-gh.circle-artifacts.com/0/home/circleci/project/test/visual/screenshots-current/search--default-diff.png
https://2249-158297197-gh.circle-artifacts.com/0/home/circleci/project/test/visual/screenshots-current/sidenav--default-diff.png
https://2249-158297197-gh.circle-artifacts.com/0/home/circleci/project/test/visual/screenshots-current/sidenav--hrefs-diff.png
https://2249-158297197-gh.circle-artifacts.com/0/home/circleci/project/test/visual/screenshots-current/statuslight--all-variants-diff.png
https://2249-158297197-gh.circle-artifacts.com/0/home/circleci/project/test/visual/screenshots-current/textfield--default-diff.png

At one point I thought that this was due to font load timing, but looking closer at those diffs, it's hard to believe that to really be true. We _do_ run our test with `waitUntil: networkidle2`, which could theoretically allow for the test to happen before the font was loaded. We could use `waitsUntil: networkidle0` to try and avoid this, I looked into it originally and it was both much slower and still exhibited issues, but much has changed in the intervening time. However, I can't remember 100% if I'd recognized that the test had to always happen in the same environment at that point and I've added some browser caching since then, which should help mitigate the load times. That caching should generally absolve this issue, was it actually font load specific and not general alias flakiness, as the font _should_ just be around to use after the first test. I'll double-check what sort of cache headers the TypeKit fonts are shipped with to confirm.

### To-dos:
- [ ] review costs/benefits of `waitsUntil: networkidle0`
- [ ] review TypeKit caching policies

test/visual/visual.js
Westbrook: Welp, I don't know when it stopped, but I can't seem to trigger it anymore. Maybe something with an upgrade to puppeteer since I originally set this up? It's looking like Ben's note above _might_ also play out, and if so I'll bring both of these notes together in a new PR 🤞

Westbrook:
It is looking like _maybe_ `waitsUntil: networkidle0` will be helping remove the need to a retry 🤞 _and_ that the caching that had been added will ameliorate the time loss of the longer loads 🎉. The process may have also caught the fact that we had one screenshot saved from the wrong side of the load barrier 🙌. I did catch one need for a retry the first time it went through CI that would appear to be a separate timing situation, so I'll run the CI on loop for a little while to see if it feels better enough or if there might be some more tuning to be done. If things are looking good, I'll close this PR in favor of this alternate approach. Thanks @benjamind for pushing for another look at some past prejudices!

Westbrook:
Closing in favor of #444 looks like even if there are some alternate types of failures hiding somewhere we are in the green on the font-related ones with the changes therein!

PR #452, opened 2020-01-24
feat: rework overlays to use popper

### LINKED ISSUE #321: Feature Request: A way to manually control overlays

### LINKED ISSUE #366: Manage carat position when overlay cannot be centered to its root.
### Expected Behaviour
When a carated overlay is presented that cannot be in default alignment with its root element, the carat should be moved to support pointing directly to that root element.

#### Affected elements
- tooltip
- overlay
- more?

### Actual Behaviour
image

### LINKED ISSUE #451: Make overlay positioning more robust using popper
### Expected Behaviour

Overlays should open reliably positioned with respect to their trigger element. We should use the popper library to improve our overlays.

### Actual Behaviour

Our overlay positioning code is cryptic and not well tested.

cuberoot:
## Description

Make our overlay positioning more robust using the popper library.
- refactor attributes for tooltip and popover (breaking change)
- change `delay` attribute to `delayed` and use warmup/cooldown
- fix animation for opening/closing overlays

## Related Issue

fixes #451
fixes #366
fixes #321

## Motivation and Context

This change improves our overlay functionality while reducing the complexity of our code. The complicated job of positioning components is handled by a well tested and maintained open source project.

## How Has This Been Tested?

I have added new unit tests for the warmup timer and I have added new stories to Storybook for dynamically testing the positioning code.

## Types of changes

## Checklist:


cuberoot:
This sort of replaces PR #422. However, I have not addressed some of the issues discussed in #422. e.g. https://github.com/adobe/spectrum-web-components/pull/422#issuecomment-573131510

I also still need to make a pass over the docs. 

I wanted to get this in peoples hands to start reviewing before the weekend though.

cuberoot:
Also, the build seems to be hitting the same issues that I was locally. We will need to figure that out.

Westbrook:
The build issue you're running into here is about `"references": [{ "path": "../overlay" }]` not being in the `tsconfig.json` for `popover`.

packages/overlay/src/overlay-types.ts
```diff
@@ -0,0 +1,55 @@
...
+export type Placement = PopperPlacement | 'none';
+
+export function spectrumPlacement(
+    placement: Placement
+): SpectrumPlacement | undefined {
+    if (!placement || placement === 'none' || /^auto/.test(placement)) return;
+    const match = /^([^-]+)/.exec(placement);
+    if (!match) return;
+    return match[1] as SpectrumPlacement;
+}
```
benjamind: Can you explain a bit the purpose of this function? Seems quite complicated. Is this just to normalize spectrum placement strings with popper strings?

packages/overlay/src/popper-arrow-rotate.ts
```diff
@@ -0,0 +1,58 @@
...
+}
+
+export const computeArrowRotateStyles: Modifier<{}> = {
+    name: 'computeArrowRotateStyles',
+    enabled: true,
+    phase: 'beforeWrite',
+    requiresIfExists: ['arrow'],
+    fn: computeArrowRotateStylesFn,
+    data: {},
+};
```
benjamind: Shame we have to do this, but I like that popper makes this easy to add to its pipeline.

packages/overlay/src/popper.ts
```diff
@@ -0,0 +1,58 @@
...
+const windowNodeWorkaround = window as WindowNodeWorkaround;
+
+// Popper code references process.env
+if (!windowNodeWorkaround.process) {
+    windowNodeWorkaround.process = {
+        env: {
+            NODE_ENV: 'production',
+        },
+    };
+}
```
benjamind: Where did the popper team land on this? Do we want to do this workaround, or just specify usage of the babel plugin for consumers and in our es-dev-server? I'm honestly not sure what the best approach is here.

packages/popover/src/popover.ts
```diff
@@ -44,6 +55,39 @@ export class Popover extends LitElement {
...
+
+    updated(changedProperties: PropertyValues): void {
+        if (
+            changedProperties.has('placement') ||
+            changedProperties.has('dataPopperPlacement')
+        ) {
+            this.computedPlacement =
+                spectrumPlacement(this.dataPopperPlacement || this.placement) ||
+                'top';
+        }
```
benjamind: If we modify the overlay implementation to rename the `data-popper-placement` attribute on content, then maybe we can make this attribute name configurable, and do our spectrumPlacement translation there?

Westbrook:
Looks like this also:

fixes #366

🎉

packages/tooltip/src/tooltip.ts
```diff
@@ -61,3 +107,9 @@ export class Tooltip extends LitElement {
         `;
     }
 }
+
+declare global {
```
Westbrook: TS practices question: when working with an event that can be emitted in multiple places but has a centralized type, should this interface be declared in multiple places or just in `@spectrum-web-components/overlay` where the `OverlayDisplayQueryDetail` declaration is?

packages/tooltip/src/tooltip.ts
```diff
@@ -31,7 +38,13 @@ export class Tooltip extends LitElement {
...
 
     @property({ reflect: true })
-    public tip: 'top' | 'bottom' | 'left' | 'right' = 'top';
+    public placement: Placement = 'top';
+
+    @property({ attribute: 'data-popper-placement' })
+    public dataPopperPlacement?: Placement;
+
+    @property({ attribute: 'computed-placement', reflect: true })
+    public computedPlacement: SpectrumPlacement = 'top';
```
Westbrook: Is it possible to simplify the usage of `placement` based attributed in the style APIs?

Most importantly, I don't like the idea of surfacing a `computed-placement` attribute, if we need all three is there better naming there? Also, should we do the work to make `computed-placement` or it's replacement naming actually `readonly`? Maybe `readonly` is overkill?

packages/tooltip/src/tooltip.ts
```diff
@@ -53,6 +66,39 @@ export class Tooltip extends LitElement {
         this._variant = '';
     }
 
+    public connectedCallback(): void {
+        super.connectedCallback();
+        this.addEventListener('sp-overlay-query', this.onOverlyQuery);
```
Westbrook: Possible over optimization/complexity incoming...

Is it reasonable that `sp-overlay-query` would be handled while `this.open===false` should we move to track `open` as a property and add/remove this listener on it's value change? In theory tooltips are closed more than they are open allowing for fewer listeners active at a time.

packages/popover/src/popover.ts
```diff
@@ -44,6 +55,39 @@ export class Popover extends LitElement {
...
     }
 
+    public connectedCallback(): void {
+        super.connectedCallback();
+        this.addEventListener('sp-overlay-query', this.onOverlyQuery);
+    }
+
+    public disconnectedCallback(): void {
+        super.disconnectedCallback();
+        this.removeEventListener('sp-overlay-query', this.onOverlyQuery);
```
Westbrook:
```suggestion
        this.removeEventListener('sp-overlay-query', this.onOverlayQuery);
```

packages/overlay/test/overlay.test.ts
```diff
@@ -0,0 +1,231 @@
...
+            '#outer-popover'
+        ) as HTMLElement;
+
+        expect(button).to.exist;
+        expect(customOverlay).to.exist;
+
+        expect(isVisible(customOverlay)).to.be.false;
+        expect(isVisible(clickOverlay)).to.be.false;
+
+        Overlay.open(button, 'custom', customOverlay, {
```
Westbrook: Does the ability to do this fix #321? If so, can you add that to the description so it auto closes on merge. If not, could you add some information about next steps to that issue?

packages/overlay/test/overlay-timer.test.ts
```diff
@@ -0,0 +1,140 @@
...
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import { useFakeTimers, SinonFakeTimers } from 'sinon';
+import { OverlayTimer } from '../lib/overlay-timer';
+import { expect } from '@open-wc/testing';
+
+describe('Overlays timer', () => {
+    let clock: SinonFakeTimers;
```
Westbrook: This `SinonFakeTimers` is fancy! Lots of stuff to learn from in here.

packages/overlay/stories/overlay-story-components.ts
```diff
@@ -0,0 +1,252 @@
...
+                    fill="currentColor"
+                    d="M248 8C111.03 8 0 119.03 0 256s111.03 248 248 248 248-111.03 248-248S384.97 8 248 8zm0 432c-101.69 0-184-82.29-184-184 0-101.69 82.29-184 184-184 101.69 0 184 82.29 184 184 0 101.69-82.29 184-184 184zm0-312c-70.69 0-128 57.31-128 128s57.31 128 128 128 128-57.31 128-128-57.31-128-128-128zm0 192c-35.29 0-64-28.71-64-64s28.71-64 64-64 64 28.71 64 64-28.71 64-64 64z"
+                ></path>
+            </svg>
+        `;
+    }
+}
+customElements.define('overlay-target-icon', OverlayTargetIcon);
+
+class OverlayDrag extends LitElement {
```
Westbrook: Nit: Since we have `hover` and `click` content, would it make sense for this to follow without having to mouse down?

packages/overlay/src/popper-arrow-rotate.ts
Westbrook: Is it possible that there are notes around this that we could pass to the Spectrum CSS team that would help make their output more resilient to this style of usage. It sounds like they're always interested in hearing about things that could benefit their tools, so even if we just point out `Because of x, when using popper, we have to y.` might be enough for them to churn on a nicer API on their end. I think FireFox just started listing the transforms as separate CSS properties, which would be a great alternative here...

packages/overlay/src/overlay-types.ts
Westbrook: This is something else we might be able to partner with Spectrum CSS on overtime. If those selectors were `computed-placement^="top"` it seems like we'd be able to avoid this?

packages/tooltip/src/tooltip.ts
benjamind: Good question!

I don't think there's any specific 'best practice' around this. I have generally leant towards doing this declaration in the file/module that emits the event. This seems most logical since then as long as you have imported that module the declaration will be imported. Having said that as you point out if it can be emitted in multiple places, it might make sense to just move it out to its own file so it can be imported appropriately?

I'm not sure if this is best or not though.

packages/overlay/src/overlay-types.ts
cuberoot: Basically, popper supports more placement options than Spectrum does. Our API can support all of the popper options, but we have to map the popper options `top-start`, `top-end` and `top` to the `top` option that Spectrum supports.

I hemmed and hawed a little about this function. It would be more code, but the purpose would be more obvious if I just made it a big mapping.

Westbrook is right that we could just fix the Spectrum CSS.

packages/overlay/src/overlay-trigger.ts
```diff
@@ -180,48 +116,39 @@ export class OverlayTrigger extends LitElement {
...
+    private onTargetSlotChange(event: Event): void {
+        const content = this.extractSlotContentFromEvent(event);
+        this.targetContent = content || undefined;
+    }
+
+    private extractSlotContentFromEvent(event: Event): HTMLElement | null {
         /* istanbul ignore if */
         if (!event.target) {
-            return;
+            return null;
```
Westbrook: Doesn't `return;` inherently return `undefined` preventing the need to clearly define the fallback in the three methods above? If not, should we explicitly return `undefined` for the same reason?

packages/overlay/src/overlay-trigger.ts
```diff
@@ -55,104 +48,44 @@ export class OverlayTrigger extends LitElement {
...
+                {
+                    offset: this.offset,
+                    placement: this.placement,
+                }
+            );
         }
     }
 
-    public onTriggerMouseOver(event: Event): void {
+    public onTriggerMouseOver(): void {
```
Westbrook: Nit: this method name is out of sync with its listener `@mouseenter=${this.onTriggerMouseOver}` Can we think of a less specific naming here that would be more lenient to ongoing refactoring?

packages/overlay/src/overlay-trigger.ts
```diff
@@ -55,104 +48,44 @@ export class OverlayTrigger extends LitElement {
     public disabled = false;
 
     private clickContent?: HTMLElement;
-
     private hoverContent?: HTMLElement;
+    private targetContent?: HTMLElement;
```
Westbrook: What sort of context would the mouse events be triggered where this content wasn't available that we've added the gating associated to this for? Might be a good add to comments or documentation.

packages/overlay/src/overlay-trigger.ts
```diff
@@ -39,12 +31,13 @@ let overlayRoot: OverlayRoot;
  * @slot click-content - The content that will be displayed on click
  */
 export class OverlayTrigger extends LitElement {
+    private clickOverlay?: Overlay;
+    private hoverOverlay?: Overlay;
```
Westbrook: The only method we leverage on these `Overlay`s is `close`...do we see the possibility of expanding to others later? Should we just store the close method instead of the whole overlay?

packages/overlay/src/overlay-stack.ts
```diff
@@ -164,9 +183,9 @@ export class OverlayStack {
 
         this.handlingResize = true;
         requestAnimationFrame(() => {
-            this.overlays.forEach((overlay) => {
+            for (const overlay of this.overlays) {
```
Westbrook: What doe we get out of the `for..of` that `forEach` wasn't giving us here? I've not used it much yet, and not sure the specific benefits that come with it.

packages/overlay/src/overlay-stack.ts
```diff
@@ -67,34 +73,43 @@ export class OverlayStack {
...
+                return promise;
             }
+        }
 
-            const activeOverlay = ActiveOverlay.create(event, this.root);
-            this.overlays.push(activeOverlay);
-
-            this.onChange(this.overlays);
+        return new Promise((resolve) => {
+            requestAnimationFrame(() => {
```
Westbrook: As you've been looking at this closely, do you think we still need the various `rAF`s across this file? Previously, I had added more to make the various functionality more normalized, but seeing it again I should have spent some time understanding fully what they were for as it would make the code here much easier to read without them...

packages/overlay/src/popper.ts
cuberoot: Oh, yah. I let this ball drop. They still seem to be building the "esm" version as `.mjs` files and they don't have Typescript typings. So, we should continue to use the "lib" versions. I guess that means your babel plug-in solution is probably the way to go. I will update this.

packages/overlay/src/overlay-types.ts
cuberoot: I found a way to remove the extra computed property. I just cache the `placement` attribute from the overlay content node before I run popper. I put back the old value when I tear down the popper.

packages/popover/src/popover.ts
cuberoot: I removed the need for `data-popper-placement` and I added a custom transform to the CSS when processing to loosen up the matching of the `placement` attribute by changing the operator from `=` to `*=`

packages/overlay/src/active-overlay.ts
```diff
@@ -134,58 +115,115 @@ export class ActiveOverlay extends LitElement {
...
+    private updateOverlayPopperPlacement(): void {
+        if (!this.overlayContent) return;
+
+        if (this.dataPopperPlacement) {
+            // Copy this attribute to the actual overlay node so that it can use
+            // the attribute for styling shadow DOM elements based on the side
+            // that popper has chosen for it
+            this.overlayContent.setAttribute(
+                'data-popper-placement',
+                this.dataPopperPlacement
```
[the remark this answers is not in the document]
cuberoot: Resolving this as I managed to remove the need for the extra attributes on the overlay content node.

packages/overlay/src/overlay-stack.ts
```diff
@@ -163,10 +182,11 @@ export class OverlayStack {
...
-        requestAnimationFrame(() => {
-            this.overlays.forEach((overlay) => {
-                overlay.updateOverlayPosition();
-            });
+        requestAnimationFrame(async () => {
+            const promises = this.overlays.map((overlay) =>
+                overlay.updateOverlayPosition()
+            );
+            await Promise.all(promises);
             this.handlingResize = false;
```
cuberoot: @Westbrook This is where I had the `for of` loop. I suspect that I did that because I was going to `await` the popper resizes. You can do that in a `for of`, but not in a `forEach`. The only other advantage of `for of` is that it avoids the overhead of a function call.

However, I realized that would serialize the updates (not that there should be many poppers open at a time). I changed it to do the updates in parallel.

packages/overlay/src/overlay-stack.ts
cuberoot: I tried removing this rAF, but things failed when I did. I would like to revisit this and understand it better.

packages/overlay/src/overlay-trigger.ts
cuberoot: I have updated the code as per this suggestion. I think it is good to hide the implementation details of the overlay until we think of a reason why we would need to access the overlay object.

packages/overlay/src/overlay-trigger.ts
cuberoot: I only added the `targetContent` because I need to pass it through to the `Overlay` and `OverlayStack`. The checks for its existence are to make Typescript happy.

packages/overlay/src/overlay-trigger.ts
cuberoot: I fixed the event name mismatch. I shifted the methods around a bit. I wanted the specific naming because there are three different DOM nodes that we register events on and I wanted it to be obvious which callbacks relate to which nodes

packages/overlay/src/overlay-trigger.ts
cuberoot: Yep, `return` is the same as `return undefined.` I new that for JS, but just worked that out for TS in this PR. I tidied this up.

packages/overlay/stories/overlay-story-components.ts
cuberoot: Hmm, I started to implement that, but it strikes me that it would be harder to test mouse hover if the target sticks to the mouse. It gets hard to un-hover. So, it is awkward either way. I'm open to suggestions if you have ideas on how to make the interaction better...

packages/overlay/test/overlay-timer.test.ts
cuberoot: Yah, I was happy with how this landed. I didn't know this stuff before...

packages/overlay/test/overlay.test.ts
cuberoot: I think this solves #321. If we hit other issues we can follow up in a later issue. I added the issue to the description for this PR.

packages/popover/src/popover.ts
cuberoot: Fixed...

packages/tooltip/src/tooltip.ts
cuberoot: Yah, I'm not sure the added complexity and potential bugs are worth it. I'd like to punt on this one.

packages/tooltip/src/tooltip.ts
cuberoot: Looks like this is a duplicate definition. I had already moved this one to [here](https://github.com/adobe/spectrum-web-components/blob/ad1554d3c5648f5b86efc6d7e2d3c0e0cf57897a/packages/overlay/src/overlay-types.ts#L41-L45)

karma.conf.js
```diff
@@ -25,6 +25,9 @@ module.exports = (config) => {
             ],
             esm: {
                 nodeResolve: true,
+                babelConfig: {
+                    plugins: ['transform-node-env-inline'],
+                },
```
Westbrook: No need to do the work here, but can you create an issue to merge the ESM configs between here at `.storybook/main.js`?

packages/overlay/README.md
```diff
@@ -45,8 +45,8 @@ yarn add @spectrum-web-components/overlay
             <sp-button>Press Me</sp-button>
         </div>
     </sp-popover>
-    <div slot="hover-content" class="tooltip" delay="100">
+    <sp-tooltip slot="hover-content" delayed open>
```
Westbrook: 👍

packages/overlay/src/active-overlay.css
```diff
@@ -30,48 +30,58 @@ governing permissions and limitations under the License.
...
 }
 
 :host([state]) {
-    display: block;
+    display: inline-block;
 }
 
-:host([state='visible']) {
+:host([state][position]) {
+    display: inline-block;
```
Westbrook: Is it needed to set `display: inline-block;` on `:host` three different ways?

packages/overlay/src/overlay-stack.ts
Westbrook: Thanks for testing it!

packages/overlay/src/overlay-trigger.ts
```diff
@@ -179,49 +77,78 @@ export class OverlayTrigger extends LitElement {
...
+    }
+
     private onHoverSlotChange(event: Event): void {
+        const content = this.extractSlotContentFromEvent(event);
+        this.hoverContent = content || undefined;
+    }
+
+    private onTargetSlotChange(event: Event): void {
+        const content = this.extractSlotContentFromEvent(event);
+        this.targetContent = content || undefined;
```
Westbrook: You fixed the below return, but I was hoping it would allow this and the two previous methods to not have this:
```suggestion
        this.targetContent = content;
```
With `private extractSlotContentFromEvent(event: Event): HTMLElement | undefined {` these are the same?

packages/overlay/src/overlay.ts
```diff
@@ -10,22 +10,144 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+    private isOpen = false;
+    private overlayElement: HTMLElement;
+    private owner: HTMLElement;
+    private interaction: TriggerInteractions;
+
+    /**
+     *
+     * @param owner the parent element we will use to position the overlay element
+     * @param interaction the type of interaction that caused this overlay to be shown
+     * @param overlayElement the item to display as an overlay
```
Westbrook: All of this documentation is just such a nice addition!

packages/overlay/src/spectrum-config-utils.js
```diff
@@ -0,0 +1,23 @@
...
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+// This transformation loosens up placement selector matching for positioned
+// overlays (e.g. [position="top"] => [position*="top"])
+module.exports.wildcardPlacementSelectorTransform = (selector) => {
```
Westbrook: It would be nice to see a Spectrum CSS issue with the altering of this selector at the source brought up. Or we could include it in our sync next week, just the same.

Westbrook:
Technical point, will this fix #451 or just reference it? Love it when the PRs close the issues on their own...

packages/overlay/src/spectrum-config-utils.js
cuberoot: the `placement="top"` is already a translation from what is in Spectrum CSS. They just have: https://github.com/adobe/spectrum-css/blob/master/components/tooltip/index.css#L107-L115

Maybe we just need to add an option to the css processing for fuzzy matches?

Westbrook: Oh right! That's what I really meant then. Thanks for clarifying that.

cuberoot: I took another pass at this and am happier. I added a `wildcard` option for enums in the spectrum processing and eliminated the custom replacement function.

Westbrook: 👏

scripts/process-spectrum-postcss-plugin.js
```diff
@@ -662,12 +662,13 @@ class ComponentConfig {
                             throw new Error(message);
                         }
                     }
+                    const operator = attribute.wildcard ? '*=' : '=';
```
cuberoot: @Westbrook Here's where I added the `wildcard` option for enum attribute conversion in CSS processing. This is better...

packages/overlay/src/active-overlay.css
cuberoot: Wow, good catch. Historical cruft. Fixed

packages/overlay/src/overlay-trigger.ts
cuberoot: Fixed

cuberoot:
@Westbrook I think I've implemented all of your suggestions. I force-pushed so it may not be obvious above.

cuberoot:
Forgot to upgrade popper. Just did that and force-pushed again.

PR #455, opened 2020-01-24
[DO NOT MERGE!] [RFC] fix(slider): request for comment on gradient backgrounds

Westbrook:
This is possibly what it would look like to ship gradient background support as a helper file for `sp-slider`.

Pros:
- shippable
- flexible
   - even colors with no stop info
   - colors with specific stop info

Cons:
- currently in HTML vs CSS
- `css` wouldn't be able to support the internal usage of looping, etc.
- leverages `variant=filled`, which is a bit weird.

Alternatives:
- A similar algorithm could be achieved in SASS, but then it's in SASS 😞 
- Spectrum CSS could revisit style application
- this could be baked into `sp-slider` or a class extension
   - could look like `<sp-slider trackGradientLeft="#00FF00" trackGradientRight="#FF0000" />`, etc.

Tell me more!

### DIFF
--- a/packages/slider/src/gradients.ts
+++ b/packages/slider/src/gradients.ts
@@ -1,0 +1,76 @@
+/*
+Copyright 2019 Adobe. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License. You may obtain a copy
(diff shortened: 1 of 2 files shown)


Westbrook:
Closing in favor of #456

PR #456, opened 2020-01-25
fix(slider): simplify application of the gradient backgrounds

Westbrook:
## Description
Simplify the application of gradients by addressing `#track-left` and `#track-right` more uniformly and applying a `clip-path` rather than enforcing width via `left` and `right` properties.

## Related Issue
fixes #455 

## Motivation and Context
*Pros*
- requires only one deviation from Spectrum CSS `width: 100%`, which should be both easy to convince the application of and resilient to upstream change if not.
- much simpler
- doesn't rely on `variant=filled` so that you can have a fill and a gradient and that the addition of `fill-start` at a later date should be unblocked

*Cons*
- needs checking in Edge/IE 11 as `clip-path` is listed as "only supporting `url()`" in those browsers depending on which Can I Use you rely on: https://caniuse.com/#search=clip-path

## How Has This Been Tested?
Manually

image

## Types of changes
- [x] refactor

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,7 +36,7 @@ commands:
             - restore_cache:
                   name: Restore Golden Images Cache
                   keys:
-                      - v2-golden-images-d79a43f1c33fd53928ac1767521d4350daffd96e-<< parameters.regression_color >>-<< parameters.regression_scale >>-<< parameters.regression_dir >>
+                      - v2-golden-images-76e51ef79b854a5531a0b65bb2a104a74792d983-<< parameters.regression_color >>-<< parameters.regression_scale >>-<< parameters.regression_dir >>
                       - v2-golden-images-main-<< parameters.regression_color >>-<< parameters.regression_scale >>-<< parameters.regression_dir >>-
             - run: yarn test:visual:ci --color=<< parameters.regression_color >> --scale=<< parameters.regression_scale >> --dir=<< parameters.regression_dir >>
             - run:
--- a/packages/slider/src/Slider.ts
+++ b/packages/slider/src/Slider.ts
@@ -172,7 +172,7 @@ export class Slider extends Focusable {
             <div
                 class="track"
                 id="track-left"
-                style=${this.trackLeftStyle}
+                style=${this.trackStartClipPath}
                 role="presentation"
             ></div>
         `;
@@ -186,7 +186,7 @@ export class Slider extends Focusable {
             <div
                 class="track"
                 id="track-right"
-                style=${this.trackRightStyle}
+                style=${this.trackEndClipPath}
                 role="presentation"
             ></div>
         `;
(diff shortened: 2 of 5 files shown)


adixon-adobe:
@cuberoot / @charlessuh  can you take a look at this? Needs to get updated since it's an old PR, but I'd like to at least make sure we give some feedback on the approach. This would get us a step closer to a useable slider.

Westbrook:
Rebased to support possibly renewed interest. 

Also, I had created https://github.com/adobe/spectrum-css/issues/521 in support of the possibility this moved forward, but still no response. Maybe @lazd is interested in checking it out...

Westbrook:
Double check that the `fill` variant is appropriately supported here.

andrewhatran:
Looks great! Just needs a rebase and a conversion to the new directory structure.

I checked the filled variant of the gradient slider and it matches with what we see here in the [Spectrum docs](https://spectrum.corp.adobe.com/page/slider/).

Screen Shot 2020-07-17 at 3 58 28 PM

lazd:
@Westbrook I really don't know if pre-Chromium Edge is a browser we target. Edge is evergreen, and mine was force-upgraded with a nice little forced intro when I rebooted recently.

@GarthDB @misterbrownlee can you weigh in?

Westbrook:
@lazd thanks for chiming in. If that's true, then I would say that Spectrum CSS should 100% move in this direction. I need to get some input from @adixon-adobe, @cuberoot, et al in the LRWeb team to ensure that they can say the same. I very much prefer the `clip-path` solution as it's just so much cleaner when applying this gradient vs other options.

GarthDB:
@lazd agreed. If it is because of a feature that is going into a main release of chrome/blink and/or WebKit, we should be concerned, otherwise we don’t test or build for prerelease browsers.

misterbrownlee:
These feels to me like we're taking on work to benefit the subset of people who are in Edge but can't / won't update their browser?  Wouldn't that be a pretty small set of people that would be affected.  Can you clarify @Westbrook; are we talking a doesn't work that is a visual defect?  Or it's actually not usable?

lazd:
@GarthDB hmmm, according to [Edge's release history](https://en.wikipedia.org/wiki/Microsoft_Edge#Release_history_(Chromium)), the latest stable version is 84. The first version of Edge with Chomium was 79. If we stick to evergreen - 2, it seems we don't support non-Chomium Edge.

Westbrook:
Thanks for all the input on this. It seems that if Spectrum CSS is "current - 2" and LRWeb is "Chromium Edge with degraded support when prior to that" then my cross browser worries are assuaged. 

To clarify for @misterbrownlee, the changes included herein alter the DOM output/styling to allow for modern browsers to support a gradient full of the track by directly setting the `--spectrum-slider-track-color` Custom Property as currently vended by Spectrum CSS, like follows:
image

There's only the smallest change at the Spectrum CSS level that I'd love to see (because I prefer that the SWC project have as few local overrides of your output as possible) to better support this and it's [outlined here](https://github.com/adobe/spectrum-css/issues/521#issuecomment-661294521). When this originally was proposed in January, this might not have been possible (due to version constraints) as the pre-Chromium builds of Edge display as follows, even though this does not alter the functionality in that browser in any way:
image

I'm reopening this PR now in accordance with everyone's support policy, CI being happy, and looking forward to having this change in the library.

Westbrook:
@andrewhatran I think we're done here? 🤞

andrewhatran (approved):
Lgtm!

Issue #355, opened 2019-12-09
Make updating baseline screenshots nicer.


Westbrook:
When you `yarn test:visual:baseline:ci ${circleToken} ${buildNumber}` it updates all of the shots. Make it only update the changed shots, if we can... 🤞

Westbrook:
Sometimes this "just work ™️", but why?

PR #457, opened 2020-01-27
Run visual regressions against all themes

### LINKED ISSUE #436: Run visual regressions across theme settings
### Expected Behaviour
Visual regression in CI covers all possible thematic delivery of our components

### Actual Behaviour
Currently only `color=light` and `scale=medium` is being tracked. Fixes in this are should also attempt to move us away from a `string[]` based test collection.

After #435 we'll have stable visual delivery for these settings and we'll want to keep it that way!

Westbrook:
## Description
fixes #436
refs #355 

Updating the visual regression suite to cover all six of the current theme options:
- Color: Light, Scale: Large
- Color: Light, Scale: Medium
- Color: Lightest, Scale: Large
- Color: Lightest, Scale: Medium
- Color: Dark, Scale: Large
- Color: Dark, Scale: Medium

This creates a reality that for every `color` we add (there is still one we don't support) there will be >200 golden images added, and for every `scale` (I know of no other scales, currently) > 300. That's a LOT of binaries to add to the repo! I'm not super happy about it, but I'm not sure I like the alternative option that much better. This large number of binaries seems to create some trouble when capturing the golden image off of CircleCI via `yarn test:visual:baseline:ci` in that there seems to be some rate-limiting going on there. This can be avoided by capturing updates one build at a time (though the script has been updated to capture multiple builds), but there is likely some tuning in the script that could make it so this sort of delay happens internally. Due to this, I've not documented the addition of multi-build artifact capturing in the script. At which point this approach can be considered "stable" we should add information on using it.

### Alternative
We could move to rely on the CircleCI cache to maintain the golden images, which would allow us to stop tracking them in the repo all together. The way I could see making this happen is to add all of the generated images (created across 6 different jobs) to the Circle Workspace for every build and then gate a job to the `master` branch that takes all of those images and saves them to a `golden-images-{{ epoch }}` cache for comparison. This has the side benefit of ensuring the golden images are _always_ a current representation of `master`. However, where I start to have issue with this is that you have to commit to merging non-passing branches when you update images. There is also an issue around what the right way to ensure we can compare the golden images to new screenshots, but with extended caching/artifact management we could work out the right process for that if we liked.

## Motivation and Context
We should test ALL of the things.

## How Has This Been Tested?
It _is_ tests!

image

## Types of changes
- [x] Tests

## Checklist:


davidyxu:
Alternatively, could we have every job push up `golden-images-{{BUILD_NUMBER}}` into the cache, and we just manually update to a new `golden-images-{{BUILD_NUMBER}}` in our circle ci config every time we want to switch to a new set of golden images?

So the workflow for updating golden images would be to make a PR, wait for the build to fail, then update the `golden-images-{{BUILD_NUMBER}}` to use the failing build's cache.

Westbrook:
Relying on a specific build number for golden images is certainly an interesting idea, and would work very well if not for the [30 day TTL](https://circleci.com/docs/2.0/caching/#cache-expiration) of the CircleCI cache. I think it does leave us with a number of paths to do additional research on. That being so, does it make sense to move forward with this PR and address more optimal solutions after, or do we think this isn't an acceptable solution either way?

davidyxu:
the changes themselves look fine to me so i'll leave it up to you to decide if you want to merge it in. i do worry that once this goes in any changes to the styling will lead to rapid growth in the git repo size (which we could compact later but would involve an annoying process for all the collaborator).

.circleci/config.yml
```diff
@@ -65,12 +65,57 @@ jobs:
...
+        executor: node8
+
+        steps:
+            - downstream
+            - run: yarn test:visual:ci --color=lightest --scale=large
+            - store_artifacts:
+                  path: test/visual/screenshots-current
+
+    visual-light-medium:
+        executor: node8
```
davidyxu: unrelated, but how come this is labelled as node8? looks like its actually node 10

PR #461, opened 2020-01-30
fix(link): process Spectrum CSS without overwriting specificity

### LINKED ISSUE #85: Focus styles for quiet and over-bg links are being overridden
### Expected Behaviour
Quiet and over-bg links should have a double underline when focused. The double underline declaration is contained within the `a:focus` rule.

### Actual Behaviour
Quiet and over-bg links show a single underline when focused.

The declarations in `a:focus` are being overriden by the ones in `:host([quiet]) a` and `:host([over-background]) a`, which have higher specificity.

 `:host([quiet]) a` and `:host([over-background]) a` contain a declaration for `text-decoration: underline`, which overrides the double underline.

Whenever quiet/over-bg link is focuses

#### Steps to Reproduce
Tab focus the quiet and over-bg link

Westbrook:
## Description
Update the Spectrum Config for the link element to process the CSS without mangling the specificity. I didn't know this was a thing until @cuberoot did it for the [Overlay work](https://github.com/adobe/spectrum-web-components/pull/452/files#diff-326ffef068d4c7c520f1332940a2684d), and I'm not sure if it's really better than #431 or not, but it seems cool.

## Related Issue
fixes #85 
closes #431 

## Motivation and Context
The `focus` states were previously being overwritten.

## How Has This Been Tested?
Manually in storybook.

image
image

## Types of changes

## Checklist:

### DIFF
--- a/packages/link/src/spectrum-config-utils.js
+++ b/packages/link/src/spectrum-config-utils.js
@@ -1,0 +1,43 @@
+/*
+Copyright 2019 Adobe. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License. You may obtain a copy
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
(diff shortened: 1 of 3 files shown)


Westbrook:
I think you're right on the money about the differences here. Luckily Michael's PR for overlays had this approach in it, or I wouldn't have even known it was possible. That being so, we may even want to document it later, but I'm not sure what that might look like yet. If you have thoughts, I'm all ears!

As the #431 only references #85, I don't think it can close it because it doesn't specifically address the root, I also favor this PR, so I'll go ahead and merge this in.

Thanks for taking the time to compare these two with me!

PR #464, opened 2020-01-30
fix(shared): fixes search input - fixes #463

### LINKED ISSUE #463: Search field does not accept input
### Expected Behaviour

I should be able to type into search fields

### Actual Behaviour

The search field placehold turns dark when you click, but you cannot type. See https://opensource.adobe.com/spectrum-web-components/components/search

#### Steps to Reproduce

1. Click on search field at https://opensource.adobe.com/spectrum-web-components/components/search
2. Type into field

benjamind:
## Description

A prior change prevented default mousedown event behavior, this change prevented focusable input elements like search from functioning correctly. This PR fixes this issue by allowing the mousedown event to do its default behavior if the event is targeting the focus element.

## Related Issue

#463 

## Motivation and Context

Search input does not work when clicked today.

## How Has This Been Tested?

Storybook testing.

## Types of changes

## Checklist:

### DIFF
--- a/packages/shared/src/focusable.ts
+++ b/packages/shared/src/focusable.ts
@@ -91,7 +91,9 @@ export class Focusable extends FocusVisiblePolyfillMixin(LitElement) {
 
     protected manageFocusIn(): void {
         this.addEventListener('mousedown', (event) => {
-            event.preventDefault();
+            if (event.composedPath()[0] !== this.focusElement) {
+                event.preventDefault();
+            }
         });
         this.addEventListener('focusin', (event) => {
             if (event.composedPath()[0] === this) {
--- a/packages/shared/src/focusable.ts
+++ b/packages/shared/src/focusable.ts
@@ -90,11 +90,6 @@ export class Focusable extends FocusVisiblePolyfillMixin(LitElement) {
     }
 
     protected manageFocusIn(): void {
-        this.addEventListener('mousedown', (event) => {
-            if (event.composedPath()[0] !== this.focusElement) {
-                event.preventDefault();
-            }
-        });
         this.addEventListener('focusin', (event) => {
             if (event.composedPath()[0] === this) {
                 this.handleFocus();


Westbrook:
This fixes in the `input` entrance, but unfixes https://github.com/adobe/spectrum-web-components/issues/318#issuecomment-566861628 

The `input` situation is certainly a bigger issue, so maybe thats the better place right now. I'm trying through some alternatives, if there's nothing to be done at this time, we should remove the entire `mousedown` listener...

benjamind:
Hmm, tried to reproduce the linked bug on this branch in Firefox but couldn't get the focus outline to misbehave on button or tab elements. Do we have a good set of steps to reproduce that issue? I don't unfortunately have a Safari to test on right now.

Westbrook:
The recreation is to go to the following storybook page `/?path=/story/button--default` in Firefox (I've only tested on MacOS, so it might not be cross OS) and click the first button twice.

Westbrook:
This alteration seems to support both issues:
```
    protected manageFocusIn(): void {
        this.addEventListener('mousedown', (event) => {
            if (this.isFocused) {
                event.preventDefault();
            }
        });
        this.addEventListener('focusin', (event) => {
            if (event.composedPath()[0] === this) {
                this.handleFocus();
                this.isFocused = true;
            }
        });
        this.addEventListener('focusout', (event) => {
            if (event.composedPath()[0] === this) {
                this.isFocused = false;
            }
        });
    }
```
Though, now that I've seen this on, I'm sure there are other `mousedown` issues that we might run into that I'm not seeing right know 🤔

benjamind:
Agreed that this feels like the 'focus solution' may have gotten complex and is now creating more edge cases than it realistically needs. Probably needs a step back and taking another look at it now that we have more components using it.

With regard to the fix proposed here, rather than tracking some new focused state, can we not use `document.activeElement === this.focusElement` or is there some shadowDOM nuance I'm missing there?

benjamind:
Also, I took a look at this on Firefox on windows and cannot reproduce the focus issue at all. Steps I've tried:

* Run this PR
* `yarn storybook`
* navigate to `http://localhost:8000/?path=/story/button--default` on Firefox 72
* Click on `Click Me` first button twice
* Do not see any focus ring

I'm not sure but if i'm reading the old issue correctly then the real problem here is 'Tab' elements getting focus state on the tab when clicked, for example your proposed change above does this on the Tabs default story:

image

The PR as proposed however results in no visible focus ring on the tab when a tab is clicked. Although I did notice that in the 'AutoFocus' story, the tabs always get the visible focus when clicked. So something is still screwy here, but then again I'm not entirely sure what the expected behavior is? It also sounds like there might be varying behavior on OS X vs Windows?

Westbrook:
I may have posted the wrong deep link... does this one point appropriately to the button focus images? https://github.com/adobe/spectrum-web-components/issues/318#issuecomment-566285878 I’m able to recreate what @cdvu1 encountered reliably with the click twice steps. However, if you’re not running into it on windows then that’s an interesting situation and may point to me having a less clear understanding of what’s actually going on than I thought. 

As for `document.activeElement`, I like that approach. There are one or two cases where the focus could be within the element and not on the `focusElement`, however. I think it sounds like a much more elegant approach, so while ‘delegatesFocus` languishes we should probably schedule work to make two separate Mixins for these diverging use cases.

Do we want to get a fix in now, with either the broad removal of the `mousedown` eventing or the use of the internal property, or do you have more time to spend on this? I can dive into a more long term fix once I’m not airport/plane restricted as I am today.

benjamind:
Checking that other photo yeah I am definitely not seeing that result with this proposed PR, on firefox on windows. So does indeed sound like there's some funky os/weirdness going on here.

I do think we should commit some temporary least-bad fix - having non-working search boxes is a very urgent issue. I probably don't have time to dig into this properly and since I can't reproduce on my machine right now it'll involve more time than I have right now to setup my OS X environment.

Will the removal of the mousedown event handling bring the focus ring issue you linked last back in?

Westbrook:
Yes it will. It’s not a deal breaker issue the way the input focusing is, so having it kick around for a while until we find a more stable fix shouldn’t hurt anyone.

benjamind:
Removed the change in favor of removing the event handler entirely. Seems to not cause any issues on windows Chrome, @cuberoot can you confirm your end this fixes the issues you were seeing?

PR #467, opened 2020-02-01
fix: correctly track "activeElement" across shadow boundaries

### LINKED ISSUE #466: Keyboard support on tabs is broken
### Expected Behaviour
Should be able to use arrow keys to move between tabs in a tab list

### Actual Behaviour
Arrow keys move to illogical tab pages

#### Steps to Reproduce

- https://opensource.adobe.com/spectrum-web-components/components/tab-list
- Press tab 4 times in the page to place focus on the "Tab 1" tab under the Example header
- Press right arrow
- Note focus does not move to Tab 2
- Press left arrow
- Note focus moves to Tab 3

#### Platform and Version
MacOS 
Chrome 79.x

Westbrook:
## Description
The use of `document.activeElement` can be problematic across shadow boundaries. This corrects the root node from which to query this value so that `sp-sidenav` and `sp-tab-list` can perform correctly in light DOM and shadow DOM contexts.

## Related Issue
fixes #466

## Motivation and Context
SWC element should be usable in all DOM contexts.

## How Has This Been Tested?
- manually
- unit tests added

## Types of changes

## Checklist:


linters/eslint/index.js
```diff
@@ -25,5 +25,26 @@ module.exports = {
                 };
             },
         },
+        'document-active-element': {
```
Westbrook: Lint for `document.activeElement` to prevent bugs from slipping back in in the future.

packages/.eslintrc.json
```diff
@@ -26,7 +27,17 @@
...
-            { "allowExpressions": true }
+            {
+                "allowExpressions": true
+            }
         ]
-    }
+    },
+    "overrides": [
+        {
+            "files": ["*.test.ts"],
```
Westbrook: Don't add this lint rule to test files where we might choose to test `document.activeElement` on purpose.

packages/.eslintrc.json
```diff
@@ -17,6 +17,7 @@
             "error",
             ["e", "ev", "evt", "err"]
         ],
+        "spectrum-web-components/document-active-element": ["error"],
```
Westbrook: Make `document.activeElement` and error for all `packages/**` files.

packages/sidenav/src/sidenav.ts
```diff
@@ -99,7 +99,8 @@ export class SideNav extends Focusable {
 
     private focusItemByOffset(direction: number): void {
         const items = [...this.querySelectorAll('sp-sidenav-item')];
-        const focused = items.indexOf(document.activeElement as SideNavItem);
+        const focused = items.indexOf((this.getRootNode() as Document)
```
msdewey: Would it be possible to have `(this.getRootNode() as Document).activeElement` as a helper? I could see this as a reusable pattern

Westbrook: What would you see a helper here looking like? Like a `getActiveElement(this)` sort of thing? 

I could also see it being used more regularly, but does that make the results seems more "magic" than the reality of the DOM APIs that they are based on?

andrewhatran:
Great work! I see an opportunity to add shadowDOM focus testing for `menu` as well.

`menu.test.ts` contains a `handle focus and keyboard input` test and could benefit from having a shadowDOM version of that test.

If you think it might be valuable, I would be happy to take a stab at it in a separate PR after this gets merged in 👍

packages/sidenav/src/sidenav.ts
msdewey: yeah, thats sort of what I had in mind. I was thinking it would be beneficial specifically for the shorthand convenience and improving the readability/prettiness of the code (less typecasting on the same line)

documentation/src/components/side-nav.css
```diff
@@ -31,6 +31,7 @@ aside {
         transform: translateX(-100%);
         z-index: 1;
         background: inherit;
+        min-height: 100vh;
```
adixon-adobe: I didn't know about this. Our team can use this now.🤯

PR #468, opened 2020-02-02
ci: move visual regression baseline to golden images cache

### LINKED ISSUE #355: Make updating baseline screenshots nicer.

### LINKED ISSUE #436: Run visual regressions across theme settings

Westbrook:
## Description
Use the CircleCI cache to avoid needing to store 700+ golden images in repo.
- regress against the cache
- add `force_update_regression_cache` parameter to allow for in-branch updating
- document the workflow

## Related Issue
closes #457 
closes #436 
closes #355 

## Motivation and Context
Golden images are hard to keep up to date and dirty up the repo.

## How Has This Been Tested?
It is tests!

image

## Types of changes
- [x] testing

## Checklist:


Westbrook:
@davidyxu thanks for your suggestion in #457 they were the inspiration for this approach. I think this takes full advantage of the cache usage you suggested while taking into account the specifics of CircleCI's constraints in that area. This also removes any issues experienced previously when updating the screenshots in the repo by removing that part of the process. Looking forward to your thoughts.

Westbrook:
@adixon-adobe you also brought up worries around storing binaries in the repo early on, so hopefully, this looks of interest to you as well!

.circleci/config.yml
```diff
@@ -35,10 +35,45 @@ commands:
...
+            regression_scale:
+                type: string
+        steps:
+            - restore_cache:
+                  keys:
+                      - v1-golden-images-<< parameters.regression_color >>-<< parameters.regression_scale >>-b9e157e40b4274683707eccf7f8d400ed4c719bc
+                      - v1-golden-images-master-<< parameters.regression_color >>-<< parameters.regression_scale >>-
+            - run: yarn test:visual:ci --color=<< parameters.regression_color >> --scale=<< parameters.regression_scale >> || << parameters.force_update_regression_cache >>
+            - run: cp -RT test/visual/screenshots-current/ci test/visual/screenshots-baseline/ci
+            - when:
```
davidyxu: would there be a significant downside to always pushing up this cache so we don't need to explicitly create a build before we can update the requested cache?

davidyxu: actually, could we just change this just always be `on_fail` and the build branch cache to be when `on_success`?

https://circleci.com/docs/2.0/configuration-reference/#the-when-attribute
looks like they already have the when hooks that we need, then we can get rid of the `force_update_regression_cache` parameter entirely and simplify the update process to just grab the cache entry generated from failures (instead of needing to request another build with specific parameters).

.circleci/config.yml
```diff
@@ -35,10 +35,45 @@ commands:
...
+                type: string
+            regression_scale:
+                type: string
+        steps:
+            - restore_cache:
+                  keys:
+                      - v1-golden-images-<< parameters.regression_color >>-<< parameters.regression_scale >>-b9e157e40b4274683707eccf7f8d400ed4c719bc
+                      - v1-golden-images-master-<< parameters.regression_color >>-<< parameters.regression_scale >>-
+            - run: yarn test:visual:ci --color=<< parameters.regression_color >> --scale=<< parameters.regression_scale >> || << parameters.force_update_regression_cache >>
+            - run: cp -RT test/visual/screenshots-current/ci test/visual/screenshots-baseline/ci
```
davidyxu: could this be moved into the force update regression cache step?

.circleci/config.yml
```diff
@@ -35,10 +35,45 @@ commands:
...
+                      - save_cache:
+                            name: Force Update Regresssion Cache
+                            paths:
+                                - test/visual/screenshots-baseline/ci
+                            key: v1-golden-images-<< parameters.regression_color >>-<< parameters.regression_scale >>-{{ .Revision }}
+            - save_cache:
+                  name: Build Branch Cache
+                  paths:
+                      - test/visual/screenshots-baseline/ci
+                  key: v1-golden-images-{{ .Branch }}-<< parameters.regression_color >>-<< parameters.regression_scale >>-{{ epoch }}
```
davidyxu: not sure if theres any additional costs from unnecessary caching, should we look into only making this cache save on master? might not be worth bothering with.

.circleci/config.yml
```diff
@@ -35,10 +35,45 @@ commands:
...
+                  keys:
+                      - v1-golden-images-<< parameters.regression_color >>-<< parameters.regression_scale >>-b9e157e40b4274683707eccf7f8d400ed4c719bc
+                      - v1-golden-images-master-<< parameters.regression_color >>-<< parameters.regression_scale >>-
+            - run: yarn test:visual:ci --color=<< parameters.regression_color >> --scale=<< parameters.regression_scale >> || << parameters.force_update_regression_cache >>
+            - run: cp -RT test/visual/screenshots-current/ci test/visual/screenshots-baseline/ci
+            - when:
+                  condition: << parameters.force_update_regression_cache >>
+                  steps:
+                      - save_cache:
+                            name: Force Update Regresssion Cache
```
davidyxu: extra s in regression

.circleci/config.yml
Westbrook: I was having some trouble with this, but I'll experiment a little more, particularly with your ideas around `on_success` and `on_fail`.

.circleci/config.yml
Westbrook: I'll look into this again. I'd used a couple of different options here attempting to leverage `on_fail` but I was probably doing something wrong. Particularly, I added the `{{ .Revision }}` gate late in the process, which _should_ be able to get us to a two-step, rather than three-step process here.

.circleci/config.yml
Westbrook: To the best of my understanding, we're on the "open source" plan here, which makes CircleCI $free.99, but I'll confirm with the opensource channel.

.circleci/config.yml
Westbrook: 🙈

.circleci/config.yml
davidyxu: sounds good, even more reason to make sure we're well behaved members of their open source plan!

.circleci/config.yml
davidyxu: 👍 not a big deal if we can't get it down to a two-step process, happy to approve this if we can't get the `on_fail` approach working

Westbrook: We can chain the `on_fail` which seems like a good way to take ou the need for `force_update_regression_cache`. Updated.

Westbrook:
@adixon-adobe I tweaked the docs a bit to match the great suggestions from @davidyxu that allow the updates to be two-step instead of three, please take a look and let me know if they still make sense!

README.md
```diff
@@ -89,29 +89,18 @@ Visual testing is run against the stories in Storybook, and stories added there
...
 
-1. Update the `force_update_regression_cache` parameter in the `run-regressions` command to `true` in order to build a new cache:
+Your failing branch will have created a new cache with a key of `v1-golden-images-<< parameters.regression_color >>-<< parameters.regression_scale >>-{{ .Revision }}`. In `.circleci/config.yml`, you will use that to update the cache that is restored at the beginning of the `run-regressions`, at least for the next 30 days. Using the revision number outlined in the `Build Golden Images Revision Cache` step of your failing build, update the first cache key listed in the `Restore Golden Images Cache` steps that appeaars as follows:
```
adixon-adobe: little typo. `appeaars`

Westbrook: Correct, thanks!

Westbrook:
Netlify is prepped to work with this PR https://github.com/adobe/spectrum-web-components/pull/469 and not with branches earlier than that in the history. Shouldn't be an issue.

PR #469, opened 2020-02-03
docs: allow PR based deployment of the project for review

Westbrook:
## Description
To allow for earlier testing of proposed changes outside of the context of our local systems, this update adds commands that allow for the deployment to locations where the code can be served from the root of a domain so that PRs can be previewed via Netlify. For issues like #466 and its fix in #467 we'll be able to communicate earlier with consumers about suggested additions and corrections to the library.

## Motivation and Context
Faster feedback in the best feedback.

## Types of changes
- [x] tooling

## Checklist:


Westbrook:
While this is open for discussion all of the other PRs will red line, so please take a minute to review. I can be convinced to walk this back if it seems too much, but I feel it would make talking about our work much easier going forward. Particularly, this would have given a shared distribution of the Storybook site to possibly catch issues like https://github.com/adobe/spectrum-web-components/pull/470 a little more easily in the future.

package.json
```diff
@@ -48,7 +48,10 @@
         "storybook:build": "yarn storybook:stories:build && build-storybook",
         "docs:analyze": "wca analyze 'packages/*/src/index.ts' --format json --outFile documentation/custom-elements.json",
         "docs:build": "yarn build && yarn docs:ci",
+        "docs:deploy": "yarn build && yarn docs:analyze && gulp docsDeploy && yarn storybook:build",
```
adixon-adobe: should there be a README change in this PR about this? Any access people need to do these deploys?

Westbrook: Oh this may be a command naming question. This is "build for deploy to the root director of a server" not "build and deploy". Thoughts on better naming?

Westbrook: As this specifically would be used for deploying a staging version of the docs for reference in a PR, maybe `docs:build:staging`?

adixon-adobe: `site:build`? Any difference between a staging build & a build from master outside of where it's deployed?

Westbrook: The values in the `<base>` tag.

For the "production" site it uses `<base href="https://opensource.adobe.com/spectrum-web-components/">`, and for the "staging" site it uses `<base href="/">` which allows it to be hosted from the root dir rather than from `/spectrum-web-components`, etc.

package.json
```diff
@@ -48,7 +48,10 @@
         "storybook:build": "yarn storybook:stories:build && build-storybook",
         "docs:analyze": "wca analyze 'packages/*/src/index.ts' --format json --outFile documentation/custom-elements.json",
         "docs:build": "yarn build && yarn docs:ci",
+        "docs:build:staging": "yarn build && yarn docs:analyze && gulp docsDeploy && yarn storybook:build",
```
adixon-adobe: So `docs:ci` only has to do the `gulp docsCompile` because it's compiling directly on the prod server? (or because there's a downstream CI task that does the prod deploy?)

adixon-adobe: Now that I'm thinking about this more, is `gulp docsDeploy` the right name for that task? This isn't actually deploying to a server right? Since you're doing the storybook build after that seems unlikely.

Westbrook: What a ball of yarn this one!

Updated to:
- `docs:build` and `docsBuildStaging` (the only documented command) this is what a user would use in a standard workflow as it prepares the docs site to run at the root of a server (localhost or Netlify, etc) and allows for direct inspection of a built documentation site
- `docs:ci` and `docsBuildProduction` which prepare the public documentation site, but a user would never need in standard use

The updated `gulp` naming should also better outline what those commands are used for...

😅

PR #471, opened 2020-02-04
feat(card): Allow slotted html title

msdewey:
Previously, we were only allowing card titles to be passed in as pure text string, meaning that they could not use html. This pr allows support for slotted titles, meaning they can be text fields, headers, strong, italicized, etc. It also continues to support the old implementation, where if a title attribute is supplied and a slot is not provided, it falls back to using the old pure text string title.

## Description

This PR adds a new renderTitle property to card, which all variant render functions use. It includes a slot named `title` which by default contains the title attribute as it's text content

## Motivation and Context

Improves flexibility of the card component, and consolidates some discrepancies between variants

## How Has This Been Tested?

- unit tests
- tested manually
- works in storybook

image

## Types of changes

## Checklist:

### DIFF
--- a/packages/card/README.md
+++ b/packages/card/README.md
@@ -27,7 +27,7 @@ yarn add @spectrum-web-components/card
 
 ## Title
 
-The default application of a title for an `sp-card` is to leverage the `title` attribute. When HTML based title content is desired, the `title` slot is also available for deliverin this content.
+By default, the title for an `sp-card` is applied via the `title` attribute, which is restricted to string content only. When HTML content is desired, a slot named `title` available for applying the title.
 
 ```html demo
 <sp-card subtitle="JPG">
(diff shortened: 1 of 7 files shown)


Westbrook:
We forgot to update the snapshots after we added the slot to all of the versions: https://app.circleci.com/jobs/github/adobe/spectrum-web-components/2963/parallel-runs/0/steps/0-108

packages/card/README.md
```diff
@@ -25,6 +25,18 @@ yarn add @spectrum-web-components/card
 </sp-card>
 ```
 
+## Title
+
+The default application of a title for an `sp-card` is to leverage the `title` attribute. When HTML based title content is desired, the `title` slot is also available for deliverin this content.
```
Westbrook:
```suggestion
By default, the title for an `sp-card` is applied via the `title` attribute, which is restricted to string content only. When HTML content is desired, a slot named `title` available for applying the title.
```

English is the most difficult language we work in...

msdewey: lol, sounds good to me. done and done!

PR #486, opened 2020-02-07
feat(dropdown): open menu UI with overlay system

### LINKED ISSUE #148: Make <sp-popover /> aware of the edge of the window.
### Expected Behaviour
Content in a popover doesn't flow past the edge of the window. Height is truncated and scrolling is activated to make extended content available.

### Actual Behaviour
Content does...

image

😱

### LINKED ISSUE #157: Investigating tying `sp-dropdown` and `sp-action-menu` into overlay system

Westbrook:
## Description
Tie `sp-dropdown` and it's derivatives into the new `Overlay` system.
- ensures content is "over" other content
- ensures that the content breaks any overflow rules
- ensures that the content respects the browser edges
- add `--spectrum-dropdown-menu-width` to control overlaid menu sizing

### Notes
- This is a breaking change!
- There's lots of room for additional documentation if people like the new DOM and attribute APIs. 
- I don't particularly like the addition of `dropdownFirstReady` to the `sp-dropdown` tests, but I'm not sure what else to do there with the need to wait for the `slotchange` event...
- We should look into expanding the theme system to capture arbitrary style snippets to be registered and passed into overlays?

## Related Issue
fixes #157 
fixes #148 

## Motivation and Context
Make things work better, be easier to use, and be more like we'd expect in the browsing experience.

## How Has This Been Tested?
- manual
- unit tests

There's a possibility to add `dropdown--open` to the [list of visual regressions](https://github.com/adobe/spectrum-web-components/blob/master/test/visual/stories.js) but it's timing based (the menu opens with an animation) so I'm hesitant to add a possibly flaky test to the suite.

image

## Types of changes

## Checklist:


packages/action-menu/README.md
```diff
@@ -17,28 +17,26 @@ yarn add @spectrum-web-components/action-menu
 <!-- prettier-ignore -->
 ```html
 <sp-action-menu>
-    More actions
-    <sp-menu slot="options">
```
cuberoot: Good work getting rid of the extra `sp-menu`

packages/action-menu/README.md
```diff
@@ -50,28 +48,26 @@ The visible label that is be provided via the default `<slot>` interface can be
...
-        </sp-menu-item>
-        <sp-menu-divider></sp-menu-divider>
-        <sp-menu-item>
-            Save selection
-        </sp-menu-item>
-        <sp-menu-item disabled>
-            Make work path
-        </sp-menu-item>
-    </sp-menu>
+<sp-action-menu label="More Actions" menu-width="125">
```
cuberoot: It feels a bit weird having presentation mixed with structure here. I feels like we should be able to set a CSS custom property. However, I realise that's a problem with the overlay system. I wonder if components need to be able to specify custom properties that the overlay system should copy over to the overlay?

packages/dropdown/src/dropdown.ts
```diff
@@ -240,60 +305,94 @@ export class DropdownBase extends Focusable {
...
+            <sp-popover id="popover" open @sp-overlay-closed=${this.close}>
+                <style>
+                    sp-popover {
+                        width: ${this.menuWidth || this.offsetWidth}px;
+                    }
+                </style>
+                <sp-menu @click=${this.onClick}>
+                    ${this.optionsList.length
+                        ? html`
+                              ${this.optionsList.map(this.renderItem)}
```
cuberoot: So, this will not replicate any event handlers the user attaches directly to their menu items. I think that would be a pretty unusual use case anyway, but we may want to be conscious of it as we think about this.

packages/action-menu/README.md
Westbrook: We _should_ be able to grab this value from CSS Custom Properties, the only issue is that there is no lifecycle management of those values off of which to hang change management. Do you have thoughts on a contract that you'd be OK with keeping here? Here's a could that I can think of offhand:

- check on the `firstUpdate` of `sp-dropdown`, only (super restrictive, probably always early)
- check on every `update` of `sp-dropdown` (may become performance prohibitive)
- check on every `menuOpen` (probably my suggestion, but a difficult call)
- add an imperative API for an implementor to call when changing (awkward...)
- `resizeObserver` (not hot on the idea for x-browser purposes, but might be the most "correct")

The second question is what do we target for setting the width. The `sp-dropdown` relies on the `--spectrum-global-dimension-size-2400` custom property (which seems like a Spectrum CSS issue) for its `width` and `--spectrum-dropdown-min-width` for its `min-width`, `sp-popover` relies on `--spectrum-global-dimension-size-400` for it's `min-width`, while `sp-menu` and `sp-menu-item` don't seem to have `width` related custom properties at this point. We _could_ pull the `width` value straight out of `sp-menu-item` when we query through them so as to allow a more direct sizing API even if the implementor isn't actually sizing the right `sp-menu-item` it would feel like they are.

Thoughts?

packages/dropdown/src/dropdown.ts
Westbrook: Yes, this is something that I didn't spend a lot of time on. It's actually secondary, IMO, to something that I forgot to get to, which is "reflecting" section state back into the light DOM elements. I feel that I might want to try to punt that to follow up work, but I'm ammenable what you think about that.

For the handling of events, I might suggest we look into the inner workings of `<select>` where this doesn't seem to be offered as an option: https://stackblitz.com/edit/select-box-events?file=my-element.js

packages/action-menu/README.md
cuberoot: I can see why you had side-stepped this one 😄 

I am happy with "check on every menuOpen." That seems like the right granularity. If it isn't, I presume we will hear in the form of new issues.

So, we definitely don't want to make users override `--spectrum-global-dimension-size-2400`. That does sound like an issue with spectrum-css. Maybe we need to define `--spectrum-dropdown-width: var(--spectrum-global-dimension-size-2400);`?

Since CSS custom props are going to be a part of our exposed API, we should file issues with spectrum-css when we hit these. In the meantime, I guess we try to tidy this stuff up in our CSS overrides.

Intuitively, it makes sense to me to set min/max and let the contained items influence the size within bounds. However, I don't know how that sits with the designs. We should probably stick with the spectrum-css implementation for now and refine if needed??

packages/dropdown/src/dropdown.ts
cuberoot: I am fine with leaving this one as is. Just flagged it for discussion.

packages/action-menu/README.md
Westbrook: I'm not sure I see what you're saying about an API that relies primarily on min/max as opposed to simple width. If you have strong feelings there, I'd love to hear more of what you see that covering/supporting to see if I can get on board with it.

However, thinking this through for myself, it's not really `--spectrum-dropdown-width` that we want here (though I think that should probably be added upstream), as that should manage the width of the element visible at all times, the "button". That element should have the ability to be delivered at a different with than the actual "menu" element. To that end, I added `--spectrum-dropdown-menu-width`, and started passing it into the `sp-popover` element on each call to `openMenu()`. The code flow there makes me even more reticent to address the possibility of an API to pass more complex custom property values in the `Overlay` system at large, but I think it works while having some performance safety built-in.

Should this be more thoroughly documented? We've been punting a bit of the CSS Custom Property documentation system-wide, but maybe this _one_ deserves closer work? It's possible this is something we come back to?

benjamind: Is it possible to avoid this copying of the custom property entirely? Should content in the dropdown not follow its own natural width, and potentially allow us to have the dropdown pass on its computed width to the component to force that the dropdown content matches the width of the control? I think both cases are called for?

packages/dropdown/src/dropdown.ts
benjamind: Can you explain why we're recreating the `sp-menu-item` elements into the popup? Why don't we just have a `sp-menu` be the content passed to the `sp-dropdown` and this be passed into the slot in the `sp-popup`?

It seems that the re-creation of the `sp-menu-item` elements is creating a lot of problems in that we have to reflect state back, event handlers don't get copied across etc. Can we avoid this entirely by just having a single item that gets used as the popup content?

packages/action-menu/README.md
Westbrook: While I think that we'd want to learn from the failings of the native `select` element and surface the ability to control the width of the overlaid menu itself at some point, this is certainly a possibility for simplifying this early iteration. Thinking on this again, it also points to subtle support of the quiet variant that seems to be missing currently.

Particularly, it does look like the general expectation is for the [standard variant](https://opensource.adobe.com/spectrum-css/components/dropdown/#Standard) to fill the full width of the field button:
image
While the [quiet variant](https://opensource.adobe.com/spectrum-css/components/dropdown/#Quiet) allows for the content to decide for itself the width it'll display:
image

All three realities can be supported by a small change to the current code, a la:
```
       const menuWidthCustomPropertyValue = this.style.getPropertyValue(
            '--spectrum-dropdown-menu-width'
        );
        const hasCustomProperty =
            menuWidthCustomPropertyValue.search('-') !== -1;
        // only use `getComputedStyle(this)` if it's a custom property itself
        const computedMenuWidth = hasCustomProperty
            ? getComputedStyle(this).getPropertyValue(
                  '--spectrum-dropdown-menu-width'
              )
            : menuWidthCustomPropertyValue;
        // only use `this.offsetWidth` when no value is available && Standard varient
        const menuWidth = computedMenuWidth
            ? computedMenuWidth
            : !this.quiet && `${this.offsetWidth}px`;
        if (menuWidth) {
            this.popover.style.setProperty(
                '--spectrum-dropdown-menu-width',
                menuWidth
            );
        }
```
Future versions could even look into caching to possibly reduce repeated measuring, but locking down the expected API seems much more important at this point.

Or we could use a simplified (either in this sort of structure or directly in the `style` tag):
```
        // only use `this.offsetWidth` when Standard varient
        const menuWidth = !this.quiet && `${this.offsetWidth}px`;
        if (menuWidth) {
            this.popover.style.setProperty(
                '--spectrum-dropdown-menu-width',
                menuWidth
            );
        }
```
And then revisit the ability to deliver more granular customization of the visual delivery later.

packages/dropdown/src/dropdown.ts
Westbrook: At first blush, I was attempting to get away from forcing the following DOM API on users:
```
<sp-dropdown>
  <sp-menu>
    <sp-menu-item>Item 1</sp-menu-item>
    <sp-menu-item>Item 2</sp-menu-item>
    <sp-menu-item>Item N</sp-menu-item>
  </sp-menu>
</sp-dropdown>
```
As I think it's more responsibility on an implementor of this element than I'd like to foist. @cuberoot seems too [agree](https://github.com/adobe/spectrum-web-components/pull/486#discussion_r376711978) but likely we can both be concinced for the right benefits.

In the context of passing an `sp-menu` like this into the `sp-popup`, are you thinking that we'd use `.cloneNode(true)` to copy that deep DOM structure? I'm a little less excited about that for not knowing fully the node management that would go into not creating a memory leak via repeated clones, but that should be achievable with a little extra work and it looks like it would support the end goal that we're looking for here. Another side effect of this is we'd need to move from a `slotchange` based cloning to a `MutationObserver`, as the former doesn't observe changes beyond the first level of descendants.

benjamind: Rather than cloning why don't we take advantage of the same popover pattern and do a move of the element with a placeholder to move it back on popover close?

Westbrook: We certainly could project that content completely from the light DOM... I'm not sure how I feel about this complex of a DOM API:
```
<sp-dropdown>
  <sp-popover>
    <sp-menu>
      <sp-menu-item>Item 1</sp-menu-item>
      <sp-menu-item>Item 2</sp-menu-item>
      <sp-menu-item>Item N</sp-menu-item>
    </sp-menu>
  </sp-popover>
</sp-dropdown>
```
The way it puts even more responsibility on the implementor is a bit of a drag while also opening up questions like:
- Do we want to just "dropdown" whatever content is a child of `sp-dropdown`? That may actually be a benefit, even if it's a trip to fully document it.
- How do we message a user when they have either content that isn't accepted or is malformed (read 2 or more root nodes, but also say they are missing the `sp-menu` which manages a bit of the accessible functionality here)?
- Is the reason that `<option>` elements aren't allowed to have event listeners a way of protecting race conditions in the state management of the element?
- While `lit-html` will actually follow the DOM elements and print templated changes to them in the new location, how do other renderers deal with having nodes they've previously places somewhere be moved to somewhere else?

Other than the possible simplicity of overlaying from the light DOM, can you expand on any other benefits you think we'd get from that approach? I'm having trouble things of ones that don't open new possibilities of difficulty. For instance, it would allow for users to apply inline styling directly on those elements that will propagate to the overlaid context. However, that technique would leave even more elements open to issue creation when expected custom property interfaces that implementors would have become accustomed to using those elements in other contexts are not available to them. Not that we fully get away from that now, but explaining that we restrict one element's interface seems less of a lift long term than doing so for three levels of elements.

benjamind: I think your first point is exactly what I would like to see. Whatever content you slot into a dropdown is what is lifted to a popup, so you could do:

```
<sp-dropdown>
   <sp-menu>
      <sp-menu-item>Item 1</sp-menu-item>
      <sp-menu-item>Item 2</sp-menu-item>
      <sp-menu-item>Item N</sp-menu-item>
    </sp-menu>
</sp-dropdown>
```

We could make the rule that it is single element only if we wanted to simplify, but I don't think it'd be too hard to make it lift multiple elements if necessary. There is then no concern about messaging a user that content is not accepted.

The main benefits I see are:

* Event handler binding is maintained, this is great for things like 'preview on hover' behaviors
* Elements can provide their own custom styling inline or through custom components
* Customization is easier, if we fix to a specific element then customization becomes much harder in the future.
* We could 'optionally' apply the matching width behavior from the dropdown if we wanted (so sp-dropdown would measure itself and force an inline style width onto the popover:

```
<sp-dropdown match-width>
   <sp-menu>
      <sp-menu-item>Item 1</sp-menu-item>
      <sp-menu-item>Item 2</sp-menu-item>
      <sp-menu-item>Item N</sp-menu-item>
    </sp-menu>
</sp-dropdown>
```

I do agree that other renderers (namely react) may have issues with nodes being lifted up through the DOM, but we still have this problem anyway with the current cloning solution since those cloned nodes are lifted up in `sp-popover` internally. I think I would rather generically solve that sort of issue at a 'react wrapper' level (possibly through a specific event that is raised to indicate nodes that are about to be lifted) than restrict our approaches because of the issues in those renderers.

I would also propose that if we did go with this approach we would provide two more configurable attributes on `sp-dropdown`:

* `selected-event="selected"` - the name of an event the `sp-dropdown` will listen for from slotted content to get the selected state from which it will expect a state to apply to the label.
* `selected-attribute="selected"` - the name of an attribute that `sp-dropdown` will reflect down to the slotted content to drive selection state within the dropdown'd content.

This would then allow for a great degree of flexibility in what dropdown content could be applied.

Having said all of this however, I am not sure how 'spectrum'-y this all is. Do we want to be rigid in what our components do and how they behave, or flexible? Are we building a toolkit, or a rigid implementation of spectrum. Perhaps this is one of those cases where we have a base implementation and a specific spectrum implementation of it?

Westbrook: I'm not 100% sure we're talking from the same shared knowledge here and want to take a quick step back. Particularly, I want to make sure we're referencing the elements required across this PR in the same way. [`sp-popover`](https://opensource.adobe.com/spectrum-web-components/storybook/?path=/story/popover--default) is an element that delivers a floating "page" UI, without actually doing any of the functional work to project the DOM that it is related to alternate locations in the main DOM tree. [`sp-menu`](https://opensource.adobe.com/spectrum-web-components/storybook/?path=/story/menu--default) will deliver a list of `sp-menu-*` elements as a single list, but without the containing `sp-popover` element (to the left of the Storybook link above is no `sp-popover`, and to the right is with a wrapping `sp-popover`) does not deliver its content in the way expected by the `sp-dropdown` element. [`Overlay`](https://opensource.adobe.com/spectrum-web-components/storybook/?path=/story/overlay--default) is the internal API that is powered by popper.js to project content into other parts of the DOM and overcome containment and z-index issues. It's also possible that I've missed some of the forest for the trees here, so please let me know if you think I've missed something. If you had different expectations here, then it might be worthwhile for us to start outlining a series of issues that bring the library more inline with what you had been thinking.

If we feel like we're all on the same page, the main thing I come to issue with what you've outlined above is that if you want this DOM API:
```
<sp-dropdown match-width>
   <sp-menu>
      <sp-menu-item>Item 1</sp-menu-item>
      <sp-menu-item>Item 2</sp-menu-item>
      <sp-menu-item>Item N</sp-menu-item>
    </sp-menu>
</sp-dropdown>
```
Then we'd have to move actual DOM element twice; once moving the `sp-menu` element from the light DOM to an `sp-popover` landing zone inside of the `sp-dropdown`'s shadow DOM (or maybe an imperatively created element internally? not something we do _yet_ but could be useful functionality to build, maybe?), and then a second time to project it out of the containment/z-index blockers. None of this is _impossible_, it just doesn't seem like the API that a "dropdown" element really should be supporting, particularly with the specifics outlined by Spectrum for this element. From my vantage point, this is the main blocker remaining in this ticket, _where_ we do _what_ work to ensure that the elements required to deliver the UI of an [`sp-dropdown`](https://opensource.adobe.com/spectrum-css/components/dropdown/). That is assuming one of the [two options outlined above regard width application](https://github.com/adobe/spectrum-web-components/pull/486#discussion_r377949808) are acceptable.

The rest of the conversation you've started here seems like great work for us to do at some point, but I agree that whether or not it is "Spectrum"-y, it definitely isn't "Spectrum Dropdown"-y. I propose we focus on this sort of downward abstraction at a later date. New underlying functionality should then be able to replace the internals of whatever `sp-dropdown` API we settle on here while supporting a clear "best practice" for the more advanced sorts of overlay based interactions you've outlined above. In the case that doing so isn't acceptable, I'd have to invite you to take this branch or the learnings herein to work out the techniques you've proposed separately as that level of investment doesn't align with my current level of availability.

benjamind: Ah yes thanks for the clarification, I had assumed `sp-popover` had included the use of `Overlay` to perform the popup magic, that's interesting that it doesn't in that it means `sp-popover` consumers need to implement this themselves. Something to file away for later discussion.

For the `sp-dropdown` api specifically, I think we should be using `sp-menu-item` with a wrapping `sp-menu`, as I believe this aligns with the react spectrum implementation, and also just logically makes more sense to me, and I absolutely want to avoid cloning the content because of the issues around event binding as discussed previously.

If the `sp-dropdown` design requires that the menu be wrapped in an `sp-popover` then I think this should be either created imperatively as part of the `Overlay` creation, or as you say an internal render into the `sp-dropdown` shadowDOM which is then hoisted through `Overlay` from there. Either approach here is fine to me, and does not affect the external API in either case.

With regard to width application, the hangup I have there is the requirement to use the custom property to define the dropdown content width. I think in general you should be able to style a `sp-dropdown` element width in any of the ways css provides us (block level, absolute positioning, flexbox etc), and that width should then either carry through to the content of the dropdown or not depending on the case that we're trying to solve (the quiet case was a good example here). I think I'm afraid that the requirement to use the custom property makes it much harder to use this element in a variety of layout contexts if that makes sense and this is why I was leaning towards bounding client rect measurement and direct width assignment to the overlay container.

Finally, agreed on iterating later on the helper abstractions here which will make it easier for other component developers to produce similar behavior, this can certainly be delayed to a later date. But I can see our team internally requiring this stuff soon so we should have another conversation about that separately when you get back from vacation.

Westbrook: Ok, I felt we were talking across each other a little, glad we’re on the same page now. 

I’m not so happy about the requirement of a wrapping `sp-menu` in the light DOM, but we can shake out any trouble that brings or not as we continue towards a 1.0. 

Let me make sure I understand the next steps. 

1. Make the following DOM work again:
```
<sp-dropdown>
   <sp-menu>
      <sp-menu-item>Item 1</sp-menu-item>
      <sp-menu-item>Item 2</sp-menu-item>
      <sp-menu-item>Item N</sp-menu-item>
    </sp-menu>
</sp-dropdown>
```
2. Rely on a version of the following to ensure popovers fill the width on Standard drop downs and assume own width for Quiet variants:
```
// only use `this.offsetWidth` when Standard varient
        const menuWidth = !this.quiet && `${this.offsetWidth}px`;
        if (menuWidth) {
            this.popover.style.setProperty(
                'width',
                menuWidth
            );
        }
```
3. Punt more nuanced management of the `sp-popover`’s width to inline styling and/or a later date. 

Does that sound right?

benjamind: If you have good reasons for removing the `<sp-menu>` I'm willing to be convinced, do we know what React Spectrum v3 interface for this component looks like right now? I'd like to keep some parity as there are other Adobe internal reasons for this I won't go into here :-)

Agreed on 2, width should just be measured and copied across to the dropdown content.

Definitely agree we can punt on further inline styling of the nested `sp-popover` until later.

One point missing from this summary though is the fact that we should move rather than clone the content nodes to avoid breaking event handlers.

cuberoot: I think that React Spectrum has the options passed in as a prop. I can see the V3 specs on dropbox paper, but I haven't found the code for React Spectrum V3.

cuberoot: Found the code: https://git.corp.adobe.com/React/react-spectrum-v3

cuberoot: Ok, took some digging... They have [renamed `Dropdown` to `MenuTrigger`](https://git.corp.adobe.com/React/react-spectrum-v3/blob/master/specs/api/MenuTrigger.md)

[Here's a story showing the usage](https://git.corp.adobe.com/React/react-spectrum-v3/blob/master/packages/%40react-spectrum/menu-trigger/stories/MenuTrigger.stories.tsx)

PR #492, opened 2020-02-09
chore: update to @open-wc/demoing-storybook 1.10.1

LarsDenBakker:
A recent change broke the theming of this project. After a lot of debugging, I was able to find out why.

In `@open-wc/demoing-storybook` v1.10 we changed stories from dynamic imports:

```html
<script type="module" src="./.storybook/preview.js"></script>  
<script type="module">
  import { configure } from '/node_modules/storybook-prebuilt/web-components.js';
  configure(() => [
    import('./story-a.js'),
    import('./story-b.js'),
  ], {});
</script>
```

To static imports:

```html
<script type="module" src="./.storybook/preview.js"></script>  
<script type="module">
  import { configure } from '/node_modules/storybook-prebuilt/web-components.js';
  import * as storyA from './story-a.js'
  import * as storyB from './story-b.js'
  configure(() => [
    storyA, storyB
  ], {});
</script>
```

This is better for rollup, as it doesn't need to create a separate chunk for each story. This however doesn't guarantee that `.storybook/preview.js` is run before the stories are run, rendering the first story without the theming decorator. 

Solution is to move the import for the preview.js inside the same module script, so that it is always loaded before stories are loaded:

```html
<script type="module">
  import './.storybook/preview.js';
  import { configure } from '/node_modules/storybook-prebuilt/web-components.js';
  import * as storyA from './story-a.js'
  import * as storyB from './story-b.js'
  configure(() => [
    storyA, storyB
  ], {});
</script>
```

This worked some of the time during development, but not all the time. This is because the decorators are registered in an async function after fetching the custom-elements.json. This creates a delay, so the first story will (potentially) be rendered without the decorator. Moving the configuration outside of the async function fixes that.

It still didn't work during development, while it worked fine in every other project I tried. I found out that rollup was tree shaking `.storybook/preview.js` out of the bundle. After stripping the project down I found the `sideEffects` option in the package.json, which was basically telling rollup that `preview.js` did not have any side effects :)

With these changes, everything should work properly. As a bonus, v1.10 adds proper tracking of added/removed/renamed stories :)

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -18,50 +18,50 @@ window.process = window.process || {};
 window.process.env = window.process.env || {};
 window.process.env.NODE_ENV = window.process.env.NODE_ENV || 'production';
 
+addDecorator(withA11y);
+addDecorator(withKnobs);
+addDecorator(withWebComponentsKnobs);
+const colorOptions = {
+    Lightest: 'lightest',
+    Light: 'light',
+    Dark: 'dark',
+};
+let defaultColor = colorOptions.Light;
+const scaleOptions = {
+    Medium: 'medium',
+    Large: 'large',
+};
+let defaultScale = scaleOptions.Medium;
+addDecorator((story) => {
+    const color = select('Color', colorOptions, defaultColor, 'Theme');
+    defaultColor = color;
+    const scale = select('Scale', scaleOptions, defaultScale, 'Theme');
+    defaultScale = scale;
+    return html`
+        <sp-theme id="root-theme" color=${color} scale=${scale}>
+            ${story()}
+        </sp-theme>
+    `;
+});
+
+addParameters({
+    a11y: {
+        config: {},
+        options: {
+            checks: { 'color-contrast': { options: { noScroll: true } } },
+            restoreScroll: true,
+        },
+    },
+    docs: {
+        iframeHeight: '200px',
+    },
+});
+
 async function run() {
     const customElements = await (await fetch(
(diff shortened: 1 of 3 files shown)


LarsDenBakker:
There are some differences in illustratedmessage, I'm wondering if this is due to a transitive dependency update?

Westbrook:
@LarsDenBakker thanks for getting into this. It looks like the stories are getting written a little differently:

```
<sp-theme>
    <sp-illustrated-message></sp-illustrated-message>
</sp-theme>
```

Whereas previously there was an intervening `div` element.

```
<sp-theme>
    <div>
        <sp-illustrated-message></sp-illustrated-message>
    </div>
</sp-theme>
```

I actually think this newer version is _more_ correct, but don't see anything in this change that would cause the addition/removal of that element, which leaves me a touch wary.

Separately, the v1.10.2 build looks like it's generating full-page failures again... 🤦‍♂

Once we get those two things either understood or addressed small screenshot "corrections" like this can be managed via these instructions: https://github.com/adobe/spectrum-web-components#keeping-ci-assets-up-to-date

LarsDenBakker:
https://github.com/adobe/spectrum-web-components/issues/491 will reports failes regardless, because it doesnt contain the changes from this PR.

We could merge these changes without a version bump since they work with an earlier version too. Then at least the changes are separated.

Westbrook:
Ahh, got it! I agree. I am still a little foggy this morning.

Thoughts on the removal of the `<div>` in the story HTML?

PR #506, opened 2020-02-15
feat(dropdown): open menu UI with overlay system

### LINKED ISSUE #148: Make <sp-popover /> aware of the edge of the window.

### LINKED ISSUE #157: Investigating tying `sp-dropdown` and `sp-action-menu` into overlay system

Westbrook:
## Description
Tie `sp-dropdown` and it's derivatives into the new `Overlay` system.
- ensures content is "over" other content
- ensures that the content breaks any overflow rules
- ensures that the content respects the browser edges
- moves the content to ensure event listeners are maintained
- uses the widths of the `sp-dropdown` for the menu content in Standard variants and allows content to size itself in the Quiet variant

### Notes
- New PR as a number of the implementation details, tests, etc have changed from #486 
- This is a breaking change!
- There's lots of room for additional documentation if people like the new DOM and attribute APIs.
- continue to require an `sp-menu` in the light DOM even though it's a small leak of our implementation details as it falls closer to the API found in [React Spectrum's `MenuTrigger`](https://git.corp.adobe.com/React/react-spectrum-v3/blob/master/specs/api/MenuTrigger.md) 

## Related Issue
fixes #157 
fixes #148 
closes #486 

## Motivation and Context
Make things work better, be easier to use, and be more like we'd expect in the browsing experience.

## How Has This Been Tested?
- manual
- unit tests

There's a possibility to add `dropdown--open` to the [list of visual regressions](https://github.com/adobe/spectrum-web-components/blob/master/test/visual/stories.js) but it's timing-based (the menu opens with an animation) so I'm hesitant to add a possibly flaky test to the suite. This might point to the benefit of some sort of "no-first-transition" style attribute later down the line, both as a helper for testing but for the possibility that certain states that are traditionally animated into might be desired on load in production.

image

## Types of changes
- [x] breaking change that adds new features.

## Checklist:

### DIFF
--- a/packages/dropdown/src/dropdown.ts
+++ b/packages/dropdown/src/dropdown.ts
@@ -37,10 +37,11 @@ import {
     MenuItemQueryRoleEventDetail,
 } from '@spectrum-web-components/menu-item';
 import '@spectrum-web-components/popover';
+import { Overlay, Placement } from '@spectrum-web-components/overlay';
 
 /**
- * @slot default - The placeholder content for the dropdown
- * @slot options - The menu with options that will display when the dropdown is open
+ * @slot label - The placeholder content for the dropdown
+ * @slot {"sp-menu"} - The menu of options that will display when the dropdown is open
  */
 export class DropdownBase extends Focusable {
     public static get styles(): CSSResultArray {
@@ -68,7 +69,10 @@ export class DropdownBase extends Focusable {
     @property({ type: Boolean, reflect: true })
     public open = false;
 
-    public optionsMenu: Menu | null = null;
+    public optionsMenu?: Menu;
+
+    @property()
+    public placement: Placement = 'bottom-start';
 
     @property({ type: Boolean, reflect: true })
     public quiet = false;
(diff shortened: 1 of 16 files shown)


packages/dropdown/src/dropdown.ts
```diff
@@ -240,16 +304,23 @@ export class DropdownBase extends Focusable {
...
+                @sp-overlay-closed=${this.onOverlayClosed}
+            ></sp-popover>
         `;
     }
 
+    protected firstUpdated(changedProperties: PropertyValues): void {
+        super.firstUpdated(changedProperties);
+
+        /* istanbul ignore if */
+        if (!this.shadowRoot) return;
```
benjamind: Why the shadowRoot check here?

Westbrook:
The errors in Storybook seem to be related to https://github.com/open-wc/storybook-prebuilt/issues/4, I've rolled back the updates to the `open-wc` packages so that they can be addressed separately. There's a greenkeeper PR due to confirm versions across a number of things are OK anyway, so they won't be put off for too long.

@benjamind PTAL!

PR #515, opened 2020-02-27
Add download attribute to <sp-link>

### LINKED ISSUE #508: Support "download" attribute on sp-link
### Expected Behaviour
When I use <sp-link> and set the href to a URL formatted data string for downloading a file, I'd like to specify the filename as I do with the <a> tag.

### Actual Behaviour
The download attribute is ignored and the file naming is out of my control

bengfarrell:
Allow optional "download" attribute on <sp-link>

## Description

Look for "download" attribute for when use requires the "href" to be a downloadable URL encoded file and use case dictates that this file is given a proper custom filename.

## Related Issue

fixes #508

## Motivation and Context

To give a URL encoded file download a filename, this component wasn't able to be used before, and use cases requiring that functionality needed to fallback to using an <a> tag.

## How Has This Been Tested?

Tested in Storybook by adding a use case. Also added on to the existing unit test, but unable to test in my own environment. CI should do proper testing for validation

## Types of changes

## Checklist:

- [x I have read the **CONTRIBUTING** document.
- [?] All new and existing tests passed.


bengfarrell:
OK! This is my second attempt at this PR. My first inadvertently got another file mixed in, and I saw the failing test so I thought I'd better give it a second go. It's still failing, and the failure seems unrelated to anything regarding my changes. I'll leave it here, hoping that someone will pick this up where I left off

benjamind:
Looks like its failing due to unrelated reasons, probably the golden images for the dark large theme: https://circleci.com/gh/adobe/spectrum-web-components/4017?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link

Thanks for submitting the PR!

packages/link/test/link.test.ts
```diff
@@ -18,7 +18,9 @@ describe('Link', () => {
     it('loads', async () => {
```
Westbrook: This is exactly the test I wanted for the `download` attribute, but can we call it `loads [download]` (or similar) and have it be a second test, after what was there previously so that we test on either side of the presence of this attribute?

packages/link/stories/link.stories.ts
```diff
@@ -46,6 +46,19 @@ export const overBackground = (): TemplateResult => {
     `;
 };
 
+export const Download = (): TemplateResult => {
```
Westbrook: This is a great way for people to see how this works, thanks!

bengfarrell:
OK! Just did those changes you requested @Westbrook! Failed in the exact same spot which is unrelated to the change, so I think we be good, unless you have anything further, which I'd be happy to address

Westbrook:
Thanks again @bengfarrell, this went green again on `master`... updating Puppeteer [here](https://github.com/adobe/spectrum-web-components/pull/520/commits/2a87df21351b5628b52f2a012dc94492d1a1b1d3), which _might_ help clarify the issue you are seeing here.

PR #520, opened 2020-03-01
fix(overlay): extend state machine to manage disposal process

### LINKED ISSUE #511: Overlay Trigger doesn't trigger if browser focus is lost
So FIRST of all - I did say I'd say I could submit a PR if I found something. I'm not - because it's hecka complicated from the looks of it. There's lots of logic here to unravel and lots of use cases it looks like you want to support for me to even think about how you'd want to fix. But I can tell you exactly what's wrong

### Expected Behaviour
Popover trigger should launch consistently

### Actual Behaviour
When browser loses focus with popup open, the trigger never works again

Use the popup contents (such as an internal button) to launch a color picker or local file dialog. Even putting a debugger breakpoint on an internal button or similar should be enough to lose focus and repro the bug

#### Platform and Version
0.7.3

So basically, the issue seems to be this:

1. User clicks button to trigger popup

2. Popup is triggered, and has an internal button to launch a file dialog or color picker. Meanwhile, Internally the "click-content" slot is moved out into the document at large

3. User clicks internal popup menu button to activate OS level dialog (color picker or file dialog), which removes focus from the browser. Either clicking this button or interacting with the color/file dialog causes mouseclick events to fire which is handled by the OverlayStack class and automatically closes the overlay.

4. Closing the overlay triggers a sequence of events. The "hideAndCloseOverlay" function is at the heart of the issue here. Inside, there is an async "overlay.hide" function. Internally, this function gracefully hides the menu over time. Being async, it will block cleanup with "overlay.remove()" until "overlay.hide()" completes. 

5. There is a timing issue I can't figure out here. Removing the "overlay.hide()" function fixes the entire situation. I can only imagine its a timing issue.

6. Going deeper, the issue might lie in Active-Overlay. The function "returnOverlayContent()" successfully sets the element "this.overlayContent" slot attribute successfully to "click-content". However, when recreating the bug, "returnOverlayContent()" successfully sets slot attribute, but then immediately runs once more where it doesn't set the slot attribute successfully because "this.originalSlot" is undefined. It's looking like something about this behavior is triggering 2 overlay hides in a row. Perhaps the second hide is triggered while the first await hide animation is in progress and things don't get properly cleaned up?

### LINKED ISSUE #518: <overlay-trigger> Repeatedly clicking on trigger with click-content permanently disables overlay
### Overview

There is a bug where if you use an overlay-trigger, and slot an element into "click-content", and then repeatedly click the "trigger" slotted element in quick succession, it will permanently disable the overlay from appearing.

### Expected Behaviour

 Click-triggered overlays should be able to handle fast repeated clicks.

### Actual Behaviour

 Clicking repeatedly will permanently disable the overlay from appearing.

`overlay-trigger` with `trigger` slotted element and a `click-content` slotted element

#### Steps to Reproduce

Click repeatedly really fast on the `trigger` element

Westbrook:
## Description
Manage the disposal process of an `active-overlay` via its state machine so that steps cannot be run multiple times against the same overlay.

Long term I'd rather this state machine interaction to be more regimented and for the side effects to occur out of checks in the `updated` method, which should go a good way to making this even more predictable. However, there are some properties (like the value of `animated` in the `hide()` method) that aren't managed internally to `active-overlay`, which would need to be updated for this to be true.

## Related Issue
fixes #511 
fixes #518

## Motivation and Context
Overlays should be resilient to various use cases.

## How Has This Been Tested?
- manually built out the use case outlined in #511 and tested it in Storybook.
- manually tested quick clicking an `overlay-trigger` element
  - @msdewey would you want a canary release, or to manually copy some code into your use case to ensure that this update actually covers both contexts?

## Types of changes

## Checklist:


Westbrook:
Realized some errant files had gotten added in a previous PR and now I'm removing them here 😓 sorry.

packages/overlay/test/overlay-trigger.test.ts
```diff
@@ -109,6 +109,14 @@ describe('Overlay Trigger', () => {
         await elementUpdated(testDiv);
     });
 
+    afterEach(async () => {
```
Westbrook: Recently, I get caught by the underlying test issue that this ensures has been cleaned up more and more, so I'm very excited to get this into the repo!

PR #521, opened 2020-03-01
fix(search): ensure "reset" surfaces "input" and "change" events

### LINKED ISSUE #313: Clearing search field does not generate event or clear value
### Expected Behaviour

When you click on the 'x' the search field should send out some sort of event, either `input` or `change` to notify that it has happened. And, accessing the `.value` property on the search field should give you and empty string.

### Actual Behaviour

No event is generated and the `.value` remains as it was before the field was cleared.

#### Steps to Reproduce

- Run the documentation site from pull #312
- Type in a search such as "but"
- click the 'x'

Documentation site from #312

Westbrook:
## Description
Manual dispatch `input` and `change` events from the `input` element in the `Search` component so that an appropriate response can be made in implementing applications.

## Related Issue
fixes #313

## Motivation and Context
Tighter interaction loop.

## How Has This Been Tested?
- unit test

## Types of changes

## Checklist:

### DIFF
--- a/packages/search/src/search.ts
+++ b/packages/search/src/search.ts
@@ -64,13 +64,29 @@ export class Search extends Textfield {
         }
     }
 
-    public reset(): void {
+    public async reset(): Promise<void> {
         /* istanbul ignore if */
         if (!this.form) {
             return;
         }
         this.value = '';
         this.form.reset();
+        await this.updateComplete;
+        this.focusElement.dispatchEvent(
+            new InputEvent('input', {
+                bubbles: true,
+                composed: true,
+            })
+        );
+        // The native `change` event on an `input` element is not composed,
+        // so this synthetic replication of a `change` event must not be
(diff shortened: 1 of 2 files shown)


packages/search/src/search.ts
```diff
@@ -64,13 +64,29 @@ export class Search extends Textfield {
...
         this.form.reset();
+        await this.updateComplete;
+        this.focusElement.dispatchEvent(
+            new InputEvent('input', {
+                bubbles: true,
+                composed: true,
+            })
+        );
+        // The native `change` event on an `input` element is not composed,
+        // so this synthetic replication of a `change` event must not either
```
cuberoot: did you mean "must not *be* either"?

Westbrook: English really is the hardest part of the work we do here... corrected!

PR #538, opened 2020-03-13
fix(tab-list): position the indicator correctly when sized

### LINKED ISSUE #537: sp-tab-list places the active indicator incorrectly
### Expected Behaviour
When set to taller than the `tab` elements and `justify-content: center` the indicator is positioned correctly.

### Actual Behaviour
In this case the indicator is actually positioned from the "centered" position, instead of the top.

## Proposal
Let's add `top: 0` to the `vertical` and `vertical-right` variants to correct this!

Westbrook:
## Description
Adds support for the following type of styles on `sp-tab-list`...
```
            height: 75vh;
            flex-direction: column;
            justify-content: center;
```

## Related Issue
fixes #537

## Motivation and Context
It makes application styling easier.

## How Has This Been Tested?
- added a visual regression.

image

## Types of changes

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,7 +42,7 @@ commands:
             - restore_cache:
                   name: Restore Golden Images Cache
                   keys:
-                      - v1-golden-images-<< parameters.regression_color >>-<< parameters.regression_scale >>-7ff1646ce8e3deef0ca67ea350121658ccf81c49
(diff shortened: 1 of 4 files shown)


Westbrook:
Yep. Remove and then either just get that value from Spectrum CSS or move to the `top` value instead of the `translateY` value as outlined in the ticket on the Spectrum CSS side.

PR #540, opened 2020-03-13
fix(theme): track default theme values dynamically

### LINKED ISSUE #532: [RFC] `sp-theme` "default" resolution
### Expected Behaviour
The `default` values for the `color` and `scale` in `sp-theme` should not be `const`. This would allow for sites not to have to ship "light" and "medium" in that case that they aren't actively used in their designs.

### Actual Behaviour
The `default` values are `const` and can cause errors (particularly in overlays) when attempting to resolve those values in `Theme.themeFragments`. 

## Proposal
Right now all the theme fragments are packed into `Theme.themeFragments` as `{ kind, styles }`, which means that it is difficult to discern how many fragments of a specific `kind` are available at any one time. This currently looks like the following when registering a fragment:
```
static registerThemeFragment(
        name: FragmentName,
        kind: FragmentType,
        styles: CSSResult
    ): void {
        this.themeFragments.set(name, { kind, styles });
    }
```
I'd like to move to a two dimensional map so that it looks more like:
```
static registerThemeFragment(
        name: FragmentName,
        kind: FragmentType,
        styles: CSSResult
    ): void {
        const kindFragments = this.themeFragments.get(kind);
        kindFragments.set(name, styles);
        if (kindFragments.size === 1) {
             kindFragments.set('default', styles);
        }
    }
```
Which would allow us to use the first available styles as the `default` while also offering a user customizable `default` at the same time.

### Side effects
I believe that in regards to the current functionality that this would be 100% an implementation change and that elements that leverage this functionality would go unaffected by the change. However, the following added functionalities would then be available:
- dynamically attainable `default` (automatic and explicit)
- the ability to register new `kind`s of fragments. we'd have the ability to loop through the `kind` dimension and acquire a `default` for each of those. it might be worth thinking about how to manage multiple options in those `kind`s.

### Open questions
- do we want to focus on expanding the `kind` functionality at this time or at a later date?
- currently the `default` values are exported, but I can't find anywhere that they are used, is there functionality that is expected there that we should continue to support?

### Alternatives
We could go without the larger revision here by updating:
```
export const DefaultColor = 'light';
export const DefaultScale = 'medium';
```
to 
```
export let DefaultColor = 'light';
export let DefaultScale = 'medium';
```
and then managing this in `registerThemeFragment` a la:
```
static registerThemeFragment(
        name: FragmentName,
        kind: FragmentType,
        styles: CSSResult
    ): void {
        this.themeFragments.set(name, { kind, styles });
        if (this.kind === 'color') DefaultColor = DefaultColor || name;
        if (this.kind === 'scale') DefaultScale = DefaultScale || name;
    }
```
It's a much smaller lift, but I think it's less flexible over the course. 

Thoughts?

Westbrook:
## Description
Support manually configuring what theme fragments to load in an application by dynamically acquiring the "default" fragment for each theme "type".

## Related Issue
fixes #532 

## Motivation and Context
We shouldn't HAVE to load all the themes ALL the time.

## How Has This Been Tested?
- probably needs more tests if we agree on this path forward.

## Types of changes

## Checklist:

### DIFF
--- a/packages/overlay/src/active-overlay.ts
+++ b/packages/overlay/src/active-overlay.ts
@@ -399,7 +399,6 @@ export class ActiveOverlay extends LitElement {
     }
 
     public renderTheme(content: TemplateResult): TemplateResult {
-        import('@spectrum-web-components/theme');
         const color = this.color as Color;
         const scale = this.scale as Scale;
         return html`
--- a/packages/theme/src/index.ts
+++ b/packages/theme/src/index.ts
@@ -11,9 +11,9 @@ governing permissions and limitations under the License.
 */
 export * from './theme.js';
 
-import './theme-dark.js';
-import './theme-darkest.js';
 import './theme-light.js';
 import './theme-lightest.js';
+import './theme-dark.js';
+import './theme-darkest.js';
 import './scale-medium.js';
 import './scale-large.js';
--- a/packages/theme/src/theme.ts
+++ b/packages/theme/src/theme.ts
@@ -13,9 +13,6 @@ governing permissions and limitations under the License.
 import coreStyles from './theme.css';
 import { CSSResult, supportsAdoptingStyleSheets } from 'lit-element';
 
-export const DefaultColor = 'light';
-export const DefaultScale = 'medium';
-
 declare global {
     interface Window {
         ShadyCSS: {
@@ -40,7 +37,7 @@ declare global {
 }
 
 type FragmentType = 'color' | 'scale' | 'core';
-type FragmentMap = Map<string, { kind: FragmentType; styles: CSSResult }>;
+type FragmentMap = Map<string, { name: string; styles: CSSResult }>;
 export type Color = 'light' | 'lightest' | 'dark' | 'darkest';
 export type Scale = 'medium' | 'large';
 type FragmentName = Color | Scale | 'core';
@@ -51,12 +48,9 @@ export interface ThemeData {
 }
 
 export class Theme extends HTMLElement {
-    private static themeFragments: FragmentMap = new Map();
-    private static defaultFragments: Set<FragmentName> = new Set([
-        'core',
-        'light',
-        'medium',
-    ]);
+    private hasAdoptedStyles = false;
+    private static themeFragmentsByKind: Map<string, FragmentMap> = new Map();
+    private static defaultFragments: Set<FragmentName> = new Set(['core']);
 
     private static templateElement?: HTMLTemplateElement;
 
@@ -65,14 +59,7 @@ export class Theme extends HTMLElement {
     }
 
     get color(): Color {
-        const color = this.getAttribute('color');
-        if (!color) return DefaultColor;
-
-        const colorFragment = Theme.themeFragments.get(color);
-        if (colorFragment && colorFragment.kind === 'color') {
-            return color as Color;
-        }
-        return DefaultColor;
+        return this.getFragementNameByKind('color') as Color;
     }
 
     set color(newValue: Color) {
@@ -80,26 +67,63 @@ export class Theme extends HTMLElement {
     }
 
     get scale(): Scale {
-        const scale = this.getAttribute('scale');
-        if (!scale) return DefaultScale;
-
-        const scaleFragment = Theme.themeFragments.get(scale);
-        if (scaleFragment && scaleFragment.kind === 'scale') {
-            return scale as Scale;
-        }
-        return DefaultScale;
+        return this.getFragementNameByKind('scale') as Scale;
     }
 
     set scale(newValue: Scale) {
         this.setAttribute('scale', newValue);
     }
 
+    private getFragementNameByKind(kind: string): string {
+        const kindFragments = Theme.themeFragmentsByKind.get(
+            kind
+        ) as FragmentMap;
+        /* istanbul ignore if */
+        if (!kindFragments) {
+            throw new Error(`Unknown theme fragment kind '${kind}'`);
+            /* istanbul ignore if */
+        } else if (kindFragments.size === 0) {
+            throw new Error(`No theme fragments of kind '${kind}'`);
+        }
+        const name = this.getAttribute(kind);
+        if (name) {
+            const fragment = kindFragments.get(name);
+            if (fragment) {
+                return name;
+            }
+        }
+        const defaultFragment = kindFragments.get('default');
+        /* istanbul ignore else */
+        if (defaultFragment) {
+            return defaultFragment.name;
+        }
+        throw new Error(
+            `Incorrectly configured theme fragments of kind '${kind}'`
+        );
+    }
+
     private get styles(): CSSResult[] {
-        return [
-            coreStyles,
-            Theme.themeFragment(this.color).styles,
-            Theme.themeFragment(this.scale).styles,
-        ];
+        const themeKinds = [...Theme.themeFragmentsByKind.keys()];
+        const styles = themeKinds.reduce(
+            (acc, kind) => {
+                const kindFragments = Theme.themeFragmentsByKind.get(
+                    kind
+                ) as FragmentMap;
+                const defaultStyles = kindFragments.get('default');
+                // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+                // @ts-ignore
+                const { [kind]: name } = this;
+                const currentStyles = kindFragments.get(name);
+                if (currentStyles) {
+                    acc.push(currentStyles.styles);
+                } else if (defaultStyles) {
+                    acc.push(defaultStyles.styles);
+                }
+                return acc;
+            },
+            [] as CSSResult[]
+        );
+        return [coreStyles, ...styles];
     }
 
     private static get template(): HTMLTemplateElement {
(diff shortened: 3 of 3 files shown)


packages/overlay/src/active-overlay.ts
```diff
@@ -399,7 +399,6 @@ export class ActiveOverlay extends LitElement {
     }
 
     public renderTheme(content: TemplateResult): TemplateResult {
-        import('@spectrum-web-components/theme');
```
Westbrook: Unless you've manually placed yourself of this path, there is no other way to get to this method without having included some part of this package in your application, so there's no reason to force ALL of it on you at this point.

packages/theme/src/index.ts
```diff
@@ -11,9 +11,9 @@ governing permissions and limitations under the License.
 */
 export * from './theme.js';
 
-import './theme-dark.js';
-import './theme-darkest.js';
 import './theme-light.js';
```
Westbrook: 'light' in the default in the tests, so this ensure that it is still `default` by loading it first.

packages/theme/src/theme.ts
```diff
@@ -51,12 +48,9 @@ export interface ThemeData {
...
 export class Theme extends HTMLElement {
-    private static themeFragments: FragmentMap = new Map();
-    private static defaultFragments: Set<FragmentName> = new Set([
-        'core',
-        'light',
-        'medium',
-    ]);
+    private hasAdoptedStyles = false;
+    private static themeFragmentsByKind: Map<string, FragmentMap> = new Map();
+    private static defaultFragments: Set<FragmentName> = new Set(['core']);
```
Westbrook: With `name` in the fragment now, we might be able to generate these values from `themeFragmentsByKind`... 🤔 this is slightly faster than looking it up every time, and this data is predominately used in the polyfilled version, so it's probably better to have a static list and to not worry too much about it as it will go away over time.

packages/theme/src/theme.ts
```diff
@@ -65,41 +59,71 @@ export class Theme extends HTMLElement {
...
+    private getFragementNameByKind(kind: string): string {
+        const kindFragments = Theme.themeFragmentsByKind.get(
+            kind
+        ) as FragmentMap;
+        /* istanbul ignore if */
+        if (!kindFragments) {
+            throw new Error(`Unknown theme fragment kind '${kind}'`);
+            /* istanbul ignore if */
+        } else if (kindFragments.size === 0) {
+            throw new Error(`No theme fragments of kind '${kind}'`);
```
Westbrook: This is probably not needed, as the only place we add theme fragments right now create the kindMap and adds the fragments, but there were `errors` thrown before and I wanted to join the party, even if I'm not good at knowing when to throw what. As this would only be encountered when addressing this element imperatively after having queried for it in the DOM, maybe we can drop it?

packages/theme/src/theme.ts
```diff
@@ -65,41 +59,71 @@ export class Theme extends HTMLElement {
...
+    }
+
     private get styles(): CSSResult[] {
-        return [
-            coreStyles,
-            Theme.themeFragment(this.color).styles,
-            Theme.themeFragment(this.scale).styles,
-        ];
+        const themeKinds = [...Theme.themeFragmentsByKind.keys()];
+        const styles = themeKinds.reduce(
```
Westbrook: This is the guts of ensuring a default.

packages/theme/src/theme.ts
```diff
@@ -65,41 +59,71 @@ export class Theme extends HTMLElement {
...
+            (acc, kind) => {
+                const kindFragments = Theme.themeFragmentsByKind.get(
+                    kind
+                ) as FragmentMap;
+                const defaultStyles = kindFragments.get('default');
+                // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+                // @ts-ignore
```
Westbrook: Is there a way to type this that TS doesn't get angry about?

packages/theme/src/theme.ts
```diff
@@ -133,8 +158,13 @@ export class Theme extends HTMLElement {
     }
 
     protected connectedCallback(): void {
+        /* istanbul ignore if */
+        if (!this.hasAdoptedStyles) {
```
Westbrook: Before we were doing this in the constructor, but it means that it might run before the attributes are set. This should help reduce some edge case extra renders.

packages/theme/src/theme.ts
```diff
@@ -176,26 +209,25 @@ export class Theme extends HTMLElement {
...
-                    styleNodes.forEach((element) => element.remove());
-                }
-                styles.forEach((s) => {
-                    const style = document.createElement('style');
-                    style.textContent = s.cssText;
-                    if (this.shadowRoot) {
-                        this.shadowRoot.appendChild(style);
-                    }
-                });
+        } else if (this.shadowRoot) {
```
Westbrook: Pretty sure this next section was lifted straight from LitElement, but reading it as part of this change it seemed to make better sense with some simplified TS structures.

packages/theme/src/theme.ts
```diff
@@ -65,41 +59,71 @@ export class Theme extends HTMLElement {
...
-        }
-        return DefaultScale;
+        return this.getFragementNameByKind('scale') as Scale;
     }
 
     set scale(newValue: Scale) {
         this.setAttribute('scale', newValue);
     }
 
+    private getFragementNameByKind(kind: string): string {
```
benjamind: Typo

packages/theme/src/theme.ts
benjamind: Something like this could work:

```diff
diff --git a/packages/theme/src/theme.ts b/packages/theme/src/theme.ts
index a841e5c1..9a8f67d8 100644
--- a/packages/theme/src/theme.ts
+++ b/packages/theme/src/theme.ts
@@ -47,9 +47,16 @@ export interface ThemeData {
     scale?: Scale;
 }
 
-export class Theme extends HTMLElement {
+type ThemeKindProvider = {
+    [P in FragmentType]: string;
+};
+
+export class Theme extends HTMLElement implements ThemeKindProvider {
     private hasAdoptedStyles = false;
-    private static themeFragmentsByKind: Map<string, FragmentMap> = new Map();
+    private static themeFragmentsByKind: Map<
+        FragmentType,
+        FragmentMap
+    > = new Map();
     private static defaultFragments: Set<FragmentName> = new Set(['core']);
 
     private static templateElement?: HTMLTemplateElement;
@@ -58,6 +65,10 @@ export class Theme extends HTMLElement {
         return ['color', 'scale'];
     }
 
+    get core(): 'core' {
+        return 'core';
+    }
+
     get color(): Color {
         return this.getFragementNameByKind('color') as Color;
     }
@@ -74,7 +85,7 @@ export class Theme extends HTMLElement {
         this.setAttribute('scale', newValue);
     }
 
-    private getFragementNameByKind(kind: string): string {
+    private getFragementNameByKind(kind: FragmentType): string {
         const kindFragments = Theme.themeFragmentsByKind.get(
             kind
         ) as FragmentMap;
@@ -110,8 +121,7 @@ export class Theme extends HTMLElement {
                     kind
                 ) as FragmentMap;
                 const defaultStyles = kindFragments.get('default');
-                // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
-                // @ts-ignore
+
                 const { [kind]: name } = this;
                 const currentStyles = kindFragments.get(name);
                 if (currentStyles) {
```

i.e. provide a mapped type to declare that we will have a property for each fragment kind, implement that type, and then change the fragment map to use it.

packages/theme/src/theme.ts
benjamind: So you like `reduce` huh? :-)

packages/theme/src/theme.ts
Westbrook: That did the trick with one extra typing in there, thanks! TS is crazy.

packages/theme/src/theme.ts
```diff
@@ -211,17 +254,18 @@ export class Theme extends HTMLElement {
...
-
-    static themeFragment(
-        name: FragmentName
-    ): { kind: FragmentType; styles: CSSResult } {
-        const fragment = this.themeFragments.get(name);
-        if (!fragment) {
-            throw new Error(`Unknown theme fragment '${name}'`);
+        if (!this.themeFragmentsByKind.has(kind)) {
+            this.themeFragmentsByKind.set(kind, new Map());
+            (this.themeFragmentsByKind.get(kind) as FragmentMap).set(
```
benjamind: Not sure you need these casts anymore?

Westbrook: The default type of `this.themeFragmentsByKind.get(kind)` is `FragmentMap | null`, so as best as I understand TS I either need this or to set a variable and test its presence (both here and below), but that doesn't compile away, right? Is there a nicer in between version?

benjamind: Ah I see. You could use the non-null operator, or you could just write this a little, something like this seems easier to read and maintain, not sure how it stacks up in terms of compiled byte size though:

```typescript

        let fragmentMap = this.themeFragmentsByKind.get(kind);
        if (!fragmentMap) {
            fragmentMap = new Map();
            this.themeFragmentsByKind.set(kind, new Map());
            // we're adding our first fragment for this kind set as default
            fragmentMap.set('default', { name, styles });
            Theme.defaultFragments.add(name);
        }
        fragmentMap.set(name, {
            name,
            styles,
        });
```

PR #542, opened 2020-03-15
docs: process TS before building the site

Westbrook:
## Description 
When you run `tsc` or the `ts-loader` equivalent with the full `searchIndex.json` file that is exported from `lunr.js` the process takes 60s or more longer to export the final files than if the file simply contains `{}`. This PR ensures that a blank version of `searchIndex.json` is available when running `tsx`, and then runs `webpack` against the exported JS (which means all of the imports need to use `.js` extensions) so that out build process can reap the 60s back when run both locally and on CI. Somewhere in there we also score a couple of KB of JS back in the final build, so even better.

## Motivation and Context
Everything should be faster, always.

## How Has This Been Tested?
Runs in CircleCI, also runs in Netlify ([demo site](https://westbrook-docs-processing--stupefied-franklin-07fec4.netlify.com/components/theme))...still reviewing whether we can build fast enough to leverage that on PRs again or not.

## Types of changes
- [x] build tooling

## Checklist:


documentation/src/components/layout.ts
```diff
@@ -28,11 +28,13 @@ import '@spectrum-web-components/menu-item';
 
 const SWC_THEME_COLOR_KEY = 'swc-docs:theme:color';
 const SWC_THEME_SCALE_KEY = 'swc-docs:theme:scale';
-const COLOR_LIGHT = 'light';
+const COLOR_FALLBACK = matchMedia('(prefers-color-scheme: dark)').matches
+    ? 'dark'
+    : 'light';
```
cuberoot: Cool. I learned something here...

PR #547, opened 2020-03-17
fix(textfield): update validation path, add "allowed-keys"

### LINKED ISSUE #211: Investigate more in depth pattern and validation processes for sp-textfield
### Expected Behaviour
We can control what things get set into a textfield

### Actual Behaviour
The field will go invalid, but not prevent the input.

```
<sp-textfield
                        label="width"
                        pattern="[0-9]+([.]?[0-9]*)?"
                    ></sp-textfield>
```
If you put in letter in the above field it will go `invalid` but won't prevent you from doing so.

### LINKED ISSUE #476: Separate "pattern" from "valid" in sp-textfield
### Expected Behaviour
Using a `pattern` doesn't automatically validate an `sp-textfield`.

### Actual Behaviour
Currently any `sp-textfield` with a `pattern` will show a X or check when content is available.

### Note
Looks like native inputs do this via the `required` attribute: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/text#Making_input_required

Westbrook:
## Description
`required` means that a value _must_ be present.
`pattern` will verify that a value (when present) matches that pattern
`allowed-keys` will populate a regex to prevent input when the value is incorrect.

## Related Issue
fixes #476
fixes #211

## Motivation and Context
Don't always want validation UI.

## How Has This Been Tested?
- new tests!

## Types of changes

## Checklist:

### DIFF
--- a/packages/textfield/src/textfield.ts
+++ b/packages/textfield/src/textfield.ts
@@ -43,6 +43,9 @@ export class Textfield extends Focusable {
         ];
     }
 
+    @property({ attribute: 'allowed-keys' })
+    allowedKeys = '';
+
     @query('#input')
     private inputElement!: HTMLInputElement | HTMLTextAreaElement;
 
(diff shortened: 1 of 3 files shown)


bradee:
yay!

PR #570, opened 2020-03-26
fix(slider): use internal "input" for value sanitation

### LINKED ISSUE #481: The value of an "sp-slider" can display outside of it's restraints
### Expected Behaviour
The display value of an `sp-slider` should stick to its restraints.

### Actual Behaviour
image

Westbrook:
## Description
- Use the `input` element internal to the slider to sanitize values to avoid JS quirks in rounding values to match `step` management.
- Reduce the number of calls to `this.getBoundingClientRect()` as it can cause unnecessary layouts.

## Related Issue
fixes #481 

## Motivation and Context
Because `12.120000000000001` doesn't conform to `step="0.01"`...

## How Has This Been Tested?
Tests run just fine.

## Types of changes

## Checklist:

### DIFF
--- a/packages/slider/src/slider.ts
+++ b/packages/slider/src/slider.ts
@@ -40,12 +40,16 @@ export class Slider extends Focusable {
 
     public set value(value: number) {
         const oldValue = this.value;
+        if (this.input) {
+            this.input.value = String(value);
+        }
+        const newValue = this.input ? parseFloat(this.input.value) : value;
 
-        if (value === oldValue) {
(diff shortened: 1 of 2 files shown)


packages/slider/src/slider.ts
```diff
@@ -299,6 +303,7 @@ export class Slider extends Focusable {
         if (this.disabled) {
             return;
         }
+        this.boundingClientRect = this.getBoundingClientRect();
```
andrewhatran: Nice!

PR #575, opened 2020-03-27
feat: add workflow icons package

### LINKED ISSUE #155: Process more Spectrum CSS Icons
### Expected Behaviour
All of the icons in Spectrum CSS are processed into our project

### Actual Behaviour
Currently only `node_modules/@adobe/spectrum-css/icons/*` files are being parsed to icons, which misses `node_modules/@adobe/spectrum-css/dist/icons/spectrum-icons.svg` which includes another set of icons that are needed to display patterns like `<sp-action-menu />`.

### NPM Source
https://www.npmjs.com/package/@adobe/spectrum-css-workflow-icons

Westbrook:
## Description
These are the JS based icons discussed oh so long ago: https://github.com/adobe/spectrum-web-components/issues/155#issuecomment-539776450 While this is a JS requisite solution, it can be tree shaken and support a TON of icons (880) and it pretty easy to keep up with what is the most "official" of icon releases to the public at Adobe.

## Related Issue
fixes #155

## Motivation and Context
Who doesn't love icons?

image

## Types of changes

## Checklist:


Westbrook:
@andrewhatran, I'd love it if you could take a deeper look at this beyond the fix to the package.json we worked out earlier, this will be a great add to the project.

@adixon-adobe, understand if you're still swamped, I'd take another high lever approval, too, but I added some things around usage of this in the bundle package that I had forgotten.

Thanks!

packages/bundle/README.md
```diff
@@ -11,3 +11,17 @@ npm install @spectrum-web-components/bundle
 
 yarn add @spectrum-web-components/bundle
 ```
+
+### Icons - Workflow
+
+While this bundle directly re-exports the majority of functionality as they would be exported from their own packages, `@spectrum-web-components/icons-workflow` is renamed to `IconsWorkflow` when leveraging the bundle. This means that you can use workflow icons in your demonstration code that using `@spectrum-web-components/bundle` like the following:
+
```
andrewhatran, line 18: Possible minor type, did you mean `demonstration code that uses @spectrum-web-components/bundle`?

Westbrook: Good call! I found a version I liked a little bit better, but thanks for calling this out.

packages/icons-workflow/README.md
```diff
@@ -0,0 +1,106 @@
...
+```
+
+### Usage
+
+With the default exports of the packages prepared with the `html` template tag from `lit-html`, the default value of an icon export will be as follows:
+
+```js
+import { LitElement, html } from 'lit-element';
+import '@spectrum-web-components/icon';
+import { Circle } from '@spectrum-web-components/icons-workflow';
```
adixon-adobe, line 22: I think it'd be kind of nice if the icon exports had icon in the name. E.g. `SaveIcon` is a lot less ambiguous than `Save`.

packages/icons-workflow/README.md
```diff
@@ -0,0 +1,106 @@
+## Description
+
+Spectrum Workflow Icons delivered in a mutable template tag so that they can be leveraged across various frameworks. The default export of this package pre-applies the `html` template tag from `lit-html` for ease of use in the Spectrum Web Components library. Continue reading below for information on using these icons in projects powered by other frameworks.
```
adixon-adobe, line 3: I'd link out to https://spectrum.adobe.com/page/icons/ in this description -- I'd expect folks new to spectrum to be confused otherwise.

packages/icons-workflow/README.md
Westbrook: @andrewhatran having looked at the PR as well, do you have thoughts no which of the naming approaches would feel most easy to leverage for you? I know we've used the pattern as outlined currently no our team, but don't let that color you're response.

andrewhatran: I sorta like the idea of suffixing our exports with `Icon`. 

Especially since we have icon names like `App`. 

If I saw this for the first time, it wouldn't be clear to me if `App` were an icon or a high-level web component:

```
render(): TemplateResult {
    return html`
         ${App()}
    `
}
```

Westbrook: That makes perfect sense. Thanks for sharing a good example! Updating.

Westbrook:
@adixon-adobe and @andrewhatran thanks for the thoughts on this, I've made updates on all the notes, looking forward to getting this out there!

Issue #593, opened 2020-04-02
process-spectrum should manage "@media" rules


Westbrook:
### Expected Behaviour
`@media` rules are processed correctly when available in origin Spectrum CSS files

### Actual Behaviour
The are currently skipped.

### Reproduce Scenario (including but not limited to)
Process the `Button` CSS and compare.

Westbrook:
Released.

PR #594, opened 2020-04-02
feat: add coachmark pattern

### LINKED ISSUE #300: process-spectrum should manage CSS animations
### Expected Behaviour
Processing `circleloader` automatically ports over the `spectrum-fill-mask-1`, `spectrum-fill-mask-2`,  and `spectrum-fills-rotate` animations.

### Actual Behaviour
They are currently moved manually.

### LINKED ISSUE #564: Add Spectrum Coach Mark Pattern
### Expected Behaviour
https://opensource.adobe.com/spectrum-css/components/coachmark/ should be available in the library

### Actual Behaviour
missing

#### The future
image

Westbrook:
## Description
Add the [Spectrum Coachmark](https://opensource.adobe.com/spectrum-css/components/coachmark/) pattern, excluding the "popover" interface for now as it's pretty related to the Spectrum Dialog pattern that is also currently in development.

Update Spectrum CSS processing to allow for the adoption of `@keyframes` rules from the origin CSS. In the name of "If you give a mouse a cookie...", attempting to support these pointed out the lack of support for `@media` rules which are a different level of complexity and are now being tracked via: #593 

## Related Issue
fixes #564 
fixes #300

## Motivation and Context
Expand the component ecosystem and simplify the creation of future packages.

## How Has This Been Tested?
This is currently a visual only component that leverages animations so both unit tests and visual regression tests in their current form will not accurately support testing this component at this time. However, the three empty `<div>`s are still tested for their "accessibility" 😏 

image

## Types of changes

## Checklist:


scripts/process-spectrum-postcss-plugin.js
```diff
@@ -401,6 +401,11 @@ class SpectrumProcessor {
             ],
         });
 
+        root.walkAtRules((atRule) => {
```
Westbrook, line 404: @cuberoot this _seems_ like all the magic needed to get the `@keyframes` rules in, since they are pretty contained to feature specific syntax inside of themselves. The `@media` issue that I added would need to step into the styles that are shipped to ensure that they are required and processed correctly, which is why I left them out for right now.

packages/circleloader/src/spectrum-circleloader.css
```diff
@@ -10,6 +10,378 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 
 THIS FILE IS MACHINE GENERATED. DO NOT EDIT */
+@keyframes spectrum-fill-mask-1 {
```
Westbrook, line 13: Now that the processor pulls this in we are able to remove this code from the manually built file above.

PR #596, opened 2020-04-02
refactor: use mixin to manage shared anchor like properties

### LINKED ISSUE #595: Elements with "href" in their API should support "download"
### Expected Behaviour
i have a dropdown with a `<sp-menu-item href="whatever" download>` and i'd like that to download instead of just link - @lazd 

### Actual Behaviour
While `<sp-link>` supports `download`, `<sp-sidenav-item>` and nothing extending `button-base.ts` currently support it, even when they support the reset of the API available in `Link`.

## Path forward
Create the following interface as a mixin:
```
export interface LikeAnchorInterface {
    download?: string;
    label?: string;
    href?: string;
    target?: '_blank' | '_parent' | '_self' | '_top';
    renderAnchor(options: {
        id: string;
        anchorContent?: TemplateResult | TemplateResult[];
    }): TemplateResult;
}
```
Use it in the base of the elements requiring anchor-like functionality.

### Questions
`<sp-sidenav-item>` has some additional requirements on the `<a>` internal to it that the other patterns do no rely on: custom `href` application, `tabindex`, `data-level` (this could technically be moved in the element so as not to collide so directly with other uses of the mixin), and `@click`. I think that places it as a valid usage of the escape valve of not leveraging `renderAnchor` in that component, but I could be convinced it means either that `renderAnchor` is not robust enough to serve this overall pattern, or that the overall pattern itself is flawed.

Should instead the signature of `renderAnchor` be as follows?
```
renderAnchor(options: {
    id?: string;
    href?: string;
    click(): void;
    tabindex?: string,
    target?: '_blank' | '_parent' | '_self' | '_top';
    label?: string,
    download?: boolean,
    anchorContent?: TemplateResult | TemplateResult[];
}): TemplateResult;
```
This leaving the possibility that the options bag might expand with time.

Should the mixin only surface properties for the individual components to do with as they'd like? A middle ground of centralized properties but decentralized usage.

Should all these components list these properties themselves?

Westbrook:
## Description
Add the `LikeAnchor` mixin with a relatively general purpose `renderAnchor` method knowing that in some cases the elements that this is mixed into will leverage their own rendering to define the DOM of the anchor.

## Related Issue
fixes #595 please see this issue for further discussion

## Motivation and Context
- Less bumpy API
- Support for early adopters

## How Has This Been Tested?
Refactor passes current tests.

## Types of changes
- [x] refactor

## Checklist:

### DIFF
--- a/packages/shared/src/like-anchor.ts
+++ b/packages/shared/src/like-anchor.ts
@@ -1,0 +1,68 @@
+/*
+Copyright 2020 Adobe. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License. You may obtain a copy
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+import { UpdatingElement, property, TemplateResult, html } from 'lit-element';
+import { ifDefined } from 'lit-html/directives/if-defined.js';
+
+type Constructor<T = object> = {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    new (...args: any[]): T;
+    prototype: T;
+};
+
+export interface LikeAnchorInterface {
+    download?: string;
+    label?: string;
+    href?: string;
+    target?: '_blank' | '_parent' | '_self' | '_top';
+    renderAnchor(options: {
+        id: string;
+        anchorContent?: TemplateResult | TemplateResult[];
+    }): TemplateResult;
+}
+
+export function LikeAnchor<T extends Constructor<UpdatingElement>>(
+    constructor: T
+): T & Constructor<LikeAnchorInterface> {
+    class LikeAnchorElement extends constructor {
+        @property({ reflect: true })
+        public download?: string;
+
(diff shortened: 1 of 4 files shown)


packages/shared/src/like-anchor.ts
```diff
@@ -0,0 +1,68 @@
...
+export interface LikeAnchorInterface {
+    download?: string;
+    label?: string;
+    href?: string;
+    target?: '_blank' | '_parent' | '_self' | '_top';
+    renderAnchor(options: {
+        id: string;
+        anchorContent?: TemplateResult | TemplateResult[];
+    }): TemplateResult;
+}
```
adixon-adobe, line 30: Out of curiosity why is this interface needed? It feels like there's some intent here that would be good to add in a comment (or a common pattern being used I'm just ignorant on).

Westbrook: Namely that it's needed in like 34 below... and so that the properties and methods therein can be seen by elements this is mixed into. Other than that it probably doesn't need to exported, but I was matching style to the `ObserveSlotText` mixin pattern. I don't have enough Typescript magic to say whether there is other intent that this could fill or not.

benjamind: Yep you have it down. Its just to enhance the type exported by the mixin so that the constructed object has these properties. Probably doesn't need export its true, but might be handy in some cases where we want to accept anything that is 'anchor like'?

adixon-adobe: Ahhh -- thanks.

lazd:
This is awesome! Nice work!

PR #615, opened 2020-04-13
Add the Spectrum Rule pattern

### LINKED ISSUE #289: Implement `sp-rule` component
### Expected Behaviour
We shouldn't have to rely on Spectrum CSS directly for our documentation site. We're pointing to the `rule` sheet, we should have a component or a pattern that supports including that in a page easily.

### Actual Behaviour
https://opensource.adobe.com/spectrum-css/components/rule/

Westbrook:
## Description
- remove babel processing in exchange for manual `process.env` management
- Adds the `<sp-rule>` element and associated documentation/tests.
- includes `<sp-rule>` in the docs site in relation to `<h2>` tags as seen in the Spectrum CSS site

## Related Issue
fixes #289 

## Motivation and Context
Who doesn't love more?

## How Has This Been Tested?
- unit tests
- visual regressions

image
image
image
image
image

## Types of changes

## Checklist:


packages/rule/stories/rule.stories.ts
```diff
@@ -0,0 +1,236 @@
...
+export default {
+    title: 'Rule',
+    component: 'sp-rule',
+    decorators: [
+        (story): TemplateResult => html`
+            <style>
+                sp-theme {
+                    color: var(--spectrum-global-color-gray-800);
+                }
+                .spectrum-Heading--subtitle1 {
```
cuberoot, line 32: Kind of annoying to have to recreate these here.

Westbrook: Yeah, it has to do with needing them out of the `node_modules` scope.

I could try again to include them in Storybook's `preview.js`, but I'm pretty sure that while using the version from Open WC we don't get magic compiling there and it'll choke.

Another option is to build a demo element that has `@spectrum-web-components/styles/typography.css` in it's styles and accepts a TemplateResult, a la:
```
<typography-demo
    .content=${html`
        <h2 class="spectrum-Heading--subtitle1">Large</h2>
        <sp-rule size="large"></sp-rule>
        <p class="spectrum-Body">Page or Section Titles.</p>
    `}
></typography-demo>
```
Could keep that in the decorator, and allow the stories to go unchanged... 🤔 That's more interesting then when I thought it through before. I'll try again.

Westbrook: This pattern is definitely cleaner, but it might start to open a separate can of worms around having more of the content in `@spectrum-web-components/styles` available as `CSSResultsArrays`, which is a sledgehammer of a solution to styling that it probably best we don't actively support. I don't have enough use cases to fell conclusively on this, however, as we aren't currently using a lot of typography or styles beyond the components themselves in our project.

@cuberoot Does this seem more like what you were expecting here?

cuberoot: Yah, that seems better. At least we don't have to cut-and-paste to use the typography to use it in our stories. We can think about how we would want to expose typography in future...

PR #618, opened 2020-04-13
fix: ensure themes are registered before adopting

Westbrook:
## Description
Ensure that `adoptStyles()` is called _after_ theme fragments have been registered by blocking style adoption until three distinct fragments are available.

## Motivation and Context
PacelJS and native imports will cause the element to be registered before the fragments are registered causing the fragments to go unadopted without implementor interaction.

## How Has This Been Tested?
- I'll take suggestions.

## Types of changes

## Checklist:

### DIFF
--- a/packages/theme/src/theme.ts
+++ b/packages/theme/src/theme.ts
@@ -61,6 +61,8 @@ export class Theme extends HTMLElement implements ThemeKindProvider {
 
     private static templateElement?: HTMLTemplateElement;
 
+    private static instances: Set<Theme> = new Set();
+
     static get observedAttributes(): string[] {
         return ['color', 'scale'];
     }
@@ -131,7 +133,7 @@ export class Theme extends HTMLElement implements ThemeKindProvider {
             }
             return acc;
         }, [] as CSSResult[]);
-        return [coreStyles, ...styles];
+        return [...styles];
     }
 
     private static get template(): HTMLTemplateElement {
@@ -168,20 +170,34 @@ export class Theme extends HTMLElement implements ThemeKindProvider {
     }
 
     protected connectedCallback(): void {
-        /* istanbul ignore if */
-        if (!this.hasAdoptedStyles) {
-            this.adoptStyles();
-        }
+        this.shouldAdoptStyles();
         // Note, first update/render handles styleElement so we only call this if
         // connected after first update.
         /* istanbul ignore if */
         if (window.ShadyCSS !== undefined) {
             window.ShadyCSS.styleElement(this);
         }
+        // Add `this` to the instances array.
+        if (!Theme.instances.has(this)) {
+            Theme.instances.add(this);
+        }
+    }
+
+    protected disconnectedCallback(): void {
+        // Remove `this` to the instances array.
+        Theme.instances.delete(this);
+    }
+
+    private shouldAdoptStyles(): void {
+        /* istanbul ignore if */
+        if (!this.hasAdoptedStyles) {
+            this.adoptStyles();
+        }
     }
 
     protected adoptStyles(): void {
         const styles = this.styles; // No test coverage on Edge
+        if (styles.length < 3) return;
 
         // There are three separate cases here based on Shadow DOM support.
         // (1) shadowRoot polyfilled: use ShadyCSS
@@ -241,6 +257,7 @@ export class Theme extends HTMLElement implements ThemeKindProvider {
     }
 
     protected attributeChangedCallback(): void {
+        this.hasAdoptedStyles = false;
         if (window.ShadyCSS !== undefined && !window.ShadyCSS.nativeShadow) {
             window.ShadyCSS.styleElement(this);
         } else {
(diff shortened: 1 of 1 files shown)


packages/theme/src/theme.ts
```diff
@@ -61,6 +61,8 @@ export class Theme extends HTMLElement implements ThemeKindProvider {
 
     private static templateElement?: HTMLTemplateElement;
 
+    private static instances: Theme[] = [];
```
Westbrook, line 64: Is this an OK way to cache all the instances? It's sort of leaky... Should I use a weak map? I like better ideas.

packages/theme/src/theme.ts
```diff
@@ -261,6 +265,7 @@ export class Theme extends HTMLElement implements ThemeKindProvider {
             Theme.defaultFragments.add(name);
         }
         fragmentMap.set(name, { name, styles });
+        Theme.instances.map((instance) => instance.adoptStyles());
```
Westbrook, line 268: Should we block this after `this.hasAdoptedStyles === true`?

packages/theme/src/theme.ts
```diff
@@ -182,6 +185,7 @@ export class Theme extends HTMLElement implements ThemeKindProvider {
 
     protected adoptStyles(): void {
         const styles = this.styles; // No test coverage on Edge
+        if (styles.length < 3) return;
```
Westbrook, line 188: If this general approach works, does `// ensure all three fragments types are loaded before adopting styles`?

Or should this effect the setting of `this.hasAdoptedStyles`? Maybe `this.hasAdoptedStyles = styles.length > 3;` would allow us to stream themes into the UI, but block after there's something available for all three?

packages/theme/src/theme.ts
benjamind: You need to be able to iterate the entries, so I don't think you can use a weakset/map here.

benjamind: Probably want the theme instalces to add/remove themselves as they are connected/disconnected instead?

cuberoot: I don't think you can iterate over weak maps, so I don't think that's an option.

packages/theme/src/theme.ts
benjamind: Does this need protecting? I'd be tempted to just make `Theme.instances` a `Set<Theme>`, add to the set in connectedCallback, and delete from the set in disconnectedCallback. Then in `registerThemeFragment` just call `adoptStyles` for each instance as you did below. Or is there some problem with calling adopt styles several times?

benjamind: I guess the other approach would be to actually evalulate if the current set of reigstered styles satisfies the theme and only call adopt once it does?

packages/theme/src/theme.ts
Westbrook: Makes sense. Will update the add/remove functionality.

packages/theme/src/theme.ts
Westbrook: `adoptStyles` scales up in processing cost depending on the support the browser has for various APIs. I'd like to skip it until it's ready, but it _could_ just run.

`styles.length === 3` is technically the easiest way to actually evaluate if the current set of registered styles satisfies the theme right now. Unless we want to replay the `reduce` twice on the last instance, it seemed like a good idea to use the single call for both meaning.

benjamind: Then yeah I guess this check here is sufficient.

packages/theme/src/theme.ts
benjamind: If the theme requirements (color / scale) change dynamically, and the theme fragments are still being late loaded, would we need to re-adopt?

Westbrook: Hrm...I'll have to think about that a little. Currently, in `attributeChangedCallback` we re-adopt...if that were to be associated with further lazy loading, you _could_ reset `this.hasAdoptedStyles = false;` until the theme was satisfied again...thoughts?

packages/theme/src/theme.ts
```diff
@@ -168,20 +170,33 @@ export class Theme extends HTMLElement implements ThemeKindProvider {
...
             window.ShadyCSS.styleElement(this);
         }
+        // Add `this` to the instances array.
+        Theme.instances.push(this);
+    }
+
+    protected disconnectedCallback(): void {
+        const instanceIndex = Theme.instances.indexOf(this);
+        // Remove `this` to the instances array.
+        Theme.instances.splice(instanceIndex, 1);
```
benjamind, line 187: Maybe consider just using `Set`, since then you don't need to iterate the index and can just add/delete from the set.

packages/theme/src/theme.ts
```diff
@@ -131,7 +133,7 @@ export class Theme extends HTMLElement implements ThemeKindProvider {
             }
             return acc;
         }, [] as CSSResult[]);
-        return [coreStyles, ...styles];
+        return [...styles];
```
benjamind, line 136: Was this an intended change?

Westbrook: Yep. I was doing a test of the set and realized we were adding that style twice 🙈 so the logic actually made that length that we test whether to adopt the styles or not work different than intended.

Westbrook:
I tested it by copying the output into a Parcel app and into a raw es-dev-server app (both of which were failing) and now both are working as expected.

benjamind:
Ok good to go then!

Issue #630, opened 2020-04-16
The tab-list component throws an error on Edge


cuberoot:
### Expected Behaviour

The `sp-tab-list` should work on pre-chromium Edge with the correct polyfills

### Actual Behaviour

The `sp-tab-list` throws an error because it assumes that `document.fonts.ready` exists

#### Steps to Reproduce

Just load the docs in Edge: https://opensource.adobe.com/spectrum-web-components/

#### Platform and Version

Pre-chromium Edge on Windows

Westbrook:
Fixed by #632

PR #632, opened 2020-04-16
fix: make tab-list work on Edge

cuberoot:
- Be forgiving if `document.fonts` does not exist
- Work around issue with lit-element and event handlers on `<slot>` components

## Related Issue

#630 

## Motivation and Context

We have a use case for spectrum-web-components on pre-chromium Edge

## How Has This Been Tested?

Manual testing in the documentation application

## Types of changes

## Checklist:


packages/tab-list/src/tab-list.ts
```diff
@@ -87,8 +99,6 @@ export class TabList extends Focusable {
     protected render(): TemplateResult {
         return html`
             <slot
-                @click=${this.onClick}
-                @keydown=${this.onKeyDown}
```
cuberoot, line 91: This makes me sad, but somehow lit-element and ShadyDOM don't work well enough together to make these event handlers work. The ShadyDOM maintainers [fixed this on their end](https://github.com/webcomponents/shadydom/issues/252) but lit-element didn't work.

packages/tab-list/src/tab-list.ts
```diff
@@ -28,6 +29,14 @@ const availableArrowsByDirection = {
     horizontal: ['ArrowLeft', 'ArrowRight'],
 };
 
+declare global {
+    interface Document {
+        fonts?: {
+            ready: Promise<void>,
```
Westbrook, line 35: For uniformity, can you move the following from the `dis/connecedCallback`s up here as well?
```
removeEventListener: (
    name: string,
    callback: () => void
) => void;
```

cuberoot:
Thanks for that. I shouldn't rush PRs at the end of the day. I am leaning towards adding the extra `div` and removing the manual event handling. My issue with adding the listeners is that the `@query('slot')` isn't resolved until `firstUpdated`

packages/tab-list/src/tab-list.ts
```diff
@@ -75,6 +83,15 @@ export class TabList extends Focusable {
             this.querySelector('sp-tab')) as Tab;
     }
 
+    constructor() {
+        super()
+
+        // These can be added as @click and @keydown handlers on the
+        // slot once we no longer need web component polyfills
+        this.addEventListener('click', this.onClick);
+        this.addEventListener('keydown', this.onKeyDown);
```
cuberoot, line 92: Instead of trying to add these to the slot, I realised I can just add them to the element itself and then I don't have to worry about cleanup. You may have been trying to tell me that 🤦

packages/tab-list/src/tab-list.ts
cuberoot: I tried this, but it blew up the CSS pretty badly.

packages/tab-list/src/tab-list.ts
```diff
@@ -240,9 +255,8 @@ export class TabList extends Focusable {
             return;
         }
         await Promise.all([
-            await selectedElement.updateComplete,
-            await ((document as unknown) as { fonts: { ready: Promise<void> } })
-                .fonts.ready,
+            selectedElement.updateComplete,
+            document.fonts ? document.fonts.ready : Promise.resolve(),
```
cuberoot, line 259: Fonts might be right on pre-chromium edge, but I can live with that. We just want Edge to be functional.

PR #642, opened 2020-04-22
fix(tab): use Spectrum CSS relative values for overrides

### LINKED ISSUE #123: Base `<sp-tab vertical />` on Spectrum CSS variables.
### Expected Behaviour
The vertical view's layout should be based on Spectrum CSS variables.

### Actual Behaviour
It uses custom styles applied in #111

Westbrook:
## Description
Ensure local customization is based as close as possible on Spectrum CSS based custom properties. Add a new visual regression to manage.

## Related Issue
fixes #123 

## Motivation and Context
Maintainability.

## How Has This Been Tested?
Visually and with regression tests.

### Before
image

### After
image

## Types of changes

## Checklist:

### DIFF
--- a/packages/tab/src/tab.css
+++ b/packages/tab/src/tab.css
@@ -17,13 +17,30 @@ governing permissions and limitations under the License.
     flex-direction: column;
     justify-content: center;
     align-items: center;
-    height: auto;
+    height: var(--spectrum-tabs-vertical-item-height, auto) !important;
+
+    --sp-tab-vertial-margin-y: calc(
+        (
+                var(
+                        --spectrum-tabs-vertical-item-height,
+                        var(--spectrum-global-dimension-size-550)
+                    ) -
+                    var(
+                        --spectrum-tabs-focus-ring-height,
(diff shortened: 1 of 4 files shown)


packages/tab/src/tab.css
```diff
@@ -17,13 +17,30 @@ governing permissions and limitations under the License.
     flex-direction: column;
     justify-content: center;
     align-items: center;
-    height: auto;
+    height: var(--spectrum-tabs-vertical-item-height, auto) !important;
```
andrewhatran, line 20: Does this need to be important? The height of `tab` is also set in `spectrum-tab.css` inside the `:host` selector.

I think that his selector has a higher specificity than `:host`?

Westbrook: This is actually used to over come the `:host([direction=vertical]) ::slotted(*)` selector in `sp-tab-list`: 
https://github.com/adobe/spectrum-web-components/blob/master/packages/tab-list/src/spectrum-tab-list.css#L138-L143

Issue #279, opened 2019-11-04
`sp-card` doesn't fully support Spectrum CSS delivery


Westbrook:
### Expected Behaviour
Component is implemented based on the full Spectrum CSS import.

### Actual Behaviour
The following classes aren't processed:
- `.spectrum-Card--small`
- `.spectrum-Card-quickActions`
- `.spectrum-Card-actionButton`
- `.spectrum-Card-quickActions .spectrum-Checkbox`
- `.is-focused`

The `:focus` context isn't managed correctly in the CSS/`render()` method.

### Reproduce Scenario (including but not limited to)
- visit the stories for the component in Storybook
- attempt to focus on any of the `sp-card` implementations.

Westbrook:
These notes from the Spectrum team should be included in this work:
> Title should be bold, sentence case.
Footer should be 64 px tall including the divider and bottom border.
There should be a fade-in transition between states.
It looks like this card is using light UI colors but is placed on a solid white background which is confusing. Could we change the background to light UI (gray 100)?
The card title is 5 px too high on standard cards, and 4 px too high on quiet cards. standard card card needs to be made taller to accommodate 5 more px between the preview and the title.
Quiet and gallery card should have a hover state where a 1 px border gray 400 appears.

PR #643, opened 2020-04-22
fix(card): correctly apply :focus-visible styling to variants

Westbrook:
## Description
- correctly process `.spectrum-Card--gallery.focus-ring` to `:host([variant='gallery']:focus-visible)` instead of `:host([variant='gallery']) :focus-visible`
- style slot wrappers instead of slots so we can accurate styles `:before` elements
- apply `tabindex` and `role` for more accurate accessibility

## Related Issue
refs #279 

## Motivation and Context
- better adherence to the Spectrum specification.
- higher quality accessibility

## How Has This Been Tested?
Visually.

image
image
image

*Corrected hover styles for quiet switches*
image
was
image

## Types of changes

## Checklist:

### DIFF
--- a/packages/card/src/card.ts
+++ b/packages/card/src/card.ts
@@ -16,7 +16,9 @@ import {
     property,
     CSSResultArray,
     TemplateResult,
+    PropertyValues,
 } from 'lit-element';
+import { FocusVisiblePolyfillMixin } from '@spectrum-web-components/shared/lib/focus-visible.js';
 
 import cardStyles from './card.css.js';
 
@@ -27,13 +29,16 @@ import cardStyles from './card.css.js';
  * @slot description - A description of the card
  * @slot footer - Footer text
  */
-export class Card extends LitElement {
-    @property({ reflect: true })
-    public variant: 'default' | 'gallery' | 'quiet' = 'default';
-
+export class Card extends FocusVisiblePolyfillMixin(LitElement) {
     public static get styles(): CSSResultArray {
         return [cardStyles];
     }
+
+    @property({ reflect: true })
+    public variant: 'default' | 'gallery' | 'quiet' = 'default';
+
+    @property({ type: Boolean, reflect: true })
+    public selected = false;
 
     @property()
     public title = '';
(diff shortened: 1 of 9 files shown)


packages/card/src/card.ts
```diff
@@ -104,4 +119,10 @@ export class Card extends LitElement {
                 return this.renderDefault();
         }
     }
+
+    protected firstUpdated(changes: PropertyValues): void {
+        super.firstUpdated(changes);
+        this.setAttribute('role', 'figure');
```
jnurthen, line 125: I don't agree with role=figure for a card. We have no idea what is going to be in the card so figure seems wrong. If we are going to generalize to something it should be role=region or group depending if we want it to be landmark of not (I vote no in most cases so would lean towards group) but we also need to allow it to be overridden which it doesn't seem is allowed here
The card should probably also have aria-labelledby referencing the card's title (if it has a role)

packages/card/src/card.ts
```diff
@@ -104,4 +119,10 @@ export class Card extends LitElement {
...
+        this.tabIndex = 0;
```
jnurthen, line 126: Why are we making the card itself focusable?

packages/card/src/card.ts
Westbrook: This is the role currently outlined in the Spectrum CSS metadata for this pattern, which is applied without an `aria-labelledby`, if that's something you think should be happening, you should definitely submit an issue against: https://github.com/adobe/spectrum-css/blob/master/components/card/metadata/card-asset.yml#L8 as altering it would just be a symptom of a larger issue.

packages/card/src/card.ts
Westbrook: Looking at this again _always_ is probably not the right amount, but my understanding of the card pattern is that it will _generally_ have interactions that become available on focus, and the ability to "activate" it in some way. As there is still further work scheduled against this in #279, we will have to revisit the heuristic (and whether that deserves to be applied internally) by which that is applied at a later date.

packages/card/src/card.ts
jnurthen: I'll do that. We don't add figure in react-spectrum.

packages/card/src/card.ts
jnurthen: Things which aren't widgets should not take focus. If it needs to take focus it should have a widget role not a structural role.

For this kind of interaction we generally recommend to use a layout grid so the card would be a gridcell. Take a look at https://react-spectrum.corp.adobe.com/storybook/?path=/story/gridview--default-layout-gridlayout for more details of that implementation.

packages/card/src/card.ts
jnurthen: filed https://github.com/adobe/spectrum-css/issues/706

PR #646, opened 2020-04-23
fix: constrain overlay

### LINKED ISSUE #645: Dropdown menu doesn't constrain to visible window
### Expected Behaviour

Dropdowns should shrink in order to keep all of the elements on on the screen.

### Actual Behaviour

Dropdowns with lots of options trail off the screen

Screen Shot 2020-04-24 at 10 35 22 AM

#### Steps to Reproduce

1. Create an `sp-dropdown` with lots of items
2. Constrain your browser window to be short enough to cause overflow

https://webcomponents.dev/edit/7DLgYGoRTdZIlJHhSVfg

@Westbrook stole one of your projects as a template. Super cool.

Westbrook:
Add the `maxSize` and `applyMaxSize` modifiers to the Popper instance in the overlay system to that the contents in an overlaid element are contained to the browser window.

fixes #645 

- unit tests pass
- visual regressions pass

Type of Change
- [x] Bug fix.

### DIFF
--- a/packages/overlay/src/popper.ts
+++ b/packages/overlay/src/popper.ts
@@ -25,6 +25,8 @@ import preventOverflow from '@popperjs/core/lib/modifiers/preventOverflow';
 import arrow from '@popperjs/core/lib/modifiers/arrow';
 import offset from '@popperjs/core/lib/modifiers/offset';
 import { computeArrowRotateStyles } from './popper-arrow-rotate';
+import maxSize from 'popper-max-size-modifier';
+import { applyMaxSize } from './apply-max-size.js';
 
 export const createPopper = popperGenerator({
     defaultModifiers: [
@@ -37,4 +39,4 @@ export const createPopper = popperGenerator({
     ],
 });
 
-export { Instance, Placement };
+export { Instance, Placement, maxSize, applyMaxSize };
--- a/packages/popover/src/popover.css
+++ b/packages/popover/src/popover.css
@@ -10,3 +10,52 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 @import './spectrum-popover.css';
+
+:host([placement*='bottom'][open]) {
+    /* .spectrum-Popover--bottom.is-open */
+    max-height: calc(
+        100% -
+            var(
+                --spectrum-dropdown-flyout-menu-offset-y,
+                var(--spectrum-global-dimension-size-75)
+            )
+    );
+}
+:host([placement*='top'][open]) {
+    /* .spectrum-Popover--top.is-open */
+    margin-top: var(
+        --spectrum-dropdown-flyout-menu-offset-y,
+        var(--spectrum-global-dimension-size-75)
+    );
+    max-height: calc(
+        100% -
+            var(
+                --spectrum-dropdown-flyout-menu-offset-y,
+                var(--spectrum-global-dimension-size-75)
+            )
+    );
+}
+:host([placement*='right'][open]) {
+    /* .spectrum-Popover--right.is-open */
+    max-width: calc(
+        100% -
+            var(
+                --spectrum-dropdown-flyout-menu-offset-y,
+                var(--spectrum-global-dimension-size-75)
+            )
+    );
+}
+:host([placement*='left'][open]) {
+    /* .spectrum-Popover--left.is-open */
+    margin-left: var(
+        --spectrum-dropdown-flyout-menu-offset-y,
+        var(--spectrum-global-dimension-size-75)
+    );
+    max-width: calc(
+        100% -
+            var(
+                --spectrum-dropdown-flyout-menu-offset-y,
+                var(--spectrum-global-dimension-size-75)
+            )
+    );
+}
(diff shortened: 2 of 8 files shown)


cuberoot:
It looks good in the constrained case, but the menu expands to the boundaries.

MenuOverflow

I am not sure that I have fully grokked why you need to subtract the `--spectrum-dropdown-flyout-menu-offset-y`.

packages/overlay/src/popper.ts
```diff
@@ -25,6 +25,35 @@ import preventOverflow from '@popperjs/core/lib/modifiers/preventOverflow';
...
+                popper: {
+                    maxHeight: string;
+                    height: string;
+                };
+            };
+        };
+    }) {
+        const { height } = state.modifiersData.maxSize;
+        state.styles.popper.maxHeight = `${height}px`;
+        state.styles.popper.height = `${height}px`;
```
FezVrasta, line 54: I think this is causing the behavior that @cuberoot noted.

```suggestion
```

Westbrook: 100%, yes. It's a bit of a bag of hacks right now to offset some weird CSS intersections that don't make sense to me right now, which is why this is for discussion _only_.

Is there a way to cache the _first_ `const { height: actualHeight } = state.rects.popper;` internal to popper, or would we need to do that ourselves in this modifier?

FezVrasta: In popper we use to put state we want to persist across updates in `state.modifiersData['modifierName#persistent']`.

https://github.com/popperjs/popper-core/blob/bf9d83dd46aa059234e59f26b6a6362999b93937/src/index.js#L201-L204

It's just a convention, actually any modifier state named after something not listed in the modifiers list is ignored, but this may change in the future, it's just an implementation detail.

Westbrook:
@cuberoot, I update the initial size cache from a `WeakMap` that points to the popper element, so it should GC nicely.

This CSS is the fix I was forcing into the `active-overlay` before: https://github.com/adobe/spectrum-web-components/pull/646/files#diff-60a3881116f51ed714be1056050b08e0R14 I think it makes good sense to put this into individual element's CSS, unless you can think of a better way to address.

packages/overlay/src/popper.ts
```diff
@@ -25,6 +25,60 @@ import preventOverflow from '@popperjs/core/lib/modifiers/preventOverflow';
...
+        };
+    }) {
+        const { height: maxHeight } = state.modifiersData.maxSize;
+        if (!appliedSizeDefaults.has(state.elements.popper)) {
+            appliedSizeDefaults.set(
+                state.elements.popper,
+                state.rects.popper.height
+            );
+        }
+        const actualHeight = appliedSizeDefaults.get(state.elements.popper);
```
cuberoot, line 70: 👍

cuberoot:
This looks like a good approach to me. Please carry on!

Westbrook:
@cuberoot PTAL. If you have thoughts on the types in https://github.com/adobe/spectrum-web-components/pull/646/files#diff-278a6296effbd36162c87a51f45cd807 I'd be very interested to hear them... "it works" and it being nice aren't the same thing, right?

Also, if you can think of other overlaid content that I may have missed for the work around a la: https://github.com/adobe/spectrum-web-components/pull/646/files#diff-60a3881116f51ed714be1056050b08e0R14 let me know!

packages/popover/stories/popover.stories.ts
```diff
@@ -43,7 +43,7 @@ export const Dialog = (): TemplateResult => {
     const placement = radios('Placement', placements, placements.bottom);
     return html`
         <div
-            style="color: var(--spectrum-global-color-gray-800); position: relative"
+            style="color: var(--spectrum-global-color-gray-800); position: relative; display: contents"
```
Westbrook, line 46: This failure https://9206-158297197-gh.circle-artifacts.com/0/test/visual/screenshots-current/ci/popover--dialog__darkest__large.png points to the possibility of this breaking usage in the wild... it's something we'd actively remove in Crisp in favor of the new containment code, so it shouldn't be a problem there. Does LRWeb have any usage of `sp-overlay` that might break in this way? The issue being the containing `<div>` here had a height of 0 previously due to the `sp-overlay` being `position: absolute` which meant that the math in ```
    max-height: calc(
        100% -
            var(
                --spectrum-dropdown-flyout-menu-offset-y,
                var(--spectrum-global-dimension-size-75)
            )
    );
```
Didn't know what height to make the popover.

Not sure if this is something we should look deeper into as far as the sizing solution above or not. Thoughts?

File: packages/popover/src/popover.css [hunk 4]
```diff
@@ -10,3 +10,52 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+                var(--spectrum-global-dimension-size-75)
+            )
+    );
+}
+:host([placement*='right'][open]) {
+    /* .spectrum-Popover--right.is-open */
+    max-width: calc(
+        100% -
+            var(
+                --spectrum-dropdown-flyout-menu-offset-y,
```
cuberoot, line 43: Weird that we still use the `y` offset for right and left, but we are doing that seems to be how [spectrum-css is doing it](https://github.com/adobe/spectrum-css/blob/master/components/overlay/index.css#L14)

PR #651, opened 2020-04-27
fix: ensure aria attributes based on state

### LINKED ISSUE #650: Add aria states to checkbox
### Expected Behaviour
indeterminate should result in aria-checked="mixed"
invalid should result in aria-invalid="true"

### Actual Behaviour
Neither of these result in any changes in accessibility APIs

- Run the demo 
https://opensource.adobe.com/spectrum-web-components/components/checkbox
- Inspect the accessibility api and note that checked appears to be true for indeterminate and that invalid causes no changes in the API

#### Platform and Version
All

Westbrook:
## Description
Tracks the `aria-checked` state and `role="switch"` state as appropriate for `CheckboxBase` derived elements.
- make prettier ignore the snapshots so that they aren't updated locally all the time

## Related Issue
fixes #650

## Motivation and Context
The more accessible the better! 

## How Has This Been Tested?
- unit tests

image

## Types of changes

## Checklist:


__snapshots__/Switch.md
```diff
@@ -3,7 +3,13 @@
 #### `loads`
 
 ```html
-<input id="input" tabindex="0" type="checkbox" />
+<input
+    aria-checked="false"
+    id="input"
+    role="switch"
+    tabindex="0"
```
jnurthen, line 10: input type=checkbox shouldn't need tabindex to be set as it is natively focusable

Westbrook: This is a part of our work around for a lack of broad support for `delegatesFocus` in shadow DOM and is expected.

jnurthen: ok - its a no-op so no objection

packages/checkbox/src/checkbox-base.ts
```diff
@@ -52,11 +52,16 @@ export class CheckboxBase extends Focusable {
...
+    protected get ariaCheckedState(): 'true' | 'false' | 'mixed' {
+        return this.checked ? 'true' : 'false';
+    }
+
     protected render(): TemplateResult {
         return html`
             <input
                 id="input"
                 type="checkbox"
+                aria-checked=${this.ariaCheckedState}
```
jnurthen, line 64: unsure why you are doing this here... it would seem to be a no-op again as the checked attribute will suffice in all but the mixed case.

Westbrook: I checked out the React implementation in the v3 release and they were managing it in all three states, so I thought why not. If it's not needed in those context, very much simpler. Thanks for the clarification!

jnurthen: yeah - the react implementation isn't always optimal ;)

PR #655, opened 2020-05-04
Add Spectrum Tags Pattern

Westbrook:
## Description
- add `sp-tags` and `sp-tag` pattern. I rather like the naming convention and wonder if we shouldn't apply the same relationship to the `sp-tab-list` pattern, so `sp-tabs` and `sp-tab` instead?
- correct a display issue in storybook
- allow child selectors in Spectrum CSS input
- clean up the package templating
- generalize the clear button

## Related Issue
fixes #569
fixes #656 

## Motivation and Context
Continue to support a wider range of Spectrum patterns.

## How Has This Been Tested?
- unit tests
- visual regressions

## Screenshots (if appropriate):
image

## Types of changes
- [x] New feature (non-breaking change which adds functionality)

## Checklist:
- [x] I have signed the [Adobe Open Source CLA](http://opensource.adobe.com/cla.html).
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #657, opened 2020-05-06
feat: add dialog, dialog-wrapper and underlay patterns

### LINKED ISSUE #565: Add Spectrum Dialog Pattern
### Expected Behaviour
https://opensource.adobe.com/spectrum-css/components/dialog/ should be available in the library

### Actual Behaviour
missing

#### The future
image

Westbrook:
Blocked by #655 

## Description
- add `<sp-dialog>`
- add `<sp-dialog-wrapper>`
- add `<sp-underlay>`
- demonstrate, document, test, repeat

## Related Issue
fixes #565 

## Motivation and Context
Coverage of Spectrum pattern.

## How Has This Been Tested?
- unit tests
- visual regressions

image
image
image

## Types of changes

## Checklist:


packages/dialog/dialog-wrapper.md
```diff
@@ -0,0 +1,86 @@
+## Description
+
+`sp-dialog` display important information that users need to acknowledge. They appear over the interface and block further interactions.
```
andrewhatran, line 3: I think you may have forgotten to change this to `sp-dialog-wrapper`

packages/dialog/dialog-wrapper.md
```diff
@@ -0,0 +1,86 @@
...
+
+### Installation
+
+[![See it on NPM!](https://img.shields.io/npm/v/@spectrum-web-components/dialog?style=for-the-badge)](https://www.npmjs.com/package/@spectrum-web-components/dialog)
+[![How big is this package in your project?](https://img.shields.io/bundlephobia/minzip/@spectrum-web-components/dialog?style=for-the-badge)](https://bundlephobia.com/result?p=@spectrum-web-components/dialog)
+
+```
+npm install @spectrum-web-components/dialog
```
andrewhatran, line 11: ^ Same as above

packages/dialog/src/dialog-wrapper.ts
```diff
@@ -0,0 +1,184 @@
...
+    public noDivider = false;
+
+    @property({ type: Boolean, reflect: true })
+    public open = false;
+
+    @property({ type: String, reflect: true })
+    public mode: 'fullscreen' | 'fullscreenTakeover' | '' = '';
+
+    @property({ type: String, reflect: true })
+    public size: 'small' | 'medium' | 'large' | 'alert' | '' = '';
```
andrewhatran, line 67: Setting the default size to 'blank' is a but confusing for me. Is that equivalent to 'medium'?

packages/dialog/src/dialog.ts
```diff
@@ -0,0 +1,122 @@
...
+    public noDivider = false;
+
+    @property({ type: Boolean, reflect: true })
+    public open = false;
+
+    @property({ type: String, reflect: true })
+    public mode: 'fullscreen' | 'fullscreenTakeover' | '' = '';
+
+    @property({ type: String, reflect: true })
+    public size: 'small' | 'medium' | 'large' | 'alert' | '' = '';
```
andrewhatran, line 58: Same as above

andrewhatran:
Tested out the storybook and the component looks great! Nice job with the variation coverage and storybook examples.

Just a few comments about the README and size property, but other than that, it looks good to go 👍

packages/dialog/dialog-wrapper.md
Westbrook: Here there is only the `@spectrum-web-components/dialog`, you can choose to use one component or the other, but both are included.

packages/dialog/dialog-wrapper.md
Westbrook: Good catch!

packages/dialog/src/dialog-wrapper.ts
Westbrook: DId I write that? Hrm 🤔 It's the equivalent of it not being there...let me look deeper at what that does, and how it interacts with the styles. In some cases I've used a more complex getter/setter pair to ensure that the attribute was removed in this case, would that be more or less confusing?

Westbrook: As we're not using this internally (beyond presence testing), would it be less confusing for this to be types as follows?
```
public size?: 'small' | 'medium' | 'large' | 'alert';
```

andrewhatran: The above looks more clear to me, thanks! 

So a size of null is equivalent to the dialog not being active, that makes sense to me 👍

Westbrook: `<sp-dialog open size="small|medium|large|alert">` enforces specific widths onto the dialog.

`<sp-dialog open>` will be open and take up as much space as it needs.

`<sp-dialog>` will not be active.

Westbrook: Catch me up on how this feels now that that most recent code is up and passing CI.

andrewhatran: This API looks solid to me 👍

Issue #679, opened 2020-05-15
[RFC] Package/File Organization for Easier Bundling and Leveraging Future APIs


Westbrook:
Currently, all of our element packages are structured to leverage `index.js` (after running `tsc`) as a side effect full export of all elements included in a package. This includes both the class definitions for the various customs elements that are defined and the side effect of defining them against `window.customElements` the global (and currently only) element registry. Much of the reasoning for this comes from Justin Fagnani who explicates the belief that this should be so, here: https://justinfagnani.com/2019/11/01/how-to-publish-web-components-to-npm/#always-self-define-elements. This approach is further established by patterns like `LitElement`s [`@customElement` decorator](https://lit-element.polymer-project.org/api/modules/_lit_element_.html#customelement) that inherently requires a colocated class definition and `define()` call. However, as of late, this concept starts to run afoul of some concepts around usability in the future (and possibly at current), performance tuning of custom elements, particularly in cases when we ship a number of elements and/or sub-elements in the same package (e.g. `@spectrum-web-components/button`), and private API surfacing.

## Future Use Cases: Scoped Custom Element Registries

As discussed in https://github.com/w3c/webcomponents/issues/716 the reality of Scoped Custom Element registries may only be in development as yet. However, having been featured in the recent Web Components Face to Face held by the W3C it is highlight likely that this or a close cousin version of this API gets shipped to browsers. While presenting the current state of the API specification at this meeting Fagnani himself admitting to likely needing to revise his best practices suggestion around element registering as part of the specification getting closer to landing. It is certainly possible to leverage a trivial subclass in order to register a custom element definition in both the global and local registries, as is done in the explain like so:

```
// x-foo.js is an existing custom element module that registers a class
// as 'x-foo' in the global registry.
import {XFoo} from './x-foo.js';

// Create a new registry that inherits from the global registry
const myRegistry = new CustomElementRegistry(window.customElements);

// Define a trivial subclass of XFoo so that we can register it ourselves
class MyFoo extends XFoo {}

// Register it as `my-foo` locally.
myRegistry.define('my-foo', MyFoo);
```

However, a component package that does this by default would then be leaking details into the global scope as part of the initial registration, which means, in the case of the use case this API prepares us for: multiple registrations of elements with the same name, we could be exposing the possibility of a failed registration or in the case of our registration gating, like so:
```
if (!customElements.get('sp-action-menu')) {
    customElements.define('sp-action-menu', ActionMenu);
}
```
[...]

PR #683, opened 2020-05-15
Directory Organization Refactor

Westbrook:
This PR is for comment only as it supports the realities outlined in #679. Please take a look, compare the changes for this small subset of functionality, taking in mind what this would look like across the entire project, and share your thoughts in that issue.

While there are a lot of changes involved, I do think a conversion would need to be all at once and might be best served with a new base-level version for all packages, e.g. `0.11.0` to outline this new reality is true for a package, as I think we're still a little ways away from feeling `1.0.0`-ish.

**Breaking Changes**
-  moves side effects out of entry points for all component packages
   ```
      import '@spectrum-web-components/action-menu';
   ```
   becomes
   ```
     import '@spectrum-web-components/action-menu/sp-action-menu.js';
   ```
   etc.
- adds `@spectrum-web-components/bundle/elements.js` for rapid prototyping with all elements pre-registered
- merges the following packages to better mirror the Spectrum CSS organization:
  - `@spectrum-web-components/tab` + `@spectrum-web-components/tab-list` => `@spectrum-web-components/tabs`
  - `@spectrum-web-components/menu` + `@spectrum-web-components/menu-group` + `@spectrum-web-components/menu-item `=> `@spectrum-web-components/menu`
  - `@spectrum-web-components/radio` + `@spectrum-web-components/radio-group` => `@spectrum-web-components/radio`
  - updates to side effects above will allow you to opt into the components you want in these grouped packages, packages that were previously grouped, and allow us to group by default going forward.
- updates `lit-html` and `lit-element` to `dependencies` from `peerDependencies` to reduce requirements on implementing projects
- removed `if (!customElements.get('sp-*')) {` gating from element registration so that bundling multiple versions of the same element will error more visibly


Westbrook:
The Synk error here has to do with the package removals, it registers all the `package.json` files at import and looks for them in each PR. Assuming we move forward with this, we'll have to [re-import the project](https://support.snyk.io/hc/en-us/articles/360000910537-Manifest-not-found) after this PR and use admin privileges to merge this once it's approved.

karma.conf.js
```diff
@@ -55,7 +55,7 @@ module.exports = (config) => {
                 thresholds: {
                     global: {
                         statements: 97,
-                        branches: 90,
+                        branches: 92,
```
andrewhatran, line 58: 🙏

packages/bundle/elements.ts
```diff
@@ -0,0 +1,59 @@
...
+import '@spectrum-web-components/checkbox/sp-checkbox.js';
+import '@spectrum-web-components/circleloader/sp-circleloader.js';
+import '@spectrum-web-components/coachmark/sp-coachmark.js';
+import '@spectrum-web-components/dialog/sp-dialog.js';
+import '@spectrum-web-components/dialog/sp-dialog-wrapper.js';
+import '@spectrum-web-components/dropdown/sp-dropdown.js';
+import '@spectrum-web-components/dropzone/sp-dropzone.js';
+import '@spectrum-web-components/icon/sp-icon.js';
+import '@spectrum-web-components/icons/sp-icons-large.js';
+import '@spectrum-web-components/icons/sp-icons-medium.js';
```
andrewhatran, line 30: Did we want to include `icons-ui` and `icons-workflow` here?

packages/overlay/src/overlay.ts
```diff
@@ -64,14 +64,14 @@ export class Overlay {
      * @param options.placement side on which to position the overlay
      * @returns an Overlay object which can be used to close the overlay
      */
-    public static open(
+    public static async open(
```
andrewhatran, line 67: This static method shares the same name as the `public async open` method in line 98.  The static method creates and opens an overlay, while the other opens an existing overlay. Should we rename one of them for clarity? 

Probably out of scope for this PR so feel free to ignore

andrewhatran:
Great work! Overall, the separation of the side effect from the component files makes a lot of sense and gives our users much more flexibility around how they consume, register, and extend our components. 

I agree that this is a large enough change to warrant a new base-level version for all packages.

Maybe it would be useful to update the `README` with a section on how to import our components given the new structure.

karma.conf.js
Westbrook: If you like this, you're gonna love: https://github.com/adobe/spectrum-web-components/pull/728/files#diff-a2a3b7b0c9c3b4b93b4aebf4e3ec3cfbR57

packages/bundle/elements.ts
Westbrook: This may play into your suggestion to extend the documentation with this info, but this file is for side effectful imports, which means that were you to add `icons-*` exports here you would once again run into the problem of:
```
import '@spectrum-web-components/bundle/element.ts';
import { StarIcon } from '@spectrum-web-components/bundle/element.ts';
```
Allowing certain linters to error and or auto-delete one of the imports. As such, I think that it's important to continue to import icons from the default entry point:
```
import { StarIcon } from '@spectrum-web-components/bundle';
```
However, it's hard to say 100% as this package is specifically designed to surface _ALL_ of the component available in the larger project for prototyping efforts, without much consideration for perf, etc anyways...

Thoughts?

packages/overlay/src/overlay.ts
Westbrook: This makes sense to me. @cuberoot, I think you brought this to the project to begin with, do you have any thoughts here? If you think this makes sense too, I'll add an issue to track separately.

Something to keep in mind here is that this will be a breaking change...I don't think we have much usage of this API, but it would be good to look at doing this sort of change soon.

Westbrook:
@andrewhatran right now we outline this a la:

### Installation
<a href="https://www.npmjs.com/package/@spectrum-web-components/dropzone" class="spectrum-Link">See it on NPM!</a><a href="https://bundlephobia.com/result?p=@spectrum-web-components/dropzone" class="spectrum-Link">How big is this package in your project?</a>
```
npm install @spectrum-web-components/dropzone

# or

yarn add @spectrum-web-components/dropzone
```

---

Would it make sense to update that to something like the following:

### Usage
<a href="https://www.npmjs.com/package/@spectrum-web-components/dropzone" class="spectrum-Link">See it on NPM!</a><a href="https://bundlephobia.com/result?p=@spectrum-web-components/dropzone" class="spectrum-Link">How big is this package in your project?</a>
```
yarn add @spectrum-web-components/dropzone
```
Import the side effectful registration of `<sp-dropzone>` via:
```
import '@spectrum-web-components/dropzone/sp-dropzone.js';
```
When looking to leverage the `Dropzone` base class as a type or for extension purposes, do so via:
```
import { Dropzone } from '@spectrum-web-components/dropzone';
```

---

Packages like `@spectrum-web-components/button` and `@spectrum-web-components/menu` would of course have slightly more expansive offerings here as they export a number of base classes and feature a number of side effectful offerings.

---

Anything else you think should be included herein before I dive into making this edit across the library.

packages/bundle/elements.ts
andrewhatran: Ah I see, that makes sense. Thanks for the examples! 

Even though `bundle` is designed to surface all components, it makes sense to me to limit the scope of the `bundle/element.ts` to side effect imports only. 

Running into linter errors while prototyping sounds like a pretty bad time haha

andrewhatran:
Your example looks excellent 🙏  Nice and crystal clear

documentation/scripts/buildSearchIndex.js
```diff
@@ -43,7 +43,8 @@ async function main() {
         if (fileName !== 'README.md') {
             componentName = fileName.replace('.md', '');
         }
-        const body = await fs.readFile(path, { encoding: 'utf8' });
+        const content = await fs.readFile(path, { encoding: 'utf8' });
+        const body = content.replace(/```((.|\s)*?)```/g, '');
```
Westbrook, line 47: @andrewhatran I'm not sure whether it's size of the search index, or the specifics of indexing coded content, but the addition of the new documentation was causing trouble in the search interface, this will extract the code snippets from the documentation pages _before_ parsing them for the search index, and causes the interface to work again 🙈. This is probably a sign we're getting close to the edge of the capabilities of `lunr.js`, but a battle for another day...

This also saves about 50KB on the index as well, so working and lighter, win:win!

Westbrook:
Now with import documentation across the project in the components' `README.md` files...

Westbrook:
@andrewhatran cleaned up the Synk situation, caught up to `main`, and added the `README.md` updates that you requested, please take another look.

packages/action-menu/README.md
```diff
@@ -1,18 +1,26 @@
...
 
-yarn add @spectrum-web-components/action-menu
+```
+import '@spectrum-web-components/action-menu/sp-action-menu.js';
+```
+
+When looking to leverage the `AcitionMenu` base class as a type and/or for extension purposes, do so via:
+
+```
+import { AcitionMenu } from '@spectrum-web-components/action-menu';
```
andrewhatran, line 23:
```suggestion
import { ActionMenu } from '@spectrum-web-components/action-menu';
```

andrewhatran:
There's a small typo, but this looks great, nice job!

Thanks for adding the usage notes to the README.

Ready to approve after rebase 👍

PR #687, opened 2020-05-16
chore: improve documentation responsiveness

### LINKED ISSUE #686: Padding in docs is too high on mobile
### Expected Behaviour
`html demo` sections of the docs a reasonable amount of padding on small widths, so the content has enough room.

### Actual Behaviour
The padding takes over at smaller widths, leaving almost no room for content.

### Screenshots

html-demo

adixon-adobe:
Reduce padding at smaller widths for both the `html demo` blocks,
as well as the overall left/right padding for the docs.
## Related Issue

Fixes #686 

## Motivation and Context

The docs should look better on mobile. There are still issues on some pages, but this should give us a big boost in the right direction.

## How Has This Been Tested?

I've reviewed several of the doc pages in the chrome mobile/responsive dev tool. I want some feedback on the breakpoints before I do a complete review.

## Types of changes

## Checklist:

- [ x ] I have updated the documentation accordingly.
- [ x ] I have added tests to cover my changes.
- [ x ] All new and existing tests passed.

### DIFF
--- a/documentation/src/components/code-example.css
+++ b/documentation/src/components/code-example.css
@@ -19,11 +19,19 @@ governing permissions and limitations under the License.
     width: 100%;
 }
 
-#demo {
+.demo-example {
     flex: 1;
-    padding: 3rem 4rem;
+    padding: var(--spectrum-global-dimension-size-400)
+        var(--spectrum-global-dimension-size-500);
     border-radius: 6px 6px 0 0;
     background: var(--spectrum-global-color-gray-100);
+}
+
+@media (max-width: 768px) {
+    .demo-example {
+        padding: var(--spectrum-global-dimension-size-200)
+            var(--spectrum-global-dimension-size-200);
+    }
 }
 
 #markup {
--- a/documentation/src/components/code-example.ts
+++ b/documentation/src/components/code-example.ts
@@ -129,7 +129,7 @@ export class CodeExample extends LitElement {
         return html`
             ${this.showDemo
                 ? html`
-                      <div id="demo">
+                      <div class="demo-example">
                           <slot name="demo">
                               ${renderedCode}
                           </slot>
--- a/documentation/src/components/layout.css
+++ b/documentation/src/components/layout.css
@@ -55,6 +55,12 @@ governing permissions and limitations under the License.
     margin-right: auto;
 }
 
+@media screen and (max-width: 768px) {
+    #body #layout-content #page {
+        padding: 40px 16px 24px 16px;
+    }
+}
+
 header {
     min-height: var(--spectrum-global-dimension-size-600);
     border-bottom: 1px solid var(--spectrum-global-color-gray-200);


documentation/src/components/code-example.ts
```diff
@@ -129,7 +129,7 @@ export class CodeExample extends LitElement {
         return html`
             ${this.showDemo
                 ? html`
-                      <div id="demo">
+                      <div class="demo-example">
```
adixon-adobe, line 132: It seemed weird to have this be an `id` since we often have multiple on the page. I don't think this is causing any issues, but we should obviously do a good pass on the docs before merging.

documentation/src/components/code-example.css
```diff
@@ -19,13 +19,25 @@ governing permissions and limitations under the License.
...
+    .demo-example {
+        padding: 2.25rem 3rem;
+    }
+}
+
+@media (max-width: 568px) {
+    .demo-example {
+        padding: 1.5rem 2rem;
+    }
+}
```
adixon-adobe, line 39: 768 is iPad, so that seems like a good breakpoint (though maybe I should go `767` for the changes?).

568 is just 200 less than that. No real reason, but it seems like a good point to drop the padding down further.

documentation/src/components/layout.css
```diff
@@ -55,6 +55,18 @@ governing permissions and limitations under the License.
...
 
+@media (max-width: 768px) {
+    #body #layout-content #page {
+        padding: 40px 40px 24px 40px;
+    }
+}
+
+@media (max-width: 568px) {
+    #body #layout-content #page {
+        padding: 40px 26px 24px 26px;
```
adixon-adobe, line 66: Just tried to drop the left/right padding to ~75% & 50% of what was there. Without going to odd values.

documentation/src/components/code-example.ts
adixon-adobe: yeah -- gonna take me a while to get used to that. Will this make adding hashtag links harder out of curiousity? I see the `docs-page` element takes over.

documentation/src/components/code-example.css
adixon-adobe: I might as well to switch to those dimension variables (but try to bottom our a little lower?)

I think we'll need something different for the `sp-theme` examples where the page itself controls the sizing since it's so tight. Don't want to tackle that here, outside of getting the right followup issue written. Do you think something like `html demo-no-padding` or `html demo-custom` would be the right name? (I haven't quite figured out how that doc section actually gets mapped to the `code-example` files btw).

documentation/src/components/layout.css
adixon-adobe: yeah -- realized I should compare to what they're doing. I'll switch to the one breakpoint at a 768 and at least be consistent.

adixon-adobe:
Don't have time to wrap this up just yet, but will "soon".

documentation/src/components/code-example.css
```diff
@@ -19,13 +19,21 @@ governing permissions and limitations under the License.
...
+    padding: var(--spectrum-global-dimension-size-400)
+        var(--spectrum-global-dimension-size-500);
     border-radius: 6px 6px 0 0;
     background: var(--spectrum-global-color-gray-100);
 }
 
+@media (max-width: 768px) {
+    .demo-example {
+        padding: var(--spectrum-global-dimension-size-200)
+            var(--spectrum-global-dimension-size-200);
```
adixon-adobe, line 33: Went with 200 for the responsive block.

adixon-adobe:
Everything looks a lot better with this change.

At 360px wide, there are still some noticeable issues with Card (last example), Dialog & Popover. 

There are also pretty minor issues with:
Banner (Corner placement – pink box a few pixels too wide)
Radio Group (a little tight)
Tab list (with icons a little too wide)

documentation/src/components/code-example.css
Westbrook: That's tricky. I've been working on track reimagining of the docs site, but without any specific reason to publish it hadn't prioritized getting it into a PR. If you're going to be doing a lot more docs work, would it be useful for me to work it into PR (even if just into a feature branch) so that we can work from there. Addressing something like this might fall in line with how it prints the documentation into the light DOM, which should give us some more flexibility in customizing this padding.

documentation/src/components/code-example.ts
Westbrook: I'm fairly certain that you can't use these sorts of links to point to content in a shadow root. This is one of the reasons that the `westbrook/11ty` branch moves our content into the light DOM.

documentation/src/components/code-example.css
adixon-adobe: I wouldn't count on me doing a ton of work near term -- I just wanted to get back to fixing some issues since I'm filing a bunch, and docs were the easiest thing to start with.

PR #705, opened 2020-05-23
sp-sidenav-item spacing and docs delivery

### LINKED ISSUE #587: SideNav font and/or color is slightly off
Compare the style application of the SideNav pattern between SWC and Spectrum CSS to figure out where the slight font weight/color difference comes from.

### Spectrum CSS
image

### SWC
image

Westbrook:
## Description
Help our docs match the Spectrum CSS docs, while ensuring a more correct delivery of `sp-sidnav-item` elements.

## Related Issue
fixes #587

## Motivation and Context
Spectrum CSS adherence.

### Before
image

### After
image

## How Has This Been Tested?
- visual regressions updated

## Types of changes

## Checklist:

### DIFF
--- a/documentation/src/main.css
+++ b/documentation/src/main.css
@@ -16,4 +16,5 @@ body {
     height: 100%;
     overflow: hidden;
     margin: 0;
+    -webkit-font-smoothing: antialiased;
 }
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,7 +42,7 @@ commands:
             - restore_cache:
                   name: Restore Golden Images Cache
                   keys:
-                      - v1-golden-images-<< parameters.regression_color >>-<< parameters.regression_scale >>-314192b37a820f729dd96d24a7ef63c85d5372a5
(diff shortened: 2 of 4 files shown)


documentation/src/main.css
```diff
@@ -16,4 +16,5 @@ body {
     height: 100%;
     overflow: hidden;
     margin: 0;
+    -webkit-font-smoothing: antialiased;
```
davidyxu, line 19: should we also be applying the firefox version (`-moz-osx-font-smoothing`) of this? and/or is this something our postcss setup will handle?

Westbrook: Great question! I didn't even know that property existed... At current, I've stolen this single property from the Spectrum CSS site. `-moz-osx-font-smoothing: grayscale;` seems to be the equivalent Firefox entry, but doesn't seem to apply a similar effect, which may be why they might have avoided this there.

I'll double check with the Spectrum CSS team about this.

davidyxu: 👍 we can move forward with this PR i guess and follow up with a second if there's a need for the firefox property.

Issue #199, opened 2019-09-13
process-spectrum doesn't manage relating elements to other instances of themselves


Westbrook:
### Expected Behaviour
`.spectrum-Menu .spectrum-Menu` -> `:host sp-menu`
`.spectrum-Button + .spectrum-Button` -> `:host(:nth-of-type(n+2))`
`.spectrum-Menu > .spectrum-Menu-sectionHeading` -> `:host() > .heading`
etc.

### Actual Behaviour
`.spectrum-Menu .spectrum-Menu` -> `:host()`
```
node_modules/@adobe/spectrum-css/dist/components/button/index-vars.css
305:18	⚠  :host rules cannot use complex combinator (+): .spectrum-Button + .spectrum-Button [postcss-process-spectrum]
```
```
node_modules/@adobe/spectrum-css/dist/components/menu/index-vars.css
14:16	⚠  :host rules cannot use complex combinator (>): .spectrum-Menu > .spectrum-Menu-sectionHeading [postcss-process-spectrum]
```

### Reproduce Scenario (including but not limited to)
`npm run process-spectrum`

Westbrook:
There are no longer skipped selectors in the processing script.

PR #706, opened 2020-05-26
feat(button-group): add ButtonGroup pattern

### LINKED ISSUE #563: Add Spectrum Button Group Pattern
### Expected Behaviour
https://opensource.adobe.com/spectrum-css/components/buttongroup/ should be available in the library

### Actual Behaviour
missing

#### The future
image

Westbrook:
## Description
- adds the `sp-button-group` pattern: https://opensource.adobe.com/spectrum-css/components/buttongroup/
- adds a `complexSelectors` entry to the Spectrum Config, as follows:
   ```
   complexSelectors: [
       {
           // new complex selector
           replacement: '::slotted(sp-button:not(:first-of-type))',
           // The spectrum-css selector who's rules we wish to apply
           selector: '.spectrum-Button+.spectrum-Button',
       },
   ]
   ```
  The above is a little brute force as it is regexs directly into the individual selector (as it reaches across multiple nodes in the AST), rather than relying on the PostCSS parsing. It also relies on the `:not(:first-of-type)` selector which isn't a direct translation of the `+` combination. I do think it works here, and might be able to work in the other contexts of #199 if we chose to apply it so.
- manually extends `sp-action-button` to support managed style overrides

### Notes:
The `3.0.0-beta` series of this element simplifies itself greatly in regards to the complex selectors (centralizes on a single set, rather than 3 sets) and ends the reaching into `sp-action-button`, which will allow for the removal of the manual extension there. It does require our update to support `:dir()` selectors and the update of `@spectrum-css/button` to `3.0.0-beta` as well, soI decided to run with the included version for the time being.

## Related Issue
fixes #563 
refs #199

## Motivation and Context
Spectrum CSS coverage

## How Has This Been Tested?
- unit tests
- visual regressions

image
image

## Types of changes

## Checklist:


packages/button-group/src/spectrum-config.js
```diff
@@ -0,0 +1,71 @@
...
+                {
+                    contents: 'sp-rule[vertical]',
+                    // The spectrum-css selector who's rules we wish to apply
+                    selector: '.spectrum-Rule--vertical',
+                },
+            ],
+            complexSelectors: [
+                {
+                    replacement: '::slotted(sp-button:not(:first-of-type))',
+                    // The spectrum-css selector who's rules we wish to apply
```
andrewhatran, line 53: Pretty minor but can we remove these comments?

Westbrook: 🙈 copied form the docs site...removed!

PR #708, opened 2020-05-26
refactor(dropdown): extend updateComplete timing for overlay oven/close

Westbrook:
## Description
- allows `sp-dropdown` to lazy load the `Overlay` and `sp-popover` code
- wraps that laziness into the `updateComplete` lifecycle promise, so that timing-based interactions with the element can be managed with the appropriate fidelity

## Motivation and Context
- users should be able to prevent unused code from being downloaded by their applications
- when doing so, those same users should be able to time interactions based on the availability of that code

## How Has This Been Tested?
- updated unit tests
- manually build/test in `westbrook/11ty`

## Types of changes
- [x] refactor (update interface to render timing)

## Checklist:

### DIFF
--- a/packages/dropdown/src/dropdown.ts
+++ b/packages/dropdown/src/dropdown.ts
@@ -40,8 +40,7 @@ import {
     MenuItem,
     MenuItemQueryRoleEventDetail,
 } from '@spectrum-web-components/menu-item';
-import '@spectrum-web-components/popover';
-import { Overlay, Placement } from '@spectrum-web-components/overlay';
+import { Placement } from '@spectrum-web-components/overlay';
 
 /**
  * @slot label - The placeholder content for the dropdown
@@ -227,15 +226,17 @@ export class DropdownBase extends Focusable {
         delete this.placeholder;
     }
 
-    private openMenu(): void {
+    private async openMenu(): Promise<void> {
         /* istanbul ignore if */
         if (
             !this.popover ||
             !this.button ||
             !this.optionsMenu ||
             this.optionsMenu.children.length === 0
-        )
+        ) {
+            this.menuStateResolver();
             return;
+        }
 
         this.placeholder = document.createComment(
             'placeholder for optionsMenu'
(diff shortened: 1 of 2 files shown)


packages/dropdown/src/dropdown.ts
```diff
@@ -256,16 +257,39 @@ export class DropdownBase extends Focusable {
         if (menuWidth) {
             this.popover.style.setProperty('width', menuWidth);
         }
```
andrewhatran, line 259: I understand that we're lazy loading the `Overlay` class so that we can prevent unused code from being downloaded.

However, it seems like `Dropdown` is heavily dependent on overlay since every dropdown contains a popover. Wouldn't `Overlay` be imported every time dropdown is used anyway?

Westbrook: 100%, `Overlay` is a hard requirement of using the functionality that is delivered by an `sp-dropdown` element. However, lazy loading isn't only about preventing unused code; that's more directly a responsibility of tree shaking and side effect management (see https://github.com/adobe/spectrum-web-components/issues/679 that looks to enhance out approach to that).

In this case, `Overlay` isn't a requirement of displaying the dropdown element when closed, and if you don't need to interact with the dropdown, this would allow the page never to download that JS. As an example, checkout the documentation site: https://opensource.adobe.com/spectrum-web-components/ unless you specifically interact with the theme customization UI, or are looking to review the dropdown documentation, there is no need to load the `Overlay` scripts.

In the case that you do need the `Overlay` scripts, it is highly unlikely that you'd interact with the dropdown immediately (e.g. load the page with the dropdown open by default, an,. if needed, that can be managed into a single bundle externally). This change looks to ensure that we give download/parse/run priority to JS that is required as the page is rendered by moving this `Overlay` into the dynamic dependency graph as opposed to the static dependency graph.

andrewhatran: I see, that checks out. Giving priority to JS when the dependency is required rather than at construction makes a lot of sense. 

It makes even more sense for disabled dropdowns.

PR #711, opened 2020-05-27
docs: process all CSS Custom Properties for inclusion in the docs site

### LINKED ISSUE #71: Css variables should be displayed in the documentation of all components

Westbrook:
## Description
Add `cssProperies` values dynamically to the docs site, storybook, and the component packages.

Use `postCSS` to extract all of the CSS Custom Properties listed in the various component packages' `dist/vars.css` files to JSON. Use that JSON to extend the contents of the `custom-element.json` associated with the specific output.

Turns on the Storybook Docs addon so that the `cssProperties` can be seen, also adds the Storybook Themeing to apply the Adobe logo, `Spectrum Web Components` title, and appropriate URL to the storybook UI.

### Initial reaction...
This is a LOT of Custom Properties, and I'm not sure that this form of delivery is going to be particularly useful long term. However, it being the best (read the lowest investment) option right now, I think it'll be good to have this public for users to at least reference. Future work will likely be needed around tracking properties set from Spectrum themes back to their source values (possibly across various themes) and to ensure that packages that export more than one element definitions have those properties targeted to specific elements documented as such.

## Related Issue
fixes #71

## Motivation and Context
The more we document, the easier our components will be for users to work with.

## How Has This Been Tested?
Manual review of the docs and output files.

image
image

## Types of changes
- [x] docs

## Checklist:


documentation/src/components/component.ts
```diff
@@ -27,7 +27,7 @@ enum TabValue {
 }
 
 interface JsDocTagParsed {
```
andrewhatran, line 29: I'm curious of why we're making the `tag` property optional? 
I also can't find where the property is being used 🤔 What does `tag` represent?

andrewhatran:
I'm really happy with the change! This is going to make the customization of our spectrum components way easier.

I honestly have no problems with the quantity. Since the variables are an API for custom styling, I think that displaying all of them it in the docs is useful, especially if we want to extend swc into CC Lite web-components down the line

documentation/src/components/component.ts
```diff
@@ -27,7 +27,7 @@ enum TabValue {
...
-    tag: string;
+    tag?: string;
```
andrewhatran, line 30: I'm curious of why we're making the `tag` property optional? 
I also can't find where the property is being used 🤔 What does `tag` represent?

Westbrook: I'm not sure what this was before...I've removed it.

Also, when we update to a slightly newer version of the analyzer output, it looks like we'll be able to start using the internal typings, but we're not quite there yet.

PR #714, opened 2020-06-01
fix(theme): support lazily loading theme fragments

Westbrook:
## Description
Through experiments around reducing the initial load of the documentation site, it became clear that `sp-theme` wasn't fully delivering the flexibility of load currently outlined in its [documentation](https://opensource.adobe.com/spectrum-web-components/components/theme). The ability to load only the style fragments you need, when you need them, is particularly important for the `scale` fragments which top out around 28kb parsed, but the `color` fragments also use up about 6kb of parsed JS, so having more than you need of either is wasteful.

- better structure `Theme` for use with lazily loaded theme fragments
- use the `scale` and `color` attributes before falling back to the fragment "defaults" when establishing the current themes
- always require the presence of a fragment for each of `color, scale and core` before rendering styles
- tighten up TS
- expand tests

## Motivation and Context
Fast sites do less. This makes sure our themeing can be delivered with the smallest amount of JS, and opens the possibility of delivering first themes in CSS and waiting for user interaction to load further customization.

## How Has This Been Tested?
- update old unit tests
- add new unit tests

## Types of changes

## Checklist:

### DIFF
--- a/packages/theme/src/theme.ts
+++ b/packages/theme/src/theme.ts
@@ -10,8 +10,9 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import coreStyles from './theme.css';
 import { CSSResult, supportsAdoptingStyleSheets } from 'lit-element';
+
+import coreStyles from './theme.css';
 
 declare global {
     interface Window {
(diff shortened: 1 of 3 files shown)


Westbrook:
@msdewey this now features updates based on our collaborative review this morning, please take a look!

PR #716, opened 2020-06-01
fix(overlay): ensure undefined data is not passed into theme

Westbrook:
## Description 
Clean up the possibility of passing `undefined` into `color` or `scale` values in a `sp-theme` element.

## Motivation and Context
Lazily defined root themes might not always supply full themes to the overlay system.

## How Has This Been Tested?
Manually while developing on `westbrook/11ty`.

## Types of changes

## Checklist:

### DIFF
--- a/packages/overlay/src/active-overlay.ts
+++ b/packages/overlay/src/active-overlay.ts
@@ -25,6 +25,7 @@ import {
     property,
     PropertyValues,
 } from 'lit-element';
+import { ifDefined } from 'lit-html/directives/if-defined';
 import styles from './active-overlay.css.js';
 
 export interface PositionResult {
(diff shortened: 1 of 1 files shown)


packages/overlay/src/active-overlay.ts
```diff
@@ -401,10 +402,9 @@ export class ActiveOverlay extends LitElement {
     }
 
     public renderTheme(content: TemplateResult): TemplateResult {
-        const color = this.color as Color;
-        const scale = this.scale as Scale;
+        const { color, scale } = this;
         return html`
-            <sp-theme .color=${color} .scale=${scale}>
```
adixon-adobe, line 407: what did the `.` characters you removed here do?

Westbrook: Using `.color=${color}` applies the value `color` to `sp-theme` as a property setter, e.g. `document.querySelector('sp-theme').color = color`. If we did this without the `.` or `color=${color}` would apply it as an attribute, e.g. `document.querySelector('sp-theme').setAttribute('color', color)`. The solution used here `color=${ifDefined(color)}` essentially does `color && document.querySelector('sp-theme').setAttribute('color', color)` so that the value is only passed when `typeof color !== 'undefined'`.

PR #717, opened 2020-06-04
RFC: Use react-aria and react-stately

benjamind:
## Description
The React Spectrum team are imminently releasing their v3 version of the react spectrum components. These are based on a new set of react hooks, `react-stately` and `react-aria`. These hooks manage component state, and aria attributes and interaction event handling.

This RFC shows a possible approach to using these hooks in SWC.

## Demo
To try this out, just run `yarn storybook`, I've modified the configs to only show the `sp-switch` stories.

## General Approach
This prototype uses [haunted](https://github.com/matthewp/haunted) and the `react-stately` and `react-aria` hooks to implement the `sp-switch` component uses these hooks.

Quite a lot of the changes in this PR are merely to get things working in storybook and alias the various modules needed to make this work, these changes such as the es-dev-server config, the haunted wrapper, the yarn package aliasing etc could possibly be done in other ways if we chose to take this approach.

The basic outline however is that we would create a new base class for Haunted compatible lit-elements, this would encapsulate the creation and update of the Haunted State object. Then our render methods are changed to use the react hooks, and our components would all implement the interfaces of the react component types.

## Pros
* We would share a common implementation for handling all aria behaviors and event bindings with the react team.
* Similarly state management would be shared with the react team.
* We would define a mostly common interface with the properties on components in the react-spectrum project.

## Cons
* Some approaches that are common in web-component APIs (slotting, event interfaces etc) may not map well to react approaches so we may end up with divergent API anyway.
* Additional bundle size (unconfirmed how much this is, but there's clearly some additional dependencies here to make this all work).
* Some work required to more easily handle spreading of properties, filtering of properties for specific components etc.
* Additional mental energy required to 'unpack' the results of the aria hooks and define where they are spread onto our DOM.

## Summary
Some of the advantages here are clear, not having to re-do the work of defining aria behaviors and state management that the React team is already doing is a nice benefit. An aligned API would also be beneficial for other Adobe projects such as UXP.

The overhead here though on the SWC project itself is not trivial, there'd be a lot of work to re-implement our components using this approach. I suspect there are also many unknown dragons in this approach too, especially with regard to our DOM structure, and slotted elements etc. While I think some of the pain points in this PR could be automated away with a little library code and some base classes, I'm not convinced I have yet uncovered all of what that work would be.

Having said that however, @devongovett has expressed an interest in adjusting these hooks to better suit integration with more frameworks. Some of these are I believe already planned (the workaround I had to do for the native event binding for instance is needed to support Preact I believe).

I would be interested to hear what thoughts the team has on this approach.


packages/switch/src/switch.ts
```diff
@@ -15,31 +15,133 @@ import {
...
         if (window.hasOwnProperty('ShadyDOM')) {
             // Override some styles if we are using the web component polyfill
-            return [...super.styles, switchStyles, legacyStyles];
+            return [switchStyles, legacyStyles];
         }
-        return [...super.styles, switchStyles];
+        return [switchStyles];
     }
 
     protected render(): TemplateResult {
```
benjamind, line 120: The meat of the change is in the render method here.

packages/switch/src/switch.ts
```diff
@@ -15,31 +15,133 @@ import {
...
+        // we also want to apply some of these attributes to our host (the aria ones)
+        for (const [key, value] of Object.entries(inputProps)) {
+            if (key.startsWith('aria-')) {
+                if (value) {
+                    this.setAttribute(key, value as any);
+                } else {
+                    this.removeAttribute(key);
+                }
+            }
+        }
```
benjamind, line 137: I think this part isn't actually working right, not seeing the aria attributes on the host correctly. Will need to dig into the exact logic needed here more if we decide to take this forward.

devongovett: One question I have is whether aria attributes should go on the host element or the `<input>` inside it? Seems like it should probably be the input, but then attributes like `aria-labelledby` wouldn't work because the id space is limited to the same shadow. Is there a way to make that work with shadow DOM?

devongovett:
This is great! We'd love to collaborate more with the SWC team. 🤩 

As @benjamind mentioned, at the moment the hooks are somewhat dependent on React itself, especially the synthetic event system. But we're looking to support Preact in the near future, which doesn't implement synthetic events, so we'll need to do some work to make us less dependent on that. That should benefit things like Spectrum Web Components as well, since the events you'll get from lit-element will be pure DOM events just like Preact.

We are also open to abstracting things internally as well to avoid hackery like aliasing react in the build system. For example, we could have some registry that built-in React hooks (e.g. `useState`, `useEffect`) could be registered on to allow registering haunted/preact hooks there instead.

It would awesome to collaborate on the low level internals and at least avoid duplicating effort there. Then we can have implementations on top with multiple rendering/styling layers. I've already heard interest from outside adobe in this as well, e.g. a Vue/Angular implementation. The ARIA stuff is surprisingly hard to implement, especially with touch/mouse/keyboard support. We've put a lot of effort in there, so would be great to see SWC get the benefits of that too. Let us know what we can do to help!

Westbrook:
@benjamind this is super interesting, thanks for getting the conversation started here!

With your cons in mind, can you think of which of our elements would likely be the hardest conversion/sell on this approach? I've not gotten a chance to see how much of the new libraries have been released as yet but I'd love to stretch this RFC a bit more in search of any breaking points. Sampling the integration at the boundaries of some of the more complicated realities like slots would be very nice to see, for sure. Maybe `<sp-dropdown>` would be a good test case? It certainly taxed us a good bit when we brought the most recent version to the library, even without the sort of abstractions contemplated herein.

To continue the conversation around bundle size (something that I'd be particularly concerned about from the outside of a change like this), I made a `test` release of this branch so we could begin to see the costs we'd be taking on. Take a look at https://bundlephobia.com/result?p=@spectrum-web-components/switch@0.2.16-test.31 to see the ~3x jump in file size, even for a smaller component like a switch.
image

It's hard to say right now how much that would amortize across numerous elements, how wildly the cost might vary based on the functionality exposed, or exactly what capabilities the added KB get us. Does the stately or aria libraries have documentation on the things they actually cover for us to begin to compare before/after with?

Looking forward to spending some more time with the code in this PR and your thoughts on the questions above to see what's the right next step for the library!

.yarnrc
```diff
@@ -0,0 +1,5 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
Westbrook, line 1: Is this some new yarn thing? Doesn't seem related to the conversion, but I've also not see it in any project before.

packages/switch/package.json
```diff
@@ -45,7 +45,15 @@
...
     "dependencies": {
+        "@open-wc/lit-helpers": "^0.3.10",
+        "@react-aria/switch": "^3.0.0-rc.2",
+        "@react-aria/toggle": "^3.0.0-rc.2",
+        "@react-stately/toggle": "^3.0.0-rc.2",
+        "@react-types/checkbox": "^3.0.0-rc.2",
         "@spectrum-web-components/checkbox": "^0.3.1",
+        "es-dev-commonjs-transformer": "^0.1.1",
+        "haunted": "^4.7.0",
+        "react": "npm:haunted",
```
Westbrook, line 56: Is there a way to get this without the cross dependency? It block up less nuanced services like Stackblitz: 
image which wouldn't be the biggest loss, but is certainly something support I'd love for us to be able to keep if possible.

.storybook/es-dev-server.config.js
```diff
@@ -0,0 +1,21 @@
+const cjsTransformer = require('es-dev-commonjs-transformer');
+
+module.exports = {
+    responseTransformers: [
+        cjsTransformer(
```
Westbrook, line 5: Is this because the new libraries from the React team at CJS? Or is there something else that hasn't updated to ES modules here?

packages/switch/src/switch.ts
```diff
@@ -15,31 +15,133 @@ import {
...
     }
 
     protected render(): TemplateResult {
+        // use the state from react-stately, this manages isSelected state
+        const state = useToggleState(this);
+        // use the hook from react-aria, this gives us event bindings and aria
+        // attributes to use in our DOM rendering
+        const { inputProps } = useSwitch(this, state, this);
+        // we have to hack the spread naming to be compatible with lit-html
+        fixSpreadNaming(inputProps);
```
Westbrook, line 127: I was definitely in conversation about whether it expand support for spreading events or booleans only, which would make this a filter rather than a conversion step 🙈 We could certainly expend the `spread` directive to support that on our own. Not sure how much we'd benefit from not having to `delete` those properties on every render...

packages/switch/src/switch.ts
Westbrook: Applying these to the `<input>` is mostly the right answer. Currently, `CheckboxBase` is applying the `aria-` coverage we do have in this way and `Switch` is applying the `role` attribute similarly.

There are some cases where we've been cavalier or undecided about an `input`'s relationship to its `label`, but if we went this way it would be a good time to solidify this. Insights on how you saw this applying to slotted `input`s or elements with `elementInternals` would probably be a useful inclusion towards clarifying next steps here.

devongovett:
> Does the stately or aria libraries have documentation on the things they actually cover for us to begin to compare before/after with?

Yeah we try to cover the high level features on each docs page. e.g. https://react-spectrum.adobe.com/react-aria/useSwitch.html

For switch, I guess the main thing that's contributing to the size is actually our `usePress` hook, which handles cross device touch/mouse/keyboard events. In this case, it's really only being used to normalize focus across browsers on mousedown/touchstart (some browsers don't do this by default, e.g. Safari). If we wanted, we could probably make a smaller version of just this part since we don't actually need mouse/touch events on the switch, but as you said, the question is whether it's worth it if `usePress` is already included in a number of other components. 

Bundlephobia is also reporting `react` is somehow in there too - maybe it didn't understand the alias? We will be taking a look at trying to make our bundle sizes smaller soon, so hopefully there are some improvements. Since the hooks don't include React or React DOM and are just functions, ideally we'd end up being relatively similar to if you implemented things separately.

> Maybe `<sp-dropdown>` would be a good test case?

I think that would be a good one to try out. Certainly one of the most complex components we have. See https://react-spectrum.adobe.com/react-aria/useSelect.html.

.yarnrc
benjamind: I found I had to pin yarn to this version using the rc file to get it to work, some bug in yarn 1.21 was preventing me installing any new dependencies: https://github.com/yarnpkg/yarn/issues/7807

packages/switch/package.json
benjamind: Yeah not sure, I guess we can probably do a module resolve as part of a build process, but definitely a problem for buildless environments.

.storybook/es-dev-server.config.js
benjamind: There's a deep dependency on a library [classnames](https://github.com/JedWatson/classnames) that the react packages depend upon that appears to need the CJS treatment.

packages/switch/src/switch.ts
benjamind: Agreed I figure we could build a spread helper thats suitable for this specific operation easily enough.

benjamind:
> We are also open to abstracting things internally as well to avoid hackery like aliasing react in the build system. For example, we could have some registry that built-in React hooks (e.g. `useState`, `useEffect`) could be registered on to allow registering haunted/preact hooks there instead.

That I think would be great, the aliasing of react gets really tricky in buildless systems.

packages/switch/src/switch.ts
benjamind: Ah yes, sorry my misunderstanding then, probably don't need this loop at all to apply anything to the host then.

I think we should definitely explore how to apply to slotted inputs or elementInternals in this manner too. Probably some additional helpers around processing slotted content on slot change perhaps?

devongovett:
How about a hackathon after shutdown?

PR #719, opened 2020-06-07
test: add accessibility testing where applicable

Westbrook:
## Description
Expand the testing suite to rely on `await expect(el).to.be.accessible();` more liberally, and correct prior use of this command that omitted the `await` signifier.
- swap a lot of `.to.equalSnapshot();` tests for this approach as the DOM internal to our components is private API and doesn't need to be guaranteed across functional changes whereas the accessibility of our components does.
- clean up unused snapshots 
- ensure elements meet even the low level of accessibility that can be confirmed by automation

## Motivation and Context
- when moving to this testing strategy the following tests fail https://app.circleci.com/pipelines/github/adobe/spectrum-web-components/2916/workflows/9472cdaa-39a0-496c-b159-a877c9950e41/jobs/11857/parallel-runs/0/steps/0-108 some of them are laziness in test preparation, but others required small changes to the actual component implementation.

## How Has This Been Tested?
- unit tests in the CI

Before:
image
After:
image

## Types of changes

## Checklist:


packages/dialog/src/dialog-wrapper.ts
```diff
@@ -138,7 +138,11 @@ export class DialogWrapper extends LitElement {
             >
                 ${this.hero
                     ? html`
-                          <img src="${this.hero}" slot="hero" />
+                          <img
+                              src="${this.hero}"
+                              slot="hero"
+                              aria-hidden="true"
```
Westbrook, line 144: This is cheating! Currently, Spectrum CSS uses a background image for this, which is very close to the same level of cheating, so I wasn't sure what to do about applying an `alt` tag here. Should we use `aria-hidden`? Should we reapply the `headline` text? Should we surface an additional attribute? Should we move to a background image? I'm not sure the right path for both functionality and accessibility here...

packages/dialog/src/dialog.ts
```diff
@@ -102,7 +102,7 @@ export class Dialog extends LitElement {
                       `
                     : html``}
             </div>
-            <div class="content">
+            <div class="content" tabindex="0">
```
Westbrook, line 105: @jnurthen would love some additional insight here. This is based on the feedback in https://dequeuniversity.com/rules/axe/3.4/scrollable-region-focusable?application=axeAPI. However, this element won't always be "scrollable" even though it has `overflow: auto`, due to content size. Do we need to dynamically cover for that? Or, is it OK to make it focusable even when the content doesn't require scroll?

Westbrook:
FYI: CI fails because of an expected change in the screenshot testing. Want to make sure we work through any alterations of the actual code in this PR before I bump the golden image cache.

packages/dialog/src/dialog-wrapper.ts
andrewhatran: As discussed offline, the second alternate that leverages the `hero-label` attribute makes the most sense to me. 

Both the `headline` and `hero-label` alternatives align with how we're handling the alt tag of shadowDOM images in other components, such as `Avatar`. I prefer the `hero-label` implementation, since the headline of the dialogWrapper doesn't always align with the contents of the hero image

Westbrook: I've updated `sp-dialog-wrapper` to include `hero-label` attribute which applies to the `alt`  attribute to the `` when available. When unavailable, the `aria-hidden="true"` practice is maintained.

packages/dialog/src/dialog.ts
Westbrook: @andrewhatran I've thought about this some more, and while it seems sub-optimal to have or not have the `tabindex` 100% of the time like this, it seems slightly less so to have it 100% of the time _AND_ any updates we make to when it is available in the future would essentially be within the private API scope of the element, so there is no need to block on them now. I've created https://github.com/adobe/spectrum-web-components/issues/723 to track the additional functionality and design application that will likely be needed. Without directly addressing this question now, I think we should be OK moving forward on a more complete review of this PR when you have some time! Thanks.

andrewhatran: This totally makes sense to me. If we're planning on updating this component in the future to make the tabindex responsive to the scrollability of the element, then I feel comfortable with moving forward with this workaround 👍 

I'll continue with a more thorough/complete review as soon as I have time!

packages/actionbar/stories/actionbar.stories.ts
```diff
@@ -10,53 +10,35 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
-        <sp-actionbar
-            ?open=${boolean('Open', true, 'Element')}
-            variant=${ifDefined(
-                select('Variant', variants, variants[0], 'Element') || undefined
-            )}
-        >
+        <sp-actionbar open>
             <sp-checkbox indeterminate>228 Selected</sp-checkbox>
-            <div class="spectrum-ButtonGroup">
+            <sp-button-group>
```
andrewhatran, line 30: 👍

packages/dialog/src/dialog-wrapper.ts
```diff
@@ -54,6 +54,9 @@ export class DialogWrapper extends LitElement {
     @property()
     public hero = '';
 
+    @property({ attribute: 'hero-label' })
```
andrewhatran, line 57: 👍

PR #728, opened 2020-06-22
Manage tab order and focus placement with Overlay

### LINKED ISSUE #709: Allow overlays to trap the tab order.
As per https://codepen.io/samthor/live/b459e77f8dc7fe8a32925866bf4e0628 we could leverage shadow DOM to implement tab trapping of overlaid content with a relatively low lift. Questions we'd want to have answered:

- Is trapping off of `body` as is done in the example to invasive?
- When should we trap? For instance, the `click-content` of an `overlay-trigger` likely wants for trapping all the time, but the `sp-dropdown` element does not.
- Should trapping be an explicit part of the `Overlay` API or implicit to an overlay type?

related #710

### LINKED ISSUE #710: Throw/Manage focus when an overlay is open
When an overlay opens, the browser focus should be thrown into it. This means either the overlay itself or the first focusable item in the overlay should receive focus immediately by default. When this occurs the focus should flow through the overlay, and then be thrown back to the triggering element when focusable content has been fully stepped through. For instance, the `sp-dropdown` element should not need to throw focus into its options list when it opens, and once the list is exhausted (a single tab in this case for the floating tab index) the focus should return to the UI from which it came.

related #709

### LINKED ISSUE #723: Dialog content tab stop for scrollable content
### Expected Behaviour
The `tabindex=0` application to the `#content` element in `sp-dialog` is managed to only be present when scrolling is needed. This will reduce unnecessary tab stops.

There should also be design affordance to clarify that the element if focused when scrolling is present.

### Actual Behaviour
Currently, `tabindex=0` is never applied, but #719 will be adding it 100% of the time to support accessible interactions with this content.

### LINKED ISSUE #727: Upon closing sp-menu using 'Esc' key, focus is not restored to triggering button
### Expected Behaviour
When sp-menu is closed, keyboard focus should go back to triggering button. 
### Actual Behaviour
Keyboard focus is restored properly to triggering button when user selects a menu item but it doesn't happen when user chooses to close the menu using 'Esc' key without selecting an item.

#### Platform and Version
Tested on https://opensource.adobe.com/spectrum-web-components/components/action-menu

Westbrook:
## Description
- adds inline and modal style overlays as described in https://github.com/adobe/spectrum-web-components/issues/710#issuecomment-643331874
- adds focus throwing via `receivedFocus`
- adds tab trapping via composed shadow DOM
- leverages new API in `sp-dropdown` and extension
  - updates timing to not include arbitrary `requestAnimationFrame` calls
- adds stories and tests to the functionality
  - sets overlay story element to "submit" on `enter` press
- updates `sp-radio-group` to appropriately manage keyboard interactions

Demo available at: https://westbrook-overlay-focus--spectrum-web-components.netlify.app/storybook/index.html?path=/story/action-menu--icon-only

## Related Issue
fixes #709 
fixes #710 
fixes #723 
fixes #727

## Motivation and Context
- increase the keyboard accessibility of our components and UI powered by our APIs
- code cleanliness

## How Has This Been Tested?
- added unit testing
- updated screenshot testing

## Types of changes

## Checklist:

### DIFF
--- a/packages/overlay/src/overlay-stack.ts
+++ b/packages/overlay/src/overlay-stack.ts
@@ -32,10 +32,67 @@ export class OverlayStack {
 
     public constructor() {
         this.addEventListeners();
+        this.initTabTrapping();
+    }
+
+    private tabTrapper!: HTMLElement;
+    private overlayHolder!: HTMLElement;
+
+    private initTabTrapping(): void {
+        this.document.body.attachShadow({ mode: 'open' });
+        /* istanbul ignore if */
+        if (!this.document.body.shadowRoot) {
+            return;
+        }
+        const root = this.document.body.shadowRoot;
+        root.innerHTML = `
+            <div id="actual"><slot></slot></div>
+            <style>
+            #actual {
+                position: relative;
+                z-index: 0;
+            }
+            #holder {
+                display: flex;
+                align-items: center;
+                justify-content: center;
+                flex-flow: column;
+                height: 100%;
+                width: 100%;
+                top: 0;
+                left: 0;
+                position: fixed;
+            }
+            #holder[hidden] {
+                display: none !important;
+            }
+            #actual[tabindex="-1"] {
+                pointer-events: none;  /* just in case? */
+            }
+            </style>
+            <div id="holder" hidden><slot name="open"></slot></div>
+        `;
+        this.tabTrapper = root.querySelector('#actual') as HTMLElement;
+        this.overlayHolder = root.querySelector('#holder') as HTMLElement;
+        this.tabTrapper.attachShadow({ mode: 'open' });
+        /* istanbul ignore else */
+        if (this.tabTrapper.shadowRoot) {
+            this.tabTrapper.shadowRoot.innerHTML = '<slot></slot>';
+        }
+    }
+
+    private startTabTrapping(): void {
+        this.tabTrapper.tabIndex = -1;
+        this.overlayHolder.hidden = false;
+    }
+
+    private stopTabTrapping(): void {
+        this.tabTrapper.removeAttribute('tabindex');
+        this.overlayHolder.hidden = true;
     }
 
     private get document(): Document {
-        return this.root.ownerDocument || document;
+        return this.root.ownerDocument /* istanbul ignore next */ || document;
     }
 
     private get topOverlay(): ActiveOverlay | undefined {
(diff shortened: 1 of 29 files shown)


packages/overlay/src/overlay-stack.ts
```diff
@@ -32,10 +32,67 @@ export class OverlayStack {
...
     public constructor() {
         this.addEventListeners();
+        this.initTabTrapping();
+    }
+
+    private tabTrapper!: HTMLElement;
+    private overlayHolder!: HTMLElement;
+
+    private initTabTrapping(): void {
+        this.document.body.attachShadow({ mode: 'open' });
```
andrewhatran, line 42: I was thinking about the body shadowRoot collision issue that we discussed offline.

What are your thoughts on cancelling the tabTrapping initiation if a shadowRoot already exists in the body?

Westbrook: Not a bad idea for safety... document that users with other tools leveraging a `shadowRoot` on the `<body>` element just don't get the trapping functionality. Certainly better than just going and breaking apps that have this for some reason.

packages/overlay/src/overlay-stack.ts
```diff
@@ -142,7 +230,7 @@ export class OverlayStack {
 
     private async hideAndCloseOverlay(
         overlay?: ActiveOverlay,
-        animated = true
+        animated?: boolean
```
andrewhatran, line 233: Is there a distinction between `animated: false` and `animated: null` for overlays?

Westbrook: Technically, yes. The `overlay.hide(animated);` that receives this argument also does `animated = true` in its method signature, which means that `undefined` in this method actually means `true` as it did before when given a default, only this way we can get test coverage on both lines...

packages/overlay/src/overlay-stack.ts
```diff
@@ -32,10 +32,81 @@ export class OverlayStack {
...
+    }
+
+    private canTabTrap = true;
+    private tabTrapper!: HTMLElement;
+    private overlayHolder!: HTMLElement;
+
+    private initTabTrapping(): void {
+        /* istanbul ignore if */
+        if (this.document.body.shadowRoot) {
+            this.canTabTrap = false;
```
andrewhatran, line 45: 🎉

PR #747, opened 2020-07-07
fix(tabs): ensure only one active tab stop in the tabs

### LINKED ISSUE #740: Focus jumps from current element to selected element before leaving the component
This issue is similar to what's happening with Side Nav. https://github.com/adobe/spectrum-web-components/issues/733

When a user is arrowing through tabs, on pressing tab, focus should leave tabs component. It happens okay if the current focus is on the selected element. But otherwise, it lands on selected tab and then goes out of the component.

Westbrook:
blocked by #683

## Description
Ensure that an `<sp-tabs>` element only has one active tab stop while navigating through its children via the keyboard.

This also removes the (seemingly) completely unused `value` attribute from `<sp-tabs>`.

## Related Issue
fixes #740 

## Motivation and Context
Better keyboard accessibility. 

## How Has This Been Tested?
- new unit tests!

## Types of changes

## Checklist:

### DIFF
--- a/packages/tabs/src/Tabs.ts
+++ b/packages/tabs/src/Tabs.ts
@@ -54,9 +54,6 @@ export class Tabs extends Focusable {
     @property()
     public selectionIndicatorStyle = '';
 
-    @property({ type: String, reflect: true })
-    public value = '';
-
     @property({ reflect: true })
     public get selected(): string {
         return this._selected;
@@ -116,6 +113,10 @@ export class Tabs extends Focusable {
         this.setAttribute('role', 'tablist');
         this.addEventListener('mousedown', this.manageFocusinType);
         this.addEventListener('focusin', this.startListeningToKeyboard);
+        const selectedChild = this.querySelector('[selected]') as Tab;
+        if (selectedChild) {
+            this.selectTarget(selectedChild);
+        }
     }
 
     protected updated(changes: PropertyValues): void {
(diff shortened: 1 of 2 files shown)


packages/tabs/src/Tabs.ts
```diff
@@ -116,6 +113,10 @@ export class Tabs extends Focusable {
         this.setAttribute('role', 'tablist');
         this.addEventListener('mousedown', this.manageFocusinType);
         this.addEventListener('focusin', this.startListeningToKeyboard);
+        const selectedChild = this.querySelector('[selected]') as Tab;
```
msdewey, line 116: wondering if maybe selectedChild could be a `@query` property?

```
@query('[selected]')
private selectedChild?: Tab;
```

Westbrook: _If_ this were possible in this context, I'd very much want to use it as well! However, the `@query` decorator only queries the element's render root, which would be `this.shadowRoot` in the context of our elements, and would not query for the elements in question here. There is an issue at the `LitElement` level to expand on the `@queryAsssignedNodes` decorator in a way that should support this technique: https://github.com/Polymer/lit-element/issues/1016

Westbrook: Thinking this through a little there is some helper code that we _might_ use here:
```
        this.addressSelectedChild((child: Tab): void  => { this.selectTarget(child); });
    }

    private addressSelectedChild(addressWith: (child: Tab) => void): void {
        const selectedChild = this.querySelector('[selected]') as Tab;
        if (selectedChild) {
            addressWith(selectedChild);
        }
    }
```
I'm a little worried that this becomes harder to read, however. Let me know what you think!

Issue #739, opened 2020-07-07
Standard slider example needs to be labelled


yalagand:
URL: https://opensource.adobe.com/spectrum-web-components/components/slider

Please provide a label for Standard variant of slider.
There's a section 'With Label' that has sliders with visible label. Is that supposed to mean we're allowing users to have sliders with no visible label when you say 'Standard Slider'? If yes, how do I label it?

Westbrook:
We attempt to stay as close to the Spectrum CSS documentation as possible, which outlines this distinction. As it was a contribution, it's possible that the documentation here wasn't fully vetted, or that it might get another pass as part of later edits. We should definitely confirm usage there as well being they're the ones outlining the "standard" use as not having a label: https://unpkg.com/browse/@spectrum-css/slider@3.0.0-beta.2/metadata/slider.yml

I would suppose even without the internal label, you _could_ use `aria-label` on the `sp-slider` element and still instill value on the accessibility tree. That being optimal or not is certainly a different question.

Westbrook:
Sliders without labels removed.

Issue #669, opened 2020-05-13
sp-icon size attribute should be deprecated


adixon-adobe:
### Expected Behaviour
The sp-icon size attribute as it exists should be deprecated or removed as it doesn't map to an actual Spectrum design concept. It also defaults to the small size which isn't the default size that should be used in controls.

Instead we should make the following changes (likely with support from SpectrumCSS):

* Default the size to the equivalent of "s" in the current implementation (18x18 or 22x22 depending on the scale).
* Use an overridable variable for the default size, which will better prepare us for t-shirt sized components. E.g. we might have a large size for action button which needs the icon to be larger as well. This should "just work" if you use sp-icon. 

### Actual Behaviour
We support a size attribute that doesn't make much sense to support at this point. There are 2 upcoming spectrum initiatives that should supersede what's here: t-shirt sizing for controls & icon sizing for things with a wider range of sizes like avatars.

Westbrook:
The sizes that are currently supported come from https://unpkg.com/browse/@spectrum-css/icon@2.1.1/dist/index.css This package doesn't seem to have a beta release for a 3.0 version, which tends to mean that Spectrum CSS isn't currently testing breaking changes for their new version. It seems from the conversations I've seen you have with the team that Spectrum is planning changes to the spec, but it's unclear when/if those changes will make it into `@spectrum-css/icon` as most of the Spectrum changes seem to be in relation to a icons inside of patterns rather than icons deployed individually (so much so that neither the Spectrum public nor contribution sites seem to reference icons not otherwise contained in a larger patter).

If the timeline there is unknown or non-short, is there a specific use case that we can focus on supporting better at the SWC level or a specific way to could message a transition period to keep the RIO of using `<sp-icon>` today high while reducing any refactor costs that might be required to get to the new version of Spectrum and Spectrum CSS when it solidifies at a later date?

Westbrook:
Can you help me clarify the expected behavior you've listed here? The code seems to currently default to `size="m"` https://github.com/adobe/spectrum-web-components/blob/main/packages/icon/src/icon.ts#L36-L37

Based on the latest version of the Spectrum CSS packages, this works out to be `24px X 24px` and `30px X 30px` in the medium and large scales, respectively.
- https://unpkg.com/browse/@spectrum-css/icon@2.1.1/dist/index.css
- https://unpkg.com/browse/@spectrum-css/icon@2.1.1/dist/index-lg.css

[...]

Issue #543, opened 2020-03-16
Update "Textfield" to most recent Spectrum CSS beta


Westbrook:
### Expected Behaviour
Textfield meets the most recent Spectrum CSS specification

### Actual Behaviour
Beta releases of this pattern have breaking changes.

block #477

## Proposal
Re-implement this pattern with `@spectrum-css/textfield@3.0.0-beta`.

### Considerations:
- The beta releases rely on `[dir="rtl"]` selectors, the CSS for which needs to be processed differently than previously. This also requires some affordance in the JS to acquire `dir` from an ancestor element. The easiest approach here would be to leverage `document.dir` but it's technically possible to have multiple `dir` values across a document. Should we resolve this value to the "closest" instance or to the "root" instance? Either seems acceptable if we document it and allow it to be locally set.
- ~order of selector writing can cause processing errors where `[dir=ltr] .spectrum-Textfield--quiet.spectrum-Textfield.is-invalid .spectrum-Textfield-validationIcon` becomes `:host([quiet][invalid]) .spectrum-Textfield[dir='ltr'] .icon` due to expectations in our processing~
- uses `.is-focused` and `.is-keyboard` focused as opposed to `:focus` and `:focus-visible` so we'll need to manage those values (possibly not that big of a deal being these selectors ALWAYS duplicate an `input:focus` selector that matches it for appropriately to our current approach)

Issue #513, opened 2020-02-27
sp-textfield cannot be sized easily


benjamind:
### Expected Behaviour

This should behave like any other input control, it should obey display style rules and layout rules applied to the component. 

### Actual Behaviour

Size is not adjustable without using css custom properties. This is due to width of the input control being set to a specific value from the custom variable, the `:host` styling has `display: inline-flex` and requires this display style to behave properly, and the input cannot resize.

Current workaround is to set `--spectrum-alias-single-line-width: 300px;` variable on the textfield instance to size its internal input box.

### Reproduce Scenario (including but not limited to)

Try and set `width: 500px` on any `sp-textfield` instance.

#### Steps to Reproduce

* Create an `sp-textfield` element
* Style it with `width: 300px`
* Observe textfield does not change size

Westbrook:
This looks to be addressed in the early versions of https://www.npmjs.com/package/@spectrum-css/textfield/v/3.0.0-beta.2 🎉 

It's not completely clear when Spectrum CSS anticipates having these packages "stable" but that should also give us some time and opportunity to discuss breaking changed on our side. I'm gonna compare notes on the list of packages with breaks that I've received and opened some issues for us to discuss any changes we'd like to see on our side.

PR #754, opened 2020-07-10
[rfc|feature branch] Spectrum CSS 3.0.0 releases and '[dir="rtl|ltr"]'

Westbrook:
## Description
Spectrum CSS is getting close to a v3.0.0 release of all of its components and this branch should be able to serve as a base for testing out individual package upgrades.
- add a "base" package that vends `SpectrumMixin` and `SpectrumElement`, very open to alternate naming conventions if you have them.
  - adds `public shadowRoot!: ShadowRoot;` to the element as we currently use shadow DOM in all of our elements and it's a pain to test for it in TS _all_ the time
  - add a `connectedCallback()` that queries document direction and then applies it to the element tag so that elements extending from his can have an `[dir="rtl|ltr"]` selector to leverage in their styles
- extends the Spectrum CSS processing to include `[dir="rtl|ltr"]` selectors in the output by appending them to the host selector (by default they are included as "context" selectors the director corollary of which in shadow DOM are not available cross-browser)
- exemplifies `sp-actionbar` as a very simple upgrade (just the version bump)
- exemplifies `sp-checkbox` as a more complex upgrade (leverages mixins and requires DOM updates, etc.)

I'm gonna work with the Spectrum CSS team and catch up on how they're documenting the lack of support for `dir="auto"` in this update, as we'll need to include matching instructions in our docs so that users don't fall into any traps there. I'll also confirm whether the team is targeting any other specifically breaking changes (e.g. t-shirt sizes) before moving these releases to stable.

## Related Issue
related to #477 
related to #513 
related to #543 
related to #669 
related to #739

## Motivation and Context
Need to keep up with out source material

## How Has This Been Tested?
- snapshots for the effected elements have been updated
- hopefully visual regressions "just work" 🤞 

## Types of changes

## Checklist:

### DIFF
--- a/packages/base/README.md
+++ b/packages/base/README.md
@@ -1,0 +1,29 @@
+## Description
+
+The `SpectrumElement` base class as created by mixing `SpectrumMixin` onto `LitElement` adopts `dir` values from the `document` at connection time with a fallback to `lrt`. In a TypeScript context, it also enforces the presence of `this.shadowRoot` on extending components.
+
+### Usage
+
+[![See it on NPM!](https://img.shields.io/npm/v/@spectrum-web-components/base?style=for-the-badge)](https://www.npmjs.com/package/@spectrum-web-components/base)
+[![How big is this package in your project?](https://img.shields.io/bundlephobia/minzip/@spectrum-web-components/base?style=for-the-badge)](https://bundlephobia.com/result?p=@spectrum-web-components/base)
+
+```
+yarn add @spectrum-web-components/base
+```
+
+When looking to leverage the `SpectrumElement` base class as a type and/or for extension purposes, do so via:
+
+```
+import { SpectrumElement } from '@spectrum-web-components/base';
+
+export MyElement extends SpectrumElement {}
+
+```
+
+Similarly the `SpectrumMixin` class factory mixin is available via:
+
+```
+import { SpectrumMixin } from '@spectrum-web-components/base';
+
+export MyElement extends SpectrumMixin(HTMLElement) {}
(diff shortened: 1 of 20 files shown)


packages/base/README.md
```diff
@@ -0,0 +1,20 @@
+## Description
```
Westbrook, line 1: Describe me.

packages/base/package.json
```diff
@@ -0,0 +1,43 @@
...
+        "spectrum css",
+        "web components",
+        "lit-element",
+        "lit-html"
+    ],
+    "version": "0.0.1",
+    "description": "",
+    "main": "src/index.js",
+    "module": "src/index.js",
+    "type": "module",
```
Westbrook, line 25: include `exports`

packages/base/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "files": [
+        "custom-elements.json",
+        "*.d.ts",
+        "*.js",
+        "*.js.map",
+        "/src/"
+    ],
```
Westbrook, line 32: include `"sideEffects": false`

packages/checkbox/src/spectrum-config.js
```diff
@@ -17,7 +17,7 @@ module.exports = {
             name: 'checkbox',
             host: {
                 selector: '.spectrum-Checkbox',
-                shadowSelector: '#root',
+                // shadowSelector: '#root',
```
Westbrook, line 20: Actually remove this.

Issue #477, opened 2020-02-06
Add Spectrum CSS Number Field (Stepper) Pattern


Westbrook:
### Expected Behaviour
https://opensource.adobe.com/spectrum-css/stepper.html should be available in the library.

### Actual Behaviour
Missing

### The future
![image](https://user-images.githubusercontent.com/1156657/104777704-21ba8380-574a-11eb-829b-f013189315bf.png)

Westbrook:
This is a "contribution" pattern for Spectrum CSS (not included in Spectrum). Hopefully, the support in https://www.npmjs.com/package/@spectrum-css/stepper is still reliable to form this pattern. If not, be sure to touch base with the Spectrum team/your own design to help formulate any missing styles.

Westbrook:
This pattern is based off of the code in `@spectrum-css/textfield@3.0.0-beta.X` which has breaking changes from our current implementation, work here should be held until that work is completed.

#513 
#543

Westbrook:
When looking at the beta release of this pattern, there are two things that jump out immediately as presenting new difficulty to our library's processing of the Spectrum CSS packages:

1. A selector like `.spectrum-Stepper-stepDown` expects to apply styles directly on the `<button>` within the shadow DOM of an `<sp-field-button>` element. We can reach out to Spectrum CSS about why they aren't leveraging CSS Custom Properties here, but the idea of having `:host()` element wrapping the `<button>` in this case is pretty Custom Element specific and something we're not likely to get changed. With that true, the process of converting their output CSS into our component CSS will become much more complicated.
2. A selector like `.spectrum-Stepper-input` intends to target the `<input>` element inside of an `sp-textfield` element. We can reach out to Spectrum CSS about why they aren't leveraging CSS Custom Properties here, but in that this is a fairly BEM-like practice, I'm not sure we'll be able to get that changed at their level. This points to similar complexity of conversion as the above, or to the idea that seems to be growing larger on the horizon that our input elements might require slotted `input/textarea` elements for us to properly leverage them within a form, ensure that we can accessibly relate them to `<label>` element, or style them across the various contexts that are required.

benjamind:
Interesting!

So are you suggesting that we add slot elements for each of our input elements with a default content that provides an input element, but so that we can in cases like stepper, or where we want form integration provide an external element where we want it?

Westbrook:
[...]

PR #761, opened 2020-07-13
Move `sp-textfiield` and `sp-search` to Spectrum CSS ^3.0.0

### LINKED ISSUE #513: sp-textfield cannot be sized easily

### LINKED ISSUE #543: Update "Textfield" to most recent Spectrum CSS beta

### LINKED ISSUE #750: Textfield with error state has incorrect focus indicator
### Expected Behaviour
``` 
<sp-textfield
                label="Name"
                placeholder="Enter your name"
                invalid
            ></sp-textfield>
```
should have a blue border when it gets keyboard focus

The correct version can be seen in https://spectrum.adobe.com/static/XD/Items/Text-field/Text-field-v5.2.0.xd

### Actual Behaviour
It gets a thicker red border instead

Run https://opensource.adobe.com/spectrum-web-components/components/textfield and focus the invalid field using the keyboard

#### Steps to Reproduce
Run https://opensource.adobe.com/spectrum-web-components/components/textfield and focus the invalid field using the keyboard

#### Platform and Version
Reproduces using Safari and Chrome on MacOS

``` 
<sp-textfield
                label="Name"
                placeholder="Enter your name"
                invalid
            ></sp-textfield>
```

Westbrook:
## Description 
Directly (or as close to directly as possible) port the `textfield` and `search` patterns to `^3.0.0`.

Separately, we should review whether the `sp-textfield` pattern needs to be updated to some form of a light dom `<input/>` element in order to appropriately pair with `label` elements and extend to patterns like the `stepper` #477. 

## Related Issue
fixes #513 
fixes #543 
fixes #750

## Motivation and Context 
Keeping up with Spectrum CSS, making customization easier.

## How Has This Been Tested?
- continues to pass unit tests
- visual regressions have been slightly updated

image
image
image

## Types of changes
- [x] dependency management

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,7 +42,7 @@ commands:
             - restore_cache:
                   name: Restore Golden Images Cache
                   keys:
-                      - v1-golden-images-<< parameters.regression_color >>-<< parameters.regression_scale >>-b5462e0cb61beacbd0b9dc751bd0284687baa4d4
+                      - v1-golden-images-<< parameters.regression_color >>-<< parameters.regression_scale >>-c26ca5ae887c2c3c4dce6fff24335b878b31ca45
                       - v1-golden-images-main-<< parameters.regression_color >>-<< parameters.regression_scale >>-
             - run: yarn test:visual:ci --color=<< parameters.regression_color >> --scale=<< parameters.regression_scale >>
             - run:
--- a/__snapshots__/Search.md
+++ b/__snapshots__/Search.md
@@ -4,17 +4,16 @@
 
 ```html
 <form id="form">
+  <sp-icon
+    class="icon icon-workflow magnifier"
+    size="s"
+  >
+  </sp-icon>
   <input
     aria-label="Search"
     id="input"
     placeholder="Search"
   >
-  <sp-icon
-    class="icon magnifier"
-    id="icon"
-    size="s"
-  >
-  </sp-icon>
   <sp-clear-button
     data-js-focus-visible=""
     dir="ltr"
--- a/packages/search/package.json
+++ b/packages/search/package.json
@@ -47,7 +47,7 @@
     "author": "",
     "license": "Apache-2.0",
     "devDependencies": {
-        "@spectrum-css/search": "^2.0.6"
+        "@spectrum-css/search": "^3.0.0-beta.4"
     },
     "dependencies": {
         "@spectrum-web-components/button": "^0.8.3",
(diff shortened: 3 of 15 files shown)


andrewhatran:
The Spectrum CSS changes lgtm 👍  Ready for approval once the circle ci visual tests are addressed.

The `<input/>` element light dom refactor makes a lot of sense. Agreed that there's a lot of value in making our components more compatible with `<form>` and `<label>` elements. 

Out of curiosity, how would we leverage our input components inside of a `<form>` if we decide against moving our `<input/>` elements into the light DOM?

Westbrook:
If we can avoid having the `<input/>` in the light DOM, we would need to leverage one or both of the techniques outlined herein in order to leverage the `sp-textfield` and associated element in ` <form>`: https://web.dev/more-capable-form-controls/ I've named a possible extension of the `formdata` interface in honor of the best film in Ashton Kutcher's filmography: https://glitch.com/edit/#!/dude-wheres-my-form

Westbrook:
@andrewhatran This should be good and caught up. PTAL.

PR #780, opened 2020-07-20
feat(asset): add the asset pattern

Westbrook:
## Description
Implements the `<sp-asset>` pattern.
- leverages Spectrum CSS `^3.0.0`
- doesn't leverage `[dir="rtl|ltr"]`, so no need for `@spectrum-web-components/base`

## Related Issue
refs #279 

## Motivation and Context
Pattern availability

## How Has This Been Tested?
- Accessibility tests
- Screenshot tests

image
image

## Types of changes

## Checklist:

### DIFF
--- a/packages/asset/src/Asset.ts
+++ b/packages/asset/src/Asset.ts
@@ -1,0 +1,70 @@
+/*
+Copyright 2020 Adobe. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License. You may obtain a copy
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under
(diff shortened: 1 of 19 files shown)


packages/asset/src/Asset.ts
```diff
@@ -0,0 +1,70 @@
...
+    LitElement,
+    CSSResultArray,
+    TemplateResult,
+    property,
+} from 'lit-element';
+
+import styles from './asset.css.js';
+
+const file = (): TemplateResult => html`
+    <svg viewBox="0 0 128 128" class="file">
```
Westbrook, line 24: Does this include all of the appropriate `aria-*` tags?

adixon-adobe: I'm guessing we'll need clients to pass in a label for these null states, similar to what we do for `sp-icon`.

I'm assuming the alt tag for child  tags will work.

PR #785, opened 2020-07-24
Add Split Button, update Button to Spectrum CSS v3.0, support direct and distant sibling selector conversion

Westbrook:
## Description
- Add the `<sp-split-button>` pattern
- Update `<sp-buton>` et al to Spectrum CSS v3.0
- Extend `process-spectrum` to convert more direct sibling (`.something .class + .class`) selectors
- Extend `process-specrrum` to convert distant sibling (`.class ~ .class`) selectors
- correct minor style applications in ActionMenu/Card/Dropdown

## Related Issue
fixes #568 
refs #199 

## Motivation and Context
Support more patters.
Support more CSS.

## How Has This Been Tested?
- unit tests
- visual regressions

## Screenshots (if appropriate):
image

## Types of changes
- [x] New feature (non-breaking change which adds functionality)

## Checklist:
- [x] I have signed the [Adobe Open Source CLA](http://opensource.adobe.com/cla.html).
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #790, opened 2020-07-24
Merge Target for Packages using the latest Spectrum CSS v3.0.0 dependency

### LINKED ISSUE #268: Create `<sp-quick-actions>`
### Expected Behaviour
`<sp-quick-actions>` should be available in the library as outlined in Spectrum CSS: https://opensource.adobe.com/spectrum-css/components/quickaction/

image

### Actual Behaviour
The component is missing.

### LINKED ISSUE #279: `sp-card` doesn't fully support Spectrum CSS delivery

### LINKED ISSUE #513: sp-textfield cannot be sized easily

### LINKED ISSUE #543: Update "Textfield" to most recent Spectrum CSS beta

### LINKED ISSUE #561: Add Spectrum Accordion Pattern
### Expected Behaviour
https://opensource.adobe.com/spectrum-css/components/accordion/ should be available in the library

### Actual Behaviour
missing

#### The future
image

### LINKED ISSUE #568: Add Spectrum Split Button Pattern
### Expected Behaviour
https://opensource.adobe.com/spectrum-css/components/splitbutton/ should be available in the library

### Actual Behaviour
missing

#### The future
image

### LINKED ISSUE #662: sp-action-button doesn't fire an event when selected changes.
### Expected Behaviour
If you use the `<sp-action-button toggles>`, a change event is triggered to let you know the selected state has changed.

### Actual Behaviour
No change event is triggered. You can listen to the `click` event for the control, but that doesn't seem ideal.

### LINKED ISSUE #750: Textfield with error state has incorrect focus indicator

### LINKED ISSUE #753: Toast close button does not have a sufficient focus style
### Expected Behaviour
The focus style should be discernable to people with low vision

### Actual Behaviour
The focus style is too subtle
Screen Shot showing the keyboard focus style on the toast close button

#### Steps to Reproduce
* https://opensource.adobe.com/spectrum-web-components/components/toast
* tab through the page until you reach the close button

#### Platform and Version
Chrome 83
Safari 14
FF 78

### LINKED ISSUE #793: Review `has...` testing for slot generation.
All elements extending from `ButtonBase` as well as a few others (like the upcoming version of `sp-tag`) rely on a pattern that looks like:
```
    private get hasIcon(): boolean {
        return !!this.querySelector('[slot="icon"]');
    }
```
To decide whether to add `<slot></slot>` content into an element's shadow DOM. This works great for the first render. However, if the icons or avatar or what have you is removed over the lifecycle of the element, unless otherwise render triggering data is updated, the template will be displayed out of sync.

A more reliable pattern would be to leverage a `MutationObserver` or some other form of `slotchange` event to track the presence of such an element and trigger a render on the pertinent change.

### LINKED ISSUE #845: Specified version of lit-element is not accurate
### Expected Behaviour

Packages should depend on a version of lit-element that matches the feature set used

### Actual Behaviour

There are packages that [depend on lit-element v2.2.1](https://github.com/adobe/spectrum-web-components/blob/main/packages/radio/package.json#L58), but use `queryAssignedNodes` which was only [introduced in version 2.3](https://github.com/Polymer/lit-element/blob/master/CHANGELOG.md)

#### Steps to Reproduce

I hit this while upgrading a project that uses Parcel 2 to use yarn 2

Westbrook:
Includes:
#754 
#761
#783 
#785 
#794
#810 
#817 
#824 

Issues:
fixes #268
fixes #279
fixes #561
refs #199 only 4 left?
```
node_modules/@spectrum-css/tabs/dist/index-vars.css
44:1	⚠  :slotted() rules must be the last in the selector [postcss-process-spectrum]
48:1	⚠  :slotted() rules must be the last in the selector [postcss-process-spectrum]
120:1	⚠  :slotted() rules must be the last in the selector [postcss-process-spectrum]
222:1	⚠  :slotted() rules must be the last in the selector [postcss-process-spectrum]
```
fixes #513 
fixes #543
fixes #568
fixes #662
fixes #750
fixes #753  
fixes #793
fixes #845


Westbrook:
Liz from the Material Web Component's team shared this great research demo for expanding baseline support for `[dir]` in a base class: https://jsbin.com/qomimoxoyu/edit?html,output

PR #794, opened 2020-07-27
Spectrum CSS Updates for more straight forward packages.

Westbrook:
## Description
Get some of the lower hanging fruit components updated to Spectrum CSS v3.0.0
- table
- avatar
- banner
- bar-loader
- button-group
- circle-loader
- coachmark
- dialog
- dropdown
- dropzone
- illustrated-message
- menu
- popover
- radio
- rule
- sidenav
- slider
- status-light
- toggle
- tabs
- tags
- toast
- tooltip
There's also some added support for Spectrum that you'll see in the card package thanks to updated CSS processing, but the full work there is being tracked in a branch specifically addressing updates to the card element.

## Motivation and Context
Keeping up with Spectrum CSS

## How Has This Been Tested?
- updated Visual regressions

## Types of changes

## Checklist:

### DIFF
--- a/packages/bar-loader/src/BarLoader.ts
+++ b/packages/bar-loader/src/BarLoader.ts
@@ -12,19 +12,19 @@ governing permissions and limitations under the License.
 
 import {
     html,
-    LitElement,
+    SpectrumElement,
     CSSResultArray,
     TemplateResult,
     PropertyValues,
     property,
-} from 'lit-element';
+} from '@spectrum-web-components/base';
 
 import styles from './bar-loader.css.js';
 
 /**
  * @element sp-bar-loader
  */
-export class BarLoader extends LitElement {
+export class BarLoader extends SpectrumElement {
     public static get styles(): CSSResultArray {
         return [styles];
     }
--- a/packages/slider/src/Slider.ts
+++ b/packages/slider/src/Slider.ts
@@ -17,7 +17,7 @@ import {
     TemplateResult,
     query,
     PropertyValues,
-} from 'lit-element';
+} from '@spectrum-web-components/base';
 
 import spectrumSliderStyles from './spectrum-slider.css.js';
 import sliderStyles from './slider.css.js';
@@ -442,7 +442,7 @@ export class Slider extends Focusable {
         const percent = (offset - minOffset) / size;
         const value = this.min + (this.max - this.min) * percent;
 
-        return value;
+        return this.isDefaultDir ? value : 100 - value;
     }
 
     private dispatchInputEvent(): void {
@@ -485,7 +485,7 @@ export class Slider extends Focusable {
     private get trackRightStyle(): string {
         const width = `width: ${(1 - this.trackProgress) * 100}%;`;
         const halfHandleWidth = `var(--spectrum-slider-handle-width, var(--spectrum-global-dimension-size-200)) / 2`;
-        const offset = `left: calc(${
+        const offset = `${this.isDefaultDir ? 'left' : 'right'}: calc(${
             this.trackProgress * 100
         }% + ${halfHandleWidth})`;
 
@@ -493,6 +493,8 @@ export class Slider extends Focusable {
     }
 
     private get handleStyle(): string {
-        return `left: ${this.trackProgress * 100}%`;
+        return `${this.isDefaultDir ? 'left' : 'right'}: ${
+            this.trackProgress * 100
+        }%`;
     }
 }
--- a/packages/split-button/stories/split-button.stories.ts
+++ b/packages/split-button/stories/split-button.stories.ts
@@ -119,7 +119,11 @@ moreCta.story = {
 export const moreCtaOpen = (options = {}): TemplateResult => {
     return html`
         <div>
+<<<<<<< HEAD
             <sp-split-button type="more" open>
+=======
+            <sp-split-button open>
+>>>>>>> feat(split-button): add split-button pattern
                 ${menu(options)}
             </sp-split-button>
         </div>
--- a/packages/split-button/stories/split-button.stories.ts
+++ b/packages/split-button/stories/split-button.stories.ts
@@ -119,11 +119,7 @@ moreCta.story = {
 export const moreCtaOpen = (options = {}): TemplateResult => {
     return html`
         <div>
-<<<<<<< HEAD
             <sp-split-button type="more" open>
-=======
-            <sp-split-button open>
->>>>>>> feat(split-button): add split-button pattern
                 ${menu(options)}
             </sp-split-button>
         </div>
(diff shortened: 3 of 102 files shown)


packages/tags/src/Tag.ts
```diff
@@ -77,9 +85,23 @@ export class Tag extends LitElement {
     }
 
     protected render(): TemplateResult {
+        const slots = [];
```
andrewhatran, line 88: I was having a bit a trouble getting the typechecking to pass for this line when running `yarn storybook`

Do you think we should set the type of `const slots` to `TemplateResult[]`?

Westbrook: Not sure why I didn't run into this on the first pass, but I hit it in https://github.com/adobe/spectrum-web-components/pull/785/files#diff-f3e3a8af2e1fc7625253b6bae2053264R88 where the extended typing cleans it up. Could go either way on copying it over or prioritizing the Split Button PR to clear this up.

packages/tabs/src/tab.css
```diff
@@ -54,7 +54,7 @@ governing permissions and limitations under the License.
     height: auto;
 }
 
-:host([vertical]) #itemLabel {
+:host([dir][vertical]) slot[name='icon'] + #itemLabel {
```
andrewhatran, line 57: I'm curious about why we added the `dir` attribute selector here? I understand that v3 elements require the `dir` attribute in order to be styled properly, but it doesnt seem like the direction influences the styling in this case

Westbrook: This is extending the updated root selector in `spectrum-tab.css`, allowing us the maintain the same specificity relationship between the generated output and our additional features.
```
:host([dir='ltr']) slot[name='icon'] + #itemLabel
```

packages/bar-loader/src/BarLoader.ts
```diff
@@ -12,19 +12,19 @@ governing permissions and limitations under the License.
...
-} from 'lit-element';
+} from '@spectrum-web-components/base';
 
 import styles from './bar-loader.css.js';
 
 /**
  * @element sp-bar-loader
  */
-export class BarLoader extends LitElement {
+export class BarLoader extends SpectrumElement {
```
andrewhatran, line 27: Should the progress bar be mirrored for the `rtl` version of the determinate bar loader?

I checked out the `rtl` version of the bar loader in the [Spectrum CSS docs](https://opensource.adobe.com/spectrum-css/barloader.html) and noticed that progress bar runs right to left

packages/slider/src/Slider.ts
```diff
@@ -17,7 +17,7 @@ import {
     TemplateResult,
     query,
     PropertyValues,
-} from 'lit-element';
+} from '@spectrum-web-components/base';
```
andrewhatran, line 20: I noticed a ux bug with the `rtl` slider:

slider_bug

We may need to update our style getters to support the `rtl` direction

packages/tabs/src/Tabs.ts
```diff
@@ -16,7 +16,7 @@ import {
     CSSResultArray,
     TemplateResult,
     PropertyValues,
-} from 'lit-element';
+} from '@spectrum-web-components/base';
```
andrewhatran, line 19: The margins in the `rtl` tabs look a little bit off: 

Screen Shot 2020-08-03 at 1 27 45 PM

We may need to update some of the overriding css in `tabs.css` to support `rtl`

packages/tags/src/Tag.ts
andrewhatran: I'm down for prioritizing the Split Button PR and resolving this during the rebase  👍

packages/tabs/src/tab.css
andrewhatran: Ah okay, makes sense!

packages/tabs/src/Tabs.ts
Westbrook: Good catch, great opportunity to update out processing so we don't need that manual override anymore!

packages/slider/src/Slider.ts
Westbrook: I added some of the storybook settings to make testing this easier, but my plan is to add it systematically after all of the baseline PRs are in and then turn on visual regressions for those URLs as well.

Westbrook:
I'll hold off on managing this PR until the `Card` work is through to project branch, so no udpated golden images, etc. But, happy to take any notes in advance to help speed up the round trip on this once it is first in line!

Westbrook:
Looks like we're ready for the full review here. Then we'll be very close to getting this `[dir]` situation over the line!

packages/split-button/stories/split-button.stories.ts
```diff
@@ -119,7 +119,11 @@ moreCta.story = {
 export const moreCtaOpen = (options = {}): TemplateResult => {
     return html`
         <div>
+<<<<<<< HEAD
```
andrewhatran, line 122: Looks like there was a bad merge here?

packages/tabs/src/Tabs.ts
```diff
@@ -16,7 +16,7 @@ import {
...
 import { Tab } from './Tab.js';
 import { Focusable, getActiveElement } from '@spectrum-web-components/shared';
```
andrewhatran, line 22: I also noticed that the `seletionIndicator` line was not showing up for the `rtl` tabs. 

We may need to go into `updateSelectionIndicator()` and add this conditional to cover the rtl case:

```
if (this.direction === 'horizontal' && this.dir === 'rtl') {
    const width = tabBoundingClientRect.width;
    const offset = tabBoundingClientRect.right - parentBoundingClientRect.right - width;

    this.selectionIndicatorStyle = `transform: translateX(${offset}px) scaleX(${width});`;
} 
```

andrewhatran:
This is getting really close 🤞 

Just a couple of things and it's good to go!

packages/split-button/stories/split-button.stories.ts
Westbrook: Fixed.

packages/tabs/src/Tabs.ts
Westbrook: Yeah, rather than have two versions I overrode the RTL CSS so that the same positioning would work on the indicator in both contexts.

Westbrook:
Once this is in the feature branch, I'd say we'd want to address the following (if possible) before a full merge to `main`:
- ensure all storybooks have `dir=rtl` support
- extend visual regressions to include both `[dir]` values
- remove use of `if (!this.shadowRoot) return;`
- manage `dir` with MutationObserver on `document`/`<sp-theme>` 

@andrewhatran any thing else that's come up in our work on this so far that I'm missing.

Issue #802, opened 2020-08-04
What is the purpose of this line?


morganbarrett:
[https://github.com/adobe/spectrum-web-components/blob/331149d0f15d22e841968592043137f10bf00238/packages/overlay/src/overlay-stack.ts#L172](https://github.com/adobe/spectrum-web-components/blob/331149d0f15d22e841968592043137f10bf00238/packages/overlay/src/overlay-stack.ts#L172
)
I only ask because it adds an extra weird chunk in webpack builds, so I had a poke around and traced it back to this dynamic import, that is already imported? 😕

Westbrook:
@morganbarrett, thanks for reaching out. It's great timing as we're discussing this same line in #801. In support of both conversations, please checkout https://github.com/adobe/spectrum-web-components/pull/804 and let me know what you're thoughts are on the possibility of a slightly larger, hopefully more useful check, as well as the documentation/examples available around how to choose not to have that in webpack/rollup contexts.

Westbrook:
This is interesting:
```
new webpack.IgnorePlugin( /^\.\.\/active\-overlay\.js$/, /\@spectrum\-web\-components\/overlay/ )
```
But, I can't seem to get it to do what I think you've done. Even with it included a webpack build still segments off that small chunk.

Are you using the component that relies on this directly or are you leveraging `@specrum-web-components/bundle/elements`?

morganbarrett:
Yeah, I'm just importing `@spectrum-web-components/bundle`

Westbrook:
Strange that `@spectrum-web-components/bundle` would work appropriately with that file excluded... 🤔 I'll continue my research with this info, but unless you explicitly include `@specrum-web-components/bundle/elements` the `IgnorePlugin` settings _should_ be removing important side effects when opening overlays or dropdowns, etc. 

Generally, as outlined in the docs for that package: https://opensource.adobe.com/spectrum-web-components/components/bundle, while good for prototyping, I don't think you're going to get the highest quality build when moving to production with this technique. Thanks for beating the bushes on it, however, as getting these specific cases clarified is gonna be helpful for future users.

morganbarrett:
Just to note, I'm now importing `@spectrum-web-components/overlay/active-overlay.js`, `@spectrum-web-components/overlay/overlay-trigger.js` with no apparent issues to the functioning of the packages.

Westbrook:
[...]

PR #804, opened 2020-08-04
Sequester use interaction requisite JS behind dynamic imports

### LINKED ISSUE #801: sp-dropdown async loads active-overlay on open.
### Expected Behaviour
Loading a component should load all dependencies needed for it to function by default.

### Actual Behaviour
The default behavior adds network latency to load the overlay script the first time you open a dropdown. On high latency connections this won't be a nice experience. 

While having the option to delay-load the overlay dependencies seems good, it doesn't seem like the right default behavior. I don't think it's expected, and it also adds an extra potential failure case if the async load fails.

Westbrook:
## Description
This outlines the complete practice that was implied by [https://github.com/adobe/spectrum-web-components/blob/331149d0f15d22e841968592043137f10bf00238/packages/overlay/src/overlay-stack.ts#L172](https://github.com/adobe/spectrum-web-components/blob/331149d0f15d22e841968592043137f10bf00238/packages/overlay/src/overlay-stack.ts#L172
) that sequesters JS that requires user interaction to be needed/used behind a dynamic import.

Specifically, this allows an intelligent build process to keep 10.04/31.6 KB of JS (gzipped/parsed) out of the initial load/parse of an application.

This also expands the webpack example, and adds a Rollup example to outline how a user could choose not to leverage this optimization in your code.

## Related Issue
refs #802
fixes #801 

## Motivation and Context
Open the broader discussion of how optimized/opinionated we want to deliver out code by default and the level of documentation/examples required to outline the options for working with that code.

## How Has This Been Tested?
- existing tests

## Types of changes

## Checklist:


projects/example-project-webpack/src/index.js
```diff
@@ -0,0 +1,26 @@
...
+import '@spectrum-web-components/button/sp-button';
+import '@spectrum-web-components/dropdown/sp-dropdown';
+import '@spectrum-web-components/menu/sp-menu';
+import '@spectrum-web-components/menu/sp-menu-item';
+
+/**
+ * Uncomment this following line to force webpack to include dynamically
+ * imported code powering the `<sp-dropdown>` in the main JS bundle.
+ **/
+// export { Overlay } from '@spectrum-web-components/overlay';
```
Westbrook, line 26: By default this will structure chunks like so:
image
image

When it is uncommented, you'll receive the single bundle build, like so:
image

projects/example-project-rollup/src/example-app.ts
```diff
@@ -0,0 +1,37 @@
...
+window.process.env.NODE_ENV = window.process.env.NODE_ENV || 'production';
+
+/**
+ * Uncomment this following line to allow rollup to include dynamically
+ * imported code whereever it thinks it should best be bundled. In particular,
+ * @open-wc/building-rollup currently uses a babel feature that will still split
+ * this into more than a single bundle.
+ * https://github.com/open-wc/open-wc/issues/1625#issuecomment-632158102
+ **/
+// export { Overlay } from '@spectrum-web-components/overlay';
```
Westbrook, line 37: While https://github.com/open-wc/open-wc/issues/1625#issuecomment-632158102 forces the OpenWC tooling to _always_ break the bundle, with this commented out, you receive bundles structured by the dynamic import:
image

And with it available, the bundles come out as:
image

I taking a look with the team there to see if we can prioritize getting the default (force multiple bundle) functionality corrected.

projects/example-project-webpack/webpack.config.js
```diff
@@ -126,6 +126,20 @@ const shared = (env) => {
             host: '0.0.0.0',
             disableHostCheck: true,
         },
+        optimization: {
+            splitChunks: {
+                cacheGroups: {
+                    vendor: {
+                        // Work around to "trick" webpack to see local packages as "node_modules".
+                        test: /[\\/](node_modules|packages)[\\/]/,
```
Westbrook, line 134: Only required for the monorepo.

projects/example-project-webpack/webpack.config.js
```diff
@@ -126,6 +126,20 @@ const shared = (env) => {
...
+                        name: 'vendor',
+                        // "async" will ensure the dynamic imports mark the boundary between chunks
+                        // "all" will allow webpack to make decisions about chunk breaking for itself
+                        chunks: 'async',
```
Westbrook, line 138: Managing this directly is something that _maybe_ applications should be doing anyways? `all` ensures you're using a single "external" bundle or vendor code, and then a small bundle for your application code:
image
image
Kind on interesting as it sequesters "low change rate" vendor code to its own bundle _AND_ chooses NOT to use the dynamic import boundary to define chunks.

packages/overlay/src/loader.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export const openOverlay = async (
+    target: HTMLElement,
+    interaction: TriggerInteractions,
+    content: HTMLElement,
+    options: OverlayOptions
+): Promise<() => void> => {
+    const { Overlay } = await import('./overlay.js');
+    return await Overlay.open(target, interaction, content, options);
+};
```
Westbrook, line 23: This centralized the load pattern for lazy `openOverlay` functionality into an easily to leverage export. This will make documenting this `Overlay` API much easier, a task I've avoided a little until now. 😬 

As per @adixon-adobe:
> The default behavior adds network latency to load the overlay script the first time you open a dropdown. On high latency connections this won't be a nice experience.

This is a valid point. However, there is no avoiding latency, it's only a question of where you put it. We can block initial load, but making this eager loading, and in high latency environments this would extend that blocking. Having it lazy allows for developers to really choose how to tune this in a way that can't be broken apart after the fact. In this way tools like `rel="preload"`, `rel="modulepreload"`, service workers etc. can be leveraged to ensure that the code is available (and even parsed if you want it to be) before it's requested without effecting the initial load of the page/functionality. The service worker can then also be used to prevent that load/latency hit in the future. I'd like to prefer having high control in this area, but I can be convinced (especially for the 10KB we're talking about) that it is overkill.

morganbarrett:
#802 As a workaround, I ended up just adding the following plugin to my webpack config
`new webpack.IgnorePlugin(
	/^\.\.\/active\-overlay\.js$/,
	/\@spectrum\-web\-components\/overlay/
)`

packages/overlay/src/loader.ts
adixon-adobe: I think having high control here is good overall, but I also don't like the default behavior.

It's easy for me to reason about importing `sp-dropdown` and how to structure our lazy-loading so that we only load that code when it's needed. The current default behavior here requires me to know that dropdown uses overlay and that overlay has an async import to have control over it. And given no control, there's a hidden network error condition in the implementation (if that async load fails, the dropdown won't open, and it's likely I wouldn't have any error handling for that scenario).

While I see the appeal of the current approach (you don't need to add a separate import for the async load case), I also like the idea of having async loading like this be "opt-in".

Westbrook: The hope is that you don't have to know anything and that with a quality webpack or Rollup set up things work the best way they possibly could. 🤞 😬 

Have you seen a pattern when you're able to opt-in to lazy loading? Seems like what you're asking for, but I can't think of a way to do it. The minute you move to `import { Overlay } from '../overlay'` you trapped with a static dependency and the builder magic to make that go away is intense. A implementor _could_ move that up on level and lazily load dropdown for themselves, but that still means they're loading code a user might not use.

The only sort of practical approach that I can see if having a dropdown that is eager and another that is lazy, but then you're either trading in multiple tag names or the high likely hood of double (read one that throws an error or one that's difficult to ensure is the one you are expecting) registrations.

adixon-adobe: Well I always try to think worst-case for anything async which is why I bring up the error condition 😁 

Just to check my understanding though, say you have 2 imports that register `sp-dropdown`. `@spectrum-web-components/dropdown/sp-dropdown` and `@spectrum-web-components/dropdown/sp-dropdown-async`. And effectively the contents of the former is:

```
import '@spectrum-web-components/overlay/active-overlay';
import './sp-dropdown-async';
```

Would there be any issue with that technically? (there's an obvious maintenance issue with this approach, particularly if it becomes more widespread than overlay)

Westbrook: @adixon-adobe this version uses `@spectrum-web-components/dropdown/sp-dropdown` and `@spectrum-web-components/dropdown/sync/sp-dropdown` as I feel pretty strongly that the default should be the version that's most "optimal" on the first load. What's more, while build tooling can take away async with various settings, nothing can add it when it's not, so having it as the default seems like the best way for users to fully understand the power of these elements. If anyone else has feelings on the default of async vs sync, let's chat, happy to hear use cases that this might endanger.

This update effects the `@spectrum-web-components/overlay/overlay-trigger` registration as well, adding `@spectrum-web-components/overlay/sync/. Both packages now have extended READMEs to reference this new feature, and the exports/side-effects in package.json have been updated as well.

adixon-adobe: This seems reasonable (and with some tweaks to our linter, something we can enforce either way) -- thanks!

I do disagree about the default behavior here (or at least am not fully convinced), but definitely not as strongly as you. For me it's about making the default the simplest approach to understand, with the most expected behavior. I think clients that aren't zealous about keeping their package size to an absolute minimum might be surprised by this.

The fact that there's a side-effect of registering a shadow DOM on the `body` tag from that async import also makes this case feel a little less good to me.

I'm definitely not gonna hold things up here, and will be curious to hear what other projects think of this approach. We have the knowledge/tooling to do what we want here, and overall I *really* appreciate the focus on minimal footprint.

packages/dropdown/README.md
```diff
@@ -17,6 +17,12 @@ Import the side effectful registration of `<sp-dropdown>` via:
 import '@spectrum-web-components/dropdown/sp-dropdown.js';
 ```
 
+The default of `<sp-dropdown>` will load dependencies in `@spectrum-web-components/overlay` asynchronously via a dynamic import. In the case that you would like to import those tranverse dependencies statically, import the side effectful registration of `<sp-dropdown>` as follows:
+
+```
+import '@spectrum-web-components/dropdown/sync/sp-dropdown.js';
+```
```
adixon-adobe, line 24: Awesome

PR #812, opened 2020-08-05
build: clean up old ce.json files and ensure presence before publishing

### LINKED ISSUE #811: No custom-elements.json for sp-dropdown
### Expected Behaviour
There should be a custom-elements.json file in 0.6.4 like there was in 0.6.3

### Actual Behaviour
It's gone! https://unpkg.com/browse/@spectrum-web-components/dropdown@0.6.4/

But it was here: https://unpkg.com/browse/@spectrum-web-components/dropdown@0.6.3/

Westbrook:
## Description
- clean up old `custom-elements.json` files before creating them
- correct globbing for `custom-elements.json` creation
- ensure `custom-elements.json` files are present before allowing packages to be published

## Related Issue
fixes #811 

## Motivation and Context
Packages need `custom-elements.json` files.

## Types of changes

## Checklist:


package.json
```diff
@@ -54,8 +54,9 @@
         "lint:ts": "eslint -f pretty 'packages/**/*.ts'",
         "lint:css": "stylelint 'packages/**/*.css'",
         "lint:versions": "node ./scripts/lint-versions.js",
-        "custom-element-json": "lerna exec --ignore @spectrum-web-components/styles -- wca analyze 'src/**/*.ts' --format json --outFile custom-elements.json",
-        "postcustom-element-json": "lerna exec --ignore @spectrum-web-components/styles -- node ../../scripts/add-custom-properties.js --src='custom-elements.json'",
+        "precustom-element-json": "rm packages/*/custom-elements.json",
```
Westbrook, line 57: Clean up all files before creating them again.

package.json
```diff
@@ -54,8 +54,9 @@
...
+        "custom-element-json": "lerna exec --ignore @spectrum-web-components/styles -- wca analyze '*.d.ts' --format json --outFile custom-elements.json",
+        "postcustom-element-json": "lerna exec --ignore '{@spectrum-web-components/{bundle,,icons-ui,icons-workflow,iconset,shared,styles},example-project,swc-templates}' -- test -f custom-elements.json",
```
Westbrook, line 59: Ensure that packages where a file is expected have a file before moving on, usually to `lerna-publish`.

Westbrook: This _doesn't_ ensure content...not sure of a good way to do so right this minute. If anyone have a good idea, I'm all ears. If not, we can come back to that.

adixon-adobe: Might help to have some sanity script that just checks for the presence of something you'd expect in each package, or maybe just presence and a minimum file size?

Westbrook: This is an interesting approach. My first thought is to also to `test -f **/*.js` to ensure there is JS in the build?

What other things might be check for? Seems pretty ad hoc, has to be something better, but agree that anything we can do to ensure a quality publish would be nice. Let's follow up on options in #816

PR #818, opened 2020-08-07
feat(icons): added workflow component

morganbarrett:
## Types of changes

## Checklist:


Westbrook:
Hi Morgan, thanks for you interest in this. We've had some back and forth on this subject in some issues that might be useful to have centralized for review, but as this approach would require an non-treeshakable 937 KB file in order to deliver the icons you're looking at, we will not be able to accept a PR in this shape. If you were interested in opening an issue, I can try to collect the notes from the various conversations we've had to date to help direct your work towards something that would both deliver these icons for other users as well as prepare them for successfully delivering a performant web application.

morganbarrett:
No worries, I thought as much, but I was interested in having such a component and thought I'd at least offer it up after making it. Would splitting the icons into thematic categories improve the issue?

Westbrook:
The working idea at this point is to take the individual icon SVGs and wrap them as components themselves. So instead of:
```
import '@spectrum-web-components/icon';
import { CircleIcon } from '@spectrum-web-components/icons-workflow';

// ...

<sp-icon slot="icon">
    ${CircleIcon()}
</sp-icon>
```
You would end up with something along the lines of:
```
import '@spectrum-web-components/icons-workflow/sp-circle-icon';

// ...

<sp-circle-icon slot="icon"></sp-circle-icon>
```
There are a couple of other options that I've seen and been interested by, but I don't feel that any others accurately capture the tree shakability plus lack of pre-configuration that I think should be the default when working with icons.

In that the main benefit of the updated approach is that you could then use the icon element in a non-JS templating environment, it is not been something that we've prioritized in the library so far. It's likely more scripting than building, as you'd want these to be generated dynamically (IMO), so hopefully it's a measure twice get 800+ icons sort of thing.

morganbarrett:
I think the amount of imports required would be mad, and variable icons would be over complicated. I'll put together a little proposal now.

Westbrook:
It’s hard for me to believe that any one component would use more than 10 or so icons, and even then I’d say the component is likely managing too much to be easily maintained. Looking forward to your thoughts to see if we can’t get to a better solution, however. If you don’t mind, can you submit the proposal via an issue? I’m going to close the PR as I think we’ll have some rounds of conceptual back and forth on this before actual code can be submitted again.

PR #822, opened 2020-08-11
fix(overlay): ensure overlay addition occurs after closing previous

### LINKED ISSUE #809: Can't open a new dropdown while one is open
### Expected Behaviour
If you have a dropdown open, and click on another dropdown, the first dropdown closes, and the second opens.

### Actual Behaviour
The first dropdown stays open, and the second flickers open then closes immediately.

Easy to repro on https://opensource.adobe.com/spectrum-web-components/components/dropdown/examples

Westbrook:
## Description
When an overlay is open when another is triggered, if the previous element isn't closed before the new overlay has opened, then the wrong overlay will be closed returning to the first overlay.

The dynamic import timing by default (and in builds _not_ made by webpack) seems to have managed this fine in the past, this update ensures the timing is correct in all cases and closer to the same between contexts.

## Related Issue
fixes #809

## Motivation and Context
We expect to be able to open successive overlays without issue.

## How Has This Been Tested?
- added new unit tests, passes old ones.

## Types of changes

## Checklist:

### DIFF
--- a/packages/overlay/src/overlay-stack.ts
+++ b/packages/overlay/src/overlay-stack.ts
@@ -173,19 +173,22 @@ export class OverlayStack {
         }
 
         const activeOverlay = ActiveOverlay.create(details);
-        this.overlays.push(activeOverlay);
         document.body.appendChild(activeOverlay);
-        let updateComplete = await activeOverlay.updateComplete;
-        while (updateComplete === false) {
-            updateComplete = await activeOverlay.updateComplete;
-        }
 
-        this.addOverlayEventListeners(activeOverlay);
-        if (details.receivesFocus === 'auto') {
-            activeOverlay.focus();
-        }
-
-        return false;
+        /**
+         * The following work to make the new overlay the "top" of the stack
+         * has to happen AFTER the current call stack completes in case there
+         * is work there in to remove the previous "top" overlay.
+         */
+        return Promise.resolve().then(async () => {
+            this.overlays.push(activeOverlay);
+            await activeOverlay.updateComplete;
+            this.addOverlayEventListeners(activeOverlay);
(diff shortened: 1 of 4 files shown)


packages/overlay/src/overlay-stack.ts
```diff
@@ -173,19 +173,23 @@ export class OverlayStack {
...
-            activeOverlay.focus();
-        }
+        return new Promise((res) => {
+            setTimeout(res);
+        }).then(async () => {
+            this.overlays.push(activeOverlay);
+            let updateComplete = await activeOverlay.updateComplete;
+            while (updateComplete === false) {
+                updateComplete = await activeOverlay.updateComplete;
+            }
```
adixon-adobe, line 185: it's a little unfortunate you need 4 lines for this "wait until fully updated" check. It'd be nice if litElement had a utility. like `allPendingUpdatesComplete`

Westbrook: This is less specifically `LitElement` than the way we acquire the overlay content after the first render in `ActiveOverlay`. I should be able to clean this up...

packages/overlay/src/overlay-stack.ts
```diff
@@ -173,19 +173,23 @@ export class OverlayStack {
...
-        while (updateComplete === false) {
-            updateComplete = await activeOverlay.updateComplete;
-        }
 
-        this.addOverlayEventListeners(activeOverlay);
-        if (details.receivesFocus === 'auto') {
-            activeOverlay.focus();
-        }
+        return new Promise((res) => {
+            setTimeout(res);
```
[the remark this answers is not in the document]
Westbrook: A yes, good call on `await Promise.resolve()`, updated with comment!

PR #824, opened 2020-08-12
refactor: use shared slot presence mixin for conditional element styling

### LINKED ISSUE #793: Review `has...` testing for slot generation.

Westbrook:
## Description
- Catch up to `main` and the keyboard helper work that is there.
- Move to using `@spectrum-web-components/base` as the only entry to `lit-element` and `lit-html`.
- update to latest `lit-html`/`lit-element`
- use new `lit-element` to power the below...

Querying light DOM children to resolve template content is not reactive:
```
private get hasFooter(): boolean {	
        return !!this.querySelector('[slot="footer"]');	
}
```
Meaning that changing the presence of this content dynamically will not inherently trigger a new render.

Relying on a `MutationObserver` instead will allow the value of `hasFooter` or the like to be dynamic to content changes and ensure a more correct delivery of the elements template across its lifecycle.

Refactors `ObserveSlotText` to better leverage `queryAssignedNodes`. I see the possibility of merging the two observer mixins in some way in the future, but for now the use of `Symbol()` based properties allows them to neatly avoid stepping on each other without further refactoring.

## Related Issue
fixes #793 

## Motivation and Context
Templates should be correct. That's the tweet...oh, wait.

## How Has This Been Tested?
- existing tests
- new tests

## Types of changes
- [x] feature factor

## Checklist:


packages/shared/README.md
```diff
@@ -67,12 +67,47 @@ Use this helper to find an `activeElement` in your component. [Learn more about
 
 Mix `download`, `label`, `href`, and `target` properties into your element to allow it to act more like an `HTMLAnchorElement`.
 
-### ObserverSlotText
+### ObserveSlotPresence
+
+When working with styles that are driven by the conditional presence of `<slot>`s in a component's shadow DOM, you will need to track whether light DOM that is meant for that slot. Use the `ObserveSlotPresence` mixin to target specific light DOM to observe the presence of and trigger `this.requestUpdate()` calls when content fulfilling that selector comes in and out of availability.
```
andrewhatran, line 72: Did you mean:

`you will need to track whether light DOM that is meant for that slot exists`?

packages/shared/src/observe-slot-text.ts
```diff
@@ -9,10 +9,15 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
...
+import {
+    PropertyValues,
+    UpdatingElement,
+    queryAssignedNodes,
+    property,
+} from 'lit-element';
 
-const observedSlotElement = Symbol('observedSlotElement');
 const slotElementObserver = Symbol('slotElementObserver');
+const assignedNodesList = Symbol('assinedNodes');
```
andrewhatran, line 20:
```suggestion
const assignedNodesList = Symbol('assignedNodes');
```

packages/base/README.md
```diff
@@ -34,6 +34,8 @@ export MyElement extends SpectrumMixin(HTMLElement) {}
...
 
+When placing a `dir` attribute on an element built from `SpectrumMixin` before attaching it to the DOM, this will prevent it from resolving the text direction value to a parent `sp-theme` or `document` element. Elements applied to the page in this way will also NOT participate in observing any such elements, so their `dir` values will remain as initialized regardless of changes in other parts of your documents. If you choose to leverage this, be aware that any child (in both light DOM and shadow DOM) of this element will need to have a `dir` attribute applied as well if you do not want it resolving to a parent `sp-theme` or `document` element itself. In this way, it is likely that you would benefit from leveraging an `sp-theme` element to create scope in your document for managing this custom content direction section of your page.
```
andrewhatran, line 37:
```suggestion
Placing a `dir` attribute on an element built from `SpectrumMixin` before attaching it to the DOM will prevent it from resolving the text direction value to a parent `sp-theme` or `document` element. Elements applied to the page in this way will also NOT participate in observing any such elements, so their `dir` values will remain as initialized regardless of changes in other parts of your documents. If you choose to leverage this, be aware that any child (in both light DOM and shadow DOM) of this element will need to have a `dir` attribute applied as well if you do not want it resolving to a parent `sp-theme` or `document` element itself. In this way, it is likely that you would benefit from leveraging an `sp-theme` element to create scope in your document for managing this custom content direction section of your page.
```

packages/split-button/README.md
```diff
@@ -1,11 +1,11 @@
 ## Description
 
-`sp-splitbutton` delivers a
+An `sp-split-button` surfaces an immediately envokable action via it's main button, as well as a list of alternative actions in its toggleable menu overlay. By default, any actions envoked from the overlay will replace the main action button. When leveraging `[type="more"]` the action will be envoked, but the main button will remain the action initally persribed by the implementor.
```
adixon-adobe, line 3: I'd suggest changing `envoke` to `invoke` since it's more common (and I know my spell check is really wanting to change it as I'm trying to give this feedback). I did learn today that `envoke` is a valid alternate spelling though 🌈

PR #827, opened 2020-08-14
build: use more strict tsconfig settings

Westbrook:
## Description 
Using more strict tsconfig settings will support downstream consumers in doing the same.

## Motivation and Context
Everyone should be able to extend out classes.

## How Has This Been Tested?
- tslint

## Types of changes
- [x] refactor

## Checklist:


packages/slider/src/Slider.ts
```diff
@@ -223,7 +223,7 @@ export class Slider extends Focusable {
         return html`
             <div class="ticks">
                 ${ticks.map(
-                    (tick, i) => html`
+                    (_, i) => html`
```
CoreyDotCom, line 226: You could name this _tick to still convey intent. Ditto for above 'size' change (_size)

Westbrook: Cool that it works both ways. I'd not used this feature in TS before, but now that we've got this on it's good to know my options. That's for the support in getting this addressed

PR #837, opened 2020-08-25
test: use web-test-runner instead of karma

Westbrook:
## Description
Change our test runner from [karma](https://karma-runner.github.io/latest/index.html) to [web-test-runner](https://modern-web.dev/docs/test-runner/overview/)
- rely on Playwright as the runner
- expand testing locations to include WebKit
- Firefox exhibits [issues with focus](https://github.com/microsoft/playwright/issues/3476), so run those tests as a separate pass to manage concurrency in that sequence (somehow results turnout in the same or less time 😱)
  - it'll be a bit of a race to see which issue reduces this complexity first as web-test-runner will also be adding: https://github.com/modernweb-dev/web/issues/369
- switch to Playwrights docker image for CircleCI (has lots of added dependencies for their browser pathes.
  - Currently we have to manually install yarn, but the image is going to be extended with [that and git](https://github.com/microsoft/playwright/issues/3529) sometime soon and we can remove it.
- this change moved completely away from snapshot testing as we had already started with the move to a11y first testing
- coverage number change due to the way that istanbul slightly inflates numbers by turning `if (something) { method(); }` into `if (something) { method(); } else {}` and giving the extra line/branch/etc to the coverage reporter.
- moves to `/* c8 ignore next ... */` syntax for ignoring lines for text coverage

The console for the test runner is a little chatty as we're using Storybook Component Story Format in a number of our tests. If this is annoying we could do some work to reduce this, but I've not been bothered too much when the "all passed" returns.

## Motivation and Context
- test more
- test faster

## How Has This Been Tested?
- it _is_ tests

## Types of changes
- [x] test suite changes

## Checklist:


web-test-runner.config.js
```diff
@@ -0,0 +1,41 @@
...
+            lines: 98,
+        },
+    },
+    testRunnerHtml: (testRunnerImport) => `
+        <html>
+        <head></head>
+        <body>
+            <script type="module">
+            import '${testRunnerImport}';
+            window.process = window.process || {};
```
Westbrook, line 30: This because of Popper...

packages/dropdown/test/dropdown.test.ts
```diff
@@ -457,13 +455,12 @@ describe('Dropdown', () => {
...
     });
 
-    it('resets value when item not available', async () => {
+    it('allows event listeners on child items', async () => {
         const mouseenterSpy = spy();
         const handleMouseenter = (): void => mouseenterSpy();
         const el = await fixture<Dropdown>(
             html`
                 <sp-dropdown
-                    value="missing"
```
Westbrook, line 466: Double check this is a correct change to the test...

andrewhatran: Ah just checked the dropdown class. In the `updated` method, Dropdown clears it's value if the value doesn't match with a `Menu Item`.

Maybe we can leave these two lines in? or separate it into another test

Westbrook: Good eye!

After taking a look at the full file, I remember why it's being removed, however.

The _actual_ `'resets value when item not available'`, one test above in the file, already tests that functionality, so this was clarifying the intent of the test in question.

So, removed and separated into another test, already!

packages/action-menu/package.json
```diff
@@ -42,7 +42,7 @@
         "./sp-action-menu.ts"
     ],
     "scripts": {
-        "test": "karma start --coverage"
+        "test": "echo \"Error: run tests from mono-repo root.\" && exit 1"
```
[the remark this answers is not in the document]
adixon-adobe: cool

packages/button/test/button.test.ts
```diff
@@ -141,8 +139,6 @@ describe('Button', () => {
         await elementUpdated(el);
         expect(el).to.not.be.undefined;
         expect(el.textContent).to.include('With Target');
-        // make sure href is being passed to <a>
-        expect(el).shadowDom.to.equalSnapshot();
```
adixon-adobe, line 145: why did these need to be removed?

packages/checkbox/test/checkbox.test.ts
```diff
@@ -178,6 +178,7 @@ describe('Checkbox', () => {
         HTMLElement.prototype.focus.apply(el);
         el.focusElement.dispatchEvent(new CustomEvent('focusout'));
         await nextFrame();
+        await nextFrame();
```
adixon-adobe, line 181: never fun :/

packages/button/test/button.test.ts
adixon-adobe: (see your note about snapshot testing, so nevermind!)

packages/dropzone/test/dropzone.test.ts
```diff
@@ -87,17 +87,23 @@ describe('Dropzone', () => {
...
-        el.dispatchEvent(dragOverEvent);
-
-        expect(el.isDragged).to.be.true;
-        // We should be able to make the following test here:
-        // expect(dataTransfer.dropEffect).to.equal(el.dropEffect);
-        // However, Chrome doesn't like it in the context of a test...
+        let dataTransfer: DataTransfer | boolean = false;
+        try {
+            // Safari doesn't like this...
+            dataTransfer = new DataTransfer();
```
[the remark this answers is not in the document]
adixon-adobe: It'd be a little nicer if you could explicitly disable Safari for this test (right now if another browser broke, or someone made bad changes to this file, it has the potential to "pass" when it shouldn't.

web-test-runner.config.js
adixon-adobe: might be worth adding a comment if that's the case

packages/dropzone/test/dropzone.test.ts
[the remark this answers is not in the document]
Westbrook: I'll check out whether there's an option for that in the runner. If not, we CAN add a `console.log()` in the `catch` so that it will at least report it to us...do you want to hold this PR or come back to this?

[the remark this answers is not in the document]
Westbrook: Just saw your other comment, I think moving forward and then catching these two singularly will be the right call, thanks!

PR #846, opened 2020-08-27
ci: use playwright for visual regression testing

Westbrook:
## Description
This catches the tooling there up to #837 where we're also moving unit testing to a Playwright driven workflow with the help of web-test-runner. It's looking like we get a little speed bump this way, too, 🤞, though I feel like this is something I initially saw with some other changes in the past that later dissolved.

## Motivation and Context
One or both of Puppeteer and/or Chrome went crazy yesterday and will no longer load for the visual regression suite.

## How Has This Been Tested?
It _is_ tests.

## Types of changes

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,7 +43,7 @@ commands:
             - restore_cache:
                   name: Restore Golden Images Cache
                   keys:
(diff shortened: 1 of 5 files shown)


adixon-adobe:
Looks good. Glad the switch here is so straighforward.

PR #862, opened 2020-09-01
chore: add css/ts watch to docs:start

### LINKED ISSUE #834: docs:start doesn't build component changes
### Expected Behaviour
If you update a component while developing, the docs should update automatically.

### Actual Behaviour
It looks like only changes to the doc pages are rebuilt.

#### Steps to Reproduce
1. Run `yarn docs:watch`
2. Make an obvious css change to action-menu.css
3. Open the docs page for action menu

Westbrook:
## Description 
Allow for continuous development with CSS and TS watching while running the docs site locally.

## Related Issue
fixes #834

## Testing
@adixon-adobe can you confirm that this acts as you expect? It's not the _fastest_ live reload as we're doing `gulp css` + `tsc packages` + `webpack site`, but I was able to "live" add a `:host:before { content: 'Hello' }` style as per your issue and it updated.


adixon-adobe (approved):
Agree it's not fast, but it's 1000% improved. Thanks!!!

PR #863, opened 2020-09-02
ci: use next docker release

Westbrook:
## Description
The Docker image we updated to in order to leverage `web-test-runner` was missing a couple of packages to support publishing the docs site from CI. This has now been corrected, and this PR points to that release. We should be able to move back to `bionic` the main line at some point in the next week or two, but this will allow us to avoid manually publishing the docs site in the interim.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,23 +2,13 @@ version: 2.1
 executors:
     node:
         docker:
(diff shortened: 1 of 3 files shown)


andrewhatran (approved):
Lgtm 👍

PR #870, opened 2020-09-04
fix(top-nav): prototype top-nav pattern

### LINKED ISSUE #608: sp-tab should support "href" attribute
### Expected Behaviour

I want to be able to have Tabs that are used for navigation and consist of a number of links.

### Actual Behavior

The `href` attribute does nothing.

#### Platform and Version

`@spectrum-web-components/tab@0.3.6`
`@spectrum-web-components/tab-list@0.3.6`

```html
<sp-tab-list selected="1">
    <sp-tab href="/page1" label="Tab 1" value="1" tabindex="1"></sp-tab>
    <sp-tab href="/page2" label="Tab 2" value="2" tabindex="2"></sp-tab>
    <sp-tab href="/page3" label="Tab 3" value="3" tabindex="3"></sp-tab>
    <sp-tab href="/page4" label="Tab 4" value="4" tabindex="4"></sp-tab>
</sp-tab-list>
```

#### Context

These tabs are being used as top-level navigation for website. As such, I'll also need to apply `role="navigation"`, which I assume will work just fine on `<sp-tab-list>`, but would like some insight into whether that is the expected implementation.

image

Westbrook:
## Description 
Add the `sp-top-nav` pattern as outlined by https://opensource.adobe.com/spectrum-css/tabs.html, plus https://spectrum.adobe.com/page/headers/#Anatomy, and appropriate accessibility/UX for a "TopNav" pattern.

To do:
- [x] add screenshot testing
- [x] add documentation
- [ ] add to bundle  <= questioning if we want to do this _just_ now as this is a highly prototypical element and it may be safer to wait a beat before adding it to the bundle, which would technically require a "breaking change" if we were to remove it later for whatever reason... 

## Related Issue
fixes #608 

## Motivation and Context
Applications should be easier to build with SWC.

## How Has This Been Tested?
- unit tests
- screenshot tests, soon

image

## Types of changes

## Checklist:

### DIFF
--- a/packages/top-nav/src/TopNavItem.ts
+++ b/packages/top-nav/src/TopNavItem.ts
@@ -1,0 +1,78 @@
+/*
+Copyright 2020 Adobe. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License. You may obtain a copy
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import {
+    html,
+    CSSResultArray,
+    TemplateResult,
+    PropertyValues,
+    ifDefined,
+    property,
+    query,
+} from '@spectrum-web-components/base';
+import { Focusable, LikeAnchor } from '@spectrum-web-components/shared';
+
+import itemStyles from '@spectrum-web-components/tabs/src/tab.css.js';
+import topNavItemStyles from './top-nav-item.css.js';
+
+/**
+ * @slot icon - The icon that appears on the left of the label
+ */
+
+export class TopNavItem extends LikeAnchor(Focusable) {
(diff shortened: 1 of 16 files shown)


simshanith (approved):
excited to get my hands on this pattern! 🙌

packages/top-nav/src/TopNavItem.ts
```diff
@@ -0,0 +1,78 @@
...
+        return this.anchor;
+    }
+
+    public click(): void {
+        this.anchor.click();
+    }
+
+    protected render(): TemplateResult {
+        return html`
+            <a
```
simshanith, line 54: Could potentially abstract into a directive or even plain vanilla function to share render logic across anchor-likes

Westbrook: This is _sort of_ what goes on in https://github.com/adobe/spectrum-web-components/blob/394fc44d18f52d3d490ad30b049372acae05b70f/packages/shared/src/like-anchor.ts#L58-L80, but we're still working out the sharing kinks for elements like this one that need a few custom attributes. I've been leaning away from using this [spread directive](https://open-wc.org/developing/lit-helpers.html#regular-spread) because I worry about it feeling a bit like a hack, but with [`lit-html@2`](https://github.com/Polymer/lit-html/issues/1182) thinking out the same it may be time to lean in. For this initial release, I'm gonna push off any decisions here, however.

Issue #848, opened 2020-08-28
Safari breaks with overlays and WebGL canvas


cuberoot:
### Expected Behaviour

We should be able to use spectrum-web-components alongside WebGL canvas elements in all modern browsers

### Actual Behaviour

Overlays (e.g. dropdown) seem to cause strange buffer flipping behaviour in Safari

![ScreenFlow](https://user-images.githubusercontent.com/154097/91514895-0b975c80-e93c-11ea-80ba-9840d116b0df.gif)

### Reproduce Scenario (including but not limited to)

Use an overlay next to a 3D canvas

#### Steps to Reproduce

- browse to https://tender-goodall-47cf7b.netlify.app with Safari
- repeatedly open and close the overlay
- it should stay black 😄 

If you want to build/run locally:
- clone https://github.com/cuberoot/swc-safari-bug
- `yarn install`
- `yarn start`

#### Platform and Version

Safari 13.1.1 on Mac

#### Sample Code that illustrates the problem

https://github.com/cuberoot/swc-safari-bug

cuberoot:
This problem seems to have occurred between SWC bundle v0.12 and v0.13. There is an `swc-bundle-0.12.0` in the `swc-safari-bug` project. If you build that branch, everything works.

I am wondering if this has to do with the change to put a shadow DOM onto the `body` element. It certainly seems to have appeared around when that change went in.

This is a pain point for us because we use WebGL canvas in LrWeb.

bengfarrell:
There's gotta be some weird nuance, despite your setup being so simple. I have a project that I'm working on that overlays a 2D canvas over a Babylon.js scene with a transparent background. I have a dropdown menu that opens over the stack of canvases. It all works in Safari without the flash. I just upgraded to your bundle version and despite some UI breakage from doing so, it's all good. I also went in and replaced my 2D canvas with your WebGL canvas code and it STILL works without the flash from opening the dropdown.

That said, I still verified the weird issue in your project.

No idea if that helps, but it's definitely weird

Westbrook:
Well...the is super fun. Looks like you can't _ever_ have a webGL `canvas` in a slot and append things to the DOM? Gonna going this down one step further and submit an issue to Safari, not that I expect it to get addressed anytime soon 😞 

[...]

PR #885, opened 2020-09-14
fix(overlay): only "tab trap" when you mean to

Westbrook:
## Description 
Safari has issues with WebGL canvas in association with out tab trapping solution. This mitigated the issue by initializing the tab trapping approach _only_ when tab trapping is being used by an element.

I have another idea that is low touch that I want to try, but all of my other ideas in this area haven't been working, so having something proposed as a possible path forward feels nice...

## Related Issue
refs #848 

## Motivation and Context
You should be able to have WebGL canvas AND use SWC.

## How Has This Been Tested?
- exiting tests pass.

## Types of changes
- [x] refactor

## Checklist:

### DIFF
--- a/packages/overlay/src/overlay-stack.ts
+++ b/packages/overlay/src/overlay-stack.ts
@@ -33,10 +33,10 @@ export class OverlayStack {
 
     public constructor() {
         this.addEventListeners();
-        this.initTabTrapping();
     }
 
     private canTabTrap = true;
+    private trappingInited = false;
     private tabTrapper!: HTMLElement;
     private overlayHolder!: HTMLElement;
 
@@ -90,6 +90,11 @@ export class OverlayStack {
     }
 
     private startTabTrapping(): void {
+        if (!this.trappingInited) {
+            this.initTabTrapping();
+            this.trappingInited = true;
+        }
+        /* c8 ignore next 3 */
         if (!this.canTabTrap) {
             return;
         }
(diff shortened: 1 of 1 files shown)


packages/overlay/src/overlay-stack.ts
```diff
@@ -90,6 +90,11 @@ export class OverlayStack {
     }
 
     private startTabTrapping(): void {
+        if (!this.trappingInited) {
+            this.initTabTrapping();
```
adixon-adobe, line 94: It might work a little better to have a `disableTabTrapping` option. I'd rather not cause all clients to do extra work if they're not running into the Safari issue.

Westbrook: Other than the boolean check, I don't particularly count this as extra work. I had previously been making this call in the `constructor()` which means if an application doesn't use tab trapping this change now means they do less work.

A universal `disabledTabTrapping` options is interesting, however. Where do you think you'd want to apply that? Currently, there are no packages that leverage tab trapping on their own (IIRC), so an implementing app could also not use `type="modal"` (or manage their own usage of that value) and this would never be needed in that application. 

This does point to the benefit of a more durable solution. To prevent such a solution from adding work to non-Safari browsers, what are you thoughts on in component browser checking?

adixon-adobe: Oh never mind -- I just misunderstood this logic. I'm not even sure how 🙈. This looks perfect.

adixon-adobe (approved):
Awesome -- thanks!

PR #871, opened 2020-09-08
Potpourri of tweaks to tests, builds, template, etc.

Westbrook:
## Description
- update side effect listings for easier maintanence
- clean up method naming in `sp-tooltip`
- refactor the order of work in checked state tracking in `sp-tabs`
- get `<sp-icons-medium>` into the `sp-rule` docs
- change testing pattern for `sp-dialog`
- make `sp-button` tests more stable
- make the plop templating more useful and require less after the fact tweaks

## Motivation and Context
- prepping for the `mpa` branch and I want to get the package changes I've run into there out of the way of the documentation structure changes therein

## How Has This Been Tested?
Most of it IS tests or docs or tooling.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [x] I have signed the [Adobe Open Source CLA](http://opensource.adobe.com/cla.html).
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #892, opened 2020-09-15
fix(tabs): check if current selected value exists before setting selected attr

### LINKED ISSUE #877: sp-tabs gives a console error without a selected attribute
### Expected Behaviour
If the selected attribute isn't present, sp-tabs should just show nothing as selected

### Actual Behaviour
`Uncaught TypeError: Cannot read property 'length' of null`

#### Steps to Reproduce
Inspect one of the sp-tabs examples in the documentation, and remove the "selected" attribute using "Edit as HTML"

AndreasGalster:
When removing the selected attribute on tabs, a typeError was thrown since the passed value was null 

## Description

This is fixed by checking if the value exists before checking for length via optional chaining

## Related Issue

fixes #877 

## Motivation and Context

* This removes the typeError
* It seems tabs behavior worked despite the typeError since the if-condition throwing the error was only relevant for selection, not removal of selection

## How Has This Been Tested?

* This has been tested in storybook.
* No additional unit tests were created since there's no functional changes

## Types of changes

## Checklist:

### DIFF
--- a/packages/tabs/src/Tabs.ts
+++ b/packages/tabs/src/Tabs.ts
@@ -246,7 +246,7 @@ export class Tabs extends Focusable {
             element.removeAttribute('selected');
         });
 
-        if (value.length) {
+        if (value?.length) {
             const currentChecked = this.querySelector(`[value="${value}"]`);
 
             if (currentChecked) {


AndreasGalster:
@Westbrook I've just submitted my first fix. I just realized my mistake of not running tests, I thought they were running during storybook. It looks like all visual tests were failing, I'm not sure why that's happening.

It seems odd since that's related to any component. Am I missing something here that I need to do?

I can only test chromium on my local and all tests past for chromium except for one where an icon cannot be found. I tried running `yarn process-icons` in the hopes to fix the test but not luck.

Can you advise how to ensure I test correctly :)?

Westbrook:
@AndreasGalster thanks for the PR! Definitely a `good first issue`!

Right now, our visual regressions won't run correctly against branches in forks, and with the content of your PR I'm confident that the failures there are not due to your change. The ability to get more external contributions certainly raises the benefits of an update to those processes... they can be run locally before and after a change to confirm, but I don't believe you'd need to do so in this case.

Right now I'm trying to decide how I feel about the way the Typescript is unfurling this syntax.

```
if (value?.length) {
```

becoming

```
if (value === null || value === void 0 ? void 0 : value.length) {
```

I think this is a pretty common transpilation of the syntax, but I wonder if there is a different version we'd get with a different `target` or the update to `typescript@^4`. Naively, a usage of `if (value && value.length) {` would also be equivalent and wouldn't require unfurling. I'm going to double check what's possible there on the typescript side and report back.

While I do so, would you take a pass on what it might look like to unit test this update? I've not spent a lot of time with the error that was being thrown previously, but if we were able to catch it and fail before the change it would position us to not regress in this area in the future. And by future, I mean in PRs like #871 where we're also touching this functionality.

Westbrook (approved):
Looks like this is the expected transpilation and the when/if we move to es2020s in the lib/target entries we'll get the natural syntax again. I'm good to move this forward.

I added the `fixes #877` syntax to your description so merging this will automatically close the issue this fixes. Thanks!

Issue #764, opened 2020-07-15
Dropdown does not take focus when opens w/ iOS + VO


jnurthen:
### Expected Behaviour
opening the dropdown with iOS & VO focus should move to the dropdown entries.

### Actual Behaviour
Focus remains on the trigger when the dropdown opens. swiping right moves to the next element in the page not the dropdown entries

### Reproduce Scenario (including but not limited to)
https://opensource.adobe.com/spectrum-web-components/components/dropdown/examples

#### Steps to Reproduce
- https://opensource.adobe.com/spectrum-web-components/components/dropdown/examples
- Turn on VO
- focus onto the dropdown
- double tap to open the dropdown
on iPadOS 14 the dropdown opens but focus does not move to the choices in the dropdown.
on iOS 13 the dropdown does not even open.

#### Platform and Version
iPadOS14 - Safari - VO
iOS13 - Safari - VO

Westbrook:
I've confirmed this locally. We'll need to track down what event/interaction is being triggered by the "double-tap" interaction and ensure that it also throws focus for this and for other "replace" type overlays.

Westbrook:
Is this a timing situation?

Westbrook:
Spent a bunch of time this week reviewing the issues here, and learned a lot about the intricacies of this interaction.

Findings:
- role values + focus not enough to associate the UI elements together
- consume/apply IDs for each item
- focus the listbox not the individual items
- synthesize item focus with JS
- manage `aria-activedecendant` and `aria-selected`
- translate listbox click to click on "active" item

The changes cross a handful of elements, but should be well within reach for correcting this issue (which occurs across all VO contexts).

PR #893, opened 2020-09-21
refactor(shared): rewrite focusable to not use two layers of tabindex

### LINKED ISSUE #880: Do not sprout tabindex on sp- elements
### Expected Behaviour
sp- elements should not have tabindex (and indeed unless they should not be in the tree should not have any global aria attributes either)

### Actual Behaviour
These elements often have tabindex=0/-1 - this causes issues

These elements should generally not be focusable. Even if focus is being delegated to other child elements when they gain focus adding tabindex causes the node to appear in the accessibility tree. This causes VO to add extra groups all over the place

Run the button demo and note the extra groups being read with VO

#### Steps to Reproduce
* Use Safari
* Turn on VO (Command+F5)
* navigate to https://opensource.adobe.com/spectrum-web-components/components/button/examples
* use VO (CTRL+OPTION) and arrow right to move to the "Normal" button
* Note that it is necessary to navigate through the Normal Group first

#### Platform and Version
MacOS 10.15.6
Safari
VO

Note this issue impacts more than button - but button is the simplest to illustrate the problem.
Changes also need making to checkbox, radio, switch, textfield (amongst others)

Westbrook:
## Description
Remove the use of multiple `tabindex` attributes to manage the `tabIndex` of elements extending the `focusable` base class. Prevents screen reader content like "This is a button, button. This is a button, group." by not creating unneeded groups. This returns OS based tab order interactions (for the most part) to non-Chrome browsers on macOS.

You can preview this change at: https://5f68a7ca147fd400a9a88137--spectrum-web-components.netlify.app

## Related Issue

fixes #880

## Motivation and Context
Content should be accessible for everyone.

## How Has This Been Tested?
Existing unit tests.

## Types of changes
- [x] Refactor

## Checklist:


andrewhatran (approved):
I ran through the storybook/documentation site with `Voiceover` and tested out the tab/keyboard interactions on Chrome. The interactions look good 👍  

However I was having trouble getting the screen reader working correctly with `Menu` and `Dropdown`. For some reason, `Voiceover` didn't read out the menu items when I was navigating the menu with arrow keys.

I figured that was a separate issue out of scope for the PR. We can address this in the future

Westbrook:
@andrewhatran your attention to details always impresses me! Checkout https://github.com/adobe/spectrum-web-components/issues/764 where we're getting into that specific issue. Thanks for the review.

Issue #894, opened 2020-09-22
Components lose direction when removed/re-attached to DOM


adixon-adobe:
### Expected Behaviour
If you remove a component from the DOM and re-add, it should still have a direction attribute (appropriately inherited).

### Actual Behaviour
The direction attribute doesn't get re-added properly when the element is re-attached.

#### Steps to Reproduce
1. Go to the https://opensource.adobe.com/spectrum-web-components/components/checkbox
2. Inspect a checkbox and choose "store as global variable" in chrome.
3. Paste the following into the chrome console:
```js
parent = temp1.parentElement
parent.removeChild(temp1)
parent.appendChild(temp1)
```

Result: there's no longer any padding between the checkbox & label, and the direction attribute is no longer on the checkbox.

adixon-adobe:
Here's a screenshot: <img width="1003" alt="Screen Shot 2020-09-22 at 11 07 40 AM" src="https://user-images.githubusercontent.com/48841507/93908102-d7595680-fcc3-11ea-8fa9-7337641c7856.png">

Westbrook:
On first glance, this seems to work _in RTL_, so hopefully it's just a little bit of logic somewhere that's missing?

Westbrook:
It's right here:

https://github.com/adobe/spectrum-web-components/blob/5ebc7cf7dc50e8d8a95430eddb572d439d95f0e9/packages/base/src/Base.ts#L87

I remember specifically pointing to `this.dir` at the end, but I can't remember why... It won't be set if you get into this logic, so it'll apply `null`. We can move it back to `ltr` or to `this.dir || 'ltr'` and the issue should go away.

adixon-adobe:
I'd *guess* that you have `this.dir` for cases where there's an explicit override, but that you want `this.dir || 'ltr'` to cover the more common case where it's not.

Westbrook:
That's what I thought (looking at the code just now), but you can't get into this logic if `this.hasAttribute('dir')` have a value, which means that _always_ `this.dir` will be `null` 🙈

This may be a case of me _thinking_ "I'll do a smart thing here...", and it not actually being all the smart. Sadly all too common.

Westbrook:
fixed by #895

PR #895, opened 2020-09-22
fix: dir should never fall back to null

adixon-adobe:
## Description

Simple fix to the fallback for `dir`

## Related Issue

#894 

## Motivation and Context

For frameworks that re-use elements, this is pretty necessary!

## How Has This Been Tested?

Using the steps in the issue.

## Types of changes

## Checklist:

### DIFF
--- a/packages/base/src/Base.ts
+++ b/packages/base/src/Base.ts
@@ -84,7 +84,8 @@ export function SpectrumMixin<T extends Constructor<UpdatingElement>>(
                         ((dirParent as unknown) as ShadowRoot)
                             .host) as HTMLElement;
                 }
-                this.dir = dirParent.dir === 'rtl' ? dirParent.dir : this.dir;
+                this.dir =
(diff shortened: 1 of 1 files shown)


adixon-adobe:
@Westbrook one thing I noticed is that if I explicitly set `dir="rtl"` on the checkbox in the Chrome element debugger, that still gets lost when the element is re-added.

That seems like a separate issue to this, that might require a little more thought/refactoring.

Westbrook:
@adixon-adobe could you outlined what you see in:
> explicitly set dir="rtl" on the checkbox in the Chrome element debugger, that still gets lost when the element is re-added.

As an issue so we can review it separately? I'm not sure from this text whether it's expected or a bug.

I think this is looking good to me. Is the team looking for a pretty immediate release here?

adixon-adobe:
Yes -- I'd like to get this merged & released so we can get the Safari Canvas issue integrated.

Westbrook (approved):

Issue #292, opened 2019-11-15
Ensure web components follow accessibility standards


jnurthen:
### Expected Behaviour

Web Components should meet WCAG 2.1 AA https://www.w3.org/TR/WCAG21/
Behaviours of components should follow ARIA Authoring Practices https://www.w3.org/TR/wai-aria-practices/

### Actual Behaviour
Many components do not appear to have accessibility behaviours defined correctly. Spot checking 1 or 2 components I found many issues
For example
- [x] action-menu. Menuitems have button child elements. This is not allowed as it leads to an incorrect accessibility tree
- [ ] textfield and other input elements. I don't see any way to label a textfield
- [x] tab doesn't appear to have any roles defined. It also has no visible focus indicator and does not follow the ARIA authoring practices.

### Reproduce Scenario (including but not limited to)
View the above components on the documentation site.

#### Steps to Reproduce
Use Accessibility inspection tools to verify the behaviour of the components.

adixon-adobe:
@Westbrook did you ever get unittests with `axecore` working?

Westbrook:
They work...I've just not added them as some of the Spectrum settings fail. All new components get them by default in the templates added in #591 a la:
```
    it('loads default {{name}} accessibly', async () => {
        const el = await fixture<{{className name}}>(
            html`
                <{{> tagnamePartial}}></{{> tagnamePartial}}>
                `
            );
        
        await elementUpdated(el);

        expect(el).to.be.accessible();
    });
```
So using that pattern across other elements and taking responsibility for keeping it/removing it based on the results is a great idea!

Westbrook:
With the addition of #905 the only task explicitly outlined in this ticket is duplicated in #742.

Closing as duplicate.

PR #905, opened 2020-09-25
fix(dropdown): improve accessibility

### LINKED ISSUE #764: Dropdown does not take focus when opens w/ iOS + VO

### LINKED ISSUE #908: Menu shows focus border on click in safari
### Expected Behaviour
A menu doesn't show a blue focus border on click.

### Actual Behaviour
On safari it does. Easy to repro in the doc pages for Menu, Action Menu, & Dropdown

#### Steps to Reproduce
1. Go to https://opensource.adobe.com/spectrum-web-components/components/menu
2. Click on a menu

#### Screenshots
Screen Shot 2020-09-28 at 11 16 28 AM

Westbrook:
## Description 
Update the accessibility tree of the `Dropdown` pattern and its derivative components (`ActionMenu` and `SplitButton`) as well as the `Menu` pattern that they leverage internally.

Preview available: https://5f6e5921adb59e0095ccb7c2--spectrum-web-components.netlify.app/

*This does:*
- correct the focus target of open "dropdowns" to ensure that VO (macOS, iOS) enters/announces their content correctly
- moves from using `:focus` to `[focused]` to ensure transient visual state
- matches tests to these patterns
- updates to the latest aXe-core release
  - adjust usage of `role="group"` as per its updated rules
- populates `MenuItems` with `id`s when missing
- differentiates functionality between `option`s and `menuitem`s
- uses `div` elements over `button`s in `<sp-menu-item>`

*This does NOT:*
- use the `role=combobox` pattern
  - will follow up in the area when we have better information on making this x-browser compliant and a path to supporting the requirement of a `textfield` as a child of this role
- handle moving backward in the tab order in iOS and leaving the overlay
- applying size and count attributes

## Related Issue
fixes #764
refs #292
fixes #908

## Motivation and Context
Components should be accessible x-browser, x-context, etc.

## How Has This Been Tested?
- aXe-core unit testing
- manually review with macOS and iOS VO features
- visual regressions

image
image

## Types of changes

## Checklist:


packages/menu/src/Menu.ts
```diff
@@ -63,13 +63,11 @@ export class Menu extends SpectrumElement {
     }
 
     public focus(): void {
-        if (this.menuItems.length === 0) {
+        if (!this.menuItems.length) {
```
adixon-adobe, line 66: Why did you need this change -- is this just a code style choice, or is there something subtle I'm missing?

packages/menu/src/MenuItem.ts
```diff
@@ -83,6 +90,41 @@ export class MenuItem extends ActionButton {
...
+            : html`
+                  <div
+                      id="button"
+                      class="button"
+                      aria-label=${ifDefined(this.label)}
+                  >
+                      ${this.buttonContent}
+                  </div>
+              `;
+    }
```
adixon-adobe, line 109: Hadn't realized this extended from ActionButton -- neat. Whenever something like a core render method is overridden I try to think about maintainability. Since the only difference is the tag name, is that something you could parameterize?

adixon-adobe (commented):
A couple of questions that might lead to a few more changes. Overall looks great!

packages/menu/src/Menu.ts
Westbrook: Just style, I had removed this check at one point and then re-added it not knowing the style changed. There _may_ be some slight amount of type security outside of the TS space in the small fraction of time `this.menuItems === undefined`, but I don't know enough V8/timing to confirm.

packages/menu/src/MenuItem.ts
Westbrook: We could make the source render method more composable so that we only have to override part of it:
```
    return this.isAnchor ?  this.renderAnchor : this.renderButton;
}
protected get renderAnchor() {}
protected get renderButton() {}
```
That's probably nicer system wide, I'll take  swing at it.

Westbrook:
@adixon-adobe this has been updated. It's got a couple of settings and changes that it would be good to have in #912 to speed/stabilize the CI process while we lock down the docs/visual regression cache. Let me know if you see any other good suggestions herein!

adixon-adobe (approved):
Looks good -- thanks!

PR #907, opened 2020-09-26
fix: improve accessibility through aria-label attribute usage

### LINKED ISSUE #882: sp-bar-loader & sp-circle-loader need an accessible name
### Expected Behaviour
role=progressbar has accessible name required in the ARIA spec so it should specify an accessible name

### Actual Behaviour
It currently doesn't.
Easiest resolution would be to sprout aria-label on sp-bar-loader - it already has the role and the rest of the aria attributes

https://opensource.adobe.com/spectrum-web-components/components/bar-loader/examples

#### Steps to Reproduce
run and examine the code

Westbrook:
## Description
Ensure the correct visibility of `role=progress` elements in the accessibility tree through passing their `label` property into the `aria-label` attribute.

## Related Issue
fixes #882

## Motivation and Context
Components should be more accessible.

## How Has This Been Tested?
image
In https://5f6fca092785764a46d086a8--spectrum-web-components.netlify.app/

## Types of changes
- [x] Feature

## Checklist:

### DIFF
--- a/packages/bar-loader/src/BarLoader.ts
+++ b/packages/bar-loader/src/BarLoader.ts
@@ -91,5 +91,8 @@ export class BarLoader extends SpectrumElement {
         } else if (this.hasAttribute('aria-valuenow')) {
             this.removeAttribute('aria-valuenow');
         }
+        if (this.label && changes.has('label')) {
+            this.setAttribute('aria-label', this.label);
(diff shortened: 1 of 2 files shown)


andrewhatran (approved):
Looks good to me 🥇

PR #912, opened 2020-09-29
fix(action-menu): default to not taking a selection

### LINKED ISSUE #350: Support an "sp-action-menu" w/o value

```
                <sp-action-menu action-only>
                    <span slot="icon">More</span>
                    <sp-menu slot="options">
                            Flip Horizontally
                        </sp-menu-item>
                            Flip Vertically
                        </sp-menu-item>
                    </sp-menu>
                </sp-action-menu>
```

Westbrook:
## Description
Should likely go AFTER #905 where we are doing other accessibility and VO work on the Dropdown (base class) element that could effect this change. In particular, pre-905 we throw focus into the first/selected menu item when opening the menu, which causes a slight visual bug herein.

Add a `selectable` attribute to the `sp-action-menu` API so that by default the menu does not display a visual selection.

This doesn't alter the `change` event that triggers when the selection is made which will allow actions to be registered against individual `sp-menu-item` children or agains the `sp-action-menu` as a whole. Action on the `click` event of individual children would need to be managed with reverence to `await ActionMenu.updateComplete;` if they chose to throw focus, but I'm not sure if that should be directly documented or if preference should be given to using the `change` event here.

## Related Issue
fixes #350

## Motivation and Context
"Action" menus don't really have a selections.

## How Has This Been Tested?
Visually in storybook.
New unit tests

## Types of changes

## Checklist:

### DIFF
--- a/packages/action-menu/src/ActionMenu.ts
+++ b/packages/action-menu/src/ActionMenu.ts
@@ -31,7 +31,7 @@ export class ActionMenu extends ObserveSlotText(DropdownBase, 'label') {
     }
 
     @property({ type: Boolean, reflect: true })
-    public selected = false;
+    public selectable = false;
 
     @property({ type: Boolean, reflect: true })
     public quiet = true;
@@ -63,14 +63,29 @@ export class ActionMenu extends ObserveSlotText(DropdownBase, 'label') {
 
     protected updated(changedProperties: PropertyValues): void {
         super.updated(changedProperties);
-        if (changedProperties.has('open')) {
-            this.selected = this.open;
-        }
         if (changedProperties.has('quiet')) {
             this.quiet = true;
         }
         if (changedProperties.has('invalid')) {
             this.invalid = false;
+        }
+        if (changedProperties.has('open')) {
+            this.manageMenuItems();
+        }
+    }
+
+    private async manageMenuItems(): Promise<void> {
+        if (this.selectable || !this.optionsMenu) {
+            return;
+        }
+        if (this.optionsMenu.menuItems.length) {
+            this.optionsMenu.menuItems.forEach((el) => (el.selected = false));
+            return;
+        }
+        await this.optionsMenu.updateComplete;
+        /* istanbul ignore else */
+        if (this.optionsMenu.menuItems.length) {
+            this.manageMenuItems();
         }
     }
 }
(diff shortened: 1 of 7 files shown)


Westbrook:
@adixon-adobe let me know if you can think of some added documentation you'd want to see around this.

adixon-adobe (dismissed):
Looks good!

In terms of documentation, I could see the following being helpful:
* An example with `selectable` enabled.
* A note that the change event is fired regardless of whether it's selectable (I could see this being non-obvious)
* As long as you're there, and example with a custom icon & no label.

Westbrook:
Adding those docs and a couple of extra stories/visual regressions, too.

charlessuh:
Not something we need ASAP, but there are cases where `selectable` should be handled per-menu-item basis.

We have some menus where there's one section which contains items that should trigger an action (so shouldn't be selected) but there's another section with menu items that toggles state (so could be selected)

Westbrook:
@charlessuh, I'd personally wonder about the UX that sort of decision would actually offer a user. Is this a pattern that was adopted from some of the less public versions of the Spectrum spec? If not, we should bring it up with the design team and see if we can't get the pattern you're looking for here more standardized in the system.

charlessuh:
Just wondering if someone might use and depend on this `selectable` attribute, which might need to go away if we implement more flexible per-menu-item selectability?

This pattern is not uncommon in UX -- e.g. see the "View" menu in Finder.app, but I'll follow up on the Spectrum design side.

packages/action-menu/src/ActionMenu.ts
```diff
@@ -63,14 +63,29 @@ export class ActionMenu extends ObserveSlotText(DropdownBase, 'label') {
...
+            this.manageMenuItems();
+        }
+    }
+
+    private async manageMenuItems(): Promise<void> {
+        if (this.selectable || !this.optionsMenu) {
+            return;
+        }
+        if (this.optionsMenu.menuItems.length) {
+            this.optionsMenu.menuItems.forEach((el) => (el.selected = false));
```
adixon-adobe, line 82: Is this needed because `DropdownBase` causes this to get selected by default? To my comment on the issue, it might be nicer if the selected states were completely untouched if `selectable` isn't set.

Westbrook:
Superseded by the conversation in #350

PR #976, opened 2020-10-21
build: leverage eslint--plugin-lit-a11y

Westbrook:
## Description 
Use `eslint-plugin-lit-a11y`. Make changes to support its findings. I'll comment a bunch inline so we can discuss whether there are additional settings/considerations we'd like to take into account.

## Motivation and Context
The level of accessibility that our elements deliver should be both explainable and maintainable. 

## How Has This Been Tested?
It's lints, so it blocks the CI when things are missing.

## Types of changes

## Checklist:


documentation/src/components/side-nav.ts
```diff
@@ -86,11 +91,15 @@ class SideNav extends SpectrumElement {
...
+            <sp-underlay
+                class="scrim"
+                @click=${this.close}
+                ?open=${this.open}
+            ></sp-underlay>
             <aside>
                 <div id="nav-header">
                     <div id="logo-container">
-                        <a href="#">
+                        <a href="#index">
```
Westbrook, line 102: We've gone back and forth on whether this rule is required. `#` is technically a legit destination, but not _explicit_. Any insight on the accessibility of `#` vs `#index` or `#top` will be submitted back to the linter!

packages/.eslintrc.json
```diff
@@ -30,13 +31,51 @@
...
+                        "allowList": [
+                            "sp-button",
+                            "sp-action-button",
+                            "sp-checkbox",
+                            "sp-radio",
+                            "sp-switch",
+                            "sp-menu-item",
+                            "sp-clear-button",
+                            "sp-underlay",
+                            "sp-popover"
```
Westbrook, line 75: I know in this file this element has interactive children, but not in others. We have the ability to skip this rule in ALL custom elements, but seems like an over step...

packages/asset/stories/asset.stories.ts
```diff
@@ -23,7 +23,7 @@ export default {
 export const Default = (): TemplateResult => {
     return html`
         <sp-asset style="height: 128px">
-            <img src=${portrait} alt="Demo Image" />
+            <img src=${portrait} alt="Demo Graphic" />
```
Westbrook, line 26: Everywhere I switch `Image` to `Graphic` is in a demo or test, but it seems like good practice to get into.

packages/button/stories/button.stories.ts
```diff
@@ -196,7 +196,6 @@ export const withIcon = (): TemplateResult => {
                         viewBox="0 0 36 36"
                         focusable="false"
                         aria-hidden="true"
-                        role="img"
```
Westbrook, line 199: `role="img"` isn't really needed with `aria-hidden="true"`...

packages/overlay/src/OverlayTrigger.ts
```diff
@@ -72,6 +72,7 @@ export class OverlayTrigger extends LitElement {
             <div
                 id="trigger"
                 @click=${this.onTrigger}
+                @keypress=${this.onKeypress}
```
Westbrook, line 75: We can't guarantee that a user uses an interactive element below here, so I've added `keypress` with some gating, just in case. _Could_ be over kill...

documentation/src/components/side-nav.css
```diff
@@ -21,6 +21,12 @@ aside {
     overflow: auto;
 }
 
+.scrim {
+    display: none;
+    pointer-events: none;
+    opacity: 0;
+}
```
adixon-adobe, line 28: How do these css changes relate to the linting change?

packages/overlay/src/OverlayTrigger.ts
adixon-adobe: This does feel like overkill to me. I'd rather just see the documentation cover this.

In general I think we should document and have examples of best practices, and not try to "fix" bad practices.

adixon-adobe (changes requested):
One change request and one question. Overall looks good!

documentation/src/components/side-nav.css
Westbrook: Indeed. I should be able to take even more of this away as we're moving from a `<div>` element here that would require special rule skipping or keyboard event handling in exchange for an `<sp-underlay>` element that already has agreement with the lint rules. Will fix!

packages/overlay/src/OverlayTrigger.ts
Westbrook: This makes sense to me. I'll work on the right way to skip this rule here and add a sentence or two to the documentation for this.

packages/overlay/README.md
```diff
@@ -71,3 +71,7 @@ import {
     </sp-tooltip>
 </overlay-trigger>
 ```
+
+## Accessibility
+
+When using an `<overlay-trigger>` element, it is important to be sure the that content you project into `slot="trigger"` is "interactive". This means that an element within that branch of DOM will be able to receive focus and said element will appropriately convert keyboard interactions to `click` events similar to what you find with `<a href="#">Anchors</a>`, `<button>Buttons</button>`, etc. You can find further reading on the subject of accessible keyboard interactions at [https://www.w3.org/WAI/WCAG21/Understanding/keyboard](https://www.w3.org/WAI/WCAG21/Understanding/keyboard).
```
adixon-adobe, line 77: Nice!

adixon-adobe (approved):
Updates look great. Nice!

Issue #137, opened 2019-08-15
Techometer reporting


Westbrook:
### Expected Behaviour
Tachometer scoring is applied somewhere and covers a decent amount of an elements standard use cases?

### Actual Behaviour
You have to run it manually.

### Reproduce Scenario (including but not limited to)
`npm run test:bench`

### Notes:
Be sure to check on things like https://github.com/adobe/spectrum-web-components/pull/103#discussion_r314466082 when deciding how to cover use cases.

Westbrook:
Looks like the change to have `lit-html` be a `dependency` rather than a `devDependency` a la https://github.com/adobe/spectrum-web-components/pull/683/files#diff-62e26ea604f25c999ecde6acf4720099R48 would allow us to leverage the `--package-version` flag from tachometer to compare `@current` to a `local` build of the packages, which would allow us to leverage this scoring in CI and possibly as a PR gate. Another great win for #679.

Westbrook:
With the addition of #978 we'll be able to efficiently test a local build against what's on the `latest` channel on NPM. Steps from there to see this completed:
- test performance changes in CI (possibly filtering for changes similar to what the Lerna versioning code does)
- report results into PRs
- possibly block PRs with regressions behind an approval step

PR #978, opened 2020-10-22
test: compare local and latest builds of a component

Westbrook:
## Description 
Update to the latest version of `tachometer` so that we can easily compare local builds to the most recent version available on NPM.
- ensure all benchmarks actually run in the environment
- prep CLI to properly configure the use of `tachometer`
- add/expand some test
- bump test subjects from 10 => 100 for more clear differences in perf scores

## Related Issue
refs #137 

## Motivation and Context
We want to start tweaking the internals of our elements for performance and feature completeness and it is useful to have baseline comparisons to know that we're moving in a good or at least acceptable direction.

## How Has This Been Tested?
Manually on local against some future looking performance changes.

image
Current output where `remote` is the performance of the code available on the `latest` channel on NPM and `<none>` is that of the code available locally.

## Types of changes
- [x] Tooling

## Checklist:

### DIFF
--- a/test/benchmark/cli.ts
+++ b/test/benchmark/cli.ts
@@ -1,5 +1,17 @@
+/*
+Copyright 2020 Adobe. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License. You may obtain a copy
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
 import { main } from 'tachometer/lib/cli';
-import { readdirSync } from 'fs';
+import { ConfigFile } from 'tachometer/lib/configfile';
+import { readdirSync, writeFileSync } from 'fs';
 import { join as pathjoin } from 'path';
 import * as commandLineArgs from 'command-line-args';
 import * as commandLineUsage from 'command-line-usage';
@@ -40,6 +52,13 @@ const optionDefinitions: commandLineUsage.OptionDefinition[] = [
         defaultValue: 50,
     },
     {
+        name: 'timeout',
+        description:
+            'The maximum number of minutes to spend auto-sampling.\n(default 3)',
+        alias: 't',
+        type: Number,
+    },
+    {
         name: 'browser',
         description:
             'Which browsers to launch in automatic mode.' +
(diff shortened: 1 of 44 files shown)


davidyxu (commented):
seems good to me, although I wasn't able to complete the benchmark on field-group, not sure if thats related to the updates here:

`yarn test:bench -p field-group`

gives me this message:
```
Failed 1/3 times to get measurement(s) field-group:basic-test in chrome from http://127.0.0.1:8081/test/benchmark/bench-runner.html?bench=basic-test&package=field-group. Retrying.
```

in the browser i see this error:
```
Uncaught (in promise) TypeError: Failed to resolve module specifier "@spectrum-web-components/field-group/sp-field-group.js". Relative references must start with either "/", "./", or "../".
```

Westbrook:
@davidyxu Thanks for diving so deep on this! 🙏 `@spectrum-web-components/field-group` has yet to be released to NPM, so it won't be able to be tested in this way. 😓 It is useful that you brought it us as it _was_ generated from files that didn't have the correct pathing for `test/benchmark/helpers.js`, so I've corrected that here and in the generator so we shouldn't have that issue in the future.

In collaboration with Alex from the LRWeb team, we're gathering some changes together to try to reduce the broader effects of a couple of breaking changes from Spectrum CSS before our next round of releases. That should hopefully come in the next week or two 🤞 and then `@spectrum-web-components/field-group` et al will be available for testing in this way.

test/benchmark/cli.ts
```diff
@@ -132,18 +148,53 @@ $ node test/benchmark/cli -n 20
             .map((dirEntry) => dirEntry.name.replace(/\.js$/, ''));
 
         for (const benchmark of benchmarks) {
-            runCommands.push(
-                `${packageName}:${benchmark}=test/benchmark/bench-runner.html` +
-                    `?bench=${benchmark}` +
-                    `&package=${packageName}`
-            );
+            if (!config.benchmarks) return;
+            config.benchmarks.push({
```
davidyxu, line 152: In regards to the unpublished npm package issue I ran into with field-group, it would be a nice-to-have in the future to skip the remote test if its not available.

a couple options might be to either to use some standardized proess with the package version to determine if its been published or to actually check for latest in npm (although this might be a bit more heavy)
```
yarn info @spectrum-web-components/field-group@latest
```

speaking of which, it might also make sense to limit the benchmark tests to the same major version instead of against latest, in case theres been an interface change.

davidyxu (approved):
left some thoughts on the field-group stuff, but otherwise lgtm.

test/benchmark/cli.ts
Westbrook: Great idea! I've added the following message when a package is still on the "default" (read here as "unpublished) version:
```
⚠️  It looks like '${packageName}' has yet to be published to NPM. Skipping!
```
I've also added the `-c/--compare` flag which allows for the use of a specific version of `@spectrum-web-components/bundle` to be used as the "remote" source against which to test. This is done so that packages that require child components both public or private are able to rely of releases of all the comparable version at a time.

Issue #815, opened 2020-08-06
sp-menu-item renders too slow for large menus


adixon-adobe:
### Expected Behaviour
A dropdown with 200 menu items should load quickly.

### Actual Behaviour
Each menu item takes 1-5ms to render (in Chrome on my MacbookPro). It takes ~400ms to render on my machine.

#### Steps to Reproduce
Create a large dropdown (e.g. a country selector) & profile the page.

adixon-adobe:
I actually have a more abusive case on our test page (a checkbox that switches 20 dropdown to all have 100 items). It takes 4 seconds for the checkbox to check. I was hoping it was more of an issue with over-updating, but following the trace down it leads to a whole lot of this:

<img width="458" alt="lots-of-ms" src="https://user-images.githubusercontent.com/48841507/89552337-a08db680-d7d1-11ea-9d45-5c92728395c8.png">

Westbrook:
I'd been tracking a similar(ish) context in the `<sp-sidenav-item>` elements, both of which handle some context resolutions synchronously in their `connectedCallback`, a la: 

https://github.com/adobe/spectrum-web-components/blob/4af755e7111da12b9ad97caef92ba7713336c470/packages/sidenav/src/SidenavItem.ts#L155-L165

Couple of options that I'd been thinking about are as follows:
- make first render lazy
  - had approached this as directly hacking the initialize functionality in `LitElement`, which I'm not super excited about, but is a possibility
- lazily resolve the value in the event
  - requires a "double render" but shouldn't be an issue in real time with `lit-html` backing up the render
- move to a `MutationOverserver` for top down, instead of bottom up, management of attributes like this
  - a deeper refactor, but likely the most powerful. not sure if mutuation observers works async by default or not. likely still requires the "double render" context

Westbrook:
@adixon-adobe you wouldn't happen to have that `more abusive case` in a code sample that you could easily share, would ya?

adixon-adobe:
Not in an easy to translate way, no, but I think if you use a country list selector sourced from one of these lists you'll be able to test some things out pretty easily: https://gist.github.com/incredimike/1469814

adixon-adobe:
(also, my test page switches over ~16 dropdowns to have 100 items all at once (named `Item 1`, `Item 2`, etc).

Westbrook:
[...]

PR #980, opened 2020-10-26
Various updates to performance based off of better `tachometer` results.

Westbrook:
## Description
- clean up duplicate code in `<sp-sidenav>`
- simplify logic in `focusable`
- make application of the `focus-visble` polyfill lazy. This means that "autofocus" functionality is delayed by a frame, see update to timing in tests.

## Related Issue
refs #815 

## Motivation and Context
Lazy polyfilling leads to perf trends like the following (note: `Remote` is the currently published version):

*Accordion*
image

*ActionMenu*
image

*Button*
image

*Link*
image

*Menu*
image

*Radio*
image

*Checkbox*
image

*Search*
image

*Sidenav*
image

*Slider*
image

*Switch*
image

*Textfield*
image

*Dropdown*
image

## How Has This Been Tested?
See above. See asynchrony changes to unit tests.

## Types of changes

## Checklist:

### DIFF
--- a/packages/checkbox/test/checkbox.test.ts
+++ b/packages/checkbox/test/checkbox.test.ts
@@ -18,6 +18,7 @@ import {
     triggerBlurFor,
     html,
     expect,
+    waitUntil,
 } from '@open-wc/testing';
 import { waitForPredicate } from '../../../test/testing-helpers.js';
 import '@spectrum-web-components/shared/src/focus-visible.js';
@@ -120,7 +121,10 @@ describe('Checkbox', () => {
         ) as Checkbox;
 
         expect(autoElement).to.exist;
-        expect(document.activeElement).to.equal(autoElement);
+        await waitUntil(
+            () => document.activeElement === autoElement,
+            'Autofocused'
+        );
 
         await triggerBlurFor(autoElement);
 
--- a/packages/shared/src/focus-visible.ts
+++ b/packages/shared/src/focus-visible.ts
@@ -114,20 +114,26 @@ export const FocusVisiblePolyfillMixin = <
         // document:
         connectedCallback(): void {
             super.connectedCallback && super.connectedCallback();
-            if (this[$endPolyfillCoordination] == null) {
-                this[$endPolyfillCoordination] = coordinateWithPolyfill(this);
-            }
+            requestAnimationFrame(() => {
+                if (this[$endPolyfillCoordination] == null) {
+                    this[$endPolyfillCoordination] = coordinateWithPolyfill(
+                        this
+                    );
+                }
+            });
         }
 
         disconnectedCallback(): void {
             super.disconnectedCallback && super.disconnectedCallback();
             // It's important to remove the polyfill event listener when we
             // disconnect, otherwise we will leak the whole element via window:
-            if (this[$endPolyfillCoordination] != null) {
-                // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-                this[$endPolyfillCoordination]!();
-                this[$endPolyfillCoordination] = null;
-            }
+            requestAnimationFrame(() => {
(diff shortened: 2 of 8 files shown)


packages/checkbox/test/checkbox.test.ts
```diff
@@ -120,7 +121,10 @@ describe('Checkbox', () => {
         ) as Checkbox;
 
         expect(autoElement).to.exist;
-        expect(document.activeElement).to.equal(autoElement);
+        await waitUntil(
+            () => document.activeElement === autoElement,
+            'Autofocused'
+        );
```
Westbrook, line 127: Support the move to lazy focusability by waiting for the focus to be applied rather than expecting it synchronously.

packages/shared/src/focusable.ts
```diff
@@ -135,12 +135,14 @@ export class Focusable extends FocusVisiblePolyfillMixin(SpectrumElement) {
...
-            );
-            this.focusElement.focus();
+            requestAnimationFrame(() => {
+                this.dispatchEvent(
+                    new KeyboardEvent('keydown', {
+                        code: 'Tab',
+                    })
+                );
+                this.focusElement.focus();
+            });
```
Westbrook, line 145: Wait a frame so that you can know that the polyfill has been applied.

packages/sidenav/test/sidenav.test.ts
```diff
@@ -222,66 +196,41 @@ describe('Sidenav', () => {
         await elementUpdated(el);
 
         expect(el.value).to.equal('Section 0');
-        expect(toBeSelected.tabIndex, '-1 when click based focus').to.equal(-1);
+        expect(toBeSelected.tabIndex, 'will be new focusable child').to.equal(
```
Westbrook, line 199: I'm honestly not sure how these tests passed previously...like at all.

packages/shared/src/focus-visible.ts
```diff
@@ -114,20 +114,26 @@ export const FocusVisiblePolyfillMixin = <
...
         connectedCallback(): void {
             super.connectedCallback && super.connectedCallback();
-            if (this[$endPolyfillCoordination] == null) {
-                this[$endPolyfillCoordination] = coordinateWithPolyfill(this);
-            }
+            requestAnimationFrame(() => {
+                if (this[$endPolyfillCoordination] == null) {
+                    this[$endPolyfillCoordination] = coordinateWithPolyfill(
+                        this
+                    );
```
adixon-adobe, line 121: guessing you'll want to collapse 120/121 back into line 119.

packages/shared/src/focusable.ts
adixon-adobe: I find it's better to be explicit about what your waiting for in cases like this (e.g. say for some reason you need to chain an extra animation frame for the polyfill because of something else -- that'd break this code).

At the very least I'd add a comment

packages/sidenav/test/sidenav.test.ts
adixon-adobe: might be worth digging into a touch more on `main` so see if you can get them to fail.

adixon-adobe (changes requested):
I see a couple of opportunities for improvement.

packages/shared/src/focus-visible.ts
Westbrook: I don't do that... Prettier owns all the new lines 🤷‍♂️  it gets tabbed over one more time in #984, so not sure it's worth messing with?

packages/shared/src/focusable.ts
Westbrook: Great point! Found a way not to have the polyfill reach into the library code like this. Let me know what you think.

adixon-adobe (approved):
Looks good!

PR #984, opened 2020-10-26
feat(shared): conditionally apply focus-visible polyfill

Westbrook:
## Description 
Conditionally apply the `focus-visible` polyfill only where it is needed:
image
While it would be interesting to also support `:-moz-focusring`, I'm not sure it 100% has the same heuristics as `:focus-visible`, so it's not included herein.

This approach builds on #980 where we make the polyfill lazy. While the results below tend to show a small amount of performance win, there seems to be less than I expected. Due to this, it may be too early to take on this change as it is generally powered by the a change in the way we process/deliver CSS for packages.

Previously, the following CSS `a:focus-visible { ... }` would be converted directly to `a.focus-visible { ... }` keeping the support process non-negative. To support both polyfill and no polyfill delivery it is required that the previous sample be processed to (all automatic of course, no developer intervention needed):
```
a:focus-visible { ... }
a.focus-visible { ... }
```
As CSS will throw out all selectors in a rule wherein a single selector is erroneous. This means that there are some small file size alterations that come along with this change to support. 

I'd love to hear your thoughts!

## Motivation and Context
I wanna go faster!

## How Has This Been Tested?

*Accordion* vs 1-6% in #980 
image

*ActionMenu* vs 35-37% in #980 
image

*Button* vs 7-9% in #980 
image

*Link* vs 9-11% in #980 
image

*Menu* vs 21-23% in #980 
image

*Radio* vs 15-16% in #980 
image

*Checkbox* vs 4-8% in #980 
image

*Search* vs 3-5% in #980 
image

*Sidenav* vs 82-86% in #980 
image

*Slider* vs 4-6% inn #980 
image

*Switch* vs 6-9% in #980 
image

*Textfield* vs 3-5% in #980 
image

*Dropdown* vs 33-36% in #980
image

## Types of changes

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
         "build": "gulp css && tsc --build packages/**/tsconfig.json",
         "build:watch": "tsc --build packages/**/tsconfig.json -w",
         "build:tests": "tsc --build test/tsconfig.json && tsc --build test/tsconfig-node.json",
-        "build:clear-cache": "rm -rf packages/*/lib && rm -rf packages/*/tsconfig.tsbuildinfo",
+        "build:clear-cache": "rimraf packages/*/lib && rimraf packages/*/tsconfig.tsbuildinfo",
         "watch": "gulp watch",
         "prestorybook": "wca analyze 'packages/*/*.d.ts' --format json --outFile custom-elements.json && node ./scripts/add-custom-properties.js --src='custom-elements.json'",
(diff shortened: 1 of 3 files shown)


adixon-adobe (approved):
This makes sense to me, and I like keeping up with browser support. Too bad it's not as big a win as you hoped though

PR #996, opened 2020-11-03
fix(overlay): use esm build from popper and point through to types

Westbrook:
## Description
Move to the ESM build from Popper so as to remove the need to `process.env` workarounds in no build contexts.

I wish that https://github.com/popperjs/popper-core/issues/1188 and https://github.com/popperjs/popper-core/issues/1188 and https://github.com/popperjs/popper-core/issues/933 would have gotten us a better solution. Needing to use an off main entry point build while hand managing the connectivity to the types is not my favorite thing in the world.

## Related Issue
refs 963

## Motivation and Context
`process.env` should be removed from production facing code.

## How Has This Been Tested?
Works locally and in CI

## Types of changes

## Checklist:

### DIFF
--- a/packages/overlay/src/popper.ts
+++ b/packages/overlay/src/popper.ts
@@ -13,17 +13,19 @@ governing permissions and limitations under the License.
 // Bundle only what we want from popper
 // See: https://popper.js.org/docs#popper-lite-tree-shaking
 
+// eslint-disable-next-line @typescript-eslint/triple-slash-reference
+/// <reference path="../local.d.ts" />
 import {
     popperGenerator,
     defaultModifiers,
-} from '@popperjs/core/lib/popper-lite';
-import type { Instance } from '@popperjs/core/lib/types';
-import type { Placement } from '@popperjs/core/lib/enums';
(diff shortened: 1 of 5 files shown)


packages/overlay/src/popper.ts
```diff
@@ -13,17 +13,19 @@ governing permissions and limitations under the License.
 // Bundle only what we want from popper
 // See: https://popper.js.org/docs#popper-lite-tree-shaking
 
+// eslint-disable-next-line @typescript-eslint/triple-slash-reference
+/// <reference path="../local.d.ts" />
```
Westbrook, line 17: This is the weirdest part. Even though I could get this to connect via `yarn tsc --build packages/overlay/tsconfig.json` without this reference, I could never get `yarn tsc --build packages/*/tsconfig.json`, our standard usage, to work without this... 😞

cuberoot (approved):
Looks great. Thanks for this. I tried to use the esm build, but my typescript-fu was not as strong as yours. I wasn't able to get the types to work. It's good to get rid of the `Env.debug` hack.

Issue #970, opened 2020-10-20
Vend a Storybook theme helper


Westbrook:
### Expected Behaviour
Can easily theme Stroybook demonstrations of SWC based features.

### Actual Behaviour
You can, but it takes some deeper knowledge/work than expected.

- export controls for `color`, `scale`, and `dir`
- export Storybook decorator based theme application
- add default styling for easy use
- structure as composable pieces/methods for reasonable customization

Westbrook:
Initial beta release available via `@spectrum-web-components/story-decorator@beta`. I've looking at using it to build out a family of component starters at https://webcomponents.dev/org/adobe

Westbrook:
Released

PR #999, opened 2020-11-07
chore: update to @web/dev-server-storybook

LarsDenBakker:
This updates the storybook to use the new dev server storybook plugin: https://modern-web.dev/docs/dev-server/plugins/storybook/ which uses storybook v6.

I used spectrum to see if it works in a real world project. I'm PRing my work so far, feel free to use it or discard it.

This is not yet a finished product, I only update some of the stories. Serve and build commands are working however, and so is the theming and new story args.


Westbrook:
Thanks for this Lars. Great to see that there's a path forming on the way to getting Storybook up to v6. Very cool capabilities therein, and getting up to it as soon as possible will be good for the project (and for Modern Web). I'm gonna take this branch and see where I can get it to. Beyond the simple step of getting all of the stories updated, there are a couple of other concepts to be worked out appropriately.

1. When using knobs we were able to cache theme values in module scope and apply them across multiple stories, meaning we'd keep a set themes across all stories. I followed up on this base line to see if I could see any other way to cache values, and while were are a couple of options, none of them really get the experience 100%. Namely, I could get the values to cache in the story, but not "display" as set in the controls UI, which was annoying.

2. We'd previously used "Knobs Groups" to separate these theme values from the story customization in the Storybook UI which is not currently available for controls. It looks like they'll be working on that at some point based on https://github.com/storybookjs/storybook/issues/10979, but not yet.

3. Similarly, we'd used another feature of knobs that allowed you to create a deep linking URL to a specific set of knobs values as the basis of our visual regression tooling. Being able to use our currently available stories and jump straight to the iframe version of the story with specific theme values applied make testing across the 16 different theme variations of each story much easier. It looks like it may be coming in the near future as well: https://github.com/storybookjs/storybook/issues/12291

I _think_ much of this can be addressed in conjunction with the work going into #970 to make leveraging Spectrum theming in external Storybook instances a drop-in decorator. It's currently using a working approach like so:
image
And would allow for "grouping" the theme properties in a custom element that then "managed" them across stories and could accept them from URL vars. I been working on it a little in advance on leveraging it in a webcomponents.dev library of our elements, so it might be good to use it everywhere...

Adding additional UI to the "story" will invalidate all of our golden images, but not such a large problem knowing that's the only change. Alternatively, this may make room for attempting to leverage other tools for visual regression testing so we're not bound to Storybook for that. Beyond making managing the regression cases easier (no array of strings in a special file) the pages will likely be much faster if they are in no way associated with the Storybook build.

With all that being said, I will be closing the PR for now. Thanks again for all your hard work in getting this to a workable baseline. Hope to have more news on being able to leverage this in the project before too long.

PR #1015, opened 2020-11-19
fix: stop merging selectors in a way that alters the cascade

Westbrook:
## Description 
It turns out that `yarn process-spectrum` has been collapsing matching selectors regardless of their relation to each other in the cascade. Here I've turned that off to support the use of `<sp-action-button quiet emphasized>` and it also allows for `<sp-checkbox disabled checked>` and `<sp-checkbox invalid indeterminate>` to display correctly without a manual work around.

This is a super noisy change, thought there are really only TWO operational alterations and I will point them out directly.

## Motivation and Context
The quiet emphasized action button wasn't displaying appropriately in application code.

## How Has This Been Tested?
Additional visual regressions to maintain this functionality going forward.

image
image
image

## Types of changes

## Checklist:

### DIFF
--- a/packages/checkbox/src/checkbox.css
+++ b/packages/checkbox/src/checkbox.css
@@ -20,30 +20,3 @@ governing permissions and limitations under the License.
 :host(:empty) label {
     display: none;
 }
-
-/* Work around specificity changes in the selector processing from Spectrum CSS */
-/* stylelint-disable no-descending-specificity */
-:host([disabled][checked]) #input + #box:before,
-:host([disabled][checked]) #input:checked + #box:before {
-    /* .spectrum-Checkbox .spectrum-Checkbox-input:checked:disabled+.spectrum-Checkbox-box:before,
-   * .spectrum-Checkbox .spectrum-Checkbox-input:disabled+.spectrum-Checkbox-box:before */
-    border-color: var(
-        --spectrum-checkbox-emphasized-box-border-color-disabled,
-        var(--spectrum-global-color-gray-400)
-    );
-    background-color: var(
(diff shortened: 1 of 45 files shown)


scripts/process-spectrum-postcss-plugin.js
```diff
@@ -590,17 +590,8 @@ class SpectrumProcessor {
...
-            parentRule = rule;
-            return false;
-        });
-
-        if (!parentRule) {
-            parentRule = postcss.rule({ selectors });
-            this.result.root.append(parentRule);
-        }
+        const parentRule = postcss.rule({ selectors });
+        this.result.root.append(parentRule);
```
Westbrook, line 594: And, the real guts...

packages/checkbox/src/checkbox.css
```diff
@@ -20,30 +20,3 @@ governing permissions and limitations under the License.
...
-:host([invalid][indeterminate]) #box:before,
-:host([invalid][indeterminate]) #input:checked + #box:before {
-    /* .spectrum-Checkbox.is-invalid .spectrum-Checkbox-box:before,
-   * .spectrum-Checkbox.is-invalid .spectrum-Checkbox-input:checked+.spectrum-Checkbox-box:before */
-    border-color: var(
-        --spectrum-checkbox-box-border-color-error,
-        var(--spectrum-global-color-red-500)
-    );
-}
-/* stylelint-enable no-descending-specificity */
```
cuberoot, line 49: Yay for pink.

cuberoot (approved):
This looks good. With CSS, order matters, so my consolidation of CSS blocks was bound to cause issues in something as complicated as `spectrum-css`. So, this is the right way to go.

PR #1018, opened 2020-11-22
fix(menu): prevent infinite loop when focus()

### LINKED ISSUE #1009: Focusing a sp-menu with no enabled sp-menu-item causes infinate loop
### Expected Behaviour
A sp-menu with no enabled sp-menu-item can be focused normally, or can't obtain a focus.

### Actual Behaviour
Call focus on such a sp-menu causes infinate loop, makes the browser stop responding.
For example, a dropdown will try to focus its menu when opened.
```html
<sp-dropdown label="Select a Country with a very long label, too long in fact">
    <sp-menu>
        <sp-menu-item disabled>
            Deselect
        </sp-menu-item>
        <sp-menu-item disabled>
            Select inverse
        </sp-menu-item>
        <sp-menu-item disabled>
            Feather...
        </sp-menu-item>
        <sp-menu-item disabled>
            Select and mask...
        </sp-menu-item>
        <sp-menu-divider></sp-menu-divider>
        <sp-menu-item disabled>
            Save selection
        </sp-menu-item>
        <sp-menu-item disabled>
            Make work path
        </sp-menu-item>
    </sp-menu>
</sp-dropdown>
```
#### Platform and Version
Latest Edge & Firefox
@spectrum-web-components/bundle:  0.15.5

Westbrook:
## Description
Make sure there are active menu items available when focusing items within an `<sp-menu>` element.

## Related Issue
fixes #1009 

## Motivation and Context
Prevent infinite loops.

## How Has This Been Tested?
Unit tests.

## Types of changes

## Checklist:

### DIFF
--- a/packages/menu/src/Menu.ts
+++ b/packages/menu/src/Menu.ts
@@ -63,7 +63,10 @@ export class Menu extends SpectrumElement {
     }
 
     public focus(): void {
-        if (!this.menuItems.length) {
+        if (
+            !this.menuItems.length ||
+            this.menuItems.every((item) => item.disabled)
+        ) {
             return;
         }
         this.focusMenuItemByOffset(0);
--- a/packages/menu/test/menu.test.ts
+++ b/packages/menu/test/menu.test.ts
@@ -59,6 +59,24 @@ describe('Menu', () => {
             .false;
         expect(document.activeElement === anchor, 'anchor').to.be.true;
     });
+    it('renders w/ [disabled] menu items', async () => {
+        const el = await fixture<Menu>(
+            html`
+                <sp-menu tabindex="0">
+                    <sp-menu-item disabled>Disabled item</sp-menu-item>
+                </sp-menu>
+            `
+        );
+
+        await elementUpdated(el);
+        expect(document.activeElement === el, 'self not focused, 1').to.be
+            .false;
+
+        el.focus();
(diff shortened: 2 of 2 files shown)


packages/menu/src/Menu.ts
```diff
@@ -63,7 +63,10 @@ export class Menu extends SpectrumElement {
     }
 
     public focus(): void {
-        if (!this.menuItems.length) {
+        if (
+            !this.menuItems.length ||
+            !this.menuItems.filter((item) => !item.disabled).length
```
kaiortmanns, line 68: You maybe could use menuItems.every(..) here to make it more obvious what you are testing.

kaiortmanns (approved):

packages/menu/src/Menu.ts
Westbrook: Good idea. Is `!this.menuItem.some((item) => !item.disabled)` similarly clear, or no? It has the possibility of not having to traverse the entire array.

kaiortmanns: Well. `every` returns a boolean and you don´t have to test for length again. And `every` would break immediately if the condition is not met. So you don´t iterate over the whole array if you don´t need to.
So `this.menuItem.every((item) => item.disabled)` would test if all items are disabled.

PR #1051, opened 2020-12-09
Latest CSS Beta

Westbrook:
## Description
- add `@spectrum-web-components/action-button` package
- update to latest Spectrum CSS betas
- work around issues in Checkbox, Radio, Switch
- updates docs and Storybook with changes

## Motivation and Context
Keeping up with Spectrum

## How Has This Been Tested?
Thorough review via Visual Regression Tests
See it for your self at: https://5ff3b2bb4e4c5a121015d63f--spectrum-web-components.netlify.app

## Types of changes
- [x] Keeping up with deps

## Checklist:
- [x] I have signed the [Adobe Open Source CLA](http://opensource.adobe.com/cla.html).
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #1054, opened 2020-12-14
feat: use fully registered icon components

Westbrook:
## Description
Publish a registered component for each of the Spectrum Workflow Icons so that they can be used directly in HTML.
- prepares for the culling of the icons that occurs in the current Spectrum CSS beta where a number of the UI icons are removed
- shows how the icons can be used by updating the documentation and the tests
- reduces the manual use of `size`
- opens the door for deprecating the `iconset` package all together
- separates _individual_ icons from `iconset` which reduces the work needed to finalize the render

If this feels good for people, we could apply this approach to the remaining UI icons as well.

Thoughts?

## Related Issue
refs #155 
refs #669

## Motivation and Context
Make it easier to use Spectrum Icons regardless of your deployment context.

## How Has This Been Tested?
- visual regressions mainly

## Types of changes

## Checklist:

### DIFF
--- a/test/benchmark/cli.ts
+++ b/test/benchmark/cli.ts
@@ -76,6 +76,15 @@ const optionDefinitions: commandLineUsage.OptionDefinition[] = [
         type: String,
         defaultValue: 'latest',
     },
+    {
+        name: 'start',
+        description:
+            'Test from "page" start (includes element registration and upgrade) or "element" start (only includes upgrade).' +
+            '\n(default element)',
+        alias: 's',
+        type: String,
+        defaultValue: 'element',
+    },
 ];
 
 interface Options {
@@ -86,6 +95,7 @@ interface Options {
     timeout: string;
     browser: string;
     compare: string;
+    start: string;
 }
 
 (async () => {
@@ -131,6 +141,7 @@ $ node test/benchmark/cli -n 20
             .filter((dirEntry) => dirEntry.isDirectory())
             .map((dirEntry) => dirEntry.name);
     }
+    const start = opts.start;
 
     const printResults: string[] = [];
     for (const packageName of packages) {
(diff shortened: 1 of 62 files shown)


adixon-adobe (commented):
Are there any issues in browsers with how many web components you register? That's one thing that would give me pause about this approach. I'll want @charlessuh to weigh in as well, since he's dug deeper into the integration here recently.

Westbrook:
@adixon-adobe that combined with ease of use are my concerns as well.

Do have any thoughts on what would be a good test case to assuage this? I look at the "real world" case of our docs site, both the diff: https://github.com/adobe/spectrum-web-components/pull/1054/files#diff-7f9e7e0373bcb839ff499c8bad676435c24cc71288978b607bd89afc990546c7R17-R26 and the perf/experience: https://5fd792d842401d0b584f3bf0--spectrum-web-components.netlify.app/ 

Generally, the situation here is bound directly to usage. If you use ALL of the icons (I can't say where the boundary would lie), then you'll have trouble. First, you'll have the same trouble you would have had in any other approach of there being a lot of icons leaving you with a large about of JS KBs. Second, where you actually to have ALL of the icons on the page at once, the registration/upgrade process is a blocking synchronous activity, you'd have additional registrations, but you'd have to excessive upgrades either way. Most other situations of perf that I can think of would be roughly a wash for things you give and take between the various techniques, but having a test case to lock that down definitively one way or another does sound like a good goal here.

adixon-adobe:
I could see us at a point where we have 50 distinct icons in an initial page view (I can get about 30 right now). I'm sure some tool heavy apps could end up with over 100, so maybe a test case with 200 icons?

The part where the registrations are blocking/synchronous sounds concerning if that's taking much time at all if the other solutions aren't.

Westbrook:
In the following tachometer report we compare the application of 35 icons (because that's how many UI icons are available to fairly compare) to a page via the `attribute`, the `injected` and the `registered` patterns. Each pattern submits tests of a single icons 35 times as well as 35 icons once. Tachometer then runs the test HTML 100 times on the same page and include the registration of all elements used in the time reported..

*attribute*
```
<sp-icon name="ui:CheckmarkMedium"></sp-icon>
```

*injected*
```
<sp-icon>${CheckmarkMediumIcon({hidden: true})}</sp-icon>
```

*registered*
```
<sp-icon-checkmark-medium></sp-icon-checkmark-medium>
```

image

Here the registered icon show performance at or above the alternative deployment strategies available, which is a bit surprising. I'll review my tests tomorrow to make sure I've not missed anything. I've also committed the expanded testing so that anyone interested in comparing results can do so locally.

I'm also a bit surprised that the `injected-many` test and the `registered-many` test diverge as much as they do in file size. Though both using more KBs than the `attributes-many` test is as expected being the `iconset` approach is minified by default.

test/benchmark/cli.ts
```diff
@@ -176,32 +176,34 @@ $ node test/benchmark/cli -n 20
             const pjson = await import(
                 pathjoin(process.cwd(), 'packages', packageName, 'package.json')
             );
-            if (pjson.version === '0.0.1') {
+            if (pjson.version === '0.0.1' && opts.compare !== 'none') {
```
Westbrook, line 179: Double check the logic here to see if there is some way to cover this `0.0.1` logic when adding new features as well as when adding completely new packages.

adixon-adobe (commented):
Overall this looks good, and it's definitely nice to have a few benchmarks. I've got a few questions before I mark for approval.

1. What does the average time mean in the benchmarks? Does that include a page load with the icons, or is it just the time to get the icons displayed once the page is loaded?
2. Should those benchmarks get moved to `icons-ui` since that's what you're using for comparison?

Westbrook:
@adixon-adobe The report in https://github.com/adobe/spectrum-web-components/pull/1054#issuecomment-744836842 tracks from `0` to `performance.now()` once all of the elements in the test have `await el.updateComplete;` which includes the full registration and upgrade process of ALL of the icons used in each test.

By default, the benchmarking wasn't prepared to do that, so locally you would have only gotten the upgrade segment of the test, which is showing much less benefit over the full page load than I had expected. To allow for both with less investment, I've added the `-s` argument to the benchmarking, which allows `page` or `element` (default), in order to easily switch between the two styles of testing.

DISCLAIMER: while the recent updates to Spectrum CSS in #1051 are pending you can only fully build `icons-workflow` or `icons-ui` singularly for test, due to icon name collisions. In order to get around this, you can run: `yarn build && rm -rf packages/icons-workflow/icons/` in order to prepare the library to test `icons-ui` and visa versa. The icon names are updated in that PR, so this won't be an issue once the two branches are joined.

Westbrook:
Forgot to share a direct link to one of the best additions in this PR:
https://5fdc0a0c204a0d01167e6d39--spectrum-web-components.netlify.app/storybook/?path=/story/icons--ui-elements
https://5fdc0a0c204a0d01167e6d39--spectrum-web-components.netlify.app/storybook/?path=/story/icons--workflow-elements

You can know search for the icons you're looking for in the Storybook build, a feature that could later be expanded for inclusion in the documentation site:
image

Issue #1055, opened 2020-12-15
Want events when overlay-trigger, sp-dropdown or sp-action-menu opene/close


cuberoot:
### Expected Behaviour

This is a feature request. I'd love to get an event when menus and overlay triggers open and close. I have hit two scenarios where this would have been handy in the last week.

cuberoot:
If this seems reasonable, I'm happy to do the work.

Westbrook:
Totally agree. Would love to see a rough outline of your specific use cases here in the issue for reference, too.

Should this event take into account anything else about the overlay? Type, etc. My use case had always been generically "when did this close" so maybe that's the long and short of it, but having your cases might support the desire for slightly more as far as the API goes here.

Please note that "close" is currently dispatched here: https://github.com/adobe/spectrum-web-components/blob/main/packages/overlay/src/ActiveOverlay.ts#L328 and that the `next` branch is pretty far ahead of the `main` branch (due to some release blocking that hopefully #1051 will release) so it's possible that this change should be branched from there.

cuberoot:
My two use cases are:
- I have a popover within an overlay trigger. The trigger element needs to be styled differently when the popover is open. In this case, I am using a popover to simulate an action menu to work around the issue of needing two separate lists of options in one menu
- In the second case, I have an action menu containing some non-standard controls that need to update using network requests when the menu opens

Westbrook:
Released

PR #1059, opened 2020-12-16
Add open close events to dropdowns and overlay-triggers

cuberoot:
## Related Issue

#1055

## Motivation and Context

I have hit a couple of use cases where it is important to know when an overlay-trigger or a dropdown-related component (e.g. sp-action-menu) opens and closes.

Use cases include:
* Changing the styling of parent control when overlay is open
* Updating menu content on demand

## How Has This Been Tested?

I have added tests to accompany the changes

## Types of changes

## Checklist:

### DIFF
--- a/packages/dropdown/src/Dropdown.ts
+++ b/packages/dropdown/src/Dropdown.ts
@@ -45,6 +45,9 @@ import {
 /**
  * @slot label - The placeholder content for the dropdown
  * @slot {"sp-menu"} - The menu of options that will display when the dropdown is open
+ *
+ * @fires sp-open - Announces that the overlay has been opened
+ * @fires sp-close - Announces that the overlay has been closed
  */
 export class DropdownBase extends Focusable {
     public static openOverlay = async (
--- a/packages/overlay/test/overlay-trigger.test.ts
+++ b/packages/overlay/test/overlay-trigger.test.ts
@@ -10,6 +10,7 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 import { waitForPredicate, isVisible } from '../../../test/testing-helpers.js';
+import { spy } from 'sinon';
 import {
     fixture,
     aTimeout,
@@ -21,7 +22,7 @@ import {
 } from '@open-wc/testing';
 
 import '../overlay-trigger.js';
-import { OverlayTrigger, ActiveOverlay, TriggerInteractions } from '../';
+import { OverlayTrigger, ActiveOverlay, TriggerInteractions, OverlayOpenCloseDetail } from '../';
 import '@spectrum-web-components/button/sp-button.js';
 import { Button } from '@spectrum-web-components/button';
 import '@spectrum-web-components/popover/sp-popover.js';
@@ -652,6 +653,7 @@ describe('Overlay Trigger', () => {
             'hoverContent should still not be visible'
         );
     });
+
     it('acquires a `color` and `size` from `sp-theme`', async () => {
         const el = await fixture<Theme>(html`
             <sp-theme color="dark">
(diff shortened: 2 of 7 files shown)


cuberoot:
@Westbrook This is my first shot at adding the events. Let me know what you think. It needs some docs for the events, but I'm not sure how we are doing that with other components.

I am out for the rest of the week, but wanted to capture these changes before I went.

Westbrook:
Thoughts on using https://github.com/adobe/spectrum-web-components/blob/a7526fb623a166ef2f6f37acba631be8367dce5a/packages/overlay/src/overlay-stack.ts#L148 and https://github.com/adobe/spectrum-web-components/blob/main/packages/overlay/src/overlay-stack.ts#L264 in some way  so as to prevent needing these events in multiple locations?

`@fires` syntax https://jsdoc.app/tags-fires.html will ensure that web component analyzer gets the event data into the API side of the documentation. Regardless of how we resolve the above, each element that leverages the APIs that dispatch these events will need an entry outlining that event. Possibly even derivatives (`sp-action-menu` and `sp-split-button`), too.

You include `cancelable` but I don't see any code to leverage it. I like the idea of making open/close cancelable from the outside, but if we don't actually surface that functionality, we should probably remove that property?

cuberoot:
Thanks for the review. I did a rework of things to shift the event dispatch to `OverlayStack` as suggested. It is definitely a cleaner approach. I did add a way to override the event detail to allow the dropdown to report the `content` as the menu instead of the constructed popover. There may be other classes (e.g. sp-split-button?) that need the override as well.

I added the `@fires` and `@event` tags. Class inheritance seems to propagate the fired events correctly in the docs.

cuberoot:
@Westbrook Looks like some visual failures above. Could these be false positives?

Westbrook:
@cuberoot looks like you were out of date with the target branch. I never expected `next` to last this long, so I'd not set up branch protections...this is why they useful 😓 Added them just now. `next` is passing, so if you rebase your branch onto it then these visual regressions should go green.

packages/overlay/test/overlay-trigger.test.ts
```diff
@@ -696,4 +698,66 @@ describe('Overlay Trigger', () => {
...
+        const el = testDiv.querySelector('#trigger') as OverlayTrigger;
+        const outerButton = testDiv.querySelector(
+            '#outer-button'
+        ) as HTMLElement;
+        const outerPopover = testDiv.querySelector('#outer-popover') as Popover;
+
+        el.addEventListener('sp-opened',openedSpy);
+        el.addEventListener('sp-closed',closedSpy);
+
+        // el.addEventListener('sp-opened', (event) => {
```
Westbrook, line 715: Please remove.

Westbrook (commented):
Overall this looks good. However, I'm not sure why you've set up the `content` value to be overridden like this.

Can you add some context as to capability this surfaces? Some notes on usage and whether or not to do this in the future would be helpful. I've been working on expanding the documentation for the overlay pattern in `westbrook/overlay-docs` and https://5fc82759eb1a831e29768d55--spectrum-web-components.netlify.app/#index (no deep linking in the preview site) and it'll be good to have this info somewhere to apply it to that.

packages/dropdown/src/Dropdown.ts
```diff
@@ -59,9 +62,16 @@ export class DropdownBase extends Focusable {
         target: HTMLElement,
         interaction: TriggerInteractions,
         content: HTMLElement,
+        menuContent: HTMLElement,
```
cuberoot, line 65: Here is my reason for this override (you can talk me out of it if you like).

As a consumer of an API, such as `sp-action-menu` I felt that I would expect the `sp-open` event to report to me on the opening of the item I have put in the slot. I don't think I would expect it to report a component that is an internal implementation artefact of the `sp-action-menu`. In that case, the `sp-action-menu` constructs an `sp-popover` to contain the contents that I have passed in.

This is a minor detail, but worth making an early decision on. I am not dead set on this, I'm just going with my gut feel.

Westbrook: This differentiation does seem useful, I'm still a little unclear on how the `content` value would be leveraged, however. Can you share some context on what you do with this value, as opposed to simply knowing that open/closed has happened? Maybe a test or example of what can/should be done with that additional data when it's received...

cuberoot: A good question. My main reasoning for including it is because of the way overlays work. They reparent the content. So, unless you cache a reference to the content, you will have trouble finding via the DOM it after the overlay has opened because it will have been reparented into an `active-overlay`. This isn't a problem for my current use case, so we could remove the `content` part until it is requested.

Westbrook: If you're not leveraging this API/data, and you felt that we could get by without it for a while, I'd be quite happy moving forward without it for now. Having less code to document/maintain/test feels like the right call at this moment. I worry there likely to be some additional API surface for overlay work soon, anyways, and we could benefit with a little more systematic approach to what opens/closes/etc at that time.

cuberoot:
@Westbrook I have pulled out the `content` field from the open/close event for now.

packages/dropdown/sync/sp-dropdown.ts
```diff
@@ -24,5 +24,10 @@ Dropdown.openOverlay = async (
     content: HTMLElement,
     options: OverlayOptions
 ): Promise<() => void> => {
-    return await Overlay.open(target, interaction, content, options);
+    return await Overlay.open(target, interaction, content, {
+        ...options,
+        openCloseDetail: {
```
Westbrook, line 29: Now that we're not relying on custom data here, should we close over the API by allowing the internals to always set the value of https://github.com/adobe/spectrum-web-components/pull/1059/files#diff-c6eb6d3cfbad5923aeba984b9ed3a1f7fa7bf8d173860247d4d346fb4726c9ffR370, etc. so that it doesn't accidentally get used before we're able to more formally structure it?

cuberoot: Good catch. Change made.

Westbrook: I don't think I was clear. With the following being the current payload of the `open` and `close` events:
```
detail: {
    interaction: details.interaction
},
```
Do we need to add `openCloseDetail` to the options bag on the `open()` method at this time? Maybe there's something you're using that for that I'm missing?

Also, does this want the same simplicity as the above usage? https://github.com/adobe/spectrum-web-components/pull/1059/files#diff-c6eb6d3cfbad5923aeba984b9ed3a1f7fa7bf8d173860247d4d346fb4726c9ffR370

cuberoot: Sorry for that. I think you were clear, I just didn't go back through as carefully as I should have, which meant I wasted some of your time. Sorry about that! I think this is now carved down to the simple change it should have been :-)

Westbrook (approved):
LGTM. Thanks for this!

Once this is through CI, let's merge. 🚀

Issue #1060, opened 2020-12-21
setting the selected item in an sp-dropdown to disabled causes the page to become unresponsive


adixon-adobe:
### Expected Behaviour
It the selected menu item becomes disabled in a dropdown, either the selection is auto-cleared or remains selected even though the item is disabled.

### Actual Behaviour
The page hangs -- I'm guessing there's an infinite loop here.

#### Steps to Reproduce
Select an item in one of the examples on the dropdown documentation page. Using the web inspector, add a `disabled` attribute to the same menu item.

Note that this doesn't seem to reproduce with just `sp-menu`

adixon-adobe:
Our test page makes this reproducible, but it doesn't seem like that likely a scenario. Still... hanging the page is bad!

Westbrook:
Some of this effect is _hopefully_ mitigated by the changes in `next` that prevent a similar infinite loop:  https://github.com/adobe/spectrum-web-components/commit/98cc45b95e99a53e39576ecc43365f9c020d3409#diff-b9cace30081552e1a2c474c35ae00ff7a5c7abe3d7c6b4722179e06d56d4ceb0

We should align on getting `next` into `main` as fast as possible after the new year. There will bumpy releases for consumer between now and when Spectrum CSS goes stable 3.0, but at some point preventing that pain causes divergence issues and fix blockage...

Westbrook:
@adixon-adobe this issue came up in your test suite, right? Can you confirm whether it persists in the new version? It should be good to go, and I'd love to close this issue, but I was never able to successfully recreate in order to do so.

adixon-adobe:
It's fixed -- thanks!

PR #1073, opened 2021-01-08
fix: prevent infinite loops when all children are [disabled]

Westbrook:
## Description
- prevent child searching loops from going infinite by keeping track of available vs disabled item counts
- prevent `sp-menu` from getting stuck on the first item (when `[disabled]`) when focusing programattically
- test these additions.

## Related Issue
ref #1060 
Brought up in Slack

## Motivation and Context
Infinite loops are bad.

## How Has This Been Tested?
- unit tests

## Types of changes

## Checklist:

### DIFF
--- a/packages/menu/src/Menu.ts
+++ b/packages/menu/src/Menu.ts
@@ -59,7 +59,6 @@ export class Menu extends SpectrumElement {
         this.onClick = this.onClick.bind(this);
         this.addEventListener('click', this.onClick);
         this.addEventListener('focusin', this.startListeningToKeyboard);
-        this.addEventListener('focusout', this.stopListeningToKeyboard);
     }
 
     public focus(): void {
@@ -90,10 +89,16 @@ export class Menu extends SpectrumElement {
     }
 
     public startListeningToKeyboard(): void {
-        if (this.menuItems.length === 0) {
-            return;
-        }
         this.addEventListener('keydown', this.handleKeydown);
+        this.addEventListener('focusout', this.handleFocusout);
+    }
+
+    public handleFocusout(): void {
+        this.stopListeningToKeyboard();
+        const focusedItem = this.menuItems[this.focusedItemIndex] as MenuItem;
+        if (focusedItem) {
+            focusedItem.focused = false;
+        }
     }
 
     public stopListeningToKeyboard(): void {
@@ -119,17 +124,25 @@ export class Menu extends SpectrumElement {
     }
 
     public focusMenuItemByOffset(offset: number): void {
+        const step = offset || 1;
         const focusedItem = this.menuItems[this.focusedItemIndex] as MenuItem;
         focusedItem.focused = false;
         this.focusedItemIndex =
             (this.menuItems.length + this.focusedItemIndex + offset) %
             this.menuItems.length;
         let itemToFocus = this.menuItems[this.focusedItemIndex] as MenuItem;
-        while (itemToFocus.disabled) {
+        let availableItems = this.menuItems.length;
+        // cycle through the available items in the directions of the offset to find the next non-disabled item
+        while (itemToFocus.disabled && availableItems) {
+            availableItems -= 1;
             this.focusedItemIndex =
-                (this.menuItems.length + this.focusedItemIndex + offset) %
+                (this.menuItems.length + this.focusedItemIndex + step) %
                 this.menuItems.length;
             itemToFocus = this.menuItems[this.focusedItemIndex] as MenuItem;
+        }
+        // if there are no non-disabled items, skip the work to focus a child
+        if (itemToFocus.disabled) {
+            return;
         }
         itemToFocus.focused = true;
         this.setAttribute('aria-activedescendant', itemToFocus.id);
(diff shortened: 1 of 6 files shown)


adixon-adobe:
I think it'd be worth adding some keyboard events to the test to be sure nothing gets focused, and also make sure the keyboard/focus tests are testing that we skip over disabled items.

packages/menu/src/Menu.ts
```diff
@@ -119,18 +119,26 @@ export class Menu extends SpectrumElement {
...
+            availableItems -= 1;
             this.focusedItemIndex =
-                (this.menuItems.length + this.focusedItemIndex + offset) %
+                (this.menuItems.length + this.focusedItemIndex + step) %
                 this.menuItems.length;
             itemToFocus = this.menuItems[this.focusedItemIndex] as MenuItem;
         }
+        // if there are no non-disabled items, skip the work to focus a child
+        if (itemToFocus.disabled) {
+            return;
```
adixon-adobe, line 140: do you need to do anything to `this.focusedItemIndex` here? I think the other places it's used are safe, but I'm not positive about all the code. I'd at least make sure the places it's used are covered in the all-disabled tests.

adixon-adobe (changes requested):
Overall this looks good. I've got some ideas around test coverage.

packages/menu/src/Menu.ts
Westbrook: Interesting...
```
- item /// focusedItemIndex, because it's the only focusable thing or because === 0 by default
- item[disabled]
- item[disabled]
```
`focusMenuItemByOffset(-1 | 0 | 1)` In this case it'll loop and rest `focusedItemIndex` to what it was as it actually focused the first item.
```
- item[disabled] /// focusedItemIndex === 0 by default
- item[disabled]
- item[disabled]
```
`focusMenuItemByOffset(0)` the case that I think Charles ran into will cause `focusedItemIndex` to start at 0 loop back around to 0 and then do nothing.

Can you share a little more about what you'd like to see in expanded tests here?

Westbrook:
@adixon-adobe we're testing the arrow keys don't allow focus of disabled elements here:

https://github.com/adobe/spectrum-web-components/blob/678d3c9d84a83589effcd89997f9c4f6db752834/packages/menu/test/menu.test.ts#L116-L171

packages/menu/test/menu.test.ts
```diff
@@ -76,6 +78,35 @@ describe('Menu', () => {
         await elementUpdated(el);
         expect(document.activeElement === el, 'self not focused, 2').to.be
             .false;
+        expect(focusinSpy.callCount).to.equal(0);
+    });
+    it('renders w/ all [disabled] menu items', async () => {
+        const focusinSpy = spy();
+        const el = await fixture<Menu>(
+            html`
+                <sp-menu tabindex="0" @focusin=${() => focusinSpy()}>
```
Westbrook, line 87: @adixon-adobe `keydown` listening is only bound _after_ a `focusin` event, this ensure that we don't do any work around this as `focusin` is prevented when the `sp-menu` element is `focus()`ed and there are no children that can accept focus as well as when a child that cannot accept focus is `focus()`ed.

Westbrook:
I'll see if there's a good test case to apply here to prove whether this also fixes #1060.

charlessuh (commented):
Awesome! Thanks so much for fixing this.

packages/menu/src/Menu.ts
```diff
@@ -119,18 +116,26 @@ export class Menu extends SpectrumElement {
     }
 
     public focusMenuItemByOffset(offset: number): void {
+        const step = offset || 1;
         const focusedItem = this.menuItems[this.focusedItemIndex] as MenuItem;
```
adixon-adobe, line 120: If `this.focusedItemIndex` is the last menu item, and I programmatically remove that item, could the logic here lead to issues?

The thing I'm struggling with is that `focusedItemIndex` can be a valid index, but point to something that's not focused. While that doesn't cause issues with the code as-is (except maybe this), it feels fragile.

Westbrook: This is a great question! 

The `childList` or `subtree` of the `sp-menu` element are currently being observed for mutations at which time the following method is called:
```
public updateSelectedItemIndex(): void {
        let index = this.menuItems.length - 1;
        let item = this.menuItems[index] as MenuItem;
        while (index && item && !item.selected) {
            index -= 1;
            item = this.menuItems[index] as MenuItem;
        }
        index = Math.max(index, 0);
        this.menuItems.forEach((item, i) => {
            if (i !== index) {
                item.focused = false;
            }
        });
        this.focusedItemIndex = index;
        this.focusInItemIndex = index;
    }
```
This steps through all of the updated items and reassigns the value of `this.focusedItemIndex` to either the index of an item with `selected` or the zeroth item in the list. That means almost all cases `this.focusedItemIndex` will point to am existing item. I'm seeing this case:
```
selectedItem.remove(); // remove the focused item
menu.focus(); // immediately focus the menu
// in the case that it was the last item and no replacement item is applied nothing will be visual selected.
```
I'll address this and add a test to track regressions.

Westbrook: @adixon-adobe I think we're coving this use case in a good way now. Let me know what you think!

adixon-adobe (approved):
I think you've covered all my concerns in the tests now. Thanks for the extra explanations.

PR #1076, opened 2021-01-11
fix: use icons without "size" values

### LINKED ISSUE #669: sp-icon size attribute should be deprecated

Westbrook:
## Description
Remove the use of the `size` attribute in `sp-icon` when scoped to `sp-button` and `sp-action-button` components.
- default to `size='m'` in the `:host` CSS so that it's easily overridden by the specific sizes and by scoped values of CSS Custom properties
- surface upcoming CSS Custom Property application in our local implementation until Spectrum CSS ships this update
- no usage changes required new golden images, but added icons to `sp-button` VRT for security

This doesn't "deprecate" the size attribute, as it can still be used as an override, but it's usage is lowered in priority.

Available for live review:
https://5ffc5bf40cc6c30ed2338e23--spectrum-web-components.netlify.app/storybook/index.html?path=/story/action-button-sizes--s
https://5ffc5bf40cc6c30ed2338e23--spectrum-web-components.netlify.app/storybook/index.html?path=/story/button-primary-sizes--s

## Related Issue
fixes #669

## Motivation and Context
Simplify!

## How Has This Been Tested?
VRTs passed for the `sp-action-button` et al modifications.
VRTs were added for `sp-button` with icon usage

## Types of changes
- [x] refactor

## Checklist:


najikahalsema (approved):

packages/accordion/src/AccordionItem.ts
```diff
@@ -97,11 +97,7 @@ export class AccordionItem extends Focusable {
                 >
                     ${this.label}
                 </button>
-                <sp-icon
-                    id="indicator"
-                    size="xs"
-                    class="spectrum-UIIcon-ChevronRight100"
-                >
+                <sp-icon id="indicator" class="spectrum-UIIcon-ChevronRight100">
```
adixon-adobe, line 100: was this just not supposed to be `xs` before?

packages/action-group/stories/action-group.stories.ts
```diff
@@ -322,17 +322,17 @@ export const compactIconsOnlyJustified = (): TemplateResult => {
...
             </sp-action-button>
             <sp-action-button label="Info">
-                <sp-icon slot="icon" size="m">
+                <sp-icon slot="icon">
                     ${InfoIcon({ hidden: true })}
                 </sp-icon>
             </sp-action-button>
             <sp-action-button label="View All Tags">
-                <sp-icon slot="icon" size="m">
+                <sp-icon slot="icon">
```
adixon-adobe, line 335: I like the changes to this file!

packages/button/stories/button-cta.stories.ts
```diff
@@ -67,6 +67,22 @@ export const withIcon = (): TemplateResult => {
...
+        <sp-button label="Edit" size="xl" variant=${variant}>
+            <sp-icon slot="icon" size="s">
+                ${HelpIcon({ hidden: true })} Testing
+            </sp-icon>
+        </sp-button>
+        <h1>For testing purposes only</h1>
+        <p>
+            Spectrum specifies that icon elements inside of buttons should share
+            the same "size" rating or have icon accept size values as scoped in
+            the CSS by the button element.
```
adixon-adobe, line 81: I'm not totally following what this is trying to say. I'm guessing there's a clearer way to word this.

adixon-adobe (approved):
I've got a couple of questions, and am a little confused by some wording in the override stories, but the actual changes look great. I don't see anything that'd block merging.

Very glad to finally see this change!

packages/accordion/src/AccordionItem.ts
Westbrook: `xs` was always a _approximation_ of the size here. The class was actually overriding it as this is a UI icon. With the update directly before this I could have used `size="none"` and be more semantically correct, but this seems like a good opportunity to flush things we don't need. What's more, searching for `<sp-icon` and removing _every_ `size` attribute seemed like a good test for this change...

packages/button/stories/button-cta.stories.ts
Westbrook: This is a test to ensure that sizing the icon will still work when it's in the scope of a parent element. You shouldn't do this on a regular basis as it is a departure from the Spectrum specification. 

Is that better, or worse, or does it give you any good ideas here?

PR #1104, opened 2021-01-15
fix: missing dependency

### LINKED ISSUE #1103: Undeclared dependency in styles package
### Expected Behaviour
I can use Spectrum Web Components with Yarn 2 without seeing dependency errors.

### Actual Behaviour
.yarn/cache/@spectrum-web-components-styles-npm-0.7.0-769d2cfc5c-b606d82381.zip/node_modules/@spectrum-web-components/styles/src/spectrum-base.css.js
Module not found: @spectrum-web-components/styles tried to access @spectrum-web-components/base, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.

#### Additional info

You can see this in the **generated** code here: https://unpkg.com/browse/@spectrum-web-components/styles@0.7.0/src/spectrum-base.css.js

There isn't actually any reference to '@spectrum-web-components/base' in the source code for the styles package though,  so there's something about the tooling that adds it.

adixon-adobe:
The dependency is added by the build, so you won't find it in the code.

## Related Issue

Fixes #1103
 
## Motivation and Context

Yarn 2 has strict dependency checks, so this will fix errors with that.

## How Has This Been Tested?

It's not really testable until we release.

## Types of changes

## Checklist:

### DIFF
--- a/packages/styles/package.json
+++ b/packages/styles/package.json
@@ -42,6 +42,7 @@
     "author": "",
     "license": "Apache-2.0",
     "devDependencies": {
+        "@spectrum-web-components/base": "^0.2.0",
         "@spectrum-css/commons": "^3.0.0",
         "@spectrum-css/typography": "^3.0.0-beta.1",
         "@spectrum-css/vars": "^3.0.0-beta.2"


Westbrook (approved):
LGTM.

LGTci.

Thanks!

PR #1105, opened 2021-01-15
Westbrook/textfield labels

### LINKED ISSUE #742: All textfield examples should be labelled
Please provide examples with visible labels that are associated with the textfield component. 
URL tested: https://opensource.adobe.com/spectrum-web-components/components/textfield

Westbrook:
## Description
All `sp-textfields` and `sp-textfield multiline` usage in our documentation site are now labelled.

- catches an assumption by `sp-field-label` that its `for` element had to be present
- expands delivery to include t-shirt sizing

## Related Issue
fixes #742

## Motivation and Context
Accessibility and availability of features.

## How Has This Been Tested?
Manually

image

## Types of changes
- [x] documentation

## Checklist:

### DIFF
--- a/packages/field-label/src/FieldLabel.ts
+++ b/packages/field-label/src/FieldLabel.ts
@@ -17,6 +17,7 @@ import {
     TemplateResult,
     property,
     PropertyValues,
+    SizedMixin,
 } from '@spectrum-web-components/base';
 import type { Focusable } from '@spectrum-web-components/shared';
 import { Asterisk100Icon } from '@spectrum-web-components/icons-ui';
@@ -30,7 +31,7 @@ type AcceptsFocusVisisble = HTMLElement & { forceFocusVisible?(): void };
 /**
  * @element sp-field-label
  */
-export class FieldLabel extends SpectrumElement {
+export class FieldLabel extends SizedMixin(SpectrumElement) {
     public static get styles(): CSSResultArray {
         return [styles, asteriskIconStyles];
     }
@@ -51,9 +52,6 @@ export class FieldLabel extends SpectrumElement {
 
     @property({ type: String, reflect: true, attribute: 'side-aligned' })
     public sideAligned?: 'start' | 'end';
-
-    @property({ type: String, reflect: true })
-    public size = 'm';
 
     private target?: HTMLElement;
 
@@ -77,6 +75,9 @@ export class FieldLabel extends SpectrumElement {
         }
         const parent = this.getRootNode() as HTMLElement;
         const target = parent.querySelector(`#${this.for}`) as Focusable;
+        if (!target) {
+            return;
+        }
         if (typeof target.updateComplete !== 'undefined') {
             await target.updateComplete;
         }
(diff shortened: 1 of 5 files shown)


packages/field-label/src/FieldLabel.ts
```diff
@@ -77,6 +75,9 @@ export class FieldLabel extends SpectrumElement {
         }
         const parent = this.getRootNode() as HTMLElement;
         const target = parent.querySelector(`#${this.for}`) as Focusable;
+        if (!target) {
+            return;
+        }
```
adixon-adobe, line 80: Can you add a unittest that checks this as well?

adixon-adobe (changes requested):
Better docs! Nice!

Just one test request.

packages/field-label/src/FieldLabel.ts
Westbrook: I was trying to make that work, but was having trouble getting that error to actually catch anywhere beyond it printing to the console. I will take another pass, but if you have any experience to share around testing that errors are/are not thrown, I'd very much appreciate it!

adixon-adobe: Since this is an async function the throw will convert to a rejected promise. That might be why you're not finding it with exceptions? I've made some progress by changing `manageFor` to be public so I can spy on it. Didn't quite crack it though -- here's where I stopped:

```
    it('loads with invalid "for"', async () => {
        const el = await fixture<FieldLabel>(
            html`
                <sp-field-label for="not_exist">Input label</sp-field-label>
            `
        );
        const manageSpy = sinon.spy(el, 'manageFor')
        await elementUpdated(el);

        const result = manageSpy.returnValues[0]; // result is null here, and not a promise :(
        try {
            await result;
        } catch {
            expect(true).to.be.false;
        }
    });
```

adixon-adobe: I'm also changing this PR to approved. It does seem like we'll want to figure out a way to do a little deeper testing/validation (but maybe in some of these cases we just need to add some error state we can check when things aren't right?)

adixon-adobe: or maybe I'll check in a really basic test so that you'd at least see the console errors in the test output pre-fix.

adixon-adobe (dismissed):
Added an imperfect test to cover the fix.

packages/field-label/src/FieldLabel.ts
Westbrook: This got me doing in the right direction, @adixon-adobe! I've got a test outlining the ability to have `for` that doesn't point to anything without throwing an error.

This does point to needing to add a MutationObserver later to support late additions of `for` value resolutions, but let's address that it a separate PR.

packages/field-label/test/field-label.test.ts
```diff
@@ -64,6 +65,35 @@ describe('FieldLabel', () => {
...
+                try {
+                    await ((FieldLabel.prototype as unknown) as {
+                        manageFor(): Promise<void>;
+                    }).manageFor.apply(el, ...args);
+                } catch (error) {
+                    return 'Error was thrown.';
+                }
+                return 'No error was thrown.';
+            }
+        );
```
adixon-adobe, line 90: when your tests are more complex than the underlying code... ;-)

adixon-adobe (dismissed):
Awesome -- glad you were able to figure that out!

PR #1110, opened 2021-01-16
ci: parallelize the setup work

Westbrook:
## Description
Rework the CI flow to increase parallelization and reduce duplicated work.

## Motivation and Context
Further, faster!


adixon-adobe (approved):
I like faster. Nice!

PR #1111, opened 2021-01-16
fix(tooltip): correct arrow orientation in horizontal placements

### LINKED ISSUE #1099: Horizontal tooltip arrows mis-render when nested in overlay triggers
I isolated that tooltip issue. Maybe I'm using overlay incorrectly?

### Expected Behaviour

`<sp-tooltip placement="left">` renders its shape identically whether or not it's within `<overlay-trigger>`.

### Actual Behaviour

A `<sp-tooltip>` within `<overlay-trigger>` with a horizontal `placement` (left, right, left-end, etc) will render its arrow as if it's in a "bottom" placement.

https://webcomponents.dev/edit/tf6L6yRntuzjJsCUE9Ek

#### Steps to Reproduce

Hover over the various action buttons:

https://user-images.githubusercontent.com/364501/104680204-57655b00-56bd-11eb-8db4-b8792c659b75.mp4

hunterloftis:
## Description

This adds rotation to the existing scale transforms for tooltip overlays.

## Related Issue

Fixes #1099 

## How Has This Been Tested?

`yarn test` yields missing library errors on my machine (eg "browserType.launch: Host system is missing dependencies!  (failed on Firefox)"), so I created a story that reproduced the issue.

image

## Types of changes

## Checklist:

- [X] I have read the **CONTRIBUTING** document.
- [X] I have added tests to cover my changes.
- [ ] All new and existing tests passed. (we'll see in CI!)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ yarn-error.log
 .npmrc
 .tsbuildinfo
 .rush
+.devcontainer
 
 documentation/components/searchIndex.json
 
--- a/packages/tooltip/src/tooltip.css
+++ b/packages/tooltip/src/tooltip.css
@@ -35,3 +35,49 @@ governing permissions and limitations under the License.
 :host([data-popper-placement*='bottom']) #tip {
     margin-left: 0;
 }
+
+:host([placement*='right']) #tip,
+:host([placement*='left']) #tip {
+    border: solid
+        var(
+            --spectrum-tooltip-tip-height,
+            var(--spectrum-global-dimension-size-50)
+        )
+        transparent;
+    border-left-color: var(
+        --spectrum-tooltip-background-color,
+        var(--spectrum-global-color-static-gray-700)
+    );
+}
+
+:host([placement*='left']) #tip {
+    transform: none;
+}
+
+:host([placement*='right']) #tip {
+    transform: scaleX(-1);
+}
+
+:host([variant='negative'][placement*='right']) #tip,
+:host([variant='negative'][placement*='left']) #tip {
+    border-left-color: var(
+        --spectrum-tooltip-negative-background-color,
+        var(--spectrum-global-color-static-red-700)
+    );
+}
+
+:host([variant='info'][placement*='right']) #tip,
+:host([variant='info'][placement*='left']) #tip {
+    border-left-color: var(
+        --spectrum-tooltip-info-background-color,
+        var(--spectrum-global-color-static-blue-700)
+    );
+}
+
+:host([variant='positive'][placement*='right']) #tip,
+:host([variant='positive'][placement*='left']) #tip {
+    border-left-color: var(
+        --spectrum-tooltip-positive-background-color,
+        var(--spectrum-global-color-static-green-700)
+    );
+}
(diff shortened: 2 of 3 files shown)


Westbrook:
Can you check into whether this inverts the regression and breaks the arrow in popover content?

We’ve got an issue open to allow overlay content to be open declaratively which would surface the ability to check this in visual regression testing and ensure these sorts of regressions don’t catch us again.

hunterloftis:
Good call; it does break horizontal popover overlays:

image

Looking at how the tip arrow is implemented - tooltips have a single arrow that adapts with CSS to the different placements (the `transform: rotate` here is what's overridden by the inline `transform` when they're used in an overlay). Popovers have two different tip arrows, one rendered for left/right and the other for top/bottom placements.

I'll see about a fix that doesn't regress the popovers...

hunterloftis:
Huh, there may be an opportunity to remove a bunch of code related to the arrows. Going to play with it for a minute to see.

hunterloftis:
I kept the repro story ("overlaid") and replaced the implementation with a change that makes tooltips work like popovers: two different rendering modes (vertical/horizontal) + scaleX to mirror. Now, the arrow-rotation code that works for popovers works the same way for tooltips.

image

image

image

Along the way I noticed an opportunity to simplify -

What would the tradeoffs be for wrapping the tooltip's `#tip` in a containing element like `#tip-position` upstream? I avoided making DOM changes, since SWC consumes Spectrum CSS, but wrapping `#tip` would prevent conflicts with PopperJS over the `transform` property. That would enable removing `popper-arrow-rotate` altogether and would unify the rendering paths of Spectrum CSS (static) and Spectrum Web Components (with popper).

- Current: Spectrum CSS <- overrides to `#tip` in `tooltip.css` <- another layer of overrides from `popper-arrow-rotate`
- Positioning wrapper: Spectrum CSS <- (used directly, both popper and SCSS place `#tip-position`)

hunterloftis:
[Here's](https://github.com/hunterloftis/spectrum-web-components/pull/2) a working demo of that last idea.

Westbrook:
Interesting take on this. With the attribute setting the "expected" location of an overlay and popper deciding based on actual run time realities the "actual" location, does this `:after` approach work what a `:host([placement*='right']) #tip` tooltip is actually placed on the "left" due to lack of page real estate?

Thanks for digging deep into this and experimenting with various demos/approaches to this situation!

hunterloftis:
Yep, since ActiveOverlay already [updates the placement whenever popper changes it](https://github.com/adobe/spectrum-web-components/blob/main/packages/overlay/src/ActiveOverlay.ts#L257), the component does the right thing. Here's that demo branch on the situation you described:

https://user-images.githubusercontent.com/364501/104858741-67fd1780-58ef-11eb-84f9-646d44f10e46.mp4

Westbrook:
This process looks pretty nice to me. I'm having a little trouble tracking this across the multiple PRs, however. To support simplifying, I've repointed this PR to a new base, could you rebase to that and bring all your changes together. I like the idea of not having to do the rotation management, 100%. Making sure that the delta in DOM from Spectrum CSS to do that in the `sp-overlay` elements will be important for maintenance long term, so it'll be really useful to see what you see that looking like.

All things being equal, I'd love to get this cleaned up and into the release for next week!

Westbrook:
Separate benefit of the new base branch, once we get this through, I want to look into whether we can have these things available in visual regression so that this doesn't occur again in the future.

hunterloftis:
Awesome - I'll finish it up by next week then!

hunterloftis:
Closing in favor of #1137

PR #1114, opened 2021-01-18
Dialog updates

### LINKED ISSUE #1036: Esc key does not dismiss dialog opened with sp-dialog-wrapper
### Expected Behaviour

Pressing **Esc** key should close a modal dialog. Since the dialog could be opened via keyboard, it should be dismissable via keyboard.

### Actual Behaviour

Pressing **Esc** has no effect. Workaround is to click the X in the top right corner or outside of the dialog area to dismiss it. 

#### Steps to Reproduce

1. Go to https://opensource.adobe.com/spectrum-web-components/components/dialog-wrapper
2. Click the first Toggle Dialog button on the page (under the "Small" Example)
3. With the dialog now open, try pressing Esc. Nothing happens
image

#### Platform and Version
Tested on Chrome 87 (Mac)

See docs

N/A

### LINKED ISSUE #1037: After opening sp-dialog, user should not be able to tab through content beneath the dialog
### Expected Behaviour

After a modal dialog is opened, the dialog should contain its tab sequence; i.e. the user should only be able to tab to UI elements within the modal (see [WAI ARIA practices](https://www.w3.org/TR/wai-aria-practices-1.2/#dialog_modal)).

For reference see [React Spectrum Dialog](https://react-spectrum.adobe.com/react-spectrum/Dialog.html). Upon opening a dialog, focus goes to the dialog element, then is contained within the dialog.

I'm not clear what defines whether an sp-dialog is 'modal', it looks like perhaps those with "underlay" should be considered modal.

### Actual Behaviour

After opening a dialog on the SWC documentation site, focus unexpectedly goes to the `<docs-component>` element beneath it, and pressing tab cycles through the many elements beneath the dialog. Eventually focus is eventually give to the dialog.

#### Steps to Reproduce

1. Go to https://opensource.adobe.com/spectrum-web-components/components/dialog-wrapper
2. Click the first Toggle Dialog button on the page
3. With the dialog now open, notice focus is now on the `<docs-component>` element. Judging by RS3, it should be on the `<sp-dialog>`. 
image
4. Press tab. Notice focus is still on some element beneath the dialog. It should instead be on the X button in the modal.
image

#### Platform and Version
Tested on Chrome 87 (Mac)

See docs

N/A

### LINKED ISSUE #1112: sp-dialog-wrapper in not placed in the middle on firefox
### Expected Behaviour
On Chrome:
图片

### Actual Behaviour
On Firefox:
图片

#### Platform and Version
Firefox Developer Edition 85.0b9
https://opensource.adobe.com/spectrum-web-components/components/dialog-wrapper

Westbrook:
## Description
- update settings in PostCSS Preset Env to leverage `browsers` listing, similar to Spectrum CSS
  - prevents the use of `height: -moz-available;` which was causing layout issues in Firefox
  - prevents use of `-moz-user-select: none;` which is not needed as of Firefox 68, current is 84
- leverage `:focus-visible` polyfil mixin in `sp-dialog` and `sp-dialog-wrapper`
  - allows reparented elements to ensure the availability of the polyfill for their content
- update `sp-dialog-wrapper` documentation to leverage `overlay-trigger`

## Related Issue
fixes #1112
fixes #1037
fixes #1036

## Motivation and Context
- clearer usage instructions
- better delivery in Firefox and Safari

## How Has This Been Tested?
manually

## Types of changes

## Checklist:


scripts/css-processing.js
```diff
@@ -21,6 +21,12 @@ const postCSSPlugins = (resourcePath) => {
         require('postcss-inherit')(),
         require('postcss-preset-env')({
             stage: 0,
+            browsers: [
+                'last 2 Chrome versions',
+                'last 2 Firefox versions',
+                'last 2 Safari versions',
+                'last 2 iOS versions',
```
adixon-adobe, line 28: Can you go with the last 3 iOS & Safari versions? While Chrome/Firefox auto-update, Safari can get blocked from updating on older OS versions, so it's a little more likely to be behind.

That doesn't mean we need to actively support those browsers, but I'd like to keep css from breaking because of this rule.

adixon-adobe (changes requested):
One small change request.

scripts/css-processing.js
Westbrook: I copied this from here: https://github.com/adobe/spectrum-css/blob/a3db903a3f6a1cab4b1c164d66f7eaf9813afe99/tools/component-builder/css/processors.js#L320-L328 but am by no means married to it.

I should also double check, but 2 versions of Chrome is roughly 12 - 18 weeks, where as 2 versions of Safari is roughly 2 years...

Will change, unless some of this changed your oppinion.

adixon-adobe: Yeah -- it's roughly 2 years (but can be compounded by delayed OS updates), so the updates also lag quite a bit more for that last 5-10%.

Westbrook: Cool. I'll bump that up. As long as it's "the same" enough, work for me!

scripts/css-processing.js
```diff
@@ -21,6 +21,12 @@ const postCSSPlugins = (resourcePath) => {
         require('postcss-inherit')(),
         require('postcss-preset-env')({
             stage: 0,
+            browsers: [
+                'last 2 Chrome versions',
+                'last 2 Firefox versions',
+                'last 3 Safari versions',
```
Westbrook, line 27: @adixon-adobe updated, and the one thing it "added" was already being shipped, so it doesn't even show up here...good to go?

adixon-adobe (approved):
Thanks!

PR #1127, opened 2021-01-20
fix(textfield): reimplement min/maxlength

### LINKED ISSUE #1118: Textfield: Maxlength attribute isn't working as expected
### Expected Behaviour

The `maxlength` attribute should prevent the user from typing indefinitely once the character count is reached.

### Actual Behaviour

The `maxlength` attribute has no effect and the user can type indefinitely.

#### Steps to Reproduce

This should limit entry to 2 characters:

`<sp-textfield maxlength="2" id="name-0" placeholder="Enter your name"></sp-textfield>`

Westbrook:
## Description
- reapply min/maxlength properties to `sp-textfield` form elements
- leverage `sendKeysPlugin` at test time to allow for interaction with Playwright to do integration testing on these values
  - NOTE: JS based faking of these interactions do not suffice to appropriately test these values
  - surfaces the `type` and `press` interfaces of `page.keyboard` but more could be added later
- use the `live()` binding for the values as realtime rebinding to the `value` property on an `<input />` causes the `tooShort` validity check to be overridden

## Related Issue
fixes #1118

## Motivation and Context
Actually implementing the APIs we surface... 🙈 

## How Has This Been Tested?
- unit (integration?) tests

## Types of changes

## Checklist:


adixon-adobe (approved):
The real question is how many people know that this is enforcing UTF-16 code points. "👍😻👺🧟‍♀️👑🦑🐲🙈🙉🙊".length == 23

PR #1130, opened 2021-01-21
build: ensure -webkit-sticky in CSS output

Westbrook:
Somehow the code landing in main left this change uncommitted. 🤔


adixon-adobe:
It's possible you forgot to push your last change (and that I didn't catch the subtlety).

Westbrook:
The correction required here means that I did some combination of not pushing, not building, or something, somewhere, yes. We agreed to go to Safari -3 and it looked like it added back the output I expected, but then when everything got into `main` it wanted to remove `position: -webkit-sticky;` listings again. This goes enough versions back to maintain them...

Issue #1024, opened 2020-11-25
Importing components forces registration without check


cuberoot:
### Expected Behaviour

I should be able to use SWC in two different javascript bundles from a single page.

### Actual Behaviour

We get errors when `customElements.define` is called and I cannot see any way around it.

### Reproduce Scenario (including but not limited to)

Build two different code bundles that include SWC components. Try to load them into the same page.

#### Steps to Reproduce

Unpack the sample code below and then do:
```
yarn install
yarn build
open index.html
```

#### Sample Code that illustrates the problem

[swc-collision.zip](https://github.com/adobe/spectrum-web-components/files/5593826/swc-collision.zip)

#### Logs taken while reproducing problem

Console logs:

```
Uncaught DOMException: Failed to execute 'define' on 'CustomElementRegistry': the name "sp-icon" has already been used with this registry
    at eval (webpack:///./node_modules/@spectrum-web-components/icon/sp-icon.js?:15:16)
    at Module../node_modules/@spectrum-web-components/icon/sp-icon.js (file:///Users/mitaylor/Scratch/swc-collision/dist/library_b.js:889:1)
    at __webpack_require__ (file:///Users/mitaylor/Scratch/swc-collision/dist/library_b.js:20:30)
    at eval (webpack:///./node_modules/@spectrum-web-components/dropdown/src/Dropdown.js?:13:98)
    at Module../node_modules/@spectrum-web-components/dropdown/src/Dropdown.js (file:///Users/mitaylor/Scratch/swc-collision/dist/library_b.js:865:1)
    at __webpack_require__ (file:///Users/mitaylor/Scratch/swc-collision/dist/library_b.js:20:30)
    at eval (webpack:///./node_modules/@spectrum-web-components/dropdown/sp-dropdown.js?:2:74)
    at Module../node_modules/@spectrum-web-components/dropdown/sp-dropdown.js (file:///Users/mitaylor/Scratch/swc-collision/dist/library_b.js:853:1)
    at __webpack_require__ (file:///Users/mitaylor/Scratch/swc-collision/dist/library_b.js:20:30)
    at eval (webpack:///./library_b.js?:3:103)
```

Westbrook:
[...]

PR #1134, opened 2021-01-22
fix to allow multiple imports

dsadhanala:
## Description
This PR attempts to prevent register/defining elements more than once when they are loaded multiple times on the same page or DOM tree.

Added new module `defineElement` which is simple function which checks for existing element registration before attempting to define as customElement.

## Related Issue
https://github.com/adobe/spectrum-web-components/issues/1024

## Motivation and Context
In one of our project we load multiple micro front-end modules from CDN, they are bundled separately but underneath are reusing some of the SWC components in all of them. Given that the way SWC define customElements it's throwing exception as components from different bundles are trying to register/define multiple times.

Below error for reference:
```JS
header-f8fa9b39.js:1839 Uncaught DOMException: Failed to execute 'define' on 'CustomElementRegistry': the name "sp-icon" has already been used with this registry.
```

## How Has This Been Tested?
Similar workaround has been in use for our own component library to allow such cases, which is working as expected in prod. I'm hoping automation in PR will validate this, if any specific testing needs to be done manually, let me know I'll make sure to run through.

## Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [ ] I have signed the [Adobe Open Source CLA](http://opensource.adobe.com/cla.html).
- [ ] My code follows the code style of this project.
- [ ] I have read the **CONTRIBUTING** document.

### DIFF
--- a/packages/base/src/defineElement.ts
+++ b/packages/base/src/defineElement.ts
@@ -1,0 +1,9 @@
+export const defineElement = (elementName: string, elementConstructor: CustomElementConstructor, options?: { extends: string; }): void => {
+    // check for element existence
+    if (customElements.get(elementName)) return;
+
+    // define as custom element
+    window.customElements.define(elementName, elementConstructor, options);
+};
+
+export default defineElement;
--- a/documentation/src/components/adobe-logo.ts
+++ b/documentation/src/components/adobe-logo.ts
@@ -40,4 +40,6 @@ class SpectrumLogo extends LitElement {
         `;
     }
 }
-customElements.define('docs-spectrum-logo', SpectrumLogo);
+import { defineElement } from '@spectrum-web-components/base';
+
+defineElement('docs-spectrum-logo', SpectrumLogo);
--- a/documentation/src/components/component.ts
+++ b/documentation/src/components/component.ts
@@ -298,4 +298,6 @@ class ComponentElement extends RouteComponent {
         `;
     }
 }
-customElements.define('docs-component', ComponentElement);
+import { defineElement } from '@spectrum-web-components/base';
+
+defineElement('docs-component', ComponentElement);
--- a/documentation/src/components/guide.ts
+++ b/documentation/src/components/guide.ts
@@ -53,4 +53,6 @@ class GuideElement extends RouteComponent {
         return result || html``;
     }
 }
-customElements.define('docs-guide', GuideElement);
+import { defineElement } from '@spectrum-web-components/base';
+
+defineElement('docs-guide', GuideElement);
--- a/documentation/src/components/home.ts
+++ b/documentation/src/components/home.ts
@@ -103,4 +103,6 @@ class HomeElement extends RouteComponent {
         `;
     }
 }
-customElements.define('docs-home', HomeElement);
+import { defineElement } from '@spectrum-web-components/base';
+
+defineElement('docs-home', HomeElement);
--- a/documentation/src/components/page.ts
+++ b/documentation/src/components/page.ts
@@ -37,4 +37,6 @@ export class Page extends LayoutElement {
     }
 }
 
-customElements.define('docs-page', Page);
+import { defineElement } from '@spectrum-web-components/base';
+
+defineElement('docs-page', Page);
--- a/documentation/src/components/side-nav-search.ts
+++ b/documentation/src/components/side-nav-search.ts
@@ -174,4 +174,6 @@ class SearchComponent extends LitElement {
     }
 }
 
-customElements.define('docs-search', SearchComponent);
+import { defineElement } from '@spectrum-web-components/base';
+
+defineElement('docs-search', SearchComponent);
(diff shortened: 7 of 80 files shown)


Westbrook:
Please update your commit messages to follow our contributing guidelines: https://github.com/adobe/spectrum-web-components/blob/main/CONTRIBUTING.md#commitlint

packages/base/src/defineElement.ts
```diff
@@ -0,0 +1,9 @@
+export const defineElement = (elementName: string, elementConstructor: CustomElementConstructor, options?: { extends: string; }): void => {
+    // check for element existence
+    if (customElements.get(elementName)) return;
```
Westbrook, line 3: Relying on something like this at the library level leaves a number of important questions and concerns found in #1024 unanswered. As you've used a pattern similar to this in the past, maybe you can help to answer them?

1. As a consumer who ends up with multiple versions of the same package in their project without my knowledge, I'd like to be made aware of this situation so that I can further optimize the code I deliver to my end users. With this gate, subsequent registration of the same element will be squelched and other than doing a deep dive of their built code or project lock, how do you see such a consumer discovering this situation?
2. As a consumer who chooses to use dependencies with mismatched package versions, I'd like to be able to know what version actually gets preferred by my build/application. As this technique prefers the first registration, how can a consumer know that this pattern has registered the one they'd actual like registered? This can be particularly useful when leveraging bundled code.
3. As a consumer who chooses to leverage multiple copies of the same version of a package in a projects, I'd like to be made aware so that I can reduce the code I ship to my users. How would a consumer in this case know that they've been building the code they deliver to their end users appropriately in this context? If I understand correct, this is very close to your use case and is an important goal to keep in mind. JS is the MOST expensive asset that we can ship to our users and patterns like this that can conceal how we're handing cost like this to our users can be a foot gun to longer term performance goals.
4. As a library vendor, I'd like to deliver quality and performance improvements to my consumers with the lowest likelihood of breaking changes. I may be overly optimistic that any later change to this can be made without it being breaking, but, particularly as this related the number 2 above, WHEN we move away from this pattern in the future I don't see a way to do it that isn't breaking. When something like Scoped Registries releases the code order prioritization of this approach there will be elements on a page that suddenly do not work the way they did before.
5. As a consumer of custom elements, I'd like to know when the libraries I leverage have diverged from the specification so I can rationalize what that means for my project. How can we document the decisions required to satisfy the above questions so that we can place consumers on the best track towards success? The best location I can think of just now is https://opensource.adobe.com/spectrum-web-components/getting-started. If we can come to an agreement on this, adding those findings to a section on that page would likely be enogh for now. Longer term does expanding that into a "Developing with SWC" section seems like a good idea?

Westbrook:
If we decide to take this path, we'll need to include this approach to https://github.com/adobe/spectrum-web-components/blob/main/packages/icons-ui/bin/build.js#L205 and https://github.com/adobe/spectrum-web-components/blob/main/packages/icons-workflow/bin/build.js#L223

Westbrook:
Also, @dsadhanala, it looks like you have access to branch directly on this repo. That would allow your PR to leverage more of our CI infrastructure in the future.

Westbrook (changes requested):
Chatted directly with @dsadhanala to clarify short/mid/long term approaches to this. Setting the PR to "Request changes" to outline that it isn't for merging at the current time. Next steps at current are as follows:

1. Leverage monkey patching `customElements.define()` to unblock development today.
2. Vend a `webpack` (possibly others in the future as needed) plugin to convert `customElements.define('docs-spectrum-logo', SpectrumLogo);` to `if (`customElements.get(elementName)) customElements.define('docs-spectrum-logo', SpectrumLogo);` to support short term production level goals when intending to ship multiple versions on a single page. (Targeting 29 Jan)
3. Review timeline for a Scoped Element Registry solution with the Polymer/OpenWC teams on Wed, 27 January.
4. Assuming that timeline is in an acceptable short/mid term, wait to make library internal changes until that becomes available.
5. If that timeline is unacceptable, revisit this PR and/or derivatives to ship registration safety mainline of the SWC library.

Westbrook:
It looks like a rough timeline of 3-4 months for a Scoped Registries polyfill is going to be realistic as of this writing, and for that we will be closing this PR. In the short term it is possible to leverage the following webpack configuration to apply something similar to this PR at time of use:
```
{
    test: /\.js$/,
    exclude: /node_modules\/(?!(@spectrum-web-components)\/).*/,
    loader: 'string-replace-loader',
    options: {
        search: /customElements.define\('(.*)'/,
        replace: (_match, p1) => `!customElements.get('${p1}') && customElements.define('${p1}'`,
        flags: 'g'
    }
}
```

packages/base/src/defineElement.ts
chnicoloso: I understand wanting the library to offer opportunities for consumers to catch potential mistakes and opportunities for optimization, but having the library crash does not seem like the right way to do that. Could we guard against crashing and add a warning in the console? There are [cases] beyond version mismatches (https://wiki.corp.adobe.com/display/assetscollab/Express+Icons+Strategy) where we may want to reuse a common namespace

PR #1137, opened 2021-01-24
fix(tooltip): correct arrow orientation, remove popper-arrow-rotate

### LINKED ISSUE #1099: Horizontal tooltip arrows mis-render when nested in overlay triggers

hunterloftis:
## Description

This fixes #1099.

The root cause is Popper overriding `transform` on any arrow element passed to it (`#tip`). This solution treats `transform` as an unavailable property on the top-level arrow element. Instead, it uses `:after` on Tooltip and `.tip` on Popover to orient the arrows, leaving `transform` free for Popper to position the arrows.

Two reasons for a CSS-oriented solution:

1. It lets us delete [popper-arrow-rotate.ts](https://github.com/adobe/spectrum-web-components/blob/b06e92d43a921bd53b47f635bc7c11e54a3b52c4/packages/overlay/src/popper-arrow-rotate.ts)
2. These changes could be integrated upstream into Spectrum CSS.

## How Has This Been Tested?

Lots of manual testing! Here's my setup, please lmk if there's anything else I should look at:

https://user-images.githubusercontent.com/364501/105639875-94151d00-5e48-11eb-963f-159c95016fcc.mp4

## Types of changes

## Checklist:

- [X] I have read the **CONTRIBUTING** document.
- [X] I have added tests to cover my changes.
- [ ] All new and existing tests passed. (we'll see in CI!)

My vscode `source.organizeImports` has added some noise to the diffs. If that's unwanted, I can revert the imports back to their original order.


Westbrook:
@cuberoot as you lead the investigation that lead us to the original solution here, could you take a look at this update and share your thoughts? I'll dive in more deeply on this tomorrow (I really want to find a way to visually regression test this), but if we can steal any of those New Zealand hours in the mean time, I'd be much obliged!

packages/tooltip/stories/tooltip.stories.ts
```diff
@@ -104,3 +103,54 @@ export const wIcon = (): TemplateResult => {
...
+        html,
+        body,
+        #root,
+        #root-inner,
+        sp-story-decorator {
+            height: 100%;
+            margin: 0;
+        }
+
+        sp-story-decorator::part(container) {
```
Westbrook, line 118: This is nice to see someone leverage!

Westbrook (approved):
This is looking good to me!

I'm not a huge fan of diverging from Spectrum CSS, but I think we have a solid case here to request they follow our lead here.

I'll wait to merge, just in case @cuberoot can think of any other edge cases we might not have covered here, but if not we'll be good to go. Definitely want to get this fix released, and then we can follow up with any regression testing we aren't able to come up with before Wednesday, later.

packages/popover/src/Popover.ts
```diff
@@ -50,24 +49,14 @@ export class Popover extends SpectrumElement {
             <div id="tip">
                 <svg
                     xmlns="http://www.w3.org/svg/2000"
-                    class="tip triangle-top-bottom"
+                    class="tip triangle-top"
```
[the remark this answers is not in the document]
hunterloftis: I think you're right - removing...

packages/tooltip/stories/tooltip.stories.ts
hunterloftis: tbh I didn't look at it very closely, I just noticed that Overlay was doing what I needed so I copy-pasted the styles!

Westbrook: OK, so you're saying I'm congratulating myself for using the thing I made for myself 😜

hunterloftis: I mean, it worked! :tada:

hunterloftis:
> I'm not a huge fan of diverging from Spectrum CSS, but I think we have a solid case here to request they follow our lead here.

The exact implementation is somewhat flexible; as long as `transform` isn't used on the top-level element of something that will be an `arrow`.

This PR doesn't change the DOM for Tooltip, and it makes the DOM for Popover a little closer to original Spectrum CSS. SWC could even use [these classes](https://opensource.adobe.com/spectrum-css/popover.html#popover-dialogstyle) directly now:

```html
<svg xmlns="http://www.w3.org/svg/2000" width="23" height="12" class="spectrum-Popover-tip" style="left: 201px;">
    <path class="spectrum-Popover-tip-triangle" d="M 0.7071067811865476 0 L 11.414213562373096 10.707106781186548 L 22.121320343559645 0"></path>
```

What I have the least confidence in with this change is its behavior across all the various permutations in which it could live. Chrome/FF/Safari * medium/large scale * dark/light/etc themes * static tooltip/popover vs overlay. I've been so impressed with how well Spectrum components have worked together even when I'm changing colors, sizes, browsers, etc, (minus the native button styles left in Safari!)... so if there weren't already an issue with tooltips I would be hesitant to propose the change.

cuberoot:
These changes look good to me. I am aware of the issue with the transform collision as I was trying to work around that. This seems like a promising solution. Let get it in and put it through its paces. Great to see code getting deleted too.

Westbrook:
Merging into a local branch. This should run the full CI for a final review. Thanks @hunterloftis for digging deep on this one for us!

Issue #1153, opened 2021-01-28
Add new component SplitView


spdev3000:
### Expected Behaviour
SWC should contain a component SplitView, reg. to the spectrum-css contribution https://opensource.adobe.com/spectrum-css/splitview.html

### Actual Behaviour
No component exists

PR #1154, opened 2021-01-28
feat: setup SplitView component

spdev3000:
This PR adds a new component SplitView to the spectrum-web-components. This is set as a Contribution from Spectrum-CSS at https://opensource.adobe.com/spectrum-css/splitview.html

## Description

This PR uses the spectrum css contribution to setup a new component, that shows a user two panels, that can be resized vertically or horizontally by dragging the splitter. It offers also a collapsible feature, that can be used to display the 1st or 2nd panel in full size and hide the other panel.

## Related Issue

https://github.com/adobe/spectrum-web-components/issues/1153

## Motivation and Context
This component is first needed by Adobe Ps project to enable the UI of layer properties.

## How Has This Been Tested?

A storybook, README and full test files (split-view.test.ts and benchmark tests) have been added to the PR. Current test coverage is 98.56% (lines).

## Types of changes

## Checklist:


Westbrook:
First note:

Please make sure that listings for this component are added to the `bundle` package. This is leveraged for both lazy consumers and the documentation site: https://601298f1879709fd35873b8b--spectrum-web-components.netlify.app/components/split-view notice how the element isn't currently upgrading. You can test this locally with `yarn docs:start`, and should see the element upgrade once you've included this as a dependency and listed it in the `elements.ts` and `src/index.ts` files appropriately.

packages/split-view/src/SplitView.ts
```diff
@@ -0,0 +1,416 @@
...
+        this.onPointerDragged = this.onPointerDragged.bind(this);
+        this.onPointerUp = this.onPointerUp.bind(this);
+        this.resize = this.resize.bind(this);
+        this.updatePosition = this.updatePosition.bind(this);
+        this.getSplitterSize = this.getSplitterSize.bind(this);
+    }
+
+    public connectedCallback(): void {
+        super.connectedCallback();
+        window.addEventListener('resize', this.resize);
```
Westbrook, line 142: I think `resizeObserver` fits out support matrix enough to leverage to more accurate specificity here over the use of window `resize`. Thoughts?

packages/split-view/stories/split-view.stories.ts
```diff
@@ -0,0 +1,205 @@
...
+import '../sp-split-view.js';
+
+export default {
+    title: 'Split View',
+    component: 'sp-split-view',
+};
+
+export const Horizontal = (): TemplateResult => {
+    return html`
+        <sp-split-view style="height: 200px" primary-default="100">
```
Westbrook, line 24: Does the use of `primary-default` here mean that the primary pane should be `100px` at runtime? I think there's a bug in `updatePosition()` that's reseting user input to the default if that's supposed to be the case.

packages/split-view/src/SplitView.ts
```diff
@@ -0,0 +1,416 @@
...
+        const dimension = this.vertical
+            ? ORIENTATIONS.vertical
+            : ORIENTATIONS.horizontal;
+
+        return html`
+            <div
+                class="pane"
+                style=${`${dimension}: ${this.dividerPosition}px`}
+            >
+                <slot name="primary"></slot>
```
Westbrook, line 160: I can be convinced that this is the right thing, but I'm not excited with how it complicates the external API here by adding two `slot`s for a user to think about. I've just pushed a branch with the approach I discussed last week `westbrook/splitview`. Thoughts?

packages/split-view/stories/split-view.stories.ts
```diff
@@ -0,0 +1,205 @@
...
+        </sp-split-view>
+    `;
+};
+
+export const HorizontalResizableCollapsible = (): TemplateResult => {
+    return html`
+        <sp-split-view
+            resizable
+            collapsible
+            primary-min="50"
```
Westbrook, line 64: When `collapsible` and `primary-min` are used together I feel a discernible jank in the pixels between the `min` and the `collapse`.

Is that just me? Probably. 

Should we do something to prevent these two values from being used together? Probably overkill.

Can you think of any sort of affordance we could make here to clarify to users the there's something _after_ the min?

Westbrook, line 64: Using these two values together also prevents the keyboard interface from reaching `collapsed`. Should we just skip from the `min` to `collapsed` in that case?

packages/split-view/src/SplitView.ts
```diff
@@ -0,0 +1,416 @@
...
+        if (this.hovered) {
+            if (this.primarySize !== undefined) {
+                return;
+            }
+            window.addEventListener(
+                'pointermove',
+                this.onPointerDragged,
+                false
+            );
+            window.addEventListener('pointerup', this.onPointerUp, false);
```
Westbrook, line 202: Can you talk to me about why this is on the `window`? Seems like with https://developer.mozilla.org/en-US/docs/Web/API/Element/setPointerCapture we'd be able to listen to this event on the same element as the `pointerdown` event.

packages/split-view/src/SplitView.ts
```diff
@@ -0,0 +1,416 @@
...
+
+    private onPointerMove(event: PointerEvent): void {
+        this.isOver = true;
+        if (this.dragging) {
+            return;
+        }
+        this.updateCursor(event);
+    }
+
+    private onPointerDown(event: PointerEvent): void {
```
Westbrook, line 189: Nit, this event is `pointerdown`. Please only camel case the `P`, both here and on others.

packages/split-view/src/SplitView.ts
```diff
@@ -0,0 +1,416 @@
...
+    private isOver = false;
+    private lastPosition = 0;
+
+    public constructor() {
+        super();
+        this.onPointerDragged = this.onPointerDragged.bind(this);
+        this.onPointerUp = this.onPointerUp.bind(this);
+        this.resize = this.resize.bind(this);
+        this.updatePosition = this.updatePosition.bind(this);
+        this.getSplitterSize = this.getSplitterSize.bind(this);
```
Westbrook, line 137: If we're able to remove the `window.addEventListener()` code based on notes below, I'd like to not be binding these event like this any more.

packages/split-view/src/SplitView.ts
```diff
@@ -0,0 +1,416 @@
...
+    public isCollapsedStart = this.dividerPosition === 0;
+
+    @property({ type: Boolean, reflect: true, attribute: 'is-collapsed-end' })
+    public isCollapsedEnd = false;
+
+    @property({ type: Boolean, reflect: true })
+    public dragging = false;
+
+    @property({ type: Boolean, reflect: true })
+    public hovered = false;
```
Westbrook, line 121: I'm not so happy with the way the upstream implementation forces the use of more complex JS to handle the styling of this. It's an internal implementation, however, so we can refactor this at a later date if you're comfortable with this approach.

Leveraging the matched `:hover` styles on `.spectrum-SplitView-splitter` we could do some special magic with:
```
:host(:not([vertical])) #splitter {
  padding: 0 10px; // this way we could also use Custom Properties to allow this math to be managed by the Scaling system
  margin: 0 -10px;
}
```
To expand the addressable area of the splitter, allowing CSS to chose when "hover" has occurred. This would mean we'd only need to manage what cursor to show, not whether to show it.

packages/split-view/src/SplitView.ts
Westbrook: With this approach we could also limit the calls to `getBoundingClientRect()` to only when the element is resized, rather than on every `onPointerMove()` like we have now.

Westbrook (changes requested):
For such a "simple" looking element, there really is lots of cool stuff going on here internally. This is a great first pass! Some of my notes my alter the code flows a good bit, so I wanted to get this back to you before I dug into it too much more. Let me know if you'd prefer to chat on any of this in person.

Westbrook:
Also, once we've got the stories locked down the way we'd like, we'll need to add them to the visual regression list in [`test/visual/stories.js`](https://github.com/adobe/spectrum-web-components/blob/main/test/visual/stories.js) as per https://github.com/adobe/spectrum-web-components#screenshot-testing

packages/split-view/src/SplitView.ts
```diff
@@ -0,0 +1,416 @@
...
+const SPLITTERSIZE = 2;
+
+/**
+ * @element sp-split-view
+ */
+export class SplitView extends SpectrumElement {
+    public static get styles(): CSSResultArray {
+        return [
+            styles,
+            css`
```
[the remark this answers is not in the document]
spdev3000: done in 7be37f0a7a62320e36fcef7575868809b961766e

packages/split-view/src/SplitView.ts
```diff
@@ -0,0 +1,416 @@
...
+                .pane {
+                    overflow: hidden;
```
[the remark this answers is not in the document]
spdev3000: done in 7be37f0a7a62320e36fcef7575868809b961766e

packages/split-view/stories/split-view.stories.ts
spdev3000: Couldn't find any bug for that, I test with various values without breaking the behavior.

packages/split-view/src/SplitView.ts
spdev3000: I used your slot css solution now, works perfect :-)

packages/split-view/stories/split-view.stories.ts
spdev3000: I agreed thats misleading the user, I removed now the attribute `collapsible` and changed the behavior that - if no min/max values are given and `resizable=true` then the SplitView can be collapsed in both start/end positions. If min/max values are given, it's not possible to drag the splitter to either end. 
We might discuss and extend this later, if we need a collapsed and animated behavior.

packages/split-view/src/SplitView.ts
spdev3000: Use setPointerCapture now (fixed in 7be37f0a7a62320e36fcef7575868809b961766e)

packages/split-view/src/SplitView.ts
spdev3000: Removed all window.addEventListener() completely.

packages/split-view/src/SplitView.ts
spdev3000: I tried so, but could not successfully use this pattern as due to the bg-color the splitter gets 10px width colored. Maybe we can come up with another solution later.

packages/split-view/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "scripts": {
+        "test": "echo \"Error: run tests from mono-repo root.\" && exit 1"
+    },
+    "author": "",
+    "license": "Apache-2.0",
+    "devDependencies": {
+        "@spectrum-css/splitview": "^3.0.0-beta.5"
+    },
+    "dependencies": {
+        "@spectrum-web-components/base": "^0.1.0",
```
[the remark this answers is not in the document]
spdev3000: done

packages/split-view/src/SplitView.ts
spdev3000: I setup resizeObserver, it works well but still need to fight the TS issues...

packages/split-view/stories/split-view.stories.ts
spdev3000: resolved this now by removing the `collapsible` flag.

packages/split-view/src/SplitView.ts
spdev3000: fixed

packages/split-view/stories/split-view.stories.ts
Westbrook: Yep, no longer defaulting to 403px, or whatever it was!

packages/split-view/src/SplitView.ts
Westbrook: Sorry, I forgot the most important part of the CSS above:
```
background-clip: content-box;
```
Then the background color should no longer grow that way 🤞🏼

packages/split-view/README.md
```diff
@@ -0,0 +1,156 @@
...
+import { SplitView } from '@spectrum-web-components/split-view';
+```
+
+### Variants
+
+#### Horizontal
+
+```html
+<sp-split-view>
+    <div slot="primary">Left panel</div>
```
Westbrook, line 30: Make sure we update this to the all default slot API.

spdev3000: Done, removed it from README, stories and tests.

packages/split-view/src/SplitView.ts
spdev3000: it works!

spdev3000:
> Also, once we've got the stories locked down the way we'd like, we'll need to add them to the visual regression list in [`test/visual/stories.js`](https://github.com/adobe/spectrum-web-components/blob/main/test/visual/stories.js?rgh-link-date=2021-01-28T14%3A44%3A07Z) as per [adobe/spectrum-web-components#screenshot-testing](https://github.com/adobe/spectrum-web-components#screenshot-testing)

I setup those 7 visual stories as a first suggestion.

packages/split-view/README.md
```diff
@@ -0,0 +1,139 @@
+## Description
```
Westbrook, line 1: We need some sort of intro to the pattern. Sadly, neither CSS or the specification outline a place to start here. How does this maybe sound to start?

> An `<sp-split-view>` element delivers its first two direct child elements in a horizontal or vertical (`<sp-split-view vertical>`) orientation the distributes the available page real estate as per the supplied attribute API. When leveraging the `resizable` attribute a mouse and keyboard accessible affordance is provided to the user to customize the distribution of that area between the available children.

packages/split-view/package.json
```diff
@@ -0,0 +1,56 @@
...
+        "./sp-*.js",
+        "./sp-*.ts"
+    ],
+    "scripts": {
+        "test": "echo \"Error: run tests from mono-repo root.\" && exit 1"
+    },
+    "author": "",
+    "license": "Apache-2.0",
+    "devDependencies": {
+        "@spectrum-css/splitview": "^3.0.0-beta.5"
```
Westbrook, line 50: There'll be a tiny race between this and #1169 as to the version that needs to be here. I'm not sure this PR can be updated without other updates in that PR, but the initial release of this pattern should use the `@3.0.0` release of the Spectrum CSS output. No change needed right now, but wanted to make sure I said this out loud.

packages/split-view/src/split-view.css
```diff
@@ -0,0 +1,148 @@
...
+}
+
+:host([vertical][resizable]) #splitter {
+    padding: var(--spectrum-global-dimension-static-size-125) 0;
+    margin: calc(-1 * var(--spectrum-global-dimension-static-size-125)) 0;
+    background-clip: content-box;
+    cursor: ns-resize;
+}
+
+:host(:not([vertical])[resizable]) #splitter {
```
Westbrook, line 56: Not sure if it's a nit or a good thing, but if you reverse the order of these rules, you should be able to forgo the `:not()` modifier/complexity by assuming the "horizontal" layout is the default?

```
host([resizable]) #splitter { ... }
:host([vertical][resizable]) #splitter { ... }
```

If that's helpful, should apply to the rest of the rules in this file, as well.

This file also gets process for nested selectors (though editor linting might not like it 🤔 ), if you felt approaching this in that way would make it easier to read/manage long term.

I'll open a conversation with the CSS team about getting them interested in adopting these changes so we can hopefully move them out of our code all together, too.

packages/split-view/stories/split-view.stories.ts
```diff
@@ -0,0 +1,257 @@
...
+        </sp-split-view>
+    `;
+};
+
+export const HorizontalResizable = (): TemplateResult => {
+    return html`
+        <sp-split-view
+            resizable
+            primary-min="50"
+            primary-max="Infinity"
```
Westbrook, line 36: Same suggestion as the docs about the use of `Infinity`.

packages/split-view/stories/split-view.stories.ts
```diff
@@ -0,0 +1,257 @@
...
+                    sites still in their infancy. Various versions have evolved
+                    over the years, sometimes by accident, sometimes on purpose
+                    (injected humour and the like).
+                </p>
+            </div>
+        </sp-split-view>
+    `;
+};
+
+export const MultipleLevels = (): TemplateResult => {
```
Westbrook, line 201: This story is so great!! 🔥

packages/split-view/test/split-view.test.ts
```diff
@@ -0,0 +1,728 @@
...
+            `
+        );
+
+        await elementUpdated(el);
+        expect(el).to.be.accessible();
+        expect(el.primaryDefault).to.equal(100);
+        const gripper = el.shadowRoot.querySelector(
+            '#gripper'
+        ) as HTMLDivElement;
+        expect(gripper).not.to.be.accessible();
```
Westbrook, line 46: What about the gripper is "not accessible"? I'm unsure what this is testing.

Westbrook: Oh, actually, I see that it's not testing anything, really. All of the `.accessible()` tests need to be `await`ed or you're not actually getting results.

Westbrook (changes requested):
This is really starting to come together! One thing that I'm not seeing right now is reactivity to changes in `primary-max`. For instance if I visit the `split-view--multiple-levels` story and set `el.primaryMax = 100` it doesn't seem to prevent me from freely changing the size of the content areas. Is that expected?

packages/split-view/src/SplitView.ts
```diff
@@ -0,0 +1,324 @@
...
+        this.resize();
+    }
+
+    protected updated(changed: PropertyValues): void {
+        super.updated(changed);
+
+        if (changed.has('dividerPosition')) {
+            this.paneSlot.style.setProperty(
+                '--spectrum-split-view-first-pane-size',
+                `${this.dividerPosition}px`
```
Westbrook, line 320: We need to round this number to whole numbers at some point in the process to prevent subpixel layout from bubbling up in the UI across various screen resolutions. See here the handle is no longer "centered": 
image

packages/split-view/README.md
```diff
@@ -0,0 +1,139 @@
...
+</sp-split-view>
+```
+
+#### Horizontal Resizable
+
+```html
+<sp-split-view
+    resizable
+    primary-min="50"
+    primary-max="Infinity"
```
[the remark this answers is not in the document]
spdev3000: Agree, will use 3840 (4k) as default_max_size.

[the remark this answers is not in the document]
Westbrook: Probably fine either way.

My point here is that a max that could be experienced on the page, like `primary-max="400"`, would allow the user to see that it actually does something, e.g. prevents the main pane from getting larger than 400px.

packages/split-view/global.d.ts
```diff
@@ -0,0 +1,223 @@
+/* eslint-disable */
+declare class ResizeObserver {
```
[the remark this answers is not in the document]
spdev3000: Isn't this included by
```
"files": [
        "custom-elements.json",
        "*.d.ts",
        ...
```
?

packages/split-view/src/SplitView.ts
```diff
@@ -0,0 +1,324 @@
...
+    @property()
+    public label?: string;
+
+    @query('slot')
+    private paneSlot!: HTMLSlotElement;
+
+    @query('#splitter')
+    private splitter!: HTMLDivElement;
+
+    public get size(): number {
```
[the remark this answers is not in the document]
spdev3000: Agree, I renamed to `offsetValue`

packages/split-view/global.d.ts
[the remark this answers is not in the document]
Westbrook: Oh, indeed. Thanks!

packages/split-view/src/SplitView.ts
```diff
@@ -0,0 +1,324 @@
...
+            this.dividerPosition = pos;
+        }
+        this.isResizedStart = this.dividerPosition === this.minPos;
+        this.isResizedEnd = this.dividerPosition === this.maxPos;
+        this.isCollapsedStart = this.dividerPosition === 0;
+        this.isCollapsedEnd =
+            this.dividerPosition >= this.size - this.getSplitterSize();
+    }
+
+    public getSplitterSize(): number {
```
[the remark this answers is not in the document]
spdev3000: yes, I used a get function with private Variable _splitterSize to achieve this.

packages/split-view/src/SplitView.ts
```diff
@@ -0,0 +1,324 @@
...
+        if (!this.resizable || this.primarySize !== undefined) {
+            return;
+        }
+        switch (event.key) {
+            case 'ArrowLeft':
+                this.movePosition(
+                    event,
+                    this.isLTR || this.vertical
+                        ? -1 * ARROW_KEY_CHANGE_VALUE
+                        : ARROW_KEY_CHANGE_VALUE
```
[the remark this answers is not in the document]
spdev3000: Hmm, moving it completely to `movePosition()` would make that function rather complex as we have 3 different use cases: Horizontal+LTR, Horizontal+RTL and Vertical. Each key event is slightly different for each use case. I could create a switch block for each of them (for clarity reason), but that would triple the switch/case construct. Not sure if this is better though.
Moving it to movePosition would require to decide on that conditions:
- which use case for what key => amount increase/decrease

packages/split-view/src/SplitView.ts
```diff
@@ -0,0 +1,324 @@
...
+            (event.button && event.button !== 0)
+        ) {
+            return;
+        }
+        this.splitter.setPointerCapture(event.pointerId);
+        this.onpointermove = this.onPointerdragged;
+        this.offset = this.getOffset();
+    }
+
+    private onPointerdragged(event: PointerEvent): void {
```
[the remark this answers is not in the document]
spdev3000: Agree

packages/split-view/README.md
spdev3000: Yes, I extended it a bit: 

> An <sp-split-view> element delivers its first two direct child elements in a horizontal or vertical (<sp-split-view vertical>) orientation that distributes the available page real estate as per the supplied attribute API. When leveraging the resizable attribute a pointer and keyboard accessible affordance is provided to the user to customize the distribution of that area between the available children. Providing minimum and maximum panel size values ("primary" and "secondary") disables the collapsing state. This state would allow the user to display one panel at a time completely.

packages/split-view/src/SplitView.ts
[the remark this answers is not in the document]
Westbrook: Ah, I look at this and the way the shape repeats and I see it as the same, lazy developers 😞 ... how do you feel about something like this instead:
```
let direction = 0;
const isLTRorVertical = this.isLTR || this.vertical;
const isRTLorVertical = !this.isLTR || this.vertical;
switch (event.key) {
    case 'Home':
        event.preventDefault();
        this.updatePosition(this.minPos);
        return;
    case 'End':
        event.preventDefault();
        this.updatePosition(this.maxPos);
        return;
    case 'Left':
    case 'ArrowLeft':
        direction = isLTRorVertical ? -1 : 1;
        break;
    case 'Right':
    case 'ArrowRight':
        direction = isLTRorVertical ? 1 : -1;
        break;
    case 'Up':
    case 'ArrowUp':
        direction = isRTLorVertical ? -1 : 1;
        break;
    case 'Down':
    case 'ArrowDown':
        direction = isRTLorVertical ? 1 : -1;
        break;
    case 'PageUp':
        direction = isRTLorVertical ? -1 : 1;
        break;
    case 'PageDown':
        direction = isRTLorVertical ? 1 : -1;
        break;
}
if (direction !== 0) {
    const moveBy = event.key.startsWith('Page')
        ? PAGEUPDOWN_KEY_CHANGE_VALUE
        : ARROW_KEY_CHANGE_VALUE;
    this.movePosition(event, moveBy * direction);
}
```

packages/split-view/README.md
Westbrook: I like the information you're adding here, but I'm not sure it's quite clear yet. @najikahalsema as a bit of a third party here, can you share any thoughts you have?

spdev3000: Yes, I am also not 100% sure if the last 2 sentences are clear about the intention. So to say if you really *want* to have "collapsed" panels, either remove all min/max values or set min to `0` and max to max-width of the whole component. Or should this be the default and doesn't need any additional explanation?

packages/split-view/README.md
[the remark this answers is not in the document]
spdev3000: Hm, can you please explain where it wasn't happen for you? Tried to find why it may not working as expected with -max values. It worked for me so far.

packages/split-view/src/SplitView.ts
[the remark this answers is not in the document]
spdev3000: Agreed!

packages/split-view/src/split-view.css
spdev3000: Found a better solution here.

packages/split-view/test/split-view.test.ts
spdev3000: Did you found out if we really need the "await" of if we can follow the linter recommendations to remove it?

packages/split-view/src/SplitView.ts
spdev3000: Fixed

packages/split-view/stories/split-view.stories.ts
spdev3000: replaced Infinity with real number

packages/split-view/src/SplitView.ts
```diff
@@ -0,0 +1,419 @@
...
+        super.connectedCallback();
+        this.observer?.observe(this);
+    }
+
+    public disconnectedCallback(): void {
+        this.observer?.unobserve(this);
+        super.disconnectedCallback();
+    }
+
+    public get viewSize(): number {
```
Westbrook, line 140: I think this is where the visual regressions are failing. Try adding the following and running them again: 
```suggestion
    /**
     * @private
     **/
    public get viewSize(): number {
```

packages/split-view/README.md
Westbrook: Let's remove `Providing minimum and maximum panel size values ("primary" and "secondary") could disable the collapsing state. This state would allow the user to display one panel at a time completely.` and come back to this in a later PR.

spdev3000: Done!

packages/split-view/stories/split-view.stories.ts
```diff
@@ -0,0 +1,272 @@
...
+                    </div>
+                </sp-split-view>
+            </div>
+        </sp-split-view>
+    `;
+};
+
+export const OnePaneNoSplitter = (): TemplateResult => {
+    return html`
+        <sp-split-view style="height: 200px"></sp-split-view>
```
Westbrook, line 260: Is this intended? Seems like this should have a single `div` (or similar) with some content in it based on that name?

spdev3000: There was indeed a inner <div> missing, I fixed this.

packages/split-view/package.json
spdev3000: done with latest rebase, new: 
```
"devDependencies": {
        "@spectrum-css/splitview": "^3.0.0"
```

packages/split-view/src/SplitView.ts
```diff
@@ -0,0 +1,429 @@
...
+    /** The start size of the primary pane */
+    @property({ type: String, attribute: 'primary-size' })
+    public primarySize?: string;
+
+    /** The minimum size of the secondary pane */
+    @property({ type: Number, attribute: 'secondary-min' })
+    public secondaryMin = 0;
+
+    /** The maximum size of the secondary pane */
+    @property({ type: Number, attribute: 'secondary-max' })
```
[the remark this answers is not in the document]
Westbrook: Particularly outlining when properties can be `123px` and `12%` and `auto` etc.

packages/split-view/src/spectrum-split-view.css
```diff
@@ -0,0 +1,364 @@
...
+    --spectrum-splitview-vertical-gripper-outer-width: 100%;
+    --spectrum-splitview-vertical-gripper-reset: 0;
+    display: flex;
+    overflow: hidden;
+}
+::slotted(*) {
+    /* .spectrum-SplitView-pane */
+    height: 100%;
+}
+:host([dir='ltr']) #gripper {
```
Westbrook, line 26: I feel like this matches https://github.com/adobe/spectrum-web-components/pull/1154/files#diff-21070cfdf1a047d11b4140e4f2838f99b136ad0fcfa5da0efa8b1bf6b2b43652R109, can you confirm that we need the custom styles, too.

Westbrook: It feels like we may be doing this in a couple of places in the CSS of `split-view.css`. To clarify, could we maybe add some comments to the customization to clarify why they are there?

Westbrook: All of the measurements that we persist in `split-view.css` may also need updated CSS Custom Property names.

packages/split-view/stories/split-view.stories.ts
```diff
@@ -0,0 +1,274 @@
...
+        </sp-split-view>
+    `;
+};
+
+export const ShowFirstTwoPanes = (): TemplateResult => {
+    return html`
+        <sp-split-view style="height: 200px">
+            <div>First panel</div>
+            <div>Second panel</div>
+            <div>Third (invisible) panel</div>
```
Westbrook, line 271: Great add here!

Westbrook (commented):
I'm very hopeful that this is my last round of notes before being able to share with the Slack channel.

packages/split-view/src/spectrum-split-view.css
spdev3000: We need both matchers, as we have a custom splitter size (with margin+padding to allow hover states) and the difference is the first offset: `var(--spectrum-global-dimension-static-size-125) +` - I can add some comment in the custom css.

packages/split-view/stories/split-view.stories.ts
spdev3000: Thx

packages/split-view/src/spectrum-config.js
```diff
@@ -0,0 +1,62 @@
...
+                {
+                    type: 'boolean',
+                    selector: '.spectrum-SplitView--vertical',
+                    name: 'vertical',
+                },
+                {
+                    type: 'boolean',
+                    selector: '.is-collapsed-start',
+                    name: 'is-collapsed-start',
+                },
```
Westbrook, line 36: If you don't want to manage `[is-collapsed-start]` on the `<sp-split-view>` element, exclude this configuration and update the internals to add these classes to `id="splitter"` in the rendered template.

packages/split-view/src/SplitView.ts
```diff
@@ -0,0 +1,429 @@
...
+            <slot
+                @slotchange=${() => {
+                    this.enoughChildren = this.children.length > 1;
+                    this.checkResize();
+                }}
+            ></slot>
+            ${this.enoughChildren
+                ? html`
+                      <div
+                          id="splitter"
```
Westbrook, line 183:
```suggestion
                          id="splitter"
                          class="${this.splitterPos === 0 ? 'start' : ''} ${this.splitterPos >= this.viewSize - this.splitterSize ? 'end' : ''}"
```

packages/split-view/src/split-view.css
```diff
@@ -0,0 +1,176 @@
...
+
+:host([vertical][resizable][collapsible][is-collapsed-end]) #splitter {
+    cursor: n-resize;
+}
+
+:host([vertical][resizable][collapsible][is-resized-start]) #splitter,
+:host([vertical][resizable][collapsible][is-resized-end]) #splitter {
+    cursor: ns-resize;
+}
+
```
Westbrook, line 108: Comment how we are covering over the gripper intractable size changes we made.

Westbrook: Alternative: 
```
transform: translateX(var(--spectrum-global-dimension-static-size-125));
```

packages/split-view/src/SplitView.ts
```diff
@@ -0,0 +1,429 @@
...
+    public isResizedStart = false;
+
+    @property({ type: Boolean, reflect: true, attribute: 'is-resized-end' })
+    public isResizedEnd = false;
+
+    @property({ type: Boolean, reflect: true, attribute: 'is-collapsed-start' })
+    public isCollapsedStart = false;
+
+    @property({ type: Boolean, reflect: true, attribute: 'is-collapsed-end' })
+    public isCollapsedEnd = false;
```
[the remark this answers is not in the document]
spdev3000: Decided not to allow users to set these properties, moved to css classes only.

packages/split-view/src/SplitView.ts
```diff
@@ -0,0 +1,429 @@
...
+            `${x}px`
+        );
+    }
+
+    private setFirstPaneAutoSize(): void {
+        this.paneSlot.style.setProperty(
+            '--spectrum-split-view-first-pane-size',
+            'auto'
+        );
+    }
```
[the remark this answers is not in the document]
spdev3000: Yes, implemented this.

packages/split-view/src/spectrum-split-view.css
spdev3000: I tried out the translateX/Y css function, but that needed to much restructuring, so I re-formatted and added further comments about the additional added offset.

packages/split-view/src/spectrum-config.js
spdev3000: removed that

packages/split-view/src/SplitView.ts
spdev3000: Done!

packages/split-view/src/split-view.css
spdev3000: Add comment to css classes

packages/split-view/src/SplitView.ts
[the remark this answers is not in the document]
spdev3000: Done

packages/split-view/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "scripts": {
+        "test": "echo \"Error: run tests from mono-repo root.\" && exit 1"
+    },
+    "author": "",
+    "license": "Apache-2.0",
+    "devDependencies": {
+        "@spectrum-css/splitview": "^3.0.0"
+    },
+    "dependencies": {
+        "@spectrum-web-components/base": "^0.3.1",
```
Westbrook, line 55: This bumped under you feet to `^0.3.3`. I wonder if our tooling would auto-manage this on release, but I've never tested it.

najikahalsema:
This is looking really great!

packages/split-view/src/SplitView.ts
```diff
@@ -0,0 +1,410 @@
...
+
+    private async calcStartPos(): Promise<number> {
+        if (
+            this.primarySize !== undefined &&
+            /^\d+(px)?$/.test(this.primarySize)
+        ) {
+            return parseInt(this.primarySize, 10);
+        }
+        if (this.primarySize?.endsWith('%')) {
```
Westbrook, line 362: What happens if I put `pirmary-size="something%"`?

Westbrook (commented):
Some last tweaks/questions before I'm ready to merge and start doing some test releases!

packages/split-view/src/SplitView.ts
```diff
@@ -0,0 +1,410 @@
...
+    @property({ type: Number, attribute: 'primary-max' })
+    public primaryMax = DEFAULT_MAX_SIZE;
+
+    /**
+     * The start size of the primary pane, can be a real pixel number|string, percentage or "auto"
+     * For example: "100", "120px", "75%" or "auto" are valid values
+     * @type {number | number + "px" | number + "%" | "auto"}
+     * @attr
+     */
+    @property({ type: String, attribute: 'primary-size' })
```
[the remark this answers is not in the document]
spdev3000: thx

packages/split-view/src/SplitView.ts
```diff
@@ -0,0 +1,410 @@
...
+    public get viewSize(): number {
+        return this.vertical ? this.offsetHeight : this.offsetWidth;
+    }
+
+    /**
+     * @private
+     **/
+    public get splitterSize(): number {
+        if (!this._splitterSize) {
+            const el = this.shadowRoot.querySelector(
```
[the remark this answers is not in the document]
spdev3000: sure, sry.

packages/split-view/package.json
spdev3000: increased to ^0.3.3

packages/split-view/src/SplitView.ts
```diff
@@ -0,0 +1,410 @@
...
+                    )) ||
+                SPLITTERSIZE;
+        }
+        return this._splitterSize;
+    }
+
+    protected render(): TemplateResult {
+        return html`
+            <slot
+                @slotchange=${() => {
```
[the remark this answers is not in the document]
spdev3000: done

packages/split-view/src/SplitView.ts
```diff
@@ -0,0 +1,410 @@
...
+                    this.checkResize();
+                }}
+                style="--spectrum-split-view-first-pane-size: ${this
+                    .firstPaneSize}"
+            ></slot>
+            ${this.enoughChildren
+                ? html`
+                      <div
+                          id="splitter"
+                          class="${this.splitterPos === this.minPos
```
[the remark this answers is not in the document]
spdev3000: implemented, added further tests preventing to have *-start and *-end classes at the same position.

packages/split-view/src/SplitView.ts
spdev3000: added another regEx here

packages/split-view/src/SplitView.ts
```diff
@@ -0,0 +1,410 @@
...
+const SPLITTERSIZE = 2;
+
+const ARROW_KEY_CHANGE_VALUE = 10;
+
+const PAGEUPDOWN_KEY_CHANGE_VALUE = 50;
+
+const COLLAPSE_THREASHOLD = 50;
+
+/**
+ * @element sp-split-view
```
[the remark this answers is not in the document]
spdev3000: Added

packages/split-view/test/split-view.test.ts
```diff
@@ -0,0 +1,1035 @@
...
+
+        await elementUpdated(el);
+        await expect(el).to.be.accessible();
+        expect(el.splitterPos || 0).to.equal(100);
+        expect(el.resizable).to.be.false;
+        expect(el.collapsible).to.be.false;
+        const gripper = el.shadowRoot.querySelector(
+            '#gripper'
+        ) as HTMLDivElement;
+        // expect(gripper).not.to.be.accessible();
```
Westbrook, line 49: Remove.

packages/split-view/test/split-view.test.ts
```diff
@@ -0,0 +1,1035 @@
...
+        );
+
+        await elementUpdated(el);
+        await expect(el).to.be.accessible();
+        expect(el.splitterPos || 0).to.equal(100);
+
+        const gripper = el.shadowRoot.querySelector(
+            '#gripper'
+        ) as HTMLDivElement;
+        expect(gripper).to.be.accessible();
```
Westbrook, line 74: `await` or if the linter complains move to something like `.to.not.be.null`

Westbrook (commented):
Saw we had one last open comment and then I'm ready to approve.

packages/split-view/test/split-view.test.ts
spdev3000: I've added `await` now to all of this `to.be.accessible();`

packages/split-view/test/split-view.test.ts
spdev3000: done

packages/split-view/test/split-view.test.ts
spdev3000: awaited all

packages/split-view/test/split-view.test.ts
```diff
@@ -1001,7 +1000,7 @@ describe('SplitView', () => {
             splitter = el.shadowRoot.querySelector(
                 '#splitter'
             ) as HTMLDivElement;
-            expect(splitter).to.be.accessible;
+            await expect(splitter).to.be.accessible;
```
Westbrook, line 1003: Oh, I missed this one... `.accessible()` is a method to be called.

spdev3000: fixed it.

Westbrook (approved):
Thanks @spdev3000 for the hard work on this, looks good to me!

I'll leave this open a little longer, just in case anyone else takes a peak, but will start doing test release work today.

Westbrook:
@spdev3000 missed that there were some branch conflicts here. Once this is cleared up, I look forward to merging and releasing this new package!

spdev3000:
> @spdev3000 missed that there were some branch conflicts here. Once this is cleared up, I look forward to merging and releasing this new package!
@Westbrook I git rebased main and fixed the conflicts in bundle/package.json

Westbrook (approved):

PR #1167, opened 2021-02-03
fix: reduce cycles

Westbrook:
## Description
Looks like this was the only thing that cycled and then _everything_ cycled. We had removed the use of this dependency a while back for TS cycles and forgot to remove the JS dependency.

## Motivation and Context
Actual cycles are bad.

## How Has This Been Tested?
`yarn lerna changed` no longer warns of cycles

## Types of changes

## Checklist:

### DIFF
--- a/packages/base/package.json
+++ b/packages/base/package.json
@@ -44,7 +44,6 @@
     "author": "",
     "license": "Apache-2.0",
     "dependencies": {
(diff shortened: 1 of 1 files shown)


adixon-adobe (approved):
Pretty straightforward...

PR #1185, opened 2021-02-09
[rfc] fix(icons-workflow): rename icons/files to avoid ad blocking

### LINKED ISSUE #1160: Some SWC modules blocked by ad blockers
Full disclaimer - Yes, I know this isn't really a bug. But on the other hand, it did stump someone on my team for a bit and might be useful to take care of in case anyone else gets stumped.

### Expected Behaviour
Spectrum modules should all load in a buildless dev environment (@web/dev-server)

### Actual Behaviour
1 or 2 specific modules are blocked

To explain - there are 1 or 2 files in the `icons-workflow` package that seem to trigger ad blockers just by virtue of their name. The first is `AdDisplay.js` and I would assume that `AdPrint.js` would also be problematic.

My team member actually forgot he was using an ad blocker, but just got the following error (which wasn't that helpful for him to figure this out):

`GET https://localhost:3000/node_modules/@spectrum-web-components/icons-workflow/lib/icons/AdDisplay.js net::ERR_BLOCKED_BY_CLIENT`

Unfortunately, the net result is that our entire app wasn't functioning (just a white page), and I don't think we even have this module as a direct dependency, and it looked like the build system was broken.

Perhaps a simple name change on these 2 files would avoid the hassle for folks in the future?

Westbrook:
## Description
Prevent ad blocking extensions from blocking unbuilt SWC code by renaming `AdPrint` to `PrintAdvert` and `AdDisplay` to `DisplayAdvert`. 

This is a "breaking" change, and I'm also unsure whether it goes far enough for ALL ad blockers and would appreciate feedback on whether there's a better name to use, whether it matters, whether we should do something else...

## Related Issue
fixes #1160

## Motivation and Context
Unbuilt apps should not be blocked by ad blockers.

## How Has This Been Tested?
Locally with adblockplus.org installed in my browser.

## Types of changes

## Checklist:

### DIFF
--- a/packages/icons-workflow/bin/build.js
+++ b/packages/icons-workflow/bin/build.js
@@ -69,10 +69,14 @@ glob(`${rootDir}/node_modules/${iconsPath}/**.svg`, (err, icons) => {
 
     icons.forEach((i) => {
         const svg = fs.readFileSync(i, 'utf-8');
-        const id = path
+        let id = path
             .basename(i, '.svg')
             .replace('S_', '')
             .replace('_22_N', '');
+        if (id.search(/^Ad[A-Z]/) !== -1) {
+            id = id.replace(/^Ad/, '');
+            id += 'Advert';
+        }
         const ComponentName = id === 'github' ? 'GitHub' : Case.pascal(id);
         const $ = cheerio.load(svg, {
             xmlMode: true,


najikahalsema (dismissed):
At last, we shall finally be free

Issue #1189, opened 2021-02-11
Add more menu selection customizability


adixon-adobe:
This is a proposal to better support multiselect and un-selectable menu elements, and allow structured sub-handling in menu groups when necessary. It's modeled after the selects attribute already in sp-action-group. Creating this after a discussion with @Westbrook, and this is intended to consolidated/close issues #350 & #715

### multiselect

With selects="multiple" any number of the items in the menu can be selected, and activating a menu item will toggle it's selected state.

```html
<sp-action-menu selects="multiple">
	<sp-menu-item selected>item 1</sp-menu-item>
	<sp-menu-item>item 2</sp-menu-item>
	<sp-menu-item selected>item 3</sp-menu-item>
</sp-action-menu>
```

### single

With selects="single" zero or 1 item can be selected.

```html
<sp-action-menu selects="single">
	<sp-menu-item selected>item 1</sp-menu-item>
	<sp-menu-item>item 2</sp-menu-item>
	<sp-menu-item>item 3</sp-menu-item>
</sp-action-menu>
```

### no selection (BREAKING CHANGE)

When the `selects` attribute is ommitted in sp-action-menu, items are not selectable. This better matches the general intent of action menu, where most items are actions that shouldn't be selected (e.g. `File -> save...` isn't an item you'd expect to be selected)

```html
<sp-action-menu>
	<sp-menu-item>item 1</sp-menu-item>
	<sp-menu-item>item 2</sp-menu-item>
	<sp-menu-item>item 3</sp-menu-item>
</sp-action-menu>
```

## sp-picker & sp-split-button are single-select

There is no `selects` attribute for these controls, as they're designed to have a single value selected.

## sp-menu-group selects attribute

To allow for more granular selection handling, `sp-menu-group` also can provide a `selects` attribute, in which case it takes over managing selection for that sub menu. When present, the parent control is not notified of selection changes.

### Example 1

This example shows an action menu that has both a multi-select and single-select group, along with an unselectable group. The `sp-menu-group` elements with the `selects` attribute present should fire a change event when their selection state changes, while the action menu would not as the handling is managed by the menu group.

```html
<sp-action-menu>
	<sp-menu-group selects="multiple">
		<sp-menu-item selected>multiitem 1</sp-menu-item>
		<sp-menu-item>multiitem 2</sp-menu-item>
		<sp-menu-item selected>multiitem 3</sp-menu-item>
	</sp-menu-group>
	<sp-menu-group selects="single">
[...]

PR #1198, opened 2021-02-19
1189/menu removal spike

adixon-adobe:
## Description

This is just a quick spike on changing the re-parenting for picker so sp-menu isn't necessary.

## Related Issue

#1189 

## Motivation and Context

Not looking to merge this. Opening this to discuss the implications of this change.

### DIFF
--- a/packages/picker/src/Picker.ts
+++ b/packages/picker/src/Picker.ts
@@ -26,6 +26,7 @@ import pickerStyles from './picker.css.js';
 import chevronStyles from '@spectrum-web-components/icon/src/spectrum-icon-chevron.css.js';
 
 import { Focusable } from '@spectrum-web-components/shared/src/focusable.js';
+import ElementReparenter from '@spectrum-web-components/shared/src/element-reparenter.js';
 import '@spectrum-web-components/icon/sp-icon.js';
 import { Chevron100Icon } from '@spectrum-web-components/icons-ui';
 import '@spectrum-web-components/icons-workflow/icons/sp-icon-alert.js';
@@ -91,6 +92,7 @@ export class PickerBase extends SizedMixin(Focusable) {
     public open = false;
 
     private reparentableChildren?: Element[];
+    private elementReparenter?: ElementReparenter;
     public optionsMenu?: Menu;
 
     /**
@@ -117,7 +119,6 @@ export class PickerBase extends SizedMixin(Focusable) {
 
     protected listRole = 'listbox';
     protected itemRole = 'option';
-    private placeholderArray?: Comment[];
 
     public constructor() {
         super();
@@ -234,24 +235,10 @@ export class PickerBase extends SizedMixin(Focusable) {
 
     protected onOverlayClosed(): void {
         this.close();
-        if (
-            this.reparentableChildren &&
-            this.placeholderArray &&
-            this.reparentableChildren.length == this.placeholderArray.length
-        ) {
-            this.placeholderArray.forEach((item, index) => {
-                const parentElement = item.parentElement || item.getRootNode();
-
-                if (parentElement && this.reparentableChildren) {
-                    parentElement.replaceChild(
-                        this.reparentableChildren[index],
-                        item
-                    );
-                }
-            });
+        if (this.elementReparenter) {
+            this.elementReparenter.restore();
         }
-
-        delete this.placeholderArray;
+        this.elementReparenter = undefined;
 
         this.menuStateResolver();
     }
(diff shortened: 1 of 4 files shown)


packages/picker/src/Picker.ts
```diff
@@ -349,14 +368,17 @@ export class PickerBase extends SizedMixin(Focusable) {
...
+            </sp-popover>
         `;
     }
 
     protected firstUpdated(changedProperties: PropertyValues): void {
         super.firstUpdated(changedProperties);
 
-        this.optionsMenu = this.querySelector('sp-menu') as Menu;
+        this.optionsMenu = this.shadowRoot.querySelector('sp-menu') as Menu;
+        this.reparentableChildren = Array.from(this.children);
```
adixon-adobe, line 381: I think we'll want to keep the existing usage backward-compatible for now. So here we'd look for an `sp-menu` and just reparent the children of that (and I was thinking of adding a `console.warn` DEPRECATION notice for that).

adixon-adobe (commented):
To our discussion earlier about adding a mixin that manages the reparenting, I'm trying to think a bit about how that would work.

I could see doing this in a couple of ways, but am curious to get your thoughts.

```
    this.beginReparentContent(targetElements, newParent) // do the reparenting & add placeholders
    this.endReparentContent() // puts everything back
```

Though rather than a mixing, maybe just a class that manages the above would be simpler?

```
   this.reparentManager = new ReparenterManager()
   // same behavior as above here
   this.reparentManager.begin(targetElements, newParent)
   this.reparentManager.end()
```

Also, I'm definitely not naming things well today...

packages/picker/src/Picker.ts
```diff
@@ -253,26 +248,20 @@ export class PickerBase extends SizedMixin(Focusable) {
...
+            !this.reparentableChildren ||
+            this.reparentableChildren.length === 0
         ) {
             this.menuStateResolver();
             return;
         }
 
-        this.placeholder = document.createComment(
-            'placeholder for optionsMenu'
+        this.elementReparenter = new ElementReparenter(
```
Westbrook, line 258: Being you've chosen to make this an object oriented API would it make better sense to recycle the reparenter rather than newing it every time?

I could alternatively see it as a functional API, a la:
```
import { reparentChildren } from 'shared';

this.reclaimChildren = reparentChildren(children, target); 

if (this.reclaimElements) {
    this.reclaimChildren();
    this.reclaimChildren = undefined;
}
```
I know right, a web component admiring suggesting a functional API... 😏

packages/shared/src/element-reparenter.ts
```diff
@@ -0,0 +1,40 @@
+export default class ElementReparenter {
```
Westbrook, line 1: This seems to be on the right track. Would be interested in how this API supports `overlay-trigger` or not as a confirmation of the approach.

Westbrook (commented):
Beyond the functional failures, this branch has some new a11y failures (https://app.circleci.com/pipelines/github/adobe/spectrum-web-components/5481/workflows/768f1f48-9152-4f38-a045-01da666c3a26/jobs/45167/parallel-runs/0/steps/0-105) due to how `sp-menu-item` currently self applies role attributes 100% of the time: https://github.com/adobe/spectrum-web-components/blob/main/packages/menu/src/MenuItem.ts#L107 we'd need to address that as well to do forward with this approach. Interestingly enough this is managed in `sp-picker` and extensions at current, so you're already in the right place.

packages/picker/src/Picker.ts
Westbrook: In the realm of a class, you will find many parallels with reactive controller approach that the Polymer team is formalizing for the upcoming release of lit-element: https://github.com/Polymer/lit-html/blob/lit-next/packages/reactive-element/CHANGELOG.md

adixon-adobe: The functional approach definitely seems better here, and I even tried that first. I was passing back state and not a callback though which I didn't like.

packages/shared/src/element-reparenter.ts
adixon-adobe: I think it should be pretty easy to rework the core reparenting logic here to use the same helper: https://github.com/adobe/spectrum-web-components/blob/main/packages/overlay/src/ActiveOverlay.ts#L328

I also know we were talking about potentially supporting an element array for the overlay contents, but that looks a lot more disruptive (e.g. it looks like there's some `delayed` attribute handling on the content element).

Issue #1187, opened 2021-02-10
Make arrow keys move to next and previous item with one press in Picker (dropdown list)


velara3:
### Expected Behaviour
When a dropdown picker is focused I'd like to be able to select the next or previous items using the up and down arrows pressing them only one time. 

### Actual Behaviour
When the dropdown picker is focused and the menu options are not visible I press the up key or down key and the picker opens. Then on the second up or down key press the next or previous item is highlighted and then you must press the Enter key to select it (or click on the row). 

https://user-images.githubusercontent.com/1084127/107487650-3fa5b900-6b3b-11eb-962d-a635ff071c6b.mp4

### Reproduce Scenario (including but not limited to)

1. Go here:  https://opensource.adobe.com/spectrum-web-components/components/picker
2. Click on a picker so it has focus (and select something so the menu disappears again)
3. Press the up or down arrow key (the picker opens)
4. Press the up or down arrow key again (the next or previous row is highlighted)
5. Press the enter key to select the highlighted row

The behavior in the Picker is much better than the native dropdown but it is still horrible when you want to step through a long list of options and you can't use the arrow keys. 

For example, let's say you are in Photoshop and you want to change the blend mode of a layer. You have to open the dropdown and click on the row each time. The pop up menu items close each time you make a selection. 

Imagine how you would like to select each item in a list with the arrow keys and you can tell what I mean.  

### Platform
OSX using Chrome or Brave Browser

I don't think this is a Spectrum components problem but a problem with ALL dropdowns. The keyboard support is horrible at least on OSX (have not tried Windows in a while). 

Please consider it. Also, I would be satisfied if the behavior was enabled by flag or enabled by modifier key. For example, if you hold down ALT + Up arrow the next item is automatically selected (pop up doesn't open - simply the next item is selected).

Westbrook:
Thanks for this issue. Very glad to hear we've surpassed the native `<select>` element in your view, and sorry to hear we've not gotten all the way to full coverage of your needs just yet. Further, I can tell you that we'd not be able to support this feature directly, though I really appreciate your giving us some alternatives that you'd be interested in instead.

[...]

PR #1220, opened 2021-02-25
fix(picker): add "quick select" action to right/left arrow keys

### LINKED ISSUE #254: Cache a `selectedItem` in Dropdown
### Expected Behaviour
The `selectedItem` data shouldn't have to be acquired differently based on the workflow it is being used in. Let's cache this data point so that we can respond to _IT_ changing, as opposed to changing other things while querying this data as we go.

### Actual Behaviour
We query the `selectedItem` differently in `updated()` vs `setValueFromItem()`, let's save bytes by not doing that, and run time by caching an element to work with! 

cc: @cdvu1

Westbrook:
## Description
Add support for making a "quick selection" when an `sp-picker` element is focused via the right and left arrow keys.

Uses a cached `selectedItem` property to reduce work across the various interactions.

Check it out at: https://6037cd0dd41f080c0b2563ad--spectrum-web-components.netlify.app/components/picker

## Related Issue
fixes #254
refs #1187

## Motivation and Context
Make users lives easier and parity with other Spectrum component offerings.

## How Has This Been Tested?
Manually and unit tests.

## Types of changes

## Checklist:

### DIFF
--- a/packages/picker/src/Picker.ts
+++ b/packages/picker/src/Picker.ts
@@ -433,4 +433,42 @@ export class Picker extends PickerBase {
     public static get styles(): CSSResultArray {
         return [pickerStyles, chevronStyles];
     }
+
+    public onKeydown = (event: KeyboardEvent): void => {
+        const { code } = event;
+        if (!code.startsWith('Arrow')) {
+            return;
+        }
+        event.preventDefault();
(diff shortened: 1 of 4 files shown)


najikahalsema (approved):
Looks good to me!

PR #1227, opened 2021-02-27
fix(styles): move `@spectrum-web-components/base` from devDependencies to dependencies

charlessuh:
## Description

Compiled files like [spectrum-base.css.js](https://unpkg.com/browse/@spectrum-web-components/styles@0.8.0/src/spectrum-base.css.js) reference the `css` directive from the `@spectrum-web-components/base` package at runtime, so needs to be in `dependencies` section of the `package.json` (not `devDependencies`.)

## Motivation and Context

Package managers like yarn in strict mode will complain about not explicitly declaring all of your dependencies.

## Types of changes

## Checklist:

### DIFF
--- a/packages/styles/package.json
+++ b/packages/styles/package.json
@@ -45,7 +45,9 @@
     "devDependencies": {
         "@spectrum-css/commons": "^3.0.1",
         "@spectrum-css/typography": "^3.0.0",
-        "@spectrum-css/vars": "^3.0.0",
+        "@spectrum-css/vars": "^3.0.0"
+    },
+    "dependencies": {
         "@spectrum-web-components/base": "^0.3.3"
     }
 }


Westbrook:
Great catch!! One goal for this year is to have some example projects that depend up in the project and are attached to CI with different project shapes (bundler, package manager, framework) so we can catch these issues more quickly. That’s for submitting a fix. 

Can you rebase `main` onto this branch? The change looks good to me, there’s just one or two changes upstream that I want to make sure don’t get lost when we bring in the fix.

Westbrook:
Rebase will also ensure the unit tests run in CI. The VRTs are just about to get a revision, so we may move forward with this even if those have issue as they've been on their last legs this week.

Westbrook (approved):
LGTM.

PR #1231, opened 2021-03-01
build: use @web/test-runner-visual-regression instead of custom suite

Westbrook:
## Description
Update the visual regression tooling to leverage `@web/test-runner-visual-regression` instead of custom mocha/WDS/playwright code. 
- update storybook files that use non-exported assets from other packages to use relative paths
- refactor storybook-decorator to leverage sp-field-label

## Motivation and Context
- ci stability
- ci speed
- ability to expand to Firefox and Webkit in the future

## How Has This Been Tested?
It's tests. I've run this branch through it's passes, but also rebased a couple of projects to it and found that it works as expected there as well.

image

## Types of changes
- [x] Test tooling

## Checklist:


test/visual/story-imports.js
```diff
@@ -0,0 +1,92 @@
+/*
```
Westbrook, line 1: I'd target long term being able to build this file from content, but I've messed with this VRT thing a little too much and I didn't want to add any extra responsibilities to getting this change in at this time. Hopefully, it's already more clear than the stories array we were maintaining previously.

web-test-runner.utils.js
```diff
@@ -0,0 +1,112 @@
...
+                color,
+                scale,
+                dir,
+                reduceMotion,
+            });
+            vrtGroups.push({
+                name: `vrt-${color}-${scale}-${dir}`,
+                files: 'test/visual/test.js',
+                testRunnerHtml: testHTML,
+                browsers: [playwrightLauncher({ product: 'chromium' })],
```
Westbrook, line 71: Here's where we focus on just testing in Chormium. We could take this away and it would switch to include Firefox and Safari as well. Two things I noticed when doing so:
- CircleCI seems to scale the final "Uploading Artifacts" seems to scale less than linearly, so if we have two or three times as baseline images to cache the task starts to take a good deal longer. 
- The above along with the somewhat synchronous nature of our current test configuration (brought on by the available RAM and CPU in CircleCI) points to a need to fork the various extended browser coverage into distinct CI jobs to prevent the job time from swelling again. Relatively straight forward with the advent of test groups, but not included in this work.
- Firefox some how unhides hidden form elements when taking the snapshot, so things like the native checkbox are show through the checkbox and switch elements. This doesn't happen when viewing the pages in browser, so it would need a little more investigation before being fully useful.

web-test-runner.config.js
```diff
@@ -41,9 +44,33 @@ module.exports = {
...
+            timeout: 100000,
         },
     },
+    groups: [
+        {
+            name: 'unit',
+            files: 'packages/*/test/*.test.js',
+        },
+        ...vrtGroups,
+        ...packages.reduce((acc, pkg) => {
```
Westbrook, line 56: This allows us to call `yarn test:focus picker` and only run the tests for the `picker` package, which is quite nice.

README.md
```diff
@@ -85,22 +85,22 @@ During development you may wish to use `yarn test:watch` to automatically build
...
 
 To create a local baseline so that you can compare your changes to it later in the development cycle, use the following commands:
 
 ```bash
-yarn storybook:build # creates the test assets
-yarn test:visual:baseline:local --color=light --scale=medium
+yarn test:visual:clean # start with a clean slate
+# yarn test:visual:clean:baseline # removes only baseline images
+# yarn test:visual:clean:current # removes only images updated in the most recent test pass
+yarn test:visual vrt-light-medium-ltr # vrt-${color}-${scale}-${direction} to access all theme options
```
Westbrook, line 96: A more wholistic local testing scheme likely benefits from running all of the baselines, however that's a lot of work to have done on your local system... I think there's some more that can be done here, but I hope that this is a good enough baseline for now. Happy to add specific docs or scripts here if it's seen necessary.

Westbrook (commented):
A couple of pieces of follow up work that we should take on later:
- Covert the library to `"type": "module"` so we can use module syntax in all of our tools by default. We could make the switch and cuardon current tools off into `*.cjs` land, and then make the transition gradually while requiring modules for everything new.
- Covert what can be converted to TypeScript. JS is great for quick, prototype work, but most of pieces that can be TS here would likely benefit from doing so over time.

najikahalsema (approved):
Yeehaw! This looks great.

Issue #1243, opened 2021-03-03
SplitView: Correct auto height calculation when using primary-size="auto" 


spdev3000:
### Expected Behaviour
The height of the primary panel should be rounded up (ceiling) to the next full pixel value. And not shrank to the floor pixel value. 

### Actual Behaviour
When using primary-size="auto" the current height of the primary panel is computed. It gives back a float value. Current implementation converts that to integer, meaning it could potentially shrink the given height of the primary panel. As we use full pixel value internally to set the splitter we need to make sure the primary height won't be shrunk by only processing the full pixel amount.

### Reproduce Scenario (including but not limited to)
Checked with different scenarios, where primary panel height is not an integer value, e.g. `50.42px` - resulting start height of primary panel will be `50px`

#### Platform and Version
all

#### Sample Code that illustrates the problem
```
const size = window
    .getComputedStyle(firstEl)
    .getPropertyValue(this.vertical ? 'height' : 'width');
const size_i = parseInt(size, 10);
if (!isNaN(size_i)) {
    return this.getLimitedPosition(size_i);
}
```

PR #1244, opened 2021-03-03
[SplitView] Correct calculation of height when using primary-size='auto'

spdev3000:
## Description

This PR fix the calculation of initial height of primary panel when using primary-size="auto".
It replaces the parseInt function by parseFloat and rounding it up to the next full pixel value to prevent shrinking the initial height of the primary panel.

## Related Issue

https://github.com/adobe/spectrum-web-components/issues/1243

## Motivation and Context

it solves the problem of current client experiencing a shrink of initial height when adding a second panel and using the attribute `primary-size="auto"`. It leads to display scrollbars where in that use case it shouldn't be displayed of the first panel content. 

## How Has This Been Tested?

Visual testing, correct behavior already in local installation of the package to verify correct solution.

## Types of changes

## Checklist:

### DIFF
--- a/packages/split-view/package.json
+++ b/packages/split-view/package.json
@@ -18,7 +18,7 @@
         "lit-element",
         "lit-html"
     ],
-    "version": "0.1.0",
+    "version": "0.1.1",
     "description": "",
     "main": "src/index.js",
     "module": "src/index.js",
--- a/packages/split-view/package.json
+++ b/packages/split-view/package.json
@@ -18,7 +18,7 @@
         "lit-element",
         "lit-html"
     ],
-    "version": "0.1.1",
+    "version": "0.1.0",
     "description": "",
     "main": "src/index.js",
     "module": "src/index.js",
--- a/packages/split-view/src/SplitView.ts
+++ b/packages/split-view/src/SplitView.ts
@@ -377,9 +377,9 @@ export class SplitView extends SpectrumElement {
                 const size = window
                     .getComputedStyle(firstEl)
                     .getPropertyValue(this.vertical ? 'height' : 'width');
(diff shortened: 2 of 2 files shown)


packages/split-view/package.json
```diff
@@ -18,7 +18,7 @@
         "lit-element",
         "lit-html"
     ],
-    "version": "0.1.0",
```
Westbrook, line 21: This will change on its own later based on your usage of the `fix` syntax in your commit. Please revert.

packages/split-view/src/SplitView.ts
```diff
@@ -377,9 +377,9 @@ export class SplitView extends SpectrumElement {
                 const size = window
                     .getComputedStyle(firstEl)
                     .getPropertyValue(this.vertical ? 'height' : 'width');
-                const size_i = parseInt(size, 10);
+                const size_i = parseFloat(size);
                 if (!isNaN(size_i)) {
-                    return this.getLimitedPosition(size_i);
+                    return this.getLimitedPosition(Math.ceil(size_i));
```
Westbrook, line 382: Is there benefit of adding a visual regression that supports this?

Westbrook (changes requested):
I could go either way on adding the regression, but please remove the version change and I can approve.

packages/split-view/package.json
spdev3000: done, reverted.

packages/split-view/src/SplitView.ts
spdev3000: I would say no. As this is a real minor change only.

Westbrook (approved):
LGTM!

PR #1266, opened 2021-03-10
feat(picker): process field-label content for more accurate a11y tree

### LINKED ISSUE #1077: Dropdown accessibility review
Can remove `aria-controls` attribute as it doesn't always have an element to actually reference and even then the element doesn't yet have the `listbox` content.

Dropdown usage requires a visible label as reference; `aria-label` is not enough in this case.

When a selection is made that value needs to be tracked accessibly. Leverage `aria-labelledby="label"` to appropriately reference the internal visible element. In cases where the physical label is also referenced here, a space delineated list will reference multiple elements.

Leverage https://w3c.github.io/aria-practices/examples/listbox/listbox-collapsible.html as a backup reference for this work.

Westbrook:
## Description
- ensure that both the `field-label` content and the current `value` are represented in the accessibility tree of the picker
- prevent element errors in the "quick select" functionality
- label all of the pickers
- add accessibility snapshot tool for Playwright

## Related Issue
fixes #1077
refs #1269

## Motivation and Context
Screen reader users should have a clear representation of the content otherwise made available when seeing the UI.

## How Has This Been Tested?
- manual review with the accessibility team
- unit tests
- visual regressions

image
image

## Types of changes

## Checklist:

### DIFF
--- a/packages/picker/src/Picker.ts
+++ b/packages/picker/src/Picker.ts
@@ -59,10 +59,14 @@ type PickerSize = Exclude<ElementSize, 'xxl'>;
  * @element sp-picker
  * @slot label - The placeholder content for the picker
  *
- * @fires sp-open - Announces that the overlay has been opened
- * @fires sp-close - Announces that the overlay has been closed
+ * @fires change - Announces that the `value` of the element has changed
+ * @fires sp-opened - Announces that the overlay has been opened
+ * @fires sp-closed - Announces that the overlay has been closed
  */
 export class PickerBase extends SizedMixin(Focusable) {
+    /**
+     * @private
+     */
     public static openOverlay = async (
         target: HTMLElement,
         interaction: TriggerInteractions,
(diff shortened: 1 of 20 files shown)


packages/picker/src/Picker.ts
```diff
@@ -319,9 +323,8 @@ export class PickerBase extends SizedMixin(Focusable) {
         return html`
             <button
                 aria-haspopup="true"
-                aria-controls="popover"
                 aria-expanded=${this.open ? 'true' : 'false'}
-                aria-label=${ifDefined(this.label || undefined)}
+                aria-labelledby="button label"
```
Westbrook, line 327: Referencing _self_ chooses to `aria-label` that will be managed by `sp-field-label` elements.

Referencing `label` points to the current "value" of the picker.

test/a11y-snapshot-plugin.ts
```diff
@@ -0,0 +1,54 @@
...
+                // handle specific behavior for playwright
+                if (session.browser.type === 'playwright') {
+                    const page = session.browser.getPage(session.id);
+                    const options: {
+                        root?: ElementHandle;
+                    } = {};
+                    if (payload.selector) {
+                        console.log(payload.selector);
+                        const root = await page.$(payload.selector);
+                        console.log(root);
```
najikahalsema, line 39: You know what you have to do. Take him (the console.log) out.

najikahalsema (approved):

PR #1290, opened 2021-03-12
refactor(overlay): leverage "reparentChildren" helper

### LINKED ISSUE #1249: Overlay should use new reparentChildren helper
### Expected Behaviour
DRY code

### Actual Behaviour
Overlay could be simplified by adopting the reparentChildren helper added in #1237

Westbrook:
## Description
Use the `reparentChildren` helper in `overlay-trigger`.

## Related Issue
fixes #1249

## Motivation and Context
Simplify code by leveraging helper.

## How Has This Been Tested?
CI.

## Types of changes
- [x] Refactor

## Checklist:

### DIFF
--- a/packages/overlay/src/ActiveOverlay.ts
+++ b/packages/overlay/src/ActiveOverlay.ts
@@ -19,6 +19,7 @@ import {
     SpectrumElement,
     TemplateResult,
 } from '@spectrum-web-components/base';
+import { reparentChildren } from '@spectrum-web-components/shared';
 import { Color, Scale } from '@spectrum-web-components/theme';
 import styles from './active-overlay.css.js';
 import {
@@ -115,9 +116,7 @@ export class ActiveOverlay extends SpectrumElement {
     public overlayContentTip?: HTMLElement;
     public trigger!: HTMLElement;
 
-    private placeholder?: Comment;
     private popper?: Instance;
-    private originalSlot?: string;
 
     @property()
     public _state = stateTransition();
@@ -155,6 +154,7 @@ export class ActiveOverlay extends SpectrumElement {
 
     public tabbingAway = false;
     private originalPlacement?: Placement;
+    private restoreContent?: () => Element[];
 
     /**
      * @prop Used by the popper library to indicate where the overlay was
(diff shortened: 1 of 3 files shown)


packages/shared/src/reparent-children.ts
```diff
@@ -25,13 +31,15 @@ export const reparentChildren = (
...
         const parentElement =
             srcElement.parentElement || srcElement.getRootNode();
         parentElement.replaceChild(placeholderItem, srcElement);
         newParent.append(srcElement);
     }
 
-    return function () {
-        restoreChildren(placeholderItems, srcElements);
+    return function (): Element[] {
+        return restoreChildren(placeholderItems, srcElements, slotNames);
```
Westbrook, line 43: Use return so that consumers that want to do something more with the elements after their return can.

packages/overlay/src/ActiveOverlay.ts
```diff
@@ -392,37 +372,15 @@ export class ActiveOverlay extends SpectrumElement {
...
-        }
+        const [element] = this.restoreContent();
+        this.restoreContent = undefined;
 
         if (this.originalPlacement) {
-            this.overlayContent.setAttribute(
-                'placement',
-                this.originalPlacement
-            );
+            element.setAttribute('placement', this.originalPlacement);
```
Westbrook, line 381: This is managed here, and `slot` is managed there... it begs a question should the API be:
```
reparentChildren(Element[], Element, () => () => void);
```
Where the double callback at the end amount to `prepare` and `cleanup` functionality?

In this way, `slot` wouldn't have to be managed on _everything_, and `placement` or whatever else came up over time, wouldn't have to be cached in the originating code location. 

Everything could just be managed/cached at the originating location, just the same, but want to make the workflow clear as one of the follow up tasks to the overall refactor here will be to add this method to: https://opensource.adobe.com/spectrum-web-components/components/shared

packages/picker/src/Picker.ts
```diff
@@ -102,7 +102,7 @@ export class PickerBase extends SizedMixin(Focusable) {
     public readonly = false;
 
     public menuItems: MenuItem[] = [];
-    private restoreChildren?: Function;
+    private restoreChildren?: () => void;
```
adixon-adobe, line 105: Do you need the `?` with this change?

packages/overlay/src/ActiveOverlay.ts
adixon-adobe: I'm assuming the callbacks would be optional -- definitely makes sense to add more control to the API.

packages/picker/src/Picker.ts
adixon-adobe, line 105: Ahhh -- still learning typescript here. This makes sense to me.

adixon-adobe (approved):
Looks good!

packages/overlay/src/ActiveOverlay.ts
Westbrook: Check it out in #1310

PR #1307, opened 2021-03-19
fix(overlay): persist hover overlay when there is not click content

### LINKED ISSUE #277: Allow a "hover" based `active-overlay` that only closes on "mouseout"
### Expected Behaviour
A triggers hover overlay can stay open even if I click on the item in question. This is particularly useful in contexts where the meaning of the interaction point toggles when clicked. Should likely be a setting of some kind on the trigger?

### Actual Behaviour
Clicking closes all overlays.

Westbrook:
## Description
"hover-content" in an `<overlay-trigger>` will not persist through click events on the trigger element when there is not "click-content" the needs to be shown in response to that interaction. Further content centric events, `longpress`, `mouseout`, `blur`, etc. will cause the hover content to be closed. This is particularly useful in cases where clicking the trigger toggles its state and that state is represented in the hover content that has been projected.

## Related Issue
fixes #277

## Motivation and Context
Better tooltips for triggers that toggle.

## How Has This Been Tested?
New unit tests. https://westbrook-persist-hover-overlay--spectrum-web-components.netlify.app/storybook/?path=/story/tooltip--overlaid-top

## Types of changes

## Checklist:

### DIFF
--- a/packages/overlay/src/OverlayTrigger.ts
+++ b/packages/overlay/src/OverlayTrigger.ts
@@ -24,6 +24,7 @@ import {
     Placement,
     TriggerInteractions,
     OverlayOptions,
+    OverlayOpenCloseDetail,
 } from './overlay-types';
 import { openOverlay } from './loader.js';
 import overlayTriggerStyles from './overlay-trigger.css.js';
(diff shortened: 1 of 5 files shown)


najikahalsema (approved):
Nice catch! Thank you for fixing this.

PR #1310, opened 2021-03-22
refactor: use a callback to handle extended reparenting management

Westbrook:
@adixon-adobe I'm not 100% sure it's required, just yet, but if we were to expand on #1290 with the callback pattern to manage extended interaction with the reparented elements, it might look like this.

### DIFF
--- a/packages/overlay/src/ActiveOverlay.ts
+++ b/packages/overlay/src/ActiveOverlay.ts
@@ -362,7 +362,17 @@ export class ActiveOverlay extends SpectrumElement {
 
     private stealOverlayContent(element: HTMLElement): void {
         this.originalPlacement = element.getAttribute('placement') as Placement;
-        this.restoreContent = reparentChildren([element], this);
+        this.restoreContent = reparentChildren(
+            [element],
+            this,
+            (el: Element) => {
+                const slotName = el.slot;
+                el.removeAttribute('slot');
+                return (el: Element) => {
+                    el.slot = slotName;
+                };
+            }
+        );
         this.stealOverlayContentResolver();
     }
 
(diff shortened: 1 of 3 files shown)


najikahalsema (dismissed):
Looks good, but I know @adixon-adobe has been working with you on this, so I think his opinion/review is going to be more important in case I missed something.

packages/overlay/src/ActiveOverlay.ts
```diff
@@ -362,7 +362,17 @@ export class ActiveOverlay extends SpectrumElement {
...
+        this.restoreContent = reparentChildren(
+            [element],
+            this,
+            (el: Element) => {
+                const slotName = el.slot;
+                el.removeAttribute('slot');
+                return (el: Element) => {
+                    el.slot = slotName;
+                };
+            }
```
adixon-adobe, line 374: Does typescript automatically assign this to a single-entry array?

adixon-adobe (changes requested):
I've got one question. It also seems like it'd be good to add a multi-element unit test for reparent-children.

packages/overlay/src/ActiveOverlay.ts
Westbrook: Assign which part to a single entry array?

Westbrook (commented):
Good idea. I'll isolate this functionality to a specific unit test so it's not only tested as an integration in the larger element where it's leveraged.

packages/overlay/src/ActiveOverlay.ts
adixon-adobe: ahh -- I just misread the parameters here (was looking at `restoreChildren` not `reparentChildren`. Now that I'm reading this right this looks great.

Westbrook:
@adixon-adobe let me know if these test cover what you're hoping for here. 🙏🏼

adixon-adobe (approved):
Looks great!

PR #1315, opened 2021-03-22
fix(overlay): handle hover/longpress more directly via the "open" attribute

### LINKED ISSUE #1313: Dialog / Dialog Wrapper examples only allow the dialog to be shown once
I'm not sure if this is exactly a bug, but it does _seem_ like a bug to a new user. When I goto these components, I can only pop up the dialog once.

https://opensource.adobe.com/spectrum-web-components/components/dialog-wrapper

Likewise, the demo pages show similar behavior, and since we toggle the `this.open` property, it seems like it should be reopenable. 
https://webcomponents.dev/edit/collection/fO75441E1Q5ZlI0e9pgq/RSDikStPmUPSioVpCsYb/src/index.ts?branch=master

Westbrook:
## Description
- add state (rather than a workaround) to handle whether a `longpress` overlay has been opened via the mouse or via the keyboard
- correct the management of whether an `sp-closed` event should close ALL overlays or not
- update/add tests to support maintaining this long term

## Related Issue
fixes #1313

## Motivation and Context
Non-hover/longpress (mouse only) overlays wouldn't open a second time.

## How Has This Been Tested?
- manually
- new tests
- updated tests

https://westbrook-overlay-trigger--spectrum-web-components.netlify.app/components/dialog-wrapper

## Types of changes

## Checklist:

### DIFF
--- a/packages/overlay/src/OverlayTrigger.ts
+++ b/packages/overlay/src/OverlayTrigger.ts
@@ -77,9 +77,13 @@ export class OverlayTrigger extends LitElement {
     private targetContent?: HTMLElement;
 
     private handleClose(event?: CustomEvent<OverlayOpenCloseDetail>): void {
-        if (!event || this.open === event.detail.interaction) {
-            this.removeAttribute('open');
+        if (
+            event?.detail.interaction !== this.open &&
+            event?.detail.interaction !== this.type
+        ) {
+            return;
         }
+        this.removeAttribute('open');
     }
 
     protected render(): TemplateResult {
@@ -200,7 +204,7 @@ export class OverlayTrigger extends LitElement {
         switch (event.type) {
             case 'mouseenter':
             case 'focusin':
-                if (this.hoverContent) {
+                if (!this.open && this.hoverContent) {
                     this.open = 'hover';
                 }
                 return;
@@ -219,6 +223,7 @@ export class OverlayTrigger extends LitElement {
                 return;
             case 'longpress':
                 if (this.longpressContent) {
+                    this._longpressEvent = event;
                     this.open = event.type;
                 }
                 return;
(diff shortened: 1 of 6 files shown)


packages/overlay/src/OverlayTrigger.ts
```diff
@@ -77,9 +77,13 @@ export class OverlayTrigger extends LitElement {
     private targetContent?: HTMLElement;
 
     private handleClose(event?: CustomEvent<OverlayOpenCloseDetail>): void {
-        if (!event || this.open === event.detail.interaction) {
-            this.removeAttribute('open');
+        if (
+            event?.detail.interaction !== this.open &&
+            event?.detail.interaction !== this.type
```
Westbrook, line 82: An overlay received either the content type or the overlay type as its `interaction` when opening, meaning that the interaction will be one of the current value of `open` or the value for `type` when a single overlay is being closed. 

Hover overlays _may_ close first and shouldn't close a second time. Particularly, this is important when a hover overlay is "closed" by opening a new overlay.

packages/overlay/src/OverlayTrigger.ts
```diff
@@ -251,26 +256,27 @@ export class OverlayTrigger extends LitElement {
...
+            this._longpressEvent?.detail.source !== 'keyboard';
         this.closeLongpressOverlay = await OverlayTrigger.openOverlay(
             targetContent,
-            interaction,
+            this.type ? this.type : 'longpress',
             longpressContent,
             {
                 ...this.overlayOptions,
                 receivesFocus: 'auto',
+                notImmediatelyClosable,
```
Westbrook, line 276: Expand the state available to the overlay so that `longpress` never needs to be overridden as `click`.

najikahalsema (approved):
Looks good to me! Thank you for the explanations.

PR #1341, opened 2021-03-31
docs: add reactivity to the code-example themeing

Westbrook:
## Description
https://westbrook-reactive-code-themes--spectrum-web-components.netlify.app/
- allow the code-example themes to associate themselves to the site wide theme reactively
- simplify to a single code-example element, rather than a parent with children of each theme type.

## Related Issue
- user feedback

## Motivation and Context
The examples should be easy to read.

## How Has This Been Tested?
Manually at the URL above.

image
image

## Types of changes
- [x] Docs feature

## Checklist:


adixon-adobe (commented):
Changes look good -- I'm wondering if the package.json changes were intended for this PR though

Westbrook:
In order to collapse the 3 elements into 1, I needed a way to wrap the PrismJS styles for application based on a parent selector.  In that way https://github.com/adobe/spectrum-web-components/pull/1341/files#diff-42697c00ff27517e4da92b7a3c65903ee913403b2d401c0d314036f85f431ffd and https://github.com/adobe/spectrum-web-components/pull/1341/files#diff-fa81c17a9eb37f7487882a241b11eeb7542e60e4e82c4f83f22b862c7ae0918f are wrapped with a prefix via: https://github.com/adobe/spectrum-web-components/pull/1341/files#diff-11c58135e2334a5478d2c0fc5390bf2a0727903dc516575b39a6c41a78ed8569R105 which requires `postCSS@8`. While _we_ use `postcss@8`, one of those tools doesn't and is preferred by webpack when deciding which version to use, so I started upgrading them all, and only in the last one that I updated (`style-loader`, wanted to do webpack specific things last so that I didn't get tricked into moving to `webpack@5`) did it actually help. I could revisit and try to get something more specific to the requirements there if you prefered.

adixon-adobe:
Sounds good -- just wanted to be sure that was intentional

adixon-adobe (approved):

Issue #1190, opened 2021-02-14
Getting startet 404


Fabioni:
## Problem

When I am browing to the side https://opensource.adobe.com/spectrum-web-components/getting-started I get the following errors in my console:
<img width="950" alt="Bildschirmfoto 2021-02-14 um 15 55 36" src="https://user-images.githubusercontent.com/45362676/107880109-15f2e600-6edd-11eb-9678-9d58a37c25d3.png">

The same thing happends when I want to try out the WebComponents in my own project with:
```
<script
    src="https://jspm.dev/@spectrum-web-components/bundle/elements.js"
    type="module"
></script>
```

The example click-me button also render just as this, no component is working at all:

<img width="731" alt="Bildschirmfoto 2021-02-14 um 15 58 00" src="https://user-images.githubusercontent.com/45362676/107880172-6bc78e00-6edd-11eb-9b46-53292d8a76f5.png">

## What I tried:
- deactivating adblocker

I appreciate your help to get the "getting started" working for me. Its hard to dive into the topic if even the getting started code does not seem to work.

Westbrook:
Thanks for reporting this, I'm not sure why this stopped work, but will dive deeper on this during the week. It's possible JSPM has decided to alter its use of the `exports` entry in `package.json` to a newer version. The specification there has been pretty fluid and while an important tool for auto-processing CDNs like this and unpkg or Skypack, it's been trouble to keep up with as a library vendor. My apologies for this happening!

In the mean time, remember that this quick start approach does not preclude using NPM packages directly, as would be important to do when you were ready to tune your application for production. Each of our packages has installation instructions on their individual pages of the documentation. You can also get started by forking this project: https://webcomponents.dev/edit/JWGqd9YV0d63ZVKERLuj/src/index.ts or the starter for any one of our element starters here: https://webcomponents.dev/org/adobe

I'll try to get back to you with more information later in the week.

Westbrook:
There are deeper issues at play here that we've yet to find.

PR #1350, opened 2021-04-06
Various package structure changes for easier use and better CDN support

Westbrook:
## Description
Working through the CDN failures brings quality findings, even if I've not actually figured out for certain the source of some of the CDN failures.
- move `custom-elements.json` out of the `exports` listings
- remove dev dependencies of in monorepo packages to reduce cycles
- ensure full dependency listings when missing
- move `icons` into a stand alone export in `bundle`
- include `custom-elements.json` listings in the icon packages
- update theme resolution process to handle late loading theme roots
  - this is possible when the child element definition is loaded before the style root's even if they are "loaded at the same time"

## Related Issue
refs #1190

## Motivation and Context
Things should work on the CDN! Thins should work in more contexts.

## How Has This Been Tested?
Manually, very manually.

## Types of changes
- [ ] refactor

## Checklist:


packages/accordion/package.json
```diff
@@ -27,13 +27,13 @@
...
         ".": "./src/index.js",
         "./src/*": "./src/*",
-        "./custom-elements.json": "./custom-elements.json",
         "./package.json": "./package.json",
         "./sp-accordion": "./sp-accordion.js",
         "./sp-accordion.js": "./sp-accordion.js",
         "./sp-accordion-item": "./sp-accordion-item.js",
         "./sp-accordion-item.js": "./sp-accordion-item.js"
     },
+    "customElementsManifest": "custom-elements.json",
```
adixon-adobe, line 36: Will this still be in the package when this is released? We're using custom-elements.json to construct some linting rules.

adixon-adobe (commented):
Just want to make sure these custom-elements.json changes won't break some integration on our side

packages/accordion/package.json
Westbrook: This will still be in the package. 

However, talk to me about these lint rules...they sound exciting! While I am interested in whether we could surface them more broadly, tell me specifically about how they are built. It _is_ possible that certain tooling will cease to find them if they are listed this way.

Depending on what that looks like, I can also see removing this one commit from the PR to prevent having to worry about any complication here.

Westbrook: I _think_ there should be no issue here in the linting code we've looked at so far. However, to support proving that, I've moved this change into its own PR. This way we can work through all of the currently open PRs, and then do a test release with the `custom-elements.json` change, and depending on how that effects these lint rules move forward with a full release on one side or the other of that commit. Sound reasonable?

adixon-adobe: Definitely -- although in this case you might be doing more work than necessary. I just wanted to be sure the file would still exist -- I feel like I might have inadvertently led you down more of a rabbit hole than necessary.

adixon-adobe (approved):

PR #1359, opened 2021-04-07
refactor: update exports listings, exclude custom-elements.json

Westbrook:
## Description
Sequesters this change into a stand alone PR so that we can confirm any ecosystem side effects of it specifically.

Let's get the currently open PRs moved through the system and do a canary release once we've rebased this onto those changes and see what we can see in the lint rules that Alex brought up.

## Motivation and Context
CDNs/Node likely shouldn't be building with this directly, so particular to large ones (the icons listings) this can save a lot of trouble at the CDN level.

## Types of changes
- [x] Metadata refactor

## Checklist:

### DIFF
--- a/packages/accordion/package.json
+++ b/packages/accordion/package.json
@@ -22,7 +22,6 @@
     "exports": {
         ".": "./src/index.js",
         "./src/*": "./src/*",
-        "./custom-elements.json": "./custom-elements.json",
         "./package.json": "./package.json",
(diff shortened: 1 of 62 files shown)


adixon-adobe (dismissed):
Approving -- I don't think we'll need a canary release to test this (it seems pretty unlikely to be problematic based on our discussions). Either way though I'll be sure to have us test after the release.

PR #1362, opened 2021-04-09
fix(menu): only scrollIntoView when keyboard navigating

### LINKED ISSUE #1361: sp-picker requires multiple clicking to choose item when in scrolling state
### Expected Behaviour
Menu items in an sp-picker should be easily clickable (one click and done)

### Actual Behaviour
When the sp-picker doesn't have enough room and needs to scroll there is some weird jumpy scroll position behavior when clicking, and selecting a menu item requires multiple clicks

Create a long-ish menu with an sp-picker. Resize your browser such that it needs to scroll when open. Now try to use the picker.

https://webcomponents.dev/edit/rlz62LTQSuyMdK4cCf8W/src/index.ts
Make sure to resize the browser small such that the popover menu needs to scroll

Westbrook:
## Description
The `scrollIntoView()` functionality to support keyboard interactions actually moves the `sp-menu-item` when scrolling is possible making the `click` event not trigger correctly. This sequester the `scrollIntoView()` call to _only_ the keyboard workflow to support this.

TODO:
- [x] ensure open/close visibility of selected/focused item
- [x] think through how to capture this functionality in a unit test

## Related Issue
fixes #1361 

## Motivation and Context
Works as expected.

## How Has This Been Tested?
Manually

## Preview:
https://westbrook-menu--spectrum-web-components.netlify.app/storybook/?path=/story/picker--custom

## Types of changes

## Checklist:

### DIFF
--- a/packages/menu/src/Menu.ts
+++ b/packages/menu/src/Menu.ts
@@ -138,12 +138,17 @@ export class Menu extends SpectrumElement {
         if (code !== 'ArrowDown' && code !== 'ArrowUp') {
             return;
         }
-        event.preventDefault();
+        const lastFocusedItem = this.menuItems[this.focusedItemIndex];
         const direction = code === 'ArrowDown' ? 1 : -1;
-        this.focusMenuItemByOffset(direction);
+        const itemToFocus = this.focusMenuItemByOffset(direction);
+        if (itemToFocus === lastFocusedItem) {
+            return;
+        }
+        event.preventDefault();
+        itemToFocus.scrollIntoView({ block: 'nearest' });
     }
 
-    public focusMenuItemByOffset(offset: number): void {
+    public focusMenuItemByOffset(offset: number): MenuItem {
         const step = offset || 1;
         const focusedItem = this.menuItems[this.focusedItemIndex] as MenuItem;
         focusedItem.focused = false;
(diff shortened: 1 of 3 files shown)


bengfarrell (dismissed):
Ahhhh, that makes sense. I didn't quite know what was going on with this behavior, it felt like it was scrolling to some sort of state.

packages/menu/src/Menu.ts
```diff
@@ -262,6 +262,12 @@ export class Menu extends SpectrumElement {
         }
         this.observer.observe(this, { childList: true, subtree: true });
         this.updateComplete.then(() => this.prepItems());
+        const selectedItem = this.querySelector('[selected]');
```
adixon-adobe, line 265: Seems like this should be good with multi-select too, as it should scroll the first selected item into view.

adixon-adobe (approved):
The one semi-contrived case I'm seeing here is if you have a long (scrollable) menu where everything is disabled, or only one item is enabled, it looks like you can't easily view the scrolled-out-of-view items with keyboard navigation.

I'm really not sure if that matters, and I don't think it should hold this up. Just something I noticed while working through the logic here.

Westbrook:
@adixon-adobe added a little gate around the `preventDefault()` to allow for your edge case. _Likely_ not something that happens a lot, but nice little spice on the API.

packages/menu/src/Menu.ts
```diff
@@ -138,12 +138,18 @@ export class Menu extends SpectrumElement {
         if (code !== 'ArrowDown' && code !== 'ArrowUp') {
             return;
         }
-        event.preventDefault();
+        const lastFocusedItem = this.menuItems[this.focusedItemIndex];
         const direction = code === 'ArrowDown' ? 1 : -1;
-        this.focusMenuItemByOffset(direction);
+        const itemToFocus = this.focusMenuItemByOffset(direction);
+        console.log(itemToFocus, lastFocusedItem);
```
CoreyDotCom, line 144: Will assume lint checks will catch this console.log. :)

Westbrook: `eslint-plugin-lucier` was a great add to the project 😬 

I'll look at getting something official in here soon!

adixon-adobe (approved):
Looks good!

Issue #715, opened 2020-06-01
Menu should manage its own state


Westbrook:
### Expected Behaviour
`value` of `sp-menu` and `selected` of `sp-menu-item` should not rely on parent elements that are not in the `menu` family to manage their state.

### Actual Behaviour
`sp-dropdown` manages this for the cases that this occurs now which makes transporting those interactions difficult. See the `sp-split-button` pattern that would leverage some version of this.

Westbrook:
This will soon be possible a la the following demos:
- `selects="single"`: https://westbrook-menu-selects--spectrum-web-components.netlify.app/storybook/?path=/story/menu--single-select
- `selects="multiple"`: https://westbrook-menu-selects--spectrum-web-components.netlify.app/storybook/?path=/story/menu--multiple-select

With various "complex" use cases outlined in:
- https://westbrook-menu-selects--spectrum-web-components.netlify.app/storybook/?path=/story/menu--menu-group-selects
- https://westbrook-menu-selects--spectrum-web-components.netlify.app/storybook/?path=/story/menu--menu-group-selects-multiple
- https://westbrook-menu-selects--spectrum-web-components.netlify.app/storybook/?path=/story/menu-group--mixed
- https://westbrook-menu-selects--spectrum-web-components.netlify.app/storybook/?path=/story/menu-group--inherit

joekukish:
@Westbrook `HTMLSelectElement` uses `multiple` (boolean) for this use case. Why not rely on a standard HTML vocabulary instead of introducing `selects`?

Westbrook:
@joekukish This is a great question, and I'm always remiss to depart from the HTML spec. However, `multiple` or not only outlines two states for this functionality, whereas in this case there are actually three: no selection, one selection, and multiple selections. Due to this, and the presence of `selects` already available in the library, I thought it would be best to perpetuate this approach. 

Can you think of a useful way in which we could leverage the `multiple` attribute, and get the trinary state covered? If there's something better than what we're doing now, it might be useful to port other APIs to leverage it, so I'm all ears on other options!

joekukish:
@Westbrook what is the use-case of a menu without selection? What happens when you click an item? Could `readonly` be used for this use case?

Westbrook:
[...]

PR #1363, opened 2021-04-10
Adixon 1189/menu customization

adixon-adobe:
Initial work to add selects support for #1189
refs #715

## Description

I figured I'd just start by adding support to `sp-menu` and `sp-menu-group` and get some feedback on the approach before integrating with our existing menu components. The storybooks for menu should work with the new changes

## Menu work to still be done (not to mention all the integration work)

* add selection get/set to menu group
* add tests!
* update documentation 
* look for low-hanging fruit accessibility issues that can be addressed here

## Checklist:

### DIFF
--- a/packages/menu/src/Menu.ts
+++ b/packages/menu/src/Menu.ts
@@ -98,6 +98,8 @@ export class Menu extends SpectrumElement {
         }) as MenuItem;
         /* c8 ignore next 3 */
         if (target) {
+            event.preventDefault();
+            event.stopImmediatePropagation();
             this.selectOrToggleItem(target);
         } else {
             return;
--- a/packages/menu/src/Menu.ts
+++ b/packages/menu/src/Menu.ts
@@ -29,6 +29,7 @@ export interface MenuQueryRoleEventDetail {
 /**
  * Spectrum Menu Component
  * @element sp-menu
+ * @fires change - Announces that the `value` of the element has changed
  *
  */
 export class Menu extends SpectrumElement {
@@ -36,12 +37,21 @@ export class Menu extends SpectrumElement {
         return [menuStyles];
     }
 
-    @property({ type: Boolean, reflect: true })
-    public selectable = false;
+    @property({ type: String, reflect: true })
+    public selects: undefined | 'single' | 'multiple';
+
+    @property({ type: String })
+    public value = '';
+
+    // TODO: allow setting this in the API to change the values
+    @property({ attribute: false })
+    public selected = [] as string[];
 
     public menuItems = [] as MenuItem[];
     public focusedItemIndex = 0;
     public focusInItemIndex = 0;
+
+    private selectedItemsMap = new Map() as Map<MenuItem, boolean>;
 
     /**
      * Hide this getter from web-component-analyzer until
@@ -87,7 +97,9 @@ export class Menu extends SpectrumElement {
             return el.getAttribute('role') === this.childRole;
         }) as MenuItem;
         /* c8 ignore next 3 */
-        if (!target) {
+        if (target) {
+            this.selectOrToggleItem(target);
+        } else {
             return;
         }
         this.prepareToCleanUp();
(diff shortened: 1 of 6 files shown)


packages/menu/src/MenuGroup.ts
```diff
@@ -32,6 +34,9 @@ export class MenuGroup extends SpectrumElement {
         return [menuGroupStyles];
     }
 
+    @property({ type: String, reflect: true })
+    public selects: undefined | 'single' | 'multiple';
```
adixon-adobe, line 38: At least menu group needs a `selects="none"` option, as the default is really "not directly managed, but a parent element can manage". What I was curious about though is whether I should also add that to `sp-menu`.

In other words, at least menu group needs to distinguish between "my parent menu can manage my selection" and "the item should not not be selectable". This is a nuanced case I missed in the issue description so I'll update with what we decide here. My gut says we might as well add it to menu, since it's simpler to manage the overall logic and won't hurt anything.

packages/menu/src/Menu.ts
```diff
@@ -29,20 +29,30 @@ export interface MenuQueryRoleEventDetail {
...
         return [menuStyles];
     }
 
-    @property({ type: Boolean, reflect: true })
-    public selectable = false;
+    @property({ type: String, reflect: true })
+    public selects: undefined | 'single' | 'multiple';
+
+    @property({ type: String })
+    public value = '';
```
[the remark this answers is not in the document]
Westbrook: I had no idea that's how this worked 🤯

What's the expected path in a `<select multi>` element to acquire the `value`? If we didn't use `selected` below, we'd need to be sure that this was in the documenation.

packages/menu/src/Menu.ts
```diff
@@ -124,6 +138,67 @@ export class Menu extends SpectrumElement {
         this.removeEventListener('keydown', this.handleKeydown);
     }
 
+    public async selectOrToggleItem(item: MenuItem): Promise<void> {
```
Westbrook, line 141: This workflow also needs a similar initialization as found in the `sp-action-group` element here: https://github.com/adobe/spectrum-web-components/blob/main/packages/action-group/src/ActionGroup.ts#L242

This ensures that all of the aria attributes are appropriately applied.

packages/menu/src/MenuGroup.ts
Westbrook: I think this makes sense. However, this might interest with the way the the `[selects]` selector is used to apply sizing rules from Spectrum CSS.

packages/menu/src/MenuGroup.ts
```diff
@@ -41,13 +46,24 @@ export class MenuGroup extends SpectrumElement {
         MenuGroup.instances += 1;
     }
 
+    private get menuRole() {
```
Westbrook, line 49: We'll need to confirm that this actually associates the menu with its items through the slot... `sp-action-group` has a slightly different relationship between the `group` and `item` elements.

packages/menu/src/MenuGroup.ts
```diff
@@ -41,13 +46,24 @@ export class MenuGroup extends SpectrumElement {
...
+
     public render(): TemplateResult {
         const labelledby = `menu-heading-category-${this.instanceCount}`;
         return html`
             <span class="header" id=${labelledby} aria-hidden="true">
                 <slot name="header"></slot>
             </span>
-            <sp-menu role="presentation">
+            <sp-menu
+                selects=${ifDefined(this.selects ? this.selects : undefined)}
```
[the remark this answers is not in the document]
Westbrook: A relatively need addition.

packages/menu/src/Menu.ts
Westbrook: Similarly, if there was a way for this to hew as closely to the functionality in `sp-action-group` it's likely that a path toward shared code here could be of benefit down the line.

packages/menu/src/Menu.ts
```diff
@@ -29,20 +29,30 @@ export interface MenuQueryRoleEventDetail {
 /**
  * Spectrum Menu Component
  * @element sp-menu
+ * @fires change - Announces that the `value` of the element has changed
```
Westbrook, line 32: 👍🏼

packages/menu/src/MenuGroup.ts
Westbrook: In this way, maybe it's not `selects="none"` but `selects="inherit"` or similar and then no value for `selects` is the never select option?

packages/menu/src/Menu.ts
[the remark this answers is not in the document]
adixon-adobe: Had to do some light reading to figure that out. In a `select` element, you need to iterate through the selectedOptions to get the values: https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/selectedOptions

It feels like deviating from the select element a bit is good to make it more convenient.

packages/menu/src/MenuGroup.ts
adixon-adobe: I'd definitely want menu group to default to `inherit`. I think the API would be pretty awkward otherwise.

So maybe that's specific to `menu-group`?

packages/menu/src/Menu.ts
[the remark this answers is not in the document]
Westbrook: Yes, please!

packages/menu/src/MenuGroup.ts
Westbrook: I was thinking on this in the realm of styling. There is specific styling for the menu items when a selection can be made. If `select="none"` is the attribute to removes selection, but sometimes selection can happen when no attribute at all, then you add complexity to what the CSS rules should be processed to in that case. None of this is a blocker, just complexity to keep in mind.

adixon-adobe: I think this would work, and as I'm thinking about it more I think the API different here is okay (but also something that should get explicitly called out in the `sp-menu-group` documentation). Since sp-menu-group manages the underlying sp-menu, I think this would avoid some of the css complexity you've mentioned.

**sp-menu** `selects: undefined | 'inherit' | 'single' | 'multiple' // default to not selectable
**sp-menu-group** `selects: undefined | 'none' | 'single' | 'multiple' // default to 'inherit' in the underlying sp-menu

adixon-adobe:
should have closed this a while back

PR #1380, opened 2021-04-15
fix(button): no double link clicks

### LINKED ISSUE #1376: sp-button link variant double-fires the click
### Expected Behaviour
If I click on a sp-button with an href defined, it only triggers once.

### Actual Behaviour
It can fire twice (most easily with a touchpad). If you use `target="_blank"` it can open the link in 2 tabs.

I've set up a repro case here: https://webcomponents.dev/edit/ykAbF0lehaZnRLAY73n3/src/index.ts

#### Steps to Reproduce
Click the CTA button in the above sample using a touchpad. I'm not reproducing every time, but pretty often.

#### Platform and Version
Mac (Chrome & FF at least).

Westbrook:
## Description
Commit to one version of surfacing the anchor element in `sp-button href` and derivative elements.
- the single placement of an invisible link over the entire button element prevents the need to proxy a click _unless_ the implementing developer uses `button.click()`
- gate the click event listening to proxy `sp-button type="submit"` on the availability of an `href` value

## Related Issue
fixes #1376 

## Motivation and Context
Should only click once.

## How Has This Been Tested?
Manually in Firefox and with the existing unit tests.

## Types of changes

## Checklist:

### DIFF
--- a/packages/button/src/ButtonBase.ts
+++ b/packages/button/src/ButtonBase.ts
@@ -112,7 +112,6 @@ export class ButtonBase extends LikeAnchor(
         if (this.anchorElement) {
             this.anchorElement.click();
             handled = true;
-            this.anchorElement.classList.remove('clicking');
         } else if (this.type !== 'button') {
             const proxy = document.createElement('button');
             proxy.type = this.type;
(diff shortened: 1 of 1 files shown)


adixon-adobe (dismissed):
Thanks!

Issue #769, opened 2020-07-15
Tags should only be focusable if they can be actioned (deleted / selected)


jnurthen:
### Expected Behaviour
There is no need to be able to tab to a Tag

### Actual Behaviour
Each tag group is tabable and then arrow keys are used to move between tags. When tags are a simple list they are not actionable elements and should not be in the tab order. When tags are removable then allow tabbing to the delete button directly.

#### Steps to Reproduce
https://opensource.adobe.com/spectrum-web-components/components/tags/examples
tab to the tag

#### Logs taken while reproducing problem

Note - lists/listitem are structural roles in ARIA not widget roles. If they really need to be tabable then they should be a layout grid not a list.

Westbrook:
The Spectrum CSS and Spectrum Specification documentation for this seems like it is laking some clarity based on this extended context. @jnurthen do you have some time next week to touch base on this in-person to ensure that I fully understand the nuance here?

jnurthen:
Sure. Please find a time in calendar. I'm sure there are probably a few more issues to discuss too.

Westbrook:
Ensure focus of tag and tags when any subset of tag in tags in deletable / selectable.

PR #1400, opened 2021-04-22
Fix(tags): tags should only be focusable if they can be actioned (deleted / selected)

najikahalsema:
## Description

The tabs are now only included in the tab order if they can be actioned (ie deleted). 

## Related Issue

fixes https://github.com/adobe/spectrum-web-components/issues/769 

## Motivation and Context

Our tags could be tab indexed regardless of their ability to be changed by the user. They're not actionable elements, so they shouldn't be included in the tab order. 

## How Has This Been Tested?

Manually and with added tests

## Types of changes

## Checklist:

### DIFF
--- a/packages/tags/src/Tag.ts
+++ b/packages/tags/src/Tag.ts
@@ -129,8 +129,13 @@ export class Tag extends SpectrumElement {
         if (!this.hasAttribute('role')) {
             this.setAttribute('role', 'listitem');
         }
-        if (!this.disabled) {
-            this.setAttribute('tabindex', this.deletable ? '0' : '-1');
+        if (this.deletable) {
+            this.setAttribute(
+                'tabindex',
+                !this.disabled && this.matches(':first-of-type:not([disabled])')
+                    ? '0'
+                    : '-1'
+            );
         }
     }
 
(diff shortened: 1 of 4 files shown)


packages/tags/src/Tag.ts
```diff
@@ -129,12 +129,9 @@ export class Tag extends SpectrumElement {
...
             this.setAttribute('role', 'listitem');
         }
-        this.setAttribute(
-            'tabindex',
-            !this.disabled && this.matches(':first-of-type:not([disabled])')
-                ? '0'
-                : '-1'
-        );
+        if (!this.disabled) {
+            this.setAttribute('tabindex', this.deletable ? '0' : '-1');
```
Westbrook, line 133: Great start! 

This previously applied a pattern called "floating tab index" wherein only one time in the `sp-tags` element at a time should have `tabindex=0`, which was powered by `:first-of-type:not([disabled])` in the previous code. That might not be the best approach, but we _should_ continue to surface that pattern. The fact that removing it didn't cause a test to fail also means we likely want to add a test to support the continued presence of the "floating tab index" pattern.

Please update the `tabindex` management here and add a test to ensure that such a pattern is not regressed.

Westbrook (approved):

Westbrook (approved):
LGTM!

PR #1140, opened 2021-01-26
fix(slider): manage focus more like a native rage input

Westbrook:
## Description
Previously throwing focus into the `<input>` element of a majority of interactions left the browser a difficult task of discerning whether a focus event was "focus visible" (keyboard triggered) or not. This meant that if the slider was interacted with (focused) via a pointing device and the the browser tab was blurred/focused again the "focus" would be applied to the slide as "focus visible", igniting the blue focus ring. This throws the focus to the `<label>` element via an imperative `click` event, which under the browser covers throws focus to the `<input>` without directly assigning additional event origin information. This leads to the "focus visible" check being more natural and the focus when returning to the tab active more like an `<input type="range">` element.

## Related Issue
fixes #517 

## Motivation and Context
We want to be "native" element, but _better_.

## How Has This Been Tested?
Manually and with additional and altered unit testing.

## Screenshots (if appropriate):
image

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [x] I have signed the [Adobe Open Source CLA](http://opensource.adobe.com/cla.html).
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #1416, opened 2021-04-28
Multi-handle slider implementation

### LINKED ISSUE #1385: [range-slider]
## Multi-Handle Slider Proposal

This began as a proposal for an implementation of the range slider as documented in [Spectrum Contributions](https://spectrum-contributions.corp.adobe.com/page/slider-beta/#Additional-variants) and as implemented in [spectrum-css](https://opensource.adobe.com/spectrum-css/slider.html#range).

The discussion (see below) has led us to a more general extension to sliders to allow multiple handles. This is the updated reworked proposal.

The basic gist of the proposal is to allow the user to specify the handles explicitly using an `sp-slider-handle` element as per this example:

```html
<sp-slider min=0 max=10>
  <sp-slider-handle name="x" value="0" min="0" max="next"></sp-slider-handle>
  <sp-slider-handle name="y" value="1" min="0" max="8"></sp-slider-handle>
  <sp-slider-handle name="z" value="3" min="previous" max="10"></sp-slider-handle> 
</sp-slider>
```

Node that in the absence of specified name, the handles will be named `handle1`, `handle2`, etc.

## Backwards compatibility and the simple case

In the case where only one handle is needed, we will read the `value`, `min` and `max` properties off of the `sp-slider` itself. It will act as the `sp-slider-handle`.

```html
<sp-slider value="0" min="0" max="10"></sp-slider>
```

## Variants

It is likely that we will only support the default variant when there is more than one handle.

## Handle clamping

In most cases, but not all, we will want to enforce the ordering of the handles. When ordering is enforced, handles will not be able to move past each other. The API will take the form of `min` and `max` properties on the handles.
attr    | allowed values
------|-------------------
`min` | `'previous'` \| `number` \| `undefined`
`max` | `'next'` \| `number` \| `undefined`

The values `previous` and `next` will prevent a handle from moving past the adjacent handle in that direction.

## Label formatting

Currently we support a `getAriaValueText` property which is a function that does the formatting. Our solution need to take into account #1159. See comments below. This section needs to be fleshed out further.

Each `sp-slider-handle` and the `sp-slider` will have a property like this:

```typescript
    @property({ type: Object, attribute: 'format-options' })
    public formatOptions: Intl.NumberFormatOptions = {};
```

The `getAriaValueText` property will be modified to take a function of type `(values: Map<string, string>) => string`. In the case where there is only one slider, the map will just contain the the pair `'value' => <value of slider`. The string values will be the individual slider values by name as formatted according to their `formatOptions`. If an `sp-slider-handle` does not have a value for `formatOptions` than the `formatOptions` from the host `sp-slider` will be used. 

*Question*: When there is only one implicit `sp-slider-handle`, should we just pass the value as a string instead of in a `Map`? Is a `Map` the right structure. I chose it because it is both random access and ordered.

## Events

The events should look the same as for the standard slider with `change` and `input` events. The change and input events will be per-handle and will come from the `sp-slider-handle` element. Where there are no explicitly declared `sp-slider-handle` elements (there is only one default handle), the events will come from the `sp-slider` itself.

## `values` property

The `value` property that will return an array of objects with the `name` and `value` for each handle such as:
```javascript
[
    { name: 'x', value: 0 },
    { name: 'x, value: 1 },
    { name: 'z, value: 3 }
]
```

## Interactions

From playing with other implementations, it appears that the most recently dragged handle is always on top. The keyboard should interact with the currently selected handle. And you should be able to tab between handles.

## Breaking changes

- You will not longer be able to set the label via slotted text as that gets confusing when there could be `sp-slider-handle` children.
- `getAriaValueText` function property will now be passed a `Map<string, string>` with the handle names and values.

cuberoot:
## Description

Refactor sp-slider to allow for explicit declaration of handles, including multiple handles as per #1385.

## Related Issue

fixes #1385

## Motivation and Context

There are a number of use cases for multi-handle sliders in creative apps, including range-slider contribution to Spectrum.

## Types of changes

## Checklist:

### DIFF
--- a/packages/slider/README.md
+++ b/packages/slider/README.md
@@ -81,6 +81,45 @@ import { Slider } from '@spectrum-web-components/slider';
 <sp-slider label="Slider Label - Disabled" variant="ramp" disabled></sp-slider>
 ```
 
+## Advanced normalization
+
+By default, `sp-slider` assumes a linear scale between the `min` and `max` values.
+For advanced applications, it is sometimes necessary to specify a custom
+"normalization."
+
+Normalization is the process of converting a slider to a value between 0 and 1 where
+0 represents the minimum and 1 represents the maximum. See the <a href="storybook/index.html?path=/story/slider--three-handles-complex" target="_blank">"Three Handles Complex" example in the playground</a>.
+
+## Labels and Formatting
+
+An `<sp-slider>` or `<sp-slider-handle>` element will process its numeric value with `new Intl.NumberFormat(navigator.language, this.formatOptions).format(this.value)` in order to prepare it for visual delivery in the input. In order to customize this processing supply your own `Intl.NumberFormatOptions` via the `formatOptions` property, or `format-options` attribute as follows.
+
+```html
+<sp-slider
+    min="0"
+    max="1"
+    step="0.01"
+    value="0.5"
+    label="Slider Label"
+    format-options='{
+        "style": "percent"
+    }'
+></sp-slider>
+```
+
+More advanced formatting is avialable by specifying a formatting function to
+the `getAriaHandleText` property on an `sp-slider` or `sp-slider-handle`. Or,
+for a multi-handle slider, you can format the combined value label for all
+handles by passing a formatting function to the `getAriaValueText` property
+on the parent `sp-slider`.
+
+You can suppress the value label altogether using the `hide-value-label`
+attribute.
+
+```html
+<sp-slider label="No value label" hide-value-label></sp-slider>
+```
+
 ## Events
 
 Like the `<input type="range">` element after which the `<sp-slider>` is fashioned it will dispatch `input` events in a stream culminating with a `change` event (representing the final comit of the `value` to the element) once the user has discontinued with the element. Both other these events can access the `value` of their dispatching target via `event.target.value`. In this way a steaming listener patters similar to the following can prove useful:
--- a/packages/slider/package.json
+++ b/packages/slider/package.json
@@ -24,7 +24,9 @@
         "./src/*": "./src/*",
         "./package.json": "./package.json",
         "./sp-slider": "./sp-slider.js",
-        "./sp-slider.js": "./sp-slider.js"
+        "./sp-slider.js": "./sp-slider.js",
+        "./sp-slider-handle": "./sp-slider-handle.js",
+        "./sp-slider-handle.js": "./sp-slider-handle.js"
     },
     "scripts": {
         "test": "echo \"Error: run tests from mono-repo root.\" && exit 1"
@@ -43,6 +45,7 @@
         "lit-html"
     ],
     "dependencies": {
+        "@internationalized/number": "^3.0.0",
         "@spectrum-web-components/base": "^0.4.3",
         "@spectrum-web-components/shared": "^0.12.4",
         "tslib": "^2.0.0"
--- a/packages/slider/src/HandleController.ts
+++ b/packages/slider/src/HandleController.ts
@@ -1,0 +1,578 @@
+/*
+Copyright 2021 Adobe. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License. You may obtain a copy
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+import {
+    html,
+    TemplateResult,
+    ifDefined,
+    classMap,
+    styleMap,
+} from '@spectrum-web-components/base';
+import { streamingListener } from '@spectrum-web-components/base/src/streaming-listener.js';
+import { Slider } from './Slider.js';
+import {
+    SliderHandle,
+    SliderNormalization,
+    Controller,
+} from './SliderHandle.js';
+
+interface HandleReference {
+    handle: HTMLElement;
+    input: HTMLInputElement;
+}
+
+interface HandleComponents extends HandleReference {
+    model: SliderHandle;
+}
+
+interface RangeAndClamp {
+    range: { min: number; max: number };
+    clamp: { min: number; max: number };
+}
+interface ModelValue extends RangeAndClamp {
+    name: string;
+    value: number;
+    normalizedValue: number;
+    step: number;
+    highlight: boolean;
+    ariaLabel?: string;
+    normalization: SliderNormalization;
+    handle: SliderHandle;
+}
+
+interface InputWithModel extends HTMLInputElement {
+    model: ModelValue;
+}
+
+export interface HandleValueDictionary {
+    [key: string]: number;
+}
+
+export class HandleController implements Controller {
+    private observer!: MutationObserver;
+    private host!: Slider;
+    private handles: Map<string, SliderHandle> = new Map();
+    private model: ModelValue[] = [];
+    private handleOrder: string[] = [];
+    private draggingHandle?: SliderHandle;
+    private handleRefMap?: WeakMap<SliderHandle, HandleReference>;
+
+    constructor(host: Slider) {
+        this.host = host;
+    }
+
+    public get values(): HandleValueDictionary {
+        const result: HandleValueDictionary = {};
+        for (const model of this.handles.values()) {
+            result[model.handleName] = model.value;
+        }
+        return result;
+    }
+
+    public inputForHandle(handle: SliderHandle): HTMLInputElement | undefined {
+        if (this.handles.has(handle.handleName)) {
+            const { input } = this.getHandleElements(handle);
+            return input;
+        }
+        /* c8 ignore next 2 */
+        throw new Error(`No input for handle "${handle.name}"`);
+    }
+
+    public requestUpdate(): void {
+        this.host.requestUpdate();
+    }
+
+    public get language(): string {
+        return this.host.language;
+    }
+
+    /**
+     * It is possible for value attributes to be set programmatically. The <input>
+     * for a particular slider needs to have an opportunity to validate any such
+     * values
+     *
+     * @param handle Handle who's value needs validation
+     */
+    public setValueFromHandle(handle: SliderHandle): void {
+        const elements = this.getHandleElements(handle);
+        /* c8 ignore next */
+        if (!elements) return;
+
+        const { input } = elements;
+        if (input.valueAsNumber === handle.value) {
+            handle.dispatchInputEvent();
+        } else {
+            input.valueAsNumber = handle.value;
+            handle.value = input.valueAsNumber;
+            this.requestUpdate();
+        }
+        handle.value = input.valueAsNumber;
+    }
+
+    public handleHasChanged(handle: SliderHandle): void {
+        if (handle !== this.host) {
+            this.requestUpdate();
+        }
+    }
+
+    public formattedValueForHandle(model: ModelValue): string {
+        const { handle } = model;
+        const numberFormat = handle.numberFormat ?? this.host.numberFormat;
+        return handle.getAriaHandleText(model.value, numberFormat);
+    }
+
+    public get formattedValues(): Map<string, string> {
+        const result = new Map<string, string>();
+        for (const model of this.model) {
+            result.set(model.name, this.formattedValueForHandle(model));
+        }
+        return result;
+    }
+
+    public get focusElement(): HTMLElement {
+        const { input } = this.getActiveHandleElements();
+        return input;
+    }
+
+    public hostConnected(): void {
+        if (!this.observer) {
+            this.observer = new MutationObserver(this.extractModelFromLightDom);
+        }
+        this.observer.observe(this.host, { subtree: true, childList: true });
+        this.extractModelFromLightDom();
+    }
+
+    public hostDisconnected(): void {
+        this.observer.disconnect();
+    }
+
+    public hostUpdate(): void {
+        this.updateModel();
+    }
+
+    private extractModelFromLightDom = (): void => {
+        let handles = [
+            ...this.host.querySelectorAll('[slot="handle"]'),
+        ] as SliderHandle[];
+        if (handles.length === 0) {
+            handles = [this.host as SliderHandle];
+        }
+        this.handles = new Map();
+        this.handleOrder = [];
+        handles.forEach((handle, index) => {
+            /* c8 ignore next */
+            if (!handle.handleName?.length) {
+                handle.name = `handle${index + 1}`;
+            }
+            this.handles.set(handle.handleName, handle);
+            this.handleOrder.push(handle.handleName);
+            handle.handleController = this;
+        });
+        this.requestUpdate();
+    };
+
+    public get activeHandle(): string {
+        return this.handleOrder[this.handleOrder.length - 1];
+    }
+
+    public get activeHandleInputId(): string {
+        const active = this.activeHandle;
+        const index = this.model.findIndex((model) => model.name === active);
+        return `input-${index}`;
+    }
+
+    public activateHandle(name: string): void {
+        const index = this.handleOrder.findIndex((item) => item === name);
+        if (index >= 0) {
+            this.handleOrder.splice(index, 1);
+        }
+        this.handleOrder.push(name);
+    }
+
+    private getActiveHandleElements(): HandleComponents {
+        const name = this.activeHandle;
+        const handleSlider = this.handles.get(name) as SliderHandle;
+        const elements = this.getHandleElements(
+            handleSlider
+        ) as HandleReference;
+        return { model: handleSlider, ...elements };
+    }
+
+    private getHandleElements(sliderHandle: SliderHandle): HandleReference {
+        if (!this.handleRefMap) {
+            this.handleRefMap = new WeakMap();
+
+            const inputNodes = this.host.shadowRoot.querySelectorAll(
+                '.handle > input'
+            );
+            for (const inputNode of inputNodes) {
+                const input = inputNode as HTMLInputElement;
+                const handle = input.parentElement as HTMLElement;
+                const model = this.handles.get(
+                    handle.getAttribute('name') as string
+                );
+                if (model) {
+                    this.handleRefMap.set(model, { input, handle });
+                }
+            }
+        }
+
+        const components = this.handleRefMap.get(
+            sliderHandle
+        ) as HandleReference;
+        return components;
+    }
+
+    private clearHandleComponentCache(): void {
+        delete this.handleRefMap;
+    }
+
+    private get boundingClientRect(): DOMRect {
+        if (!this._boundingClientRect) {
+            this._boundingClientRect = this.host.getBoundingClientRect();
+        }
+        return this._boundingClientRect;
+    }
+
+    private updateBoundingRect(): void {
+        delete this._boundingClientRect;
+    }
+
+    private _boundingClientRect?: DOMRect;
+
+    /**
+     * Receives an event from a track click and turns it into a drag
+     * of the active handle
+     * @param event Track click event
+     */
+    public beginTrackDrag(event: PointerEvent): void {
+        const { handle } = this.getActiveHandleElements();
+        const model = this.model.find(
+            (item) => item.name === this.activeHandle
+        );
+        /* c8 ignore next */
+        if (!model) return;
+
+        event.stopPropagation();
+        event.preventDefault();
+        const applyDefault = handle.dispatchEvent(
+            new PointerEvent('pointerdown', event)
+        );
+        if (applyDefault) {
+            const model = this.model.find(
+                (model) => model.name === this.activeHandle
+            );
+            if (model) {
+                this.handlePointermove(event, model);
+            }
+        }
+    }
+
+    private handlePointerdown(event: PointerEvent, model: ModelValue): void {
+        const handle = event.target as HTMLDivElement;
+        if (this.host.disabled || event.button !== 0) {
+            event.preventDefault();
+            return;
+        }
+        this.updateBoundingRect();
+        this.host.labelEl.click();
+        this.draggingHandle = model.handle;
+        model.handle.dragging = true;
+        this.activateHandle(model.name);
+        handle.setPointerCapture(event.pointerId);
+        this.host.requestUpdate();
+    }
+
+    private handlePointerup(event: PointerEvent, model: ModelValue): void {
+        // Retain focus on input element after mouse up to enable keyboard interactions
+        const handle = event.target as HTMLDivElement;
+        const input = handle.querySelector('input') as HTMLInputElement;
+        this.host.labelEl.click();
+        model.handle.highlight = false;
+        delete this.draggingHandle;
+        model.handle.dragging = false;
+        this.requestUpdate();
+        handle.releasePointerCapture(event.pointerId);
+        this.dispatchChangeEvent(input, model.handle);
+    }
+
+    private handlePointermove(event: PointerEvent, model: ModelValue): void {
+        /* c8 ignore next 3 */
+        if (!this.draggingHandle) {
+            return;
+        }
+        event.stopPropagation();
+        const { input } = this.getHandleElements(model.handle);
+        input.value = this.calculateHandlePosition(event, model).toString();
+        model.handle.value = parseFloat(input.value);
+        this.requestUpdate();
+    }
+
+    /**
+     * Keep the slider value property in sync with the input element's value
+     */
+    private onInputChange = (event: Event): void => {
+        const input = event.target as InputWithModel;
+        input.model.handle.value = input.valueAsNumber;
+
+        this.requestUpdate();
+        this.dispatchChangeEvent(input, input.model.handle);
+    };
+
+    private onInputFocus = (event: Event): void => {
+        const input = event.target as InputWithModel;
+        let isFocusVisible;
+        try {
+            isFocusVisible =
+                input.matches(':focus-visible') ||
+                this.host.matches('.focus-visible');
+            /* c8 ignore next 3 */
+        } catch (error) {
+            isFocusVisible = this.host.matches('.focus-visible');
+        }
+        input.model.handle.highlight = isFocusVisible;
+        this.requestUpdate();
+    };
+
+    private onInputBlur = (event: Event): void => {
+        const input = event.target as InputWithModel;
+        input.model.handle.highlight = false;
+        this.requestUpdate();
+    };
+
+    private dispatchChangeEvent(
+        input: HTMLInputElement,
+        handle: SliderHandle
+    ): void {
+        input.valueAsNumber = handle.value;
+
+        const changeEvent = new Event('change', {
+            bubbles: true,
+            composed: true,
+        });
+
+        handle.dispatchEvent(changeEvent);
+    }
+
+    /**
+     * Returns the value under the cursor
+     * @param: PointerEvent on slider
+     * @return: Slider value that correlates to the position under the pointer
+     */
+    private calculateHandlePosition(
+        event: PointerEvent | MouseEvent,
+        model: ModelValue
+    ): number {
+        const rect = this.boundingClientRect;
+        const minOffset = rect.left;
+        const offset = event.clientX;
+        const size = rect.width;
+
+        const normalized = (offset - minOffset) / size;
+        const value = model.normalization.fromNormalized(
+            normalized,
+            model.range.min,
+            model.range.max
+        );
+
+        /* c8 ignore next */
+        return this.host.isLTR ? value : model.range.max - value;
+    }
+
+    public renderHandle(
+        model: ModelValue,
+        index: number,
+        zIndex: number,
+        isMultiHandle: boolean
+    ): TemplateResult {
+        const classes = {
+            handle: true,
+            dragging: this.draggingHandle?.handleName === model.name,
+            'handle-highlight': model.highlight,
+        };
+        const style = {
+            [this.host.isLTR ? 'left' : 'right']: `${
+                model.normalizedValue * 100
+            }%`,
+            'z-index': zIndex.toString(),
+            // Allow setting background per-handle
+            'background-color': `var(--spectrum-slider-handle-background-color-${index}, var(--spectrum-slider-handle-default-background-color))`,
+            'border-color': `var(--spectrum-slider-handle-border-color-${index}, var(-spectrum-slider-handle-default-border-color))`,
+        };
+        const ariaLabelledBy = isMultiHandle ? `label input-${index}` : 'label';
+        return html`
+            <div
+                class=${classMap(classes)}
+                name=${model.name}
+                style=${styleMap(style)}
+                @manage=${streamingListener(
+                    {
+                        type: 'pointerdown',
+                        fn: (event) => this.handlePointerdown(event, model),
+                    },
+                    {
+                        type: 'pointermove',
+                        fn: (event) => this.handlePointermove(event, model),
+                    },
+                    {
+                        type: ['pointerup', 'pointercancel'],
+                        fn: (event) => this.handlePointerup(event, model),
+                    }
+                )}
+                role="presentation"
+            >
+                <input
+                    type="range"
+                    class="input"
+                    id="input-${index}"
+                    min=${model.clamp.min}
+                    max=${model.clamp.max}
+                    step=${model.step}
+                    value=${model.value}
+                    aria-disabled=${ifDefined(
+                        this.host.disabled ? 'true' : undefined
+                    )}
+                    aria-label=${ifDefined(model.ariaLabel)}
+                    aria-labelledby=${ariaLabelledBy}
+                    aria-valuetext=${this.formattedValueForHandle(model)}
+                    @change=${this.onInputChange}
+                    @focus=${this.onInputFocus}
+                    @blur=${this.onInputBlur}
+                    .model=${model}
+                />
+            </div>
+        `;
+    }
+
+    public render(): TemplateResult[] {
+        this.clearHandleComponentCache();
+        return this.model.map((model, index) => {
+            const zIndex = this.handleOrder.indexOf(model.name) + 1;
+            return this.renderHandle(
+                model,
+                index,
+                zIndex,
+                this.model.length > 1
+            );
+        });
+    }
+
+    /**
+     * Returns a list of track segment [start, end] tuples where the values are
+     * normalized to be between 0 and 1.
+     * @returns A list of track segment tuples [start, end]
+     */
+    public trackSegments(): [number, number][] {
+        const values = this.model.map((model) => model.normalizedValue);
+        values.sort((a, b) => a - b);
+
+        // The first segment always starts at 0
+        values.unshift(0);
+        return values.map((value, index, array) => [
+            value,
+            array[index + 1] ?? 1,
+        ]);
+    }
+
+    private updateModel(): void {
+        const handles = [...this.handles.values()];
+
+        const getRangeAndClamp = (index: number): RangeAndClamp => {
+            const handle = handles[index];
+            const previous = handles[index - 1];
+            const next = handles[index + 1];
+
+            const min =
+                typeof handle.min === 'number'
+                    ? handle.min
+                    : (this.host.min as number);
+            const max =
+                typeof handle.max === 'number'
+                    ? handle.max
+                    : (this.host.max as number);
+
+            const result: RangeAndClamp = {
+                range: { min: min, max: max },
+                clamp: { min: min, max: max },
+            };
+
+            if (handle.min === 'previous') {
+                if (previous) {
+                    for (let j = index - 1; j >= 0; j--) {
+                        const item = handles[j];
+                        if (typeof item.min === 'number') {
+                            result.range.min = item.min;
+                            break;
+                        }
+                    }
+                    result.clamp.min = Math.max(
+                        previous.value,
+                        result.range.min
+                    );
+                    /* c8 ignore next 5 */
+                } else {
+                    console.warn(
+                        'First slider handle cannot have attribute min="previous"'
+                    );
+                }
+            }
+            if (handle.max === 'next') {
+                if (next) {
+                    for (let j = index + 1; j < handles.length; j++) {
+                        const item = handles[j];
+                        if (typeof item.max === 'number') {
+                            result.range.max = item.max;
+                            break;
+                        }
+                    }
+                    result.clamp.max = Math.min(next.value, result.range.max);
+                    /* c8 ignore next 5 */
+                } else {
+                    console.warn(
+                        'Last slider handle cannot have attribute max="next"'
+                    );
+                }
+            }
+            return result;
+        };
+
+        const modelValues = handles.map((handle, index) => {
+            const rangeAndClamp = getRangeAndClamp(index);
+            const { toNormalized } = handle.normalization;
+            const clampedValue = Math.max(
+                Math.min(handle.value, rangeAndClamp.clamp.max),
+                rangeAndClamp.clamp.min
+            );
+            const normalizedValue = toNormalized(
+                clampedValue,
+                rangeAndClamp.range.min,
+                rangeAndClamp.range.max
+            );
+            const model = {
+                name: handle.handleName,
+                value: clampedValue,
+                normalizedValue,
+                highlight: handle.highlight,
+                step: handle.step ?? this.host.step,
+                normalization: handle.normalization,
+                handle,
+                ariaLabel:
+                    handle !== this.host && handle?.label.length > 0
+                        ? handle.label
+                        : undefined,
+                ...rangeAndClamp,
+            };
+            return model;
+        });
+
+        this.model = modelValues;
+    }
+}
--- a/packages/slider/src/Slider.ts
+++ b/packages/slider/src/Slider.ts
@@ -16,29 +16,29 @@ import {
     CSSResultArray,
     TemplateResult,
     query,
-    PropertyValues,
     styleMap,
     ifDefined,
+    repeat,
+    classMap,
 } from '@spectrum-web-components/base';
-import { streamingListener } from '@spectrum-web-components/base/src/streaming-listener.js';
 
 import sliderStyles from './slider.css.js';
 import { ObserveSlotText } from '@spectrum-web-components/shared/src/observe-slot-text.js';
-import { Focusable } from '@spectrum-web-components/shared/src/focusable.js';
 import { StyleInfo } from 'lit-html/directives/style-map';
+import { HandleController, HandleValueDictionary } from './HandleController.js';
+import { SliderHandle } from './SliderHandle.js';
 
 export const variants = ['filled', 'ramp', 'range', 'tick'];
 
-export class Slider extends ObserveSlotText(Focusable, '') {
+export class Slider extends ObserveSlotText(SliderHandle, '') {
     public static get styles(): CSSResultArray {
         return [sliderStyles];
     }
 
+    public handleController: HandleController = new HandleController(this);
+
     @property()
     public type = '';
-
-    @property({ type: Number, reflect: true })
-    public value = 10;
 
     @property({ type: String })
     public set variant(variant: string) {
@@ -60,31 +60,45 @@ export class Slider extends ObserveSlotText(Focusable, '') {
         return this._variant;
     }
 
+    public get values(): HandleValueDictionary {
+        return this.handleController.values;
+    }
+
+    public get handleName(): string {
+        return 'value';
+    }
+
     /* Ensure that a '' value for `variant` removes the attribute instead of a blank value */
     private _variant = '';
 
-    @property({ attribute: false })
-    public getAriaValueText: (value: number) => string = (value) => `${value}`;
+    @property({ type: String })
+    public language: string = navigator.language;
 
     @property({ attribute: false })
+    public getAriaValueText: (values: Map<string, string>) => string = (
+        values
+    ) => {
+        const valueArray = [...values.values()];
+        if (valueArray.length === 2)
+            return `${valueArray[0]} - ${valueArray[1]}`;
+        return valueArray.join(', ');
+    };
+
     private get ariaValueText(): string {
         if (!this.getAriaValueText) {
             return `${this.value}`;
         }
-        return this.getAriaValueText(this.value);
+        return this.getAriaValueText(this.handleController.formattedValues);
     }
 
-    @property()
-    public label = '';
+    @property({ type: Boolean, reflect: true, attribute: 'hide-value-label' })
+    public hideValueLabel = false;
 
-    @property({ reflect: true, attribute: 'aria-label' })
-    public ariaLabel?: string;
-
-    @property({ type: Number })
-    public max = 100;
-
-    @property({ type: Number })
+    @property({ type: Number, reflect: true })
     public min = 0;
+
+    @property({ type: Number, reflect: true })
+    public max = 100;
 
     @property({ type: Number })
     public step = 1;
@@ -98,25 +112,15 @@ export class Slider extends ObserveSlotText(Focusable, '') {
     @property({ type: Boolean, reflect: true })
     public disabled = false;
 
-    @property({ type: Boolean, reflect: true })
-    public dragging = false;
-
-    @property({ type: Boolean, reflect: true, attribute: 'handle-highlight' })
-    public handleHighlight = false;
-
-    @query('#handle')
-    private handle!: HTMLDivElement;
-
-    @query('#input')
-    private input!: HTMLInputElement;
-
     @query('#label')
-    private labelEl!: HTMLLabelElement;
+    public labelEl!: HTMLLabelElement;
 
-    private boundingClientRect?: DOMRect;
+    public get numberFormat(): Intl.NumberFormat {
+        return this.getNumberFormat();
+    }
 
     public get focusElement(): HTMLElement {
-        return this.input;
+        return this.handleController.focusElement;
     }
 
     protected render(): TemplateResult {
@@ -125,55 +129,42 @@ export class Slider extends ObserveSlotText(Focusable, '') {
         `;
     }
 
-    protected updated(changedProperties: PropertyValues): void {
-        if (changedProperties.has('value')) {
-            if (this.value === this.input.valueAsNumber) {
-                this.dispatchInputEvent();
-            } else {
-                this.value = this.input.valueAsNumber;
-            }
-        }
+    public connectedCallback(): void {
+        super.connectedCallback();
+        this.handleController.hostConnected();
+    }
+
+    public disconnectedCallback(): void {
+        super.disconnectedCallback();
+        this.handleController.hostDisconnected();
+    }
+
+    public update(changedProperties: Map<string, boolean>): void {
+        this.handleController.hostUpdate();
+        super.update(changedProperties);
     }
 
     private renderLabel(): TemplateResult {
         return html`
             <div id="labelContainer">
-                <label id="label" for="input">
+                <label
+                    id="label"
+                    for=${this.handleController.activeHandleInputId}
+                >
                     ${this.slotHasContent ? html`` : this.label}
                     <slot>${this.label}</slot>
                 </label>
-                <output id="value" aria-live="off" for="input">
+                <output
+                    class=${classMap({
+                        'visually-hidden': this.hideValueLabel,
+                    })}
+                    id="value"
+                    aria-live="off"
+                    for="input"
+                >
                     ${this.ariaValueText}
                 </output>
             </div>
-        `;
-    }
-
-    private renderTrackLeft(): TemplateResult {
-        if (this.variant === 'ramp') {
-            return html``;
-        }
-        return html`
-            <div
-                class="track"
-                id="track-left"
-                style=${styleMap(this.trackStartStyles)}
-                role="presentation"
-            ></div>
-        `;
-    }
-
-    private renderTrackRight(): TemplateResult {
-        if (this.variant === 'ramp') {
-            return html``;
-        }
-        return html`
-            <div
-                class="track"
-                id="track-right"
-                style=${styleMap(this.trackEndStyles)}
-                role="presentation"
-            ></div>
         `;
     }
 
@@ -202,7 +193,8 @@ export class Slider extends ObserveSlotText(Focusable, '') {
             return html``;
         }
         const tickStep = this.tickStep || this.step;
-        const tickCount = (this.max - this.min) / tickStep;
+        const tickCount =
+            ((this.max as number) - (this.min as number)) / tickStep;
         const partialFit = tickCount % 1 !== 0;
         const ticks = new Array(Math.floor(tickCount + 1));
         ticks.fill(0, 0, tickCount + 1);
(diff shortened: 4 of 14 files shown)


cuberoot:
@Westbrook Here's my work in progress so far. It is getting close. I haven't implemented the slider handle constraints yet and I need to work on the tests. It would be great to get early feedback.

packages/slider/src/Slider.ts
```diff
@@ -16,44 +16,69 @@ import {
...
-        if (this.input) {
-            this.input.value = String(value);
-        }
-        const newValue = this.input ? parseFloat(this.input.value) : value;
+    @property()
+    public type = '';
 
-        if (newValue === oldValue) {
-            return;
+    @property({ reflect: true, converter: ValueConverter })
```
Westbrook, line 56: I've had trouble with `@property` working as expected (usually as documented, but you know expectations...) with getter/setter usage. If this works, great, but just something to keep an eye on it.

packages/slider/src/Slider.ts
```diff
@@ -16,44 +16,69 @@ import {
...
+        if (this.handleController.hasExplicitHandles) {
+            return this.handleController.values;
+        } else {
+            return this._value;
         }
+    }
 
-        this._value = newValue;
-        this.requestUpdate('value', oldValue);
+    public set value(value: number | HandleValues) {
```
Westbrook, line 65: I would wonder if `this.handleController.hasExplicitHandles && this.handleController.count` could handle `isNumber` and if `!this.handleController.hasExplicitHandles` could handle `HandleValues.length === 1`. Some logic is too much...but could be interesting.

packages/slider/src/Slider.ts
```diff
@@ -78,18 +103,29 @@ export class Slider extends Focusable {
...
 
     @property({ attribute: false })
-    public getAriaValueText: (value: number) => string = (value) => `${value}`;
+    public getAriaValueText: (values: Map<string, string>) => string = (
+        values
+    ) => {
+        const valueArray = Array.from(values.values());
+        if (valueArray.length === 2)
+            return `${valueArray[0]} - ${valueArray[1]}`;
+        return valueArray.join(', ');
```
Westbrook, line 120: This seems like it'll need some thinking...you're likely already doing that with usage of this commented out in the handle controller.

packages/slider/src/Slider.ts
```diff
@@ -78,18 +103,29 @@ export class Slider extends Focusable {
...
+    };
 
     @property({ attribute: false })
```
Westbrook, line 123: Not your code, but not sure why this is `@property`...

packages/slider/src/Slider.ts
```diff
@@ -143,51 +165,39 @@ export class Slider extends Focusable {
...
     }
 
-    protected updated(changedProperties: PropertyValues): void {
-        if (changedProperties.has('value')) {
-            this.dispatchInputEvent();
-        }
+    public connectedCallback(): void {
+        super.connectedCallback();
+        this.handleController.hostConnected();
+        this.addEventListener(
```
Westbrook, line 171: I feel like this is an implementation for more flexible reuse, which is cool, but we might want to just allow the controller too access `this.host.requestUpdate()` to simplify this for our use case?

packages/slider/src/Slider.ts
```diff
@@ -143,51 +165,39 @@ export class Slider extends Focusable {
...
+
+    private handleSliderHandleUpdate = (): void => {
+        this.requestUpdate();
+    };
+
     private renderLabel(): TemplateResult {
         return html`
             <div id="labelContainer">
-                <label id="label" for="input"><slot>${this.label}</slot></label>
+                <label id="label" for="input">${this.label}</label>
```
Westbrook, line 193: I'd love to find an acceptable way to allow use to keep this `<slot>`. In the combobox research, we've looked at having the `sp-combobox-item` elements sprout a `slot='option'` attribute to allow for this. Thoughts on `<sp-slider-handle>` doing similar? Maybe sprouting `slot="handle"`? This is also a way to support the `assignedElements` note above.

packages/slider/src/Slider.ts
```diff
@@ -243,203 +253,60 @@ export class Slider extends Focusable {
...
             <div @pointerdown=${this.handleTrackPointerdown}>
                 <div id="controls">
-                    ${this.renderTrackLeft()} ${this.renderRamp()}
-                    ${this.renderTicks()} ${this.renderHandle()}
-                    ${this.renderTrackRight()}
+                    ${this.renderTrackSegment(...segments[0])}
+                    ${this.renderRamp()} ${this.renderTicks()}
+                    ${this.handleController.render()}
+                    ${segments
+                        .slice(1)
```
Westbrook, line 278: What are we slicing off here? Is there a way to do that in the controller?

packages/slider/src/SliderHandle.ts
```diff
@@ -0,0 +1,69 @@
...
+    public formatOptions: Intl.NumberFormatOptions = {};
+
+    public abstract get handleName(): string;
+
+    public updated(changedProperties: PropertyValues): void {
+        super.updated(changedProperties);
+    }
+}
+
+export class SliderHandle extends SliderHandleBase {
```
Westbrook, line 47: This likely wants a value for `focusElement`?

packages/slider/src/SliderHandle.ts
```diff
@@ -0,0 +1,69 @@
...
+    @property({ type: Number })
+    value: number = 10;
+
+    public get handleName(): string {
+        return this.name;
+    }
+
+    public updated(changedProperties: PropertyValues): void {
```
Westbrook, line 55: Ah, this isn't in the controller, I see. Interesting.

Westbrook: This never actually _renders_ anything, right? Might be a premature optimization, but in that case we might want to do this work in `shouldUpdate` and `return false;` to work within that context while reducing the call depth of an update.

packages/slider/stories/slider.stories.ts
```diff
@@ -97,12 +143,122 @@ export const Disabled = (): TemplateResult => {
...
+                    name="mid"
+                    value="100"
+                    min="previous"
+                    max="next"
+                ></sp-slider-handle>
+                <sp-slider-handle
+                    name="high"
+                    value="250"
+                    min="previous"
+                ></sp-slider-handle>
```
Westbrook, line 252: I'd love to see `handles` as a Storybook arg here, so that they could be dynamic to the story, but having all three versions available for the VRT is the most important part here, IMO.

Westbrook (commented):
@cuberoot this is such an exciting start! Most of this is more questions and encouragement than any really big requests, so hopefully it doesn't add too much to the process.

One "in the wild" usage to confirm here is in the VRT results project which is failing the builds: https://app.circleci.com/pipelines/github/adobe/spectrum-web-components/6745/workflows/bcc75a34-a311-460b-a448-3136b258304b/jobs/69421/parallel-runs/0/steps/0-103

Westbrook:
When you get to a place where it's possible, can you report the results of:
```
yarn build:tests
yarn test:bench -p slider
```
and
```
yarn test:bench -p slider -s page
```

packages/slider/sp-slider.ts
```diff
@@ -10,11 +10,14 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
 import { Slider } from './src/Slider.js';
+import { SliderHandle } from './src/SliderHandle.js';
 
 customElements.define('sp-slider', Slider);
+customElements.define('sp-slider-handle', SliderHandle);
 
 declare global {
     interface HTMLElementTagNameMap {
         'sp-slider': Slider;
+        'sp-slider-handle': SliderHandle;
```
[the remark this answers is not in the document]
cuberoot: I'm happy to change if you think it is worth adding in an `sp-slider-handle.ts`.

packages/slider/src/HandleController.ts
```diff
@@ -0,0 +1,371 @@
...
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+import { html, TemplateResult, ifDefined } from '@spectrum-web-components/base';
+import { streamingListener } from '@spectrum-web-components/base/src/streaming-listener.js';
+import { Slider } from './Slider';
+import { SliderHandle, HandleValues } from './SliderHandle';
```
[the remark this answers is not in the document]
cuberoot: No justification. I'm just not used to working on projects where the .js is left off. Easy fix.

packages/slider/src/HandleController.ts
```diff
@@ -0,0 +1,371 @@
...
+        // TODO: Implement
+        console.log(values);
+    }
+
+    public get formattedValues(): Map<string, string> {
+        const result = new Map<string, string>();
+        for (const handle of this.model.values()) {
+            const formatOptions =
+                handle.formatOptions ?? this.host.formatOptions;
+            const formattedValue = formatOptions
```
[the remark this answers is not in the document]
cuberoot: 👍

packages/slider/src/HandleController.ts
```diff
@@ -0,0 +1,371 @@
...
+                : handle.value.toString();
+            result.set(handle.handleName, formattedValue);
+        }
+        return result;
+    }
+
+    public get hasExplicitHandles(): boolean {
+        if (this.model.size === 0) return false;
+        if (this.model.size > 1) return true;
+        return this.model.values().next().value !== this.host;
```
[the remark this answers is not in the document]
cuberoot: If the user has explicitly defined handles, then that changes how the `value` attribute on the `sp-slider` behaves.

This comes back to an earlier suggestion of yours that I had implemented. I have a `value` property on the `sp-slider` and `sp-slider-handle` that was always a `number`. I had then added a second property called `values` on the `sp-slider` that had the structured data for all of the sliders. You asked if I could just change the shape of the `value` property.

So, at the moment, there is only a `value` property and it returns a number when there are no explicit `sp-slider-handles`. When there are, it returns the structured `HandleValues` result. This is what is causing the break in the wild that you mentioned above...

I wonder if it would be better to go back to having two attributes? What do you think?

packages/slider/src/HandleController.ts
```diff
@@ -0,0 +1,371 @@
...
+    }
+
+    public get focusElement(): HTMLElement {
```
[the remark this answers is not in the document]
cuberoot: Yah, I wasn't sure what to do here. What I was going for was that, if you click on the label, you should be able to use the keyboard on the handle that was last moved (defaulting to the first handle). The `input` elements are always rendered in the same order, so you seem to be able to tab and shift-tab between them in the correct order. This is going to need some testing.

packages/slider/src/HandleController.ts
```diff
@@ -0,0 +1,371 @@
...
+    }
+
+    public hostDisconnected(): void {
+        this.observer.disconnect();
+    }
+
+    private extractModelFromLightDom = () => {
+        let elements = [
+            ...this.host.querySelectorAll('sp-slider-handle'),
+        ] as SliderHandle[];
```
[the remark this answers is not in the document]
cuberoot: Definitely a more advanced use case than I had envisioned. Worth a follow up later once we get the simple case working.

packages/slider/src/HandleController.ts
```diff
@@ -0,0 +1,371 @@
...
+        elements.forEach((model, index) => {
+            if (!model.handleName?.length) {
+                model.name = `handle${index + 1}`;
+            }
+            this.model.set(model.handleName, model);
+        });
+
+        this.handleOrder = [];
+        for (const name of this.model.keys()) {
+            this.handleOrder.push(name);
```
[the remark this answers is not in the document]
cuberoot: Yep.

packages/slider/src/HandleController.ts
```diff
@@ -0,0 +1,371 @@
...
+        }
+        this.host.requestUpdate();
+    };
+
+    public get activeHandle(): string {
+        return this.handleOrder[0];
```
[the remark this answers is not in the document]
cuberoot: This is dynamic. When a handle is activated, I remove it from `handleOrder` and `unshift` it to the front of the list.

packages/slider/src/HandleController.ts
```diff
@@ -0,0 +1,371 @@
...
+
+    get handles(): Iterator<SliderHandle> {
+        return this.model.values();
+    }
+
+    private getHandleComponents(name: string): HandleComponents {
+        const handle = this.host.shadowRoot.querySelector(
+            `.handle[name="${name}"]`
+        ) as HTMLElement;
+        const input = handle.querySelector('input') as HTMLInputElement;
```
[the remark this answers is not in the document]
cuberoot: Yes, I have been missing having a way to get a ref to a rendered element. Good to hear that one is coming. I didn't want to complicate an already intense code change by adding caching here 😄

packages/slider/src/HandleController.ts
```diff
@@ -0,0 +1,371 @@
...
+        const model = this.model.get(this.activeHandle) as SliderHandle;
+        return { model, handle, input };
+    }
+
+    public get count(): number {
+        return this.model.size;
```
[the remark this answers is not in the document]
cuberoot: Not sure what you are asking here. `model` is a `Map`, so `size` is going to be what I call. It is possible that `model` isn't the right name. I refactored this PR twice before submitting it. The `model` started out as JSON data and turned into a `Map` of the actual slider handles.

packages/slider/src/HandleController.ts
```diff
@@ -0,0 +1,371 @@
...
+         * handlers attached to the slider are invoked before input handlers
+         */
+        event.stopPropagation();
+        const syntheticPointerEvent = new PointerEvent('pointerdown', event);
+        handle.dispatchEvent(syntheticPointerEvent);
+
+        input.value = this.calculateHandlePosition(event).toString();
+        model.value = parseFloat(input.value);
+        this.host.requestUpdate();
+        this.dispatchInputEvent(model);
```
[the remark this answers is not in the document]
cuberoot: I have merged your changes across to `beginTrackDrag` in `HandleController`. I'll push that soon.

packages/slider/src/HandleController.ts
```diff
@@ -0,0 +1,371 @@
...
+    }
+
+    /**
+     * Returns the value under the cursor
+     * @param: PointerEvent on slider
+     * @return: Slider value that correlates to the position under the pointer
+     */
+    private calculateHandlePosition(event: PointerEvent | MouseEvent): number {
+        if (!this.boundingClientRect) {
+            return 0; //return this.value;
```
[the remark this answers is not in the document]
cuberoot: I just moved to using a getter instead (DRY principle):
```typescript
   private get boundingClientRect(): DOMRect {
        return this.host.getBoundingClientRect();
    }
```

packages/slider/sp-slider.ts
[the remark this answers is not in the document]
Westbrook: Will reserve judgement till the very end, but possibly.

packages/slider/src/HandleController.ts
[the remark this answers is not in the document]
Westbrook: It makes perfect sense later on in the review when I actually see it used, so really just a comment it fine for me.

packages/slider/src/HandleController.ts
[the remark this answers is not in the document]
Westbrook: 🤯 Great!

packages/slider/src/HandleController.ts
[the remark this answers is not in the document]
Westbrook: Just looking at it is weird. `HandleController.count` reaches in to `HandleController.model.size`. Is `size` for both feels weird, I'm not worried about it that much.

packages/slider/src/HandleController.ts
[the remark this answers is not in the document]
Westbrook: One thing to take into account is that `.getBoundingClientRect()` requires style recalc, so being able to cache it is really valuable.

packages/slider/src/HandleController.ts
```diff
@@ -0,0 +1,371 @@
...
+        const classes = ['handle'];
+        if (this.draggingHandle === model) {
+            classes.push('dragging');
+        }
+        if (model.highlight) {
+            classes.push('handle-highlight');
+        }
+        return html`
+            <div
+                class=${classes.join(' ')}
```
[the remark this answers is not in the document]
cuberoot: Excellent, I thought there should be something like that, but I missed it in my original search

packages/slider/src/HandleController.ts
```diff
@@ -0,0 +1,371 @@
...
+        );
+    }
+
+    public trackSegments(): [number, number][] {
+        const values = [...this.model.values()].map((item) => item.value);
+        values.sort((a, b) => a - b);
+        const normalizedValues = values.map((value) =>
+            this.normalizeValue(value)
+        );
+        normalizedValues.unshift(0);
```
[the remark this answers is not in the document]
cuberoot: Lets say we have a slider with min=0 and max=10 and handles at 3 and 6. There will be three track segments: `[0,3], [3,6], and [6,10]`. Normalized (between 0 and 1) this will look like: `[0,0.3], [0.3,0.6], and [0.6,1]`.

The `normalizedValues.unshift(0)` puts the starting value of `0` into the list of segment start values. Then, we iterate over the start values of the segment and product tuples of the `[start, end]` values.

I will change a variable name and add a comment to make it more clear

packages/slider/src/HandleController.ts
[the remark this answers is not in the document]
cuberoot: Ok, added caching back in...

packages/slider/src/SliderHandle.ts
cuberoot: I'll change it.

packages/slider/src/HandleController.ts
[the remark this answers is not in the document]
cuberoot: Comment added

packages/slider/src/Slider.ts
cuberoot: So, the CSS assumes that the first and last things rendered are track segments. So, I render the first segment explicitly and then render the remaining ones at the end. The `slice(1)` just takes all of the remaining segments after the first one.

packages/slider/src/Slider.ts
cuberoot: Ok, giving that a try.

packages/slider/src/SliderHandle.ts
cuberoot: Yes, I'll have to work that one out...

cuberoot:
> When you get to a place where it's possible, can you report the results of:
> 
> ```
> yarn build:tests
> yarn test:bench -p slider
> ```

```
┌─────────────┬───────────────────┐
│   Benchmark │ slider:test-basic │
├─────────────┼───────────────────┤
│     Browser │ chrome-headless   │
│             │ 90.0.4430.93      │
├─────────────┼───────────────────┤
│ Sample size │ 50                │
└─────────────┴───────────────────┘

┌─────────┬────────────┬───────────────────┬───────────────────┬───────────────────┐
│ Version │ Bytes      │          Avg time │         vs remote │               vs  │
├─────────┼────────────┼───────────────────┼───────────────────┼───────────────────┤
│ remote  │ 352.12 KiB │ 69.16ms - 69.79ms │                   │            faster │
│         │            │                   │          -        │         25% - 26% │
│         │            │                   │                   │ 22.99ms - 23.83ms │
├─────────┼────────────┼───────────────────┼───────────────────┼───────────────────┤
│ <none>  │ 360.89 KiB │ 92.61ms - 93.16ms │            slower │                   │
│         │            │                   │         33% - 34% │          -        │
│         │            │                   │ 22.99ms - 23.83ms │                   │
└─────────┴────────────┴───────────────────┴───────────────────┴───────────────────┘


Update available!
The latest version of tachometer is 0.5.9
You are running version 0.5.8
See what's new at https://github.com/Polymer/tachometer/blob/master/CHANGELOG.md
slider:test-basic: 69.16ms - 69.79ms
slider:test-basic: 92.61ms - 93.16ms
✨  Done in 48.53s.
```

> 
> and
> 
> ```
> yarn test:bench -p slider -s page
> ```
```

┌─────────────┬───────────────────┐
│   Benchmark │ slider:test-basic │
├─────────────┼───────────────────┤
│     Browser │ chrome-headless   │
│             │ 90.0.4430.93      │
├─────────────┼───────────────────┤
│ Sample size │ 50                │
└─────────────┴───────────────────┘

┌─────────┬────────────┬─────────────────────┬───────────────────┬───────────────────┐
│ Version │ Bytes      │            Avg time │         vs remote │               vs  │
├─────────┼────────────┼─────────────────────┼───────────────────┼───────────────────┤
│ remote  │ 352.12 KiB │ 117.35ms - 118.22ms │                   │            slower │
│         │            │                     │          -        │         25% - 27% │
│         │            │                     │                   │ 23.68ms - 24.88ms │
├─────────┼────────────┼─────────────────────┼───────────────────┼───────────────────┤
│ <none>  │ 360.89 KiB │   93.09ms - 93.92ms │            faster │                   │
│         │            │                     │         20% - 21% │          -        │
│         │            │                     │ 23.68ms - 24.88ms │                   │
└─────────┴────────────┴─────────────────────┴───────────────────┴───────────────────┘


Update available!
The latest version of tachometer is 0.5.9
You are running version 0.5.8
See what's new at https://github.com/Polymer/tachometer/blob/master/CHANGELOG.md
slider:test-basic: 117.35ms - 118.22ms
slider:test-basic: 93.09ms - 93.92ms
✨  Done in 30.15s.
```

packages/slider/src/Slider.ts
Westbrook: Interesting. I'd want to check whether slicing vs skipping had any effect on the way the template was addressed in subsequent renders, being the slice (IIRC) makes a new array identity every time.

Westbrook:
@cuberoot these perf profiles are something I'd really like to see clarified before we could accept this into the main library. 

`yarn test:bench -p slider` analyzes upgrade and render time of the element, and here it's showing a 33% - 34% degradation of performance. I'd say a little slower (at the benefit of the added functionality) would be acceptable, but this is a pretty solid drop. Further, it seems exacerbated when you use `sp-slider-handle` to outline the one handle. In that case, I'm seeing the following (further degraded) perf locally:
```
┌─────────────┬───────────────────┐
│   Benchmark │ slider:test-basic │
├─────────────┼───────────────────┤
│     Browser │ chrome-headless   │
│             │ 90.0.4430.93      │
├─────────────┼───────────────────┤
│ Sample size │ 50                │
└─────────────┴───────────────────┘

┌─────────┬────────────┬─────────────────────┬───────────────────┬───────────────────┐
│ Version │ Bytes      │            Avg time │         vs remote │               vs  │
├─────────┼────────────┼─────────────────────┼───────────────────┼───────────────────┤
│ remote  │ 352.29 KiB │ 108.09ms - 115.22ms │                   │            faster │
│         │            │                     │          -        │         37% - 41% │
│         │            │                     │                   │ 66.46ms - 76.57ms │
├─────────┼────────────┼─────────────────────┼───────────────────┼───────────────────┤
│ <none>  │ 363.37 KiB │ 179.59ms - 186.75ms │            slower │                   │
│         │            │                     │         58% - 70% │          -        │
│         │            │                     │ 66.46ms - 76.57ms │                   │
└─────────┴────────────┴─────────────────────┴───────────────────┴───────────────────┘
```

`yarn test:bench -p slider -s page` outlines the full load, define, upgrade, render process, and I'm somewhat perplexed that it's showing about the same perf profile as that without load/define. I'm quite pleased to see this profile flip back and prefer the new version by 20% - 21%, however. It could be that this comes from splitting the JS into multiple imports, or from our `updateComplete` not being appropriately measured (which could effect both profiles). I'm not sure a concrete next step there just now, but something we'll need to revisit as the functionality here settles.

cuberoot:
Ok, this is going to have to sit for a bit. I have to take what I have so far and work with it for now. I'll try to come back to this after next week.

packages/slider/src/HandleController.ts
```diff
@@ -0,0 +1,371 @@
...
+
+    private handlePosition(model: SliderHandle): number {
+        const range = this.host.max - this.host.min;
+        const progress = model.value - this.host.min;
+
+        return progress / range;
+    }
+
+    private handleStyle(model: SliderHandle): string {
+        return `${this.host.isLTR ? 'left' : 'right'}: ${
```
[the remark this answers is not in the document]
cuberoot: Yep, `styleMap` came into play.

packages/slider/src/Slider.ts
```diff
@@ -16,44 +16,69 @@ import {
...
-    @property()
-    public type = '';
+    private handleController!: HandleController;
 
-    @property({ type: Number, reflect: true })
-    public get value(): number {
-        return this._value;
+    constructor() {
+        super();
+        this.handleController = new HandleController(this);
```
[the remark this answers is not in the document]
cuberoot: No, that was a mental leftover of using an old proprietary framework where the semantics were different. There's no reason not to have the value in an initialiser above in modern JS

packages/slider/src/Slider.ts
cuberoot: Good to know. I think I got the idea from the original slider code. I will watch it like a hawk 😄

packages/slider/src/Slider.ts
cuberoot: Will try removing the `@property`

packages/slider/src/Slider.ts
cuberoot: Are you talking about the event listener? It is listening to the slider handle, which the user owns in case they set the value programatically.

packages/slider/src/Slider.ts
Westbrook: It was mainly that either reflection or render didn't happen without actually telling it to. Might be good to focus a test on those two side effects here.

packages/slider/src/Slider.ts
Westbrook: Ah, ok,  that makes sense.

cuberoot:
@Westbrook I have made a lot of changes. My latest benchmark is below:

```
┌─────────────┬───────────────────┐
│   Benchmark │ slider:test-basic │
├─────────────┼───────────────────┤
│     Browser │ chrome-headless   │
│             │ 90.0.4430.212     │
├─────────────┼───────────────────┤
│ Sample size │ 50                │
└─────────────┴───────────────────┘

┌─────────┬────────────┬───────────────────┬──────────────────┬──────────────────┐
│ Version │ Bytes      │          Avg time │        vs remote │              vs  │
├─────────┼────────────┼───────────────────┼──────────────────┼──────────────────┤
│ remote  │ 353.82 KiB │ 76.91ms - 79.02ms │                  │           faster │
│         │            │                   │         -        │         9% - 12% │
│         │            │                   │                  │ 7.73ms - 10.65ms │
├─────────┼────────────┼───────────────────┼──────────────────┼──────────────────┤
│ <none>  │ 364.88 KiB │ 86.14ms - 88.16ms │           slower │                  │
│         │            │                   │        10% - 14% │         -        │
│         │            │                   │ 7.73ms - 10.65ms │                  │
└─────────┴────────────┴───────────────────┴──────────────────┴──────────────────┘
```

It is an improvement over the previous benchmark and I hope it is more acceptable. However, I did [cheat slightly](https://github.com/adobe/spectrum-web-components/blob/7901a0ce9d3754471b4acba11e43f37fd7c6aa77/packages/slider/test/benchmark/test-basic.ts#L25) to get there. It turns out that `Intl.formatNumber` is surprisingly expensive. It accounted for most of the degradation in performance. So, in order to get a better sense of the performance changes related to the structure of the component, I bypassed it in the performance test.

packages/slider/src/SliderHandle.ts
cuberoot: Slider handles now supply a `focusElement` correctly

packages/slider/src/Slider.ts
cuberoot: I reworked this to make sure that the various parts of rendered HTML have appropriate ids. I am not sure I gained much in performance. My ad hoc testing showed there was possibly a small difference.

cuberoot:
@Westbrook I have updated the tests and docs and I think I have addressed all of your comments. Please take another close look at it as I have refactored heavily since your last review. It is a bit simpler now.

I did eventually decide to stop overloading the `value` attribute/property on `sp-slider`. I had it so that it had structured data when there were multiple handles and only a number when there was a single implicit handle. However, as you pointed out, the getter and setter for `value` causes issues (which I guess is why you removed them). Users will probably use the `value` attributes on the individual handles anyway. There's now a `values` property on the `sp-slider` that will return all of the values.

packages/slider/src/HandleController.ts
```diff
@@ -0,0 +1,540 @@
...
+        const classes = {
+            handle: true,
+            dragging: this.draggingHandle?.handleName === model.name,
+            'handle-highlight': model.highlight,
+        };
+        const style = {
+            [this.host.isLTR ? 'left' : 'right']: `${
+                model.normalizedValue * 100
+            }%`,
+            'z-index': zIndex.toString(),
```
Westbrook, line 385: Just an FYI, but, `styleMap()` does accept JS values for CSS properties, if you wanted.

packages/slider/src/HandleController.ts
```diff
@@ -0,0 +1,540 @@
...
+        return values.map((value, index, array) => [
+            value,
+            array[index + 1] ?? 1,
+        ]);
+    }
+
+    private updateModel(): void {
+        const handles = [...this.handles.values()];
+
+        const getRangeAndClamp = (index: number): RangeAndClamp => {
```
Westbrook, line 460: This is a pretty smart function. Almost enough to want to have it moved somewhere it could be individually unit tested. Not a blocker.

packages/slider/src/Slider.ts
Westbrook: This doesn't exactly feel related to this line anymore, any thoughts on what I was going on about here?

packages/slider/src/HandleController.ts
```diff
@@ -0,0 +1,540 @@
...
+                        type: ['pointerup', 'pointercancel'],
+                        fn: (event) => this.handlePointerup(event, model),
+                    }
+                )}
+                role="presentation"
+            >
+                <input
+                    type="range"
+                    class="input"
+                    id=${ifDefined(isActive ? 'input' : undefined)}
```
Westbrook, line 414: We likely should switch from the current `<label for="id"/><output for="id"/><input id="id" />` style of labeling to a `<label id="id1" /><label id="id1" /><input aria-labelledby="id1 id2" />` style so that any number of inputs can reference the `label` and `output` element at the same time.

packages/slider/src/Slider.ts
```diff
@@ -228,194 +206,65 @@ export class Slider extends Focusable {
...
+            })),
+        ];
+
         return html`
             <div @pointerdown=${this.handleTrackPointerdown}>
                 <div id="controls">
-                    ${this.renderTrackLeft()} ${this.renderRamp()}
-                    ${this.renderTicks()} ${this.renderHandle()}
-                    ${this.renderTrackRight()}
+                    ${repeat(
```
Westbrook, line 239: Is `repeat` to support them changing order? Or something else?

packages/slider/src/Slider.ts
```diff
@@ -124,52 +120,33 @@ export class Slider extends Focusable {
...
-                id="track-right"
-                style=${styleMap(this.trackEndStyles)}
-                role="presentation"
-            ></div>
+            <div id="labelContainer">
+                <label id="label" for="input"><slot>${this.label}</slot></label>
+                ${this.hideValueLabel
+                    ? ''
+                    : html`
+                          <output id="value" aria-live="off" for="input">
```
Westbrook, line 145: We should use something like the following on this instead here:
```
.visually-hidden {
  clip: rect(0 0 0 0);
  clip-path: inset(50%);
  height: 1px;
  overflow: hidden;
  position: absolute;
  white-space: nowrap;
  width: 1px;
}
```
This will allow the output to still be available for the screen reader.

Westbrook: I've often thought to add this to our `styles` package and then ended up not fully needing the rule in the project I was working on, for instance.

packages/slider/stories/slider.stories.ts
```diff
@@ -97,12 +145,234 @@ export const Disabled = (): TemplateResult => {
...
+            <sp-slider
+                step="0.5"
+                min="0"
+                max="20"
+                label="Intensity"
+                @input=${handleEvent}
+                @change=${handleEvent}
+                ...=${spreadProps(args)}
+            >
+                <sp-slider-handle slot="handle" value="5"></sp-slider-handle slot="handle">
```
Westbrook, line 176: Double `slot`.

packages/slider/stories/slider.stories.ts
```diff
@@ -97,12 +145,234 @@ export const Disabled = (): TemplateResult => {
...
+                value="5"
+                step="1"
+                min="0"
+                max="255"
+                @input=${handleEvent}
+                @change=${handleEvent}
+                ...=${spreadProps(args)}
+            >
+                Output Levels
+                <sp-slider-handle slot="handle" name="min" value="5"></sp-slider-handle slot="handle">
```
Westbrook, line 205: `slot`

packages/slider/stories/slider.stories.ts
```diff
@@ -97,12 +145,234 @@ export const Disabled = (): TemplateResult => {
...
+                ></sp-slider-handle slot="handle">
+            </sp-slider>
+        </div>
+    `;
+};
+ThreeHandlesOrdered.args = {
+    tickStep: 10,
+};
+
+// This is a very complex example from an actual application.
```
Westbrook, line 262: Can you expand on how this work, I'll have no way of known if I break this in the future or not.

Also, what would you consider the "strangest" values this could produce? It might be good to set that as the default so that the VRT can support catching regressions here.

packages/slider/src/HandleController.ts
```diff
@@ -0,0 +1,540 @@
...
+                    /* c8 ignore next 3 */
+                    console.warn(
+                        'First slider handle cannot have attribute min="previous"'
+                    );
+                }
+            }
+            if (handle.max === 'next') {
+                if (next) {
+                    for (let j = index + 1; j < handles.length; j++) {
+                        const next = handles[j];
```
Westbrook, line 496: JS doesn't mind using `next` here in different scopes, but it's a bit hard to track why we have next but then get a new next 🤔

packages/slider/stories/slider.stories.ts
```diff
@@ -97,12 +145,234 @@ export const Disabled = (): TemplateResult => {
...
+    };
+    const computeGray = (value: number): string => {
+        let result = 1.0;
+        if (value > 0.5) {
+            result = Math.max(2 * (1 - value), 0.01);
+        } else if (value < 0.5) {
+            result = ((1 - 2 * value) * (Math.sqrt(9.99) - 1) + 1) ** 2;
+        }
+        const formatOptions = {
+            maximumFractionDigits: 2,
```
Westbrook, line 331: You might want `minimumFractionDigits: 2` here as well so that the text doesn't jump around so much. Though I guess that makes it display `1.00` vs `1` which might also not be appropriate.

packages/slider/stories/slider.stories.ts
Westbrook: It's quite cool that we can do this now, however!

packages/slider/test/benchmark/test-basic.ts
```diff
@@ -22,5 +22,6 @@ measureFixtureCreation(html`
         max="100"
         label="Opacity"
         id="opacity-slider"
+        .getAriaHandleText=${(value: number) => value.toString()}
```
Westbrook, line 25: I appreciate leveraging this in a proof context, but we should gate as outlined above or test the raw perf here going forward.

projects/vrt-compare/src/OnionSkinner.ts
```diff
@@ -32,7 +32,7 @@ export class OnionSkinner extends SpectrumElement {
     private rightThumbnail?: HTMLImageElement;
 
     private handleOnionInput(event: Event & { target: Slider }) {
-        this.onionLevel = event.target.value;
+        this.onionLevel = event.target.value as number;
```
Westbrook, line 35: 🙏🏼

Westbrook (commented):
This is very exciting, and looks even more awesome!

I need to get into the tests tomorrow. Seems like not a lot of test for the amount of changes, but I've not actually looked at it so far.

packages/slider/sp-slider.ts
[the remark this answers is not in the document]
Westbrook: Speaking through this a little:

Having this together means that if you want to use an `sp-slider` element but for some reason want to use a custom/extended `sp-slider-handle` element (possibly one that closes over some of the complexities you've shown possible in the three handle demos), a consumer would _have_ to extend and use a new element name. This works fine because you leverage `this.host.querySelectorAll('[slot="handle"]')` in the model resolver, so I don't think we have a strong need to move this. It is a little outside of other usage, none of the other paired elements register in a single file, but I'm not sure it's that big of a deal.

If you were to read this and decide "let's move it to `sp-slider-handle.ts`!", do take into account that the new file would need to be added to the `exports` map in `package.json` were you to do that.

packages/slider/README.md
```diff
@@ -81,6 +81,98 @@ import { Slider } from '@spectrum-web-components/slider';
 <sp-slider label="Slider Label - Disabled" variant="ramp" disabled></sp-slider>
 ```
 
+### Range
+
+```html
+<div style="width: 500px; margin: 12px 20px;">
```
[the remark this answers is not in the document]
cuberoot: To be honest, I'm not sure why that was there. Must have been a copy-paste mistake. Deleting that

packages/slider/sp-slider.ts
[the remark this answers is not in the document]
cuberoot: Ok, broke down and created `sp-slider-handle.ts`

packages/slider/src/SliderHandle.ts
```diff
@@ -0,0 +1,133 @@
...
+
+    @property({ type: Object, attribute: 'format-options' })
+    public formatOptions?: Intl.NumberFormatOptions;
+
+    @property({ attribute: false })
+    public getAriaHandleText: (
+        value: number,
+        formatOptions: Intl.NumberFormatOptions
+    ) => string = (value, formatOptions) => {
+        return new Intl.NumberFormat(navigator.language, formatOptions).format(
```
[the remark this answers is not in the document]
cuberoot: I implemented formatter caching

packages/slider/README.md
```diff
@@ -81,6 +81,98 @@ import { Slider } from '@spectrum-web-components/slider';
 <sp-slider label="Slider Label - Disabled" variant="ramp" disabled></sp-slider>
 ```
 
+### Range
```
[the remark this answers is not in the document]
cuberoot: I added `slider-handle.md` as suggested. That's a good idea.

packages/slider/README.md
```diff
@@ -81,6 +81,98 @@ import { Slider } from '@spectrum-web-components/slider';
...
+<div style="width: 500px; margin: 12px 20px;">
+    <sp-slider
+        variant="range"
+        step="1"
+        min="0"
+        max="255"
+    >
+        Output Levels
+        <sp-slider-handle slot="handle" name="min" value="5"></sp-slider-handle slot="handle">
+        <sp-slider-handle slot="handle" name="max" value="250"></sp-slider-handle slot="handle">
```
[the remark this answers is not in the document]
cuberoot: Fixed.

packages/slider/README.md
```diff
@@ -81,6 +81,98 @@ import { Slider } from '@spectrum-web-components/slider';
...
+        name="mid"
+        value="100"
+        min="previous"
+        max="next"
+    ></sp-slider-handle slot="handle">
+    <sp-slider-handle slot="handle"
+        name="high"
+        value="250"
+        min="previous"
+    ></sp-slider-handle slot="handle">
```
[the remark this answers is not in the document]
cuberoot: Fixed

packages/slider/README.md
```diff
@@ -81,6 +81,98 @@ import { Slider } from '@spectrum-web-components/slider';
...
+    min="0"
+    max="255"
+>
+    Output Levels
+    <sp-slider-handle slot="handle"
+        name="low"
+        value="5"
+        max="next"
+    ></sp-slider-handle slot="handle">
+    <sp-slider-handle slot="handle"
```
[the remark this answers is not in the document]
cuberoot: This was a regression. I have fixed it.

cuberoot:
I made some progress today. I'll tackle more tomorrow!

packages/slider/test/benchmark/test-basic.ts
cuberoot: Removed

packages/slider/src/HandleController.ts
cuberoot: Agreed. I don't like variable shadowing either. Good catch...fixed.

packages/slider/src/Slider.ts
```diff
@@ -59,32 +58,52 @@ export class Slider extends Focusable {
...
+        return 'value';
+    }
+
     /* Ensure that a '' value for `variant` removes the attribute instead of a blank value */
     private _variant = '';
 
-    @property({ attribute: false })
-    public getAriaValueText: (value: number) => string = (value) => `${value}`;
+    @property({ type: String, reflect: true })
+    public language: string = navigator.language;
```
cuberoot, line 73: As I mentioned in #1495, my application doesn't necessarily use the value from `navigator.language`. So, I added this here. It seems a little painful to have to provide this to each component though. I wonder if there's a better API we could provide to allow override of `navigator.language` for all SWC components on the page?

packages/slider/src/Slider.ts
cuberoot: I changed the value attribute to be simpler

packages/slider/src/Slider.ts
cuberoot: I changed to make `value` a simple number attribute

packages/slider/src/Slider.ts
cuberoot: 🤔 Not sure what this was about

packages/slider/stories/slider.stories.ts
cuberoot: I am adding a long explanation. You can decide if it is worth keeping. I have changed the values so that they exercise the most complex part of the calculations by default.

packages/slider/stories/slider.stories.ts
cuberoot: Yep, that looks better...

projects/vrt-compare/src/OnionSkinner.ts
cuberoot: We don't need this anymore since I changed `value` back to a simple number attribute.

packages/slider/package.json
```diff
@@ -24,7 +24,8 @@
         "./src/*": "./src/*",
         "./package.json": "./package.json",
         "./sp-slider": "./sp-slider.js",
-        "./sp-slider.js": "./sp-slider.js"
+        "./sp-slider.js": "./sp-slider.js",
+        "./sp-slider-handle.js": "./sp-slider-handle.js"
```
Westbrook, line 28: We should include the no extension on the left version as well.

packages/slider/src/SliderHandle.ts
```diff
@@ -0,0 +1,163 @@
...
+        | undefined;
+    protected getNumberFormat(): Intl.NumberFormat {
+        const language = this.handleController?.language ?? navigator.language;
+        if (
+            !this._numberFormatCache ||
+            language !== this._numberFormatCache.language
+        ) {
+            this._numberFormatCache = {
+                language,
+                numberFormat: new Intl.NumberFormat(
```
Westbrook, line 150: We use https://www.npmjs.com/package/@internationalized/number to better align Number Field to other Adobe offerings, should we do the same here?

packages/slider/src/Slider.ts
Westbrook: Does this specifically need to reflect?

packages/slider/README.md
```diff
@@ -81,6 +81,98 @@ import { Slider } from '@spectrum-web-components/slider';
...
+
+## Advanced normalization
+
+By default, `sp-slider` assumes a linear scale between the `min` and `max` values.
+For advanced applications, it is sometimes necessary to specify a custom
+"normalization."
+
+Normalization is the process of converting a slider to a value between 0 and 1 where
+0 represents the minimum and 1 represents the maximum. See ["Three Handles Complex"
+example in the playground](storybook/index.html?path=/story/slider--three-handles-complex).
```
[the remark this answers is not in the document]
cuberoot: I am working around this by using an actual `<a target="_blank" ...>` tag to open it in another tab.

packages/slider/README.md
```diff
@@ -81,6 +81,98 @@ import { Slider } from '@spectrum-web-components/slider';
...
+the `getAriaHandleText` property on an `sp-slider` or `sp-slider-handle`. Or,
+for a multi-handle slider, you can format the combined value label for all
+handles by passing a formatting function to the `getAriaValueText` property
+on the parent `sp-slider`.
+
+You can suppress the value label altogether using the `hide-value-label`
+attribute.
+
+```html
+<sp-slider hide-value-label></sp-slider>
```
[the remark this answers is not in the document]
cuberoot: Adding label

packages/slider/src/HandleController.ts
```diff
@@ -0,0 +1,540 @@
...
+    }
+
+    public inputForHandle(handle: SliderHandle): HTMLInputElement | undefined {
+        if (this.handles.has(handle.handleName)) {
+            const { input } = this.getHandleElements(handle);
+            return input;
+        }
+
+        /* c8 ignore next */
+        return undefined;
```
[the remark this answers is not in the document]
cuberoot: Changing to an exception. We shouldn't hit here.

packages/slider/src/HandleController.ts
```diff
@@ -0,0 +1,540 @@
...
+                    class="input"
+                    id=${ifDefined(isActive ? 'input' : undefined)}
+                    min=${model.clamp.min}
+                    max=${model.clamp.max}
+                    step=${model.step}
+                    value=${model.value}
+                    aria-disabled=${ifDefined(
+                        this.host.disabled ? 'true' : undefined
+                    )}
+                    aria-valuetext=${this.formattedValueForHandle(model)}
```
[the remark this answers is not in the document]
cuberoot: That is a good question. Are there some accessibility folks that we ask for advice? I think we need to give some indication of which handle you are interacting with...

packages/slider/src/HandleController.ts
cuberoot: I'm not sure what change you are proposing here (although happy to learn!). 

```javascript
[this.host.isLTR ? 'left' : 'right']: `${
    model.normalizedValue * 100 
}%`,
```
This is a JS value, isn't it? I need the template string to stick the `%` on the end.

packages/slider/src/HandleController.ts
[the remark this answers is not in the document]
Westbrook: @jnurthen thoughts on where we might back up our labeling approach here?

packages/slider/src/HandleController.ts
```diff
@@ -0,0 +1,540 @@
...
+                    @change=${(event: Event) =>
+                        this.onInputChange(event, model)}
+                    @focus=${(event: Event) => this.onInputFocus(event, model)}
+                    @blur=${() => this.onInputBlur(model)}
```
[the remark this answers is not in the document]
cuberoot: I wouldn't have thought of that. Taking the model on the input and then referencing the class members directly does knock off a couple of ms.

packages/slider/src/HandleController.ts
```diff
@@ -0,0 +1,540 @@
...
+                        if (typeof next.max === 'number') {
+                            result.range.max = next.max;
+                            break;
+                        }
+                    }
+                    result.clamp.max = Math.min(next.value, result.range.max);
+                } else {
+                    /* c8 ignore next 3 */
+                    console.warn(
+                        'First slider handle cannot have attribute max="next"'
```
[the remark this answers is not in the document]
cuberoot: Ah yep!

packages/slider/src/Slider.ts
cuberoot: Using `repeat` just allowed me to specifically assign an id to each element rendered to make sure that lit can track them. Previously I was using the `slice` directly in the rendering and you were concerned about that affecting the performance on subsequent renders.

packages/slider/src/HandleController.ts
cuberoot: I am a bit confused. Your suggestion above seems to show multiple labels for one input. Whereas I we have multiple inputs for one label. Am I missing something? 😕

packages/slider/src/HandleController.ts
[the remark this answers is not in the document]
jnurthen: @Westbrook have you looked at the react-spectrum RangeSlider? https://react-spectrum.adobe.com/react-spectrum/RangeSlider.html

packages/slider/src/Slider.ts
cuberoot: Nope.

packages/slider/package.json
cuberoot: Good catch!

packages/slider/src/SliderHandle.ts
cuberoot: Done.

packages/slider/src/HandleController.ts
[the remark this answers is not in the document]
cuberoot: I have pushed a reworking of the labels based on the React Spectrum implementation. A basic summary is:
- the `<input>` for each handle is labelled by the slider's label and by itself
- there is an `aria-label` attribute on the `sp-slider-handle` that is passed to the `input`

See the range slider example on the slider-handle docs page to play with it.

The one difference is that React Spectrum automatically adds the `aria-labels` "Minimum" and "Maximum" to the inputs on the range slider. I have not done that since our solution is a little more free-form. But I could add that default for the two handle case if it made sense.

packages/slider/src/HandleController.ts
cuberoot: Ok, that looks similar to what I did based on React Spectrum

packages/slider/src/SliderHandle.ts
```diff
@@ -0,0 +1,170 @@
...
+    public max?: number | 'next';
+
+    @property({ type: Number, reflect: true })
+    public step?: number;
+
+    @property({ type: Object, attribute: 'format-options' })
+    public formatOptions?: NumberFormatOptions;
+
+    @property({ type: String, attribute: 'aria-label' })
+    public ariaLabel?: string;
```
Westbrook, line 105: This need to surface as the `label` attribute as any `aria-*` attribute will feed directly into the accessibility tree. And, while we want that on the `<input>` element, we don't want that on the `sp-slider-handle` element.

Westbrook: It would be great to use the attribute in the stories and documentation, too.

packages/slider/src/HandleController.ts
Westbrook: Not that important, but you can do `zIndex: zIndex.toString(),`, IIRC. So, `zIndex` the JS property for the the property...never mind me.

jnurthen (commented):
Has any thought been put into what the behaviour should be when the minimum becomes the maximum? Currently the user can switch the slider handles which creates a confusing situation. react-spectrum limits the minimum value to be less than the maximum value so this situation does not occur

packages/slider/src/SliderHandle.ts
cuberoot: Any suggestions here? We could have a `label` attribute on `sp-slider-handle` that is passed through to `aria-label` on the `input`. It's a bit confusing to the developer using the component.

I did have one example in the docs [here](https://github.com/adobe/spectrum-web-components/blob/cuberoot/multi-handle-slider/packages/slider/slider-handle.md#L35), but more should be added.

cuberoot:
> Has any thought been put into what the behaviour should be when the minimum becomes the maximum? Currently the user can switch the slider handles which creates a confusing situation. react-spectrum limits the minimum value to be less than the maximum value so this situation does not occur

We allow the handles to slide past each other by default because it is actually a use case in our software. However, you can specify `min="previous"` or `max="next"` to have the handles be limited according to their neighbour.

packages/slider/src/SliderHandle.ts
Westbrook: `<my-el label="Text">` is generally what we do here across the library. 

Can you expand on the confusion? I'm not sure how it's a different amount of confusion that placing `aria-label` on an element and then it being applied to two different things in the accessibility tree.

Screen reader reporting of the Maximum Handle in the demo

Beyond the attribute naming, this seems to interact with the screen reader quite nicely! If we can at least have  both versions here: https://cuberoot-multi-handle-slider--spectrum-web-components.netlify.app/components/slider-handle leverage this attribute, I think we'll be good for iteration 1.

cuberoot: Ok, I will make it so. The confusion I meant was setting a `label` attribute that would never be rendered (only for aria). However, I think that is the best solution

Westbrook:
We're very close here, some small (hopefully 🤞🏼 ) final things:
- if I click and drag a handle the associated `<input>` is no longer focused, so a user can't immediate use the arrow keys to update the value after the initial drag, this is a regression. This is likely due to the way we've changed the ID relationship to the label. Previously we could click the label to achieve focus because it only related to a single input. Check out https://github.com/adobe/spectrum-web-components/pull/1140 for other things that could regress while addressing this.
- there seems to be a _slight_ variation on the vertical placement of the track when there are multiple handles:
Misaligned track UI
- all of the storybook stories need to have labels (Explicit Handle, Three Handles Ordered, Three Handles Complex)

I promise that I'll stop looking after you get these last pieces into the PR 😅

cuberoot:
@Westbrook I have fixed these two issues:

* when you click and drag, the handle now stays focused and responds to key events. I made sure that my fix matches your original implementation, which calls `click()` on the label
* All storybook sliders now have labels

I could not reproduce the alignment issue. I made two identical sliders in storybook, except one had two handles. I flipped back-and-forth, but there was no difference. See below.

ScreenFlow

Westbrook:
Exciting!

- label ✅ 
- arrow keys, seeing this work. there's a peculiarity where the `:focus-visible` styles don't trigger once you move to the arrow keys, which would be expected. I can accept that being something we look at in follow up work (if it's not something you can see a path towards addressing easily) as managing the heuristic there can be a bit of a pain. Please create an issue here if you decide not to address that just now.

I looked deeper on this layout question and you can see it explicitly between these two stories:
- https://cuberoot-multi-handle-slider--spectrum-web-components.netlify.app/storybook/index.html?path=/story/slider--two-handles
- https://cuberoot-multi-handle-slider--spectrum-web-components.netlify.app/storybook/index.html?path=/story/slider--three-handles-ordered

The track travels up 1px moving between the first and the second story, seemingly due to the `variant` attribute being moved to `range` between the two. This seems to be due to the way that `margin` is applied in this rule:

```
:host([dir="ltr"][variant="range"]) .track,
:host([dir="rtl"][variant="range"]) .track {
    padding-top: 0px;
    padding-bottom: 0px;
    padding-left: var(--spectrum-slider-track-middle-handleoffset);
    padding-right: var(--spectrum-slider-track-middle-handleoffset);
    margin: var(--spectrum-slider-range-track-reset);
}
```
Which applies `margin-top: 0;` and overrides the "standard" or expected styling in:
```
#fill, 
.track {
    height: var(--spectrum-slider-track-height);
    box-sizing: border-box;
    position: absolute;
    z-index: 1;
    top: calc(var(--spectrum-slider-height)/2);
    margin-top: calc(var(--spectrum-slider-track-height)/-2);
    pointer-events: none;
}
```
Which would compute to `margin-top: -1px;`. It's purely cosmetic, so I'd accept an issue to address this later as well.

Westbrook:
The above does ask the question, "Does `variant="range"` do anything for the element?". And, if the answer is "yes", should we resolve that `variant` value for the consumer when we recognize that there is more than one `sp-slider-handle` element present?

cuberoot:
Logged #1527 for track issue. I don't think I understand what you are saying about the arrow key issue to describe it in an issue. I may leave you to enter that one.

I have to put this down for other work now. If it needs more polish, then hopefully someone can push it over the top.

Westbrook:
@cuberoot thanks for all the hard work on this and the clean handoff. We'll take it from here and have a release for you, soon!

PR #1425, opened 2021-04-30
fix(menu): pass current focus visibility to menu items

### LINKED ISSUE #1136: menu-item keyboard focus shouldn't show
### Expected Behaviour
Hovering over a menu item should just make that item hovered, and there shouldn't be styling on menu open.

### Actual Behaviour
When you open a dropdown or action menu, the first item has hover styling (maybe for keyboard focus? But it doesn't get the blue bar). If you hover over the second item, both show the same hover styling.

Westbrook:
## Description
Afford the current `:focus-visible` context from the `<sp-menu>` element to its `<sp-menu-item>` children when navigating through them.
- only show the blue edge highlighting when the `:focus-visible` heuristic has been triggered.
- persist selected and hovered visual states without alteration

## Related Issue
fixes #1136 

## Motivation and Context
Clearer interface delivery.

## How Has This Been Tested?
New unit tests for both sides of the workflow, including some _special_ Firefox conciderations.

https://westbrook-menu--spectrum-web-components.netlify.app/components/picker

## Types of changes

## Checklist:

### DIFF
--- a/packages/menu/src/Menu.ts
+++ b/packages/menu/src/Menu.ts
@@ -167,7 +167,7 @@ export class Menu extends SpectrumElement {
         }
         // if there are no non-disabled items, skip the work to focus a child
         if (!itemToFocus.disabled) {
-            itemToFocus.focused = true;
+            this.forwardFocusVisibleToitem(itemToFocus);
             this.setAttribute('aria-activedescendant', itemToFocus.id);
         }
         return itemToFocus;
@@ -220,9 +220,25 @@ export class Menu extends SpectrumElement {
         this.updateSelectedItemIndex();
         const focusInItem = this.menuItems[this.focusInItemIndex] as MenuItem;
         if ((this.getRootNode() as Document).activeElement === this) {
-            focusInItem.focused = true;
+            this.forwardFocusVisibleToitem(focusInItem);
         }
     };
+
+    private forwardFocusVisibleToitem(item: MenuItem): void {
+        let shouldFocus = false;
+        try {
+            // Browsers without support for the `:focus-visible`
+            // selector will throw on the following test (Safari, older things).
+            // Some won't throw, but will be focusing item rather than the menu and
+            // will rely on the polyfill to know whether focus is "visible" or not.
+            shouldFocus =
+                this.matches(':focus-visible') ||
+                this.matches('.focus-visible');
+        } catch (error) {
+            shouldFocus = this.matches('.focus-visible');
+        }
+        item.focused = shouldFocus;
+    }
 
     public render(): TemplateResult {
         return html`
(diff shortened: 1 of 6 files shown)


packages/menu/src/Menu.ts
```diff
@@ -220,10 +220,22 @@ export class Menu extends SpectrumElement {
...
     };
 
+    private forwardFocusVisibleToitem(item: MenuItem): void {
+        let shouldFocus = false;
+        try {
+            shouldFocus =
+                this.matches(':focus-visible') ||
+                this.matches('.focus-visible');
+        } catch (error) {
+            shouldFocus = this.matches('.focus-visible');
```
adixon-adobe, line 234: So the first check can throw in some browsers? Might be worth a comment.

packages/picker/src/Picker.ts
```diff
@@ -253,27 +253,34 @@ export class PickerBase extends SizedMixin(Focusable) {
...
+            reparentableChildren = Array.from(
+                deprecatedMenu.children
+            ) as MenuItem[];
         } else {
             reparentableChildren = Array.from(this.children).filter(
                 (element) => {
                     return !element.hasAttribute('slot');
                 }
-            );
+            ) as MenuItem[];
```
adixon-adobe, line 268: Won't this break re-parenting of MenuDividers? There's a separate query selector to get the menu items (which also might be problematic for sub-classing, I see).

Assuming this does break for MenuDivider though, we should add a test!

adixon-adobe: (it does seem to work on your branch, but this code still looks wrong)

adixon-adobe (changes requested):

packages/picker/src/Picker.ts
Westbrook: I think it looks wrong, because it was needed for something else, but I've backed out of it 🙈

I should be able to revert this to `Element[]`.

adixon-adobe: this is why code reviews are good! It definitely felt like there was more to the story.

packages/picker/src/Picker.ts
```diff
@@ -271,10 +271,15 @@ export class PickerBase extends SizedMixin(Focusable) {
             return;
         }
 
-        this.restoreChildren = reparentChildren(
-            reparentableChildren,
-            this.optionsMenu
-        );
+        this.restoreChildren = reparentChildren<
```
Westbrook, line 274: Now with types more inline with what one might think, I hope! @adixon-adobe

adixon-adobe (approved):
Yep -- this looks good now. Thanks!

PR #1441, opened 2021-05-07
fix(slider): ensure min/max/value application order

### LINKED ISSUE #1308: [Slider] Mange unset `max` value
### Expected Behaviour
`<sp-slider value="500">` should work more clearly like `<input type="range" value="500" />`.

### Actual Behaviour
`<sp-slider>` has a max of 20 by default to the value would be restricted to 20, while the max of the `input` is _unset_ while still restricting the value to 100. Both of these are somewhat surprising, but aligning the element more closely to its native predecessors will reduce the documentation required to explain certain interactions. In both of these cases, when the `max` is set later, there is no alteration to the `value` as part of that interaction.

cc: @kaiortmanns

Westbrook:
## Description
Make sure the `value` is set _after_ `min`, `max`, and `step` in the rendered template and that validation of the `value` happens after the `<input>` element is update, instead of before.

## Related Issue
fixes #1308

## Motivation and Context
Easier, less error prone element usage.

## How Has This Been Tested?
- unit tests updated and passing

## Types of changes

## Checklist:
- [ x] I have signed the [Adobe Open Source CLA](http://opensource.adobe.com/cla.html).

### DIFF
--- a/packages/slider/src/Slider.ts
+++ b/packages/slider/src/Slider.ts
@@ -37,26 +37,7 @@ export class Slider extends Focusable {
     public type = '';
 
     @property({ type: Number, reflect: true })
-    public get value(): number {
-        return this._value;
-    }
-
-    public set value(value: number) {
-        const oldValue = this.value;
-        if (this.input) {
-            this.input.value = String(value);
-        }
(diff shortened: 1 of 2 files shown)


kaiortmanns (approved):
LGTM

PR #1445, opened 2021-05-10
[RFC] Virtual Target helper "element"

### LINKED ISSUE #1437: [Overlay] Add Virtual Element support for triggering overlays
Taking https://popper.js.org/docs/v2/virtual-elements/ into account, surface an API so that something like https://webcomponents.dev/edit/ThKC7g0JNv1sswUo63Iu/src/index.ts is relatively easy to achieve.

Initial thought is a `makeVirtualTriggerAt(x, y)` helper that would possibly extend `eventTarget` to include the `getBoundingClientRect` call so that we could still communicate trigger contexts via DOM events even when the trigger is virtual.

Westbrook:
## Description
By default overlays are dispatched from element triggers, however some magic around `getBoundingClientRects()` allows for a "virtual" trigger, e.g. https://popper.js.org/docs/v2/virtual-elements/. The ability to do so is useful for the context in the PopperJS docs where you want to overlay to follow your cursor, but also in the context of wanting to replace the `contextmenu` display in conjunction with a user interaction point along a `canvas` element, which that "point" to be the trigger of the overlay not the entire `canvas` element. Herein, you'll find a proposal for making that possible in SWC.

### Requirements
On the PopperJS side, the "VirtualTrigger" object needs to at least include a synthetic implementation of `getBoundingClientRects()` in order to know where to place the overlay. This precipitates the need to be able to easily update the results of that call, for which this PR adds the `updateBoundingClientRect()` method to the `VirtualTrigger` element. This accepts the `x` and `y` coordinates of the virtual element.

On the SWC side there are a few more responsibilities of the trigger in an overlay experience.
- it is the `EventTarget` from which `sp-query-theme` (which resolves the Spectrum theme information), as well as `sp-closed` and `sp-opened` (which announces the state of the overlay) are dispatched
- they take `focus()` in certain contexts
- their `getRootNode()` method is leveraged to resolve `activeElement` information
- their `closest()` method is used to resolve overlay "stacks" or groups of modals demarcated by `type="modal"` overlays
All of these responsibilities stem the the `HTMLElement` typing that they embody, and in relying on their DOM position for the majority of these things raise the bar on a virtual replacement

### Approach
Herein I create a custom element, which directly satisfies the type requirements of `HTMLElement`, that rather than being used as an actual element in the HTML is constructed `new VirtualTrigger()` to be leveraged in a detached state for this functionality. The constructor is passed a `host` and the initial `x` and `y` positional values for the trigger. The host needs to be a DOM element that is attached to the tree and benefits from being a part of the actual application for the sake of resolving themes, etc. but doesn't strictly need to be associated to the actual interaction point. For ease of use, in our default example of triggering an overlay from a point on a `canvas` element, while the `canvas` can't be the target, it can, and likely should, be the host. In the constructor of the `VirtualElement`, the host (and the implementations for `getBoundingClientRects` and `updateBoundingClientRect`) are wrapped in a `Proxy()` and returned so that calls to the cached virtual trigger will be made to the Proxy rather than to an actual element.

### Benefits
Providing a centralized `VirtualTrigger` API will be useful for code sharing and understandability across larger shared app surfaces while covering over foot gun situations like forgetting to resolve themes, etc. On top of adequately addressing the current needs of the `trigger` element being fairly `HTMLElement`-like, the approach of having `VirtualTrigger` extend `HTMLElement` and proxy calls as needed should be that we don't have to think about it too much when addressing future updates to the overlay system. While we could make a more synthetic "HTMLElement" to support this virtualization of the trigger, we'd likely have to take into account every API on a standard `trigger` that we leveraged and add it to the virtual one. Here we can rely fully on the fact that we have an actual `HTMLElement` for making those calls against. If at any point we need to add more virtual functionality then it simply becomes a matter of managing that forking in the Proxy, as well.

## Questions
- The proxy approach seems like a nice win both for the functional requirements, as well as the TypeScript requirements, but are there other patterns that would better serve us here?
- Like any actual element, when this Virtual Trigger is used up and the closure from which it originates the complete, it should just GC, right? I've not thought a lot about leaks in this manner, but an additional benefit of it extending `HTMLElement` should be that the realities that govern there would govern here, right?
- In the current demo, I'm using a fixed menu, which I think applies more cleanly to the initial use case. However, the demo on the PopperJS side points to the power of using this with a moving `trigger`. In that case, we already have the `updateBoundingClientRect()` method, but as outlined in the follow code on their page a consuming developer would subsequently need to update the instance:

  ```js
      document.addEventListener('mousemove', ({ clientX: x, clientY: y }) => {
          virtualElement.getBoundingClientRect = generateGetBoundingClientRect(x, y);
          instance.update();
      });
  ```
  We surface this functionality via `Overlay.update()`, which is a nice helper, but in this case would we better serve our users by closing other that functionality and calling that for them when they update the position of their virtual trigger?
  ```js
      updateBoundingClientRect(x: number, y: number): void {
          this.x = x;
          this.y = y;
          Overlay.update();      
      }
  ```
- Are there use cases that you'd like to be sure are covered by this functionality in particular that you don't think is addressed here in?

## Related Issue
fixes #1437  

## Motivation and Context
It should be easy to leverage the full capability of our overlay system.

## How Has This Been Tested?
- no tests yet, just demos for conversation

Demo available here, right click to see the overlay:
https://westbrook-virtual-element--spectrum-web-components.netlify.app/storybook/?path=/story/overlay--virtual-element
This will look particularly nice when we get #1425 merged!

## Types of changes

## Checklist:
- [ ] I have updated the documentation accordingly. <=== ADD ME!
- [ ] I have added tests to cover my changes. <=== ADD ME!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ executors:
 parameters:
     current_golden_images_hash:
         type: string
-        default: 181d8463d1de8a10c6c47d4bf51eb4024c5ea1b4
+        default: 38d50443c7f6490dccc8f0715f702165eadb7679
 commands:
     downstream:
         steps:
--- a/packages/overlay/src/VirtualTrigger.ts
+++ b/packages/overlay/src/VirtualTrigger.ts
@@ -1,0 +1,74 @@
+/*
+Copyright 2020 Adobe. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License. You may obtain a copy
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+type VirtualElement = HTMLElement & { updateBoundingClientRect: void };
+
+export class VirtualTrigger extends HTMLElement {
+    x = 0;
+    y = 0;
+    constructor(host: HTMLElement, x: number, y: number) {
+        super();
+        this.x = x;
+        this.y = y;
+
+        const target = {};
+        const getHandler = (
+            _target: Record<string, unknown>,
+            prop: keyof VirtualElement
+        ): unknown => {
+            if (
+                prop !== 'getBoundingClientRect' &&
+                prop !== 'updateBoundingClientRect'
+            ) {
+                const hostProp = host[prop];
+                if (typeof hostProp === 'function') {
+                    return hostProp.bind(host);
+                }
+                return hostProp;
+            }
+            return this[prop].bind(this);
+        };
+
+        const handler2 = {
+            get: getHandler,
+        };
+        const proxy2 = (new Proxy(
+            target,
+            handler2
+        ) as unknown) as VirtualTrigger;
+
+        return proxy2;
+    }
+
+    updateBoundingClientRect(x: number, y: number): void {
+        this.x = x;
+        this.y = y;
+    }
+
+    getBoundingClientRect(): DOMRect {
+        return {
+            width: 0,
+            height: 0,
+            top: this.y,
+            right: this.x,
+            y: this.y,
+            x: this.x,
+            bottom: this.y,
+            left: this.x,
+            toJSON() {
+                return;
+            },
+        };
+    }
+}
+
+customElements.define('sp-virtual-overlay-target', VirtualTrigger);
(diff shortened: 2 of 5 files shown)


spdev3000:
Just my 2cents here: It would really make sense for us to have this VirtualTrigger when using context menu from our drawing stage/canvas. Not sure if we would need a separation between Overlay API trigger and the PopperJS trigger though. So far it looks good to me.

Westbrook:
@adixon-adobe thinking about this some more, I wonder if it makes better sense to conditional make the PopperJS trigger and the SWC Overlay trigger two different things. In this way we could leverage the current trigger in all of our current usage, but in a case where we want to position the Overlay in a way not associated with an element, we could send in the options a `virtualTrigger` that merely managed access to a synthetic `getBoundingClientRects()` with customization for which surfaced via an `updateBoundingClientRect()` method. Thoughts?

I'll mark up a separate branch with how I see that working for comparison.

Westbrook:
Closing in favor of #1477

Issue #1454, opened 2021-05-12
Color slider user drag behavior


bengfarrell:
### Expected Behaviour
Dragging the slider should follow your pointer

### Actual Behaviour
Slider loses drag behavior mid-stream and just stops when testing on mobile, or mobile simulated dev tools

### Reproduce Scenario (including but not limited to)
Just open Chrome's mobile sim and try to drag the color slider (even with a mouse)

Westbrook:
This has been fixed!

PR #1455, opened 2021-05-12
fix: include touch-action rule for draggable content

Westbrook:
## Description
Without specific `touch-action: ...` rules scroll actions will take priority over element interactions and make it difficult to work with custom input elements. This updates the following to allow for this:

EDIT (move to `none` in all): After playing with this on a touch device a little more the `pan-*` values just don't feel right.

- `sp-color-area` and `sp-color-wheel`: these are both two dimensional inputs, so we use `touch-action: none;`. There is a secondary issue this could cause in the case that a consumer makes the Color Area larger than the viewport, but that seems like something we'd rather live with?
- `sp-color-handle`: if you touch the handle, you'd likely be wanting to interact with the input, so `touch-action: none;` this applies to the rest of the Color controls that leverage this internally.
- `sp-color-slider`: while the interactions are directional, the smallest amount of "falling off" will loose the touch event for the control, so use `touch-action: none;` here, too.
- `sp-slider`: The handle previously included this rule, add to the track as well taking into account that while the interactions are directional, the smallest amount of "falling off" will loose the touch event for the control, so use `touch-action: none;` here, too.

In testing, I found that the `sp-slider` had a small bug from the work in #1441 and corrected it and added a test to prevent the situation in the future.

## Related Issue
refs #1454 

## Motivation and Context
Touch environments should have the same ease of use as any other.

## How Has This Been Tested?
- unit tests for the slider fix
- manually for the touch-action changes:
  - https://westbrook-touch-events--spectrum-web-components.netlify.app/components/color-area
  - https://westbrook-touch-events--spectrum-web-components.netlify.app/components/color-slider
  - https://westbrook-touch-events--spectrum-web-components.netlify.app/components/color-wheel
  - https://westbrook-touch-events--spectrum-web-components.netlify.app/components/slider

## Types of changes

## Checklist:

### DIFF
--- a/packages/color-area/src/color-area.css
+++ b/packages/color-area/src/color-area.css
@@ -12,6 +12,10 @@ governing permissions and limitations under the License.
 
 @import './spectrum-color-area.css';
 
+:host {
+    touch-action: none;
+}
+
 :host:before {
     pointer-events: none;
 }
--- a/packages/color-handle/src/color-handle.css
+++ b/packages/color-handle/src/color-handle.css
@@ -11,3 +11,7 @@ governing permissions and limitations under the License.
 */
 
 @import './spectrum-color-handle.css';
+
+:host {
+    touch-action: none;
+}
--- a/packages/color-slider/src/color-slider.css
+++ b/packages/color-slider/src/color-slider.css
@@ -18,6 +18,14 @@ governing permissions and limitations under the License.
         rgb(255, 0, 255) 83%, rgb(255, 0, 0) 100%;
 }
 
+:host(:not([vertical])) {
+    touch-action: pan-y;
+}
+
+:host([vertical]) {
+    touch-action: pan-x;
+}
+
 .gradient {
     overflow: hidden;
 }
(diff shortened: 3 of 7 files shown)


bengfarrell (approved):
This looks great! I hear you on not being sure about using pan-x/y, but I think this is a nice touch. I've been testing on my Pixel 4 and I like the behavior you can afford with the horizontal slider for example. I can just accidentally tap it and my page still scrolls if I'm I'm using the vertical gesture. I suppose if I were being nitpicky, I should get the same behavior from the center of the color wheel, but I think that's a waaaaaaaaay minor request that should be battled another day. I did not test the slider dot property change as I'm not sure what use cases it impacts, though the slider still seems to work fine.

Thanks for the change!

Westbrook:
Ha, I wrote that I fixed a small bug, but not what it was 🤦🏼 

The `.value` change means we're updating the property (which had previously been done imperatively) instead of the attribute. The `value` attribute only applies its value to an `input` the first time it is set, which meant that subsequent interaction with the slider were not registered.

PR #1474, opened 2021-05-20
fix: prevent tabindex=-1 elements from placing focus on their host

### LINKED ISSUE #1038: Arrow key navigation through sp-radio-group sometimes gets stuck
### Expected Behaviour

Should be able to press Left/Right keys to enable different Radio buttons

### Actual Behaviour

In some sequences, pressing Left/Right eventually appears to have no effect, i.e. they stop changing the selected radio button.

#### Steps to Reproduce

Here's one reliable way to repro, though I've hit this with a few different assortments of keystrokes.
1. Go to https://opensource.adobe.com/spectrum-web-components/components/radio
2. With the mouse, scroll to the first Radio group example. Note Option 1 is selected.
3. Click Option 4 to select it.
4. Press left key three times. As expected, Option 1 is now selected.
5. Press right key. Unexpectedly, Option 1 is still selected.
image

#### Platform and Version
Tested on Chrome 87 on Mac.

See docs above

Nothing logged to console

### LINKED ISSUE #1426: radio elements show blue focus ring on safari when selected via mouse
### Expected Behaviour
Blue focus borders only show up with keyboard navigation

### Actual Behaviour
On safari, these are showing up in radio elements when you click on them

#### Steps to Reproduce
Click one of the radio buttons here: https://opensource.adobe.com/spectrum-web-components/components/radio

#### Platform and Version
Tested on Safari 14.0.3 (15610.4.3.1.7, 15610)

Westbrook:
## Description
- prevent the presence of a focus outline on element hosts in Safari where the `focusElement` child isn't a "focusable" element
  - Safari doesn't not by default support focusing `input[type="radio"]`, `input[type="checkbox"]`, `input[type="range"]`, or `a` elements, so while "focus" will sometimes resolve to their host the focus outline has been remove from these sorts of elements
- prevent focusing the host when interacting with elements that contain "focusable" children
- correct the time as which `tabindex` is manages for children of a `sp-radio-group` element so that it performs more predictably

## Related Issue
fixes #1426 
fixes #1038

## Motivation and Context
- correct visual delivery in Safari
- correct keyboard interaction when `input[type=radio]` is "focusable"

## How Has This Been Tested?
- unit tests
- manually

https://westbrook-focusable--spectrum-web-components.netlify.app/components/radio

## Types of changes

## Checklist:

### DIFF
--- a/documentation/src/components/code-example.css
+++ b/documentation/src/components/code-example.css
@@ -44,7 +44,7 @@ governing permissions and limitations under the License.
 #markup {
     display: grid;
     grid-template-columns: 100%;
-    grid-template-rows: 100%;
+    grid-template-rows: minmax(60px, 100%);
     padding: 0.75rem 1.5rem;
     border-radius: 0 0 6px 6px;
     border-top: 1px solid var(--spectrum-global-color-gray-100);
--- a/packages/checkbox/src/checkbox.css
+++ b/packages/checkbox/src/checkbox.css
@@ -17,10 +17,14 @@ governing permissions and limitations under the License.
     vertical-align: top;
 }
 
-:host(:empty) label {
-    display: none;
+:host(:focus) {
+    outline: none;
 }
 
 :host([disabled]) {
     pointer-events: none;
+}
+
+:host(:empty) label {
+    display: none;
 }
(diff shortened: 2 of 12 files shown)


najikahalsema (dismissed):
lgtm!

PR #1475, opened 2021-05-22
Adixon 1189 menu prep

### LINKED ISSUE #715: Menu should manage its own state

adixon-adobe:
Continued work on **selects** attribute in sp-menu & sp-menu-group

Preview link: https://adixon-1189-menu-prep--spectrum-web-components.netlify.app/

## Description

For this round I've tried to keep the changes as scoped to sp-menu as possible, given that the reparenting in PickerBase is going to be tricky to work through.

I'm stuck on initialization for sub-menus though -- `prepItems` isn't getting called once the slotted items in the underlying `sp-menu-group`'s `sp-menu` are present (meaning that after initializing menu groups always think they have 0 menu items). I've got some new unittests that are failing, and it's pretty easy to see in the storybook for "Menu Group Selects Multiple"

The only (known) work left here outside of that is getting the documentation work.

## Related Issue

Initial work for #1189 
fixes #715

## Motivation and Context

Gets us closer to more useful menus

## How Has This Been Tested?

Storybook & unit tests.

Here's an example of what's failing: the "Oakland" menu group should be single-select but it initializes wrong. If you keep using the menu it eventually starts working properly once `prepItems` has been called correctly.
Screen Shot 2021-05-22 at 11 54 57 AM

## Types of changes

- [ x ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ x ] I have added tests to cover my changes.

### DIFF
--- a/packages/menu/src/Menu.ts
+++ b/packages/menu/src/Menu.ts
@@ -17,9 +17,10 @@ import {
     TemplateResult,
     property,
     PropertyValues,
+    query,
 } from '@spectrum-web-components/base';
 
-import { MenuItem, MenuItemQueryRoleEventDetail } from './MenuItem.js';
+import { MenuItem, MenuItemUpdateEvent } from './MenuItem.js';
 import menuStyles from './menu.css.js';
 
 export interface MenuQueryRoleEventDetail {
@@ -56,11 +57,17 @@ export class Menu extends SpectrumElement {
     @property({ attribute: false })
     public selectedItems = [] as MenuItem[];
 
+    @query('slot')
+    private menuSlot!: HTMLSlotElement;
+
     public menuItems = [] as MenuItem[];
+    private menuItemSet = new Set() as Set<MenuItem>;
     public focusedItemIndex = 0;
     public focusInItemIndex = 0;
 
     private selectedItemsMap = new Map() as Map<MenuItem, boolean>;
+
+    private itemsChanged: boolean = false;
 
     /**
      * Hide this getter from web-component-analyzer until
@@ -118,10 +125,28 @@ export class Menu extends SpectrumElement {
         super();
 
         this.addEventListener(
-            'sp-menu-item-query-role',
-            (event: CustomEvent<MenuItemQueryRoleEventDetail>) => {
+            'sp-menu-item-added',
+            (event: CustomEvent<MenuItemUpdateEvent>) => {
+                const item = event.detail.item;
+                if (this.selects === 'inherit') {
+                    // We still track children for focus
+                    if (!event.detail.inherited) {
+                        event.detail.inherited = true;
+                        this.addItem(item);
+                    }
+                    return;
+                }
                 event.stopPropagation();
-                event.detail.role = this.childRole;
+                item.setRole(this.childRole);
+                this.addItem(item);
+            }
+        );
+
+        this.addEventListener(
+            'sp-menu-item-removed',
+            (event: CustomEvent<MenuItemUpdateEvent>) => {
+                event.stopPropagation();
+                this.removeItem(event.detail.item);
             }
         );
 
@@ -365,30 +390,62 @@ export class Menu extends SpectrumElement {
         this.focusInItemIndex = index;
     }
 
-    private async updateItemRoles() {
-        const childMenuItems = [
-            ...this.querySelectorAll(`sp-menu-item`),
-        ] as MenuItem[];
+    // debounce update this.menuItems so they're in DOM order
+    private async updateMenuItems(): Promise<void> {
+        this.itemsChanged = true;
 
-        const promises: Promise<any>[] = [];
+        return new Promise((resolve) => {
+            // debounce so we only actually update once
+            requestAnimationFrame(() => {
+                if (!this.itemsChanged) {
+                    resolve();
+                }
 
-        for (const item of childMenuItems) {
-            item.updateRole();
-            promises.push(item.updateComplete);
-        }
-        await Promise.all(promises);
+                this.menuItems = [];
+
+                const slotElements =
+                    this.menuSlot.assignedElements({ flatten: true }) || [];
+                for (const slotElement of slotElements) {
+                    const childItems: MenuItem[] =
+                        slotElement instanceof MenuItem
+                            ? [slotElement as MenuItem]
+                            : ([
+                                  ...slotElement.querySelectorAll(
+                                      `[role="${this.childRole}"]`
+                                  ),
+                              ] as MenuItem[]);
+                    for (const childItem of childItems) {
+                        if (this.menuItemSet.has(childItem)) {
+                            this.menuItems.push(childItem);
+                        }
+                    }
+                }
+
+                this.updateSelectedItemIndex();
+                this.updateItemFocus();
+                this.itemsChanged = false;
+                resolve();
+            });
+        });
     }
 
-    private async prepItems() {
-        await this.updateItemRoles();
+    private addItem(item: MenuItem) {
+        this.menuItemSet.add(item);
+        this.updateMenuItems();
+    }
 
-        this.menuItems = [
-            ...this.querySelectorAll(`[role="${this.childRole}"]`),
-        ] as MenuItem[];
-        if (!this.menuItems || this.menuItems.length === 0) {
-            return;
+    private removeItem(itemToRemove: MenuItem) {
+        const newMenuItems: MenuItem[] = [];
+        for (const item of this.menuItems) {
+            if (item !== itemToRemove) {
+                newMenuItems.push(item);
+            }
         }
-        this.updateSelectedItemIndex();
+        this.menuItemSet.delete(itemToRemove);
+        this.updateMenuItems();
+    }
+
+    private async updateItemFocus() {
         const focusInItem = this.menuItems[this.focusInItemIndex] as MenuItem;
         if ((this.getRootNode() as Document).activeElement === this) {
             this.forwardFocusVisibleToitem(focusInItem);
@@ -396,6 +453,7 @@ export class Menu extends SpectrumElement {
     }
 
     private forwardFocusVisibleToitem(item: MenuItem): void {
+        console.log('forwarding focus');
         let shouldFocus = false;
         try {
             // Browsers without support for the `:focus-visible`
@@ -429,11 +487,7 @@ export class Menu extends SpectrumElement {
         if (!this.hasAttribute('role')) {
             this.setAttribute('role', 'menu');
         }
-        if (!this.observer) {
-            this.observer = new MutationObserver(this.prepItems.bind(this));
-        }
-        this.observer.observe(this, { childList: true, subtree: true });
-        this.updateComplete.then(() => this.prepItems());
+        this.updateComplete.then(() => this.updateItemFocus());
         const selectedItem = this.querySelector('[selected]');
         if (selectedItem) {
             requestAnimationFrame(() => {
@@ -442,18 +496,12 @@ export class Menu extends SpectrumElement {
         }
     }
 
-    public disconnectedCallback(): void {
-        this.observer.disconnect();
-        super.disconnectedCallback();
-    }
-
     protected updated(changes: PropertyValues): void {
         if (changes.has('selects')) {
-            this.prepItems();
+            // TODO: update roles/selection and announce ownership change if
+            // we're going to/from inherits
         }
     }
-
-    private observer!: MutationObserver;
 }
 
 declare global {
--- a/packages/menu/src/Menu.ts
+++ b/packages/menu/src/Menu.ts
@@ -56,7 +56,6 @@ export class Menu extends SpectrumElement {
     @query('slot')
     private menuSlot!: HTMLSlotElement;
 
-    @property({ attribute: false, type: String, reflect: true })
     public get menuItems(): MenuItem[] {
         if (this.cachedMenuItems !== undefined) {
             return this.cachedMenuItems;
--- a/packages/menu/src/Menu.ts
+++ b/packages/menu/src/Menu.ts
@@ -99,6 +99,8 @@ export class Menu extends SpectrumElement {
         }) as MenuItem;
         /* c8 ignore next 3 */
         if (target) {
+            event.preventDefault();
+            event.stopImmediatePropagation();
             this.selectOrToggleItem(target);
         } else {
             return;
--- a/packages/menu/src/Menu.ts
+++ b/packages/menu/src/Menu.ts
@@ -48,7 +48,7 @@ export class Menu extends SpectrumElement {
 
     // For the multiple select case, we'll join the value strings together
     // for the value property with this separator
-    @property({ type: String })
+    @property({ type: String, attribute: 'value-separator' })
     public valueSeparator = ',';
 
     // TODO: which of these to keep?
@@ -318,8 +318,8 @@ export class Menu extends SpectrumElement {
             // item as the value. Also set the selected array
             // in the order of the menu items.
             let valueSet: boolean = false;
-            let selected: string[] = [];
-            let selectedItems: MenuItem[] = [];
+            const selected: string[] = [];
+            const selectedItems: MenuItem[] = [];
 
             for (const childItem of this.childItems) {
                 if (!childItem.managed) continue;
@@ -346,12 +346,14 @@ export class Menu extends SpectrumElement {
             })
         );
         if (!applyDefault) {
+            // Cancel the event & don't apply the selection
             this.selected = oldSelected;
             this.selectedItems = oldSelectedItems;
             this.selectedItemsMap = oldSelectedItemsMap;
             this.value = oldValue;
             return;
         }
+        // Apply the selection changes to the menu items
         if (resolvedSelects === 'single') {
             for (const oldItem of oldSelectedItemsMap.keys()) {
                 if (oldItem !== item) {
--- a/packages/menu/src/Menu.ts
+++ b/packages/menu/src/Menu.ts
@@ -23,10 +23,6 @@ import {
 import { MenuItem, MenuItemUpdateEvent } from './MenuItem.js';
 import menuStyles from './menu.css.js';
 
-export interface MenuQueryRoleEventDetail {
-    role: string;
-}
-
 /**
  * Spectrum Menu Component
  * @element sp-menu
@@ -104,7 +100,7 @@ export class Menu extends SpectrumElement {
                 parent = shadowRoot?.host as HTMLElement;
             }
             while (parent != null) {
-                if (parent.tagName === 'SP-MENU') {
+                if (parent.localName === 'sp-menu') {
                     const selects = parent.getAttribute('selects');
                     const role = parent.getAttribute('role');
                     if (selects === 'single' || selects === 'multiple') {
@@ -181,7 +177,6 @@ export class Menu extends SpectrumElement {
             }
             return el.getAttribute('role') === this.childRole;
         }) as MenuItem;
-        /* c8 ignore next 3 */
         if (target && this.selects !== 'inherit') {
             event.preventDefault();
             event.stopImmediatePropagation();
@@ -435,12 +430,6 @@ export class Menu extends SpectrumElement {
     }
 
     private removeItem(itemToRemove: MenuItem) {
-        const newMenuItems: MenuItem[] = [];
-        for (const item of this.menuItems) {
-            if (item !== itemToRemove) {
-                newMenuItems.push(item);
-            }
-        }
         this.menuItemSet.delete(itemToRemove);
         this.updateMenuItems();
     }
@@ -453,7 +442,6 @@ export class Menu extends SpectrumElement {
     }
 
     private forwardFocusVisibleToitem(item: MenuItem): void {
-        console.log('forwarding focus');
         let shouldFocus = false;
         try {
             // Browsers without support for the `:focus-visible`
@@ -501,11 +489,5 @@ export class Menu extends SpectrumElement {
             // TODO: update roles/selection and announce ownership change if
             // we're going to/from inherits
         }
-    }
-}
-
-declare global {
-    interface GlobalEventHandlersEventMap {
-        'sp-menu-query-role': CustomEvent<MenuQueryRoleEventDetail>;
     }
 }
--- a/packages/menu/src/Menu.ts
+++ b/packages/menu/src/Menu.ts
@@ -29,6 +29,7 @@ export interface MenuQueryRoleEventDetail {
 /**
  * Spectrum Menu Component
  * @element sp-menu
+ * @fires change - Announces that the `value` of the element has changed
  *
  */
 export class Menu extends SpectrumElement {
@@ -36,12 +37,21 @@ export class Menu extends SpectrumElement {
         return [menuStyles];
     }
 
-    @property({ type: Boolean, reflect: true })
-    public selectable = false;
+    @property({ type: String, reflect: true })
+    public selects: undefined | 'single' | 'multiple';
+
+    @property({ type: String })
+    public value = '';
+
+    // TODO: allow setting this in the API to change the values
+    @property({ attribute: false })
+    public selected = [] as string[];
 
     public menuItems = [] as MenuItem[];
     public focusedItemIndex = 0;
     public focusInItemIndex = 0;
+
+    private selectedItemsMap = new Map() as Map<MenuItem, boolean>;
 
     /**
      * Hide this getter from web-component-analyzer until
@@ -88,7 +98,9 @@ export class Menu extends SpectrumElement {
             return el.getAttribute('role') === this.childRole;
         }) as MenuItem;
         /* c8 ignore next 3 */
-        if (!target) {
+        if (target) {
+            this.selectOrToggleItem(target);
+        } else {
             return;
         }
         this.prepareToCleanUp();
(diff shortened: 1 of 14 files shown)


packages/picker/src/Picker.ts
```diff
@@ -365,7 +365,11 @@ export class PickerBase extends SizedMixin(Focusable) {
                 @click=${this.onClick}
                 @sp-overlay-closed=${this.onOverlayClosed}
             >
-                <sp-menu id="menu" role="${this.listRole}"></sp-menu>
+                <sp-menu
+                    id="menu"
+                    selects="inherit"
```
adixon-adobe, line 370: This is a little weird, since it has nothing to inherit, but it allows the menu to be unmanaged right now (so that `PickerBase` can continue managing until we actually finish reconciling this change).

packages/menu/src/MenuItem.ts
```diff
@@ -92,26 +92,41 @@ export class MenuItem extends ActionButton {
...
+            ) {
+                this.setAttribute(
+                    'aria-checked',
+                    this.selected ? 'true' : 'false'
+                );
+            }
         }
     }
 
+    public updateRole(): void {
```
adixon-adobe, line 114: This is part of what I haven't gotten quite working yet. The idea is that the parent menu asks the item to update when it sees it's children change.

packages/menu/src/Menu.ts
```diff
@@ -201,16 +345,43 @@ export class Menu extends SpectrumElement {
...
+        this.selectedItems = selectedItems;
         this.focusedItemIndex = index;
         this.focusInItemIndex = index;
     }
 
-    private prepItems = (): void => {
+    private async updateItemRoles() {
+        const childMenuItems = [
+            ...this.querySelectorAll(`sp-menu-item`),
+        ] as MenuItem[];
```
adixon-adobe, line 371: What I'm seeing after initialization in the context of MenuGroup is that this is still an empty array in the last call. I'm guessing there's something I'm not quite grokking in the lifecycle, or that `MutationObserver` doesn't actually work for re-slotted children.

packages/menu/src/Menu.ts
```diff
@@ -201,16 +345,43 @@ export class Menu extends SpectrumElement {
...
+
+        const promises: Promise<any>[] = [];
+
+        for (const item of childMenuItems) {
+            item.updateRole();
+            promises.push(item.updateComplete);
+        }
+        await Promise.all(promises);
```
adixon-adobe, line 379: I'm guessing this should be removed once we properly resolve this.

adixon-adobe:
I think I have a bit more work to do with `inherits` & selection styles here as well -- I bet we're not applying `no selection` styles in some menu group cases we should.

packages/menu/src/Menu.ts
```diff
@@ -51,11 +70,61 @@ export class Menu extends SpectrumElement {
...
+        }
+    }
+
+    private get resolvedSelects(): undefined | String {
+        return this.resolvedSelectsAndRole[0];
+    }
+
+    private get resolvedSelectsAndRole(): [String | undefined, String | null] {
+        if (this.selects === 'inherit') {
+            let parent = this.parentElement;
```
Westbrook, line 94: We'd likely want a way to cache this in the future?

packages/menu/src/Menu.ts
```diff
@@ -88,7 +157,11 @@ export class Menu extends SpectrumElement {
             return el.getAttribute('role') === this.childRole;
         }) as MenuItem;
         /* c8 ignore next 3 */
```
Westbrook, line 159: This probably needs updating, too?

packages/menu/src/Menu.ts
```diff
@@ -125,6 +198,77 @@ export class Menu extends SpectrumElement {
         this.removeEventListener('keydown', this.handleKeydown);
     }
 
+    public async selectOrToggleItem(item: MenuItem): Promise<void> {
+        const resolvedSelects = this.resolvedSelects;
+        if (resolvedSelects == null) {
```
Westbrook, line 203: This looks like it tests as `undefined` not as `null`?

```
private get resolvedSelects(): undefined | String {
```

packages/menu/src/Menu.ts
```diff
@@ -125,6 +198,77 @@ export class Menu extends SpectrumElement {
...
+            } else {
+                this.selectedItemsMap.set(item, true);
+            }
+
+            // Match HTML select and set the first selected
+            // item as the value. Also set the selected array
+            // in the order of the menu items.
+            let valueSet: boolean = false;
+            let selected: string[] = [];
+            let selectedItems: MenuItem[] = [];
```
Westbrook, line 230: I wonder why you linter doesn't complain at these two values not being `const`?

Westbrook: I'll need to see what about mine does that.

packages/menu/src/Menu.ts
```diff
@@ -125,6 +198,77 @@ export class Menu extends SpectrumElement {
...
+        if (resolvedSelects === 'single') {
+            for (const oldItem of oldSelectedItemsMap.keys()) {
+                if (oldItem !== item) {
+                    oldItem.selected = false;
+                }
+            }
+            item.selected = true;
+        } else if (resolvedSelects === 'multiple') {
+            item.selected = !item.selected;
+        }
```
Westbrook, line 269: I don't fully grok what is happening here. Can you comment a little for next me, who is unequivocally going to be less smart than current me.

packages/menu/src/Menu.ts
Westbrook: `this.querySelectorAll` won't see re-slotted children. We'll need to move to something like `slot.assignedElements()` or the like.

packages/menu/src/Menu.ts
Westbrook: Yes, especially because `role` isn't a watched property, so without a `this.requestUpdate()` or something there won't technically be work happening in the child items here.

packages/menu/src/Menu.ts
```diff
@@ -201,16 +345,43 @@ export class Menu extends SpectrumElement {
...
+
+        for (const item of childMenuItems) {
+            item.updateRole();
+            promises.push(item.updateComplete);
+        }
+        await Promise.all(promises);
+    }
+
+    private async prepItems() {
+        await this.updateItemRoles();
```
Westbrook, line 383: Do we update roles every time on purpose?

If we update roles every time, can we get the value of `this.menuItems` either set inside or returned from that method to remove the need to the below `querySelectorAll`?

packages/menu/src/Menu.ts
```diff
@@ -256,18 +427,10 @@ export class Menu extends SpectrumElement {
...
-            const queryRoleEvent = new CustomEvent('sp-menu-query-role', {
-                bubbles: true,
-                composed: true,
-                detail: {
-                    role: '',
-                },
-            });
-            this.dispatchEvent(queryRoleEvent);
-            this.setAttribute('role', queryRoleEvent.detail.role || 'menu');
+            this.setAttribute('role', 'menu');
```
Westbrook, line 430: Should this sometimes be `group`? How?

packages/menu/src/MenuGroup.ts
```diff
@@ -41,13 +46,38 @@ export class MenuGroup extends SpectrumElement {
         MenuGroup.instances += 1;
     }
 
+    private get menuRole() {
+        if (this.selects) {
+            return 'menu';
```
Westbrook, line 51: I'm not sure this can ever see children through the slot in a screen reader, so the role here may actually go somewhere else. I'll have to visit the preview site to confirm.

packages/menu/src/MenuGroup.ts
```diff
@@ -41,13 +46,38 @@ export class MenuGroup extends SpectrumElement {
...
     public render(): TemplateResult {
         const labelledby = `menu-heading-category-${this.instanceCount}`;
         return html`
             <span class="header" id=${labelledby} aria-hidden="true">
                 <slot name="header"></slot>
             </span>
-            <sp-menu role="presentation">
+            <sp-menu
+                selects=${ifDefined(
+                    this.menuSelects ? this.menuSelects : undefined
```
Westbrook, line 77: Should we just have `menuSelects` return `undefined` to simplify the templating here?

packages/menu/src/MenuItem.ts
```diff
@@ -92,26 +92,41 @@ export class MenuItem extends ActionButton {
 
     protected updated(changes: PropertyValues): void {
         super.updated(changes);
-        if (this.getAttribute('role') === 'option' && changes.has('selected')) {
-            this.setAttribute(
-                'aria-selected',
-                this.selected ? 'true' : 'false'
-            );
+        const role = this.getAttribute('role');
```
Westbrook, line 95: Can you move this into the `if` scope for easy removal in the future?

packages/menu/src/MenuItem.ts
Westbrook: If the parent has the data https://github.com/adobe/spectrum-web-components/pull/1475/files#diff-b9cace30081552e1a2c474c35ae00ff7a5c7abe3d7c6b4722179e06d56d4ceb0R124 and tells the child to update the data (this method), maybe we shouldn't use an event anymore?

packages/picker/src/Picker.ts
Westbrook: So `selects="inhereit"` is less controlled than `selects='null|undefined'`?

Westbrook (commented):
This is a code only pass. I'll take a closer look at the actual experience, particularly as it relates to your questions either later today or tomorrow morning.

packages/menu/src/Menu.ts
adixon-adobe: Perfect -- thanks!

packages/menu/src/Menu.ts
```diff
@@ -51,11 +70,61 @@ export class Menu extends SpectrumElement {
...
+
+    private get resolvedSelectsAndRole(): [String | undefined, String | null] {
+        if (this.selects === 'inherit') {
+            let parent = this.parentElement;
+            if (parent == null) {
+                const shadowRoot = this.getRootNode() as ShadowRoot;
+                parent = shadowRoot?.host as HTMLElement;
+            }
+            while (parent != null) {
+                if (parent.tagName === 'SP-MENU') {
```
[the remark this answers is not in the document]
adixon-adobe: I thought `tagName` was always capitalized in HTML: https://developer.mozilla.org/en-US/docs/Web/API/Element/tagName

I might be able to re-work this so it's not needed if I get rid of the role event though.

packages/menu/src/Menu.ts
adixon-adobe: yes -- I'll give this some thought. Might not be that hard to add.

packages/menu/src/Menu.ts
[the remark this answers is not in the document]
Westbrook: Yes, I'm just saying that you can use `localName` like `tagName` but lowercase...because I had no idea and I much prefer lowercase, but not because I expected a change on this.

Binding to a tag names does generally cause some trouble around extensions, but I'm not sure we need to pre-optimize for that.

adixon-adobe:
@Westbrook pushed up my WIP since I *think* I've finally got a solid approach to managing item ownership based on our discussions.

packages/menu/src/MenuItem.ts
```diff
@@ -82,42 +80,73 @@ export class MenuItem extends ActionButton {
...
         this.setAttribute('tabindex', '-1');
         super.firstUpdated(changes);
         if (!this.hasAttribute('id')) {
             this.id = `sp-menu-item-${MenuItem.instanceCount++}`;
         }
+
+        // Slot updates happens after the connected callback,
+        // so we need to wait a frame before announcing ourselves
+        // or the right menu might not pick this up
+        await new Promise((ready) => requestAnimationFrame(ready));
```
adixon-adobe, line 93: This is the big thing that was tripping me up before. Curious if this feels like the right approach.

packages/menu/src/Menu.ts
```diff
@@ -201,30 +370,90 @@ export class Menu extends SpectrumElement {
...
+                this.itemsChanged = false;
+                resolve();
+            });
+        });
+    }
+
+    private addItem(item: MenuItem) {
+        this.menuItemSet.add(item);
+        this.updateMenuItems();
+    }
```
adixon-adobe, line 435: This and the `updateMenuItems` method are the bulk of the new management logic. While there's a lot more to clean up in this PR, it'd be helpful to get some feedback on this approach.

packages/menu/src/Menu.ts
```diff
@@ -201,30 +370,90 @@ export class Menu extends SpectrumElement {
...
+
+    private removeItem(itemToRemove: MenuItem) {
+        const newMenuItems: MenuItem[] = [];
+        for (const item of this.menuItems) {
+            if (item !== itemToRemove) {
+                newMenuItems.push(item);
```
Westbrook, line 441: This doesn't seem to get used? I much prefer this to the full walk that happes in `updateMenuItems()` below, if it's possible to leverage this.

packages/menu/src/Menu.ts
Westbrook: IIUC, the `this.menuItems` array that is built in `this.updateMenuItems()` only needs to be maintained in the original `sp-menu` descendent, right? Like in this case:
```
<sp-menu> // I'm a menu.
  <sp-menu-group> // I own a menu
    <sp-menu-item></sp-menu-item>
  </sp-menu-group>
  <sp-menu-group> // I also own a meni
    <sp-menu-item></sp-menu-item>
  </sp-menu-group>
  <sp-menu> // I'm also a menu
    <sp-menu-item></sp-menu-item>
  </sp-menu>
</sp-menu>
```
While _technically_ there would be 4 `sp-menu` elements in the various shadow trees, only the one at the root would need to coalesce this data in order to manage focus/arrow keys/etc. Do we need additional logic to skip that work, or am I missing something?

packages/menu/src/Menu.ts
```diff
@@ -201,30 +370,90 @@ export class Menu extends SpectrumElement {
...
+                        slotElement instanceof MenuItem
+                            ? [slotElement as MenuItem]
+                            : ([
+                                  ...slotElement.querySelectorAll(
+                                      `[role="${this.childRole}"]`
+                                  ),
+                              ] as MenuItem[]);
+                    for (const childItem of childItems) {
+                        if (this.menuItemSet.has(childItem)) {
+                            this.menuItems.push(childItem);
```
Westbrook, line 419: Related to the below comment around who owns `this.menuItems` to manage keyboard and focus work, it seems like `this.menuItems` is a larger collection that `this.menuItemSet` (which is about selection, right?) in way that questions this logic...

packages/menu/src/Menu.ts
```diff
@@ -201,30 +370,90 @@ export class Menu extends SpectrumElement {
...
+                        }
+                    }
+                }
+
+                this.updateSelectedItemIndex();
```
Westbrook, line 424: Possibly nit as O(1n) and O(2n) are kind of the same...

This walks the array we just built again, is there room to merge the work?

packages/menu/src/Menu.ts
```diff
@@ -201,30 +370,90 @@ export class Menu extends SpectrumElement {
...
+            }
         }
-        this.updateSelectedItemIndex();
+        this.menuItemSet.delete(itemToRemove);
+        this.updateMenuItems();
+    }
+
+    private async updateItemFocus() {
         const focusInItem = this.menuItems[this.focusInItemIndex] as MenuItem;
         if ((this.getRootNode() as Document).activeElement === this) {
```
Westbrook, line 450: The work in `this.updateMenuItems()` _seems_ only necessary when we're about to do this work. Should we think about a second order of laziness beyond the batching/debouncing that you're doing now?

packages/menu/src/MenuItem.ts
Westbrook: This says that _every_ item should wait, what does the item really care of this? Should the menu that listens do the waiting?

You are already doing a version of this with the debounce in `Menu.updateMenuItems()`. If it were extended in include a `slotchange` resolution, could we get await without needing special handling here?

packages/menu/src/Menu.ts
adixon-adobe: yeah -- I'll remove this.

packages/menu/src/Menu.ts
adixon-adobe: there might be, but there are times we only need to do updateSelectedItemIndex. I'll take a look once I get more of the changes here cleaned up.

packages/menu/src/Menu.ts
adixon-adobe: If a sub-menu manages it's own selection things get trickier, since we still want arrow keys to behave sensibly. I haven't addressed that yet (realized this case right after I pushed up yesterday). So tweaking your example slightly:

```
<sp-menu> // I'm a menu.
  <sp-menu-group> // I own a menu
    <sp-menu-item></sp-menu-item>
  </sp-menu-group>
  <sp-menu-group selects="single"> // I'm a menu that manages my own selection
    <sp-menu-item></sp-menu-item>
  </sp-menu-group>
  <sp-menu> // I'm also a menu
    <sp-menu-item></sp-menu-item>
  </sp-menu>
</sp-menu>
```

packages/menu/src/Menu.ts
adixon-adobe: I think we still need `this.menuItems` to be up to date (it's public, and you also should be able to programmatically update selected state).

packages/menu/src/Menu.ts
adixon-adobe: Looks like I didn't finish the cleanup here. I just set the role directly in `PickerBase` now. That's the only place we were listening to this event internally.

packages/menu/src/Menu.ts
adixon-adobe: One question I have here is whether I need to manage 2 lists. One for focus (every item the menu is a parent of), and one for just the items the menu directly manages.

packages/picker/src/Picker.ts
adixon-adobe: That's true right now, but I'm thinking I should change this so that the `inherits` behavior is the default and `none` is explicit (which would disallow selection). This would also get rid of the difference between `sp-menu` and `sp-menu-group`. Originally I was thinking you'd always use a `menu-group` as the "sub-menu" pattern, but `menu-group` is really just adding a header. And now I'm wondering is `sp-menu-group` should even be it's own component...

packages/menu/src/MenuItem.ts
adixon-adobe: What I like about the event here is that it simplifies the ownership issues. The menu item sends out a "hey I'm here" event, and the right parent picks it up (but only if we do this `await` because of the timing in `sp-menu-group`).

We could also manage this in the parent menu, but we'd need to either need to have each child menu item "trigger" a similar event to determine ownership, or walk the DOM/sub-menus to determine it.

If it weren't for the `await` here I think it'd be a no brainer.

Westbrook: I always forget about the menu in the menu group…🤦‍♂️

packages/menu/src/Menu.ts
Westbrook: This is actually what I was thinking. There are _ways_ to not do this, but I'm pretty certain that while the selection context is handled locally, the focus management has to happen globally. There is likely some tricky accessibility concerns with how the screen reader relates to the data here...

packages/menu/src/Menu.ts
Westbrook: Makes sense.

packages/menu/src/Menu.ts
Westbrook: You could also have a more complicated single list?
```
type ChildItem = {
    element: MenuItem, // reference to DOM element
    managed: boolean, // whether this item is part of the selection pool
    active: boolean, // whether this is the "focused" item or not
}

type ChildItems = ChildItem[];
```

packages/picker/src/Picker.ts
Westbrook: I like `selects="none"`!

I _could_ be convinced that `<sp-menu><span slot="heading">Title of Group</span></sp-menu>` was the same as `<sp-menu-group><span slot="heading">Title of Group</span></sp-menu-group>` for sure. Or maybe it could just be an extension, instead of a second completely independent component...

packages/menu/src/MenuItem.ts
Westbrook: We need to make sure to add `sp-menu-group` (assuming we keep it) to the comment here or this is the sort of things that would get refactored without fully knowing what's going on here!

packages/picker/src/Picker.ts
adixon-adobe: On the one hand, I don't really want to add to this PR, but on the other hand if I could get that `sp-menu` out of the Shadow DOM, it would simplify the problem space quite a bit.

I'll open a separate issue for that so we can discuss.

adixon-adobe: #1505

packages/menu/src/MenuItem.ts
adixon-adobe: This is working (at least for now) with the event approach so marking this resolved.

packages/menu/src/Menu.ts
adixon-adobe: that's weird. I'd expect the linter to have at least caught this at commit time.

packages/menu/src/Menu.ts
adixon-adobe: thought I'd responded yesterday. With the approach here, `this.menuItemSet` is used to manage *what* items are owned, while `this.menuItems` has thee same items in the right DOM order.

I can't say I like it, and one thing I'm thinking is that moving more of the logic here into a `menuItems` getter (along with an invalidation flag), might help clean this up, and address some of the other feedback you have here.

packages/menu/src/Menu.ts
adixon-adobe: Okay, this is helpful!

Would only the top-level parent menu need to manage the focus list (at least until we support nested menus)?

packages/picker/test/picker.test.ts
```diff
@@ -679,7 +693,7 @@ describe('Picker', () => {
         const el = await pickerFixture();
         el.addEventListener('change', (event: Event) => {
             const { value } = event.target as Picker;
-            console.log('change', value);
+            //console.log('change', value);
```
adixon-adobe, line 696: these seemed noisy, but probably worth leaving in the code since it's useful for debugging. Happy to re-enable though!

packages/menu/test/menu-group.test.ts
```diff
@@ -12,14 +12,20 @@ governing permissions and limitations under the License.
...
+    expect,
+    waitUntil,
+} from '@open-wc/testing';
 
 describe('Menu group', () => {
     it('renders', async () => {
         const el = await fixture<Menu>(
             html`
-                <sp-menu>
+                <sp-menu selects="none">
```
adixon-adobe, line 28: Right now this outer menu won't manage the child `sp-menu-group` elements without selects being one of `none`, `single`, or `multiple`, but I'm thinking it might be better for the behavior to default to `selects="none"` for the outer parent menu (i.e. if there's nothing to inherit from the parent).

packages/menu/src/MenuItem.ts
adixon-adobe: At least with the latest code `role` is referenced in the if and the else, so I'm gonna mark this resolved.

packages/menu/src/Menu.ts
adixon-adobe: this should be cached now.

packages/menu/src/Menu.ts
adixon-adobe: Let me know what you think of the way this is re-worked now.

packages/menu/test/menu-group.test.ts
adixon-adobe: I also realize that we need a way to have an "unmanaged" state for the `PickerBase` case, since at least for this round `PickerBase` is still managing it's own selection.

Westbrook: Sub-note, styling for the `Menu` in a `Picker` will currently require a `selects` attribute (value not withstanding) in order to display correctly when going between having a selected child and not. That also means that if `<sp-menu selects="none">` actually doesn't take a selected child that it will be visual "incorrect".

This likely doesn't help. Can you clarify what information would be good for me to share my thoughts on here?

packages/menu/src/Menu.ts
Westbrook: I believe so.

For all intents and purposes, each nested menu will act as its own root: https://webcomponents.dev/edit/Aze4atcpZc7EsxqnCBve/src/index.ts anyways.

packages/menu/test/menu-group.test.ts
adixon-adobe: Oh yeah -- I even looked at that styling earlier today. I think I also need to fix styling for nested menus so I'll leave this open until that's resolved.

It'd be good to get your thoughts on having the default behavior be that selection is unmanaged (if no parent is managing selection). I think this also means that by default the styling should allow for selection. As I think about it more it seems reasonable, but we could certainly go a different direction.

Westbrook: I don't understand this. Can you share a couple of code samples and the expected output?

adixon-adobe: It's probably easiest to think about the difference between `selects="none"` and no selects.

For `selects="none"` we don't allow selection (and in the menu I'm capturing the click event, though that might be unnecessary now). This means that clients should add click events to the individual menu items here, at least the way it's coded now. When adding/updating items we also might clear the selected state on any items marked as selected.

When `selects` isn't present, selection is totally unmanaged. Click events are passed up so either a parent menu or PickerBase can handle it in the way they need. In general this probably isn't the state you want if there's not a parent managing things.

The one piece of this I haven't thought through is whether the default styling with no `selects` should be that the items have the selectable styling or not. Keeping it selectable is handy for the PickerBase case.

packages/menu/src/MenuGroup.ts
adixon-adobe: let me know if there's anything to do here.

packages/menu/src/Menu.ts
```diff
@@ -29,20 +29,39 @@ export interface MenuQueryRoleEventDetail {
...
-    public selectable = false;
+    @property({ type: String, reflect: true })
+    public selects: undefined | 'inherit' | 'single' | 'multiple';
+
+    @property({ type: String })
+    public value = '';
+
+    // For the multiple select case, we'll join the value strings together
+    // for the value property with this separator
+    @property({ type: String })
```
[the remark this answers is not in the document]
adixon-adobe: Switched to non-camelCase. I should also add some tests around this (and a bit more for selection in general).

I don't see a reason for it not to be an attribute as long as it's a string.

adixon-adobe:
@Westbrook I'm seeing one unit test failure on Firefox I might need some help with. picker.test "opens without visible focus on a menu item on click" -- I don't see the same in storybook, but the test says it is getting visible focus (also possible I just need to rebase).

packages/menu/src/MenuGroup.ts
Westbrook: Thanks for reminding me, this looks good right now!

image

Westbrook:
Closing in favor of `adixon-1189-menu-selects`

PR #1477, opened 2021-05-24
Virtual Trigger as additional option to split what's send to Popper and SWC

### LINKED ISSUE #1437: [Overlay] Add Virtual Element support for triggering overlays

Westbrook:
## Description
Add a `virtualTrigger` option to the overlay process that will supersede the element trigger, when available, as part of interaction directly with PopperJS.

`new VirtualTrigger(x, y)` will now provide a `VirtualElement` that can be passed an option to the `openOverlay` methods for use as the "trigger" in calls to PopperJS. The element makes public both a `getBoundingClientRect()` method for position the overlay and a `updateBoundingClientRect(x, y)` method for moving the "trigger" of the overlay. `updateBoundingClientRect()` also calls `Overlay.update()` internally to reduce the amount of manual interaction required to update overlays opened with the virtual trigger context.

## Related Issue
fixes #1437 

## Motivation and Context
Attempt to simplify from #1445 

## How Has This Been Tested?
Manually in Storybook. If this approach makes sense, some unit tests will be added.

## Types of changes

## Checklist:

### DIFF
--- a/packages/overlay/stories/overlay.stories.ts
+++ b/packages/overlay/stories/overlay.stories.ts
@@ -9,7 +9,13 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 import { html, TemplateResult, ifDefined } from '@spectrum-web-components/base';
-import { OverlayContentTypes, OverlayTrigger, Placement } from '../';
+import {
+    openOverlay,
+    OverlayContentTypes,
+    OverlayTrigger,
+    Placement,
+    VirtualTrigger,
+} from '../';
 import '@spectrum-web-components/action-button/sp-action-button.js';
 import '@spectrum-web-components/action-group/sp-action-group.js';
 import '@spectrum-web-components/button/sp-button.js';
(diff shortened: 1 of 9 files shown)


najikahalsema (commented):
Looks good to me so far!

packages/overlay/stories/overlay.stories.ts
```diff
@@ -637,3 +644,51 @@ export const superComplexModal = (): TemplateResult => {
...
+
+export const virtualElement = (args: Properties): TemplateResult => {
+    const pointerenter = async (event: PointerEvent): Promise<void> => {
+        event.preventDefault();
+        const trigger = event.target as HTMLElement;
+        const virtualTrigger = new VirtualTrigger(event.clientX, event.clientY);
+        openOverlay(
+            trigger,
+            'modal',
+            trigger.nextElementSibling as HTMLElement,
```
adixon-adobe, line 656: If I'm understanding this right, clients still pass in an HTMLElement, but now that's just for event bubbling (and focus?), but the virtualTrigger manages positioning. Seems good to me!

adixon-adobe (dismissed):
This seems like a much better approach. Nice!

stanvladut (approved):

PR #1498, opened 2021-06-01
fix: use ObserveSlotText mixin to prevent white space from overriding label attribute

### LINKED ISSUE #1113: Correct slider label usage isn't documented
### Expected Behaviour
Documentation (README) and on site documents adding a label to your `sp-slider` like so:
`<sp-slider label="Slider Label"...`

### Actual Behaviour
Labels don't show up when using the `label` attribute

#### Platform and Version
Spectrum 0.16.0 (tested on FF and Chrome on Mac, but I think its just a doc issue)

Looks like you've switched over to slots for the label, and using an attribute is deprecated. No biggie, but it got me stuck for a bit thinking it was me or a bug.

Westbrook:
## Description
Using an attribute fallback value for a default `slot` element can produce unexpected results when there is a text node with whitespace in side of the host element. The `ObserveSlotText` mixin allows us to apply the attribute in a non-replacing location when there is no content in the slotted text node.

## Related Issue
fixes #1113

## Motivation and Context
Labels should be easy.

## How Has This Been Tested?
Passes VRTs and unit tests.

## Types of changes

## Checklist:

### DIFF
--- a/packages/meter/package.json
+++ b/packages/meter/package.json
@@ -45,6 +45,7 @@
     "dependencies": {
         "@spectrum-web-components/base": "^0.4.3",
         "@spectrum-web-components/field-label": "^0.5.0",
+        "@spectrum-web-components/shared": "^0.12.4",
         "tslib": "^2.0.0"
     },
     "devDependencies": {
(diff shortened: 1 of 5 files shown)


najikahalsema (approved):
Nice!

Issue #1424, opened 2021-04-29
picker/action-menu/split-button menus should be modal by default


adixon-adobe:
### Expected Behaviour
Menus should open in a modal overlay, which prevents scrolling.

### Actual Behaviour
The menus scroll/re-position/re-size as you scroll. While it's cool that it works as well as it does, it leads to several odd UI situations:
* A menu's contents can display when it's not in view (e.g. https://user-images.githubusercontent.com/48841507/116596353-dcc9ef80-a8e9-11eb-999f-77b2dfbbbba9.mov)
* Multiple menus can be open simultaneously
* Page content can resize for long menus: (see: #697)

#### Steps to Reproduce
On the docs page for any of these controls you can open the menu and scroll the page.

adixon-adobe:
I could also see closing the menu on window resize. Regardless I think we'll definitely want to do a little mobile testing to make sure this really works there.

Westbrook:
For reference of what this switch feels like to the use: https://westbrook-picker-modal--spectrum-web-components.netlify.app/components/picker

It's both kind of nice and kind of jarring...looking forward to your thoughts!

Westbrook:
Using `modal` here also fully traps the screen reader in the overlaid (and thus out of flow) content, preventing moving the reader head past the end of the page, which seems like the right idea: 
![image](https://user-images.githubusercontent.com/1156657/117980466-fdd70b00-b301-11eb-89f6-cb0a62aaa405.png)

cc: @jnurthen

Westbrook:
@jnurthen We're talking with Spectrum design and React about updating the experience of picker style overlays to be modal by nature and wanted to get a read from you and the team as to whether there were any underlying issues there. This would put us roughly in line with how Chrome and Safari handle the `<select>` element in macOS, but we're looking to clarify this in the design spec so that it's clear how to implement these sort of patterns in the future. Thoughts?

jnurthen:
While not directly related to this question I'd advise that pickers should start following the read-only combobox pattern rather than the soon to be deprecated from ARIA authoring practices collapsible dropdown pattern. 
See https://github.com/w3c/aria-practices/pull/1852 for the PR which deprecates this pattern.

Westbrook:
[...]

PR #1504, opened 2021-06-01
fix: ensure focus is managed when tabbing out of a menu

### LINKED ISSUE #1488: Opened Action Menu a11y tab behavior
### Expected Behaviour

User can <kbd>tab</kbd> forward from an opened action menu to the next tabIndex.

### Actual Behaviour

When tabbing forward from opened & focused action menu, focus is thrown outside of document to browser controls instead of next element in document.

Able to <kbd>shift</kbd>+<kbd>tab</kbd> to go back as expected, but cannot advance forward from open state.

On https://opensource.adobe.com/spectrum-web-components/components/action-menu tabbing from an example action menu button when closed advances to the "Copy to Clipboard" button. When open, goes to my browser tab outside of the document.

#### Steps to Reproduce

- Open action menu
- focus anywhere inside menu (focused by default on open)
- <kbd>tab</kbd>

Westbrook:
## Description
- ensure forward tab out of an `inline` overlay goes back to the trigger before moving the focus
  - make sure `open === false` before moving the focus
- update `sp-opened` timing so that it always happens _after_ any animation that might occur
- update testing dependencies

Can be previewed, here: https://westbrook-picker-focus--spectrum-web-components.netlify.app/components/action-menu

@cuberoot you most recently worked on the `sp-opened` eventing. How does this timing update strike you?

@adixon-adobe this might change again in #1424, but I think it still matters in the interim. My question is, do you see this effecting your "focus next form control" functionality? The "can throw focus after `change`" test passes just fine, but wanted to make sure you were part of the conversation here.

## Related Issue
fixes #1488 

## Motivation and Context
Tab order experience for inline overlays should be as if the content was opened _inline_...

## How Has This Been Tested?
- unit tests

## Types of changes

## Checklist:


najikahalsema:
So, if I understand correctly, if I do the following: 
- `Tab` 
- `Enter` (to open the menu) 
- `Tab` again
I should be focused on the menu trigger, right? Instead of the "Copy to Clipboard"? I tested this manually and that wasn't the case, so I just want to make sure I'm doing it properly.

image

cuberoot (approved):
This looks good to me. The tabbing in the the docs makes sense to me. I think shift-tab should take you back to the trigger, which it does.

Westbrook:
@najikahalsema, this is a great question to clarify! You're confirming the difference between two difference "types" of overlays:
1. `inline`: https://westbrook-picker-focus--spectrum-web-components.netlify.app/storybook/index.html?path=/story/overlay--inline (the version leveraged in the `sp-picker` element) places the overlay as if it were "between" the _trigger_ and the _next_ item in the tab order:
  > inline places the overlay after the trigger but before the next element in the logical tab order. This means the shift + tab keyboard stroke will return to the trigger. - https://opensource.adobe.com/spectrum-web-components/components/overlay
  - notice that the Storybook example does not throw focus the way the Picker does
2. `replace` https://westbrook-picker-focus--spectrum-web-components.netlify.app/storybook/index.html?path=/story/overlay--replace places the overlay as if it were the _trigger_, making the _trigger_ inaccessible via the tab order while the overlay is open.
  > replace will position the overlay directly in the position of the trigger in the logical tab order. This means the shift + tab keyboard stroke will return the focusable element immediately prior to the trigger. - https://opensource.adobe.com/spectrum-web-components/components/overlay

This PR is explicitly to correct existing functionality. However, in #1424 we discussing whether we're currently using the most expected type for this overlay.

Issue #1470, opened 2021-05-19
[Color Area] Extend set color method to use HSV values from input object


spdev3000:
### Expected Behaviour
Extend `set color` method to parse and use hsv values directly from input - if `color` input is an HSV/HSVA object like `{h: hueValue, s: sValue, v: vValue}`

This will prevent TinyColor's `toHsv()` method to override any given hue values with `0` if s-value is set to `0`

### Actual Behaviour
If I provide an input color object with s (saturation) = 0, TinyColor's `toHsv()` method will also set the hue value to `0` regardless of the provided input hue value.
This leads to resetting the hue value. Currently this can be temporary fixed if `hue` , `x` and `y` values are provided directly but still a color value needs to be set for default background color rendering of ColorHandle. 

#### Steps to Reproduce
- set start color value to `#00ff00`
- move color handle to left border of color area to select grey value or provide a grey value to color area , like `#c1c1c1` 
- this will set the hue part to `0` resulting in a pure red color area

#### Platform and Version
color-area: 0.1.4 on Chrome

Westbrook:
We can use this as a recreation to review options for moving forward here: https://webcomponents.dev/edit/Bv35bBB1gfphaS83UsOQ/src/index.stories.js

We documented this as a reality of the element, but it can/should be addressed to prevent these sorts of issue from occurring. In `sp-color-slider` and `sp-color-wheel` we extract the `hue` entry as `value` internally, and we could definitely do similarly in `sp-color-area`. I think the extra work needed here would include grafting that `hue` value back into the `get color` value returned when `format === 'hsl'` or `format === 'hsv'`. The main complexity of that would be inserting the value into the string results.

@spdev3000 would you be interested in structuring a PR to support inputs in this way?

Conversely, we could raise this to https://github.com/scttcper/tinycolor and discuss/submit a PR that would allow for managing colors inside of the TinyColor object in the format that they were originally supplied. Thoughts?

spdev3000:
I have checked the current source code of TinyColor - and they use internally RGBA data completely. That means they convert everything to RGBA on input and then handle a back conversation to each respective output format when requested. see: [toHsv()](https://github.com/scttcper/tinycolor/blob/e0a400a35d202d25e017bbb83e61cbc100703774/src/index.ts#L167 )
I doubt that they will restructure the complete core of importing/exporting color data for this use case so I would prefer to start with the first option you suggested. Let me check if I can open a PR on color-area.

PR #1507, opened 2021-06-03
[Color Area] Prevent TinyColor from overriding hue value on HSV/HSL colour formats

najikahalsema:
## Description

TinyColor has a problem where it, upon using HSL/HSV colour models, does not preserve the value of `hue` if the saturation is set to 0. By extracting the `hue` in the `set color()` method, as is done in `sp-color-slider` and `sp-color-wheel`, we can preserve the hue when it's given by the user. In the `get color()` method, we use a RegEx to remove the incorrect `hue` value and replace it with the `hue` property when the colour format is a string. When it is an object, we simply inject the `hue` property into an object and return that. 

## Related Issue

fixes https://github.com/adobe/spectrum-web-components/issues/1470

## Motivation and Context

Getting rid of a long-standing bug. Thanks, TinyColor.

## How Has This Been Tested?

Manually and with unit tests

## Types of changes

## Checklist:

### DIFF
--- a/packages/color-area/README.md
+++ b/packages/color-area/README.md
@@ -36,9 +36,6 @@ The current color formats supported are as follows:
 -   RGB, RGBA
 -   Strings (eg "red", "blue")
 
-**Please note for the following formats: HSV, HSVA, HSL, HSLA**
-When using the HSL or HSV formats, and a color's value (in HSV) is set to 0, or its luminosity (in HSL) is set to 0 or 1, the hue and saturation values may not be preserved by the element's `color` property. This is detailed in the [TinyColor documentation](https://www.npmjs.com/package/@ctrl/tinycolor).
-
 ## Standard
 
 ```html
--- a/packages/color-area/test/color-area.test.ts
+++ b/packages/color-area/test/color-area.test.ts
@@ -318,7 +318,7 @@ describe('ColorArea', () => {
         expect(el.x).to.equal(0.53125);
         expect(el.y).to.equal(0.53125);
     });
-    it('retains `hue` value when s = 0 in HSL format', async () => {
+    it('retains `hue` value when s = 0 in HSL string format', async () => {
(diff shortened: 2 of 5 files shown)


packages/color-area/README.md
```diff
@@ -36,9 +36,6 @@ The current color formats supported are as follows:
 -   RGB, RGBA
 -   Strings (eg "red", "blue")
 
-**Please note for the following formats: HSV, HSVA, HSL, HSLA**
```
Westbrook, line 39: Great call to pull this out!

packages/color-area/test/color-area.test.ts
```diff
@@ -318,6 +318,122 @@ describe('ColorArea', () => {
...
+
+        el.color = 'hsl(100, 0%, 50%)';
+        await elementUpdated(el);
+
+        expect(el.hue, 'new hue').to.equal(100);
+        expect(el.x, 'new x').to.equal(0);
+        expect(el.y, 'new y').to.equal(0.5);
+        expect(el.color).to.equal('hsl(100, 0%, 50%)');
+    });
+    it('retains `hue` value when s = 0 in HSL object format', async () => {
```
Westbrook, line 343: This is nice having the object and the string formats separated like this!

Westbrook (changes requested):
I would like the make sure that this looks good for @spdev3000 as well, but I think this is a great update to the functionality here!

The one thing I'd like added is a visual regression (story) for the way this successfully displays `sp-color-area[color=${{"h": 120, "s": 0, "v": 1}}]` element as "green" as opposed to the "red" that it would have done before this change.

spdev3000 (approved):
This looks really good to me and matches the described issue. Thanks Najika for helping out that quickly!

Westbrook (approved):
Thanks Najika, this looks great!

PR #1522, opened 2021-06-09
feat: remove theme fragment minimum and allow "app" fragments

### LINKED ISSUE #1496: [Theme] Allow Theme component to provide custom styles
In order for our application to scale properly, we would need the same `sp-theme` functionality not with Spectrum CSS variables, since the `sp-theme` component already provides them, but with our custom variables that we use across our components, that might or might not depend on some Spectrum CSS variables.

### Expected Behaviour
The expected behavior is to have a possibility to provide some application related styles via `sp-theme` component, not only Spectrum CSS styles.

For example: 
Locally, for our app, depending on the `scale` & `color` combination, we compute a list of local variables to be used across our app like this:
```
.app-container[scale="medium"] {
       --label-size: 0.75rem;
       --spacing-unit: 0.5rem;
       --header-height: calc(2.5rem + 1px);
}
.app-container[color="light"] {
      --label-color: var(--spectrum-alias-label-text-color);
}       
```
This CSS has to be attached to a child of the `<sp-theme>` in the DOM hierarchy to be able to also reference Spectrum variables.
What we would need is a way to have this local computed CSSResult provided by the `sp-theme` in order for us not to create a new local provider for the app's theme.

### Actual Behaviour
The actual behavior only allows us to send a combination of `color` and `scale` attributes depending on which the `sp-theme` provides the Spectrum CSS variables to be used in the app.

Westbrook:
## Description
- removes the fragment minimum before a "theme" is applied to `sp-theme`
- moves to an async rendering pattern to prevent multiple renders as various fragments are loaded
- adds support for an `app` fragment so that consuming application can hang local extensions of the Spectrum CSS Custom Properties API on the `sp-theme` element

# Questions:
- This approach does not support for more than one `app` fragment, is that OK?
- This approach does not support removing the `app` fragment, is that OK?
- @stanvladut pointed out how his team was previously overload fragments to achieve a similar goal, should we prevent that in some way?
- Is it OK to allow "partial" themes or should we ensure one of each fragment type via a more rigorous check?
- Are there edge/use cases that I've not thought of here?

## Related Issue
fixes #1496

## Motivation and Context
Easier customization of applications consuming Spectrum Web Component and associated styling APIs

## How Has This Been Tested?
- updated existing tests
- added new tests
- found [acceptable VRT failures](https://82623-158297197-gh.circle-artifacts.com/0/test/visual/review/index.html#ThemeStories/reverseColorNestedTheme.png) and adopted them (when we agreed that they were acceptable) 

## Types of changes
- [x] Refactor
- [x] _Possibly_ Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

### DIFF
--- a/packages/theme/src/Theme.ts
+++ b/packages/theme/src/Theme.ts
@@ -40,7 +40,7 @@ declare global {
     }
 }
 
-type FragmentType = 'color' | 'scale' | 'core';
+type FragmentType = 'color' | 'scale' | 'core' | 'app';
 type SettableFragmentTypes = 'color' | 'scale';
 type FragmentMap = Map<string, { name: string; styles: CSSResult }>;
 export type ThemeFragmentMap = Map<FragmentType, FragmentMap>;
@@ -48,7 +48,7 @@ export type Color = 'light' | 'lightest' | 'dark' | 'darkest';
 export type Scale = 'medium' | 'large';
 const ScaleValues = ['medium', 'large'];
 const ColorValues = ['light', 'lightest', 'dark', 'darkest'];
-type FragmentName = Color | Scale | 'core';
+type FragmentName = Color | Scale | 'core' | 'app';
 
 export interface ThemeData {
     color?: Color;
@@ -65,7 +65,6 @@ export interface ProvideLang {
 }
 
 export class Theme extends HTMLElement implements ThemeKindProvider {
-    private hasAdoptedStyles = false;
     private static themeFragmentsByKind: ThemeFragmentMap = new Map();
     private static defaultFragments: Set<FragmentName> = new Set(['core']);
     private static templateElement?: HTMLTemplateElement;
@@ -94,7 +93,6 @@ export class Theme extends HTMLElement implements ThemeKindProvider {
     }
 
     private requestUpdate(): void {
-        this.hasAdoptedStyles = false;
         if (window.ShadyCSS !== undefined && !window.ShadyCSS.nativeShadow) {
             window.ShadyCSS.styleElement(this);
         } else {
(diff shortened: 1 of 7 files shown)


packages/theme/src/Theme.ts
```diff
@@ -160,10 +154,20 @@ export class Theme extends HTMLElement implements ThemeKindProvider {
             const kindFragments = Theme.themeFragmentsByKind.get(
                 kind
             ) as FragmentMap;
-            const { [kind]: name } = this;
-            const currentStyles = kindFragments.get(name);
-            if (currentStyles) {
-                acc.push(currentStyles.styles);
+            const addStyles = (
+                name: FragmentType,
```
stanvladut, line 158: Shouldn't this be of type `FragmentName`, as it can have values other than `'color' | 'scale' | 'core' | 'app'`?

stanvladut (commented):
Exepting the minor comment I posted, the changes look good, great job. Thanks for working on this and delivering it so fast!

My opinion on some of the questions posted:
1. Generally speaking, I think it's enough (at least for my team) to only have one `app` fragment. We can compute our `app` styles locally, depending on `color`, `scale` or other factors and register it when ready.
2. I cannot think of a scenario where we would want that, as long as we can change the styles for the app fragment. Keeping in mind that the local `app` styles might also depend on `color` or `scale`, we might want to recompute the `app` styles and update them, but not remove.

najikahalsema (dismissed):

stanvladut (dismissed):

stanvladut (approved):
Great work! 💯

stanvladut (approved):
Great work!

Issue #1469, opened 2021-05-19
[Color Slider] Add dispatchChangeEvent for input value changes


spdev3000:
### Expected Behaviour
Color Slider should dispatch a change event when input values are changed. This should work [similar to Slider](https://github.com/adobe/spectrum-web-components/blob/50c8640a95baee56b5f1b74bbc66a69ef9443167/packages/slider/src/Slider.ts#L326) when the handle has focus and user tries to change values with cursor keys.

### Actual Behaviour
No change event is dispatched

#### Steps to Reproduce
Set focus to Color Slider handle
key down cursor up/down moves the handle
no change event is dispatched to outer component 

#### Platform and Version
color-slider: 0.1.4 on Chrome

Westbrook:
Indeed it does. Easy to forget that native `<input>` elements don't have their `change` events composed and need to be re-dispatched in our case.

It appears on first look the `sp-color-wheel` shares code shape here, and will likely need this same update. However, `sp-color-area` seems to have mitigations for this. Have you see similar as part of your consumption?

spdev3000:
I currently only use Color Area (Handle) and Color Slider - and it only happens at the Slider component. Color Area works fine for me (besides #1470).

Westbrook:
@najikahalsema let me know if this helps to clarify the issue here.

Good Steps:
1. Visit https://webcomponents.dev/edit/jC0UwYyeK45myyAsMkYm/src/index.ts
2. Click, drag, and release the handle on any of the three controls
3. Notice that you get a stream of `input` events and one `change` event at the end

Bad Steps (area) keyboard:
1. Visit https://webcomponents.dev/edit/jC0UwYyeK45myyAsMkYm/src/index.ts
2. Visually focus the handle in the `sp-color-area` element (it should get bigger)
3. Use the arrow keys to adjust the value of the control
4. Notice the `change` event dispatched on each arrow keys, but no `input` event

Bad Steps (area) Voice Over:
1. Visit https://webcomponents.dev/edit/jC0UwYyeK45myyAsMkYm/src/index.ts
2. Visually focus the handle in the `sp-color-area` element (it should get bigger)
3. Focus the control
4. Enter the slider with `Shift + Control + Option + Down Arrow`
5. Press `Control + Option` => See the input event dispatch for some reason
6. Press `Control + Option + Arrow *` => See the change event dispatch but no `input` event

Bad Steps (slider) keyboard:
[...]

PR #1526, opened 2021-06-10
Add change event to color slider

najikahalsema:
## Description

The colour slider was not dispatching the input event to its external elements, and thus wasn't properly displaying the hue value when the user adjusts the slider. This can be seen in the current implementation, where adjusting the hue slider in the joint Colour Area story doesn't always match up with what's being shown in the colour area. 

## Related Issue

Fixes https://github.com/adobe/spectrum-web-components/issues/1469

## Motivation and Context

Provides a better, less buggy UX

## How Has This Been Tested?

manually and with written tests

## Types of changes

## Checklist:

### DIFF
--- a/packages/color-slider/test/color-slider.test.ts
+++ b/packages/color-slider/test/color-slider.test.ts
@@ -80,6 +80,8 @@ describe('ColorSlider', () => {
         input.dispatchEvent(arrowUpKeyupEvent);
 
         await elementUpdated(el);
+        await elementUpdated(el);
+        await elementUpdated(el);
 
         expect(el.sliderHandlePosition).to.equal(2);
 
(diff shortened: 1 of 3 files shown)


packages/color-slider/test/color-slider.test.ts
```diff
@@ -448,6 +450,26 @@ describe('ColorSlider', () => {
 
         expect(el.sliderHandlePosition).to.equal(46.875);
     });
+    it('accepts input events', async () => {
```
Westbrook, line 453: To ensure that we've appropriately addressed the originating issue, can you add a `changeSpy` style test, a la: https://github.com/adobe/spectrum-web-components/blob/1ac12931771c6d5fdbc99f5d214702ed644cb81a/packages/sidenav/test/sidenav.test.ts#L182 

Each keyboard interaction with this element should dispatch one `input` and one `change` event. In the case that you make a protracted `pointer` event, it should dispatch `input` event(s) for each change in value until `pointerup` at which point a `change` event should be dispatched.

Westbrook:
Take a more focused look at the pieces of this issue in another PR.

Issue #1528, opened 2021-06-10
Scrubbable Numeric Input


bengfarrell:
### Expected Behaviour
- Numeric input should offer an ability to turn on "scrubbing" or click and drag to increment/decrement
- ~Also, a user should be able to hold a shift key down to accelerate the steps when using the arrow keys.~

So, I could make a PR for this, but I think I'm too far ahead of the curve to even be under consideration against Spectrum design recommendations. I'm also messing with component and input element focus in ways that probably need more thought (that might impact FocusableBase maybe). It's also up to opinion whether the properties I created should act and are named the way I did them.

However, given that I thought more projects might want the feature, I thought I'd offer up a first pass, including storybook documentation as well.

For me, I'm just going to copy and paste into my project regardless, but maybe this'll help!

https://github.com/bengfarrell/spectrum-web-components

Westbrook:
# Modified Step

I really like this and would love to work out the right way to bring this into the library. Currently, we leverage a version of this in the Color elements. There we actually allow _multiple_ modifier keys to multiply the multiplier: https://github.com/adobe/spectrum-web-components/blob/main/packages/color-area/src/ColorArea.ts#L212

Maybe that's overkill? Either way would that be a nice feature to abstract in a reusable manner across elements including the Number Field, standard Slider and Color Elements broadly? 

While I'm thinking in this area, I also wonder if `Home` and `End` should move to the bounds of `min`/`max` while available? Something to consider separately, probably.

Could I convince you to submit a PR with this multiplier functionality? I think regardless of merging the nuances between this and other elements that we can likely adopt your approach here ASAP, even if it might not be 100% "Spectrum". Then we can refine its implementation and how it fits into the larger library over time.

# Scrubbing

This is functionality that we are 100% planning to bring into the library, and I'd be quite happy to start doing so with a PR here as well: separate, preferably.

It's not immediately clear to me from reading the code how it works. Particularly, I see the `scrub` functionality stemming from the same listener as the `stepper` UI, which I will have to spend some more time with to fully understand. However, if it's working for you, I'd be quite happy to leverage this as the basis for the same in the main library. 

One nuance question that I'll have to dig into the Spectrum docs for is whether you can `scrub` the input, or just the `buttons`. That may be overly restricting, in that the buttons are optional, but shouldn't be a big deal to move around as needed.

[...]

Issue #1531, opened 2021-06-11
Step in sp-number-field does not affect scrolling


davidkudera:
### Expected Behaviour
When using scroll wheel within focused `<sp-number-field>` the number should change by the configured `step` value.

### Actual Behaviour
It goes always up/down by `100`.

#### Platform and Version
* @spectrum-web-components/bundle: 0.21.2
* Windows 10
* Chrome 91.0.4472.77

#### Sample Code that illustrates the problem
```
<sp-number-field step="1"></sp-number-field>
```

Should behave similarly to:

```
<input type="number" step="1">
```

Westbrook:
@davidkudera thanks for submitting. I'm fairly certain this is due to differences in mouse settings, but I am no able to test (no mouse 🙈 ). Would you be able to visit https://westbrook-numberfield-wheel-step--spectrum-web-components.netlify.app/components/number-field and let me know if this performs more how you would expect in this context? If that's so, hopefully we can have this out in a release next week sometime.

Thanks so much! 🙏🏼

bengfarrell:
@Westbrook  Howdy, I'm actually looking at this myself now. I was considering adding a shift modifier key to the wheel event, but inline this issue, it doesn't really make sense.

The action uses the event.deltaY property instead of the step. And it looks like pixel based values are getting returned which can result in HUGE increments if you naturally scroll a bit aggressively.

Would you prefer I take a stab at doing something a bit more naturally in my shift modifier PR as well as apply the shift modifier to it as well?

My take on it would be just to do it event based a step at a time, ignoring the delta values form the mouse

davidkudera:
@Westbrook Hello, your example works better for me. Value is updated by `1` which is how the native input works when no custom step is defined.

But, there is another problem I mentioned here (https://github.com/adobe/spectrum-web-components/issues/1509) - scrolling goes in opposite direction compared to native input.

I prepared a new example - https://webcomponents.dev/edit/i9wbtQ3zlhqGQBDZk7oi/src/index.ts

Behavior when I use "scroll wheel direction - traditional":

* `sp-number-field`:
	+ scrolling up: decrease value
	+ scrolling down: increase value
* `input type=number`:
	+ scrolling up: increase value
	+ scrolling down: decrease value

And when I use "scroll wheel direction - inverted":

* `sp-number-field`:
	+ scrolling up: increase value
	+ scrolling down: decrease value
* `input type=number`:
	+ scrolling up: decrease value
[...]

PR #1534, opened 2021-06-11
fix(number-field): add support for modified stepping

bengfarrell:
Step multiplier for faster stepping

## Description

Added step multiplier - this will multiply the step when holding down shift. Originally I had this working for just the arrow keys, but added support for pressing the stepper buttons as well. The multiplier is a settable attribute called "step-modifier".

We had discussed allowing any modifier, but instantly I noticed that CTRL + ArrowUp does a spaces thing on OSX, so I'm limiting to just shift right now.

## Related Issue

https://github.com/adobe/spectrum-web-components/issues/1528

## Motivation and Context

Motivation is largely for my project which includes a canvas where objects on canvas have transforms. Stepping by 1px at a time is super slow, and our designer wanted a way to increase that 10x.

## How Has This Been Tested?

Just lightly tested in Chrome as I was developing

## Types of changes

## Checklist:

- [x ] I have signed the [Adobe Open Source CLA](http://opensource.adobe.com/cla.html).


packages/number-field/src/NumberField.ts
```diff
@@ -239,37 +244,40 @@ export class NumberField extends TextfieldBase {
...
+        this.stepBy(1 * factor);
     }
 
-    private decrement(): void {
-        this.stepBy(-1);
+    private decrement(factor = 1): void {
+        this.stepBy(-1 * factor);
     }
 
+    private shiftPressed = false;
```
Westbrook, line 255: Thought on leveraging https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/shiftKey as opposed to hanging additional state off of the element?

You likely adopted this from the fact that it isn't used in the Color elements, but likely should be. Here where it is just the one, it seems like a good idea.

Westbrook (commented):
Super conscience, I love it!

Would you have some time to drop a test onto this before we approve to merge?

packages/number-field/src/NumberField.ts
bengfarrell: Weird! For some reason I didn't see the test file for Numberfield, I remembered it being empty, but thats obviously not the case. I'll get on it!

In the meantime, the shift key is tricky. So here's the deal - I thought it should work not only for arrow key input, but also the stepper buttons and for scrubbing in my other PR as well.

For these other modalities, we're relying on PointerEvents. I really just assumed that `shiftKey` couldn't be queried from a non-keyboard event, but I guess I'm wrong. Even so, I don't see the property documented on MDN, and I'm wondering how global and reliable it is.

Secondly, the stepper buttons operate on a timer, so aren't even triggered by a proper event. I might be able to nest the PointerEvent in the closure, but setting state just seemed more natural and straightforward.

Lemme know what you think!

bengfarrell: Oops, scratch that first point - I just didn't look down the inheritance chain to MouseEvent. But I think my second point stands.

bengfarrell: So, forget everything I said. Here's what happened...
I realized that if the modifier key is going to be here, it should apply to all situations including the scroll wheel which I did not cover. Unfortunately, I found the scroll a bit crazy. I can scroll a bit and suddenly find myself in the 100k's or even millions. [This issue](https://github.com/adobe/spectrum-web-components/issues/1531) seemed to have the same complaints if I were to guess.

Your demo to the person who filed the issue seems exactly inline with how I solved, except the shift kills the input (OK for you, not for this PR though). Since you don't have a scroll wheel, I'll tell you that the shift key modifies the axis of input. When before you got deltaY, with shift held you get deltaX.

Anyway, that's my functionality change here in the latest.

I also found that my shift state that you called out wasn't being set properly for when I was scrolling. Not sure why, but I just took that as a sign to kill it. I ended up carrying the pointer events throughout the timed change requests for the stepper buttons.

In terms of testing, I added 4 tests. One each for default shift Arrow Up/Down, and one each for changing step and the step modifier values on Arrow Up/Down.

I also needed to alter a couple other tests because they were using the ArrowKey + Shift as a way to clear out the field and start over.

Lastly, I tried to make some tests around Shift + stepper buttons/scroll wheel, but wasn't sure how to send button/wheel events at the same time as a shift key. I'm assuming that what I already did is probably good enough, though.

Westbrook:
@bengfarrell sorry for the delay on this. Would you be able to clean up this one conflict?

Then I'll merge this down into the holding branch and make sure the tests/et al are passing and we can get this into next week's release!

bengfarrell:
> @bengfarrell sorry for the delay on this. Would you be able to clean up this one conflict?
> 
> Then I'll merge this down into the holding branch and make sure the tests/et al are passing and we can get this into next week's release!

Sure thing! Done. The only difference in the new logic was obviously the shift modifier, but also checking that the deltaX/Y is not 0 before running the stepBy function. So I used my logic, and added that check back in

Westbrook (approved):
Awesome. Thanks!

PR #1579, opened 2021-06-24
fix: normalize wheel input directionally for more predictable input

### LINKED ISSUE #1531: Step in sp-number-field does not affect scrolling

### LINKED ISSUE #1553: [Number Field] Step direction when scrolling.
At current, we are matching other approaches to this functionality at Adobe, and it may be difficult to move away from it. However, as reported by @davidkudera...

I prepared a new example - https://webcomponents.dev/edit/i9wbtQ3zlhqGQBDZk7oi/src/index.ts

Behavior when I use "scroll wheel direction - traditional":

* `sp-number-field`:
	+ scrolling up: decrease value
	+ scrolling down: increase value
* `input type=number`:
	+ scrolling up: increase value
	+ scrolling down: decrease value

And when I use "scroll wheel direction - inverted":

* `sp-number-field`:
	+ scrolling up: increase value
	+ scrolling down: decrease value
* `input type=number`:
	+ scrolling up: decrease value
	+ scrolling down: increase value

So no matter the settings I use, the direction is always opposite to native behavior. Maybe I should create a new issue for that?

_Originally posted by @davidkudera in https://github.com/adobe/spectrum-web-components/issues/1531#issuecomment-863847284_

Westbrook:
## Description
Use wheel directionally rather than based on "distance" for more predictable input.

## Related Issue
fixes #1531

## Motivation and Context
Users should not be surprised.

## How Has This Been Tested?
Unit tests.

https://westbrook-numberfield-wheel-step--spectrum-web-components.netlify.app/components/number-field

## Types of changes

## Checklist:

### DIFF
--- a/packages/number-field/src/NumberField.ts
+++ b/packages/number-field/src/NumberField.ts
@@ -269,7 +269,10 @@ export class NumberField extends TextfieldBase {
 
     protected onScroll(event: WheelEvent): void {
         event.preventDefault();
-        this.stepBy(event.deltaY);
+        const { deltaY } = event;
+        if (deltaY !== 0) {
+            this.stepBy(deltaY / Math.abs(deltaY));
+        }
     }
 
     protected onFocus(): void {
--- a/packages/number-field/test/number-field.test.ts
+++ b/packages/number-field/test/number-field.test.ts
@@ -194,7 +194,7 @@ describe('NumberField', () => {
             expect(el.formattedValue).to.equal('0');
             expect(el.valueAsString).to.equal('0');
             expect(el.value).to.equal(0);
-            el.dispatchEvent(new WheelEvent('wheel', { deltaY: 1 }));
+            el.dispatchEvent(new WheelEvent('wheel', { deltaY: 100 }));
             await elementUpdated(el);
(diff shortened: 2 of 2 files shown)


najikahalsema (commented):
So, I tested this with my mouse (I actually have a scroll wheel!!!) and it works great, save for the inversion. I saw that you mentioned that in your issue linked to this that you attempted to try and fix that... Is that still something we can't resolve? If not, I think it's fine as is, but we may want to denote that in the docs, create a separate issue, etc.

Westbrook:
Indeed. If there is a path to altering that later on, we'll come back to this. Also, I corrected the issue that this PR closes so that we can continue to track the direction question past this fix.

najikahalsema (approved):

PR #1580, opened 2021-06-24
fix: correct a11y representation of a radio group

### LINKED ISSUE #1278: [radio] a11y review
### Feedback

- Radio buttons are not associated with each other
  - in html the name of the buttons is used to create the group structure (and control the tab focus order)
here the group is named, and the buttons have empty name attribute
  - this results in screen readers not knowing the buttons are associated together
- Radio group needs role=”radiogroup” and a way to specify a group label
  - and example of labeling the group should be provided.
- Each radio button in the radio group is read as option 1 of 1 Each radio button is appearing to be in its own “group”

### Path forward

1. We're both tracking https://bugs.chromium.org/p/chromium/issues/detail?id=1158144#c_ts1615261031 and needing to update the structure of the `sp-radio` elements to support usage of that pattern. In this way we'll likely having it working fully in Safari/Firefox and then just have to wait for Chrome... this will also support #768 
2. General radio focus is being reviewed as part of #1038 
3. Add `radiogroup` role, add `label` => `aria-label` attribute. Do groups need a visible label?

cc @jnurthen

Westbrook:
## Description
While still tracking https://bugs.chromium.org/p/chromium/issues/detail?id=1158144#c_ts1615261031, this update allows both Firefox and Safari to show the posinset and setsize data automatically through appropriately preparing the "radiogroup" root and associating it correctly with child "radio" elements. Once the linked bug in Chrome is addressed, these updated should automatically be available in that context.

Note in the Chrome issue that the grouping issue also effects the naming in certain contexts. If this doesn't get addressed directly by the issue, we'll start a new issue to track further fixes in the browser.

## Related Issue
fixes #1278

## Motivation and Context
Accessibility!

## How Has This Been Tested?
Manually, unit tests, etc.

image

## Types of changes

## Checklist:

### DIFF
--- a/packages/radio/src/Radio.ts
+++ b/packages/radio/src/Radio.ts
@@ -17,10 +17,11 @@ import {
     CSSResultArray,
     TemplateResult,
     PropertyValues,
+    SpectrumElement,
 } from '@spectrum-web-components/base';
+import { FocusVisiblePolyfillMixin } from '@spectrum-web-components/shared/src/focus-visible.js';
 
 import radioStyles from './radio.css.js';
-import { Focusable } from '@spectrum-web-components/shared/src/focusable.js';
 
 /**
  * Spectrum Radio Button Component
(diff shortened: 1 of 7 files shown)


najikahalsema (approved):
This is so beautiful, I'm crying.

Issue #290, opened 2019-11-11
Implement `sp-table` component


Westbrook:
### Expected Behaviour
We shouldn't have to rely on Spectrum CSS directly for our documentation site. We're pointing to the `table` sheet, we should have a component or a pattern that supports including that in a page easily.

### Actual Behaviour
https://opensource.adobe.com/spectrum-css/table.html

hunterloftis:
Review pre-1.0.

Westbrook:
@lastmjs there should be no trouble implementing something in Spectrum CSS inline with things shipped via SWC. You would benefit from ensuring that the version of the Spectrum CSS package you are leveraging depends on the same Spectrum CSS Vars package that is leveraged in `<sp-theme>`, assuming you use that, in order to make sure there are no subtle rendering bugs, but other than that, you should be good to go.

We are just beginning work with a team that intends to consumer the `sp-table` pattern, so hopefully we'll also have an update on SWC native support for this pattern, soon.

Cheers!

lastmjs:
I'm super excited for `sp-table`! Once that is implemented we might make a full switch over. Thanks!

PR #1603, opened 2021-07-06
[RFC] feat(table): add table pattern

davidkudera:
## Description

This is a basic and really naive implementation of the [table](https://opensource.adobe.com/spectrum-css/table.html) pattern.

I think that it could be better if someone with more experience actually implemented the table pattern given how table components are complicated. But even if someone else was to implement it, this PR could still be used as an inspiration (maybe).

Here is a list of things that are missing (and probably many more I missed):
* Only div style tables, no `<table>` tables
* No multi-select, column dividers, and other variants
* No tests
* Position of icons in header cells is off
* I'm not sure if I used the slots correctly (I don't think I did)
* I didn't check anything related to accessibility
* I'm not sure about the names of the elements
* No documentation
* Storybook should contain more examples
* Storybook examples have wrong ordering icons
* No updates to `spectrum-config.js`
* Wouldn't using light DOM be easier and preferable?

Currently there are 2 ways for defining tables (both in storybook):

1. using properties
```html
<sp-table .header=${header} .data=${data}></sp-table>
```
2. using slots
```html
<sp-table>
  <sp-head-cell>Head cell 1</sp-head-cell>
  <sp-head-cell>Head cell 1</sp-head-cell>
  <sp-row>
    <sp-cell>Cell 1</sp-cell>
    <sp-cell>Cell 2</sp-cell>
  </sp-row>
</sp-table>
```

The second way is just an extension of the first one and I believe that both are equally useful. 1st option is suitable for constructing tables from JS (maybe with data from some API). The 2nd option is convenient when creating tables that contain eg. unrelated data or when using some backend framework (dotnet in our case):

```cshtml
<sp-table>
  <sp-head-cell>Name</sp-head-cell>
  <sp-head-cell>Email</sp-head-cell>

  @foreach (var user in Model.Users)
  {
    <sp-row>
      <sp-cell>@user.Name</sp-cell>
      <sp-cell>@user.Email</sp-cell>
    </sp-row>
  }
</sp-table>
```

But the 2nd option is slower to render. Also, it is possible to use a combination of both methods.

## Related Issue

* https://github.com/adobe/spectrum-web-components/issues/290

## Motivation and Context

Spectrum CSS contains tables but there is no implementation of tables in spectrum web components.

## How Has This Been Tested?

* Visually in storybook
* By using this implementation in our dashboard (but just briefly)

## Types of changes

## Checklist:

### DIFF
--- a/packages/table/.npmignore
+++ b/packages/table/.npmignore
@@ -1,0 +1,2 @@
+stories
+test
--- a/packages/table/README.md
+++ b/packages/table/README.md
@@ -1,0 +1,28 @@
+## Description
+
+### Usage
+
+[![See it on NPM!](https://img.shields.io/npm/v/@spectrum-web-components/table?style=for-the-badge)](https://www.npmjs.com/package/@spectrum-web-components/table)
+[![How big is this package in your project?](https://img.shields.io/bundlephobia/minzip/@spectrum-web-components/table?style=for-the-badge)](https://bundlephobia.com/result?p=@spectrum-web-components/table)
+
+```
+yarn add @spectrum-web-components/table
+```
+
+Import the side effectful registration of `<sp-table>` via:
+
+```
+import '@spectrum-web-components/table/sp-table.js';
+```
+
+When looking to leverage the `Table` base class as a type and/or for extension purposes, do so via:
+
+```
+import { Table } from '@spectrum-web-components/table';
+```
+
+## Example
+
+```html
+<sp-table></sp-table>
+```
--- a/packages/table/package.json
+++ b/packages/table/package.json
@@ -1,0 +1,58 @@
+{
+    "name": "@spectrum-web-components/table",
+    "version": "0.0.1",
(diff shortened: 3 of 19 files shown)


Westbrook:
@davidkudera very interesting to see this! Thanks for the interest. I've changed the base branch here from `main` so we could accept this change into the repo as an example for developing a more complete Table over time. If you're OK with it, I'd love to accept this PR into that branch, which will also give us a preview site, etc. and start the larger conversation in the coming month when the team grows with a couple of new members. Sound good?

davidkudera:
@Westbrook that sounds awesome, I wasn't expecting that this could be merged before all of the problems are solved. But it is exactly what I was hoping for - mainly to start the conversation. So thank you.

I'll probably send more PRs related to this based on our usage of the table component. I think that it would be nice if there was some issue (or many issues) tracking the progress of table component implementation. And then there are also some questions, like elements names and others.

Issue #1610, opened 2021-07-07
[Split view] primarySize can't be updated


davidkudera:
### Expected Behaviour
Changing the `primary-size` attribute in `<sp-split-view>` should update the size of primary and secondary views.

### Actual Behaviour
Changing it does nothing. I tried numbers, percentages, and pixels too.

### Reproduce Scenario (including but not limited to)
Using the Split view component with provided size

#### Steps to Reproduce
Here is an example with a toggle button that is changing the value from `20%` to `80%` and back: https://webcomponents.dev/edit/13P6vLAg4Fb6SGvVYtIm/src/index.ts

#### Platform and Version
* Spectrum web components ^0.22.2
* Edge 91.0.864.64

Westbrook:
First thought... updated:
```
    protected firstUpdated(changed: PropertyValues): void {
        super.firstUpdated(changed);
        this.checkResize();
    }

    protected updated(changed: PropertyValues): void {
        super.updated(changed);
        if (
            changed.has('splitterPos') &&
            this.splitterPos !== undefined &&
            this.enoughChildren
        ) {
            this.firstPaneSize = `${Math.round(this.splitterPos)}px`;
        }
    }
```
to be
```
    protected updated(changed: PropertyValues): void {
        super.updated(changed);
        if (changed.has('primarySize') {
             this.checkResize();
        }
        if (
            changed.has('splitterPos') &&
            this.splitterPos !== undefined &&
            this.enoughChildren
        ) {
            this.firstPaneSize = `${Math.round(this.splitterPos)}px`;
        }
    }
```
And you'd be good to go...in that case that you're interested in submitting a PR. @spdev3000 does that sound right-ish?

Otherwise, we've got a number of people of holiday, so it'll be another couple of weeks before we'd be able to actively take a look at this.

davidkudera:
@Westbrook I was going to try to create a PR and you just made it much easier for me. Thank you

PR #1611, opened 2021-07-08
feat(split-view): redraw when primary-size change

davidkudera:
## Description

This change automatically updates the splitter position when `primarySize` property changes.

## Related Issue

https://github.com/adobe/spectrum-web-components/issues/1610

## Motivation and Context

According to docs `primarySize` can be set with the `primary-size` attribute. Right now that applies only to the initial state. This change allows (us) to use the split view component as some kind of sidebar which is opened/closed automatically based on some user action.

## How Has This Been Tested?

* Added test
* Added `primarySize` control to storybook (just to the first example)
* By using this change with patch-package in the thing I'm working on

## Types of changes

## Checklist:

### DIFF
--- a/packages/split-view/src/SplitView.ts
+++ b/packages/split-view/src/SplitView.ts
@@ -401,6 +401,10 @@ export class SplitView extends SpectrumElement {
 
     protected updated(changed: PropertyValues): void {
         super.updated(changed);
+        if (changed.has('primarySize')) {
+            this.splitterPos = undefined;
+            this.checkResize();
+        }
         if (
             changed.has('splitterPos') &&
             this.splitterPos !== undefined &&
--- a/packages/split-view/stories/split-view.stories.ts
+++ b/packages/split-view/stories/split-view.stories.ts
@@ -19,13 +19,20 @@ export default {
     component: 'sp-split-view',
 };
 
-export const Horizontal = (): TemplateResult => {
+interface Properties {
+    primarySize?: string;
+}
+
+export const Horizontal = (args: Properties): TemplateResult => {
     return html`
-        <sp-split-view style="height: 200px" primary-size="100">
+        <sp-split-view style="height: 200px" primary-size="${args.primarySize}">
             <div>First panel</div>
             <div>Second panel</div>
         </sp-split-view>
     `;
+};
+Horizontal.args = {
+    primarySize: '100',
 };
 
 export const HorizontalResizable = (): TemplateResult => {
--- a/packages/split-view/test/split-view.test.ts
+++ b/packages/split-view/test/split-view.test.ts
@@ -1031,4 +1031,24 @@ describe('SplitView', () => {
         expect(el.splitterPos || 0).to.equal(pos - 10);
         expect(changeSpy.callCount).to.equal(1);
     });
+
+    it('resizes when primarySize changes', async () => {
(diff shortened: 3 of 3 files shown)


packages/split-view/src/SplitView.ts
```diff
@@ -303,12 +303,12 @@ export class SplitView extends SpectrumElement {
...
     }
 
-    private async checkResize(): Promise<void> {
+    private async checkResize(forceUpdate = false): Promise<void> {
         if (!this.enoughChildren) {
             return;
         }
         this.updateMinMax();
-        if (this.splitterPos === undefined) {
+        if (this.splitterPos === undefined || forceUpdate) {
```
davidkudera, line 311: I'm not really sure about this, I feel like there could be some better way instead of adding a new boolean argument. Removing this condition breaks the `keeps the splitter pos when removing and re-adding a panel` test.

davidkudera:
Failing CI does not seem to be related to my changes I think...

packages/split-view/src/SplitView.ts
```diff
@@ -401,6 +401,9 @@ export class SplitView extends SpectrumElement {
 
     protected updated(changed: PropertyValues): void {
         super.updated(changed);
+        if (changed.has('primarySize')) {
+            this.checkResize(true);
```
Westbrook, line 405: Can't spend too much time digging in, as I'm headed out of town, but you _could_...
```
this.splitterPos = undefined;
this.checkResize();
```
To prevent adding the argument, though it likely disrupts something elsewhere?

packages/split-view/stories/split-view.stories.ts
```diff
@@ -19,14 +19,21 @@ export default {
     component: 'sp-split-view',
 };
 
-export const Horizontal = (): TemplateResult => {
+interface Properties {
+    primarySize?: number;
```
Westbrook, line 23: If this was `string` could we then input `80%`, etc.? Is that helpful?

Westbrook (commented):
Yes, the CI failure is on our side. We don't get quite enough external contributions to invest in making that better, but you may be tipping us across that line. When I get back in the second half of next week, I can create a landing branch to merge this into before going to `main` in order to confirm the full CI process.

Thanks for looking into this!

packages/split-view/stories/split-view.stories.ts
davidkudera: Oh, yeah I wanted to change that, but then I forgot. Thanks

packages/split-view/src/SplitView.ts
davidkudera: I'll try that

davidkudera:
@Westbrook there is no need to hurry with this PR. Until it is merged we'll just use the patch-package. So take your time and enjoy your holiday 👋

packages/split-view/src/SplitView.ts
davidkudera: It seems to work too. All tests are passing and changing the value in the storybook redraws the component correctly.

You can pick the preferable variant.

Westbrook (approved):
@davidkudera thanks for this! I'm going to accept it into a landing branch so we can get all of the CI goodness and a preview site to test it out manually, but I'm pretty sure this is what we want here. 🙏🏼

PR #1641, opened 2021-07-23
[Color Area] Dispatch change event when using keyboard

najikahalsema:
## Description

Similar to the other colour elements, ColorArea was not correctly dispatching change events when using the keyboard. That is now fixed. Hooray. Also, altering other tests that were previously incorrect/not optimal.

## Related Issue

fixes https://github.com/adobe/spectrum-web-components/issues/1469 by extension

## Motivation and Context

Matching functionality with other colour elements and providing accessibility for users 

## How Has This Been Tested?

Manually and with tests

## Types of changes

## Checklist:

### DIFF
--- a/packages/color-area/src/ColorArea.ts
+++ b/packages/color-area/src/ColorArea.ts
@@ -211,6 +211,7 @@ export class ColorArea extends SpectrumElement {
 
     private handleKeydown(event: KeyboardEvent): void {
         const { key, code } = event;
+        event.preventDefault();
         if (['Shift', 'Meta', 'Control', 'Alt'].includes(key)) {
             this.altKeys.add(key);
             this.altered = this.altKeys.size;
@@ -247,37 +248,35 @@ export class ColorArea extends SpectrumElement {
                     break;
             }
         });
-        if (deltaX != 0 || deltaY != 0) {
+        if (deltaX != 0) {
             this.inputX.focus();
-            this.inputX.dispatchEvent(
-                new Event('input', {
-                    bubbles: true,
-                    composed: true,
-                })
-            );
+        } else if (deltaY != 0) {
             this.inputY.focus();
-            this.inputY.dispatchEvent(
-                new Event('input', {
-                    bubbles: true,
-                    composed: true,
-                })
-            );
         }
+
         this.x = Math.min(1, Math.max(this.x + deltaX, 0));
         this.y = Math.min(1, Math.max(this.y + deltaY, 0));
 
         this._previousColor = this._color.clone();
         this._color = new TinyColor({ h: this.hue, s: this.x, v: 1 - this.y });
 
-        const applyDefault = this.dispatchEvent(
-            new Event('change', {
-                bubbles: true,
-                composed: true,
-                cancelable: true,
-            })
-        );
-        if (!applyDefault) {
-            this._color = this._previousColor;
+        if (deltaX != 0 || deltaY != 0) {
+            this.dispatchEvent(
+                new Event('input', {
+                    bubbles: true,
+                    composed: true,
+                })
+            );
+            const applyDefault = this.dispatchEvent(
+                new Event('change', {
+                    bubbles: true,
+                    composed: true,
+                    cancelable: true,
+                })
+            );
+            if (!applyDefault) {
+                this._color = this._previousColor;
+            }
         }
     }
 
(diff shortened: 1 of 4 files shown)


packages/color-area/src/ColorArea.ts
```diff
@@ -247,10 +247,34 @@ export class ColorArea extends SpectrumElement {
...
+        if (deltaX != 0) {
             this.inputX.focus();
-        } else if (deltaY) {
+            this.inputX.dispatchEvent(
+                new Event('input', {
+                    bubbles: true,
+                    composed: true,
+                })
+            );
+            // this.inputX.dispatchEvent(
```
Westbrook, line 258: I think if we `event.preventDefault();` in line 213, to make this work a little more like `ColorSlider` then we can uncomment these `change` events.

If not, we should remove them.

Also, we should join the dispatches under a single conditional after the focus logic, something like:
```
if (deltaX != 0 || deltaY != 0) {
   // dispatch here, so it's in one place
}
```

packages/color-area/test/color-area.test.ts
```diff
@@ -160,13 +161,16 @@ describe('ColorArea', () => {
...
+        expect(el.x).to.equal(0.67);
         expect(el.y).to.equal(0.25);
     });
     it('accepts "Arrow*" keypresses with alteration', async () => {
         const el = await fixture<ColorArea>(
             html`
-                <sp-color-area color="hsla(100, 50%, 50%, 1)"></sp-color-area>
+                <sp-color-area
+                    color="hsla(100, 50%, 50%, 1)"
+                    style="--spectrum-global-dimension-size-2400:192px"
```
Westbrook, line 172: Can you check if we ended up needing this?

packages/color-area/test/color-area.test.ts
```diff
@@ -325,6 +333,82 @@ describe('ColorArea', () => {
...
+        el.inputY.dispatchEvent(
+            new Event('input', {
+                bubbles: true,
+                composed: true,
+            })
+        );
+        el.inputY.dispatchEvent(
+            new Event('change', {
+                bubbles: true,
+                composed: false,
```
Westbrook, line 380: Can you comment here too: `// native change events do not compose themselves by default`, it's just such a finicky reality...

packages/color-area/test/color-area.test.ts
```diff
@@ -325,6 +333,82 @@ describe('ColorArea', () => {
...
+                ></sp-color-area>
+            `
+        );
+
+        await elementUpdated(el);
+
+        el.inputX.focus();
+
+        await sendKeys({ press: 'ArrowRight' });
+        await sendKeys({ press: 'ArrowRight' });
```
Westbrook, line 405: Might be worth doing all four directions here, for "completeness"?

Westbrook (changes requested):
This looks great!

A couple of double checks to make sure we're not leaving extra code laying around and we'll be good.

Also, the Github Actions failure is their fault, not ours.

packages/color-area/test/color-area.test.ts
najikahalsema: Doesn't seem like it.

packages/color-area/src/ColorArea.ts
najikahalsema: I did check this out, and it didn't seem to work the way it was expected. I removed the change events for this reason.

packages/color-area/test/color-area.test.ts
```diff
@@ -325,6 +330,112 @@ describe('ColorArea', () => {
...
+
+        await elementUpdated(el);
+
+        expect(inputSpy.callCount).to.equal(4);
+        expect(changeSpy.callCount).to.equal(2);
```
Westbrook, line 407: These should match, I think. One input and one change event for each key listened to... can you check what's causing this? Seems like maybe the `event.preventDefault()` note I sent before, but possibly something else.

Westbrook: For reference: https://github.com/adobe/spectrum-web-components/blob/main/packages/color-slider/test/color-slider.test.ts#L80-L95

Westbrook (approved):
Thanks so much for this. Looks like the branch is a little behind `main` but once that's caught up, you're good to go!

Westbrook (approved):

PR #1646, opened 2021-07-27
fix(split-button): hide "selected" item from menu

### LINKED ISSUE #1456: More variant of split button can have value change
### Expected Behaviour
It looks like the More variant of SplitButton shouldn't actually have a value. It's definitely not supposed to have the button text change when selecting a different option. Here's the contribution docs link for reference: https://spectrum-contributions.corp.adobe.com/page/split-button-beta/

### Actual Behaviour
If you click the default action and then another action, but button text / selection chagnes.

Easy to repro in the docs.

### LINKED ISSUE #1642: action-menu & split-button should have sync import options like picker
### Expected Behaviour
Following up on the resolution to #801 Clients should have the option to do a sync import of action-menu & split-button, so that the menus don't open with a delay because of the async overlay import.

### Actual Behaviour
This is only available in picker now.

Westbrook:
## Description
- correct delivery of `sp-split-button` variants
- surface sync imports for Action Menu and Split Button patterns

## Related Issue
fixes #1642
fixes #1456

## Motivation and Context
Correctness, options.

## How Has This Been Tested?
VRT

https://westbrook-split-button--spectrum-web-components.netlify.app/storybook/index.html?path=/story/split-button--more-cta

## Types of changes

## Checklist:

### DIFF
--- a/packages/split-button/sync/sp-split-button.ts
+++ b/packages/split-button/sync/sp-split-button.ts
@@ -10,5 +10,5 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import './index.js';
-import '../sp-picker.js';
+import '@spectrum-web-components/picker/sync/index.js';
+import '../sp-split-button.js';
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ executors:
 parameters:
     current_golden_images_hash:
         type: string
-        default: c36232311ca79de3ba648fc08333d021d45a68b7
+        default: 6035abba9a5721c7647b3772a8ac1e7745d3e783
 commands:
     downstream:
         steps:
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,6 +4,7 @@
         "packages/*/src/spectrum-vars.json": true,
         "**/*.js.map": true,
         "**/*.js": { "when": "$(basename).ts" },
-        "**/*.d.ts": { "when": "$(basename).ts" }
+        "**/*.d.ts": { "when": "$(basename).ts" },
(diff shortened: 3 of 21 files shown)


packages/split-button/sync/sp-split-button.ts
```diff
@@ -0,0 +1,28 @@
...
+import '../sp-split-button.js';
+
+Picker.openOverlay = async (
+    target: HTMLElement,
+    interaction: TriggerInteractions,
+    content: HTMLElement,
+    options: OverlayOptions
+): Promise<() => void> => {
+    return await Overlay.open(target, interaction, content, options);
+};
```
adixon-adobe, line 28: it might be worth DRYing up the `openOverlay` overrides into a Picker import.

adixon-adobe (dismissed):
One suggestion but overall looks great!

packages/split-button/sync/sp-split-button.ts
Westbrook: Agreed!

web-test-runner.config.js
```diff
@@ -96,21 +96,21 @@ export default {
...
+        //             'browser.shell.checkDefaultBrowser': false,
+        //             'browser.bookmarks.restore_default_bookmarks': false,
+        //             'dom.disable_open_during_load': false,
+        //             'dom.max_script_run_time': 0,
+        //             'dom.min_background_timeout_value': 10,
+        //             'extensions.autoDisableScopes': 0,
+        //             'extensions.enabledScopes': 15,
+        //         },
+        //     },
+        // }),
```
adixon-adobe, line 114: Guessing you didn't meant to check this change in

adixon-adobe (changes requested):
Looks like you had some local test config changes that accidentally got checked in.

Otherwise looks great

PR #1661, opened 2021-07-28
Complete Menu selects implementation

### LINKED ISSUE #350: Support an "sp-action-menu" w/o value

### LINKED ISSUE #715: Menu should manage its own state

### LINKED ISSUE #1189: Add more menu selection customizability

### LINKED ISSUE #1505: sp-menu-group is overkill
### Expected Behaviour
Our code and API structure is as simple as possible.

### Actual Behaviour
`sp-menu-group` really only serves the purpose of adding a header item above menu items. We should be able to move this pattern into `sp-menu` directly.

Because `sp-menu-group` is adding an `sp-menu` in it's Shadow DOM, it's also creating some timing complexities around managing menu items. 

### Proposal
Deprecate `sp-menu-group` and just add the `header` slot it supports directly to `sp-menu`.

Westbrook:
## Description
Allow `<sp-menu>` elements to manage their own selection via `selects="single|multiple|inherit"`. By default, an `<sp-menu>` element will take no selection, as it has previously. A `selects` value of `single` or `multiple` will make this element a selection root and manage the selection of `<sp-menu-item>` children and descendants of `[selects="inherit"]` children. When a Menu element has `selects="inherit"`, it's `<sp-menu-item>` descendants will participate in the selection of the nearest ancestor with `selects="single|multiple"`. Each Menu root that manages a selection will dispatch `change` events outlining updates to its value. In this way, multiple selection roots inside of a non-selecting `<sp-menu>` parent can manage these changes via event delegation, as [seen here](https://github.com/adobe/spectrum-web-components/blob/menu-selects/packages/menu/stories/menu-group.stories.ts#L85).

**Single** - [Demo](https://menu-selects--spectrum-web-components.netlify.app/storybook/?path=/story/menu--single-select)
https://github.com/adobe/spectrum-web-components/blob/a1a2ab37be661c064477192dbdd144d22735e9f8/packages/menu/stories/menu.stories.ts#L56-L64

**Multiple** - [Demo](https://menu-selects--spectrum-web-components.netlify.app/storybook/?path=/story/menu--multiple-select)
https://github.com/adobe/spectrum-web-components/blob/a1a2ab37be661c064477192dbdd144d22735e9f8/packages/menu/stories/menu.stories.ts#L82-L90

**Inherit** - [Demo](https://menu-selects--spectrum-web-components.netlify.app/storybook/?path=/story/menu-group--inherit)
https://github.com/adobe/spectrum-web-components/blob/a1a2ab37be661c064477192dbdd144d22735e9f8/packages/menu/stories/menu-group.stories.ts#L130-L149

**Complex** - [Demo](https://menu-selects--spectrum-web-components.netlify.app/storybook/?path=/story/menu-group--mixed)
https://github.com/adobe/spectrum-web-components/blob/a1a2ab37be661c064477192dbdd144d22735e9f8/packages/menu/stories/menu-group.stories.ts#L85-L112

- `<sp-menu>` now manages selection inside of `<sp-picker>` elements
- surfaces `selects` attribute on `<sp-action-menu>` so that it can choose whether to take a selection of not. _THIS IS A BREAKING CHANGE!_
- `<sp-picker>` will validate `selects` to `single` when set for all derivative elements, preventing the application of `multiple`. This should be added in follow up work.
- `<sp-menu-group>` is now an extension of `<sp-menu>` simplifying the code there
- corrects header correlation in `<sp-menu-group>` to ensure availability in the accessibility tree.
- `<sp-split-button>` no longer takes a selection in `type="more"`, correcting the visual delivery there
- flips Alex's initial implementation where each Menu had a data representation of all of it's Menu Items to a data model of each Menu have access to it's Menu Items so that there was only a single representation of their state allowing for the event cycle there not to require `new Event()`s in each Menu Item
- the above leverages events extended from `Event` rather than `CustomEvent`s
- [ ] Mentioning here (as well as below) that I need to add some documentation to various READMEs, but with so much functionality here, I wanted to get an official PR open for reviewers while I did so.

## Related Issue
fixes #350
fixes #715
fixes #1189
fixes #1505

## Motivation and Context
Menus need the ability to do much more than look pretty, and as the basis of a number of other UIs in the ecosystem they can reduce code duplication my having a central selection management implementation.

## How Has This Been Tested?
- lots of unit tests
- lots of VRTs
- lots of manual

## Types of changes

## Checklist:


najikahalsema (approved):
This is the most beautiful PR I've ever seen. Awesome job, and thanks for your hard work!

stanvladut (approved):
All the changes look great 💯

packages/menu/src/MenuItem.ts
```diff
@@ -23,6 +23,67 @@ import { ActionButton } from '@spectrum-web-components/action-button';
...
+        this._currentAncestorWithSelects = undefined;
+        item.menuData = {
+            focusRoot: undefined,
+            selectionRoot: undefined,
+        };
+    }
+}
+
+const addOrUpdateEvent = new MenuItemAddedOrUpdatedEvent();
+const removeEvent = new MenuItemRemovedEvent();
```
adixon-adobe, line 86: I'm assuming this was a little performance improvement?

packages/menu/src/MenuItem.ts
```diff
@@ -146,41 +189,37 @@ export class MenuItem extends ActionButton {
...
+    };
 }
 
 declare global {
     interface GlobalEventHandlersEventMap {
-        'sp-menu-item-added': CustomEvent<MenuItemUpdateEvent>;
-        'sp-menu-item-update': CustomEvent<MenuItemUpdateEvent>;
-        'sp-menu-item-removed': CustomEvent<MenuItemUpdateEvent>;
+        'sp-menu-item-added': MenuItemAddedOrUpdatedEvent;
+        'sp-menu-item-update': MenuItemAddedOrUpdatedEvent;
```
adixon-adobe, line 222: This isn't actually dispatched anymore. I also think you should change the added event to `sp-menu-item-added-or-updated` (or re-work things a bit so both can get dispatched separately).

packages/picker/src/Picker.ts
```diff
@@ -371,6 +367,11 @@ export class PickerBase extends SizedMixin(Focusable) {
         if (changes.has('selectedItem')) {
             this._selectedItemContent = undefined;
         }
+        if (this.selects) {
+            // Always force `selects` to "single" when set.
+            // TODO: Add support functionally and visually for "multiple"
```
adixon-adobe, line 372: For `Picker` itself we only want Single, right? It's just adding support to `PickerBase`? (it seems like it might be good to eventually break PickerBase out to a separate file and possibly rename)

adixon-adobe (changes requested):
I've got a bunch of little comments, but overall this looks really awesome. Love seeing those added selects tests, and it's nice to see the simplification you made to item management.

These changes are deviously complex and inter-related so nice freaking job on this!

Thank you!

packages/action-menu/src/ActionMenu.ts
```diff
@@ -33,6 +36,9 @@ export class ActionMenu extends ObserveSlotText(PickerBase, 'label') {
         return [actionMenuStyles];
     }
 
+    @property({ type: String })
+    public selects: undefined | 'single' = undefined;
```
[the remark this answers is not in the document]
Westbrook: I am excluding support for `multiple` here and in `Picker.ts` for the time being. There is no design direction on supporting multi-selection in this case and while I have some ideas, I didn't want to bind decisions in both places to allowing this PR to move forward. 

I've added https://github.com/adobe/spectrum-web-components/issues/1665 to track.

packages/menu/src/Menu.ts
```diff
@@ -20,28 +20,43 @@ import {
...
+
 /**
  * Spectrum Menu Component
  * @element sp-menu
  * @fires change - Announces that the `value` of the element has changed
- *
+ * @attr selects - whether the element has a specific selection algorithm that it applies
+ *   to its item descendants. `single` allows only one descendent to be selected at a time.
+ *   `multiple` allows many descendants to be selected. `inherit` will be applied dynamically
+ *   when an ancestor of this element is actively managing the selection of its descendents.
```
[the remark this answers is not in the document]
Westbrook: Action Menu demonstrates this: https://menu-selects--spectrum-web-components.netlify.app/storybook/index.html?path=/story/action-menu--default, and likely should have a `selects` demo, now, too.

The default Menu demo does as well: https://menu-selects--spectrum-web-components.netlify.app/storybook/index.html?path=/story/menu--default

I'll clarify this in the docs, too.

packages/menu/src/Menu.ts
```diff
@@ -59,30 +74,43 @@ export class Menu extends SpectrumElement {
...
+        return this.cachedChildItems;
+    }
+
+    private cachedChildItems: MenuItem[] | undefined;
 
-    private cachedChildItems: undefined | MenuChildItem[] = [];
+    private updateCachedMenuItems(): MenuItem[] {
+        this.cachedChildItems = [];
+
+        const slotElements = this.menuSlot.assignedElements({ flatten: true });
```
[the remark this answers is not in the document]
Westbrook: While I moved this around, this is _your_ logic... 😉 
```
    @query('slot:not([name])')
    public menuSlot!: HTMLSlotElement;
```
This decorator outlines that `this.menuSlot` will be _only_ the default `<slot>` element with not `name` attribute. `flatten: true` means content directly in that slot:
```
<sp-menu>
    <sp-menu-item></sp-menu-item>
</sp-menu>
```
And content slotted into that slot:
```
#shadow-root
    <sp-menu>
        <slot></slot>
    </sp-menu>
<sp-menu-item></sp-menu-item>
```
Will be acquired by this method.

The following code then accesses the various descendants of those elements. Again this is specifically for content order and is delayed as late in the process as possible due to the amount of extra DOM walking required here.

packages/menu/src/MenuItem.ts
Westbrook: Yeah, the `new`ing of a constructor (e.g. `new Event()`) can add up. Normally, we only do one or two at a time and it's not worth trying to optimize. However, in this case, we're causing ALL of the Menu Items to do this at once, so it seemed like something we could benefit from.

packages/picker/src/Picker.ts
Westbrook: Yeah, I've been thinking about whether `Picker` is the base or is `Action Menu` the base, and when we reconcile all that in the near future with `Combobox` I'll be sure to simplify/clarify this.

adixon-adobe (approved):

adixon-adobe:
Updates look great!

spdev3000:
Awesome PR! Thanks for the hard work here!

PR #1666, opened 2021-07-29
[WIP] DNA@7 Preview

Westbrook:


packages/action-group/src/action-group.css
```diff
@@ -39,63 +39,69 @@ governing permissions and limitations under the License.
 
 :host([compact]:not([quiet])) ::slotted(:not([role])) {
     /* .spectrum-ActionGroup--compact:not(.spectrum-ActionGroup--quiet) .spectrum-ActionGroup-item */
-    --spectrum-actionbutton-m-border-radius: 0;
-    --spectrum-actionbutton-s-border-radius: var(
-        --spectrum-actionbutton-m-border-radius
+    --overriden-border-radius: 0;
```
castastrophe, line 42: @Westbrook & I chatted about this in a call but I'm wondering if we save a little size by converting this to a Sass variable instead:

```suggestion
    $overriden-border-radius: 0;
```

This allows us to retain the clarity of calling out this override in one place and showing it's use in the variables below while removing the variable interpolation from the rendered output.

castastrophe: A small sizing and minute performance gain but each of these can add up over time.

Westbrook: Was stepping back into this with the full release at the CSS level and this runtime contextually changes starting in line 71, so I'm not sure this can be moved into SASS without end use bloat. Thoughts?

Westbrook:
The changes accepted in the VRT roughly map out to the following:

- Action Button
  - nuanced spacing changes
- Button
  - nuanced spacing changes
  - icons vertically centered
- Card
  - nuanced spacing changes
  - nuanced font size changes
  - default widths updated
- Checkbox
  - a11y color updates
- Field Label
  - nuanced spacing changes
- Meter
  - nuanced spacing/color changes
  - fixes `warning` => `notice`
  - fixes `critical` => `negative`
- Progress Bar
  - nuanced spacing/color changes
- Progress Circle
  - nuanced color changes
- Radio
  - nuanced color change
- Search
  - nuanced color change
- Slider
  - nuanced spacing changes
  - fixes padding/margin when `[label-visibility='none']`
- Split Button
  - same as Button
- Tabs
  - nuanced spacing changes
- Top Nav
  - nuanced spacing changes

PR #1668, opened 2021-07-29
Menu Selects landing branch to Main

### LINKED ISSUE #350: Support an "sp-action-menu" w/o value

### LINKED ISSUE #715: Menu should manage its own state

### LINKED ISSUE #1189: Add more menu selection customizability

### LINKED ISSUE #1505: sp-menu-group is overkill

Westbrook:
This merges @adixon-adobe previous work and my recent work on Menu selection into `main`.

I've added this one commit: https://github.com/adobe/spectrum-web-components/commit/a3120abb5b1e48c025de81bde52b36f4d3c47ed5 to document the new usage of `sp-action-menu` and to ensure that the package is given a "breaking" (in pre-1.0 land _feature_) version change.

See the previous PRs for reference:
- #1661
- #1541

Issues referenced: 
fixes #350
fixes #715
fixes #1189
fixes #1505


adixon-adobe (approved):
Awesome -- very exciting!
```ascii
  _______
 |       |
(|WINNING|)
 |  #1   |
  \     /
   `---'
   _|_|_
```

PR #1670, opened 2021-07-30
fix: add "editable" option to "sp-slider"

### LINKED ISSUE #1572: [Slider] Add editable input for Slider component
### Expected Behaviour
We would like to have an editable input for the current `Slider` implementation that would not only reflect the current `Slider's` value, but would also allow us to edit the current value.

### Actual Behaviour
The current `Slider` implementation only allows changing the value via dragging the bullet on in horizontal axis.

Screenshot 2021-06-23 at 17 04 51

Westbrook:
## Description
Add editable slider delivery binding an `<sp-number-field>` into the UI of the `<sp-slider editable>` element through dynamic importing the element registration.
- also adds a synchronous entry point into the functionality as desired.

## Related Issue
fixes #1572 

## Motivation and Context
Spectrum compliance and application flexibility.

## How Has This Been Tested?
- VRT
- unit tests.

## Types of changes

## Checklist:

### DIFF
--- a/packages/slider/src/Slider.ts
+++ b/packages/slider/src/Slider.ts
@@ -25,6 +25,8 @@ import {
 import sliderStyles from './slider.css.js';
 import { ObserveSlotText } from '@spectrum-web-components/shared/src/observe-slot-text.js';
 import { StyleInfo } from 'lit-html/directives/style-map';
+import '@spectrum-web-components/field-label/sp-field-label.js';
+import type { NumberField } from '@spectrum-web-components/number-field';
 import { HandleController, HandleValueDictionary } from './HandleController.js';
 import { SliderHandle } from './SliderHandle.js';
 
@@ -36,6 +38,36 @@ export class Slider extends ObserveSlotText(SliderHandle, '') {
     }
 
     public handleController: HandleController = new HandleController(this);
+
+    /**
+     * Whether to display a Number Field along side the slider UI
+     */
+    @property({ type: Boolean, reflect: true })
+    public get editable(): boolean {
+        return this._editable;
+    }
+
+    public set editable(editable: boolean) {
+        if (editable === this.editable) return;
+        const oldValue = this.editable;
+        this._editable = this.handleController.size < 2 ? editable : false;
+        if (this.editable) {
+            this._numberFieldInput = import(
+                '@spectrum-web-components/number-field/sp-number-field.js'
+            );
+        }
+        if (oldValue !== this.editable) {
+            this.requestUpdate('editable', oldValue);
+        }
+    }
+
+    private _editable = false;
+
+    /**
+     * Whether the stepper UI of the Number Field is hidden or not
+     */
+    @property({ type: Boolean, reflect: true, attribute: 'hide-stepper' })
+    public hideStepper = false;
 
     @property()
     public type = '';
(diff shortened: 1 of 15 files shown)


stanvladut (approved):
Looks nice, great work on this!

packages/slider/src/Slider.ts
```diff
@@ -273,6 +328,22 @@ export class Slider extends ObserveSlotText(SliderHandle, '') {
...
+        const { value } = event.target;
+        if (!isNaN(value)) {
+            this.value = value;
+        }
+    }
+
+    private handleNumberChange(event: Event & { target: NumberField }): void {
+        const { value } = event.target;
+        if (isNaN(value)) {
+            this.value = 0;
```
stanvladut, line 341: Would it be a better idea to fallback on the current value instead of 0?

stanvladut (approved):

packages/slider/src/Slider.ts
Westbrook: Good question. If we do take this path, it should also take into account `this.min`. However, the idea here is the `<sp-number-field>` returns a `NaN` value when its contents have been deleted and then the empty value committed (either `Enter` key is pressed or the field is blurred). I was thinking of this as an explicit decision by the user to "reset" the value. We don't currently have a _default_ value, so I chose to default it to "nothing" or `this.min`. I can see this going both ways, but feel like the active decision to _delete_ the value is something we should respect. I'll surface this to the steering committed to get some other thoughts.

Westbrook: Based off of discussion across the team, particularly with design, to align with other offerings the new experience here will be as follows:
- `input` is no longer tracked
  - only "committed" values will be sent to the slider
  - `input` event from the Number Field will have their propagation stopped
- `change` will be tracked and _if_ the change is to `NaN` the previous "committed" value will be reapplied
- when a _valid_ `change` event is handled, it will dispatch a matching `input` event for consistency

adixon-adobe:
One thing I'm seeing is that if I click & hold the stepper up or down in the `Slider->Editable` storybook, the slider isn't updating until you let go. I'd expect the slider to update just like the input is updating, but maybe not as a committed change (so e.g. if you hit `esc` you could go back to the previous value). I know in Lightroom we have both a "tracking" value for when you're moving the slider and a final/committed value when you let go.

I won't have time to fully review this soon, but figured I'd pass on some basic feedback/thoughts.

Westbrook:
@adixon-adobe good call on the following the stepper more closely! I've added that as it balances the overall adjustments to the `input`/`change` event listening happening here. I'd say that further customization isn't blocking here, but can be convinced that's not true.

With @stanvladut leveraging similar, but maybe not exactly the same, functionality local to his work, maybe he has thoughts to share on whether moving forward now and possibly having small changes later is an issue?

Thanks everyone for helping us work through this important addition to the library!

stanvladut:
I think we can move on with it like this and update it in the future. We haven not yet implemented such functionality, so I don't see this as a blocker, but more as a new feature that can be improved in the future based on the feedback.

packages/slider/src/Slider.ts
```diff
@@ -37,6 +39,36 @@ export class Slider extends ObserveSlotText(SliderHandle, '') {
...
+    public get editable(): boolean {
+        return this._editable;
+    }
+
+    public set editable(editable: boolean) {
+        if (editable === this.editable) return;
+        const oldValue = this.editable;
+        this._editable = this.handleController.size < 2 ? editable : false;
+        if (this.editable) {
+            this._numberFieldInput = import(
```
stanvladut, line 55: 💯

stanvladut (approved):

PR #1675, opened 2021-08-03
refactor: use simpler ButtonBase

Westbrook:
Do less work, go faster!


Westbrook:
Closing in favor of #1686 where there is more direct benefit _today_ without the need to refactor all the way up the Spectrum tool chain.

PR #1697, opened 2021-08-11
refactor: use CSS over SVG for wheel UI delivery

Westbrook:
## Description
Investigate perf benefits of CSS based wheel UI instead of SVG based UI.

Previously, this was not leveraged due to browser support of either https://caniuse.com/?search=conic-gradient or https://caniuse.com/?search=clip-path. Both of these appear securely within our support matrix at this time, and the perf benefits outlined below are quite exciting!

## Motivation and Context
Things should be fast!

## How Has This Been Tested?
- VRTs updated
- show perf benefits outlined below

## Types of changes
- [x] refactor

## Checklist:


.circleci/config.yml
```diff
@@ -8,7 +8,7 @@ executors:
 parameters:
     current_golden_images_hash:
         type: string
-        default: aa8a18f56f8af4fe5526f060006ad0c810699c73
+        default: b268071af940937c61fd0b5e458a69f825c68a48
```
castastrophe, line 11: Interesting - can you tell me about this hash value?

Westbrook: The goal is for this to be explained here: https://github.com/adobe/spectrum-web-components#screenshot-testing and we can talk about it some more in person later. You and I have also talked separately about alternative approaches to this concept that I'd be very happy to see explored in due time.

packages/color-wheel/src/color-wheel.css
```diff
@@ -19,3 +19,62 @@ governing permissions and limitations under the License.
...
+    /* .spectrum-ColorWheel.is-disabled .spectrum-ColorWheel-innerCircle,
+   * .spectrum-ColorWheel.is-disabled .spectrum-ColorWheel-outerCircle */
+    border-color: var(
+        --spectrum-colorwheel-border-color-disabled,
+        var(--spectrum-global-color-gray-300)
+    );
+}
+:host([disabled]) .wheel {
+    /* .spectrum-ColorWheel.is-disabled .spectrum-ColorWheel-outerCircle */
+    background: var(
```
castastrophe, line 76:
```suggestion
    background-color: var(
```

Would it make sense to include more specificity here in the property name to prevent shorthand input?  Or if someone were to add an image to `--spectrum-colorwheel-fill-color-disabled` an okay use of the hook?

packages/color-wheel/src/ColorWheel.ts
```diff
@@ -23,7 +23,6 @@ import { WithSWCResizeObserver, SWCResizeObserverEntry } from './types';
 import { Focusable } from '@spectrum-web-components/shared/src/focusable.js';
 import '@spectrum-web-components/color-handle/sp-color-handle.js';
 import styles from './color-wheel.css.js';
-import { wheel } from './wheel-svg.js';
```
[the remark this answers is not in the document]
Westbrook: Hedged originally, but the export map was protecting users from accidentally importing that, so it can be removed. Nice!

packages/color-wheel/src/color-wheel.css
Westbrook: I can't put a `conic-gradient` into `background-color`, can I? This value needs to override the `conic-gradient` that applied the hue visuals by default.

Also, great question about `--spectrum-colorwheel-fill-color-disabled` being given an image. If it was given an image, regardless of its use in `background` or `background-color`, the results would be somewhat unexpected. Not sure there's safety there outside of a future where we have typed custom properties.

packages/color-wheel/src/color-wheel.css
```diff
@@ -19,3 +19,62 @@ governing permissions and limitations under the License.
...
+    inset: 0;
+}
+
+.wheel:before {
+    inset: 24px;
+}
+
+:host([disabled]) .wheel:before,
+:host([disabled]) .wheel:after {
+    /* .spectrum-ColorWheel.is-disabled .spectrum-ColorWheel-innerCircle,
```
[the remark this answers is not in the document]
Westbrook: Yes. Generally this means I copied something out of the `spectrum-{component-name}.css` file that either needed rules that Spectrum CSS didn't ship, or needed to be converted to a selector I couldn't figure out how to process with our PostCSS plugin.

In this case, we added new rules with the originals being:
https://github.com/adobe/spectrum-web-components/blob/fadb7a1c80e8090c29b590111e136efa91b471aa/packages/color-wheel/src/spectrum-color-wheel.css#L115-L134

packages/color-wheel/src/color-wheel.css
castastrophe: Ah good point, that function is creating an image. `background-image` then maybe? https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/conic-gradient()#gradient_at_40-degrees

Though I see a lot of the examples just use the shorthand so I'm sure it's not a problem.

Westbrook: Cool. I'll leave as `background` for now but these are good lines of inquiry to make time for with each change we make to the library!

castastrophe (approved):
Loki Giving Thor Mischief (LGTM!) 🙃

PR #1709, opened 2021-08-17
[RFC] feat: allow native :dir(...) selector usage where availables

Westbrook:
Testing perf updates in CI for Firefox where the native`:dir()` selector is currently correctly enough implemented to leverage.

EDITOR'S NOTE: While perf testing is generally done against Chrome, this branch specifically looks at performance in Firefox as it is the only browser that has an implementation in this area that we can rely on.


Westbrook:
Core Token's releases in Spectrum CSS is slowly moving to CSS Logical Properties. Once that process is done, we can discuss removing this workaround all together!!

PR #1716, opened 2021-08-17
chore: use latest globby version

Westbrook:
fixes #1712


Westbrook:
GitHub actions and CJS don't like this. Will revisit later.

PR #1718, opened 2021-08-18
Propogate open to child sp-popover 

### LINKED ISSUE #1417: [Action Bar] a11y review
### Feedback

- Action bar is in the DOM (not hidden) even when not visible. This will cause screen readers to read it even when not displayed visually

### Path forward

1. Interest, most of the other elements use `display: none` in this case, which prevents this (right?). We can sprout `aria-hidden="true"` when `open === false` and we should be set here.

cc @jnurthen

jnurthen:
fixes #1417 

## Description

propogates open attribute from sp-action-bar to the child sp-popover

## Related Issue

#1417

## Motivation and Context

the sp-popover was always rendered as open which made it visible to the accessibility tree as it was being rendered with visibility:visible which will expose content in the tree even if a child of visibility:hidden

## How Has This Been Tested?

Running in the docs and ensuring that the content is only rendered in the a11y tree when the action-bar is open

## Types of changes

## Checklist:

### DIFF
--- a/packages/action-bar/src/ActionBar.ts
+++ b/packages/action-bar/src/ActionBar.ts
@@ -68,7 +68,7 @@ export class ActionBar extends SpectrumElement {
 
     public render(): TemplateResult {
         return html`
-            <sp-popover open id="popover">
+            <sp-popover ?open=${this.open} id="popover">
                 <slot></slot>
             </sp-popover>
         `;


Westbrook (approved):
Love how this is even simpler than I thought it would be. Thanks @jnurthen!

Sorry to leave these smaller issues are Open Source bait, but we've got a couple of new team members who are just about ramped up to a place to start picking things like this off!

PR #1728, opened 2021-08-24
chore: remove duplicates, comment groupings in .gitignore

castastrophe:
## Description

- Removed duplicate `custom-elements.json` from .gitignore
- Added grouping comments
- Q @Westbrook: Is it possible to commit the .npmrc so that users don't have to do that step during the quick start? 

## Motivation and Context

I was poking around to better understand the tooling 😄 

## Types of changes

- [x] Chore

## Checklist:


castastrophe:
@Westbrook I'm not sure if all my comments are accurate or make sense - would super appreciate feedback on if I got those right!

.gitignore
```diff
@@ -1,38 +1,34 @@
...
+# documentation
 documentation/components/searchIndex.json
+documentation/api-docs
+documentation/dist
 
+# dependency folders
 coverage
+lib
+
+# compiled / pulled-down assets
```
Westbrook, line 22: What did you have in mind with "pulled-down" here? These are all various build artifacts, IIRC.

.gitignore
```diff
@@ -42,26 +38,37 @@ packages/*/*/*.d.ts
 packages/**/*.test-vrt.ts
 *.tsbuildinfo
 
+# ts + css assets
+src/**/*.css.ts
```
Westbrook, line 42: This is repeated in line 24.

.gitignore
```diff
@@ -42,26 +38,37 @@ packages/*/*/*.d.ts
...
+packages/**/*.css.ts
+styles/**/*.css.ts
+
+# compiled assets
```
Westbrook, line 46: Should we merge this with line 31?

.gitignore
```diff
@@ -42,26 +38,37 @@ packages/*/*/*.d.ts
...
 projects/**/*.js
 projects/**/*.js.map
 projects/**/*.d.ts
 !projects/*/global.d.ts
 
+# build logs
```
Westbrook, line 52: I wonder what outputs these two... 🤔

.npmrc
```diff
@@ -0,0 +1,2 @@
+@adobe:registry=https://registry.npmjs.org/
```
Westbrook, line 1: What do we get from adding these?

Westbrook (commented):
Good call in looking at this! Definitely has been a lot of churn in here over time.

.npmrc
Westbrook: Now I see what this is related to. That's been in the README.md forever, but I'm not sure I've ever done it... Listing here makes sense, and in doing so does that means we can remove it from the README?

.gitignore
castastrophe: Ah good, that's what I wasn't sure about - if custom-elements.json for example was a build artifact or something pulled-down externally

castastrophe:
```suggestion
# build artifacts
```

.gitignore
castastrophe: is `src/**/*.css.ts` the same as `packages/*/src/**/*.css.js`?

.gitignore
castastrophe: Do you think they might be outdated?  I could remove them and we can always add back if they reappear.

.gitignore
Westbrook: Either it _is_ the same or it points to something that no longer exists... I think there was an `src` directory in the root at some point 🤔. If the shorter version covers, it may be nice that is applies to both `packages` AND `projects` but that auto coverage might be unexpected. Thoughts?

.gitignore
Westbrook: Sounds good to me!

.gitignore
castastrophe: I don't believe it will, it would need the `*/src` for it to delve into parent folders.  This would exclusively apply to a src directory in the root.

Westbrook: Then remove we shall, as in that case it references no files!

PR #1750, opened 2021-08-27
ci: block PRs on out-of-date

Westbrook:
## Description

Test to see if we could just remove the "Update Branch" button, but still get warned when we're not up-to-date via CI....

## Motivation and context
An alternative, or partner, to #1729 

## How has this been tested?

-   [ ] _Test case 1_
    1. Open PR.
    2. This PR currently fails CI for being out-of-date.
-   [ ] _Test case 2_
    1. Rebase this branch
    2. This PR passes for being up-to-date

## Types of changes
-   [x] CI

## Checklist

## Best practices

### DIFF
--- a/.github/workflows/up-to-date.yml
+++ b/.github/workflows/up-to-date.yml
@@ -1,0 +1,36 @@
+name: Banch up-to-date with target
+
+on:
+    pull_request:
+        types: [opened, synchronize, edited, reopened]
+
+jobs:
+    compare:
+        name: Confirm that the branch is up-to-date with its target
+
+        # We can't currently run benchmarks on PRs from forked repos, because the
+        # tachometer action reports results by posting a comment, and we can't post
+        # comments without a github token.
+        if: github.event.pull_request.head.repo.full_name == github.repository
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v2
+              with:
+                  fetch-depth: 0
+
+            - name: Setup Node 14
+              uses: actions/setup-node@v2
+              with:
+                  node-version: '14'
+                  cache: 'yarn'
+                  registry-url: 'https://registry.npmjs.org'
+
+            - name: Install dependencies
+              run: yarn --frozen-lockfile --ignore-scripts
+
+            - name: Find if target has been merged
+              run: |
+                  git checkout $GITHUB_BASE_REF
+                  git checkout $GITHUB_HEAD_REF
+                  merged=$(git branch --merged)
+                  yarn is-merged --branch $GITHUB_BASE_REF --merged "$merged"
--- a/scripts/is-merged.js
+++ b/scripts/is-merged.js
@@ -1,0 +1,35 @@
+#!/usr/bin/env node
+
+/*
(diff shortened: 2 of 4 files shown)


.github/workflows/up-to-date.yml
```diff
@@ -0,0 +1,25 @@
+name: Banch up-to-date with target
+
+on: pull_request
```
castastrophe, line 3: Which pull_request events do you want this to fire on? `synchronize` maybe?

.github/workflows/up-to-date.yml
```diff
@@ -0,0 +1,25 @@
...
+
+jobs:
+    compare:
+        name: Confirm that the branch is up-to-date with its target
+
+        # We can't currently run benchmarks on PRs from forked repos, because the
+        # tachometer action reports results by posting a comment, and we can't post
+        # comments without a github token.
+        if: github.event.pull_request == null || github.event.pull_request.head.repo.full_name == github.repository
```
castastrophe, line 12: Since you're triggering this workflow from the pull_request event, the value of event.pull_request should never be null.  Here's the object you should have access to on this event: https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#pull_request

.github/workflows/up-to-date.yml
Westbrook: Interesting. I had no idea `pull_request` resolved to so many different events. I'll look at this closer (and take a look at how this might be applied to other actions), thanks!

.github/workflows/up-to-date.yml
Westbrook: Nice! Lots of copy/pasta in the actions code to date. Anytime we can get rid of things is win 🎊

scripts/is-merged.js
```diff
@@ -0,0 +1,35 @@
...
+    let mergedBranched = merged;
+    if (!merged) {
+        const command = execSync('git branch --merged');
+        mergedBranched = command.toString();
+    }
+    const isMerged = mergedBranched.search(branch) > -1;
+    if (isMerged) {
+        console.warn('This branch is up to date.');
+    } else {
+        console.warn('This branch needs to be rebased.');
```
castastrophe, line 30: If you like, you could try a rebase here too to see if it would rebase cleanly.  Just abort if it requires manual intervention.

Westbrook: That could be interesting. Auto history editing could be surprising, on top of everything else, but if it's actually an option, knowing the full possibilities here would allow us to review and construct the right workflow together going forward!

Where I'm struggling more directly with this is in finding a way to trigger this workflow when there are changes on the target branch. Without getting auto run in that context I'm not sure that any work we do here is really _enough_ for it to replace the "This branch is out-of-date with the base branch" functionality that this mimics in hope of preventing the merge commits that come along with it.

castastrophe: This Action can auto-rebase: https://github.com/marketplace/actions/auto-rebase

Westbrook:
Not sure I can get this to work how I would like. Closing for now.

Issue #1770, opened 2021-09-04
[Bug]: picker is not compatible with Safari's tab navigation


hunterloftis:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Picker

### Expected behavior

Tab and shift+tab keypresses focus Picker elements on the page in tab order across all browsers.

### Actual behavior

Safari skips Picker elements during tab navigation (and the relevant test fails locally).

### Screenshots

![image](https://user-images.githubusercontent.com/364501/132083778-530802ef-0302-4457-bd03-7f968fda9355.png)

### What browsers are you seeing the problem in?

Safari

### How can we reproduce this issue?

1. run `yarn test`

PR #1771, opened 2021-09-04
fix: give Picker a focus helper to enable tab navigation in Safari

hunterloftis:
## Description

Safari skips `button` elements in the tab order, so we need to provide a synthetic target with a tabindex to receive and then throw focus onto the Picker's button.

## Related issue(s)

Resolves https://github.com/adobe/spectrum-web-components/issues/1770

## Motivation and context

Fixes a bug.

## How has this been tested?

`yarn test` and manual testing in Chrome & Safari.

## Types of changes

-   [X ] Bug fix (non-breaking change which fixes an issue)

## Checklist

-   [ X] I have signed the [Adobe Open Source CLA](http://opensource.adobe.com/cla.html).
-   [ X] My code follows the code style of this project.
-   [ X] If my change required a change to the documentation, I have updated the documentation in this pull request.
-   [ X] I have read the **[CONTRIBUTING](<(https://github.com/adobe/spectrum-web-components/blob/main/CONTRIBUTING.md)>)** document.
-   [ X] I have added tests to cover my changes.
-   [ X] All new and existing tests passed.

## Best practices

### DIFF
--- a/packages/picker/src/Picker.ts
+++ b/packages/picker/src/Picker.ts
@@ -160,6 +160,12 @@ export class PickerBase extends SizedMixin(Focusable) {
         this.toggle();
     }
 
+    public onHelperFocus(): void {
+        // set focused to true here instead of onButtonFocus so clicks don't flash a focus outline
+        this.focused = true;
+        this.button.focus();
+    }
+
     public onButtonFocus(): void {
         (this.target as HTMLButtonElement).addEventListener(
             'keydown',
@@ -348,8 +354,15 @@ export class PickerBase extends SizedMixin(Focusable) {
         ];
     }
 
+    // a helper to throw focus to the button is needed because Safari
+    // won't include buttons in the tab order even with tabindex="0"
     protected get renderButton(): TemplateResult {
         return html`
+            <span
+                id="focus-helper"
+                tabindex="${this.focused ? '-1' : '0'}"
+                @focus=${this.onHelperFocus}
+            ></span>
             <button
                 aria-haspopup="true"
                 aria-expanded=${this.open ? 'true' : 'false'}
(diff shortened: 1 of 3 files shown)


hunterloftis:
I think this test failure is related to https://github.com/adobe/spectrum-web-components/issues/1772 rather than this PR.

Westbrook:
This looks good so far. Could you expand the test found [here](https://github.com/adobe/spectrum-web-components/blob/main/packages/picker/test/picker-sync.test.ts#L816) and [here](https://github.com/adobe/spectrum-web-components/blob/main/packages/picker/test/picker-sync.test.ts#L816) to cover Tabing forward through the interface as well as backward through the interface as seen in https://github.com/adobe/spectrum-web-components/blob/main/packages/color-area/test/color-area.test.ts#L33-L92, please. We should be able to remove those comments in the test files while you're at it.

In particular, I'm looking to make sure that the CI version of Safari don't accidentally stop on the `focus-helper` while `tabindex=-1` like I saw it do with the Color Area tests.

hunterloftis:
Sure! Could you double-check those test links? They both open the same line at the same test for me.

Westbrook:
Updated. They're the same test in the async and sync version of the Picker.

hunterloftis:
@Westbrook I updated those tests to go through, back, in, out, up. They pass! Also, rebased.

packages/picker/src/Picker.ts
```diff
@@ -160,6 +160,15 @@ export class PickerBase extends SizedMixin(Focusable) {
         this.toggle();
     }
 
+    public onHelperFocus(): void {
+        // set focused to true here instead of onButtonFocus so clicks don't flash a focus outline
+        this.focused = true;
+        const button = this.shadowRoot.querySelector(
```
Westbrook, line 166: This should already be available as `this.button`. Swap this out and we'll be good to go!

hunterloftis: Good catch!

hunterloftis: ✅

Westbrook (approved):
Awesome!

Issue #1772, opened 2021-09-04
[Bug]: Color Area "keypress" test is flappy


hunterloftis:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Color Area

### Expected behavior

Tests are deterministic

### Actual behavior

The check `expect(el.x, 'x').to.equal(0.67);` is sometimes off by 0.01, leading to flappy test failures.

### How can we reproduce this issue?

Run `yarn test` repeatedly and notice that sometimes all tests pass, and sometimes this test fails. For example:

https://app.circleci.com/pipelines/github/adobe/spectrum-web-components/9063/workflows/b5076bff-7557-4d76-b914-fdb3cbe38a02/jobs/111639

Westbrook:
fixed by #1773

PR #1773, opened 2021-09-06
fix: flappy Slider/Color Area tests

hunterloftis:
## Description

- replace `elementUpdate` with `nextFrame`, which seems more reliable, in intermittently failing tests
- fix visibility of `ariaValueText` in Slider to be compatible with its type signature

## Related issue(s)

- https://github.com/adobe/spectrum-web-components/issues/1772

## Motivation and context

While developing https://github.com/adobe/spectrum-web-components/pull/1771, I noticed that these tests would intermittently fail. The failures were rare and non-deterministic, which makes it difficult to know whether or not your work created a regression.

## How has this been tested?

Experimentation (by running `yarn test` repeatedly) suggests that this PR will more-reliably `await` the state changes that the tests are confirming.

## Types of changes

## Checklist

-   [ X] My code follows the code style of this project.
-   [ X] If my change required a change to the documentation, I have updated the documentation in this pull request.
-   [ X] I have read the **[CONTRIBUTING](<(https://github.com/adobe/spectrum-web-components/blob/main/CONTRIBUTING.md)>)** document.
-   [ X] I have added tests to cover my changes.
-   [ X] All new and existing tests passed.

## Best practices

### DIFF
--- a/packages/slider/test/slider.test.ts
+++ b/packages/slider/test/slider.test.ts
@@ -14,12 +14,30 @@ import '../sp-slider.js';
 import '../sp-slider-handle.js';
 import { Slider, SliderHandle } from '../';
 import { tick } from '../stories/slider.stories.js';
-import { fixture, elementUpdated, html, expect } from '@open-wc/testing';
+import {
+    fixture,
+    elementUpdated,
+    html,
+    expect,
+    oneEvent,
+    nextFrame,
+} from '@open-wc/testing';
 import { sendKeys, executeServerCommand } from '@web/test-runner-commands';
-import { spy } from 'sinon';
 import { ProvideLang } from '@spectrum-web-components/theme';
 
 describe('Slider', () => {
+    // At least one browser (Webkit) maintains shared global state for the mouse,
+    // such that `send-mouse` type: 'down' without subsequently sending 'up'
+    // can prevent a subsequent test from sending 'down' correctly.
+    afterEach(async () => {
+        await executeServerCommand('send-mouse', {
+            steps: [
+                {
+                    type: 'up',
+                },
+            ],
+        });
+    });
     it('loads', async () => {
         const el = await fixture<Slider>(
             html`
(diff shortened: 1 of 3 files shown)


Westbrook:
I'm a little uncomfortable with needing to move to `nextFrame` here, however due to the fact that we're leveraging the `executeCommand` API to communicate back and forth between the client side tests and the server side browser runner, I could be convinced that this is OK.

Context: `await elementUpdated(el);` is an abstraction on `await el.updateComplete;` which is what a consuming developer would normally wait for in this context. If that timing isn't correct, we should confirm whether there is additional work inside of the element to ensure that this workflow is stable for users that need to leverage it in their code. An alternative could be to leverage the `oneEvent` helper, to wait for expected events in the user workflow before we test the current realities fo the code. This may also help us to fully confirm that the validation is happening synchronous to the events dispatched as expected.

hunterloftis:
I think that's a reasonable concern. The reason I suspect `nextFrame` may be ok is that clues indicate that the failures are due to our testing framework rather than the underlying library. _Most_ of the time, the tests pass with `elementUpdated`, suggesting a race condition in transferring updated state between the browser and the testing infrastructure.

I'll change this to check state after the events are received instead; it'd be good to confirm that the updated state is available synchronously.

hunterloftis:
@Westbrook this is now both event-based and rebased!

packages/slider/test/slider.test.ts
```diff
@@ -14,12 +14,30 @@ import '../sp-slider.js';
...
+    nextFrame,
+} from '@open-wc/testing';
 import { sendKeys, executeServerCommand } from '@web/test-runner-commands';
-import { spy } from 'sinon';
 import { ProvideLang } from '@spectrum-web-components/theme';
 
 describe('Slider', () => {
+    // At least one browser (Webkit) maintains shared global state for the mouse,
+    // such that `send-mouse` type: 'down' without subsequently sending 'up'
+    // can prevent a subsequent test from sending 'down' correctly.
```
Westbrook, line 31: 👏🏼 👏🏼 👏🏼 👏🏼

Westbrook (approved):
lgtm

PR #1785, opened 2021-09-10
Support "px" and other non-native units in Number Field and Slider

### LINKED ISSUE #1715: NumberField: Support of any generic unit string
### Expected Behaviour
We'd like to provide a generic unit string (such as `px`) to `sp-number-field` in order to let it be appended to the current value (and not parsed by and NumberFormatter added).

For example: To achieve a Pixel NumberField with display value `1234 px` we'd like to use something like
```
<sp-number-field custom-unit="px" value="1234"></sp-number-field> 
```

### Actual Behaviour
Current NumberFormatter doesn't provide any `px`-related unit. 

#### Platform and Version
MacOs, Chrome v92 

Unfortunately this doesn't work:
```
.formatOptions=${{ useGrouping: false, unit: 'graphics-pixel', style: 'unit', unitDisplay: 'narrow' }}
```
Also - as we need to provide a localized version of `px` we'd like to see such a generic/polyfill solution.

Westbrook:
## Description
- add support for non-native `unit` values in `formatOptions` object supplied to Number Field and Slider elements
- test it
- demo it
- document it

## Related issue(s)
- fixes #1715 

## Motivation and context
We work with a lot of graphics software that needs extended unit support.

## How has this been tested?

-   [ ] _Test case 1_
    1. Visit https://support-px--spectrum-web-components.netlify.app/storybook/?path=/story/slider--px, https://support-px--spectrum-web-components.netlify.app/storybook/?path=/story/slider--two-handles-pt&args=variant:range;tickStep:10;labelVisibility:!undefined, or https://support-px--spectrum-web-components.netlify.app/storybook/?path=/story/slider--three-handles-pc
    2. Move around the handles
    3. Ensure that the `pc`, `pt`, or `px` string is included appropriately in the value labeling
-   [ ] _Test case 2_
    1. Visit https://support-px--spectrum-web-components.netlify.app/storybook/?path=/story/number-field--pixels&args=disabled:false;readonly:false;quiet:false;value:800;placeholder:;min:!undefined;max:!undefined;step:!undefined
    2. Edit the value managed by the Number Field
    3. Ensure that while focused the Number Field does not support a string based unit value
    4. Blur the field
    5. Ensure that the unit string is reapplied
    6. Repeat to taste across the other Number Field demos

## Types of changes

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ executors:
 parameters:
     current_golden_images_hash:
         type: string
-        default: f23812837fc051e2770bf04b7b938a6a0927c98d
+        default: 63bd3cbbc7c85c252bfd6353605c47ad9abae1a8
 commands:
     downstream:
         steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ executors:
 parameters:
     current_golden_images_hash:
         type: string
-        default: 4dd1cb3933d1d5b8b542916d0b8e775a1d5a397e
+        default: f23812837fc051e2770bf04b7b938a6a0927c98d
 commands:
     downstream:
         steps:
--- a/packages/number-field/README.md
+++ b/packages/number-field/README.md
@@ -112,6 +112,25 @@ Note: The unit style is not currently supported in Safari. A [polyfill](https://
 ></sp-number-field>
 ```
 
+### Units not included in `Intl.NumberFormatOptions`
+
+While `Intl.NumberFormatOptions` does support a [wide range of units](https://tc39.es/proposal-unified-intl-numberformat/section6/locales-currencies-tz_proposed_out.html#sec-issanctionedsimpleunitidentifier), it is possible to encounter units (e.g. the graphics units of `pixel`, `pixels`, `points`, etc.) that are not supported therein. When this occurs, an `<sp-number-field>` element will attempt to polyfill support for this unit. See the following example delivering `{ style: "unit", unit: "px" }` below:
+
+```html
+<sp-field-label for="units">Document width in pixels</sp-field-label>
+<sp-number-field
+    id="units"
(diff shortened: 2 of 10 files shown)


najikahalsema (approved):
This looks great! Just curious--why can't we have the "px" when focused on the NumField?

spdev3000 (approved):
LGTM, thx Westbrook

Westbrook:
@najikahalsema great question! I discovered on a closer read of the Spectrum documentation, that units should not show while editing (at all), and shared a screen grab of the internal docs, here: https://github.com/adobe/spectrum-web-components/issues/1715#issuecomment-916993384

This is applied via `formatOptions` objects that use the `style: "unit"` entry, so you'll see this in the "Units" demo that asks for "inches", here: https://support-px--spectrum-web-components.netlify.app/storybook/index.html?path=/story/number-field--units&args=disabled:false;readonly:false;quiet:false;value:24;placeholder:;min:!undefined;max:!undefined;step:!undefined

It could be argued that other formatted numbers `style: "percent"` or `style: "currency"`, for example, should also follow this rule. Do you or @spdev3000 think we should expand this change to include that?

hunterloftis (approved):
lgtm, with one idea for future consideration.

Could the extra state (`_forcedUnit`) and duplication of logic (between `NumberField` & `Slider`) be factored out by following a strategy pattern wrapping `NumberFormatter`/`NumberParser`? Alternatively, we could extend those classes in [@internationalized/number](https://www.npmjs.com/package/@internationalized/number) to support these extra units. 

Either way, to the consuming code in NumberField & Slider, there would just be a single `NumberFormatter`/`NumberParser` interface, with the error catching and forced unit fallback encapsulated inside.

Westbrook:
@hunterloftis that does sound like a great approach to follow up work! I've created #1786 to track.

Issue #1768, opened 2021-09-02
[Feat]: [Color Area] Separate label application of "saturation" and "luminosity" values in Color Area


Westbrook:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-color-area

### Description of the requested feature

Surface label customization for the hue and saturation sliders individually so that it can be made clear to screen reader users the value with which they are interacting.

### Implementation notes or ideas

Possibly, a usage of something along the lines of...
```
<sp-color-area
-  label="saturation and luminosity"
+  label-x="saturation"
+  label-y="luminosity"
></sp-color-area>
```
or, maybe:
```
<sp-color-area
-  label="saturation and luminosity"
+  labels='["saturation", "luminosity"]'
></sp-color-area>
```

PR #1801, opened 2021-09-16
feat(color-area): separate X and Y aria labels to improve accessibility

hunterloftis:
## Description

This provides distinct aria labels for the X and Y inputs so that accessibility tools can expose them as `${saturation}, saturation` and `${luminosity}, luminosity` instead of `${saturation_or_luminosity}, saturation and luminosity`.

To ensure backwards-compatibility, so that this can be shipped as a non-breaking change, if the developer specifies a `label` attribute, it will override the `label-x` and `label-y` attributes, maintaining previous behavior.

## Related issue(s)

resolves https://github.com/adobe/spectrum-web-components/issues/1768

## Motivation and context

This makes the Color Area component accessible to users who cannot use its visual representation on a 2D graphic.

## How has this been tested?

1. go to https://separate-color-area-labels--spectrum-web-components.netlify.app/components/color-area
2. activate VoiceOver
3. ensure that you can independently read and edit the saturation / luminosity values in a color area

## Screenshots

https://user-images.githubusercontent.com/364501/133684327-7488d1b0-5935-4d79-ad87-0138e53f62c5.mov

## Types of changes

-   [X] New feature (non-breaking change which adds functionality)

## Checklist

## Best practices

### DIFF
--- a/packages/color-area/src/ColorArea.ts
+++ b/packages/color-area/src/ColorArea.ts
@@ -49,7 +49,13 @@ export class ColorArea extends SpectrumElement {
     public focused = false;
 
     @property({ type: String })
-    public label = 'saturation and luminosity';
+    public label: string | undefined;
+
+    @property({ type: String, attribute: 'label-x' })
+    public labelX = 'saturation';
+
+    @property({ type: String, attribute: 'label-y' })
+    public labelY = 'luminosity';
 
     @query('.handle')
     private handle!: ColorHandle;
@@ -498,7 +504,7 @@ export class ColorArea extends SpectrumElement {
                     type="range"
                     class="slider"
                     name="x"
-                    aria-label=${this.label}
+                    aria-label=${this.label ?? this.labelX}
                     min="0"
                     max="1"
                     step=${this.step}
@@ -513,7 +519,7 @@ export class ColorArea extends SpectrumElement {
                     type="range"
                     class="slider"
                     name="y"
-                    aria-label=${this.label}
+                    aria-label=${this.label ?? this.labelY}
                     min="0"
                     max="1"
                     step=${this.step}
--- a/packages/color-area/README.md
+++ b/packages/color-area/README.md
@@ -63,3 +63,17 @@ An `<sp-color-area>`’s height and width can be customized appropriately for it
         height: var(--spectrum-global-dimension-size-900)"
 ></sp-color-area>
 ```
+
+## Labels
+
+An `<sp-color-area>` renders accessible labels for each axis: _"saturation"_ and _"luminosity"_.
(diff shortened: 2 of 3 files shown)


packages/color-area/src/ColorArea.ts
```diff
@@ -498,7 +504,7 @@ export class ColorArea extends SpectrumElement {
                     type="range"
                     class="slider"
                     name="x"
-                    aria-label=${this.label}
+                    aria-label=${this.label ?? this.labelX}
```
Westbrook, line 507: I may not fully understand the `??` usages here, but should these resolve the other way so that users that accidentally make a PR like the following they get the more custom values?
```
<sp-color-area
    label="Both things"
+  label-x="One thing"
+  label-y="Other thing"
></sp-color-area>
```

Westbrook (changes requested):
I think we consider this to be the optimal approach which means we should apply it to the docs and storybook files as well, right?

packages/color-area/src/ColorArea.ts
Westbrook: Also, we should make a note somewhere, an issue or code comment or something else you  might prefer, to deprecate support for this in the future, hopefully when we got 1.0 or make a specifically breaking `0.x.0` release.

hunterloftis: That *is* how I would do it if this were a new component (is it new enough that we could consider it a new component?).

The reason I have `label` overriding `label-x` / `label-y` is so that it can be totally backwards-compatible. If someone today is using this and specifying `label`, their `label` value will not break if they upgrade to this version. Anyone who isn't specifying a custom `label` will get the new, more-accessible behavior.

hunterloftis:
+1 on applying it to the docs and storybook - right now, there are no color-area stories that mention ("label"). Ditto with the docs.

I'll add a story and a description of the label/label-x/label-y behavior in the docs.

packages/color-area/src/ColorArea.ts
Westbrook: OK, I see now, it's the fact that you moved default values into the `label-*` properties. Mystery of the `??` solved.

We could technically put some sort of getter/setter or change tracking to _unset_ `label` when `label-*` values are set from the outside to ensure that new values are preferred, but that's likely too much extra paths to protect accidentally setting all three properties.

Let's still comment/issue create, but I'm onboard with this!

hunterloftis:
- [x] Documented `label-x`, `label-y`, and the deprecation of `label`
- [x] Created an issue to drop the `label` attribute at 1.0

Westbrook (approved):

PR #1804, opened 2021-09-20
fix: correct add/remove timing of overlay events

Westbrook:
## Description
Close the hole that exists between adding and removing overlay events due to events being added at the beginning of an asynchronous process and them being removed at the end of another which could prevent users from being able to close overlays after interactions of a specific timing.

## Motivation and context
Overlays should be closable.

## How has this been tested?

-   See unit test, it's difficult to recreate the context that causes this without a fairly complex user interface

## Types of changes

## Checklist

### DIFF
--- a/packages/overlay/src/overlay-stack.ts
+++ b/packages/overlay/src/overlay-stack.ts
@@ -43,6 +43,7 @@ export class OverlayStack {
     private trappingInited = false;
     private tabTrapper!: HTMLElement;
     private overlayHolder!: HTMLElement;
+    private _eventsAreBound = false;
 
     private initTabTrapping(): void {
         /* c8 ignore next 4 */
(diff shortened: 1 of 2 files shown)


packages/overlay/src/overlay-stack.ts
```diff
@@ -187,33 +188,21 @@ export class OverlayStack {
     }
 
     private addEventListeners(): void {
+        this._eventsAreBound = true;
```
hunterloftis, line 191: Can we move the `_eventsAreBound` check here so that callers of `addEventListeners` don't need to be aware of it?

```ts
if (this._eventsAreBound) {
  return;
}
this._eventsAreBound = true;
```

Then `openOverlay` could safely call `this.addEventListeners()` directly.

hunterloftis (dismissed):

hunterloftis (approved):

spdev3000:
@Westbrook Thanks for that quick solution here!

Issue #1803, opened 2021-09-19
[Bug]: Range slider throws an error when trying to drag the handles


hkh12:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-slider, sp-slider-handle

### Expected behavior

It just should work.

### Actual behavior

One of the handles doesn't show up, and when I try to drag the handles I get this error:
```
Uncaught TypeError: Cannot destructure property 'input' of 'this.getHandleElements(...)' as it is undefined.
    at HandleController.handlePointermove (HandleController.ts:321)
    at HTMLElement.fn (HandleController.ts:437)
    at HTMLDivElement.handleStream (streaming-listener.ts:72)
```

### Screenshots

![Slider](https://user-images.githubusercontent.com/24871883/133917546-00e02dc2-e723-4d2e-bd25-0c8ac4473bad.png)

### What browsers are you seeing the problem in?

Chrome

### How can we reproduce this issue?

1. Add an `sp-slider[variant=range]`
2. Try to drag the handles
4. Check console
3. See error

### Sample code that illustrates the problem

Copied from [the docs](https://opensource.adobe.com/spectrum-web-components/components/slider-handle/examples)
```html
<sp-slider variant="range" step="1" min="0" max="255">
    Output Levels
    <sp-slider-handle
        slot="handle"
        name="min"
        label="Minimum"
        value="5"
    ></sp-slider-handle>
    <sp-slider-handle
        slot="handle"
        name="max"
        label="Maximum"
        value="250"
    ></sp-slider-handle>
</sp-slider>
```

### Logs taken while reproducing problem

I'm using `@spectrum-web-components/bundle@0.22.7` and I'm testing on Chrome 93

hunterloftis:
Ok, I'm able to repro by loading the bundle via `<script>` in HTML:

https://webcomponents.dev/edit/T2jOTTTBmRDtrh4RYod5/www/index.html

re: @Westbrook this may be due to `<sp-slider>` trying to load state, on upgrade, from `<sp-slider-handle>` before the handle elements are registered.

PR #1829, opened 2021-10-01
fix(slider): make implicit dependency on sp-slider-handle explicit

hunterloftis:
## Description

`<sp-slider>` depends on `<sp-slider-handle>`. Since that dependency is only implicit, it's possible to create a document that upgrades `<sp-slider-handle>` _after_ `<sp-slider>`, breaking the Slider's functionality.

This change specifies the dependency and ensures that `<sp-slider-handle>`s will exist before `<sp-slider>` tries to use them.

## Related issue(s)

Fixes https://github.com/adobe/spectrum-web-components/issues/1803

## Motivation and context

Sliders can currently throw and crash if the document upgrades them out-of-order.

## How has this been tested?

1. Added `slider-handle-upgrade.test.ts` (run with `yarn test`)
2. Manually verified that https://fix-slider-handle-upgrade-order--spectrum-web-components.netlify.app/components/slider-handle continues to work as expected

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/slider/sp-slider.ts
+++ b/packages/slider/sp-slider.ts
@@ -9,6 +9,7 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
 OF ANY KIND, either express or implied. See the License for the specific language
 governing permissions and limitations under the License.
 */
+import './sp-slider-handle.js'; // codify sp-slider's implicit dependency on sp-slider-handle
 import { Slider } from './src/Slider.js';
 
 customElements.define('sp-slider', Slider);
--- a/packages/slider/test/slider-handle-upgrade.test.ts
+++ b/packages/slider/test/slider-handle-upgrade.test.ts
@@ -1,0 +1,42 @@
+/*
+Copyright 2021 Adobe. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License. You may obtain a copy
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import { expect, fixture, html } from '@open-wc/testing';
+import '../sp-slider.js';
+import { Slider } from '../src/Slider.js';
+
+describe('Slider - upgrade order', () => {
+    it('loads both handles when document imports sp-slider-handle after sp-slider', async () => {
+        const el = await fixture<Slider>(html`
+            <sp-slider variant="range" step="1" min="0" max="255">
+                Output Levels
+                <sp-slider-handle
+                    slot="handle"
+                    name="min"
+                    label="Minimum"
+                    value="5"
+                ></sp-slider-handle>
+                <sp-slider-handle
+                    slot="handle"
+                    name="max"
+                    label="Maximum"
+                    value="250"
+                ></sp-slider-handle>
+            </sp-slider>
+        `);
+
+        import('../sp-slider-handle.js');
+
+        const handles = el.shadowRoot.querySelectorAll('.handle');
+        expect(handles).to.have.lengthOf(2);
+    });
+});


packages/slider/sp-slider.ts
```diff
@@ -9,6 +9,7 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
 OF ANY KIND, either express or implied. See the License for the specific language
 governing permissions and limitations under the License.
 */
+import './sp-slider-handle'; // codify sp-slider's implicit dependency on sp-slider-handle
```
Westbrook, line 12: I thought we'd added linting to help prevent this, but apparently not. We want to use file extensions on our imports, so `.js`, here.

We use export maps, now, so this won't break in almost all cases, but it's safer to include them.

packages/slider/test/slider-handle-upgrade.test.ts
```diff
@@ -0,0 +1,42 @@
...
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import { expect, fixture, html } from '@open-wc/testing';
+import '../sp-slider.js';
+import { Slider } from '../src/Slider.js';
+
+describe('Slider', () => {
```
Westbrook, line 17: Nit: can you use a longer name to differentiate this from the tests in https://github.com/adobe/spectrum-web-components/blob/main/packages/slider/test/slider.test.ts#L28 Maybe just `Slider - upgrade order` or something.

packages/slider/src/Slider.ts
```diff
@@ -11,23 +11,22 @@ governing permissions and limitations under the License.
...
-
-import sliderStyles from './slider.css.js';
-import { ObserveSlotText } from '@spectrum-web-components/shared/src/observe-slot-text.js';
-import { StyleInfo } from 'lit-html/directives/style-map';
 import '@spectrum-web-components/field-label/sp-field-label.js';
 import type { NumberField } from '@spectrum-web-components/number-field';
+import { ObserveSlotText } from '@spectrum-web-components/shared/src/observe-slot-text.js';
+import { StyleInfo } from 'lit-html/directives/style-map';
 import { HandleController, HandleValueDictionary } from './HandleController.js';
+import sliderStyles from './slider.css.js';
```
Westbrook, line 29: If this really is just import reordering, can you remove it from this PR?

We should include adding linting to manage order in our 1.0.0 release checklist and have a single _massive_ PR that does this across the library so we dont' end up with this happening in various one of PRs over time.

Westbrook (changes requested):
There's the smallest chance of a surprising bug here wherein a consumer registers/wants to register a custom `sp-slider-handle` element, and this addition leads to a runtime error. Probably an acceptable risk, however.

Do you envision your larger refactor of this interface requiring this? I do think at some point in the future we should fully support late upgrades of handle elements, and could be 100% OK with this in the near term.

packages/slider/sp-slider.ts
hunterloftis: Good catch!

packages/slider/test/slider-handle-upgrade.test.ts
hunterloftis: Good idea 👍

packages/slider/src/Slider.ts
hunterloftis: This is from the automatic vscode setting:

```
"editor.codeActionsOnSave": {
        "source.organizeImports": true
    }
```

...iirc there's one that just removes unused imports; I'll swap to that instead. Otherwise I think all of my PRs will get little changes like this auto-added.

hunterloftis:
> Do you envision your larger refactor of this interface requiring this? I do think at some point in the future we should fully support late upgrades of handle elements, and could be 100% OK with this in the near term.

Agreed! I haven't built anything with this use-case before (implicit dependencies on possibly-extended child custom elements) so I'm curious if you know of an established pattern for it.

My initial idea for a future refactor is to emit events from child Handles when they're ready to be added to their parent Slider. The Slider's `newChildAdded` (or whatever) handler would need to be idempotent in order to gracefully handle 1-N events that may not all happen in the same frame.

Westbrook (approved):
lgtm

PR #1401, opened 2021-04-23
feat: use "lit" instead of "lit-html" and "lit-element"

Westbrook:
## Description
Of the below tests (47), 42 show a clear timing preference for the new `lit` versions vs the currently published versions.

In ALL cases the byte count of the two versions prefers the new `lit` versions by about 113KB, a couple (which include as yet unreleased changes to `sp-action-button` show a very large change).

~One unfortunate mitigation was required to work around the issue in https://github.com/lit/lit/issues/1789, for which a PR has already been approved: https://github.com/lit/lit/pull/1790~

We also needed to internalize the `spread` helper to support its ongoing use in some of our tests. With the creation of `@lit-labs`, it may be that a new community supported version of that functionality surfaces soon, but being we only leverage this in tests at current it seemed that bringing it internal to our repo made sense for now.

Assuming the one issue noted above is included in the next RC, I'd like to suggest that we discuss with seriousness moving to `lit` pre-2.0 in our library. In support of that, I could see this change being moved to a `minor` change from its current `patch` level, however I was quite pleased by the lack of changes needed by our library (outside of the `streamingListener` refactor).

## Motivation and Context
Smaller. Faster. Easier to update. New features. 

## How Has This Been Tested?
The full suite of tests, etc.

Perf comments: https://github.com/adobe/spectrum-web-components/pull/1401#issuecomment-885036800
Preview comments: https://github.com/adobe/spectrum-web-components/pull/1401#issuecomment-889079203

## Types of changes
- [x] Dependency update.

## Checklist:
- [x] I have signed the [Adobe Open Source CLA](http://opensource.adobe.com/cla.html).
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #1908, opened 2021-11-05
docs: add migration instructions to the first lit@2.0 releases

Westbrook:
## Description
Add migration instructions for the pending `lit@2.0` release!

## Motivation and context
Docs are great.

## How has this been tested?

-   [ ] _Test case 1_
    1. Go to https://lit-migration--spectrum-web-components.netlify.app/guides/migration-guide/
    2. Read the page.

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/projects/documentation/content/guides/migration-guide.md
+++ b/projects/documentation/content/guides/migration-guide.md
@@ -1,0 +1,132 @@
+---
+layout: guide.njk
+title: 'Migration Guide: Spectrum Web Components'
+displayName: Migration Guide
+slug: migration-guide
+---
+
+## Migration: 8/11/2021
+
+As of 8/11/2021, Spectrum Web Components will cease to leverage `lit-element` and `lit-html` dependencies directly in favor of [`lit2.0`](https://lit.dev/blog/2021-09-21-announcing-lit-2/). For most of you this will be a 100% transparent change as the usage of these libraries is encapsulated within our elements and consuming these new versions within your HTML of Javascript application should do little more than pass the bundle size and performance benefits of this change on to your applications. If you are currently using any of the following package version you should be seeing these benefits in your applications today. In the case that you are using those packages in concert with versions before those listed below we highly suggest that you align your version consumption to the most recent versions as soon as is comfortable for your project.
+
+<details>
+	<summary><strong>Package Versions</strong></summary>
+	<div class="table-container">
+		<table class="spectrum-Table spectrum-Table--sizeM">
+			<thead class="spectrum-Table-head">
+				<tr>
+					<th class="spectrum-Table-headCell">Package</th>
+					<th class="spectrum-Table-headCell">Version</th>
+				</tr>
+			</thead>
+			<tbody class="spectrum-Table-body">
+				{% for pkg in versions %}<tr class="spectrum-Table-row">
+					<td class="spectrum-Table-cell">
+						<a href="{{ pkg.dir }}">{{ pkg.name }}</a>
+					</td>
+					<td class="spectrum-Table-cell">
+						<a href="https://www.npmjs.com/package/{{ pkg.name }}/bundle/v/{{ pkg.version }}">{{ pkg.version }}</a>
+					</td>
+				<tr>{% endfor %}
+			</tbody>
+		</table>
+	</div>
+</details>
+
+### Using `lit@2.0` inside of `lit-html` and/or `LitElement`
+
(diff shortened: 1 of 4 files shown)


projects/documentation/content/guides/migration-guide.md
```diff
@@ -0,0 +1,132 @@
...
+					<td class="spectrum-Table-cell">
+						<a href="https://www.npmjs.com/package/{{ pkg.name }}/bundle/v/{{ pkg.version }}">{{ pkg.version }}</a>
+					</td>
+				<tr>{% endfor %}
+			</tbody>
+		</table>
+	</div>
+</details>
+
+### Using `lit@2.0` inside of `lit-html` and/or `LitElement`
```
Westbrook, line 36: Clarifies the required changes we've seen in test upgrades to date.

projects/documentation/content/guides/migration-guide.md
```diff
@@ -0,0 +1,132 @@
...
+        return html`
+            // ...
+            <sp-icon-redo></sp-icon-redo>
+            // ...
+        `;
+    }
+}
+```
+
+### Javascript re-exports
```
Westbrook, line 87: Clarifies some changes that are not yet in #1401 but have been previously announced the the leads team.

projects/documentation/content/guides/migration-guide.md
```diff
@@ -0,0 +1,132 @@
...
+        return html`
+            // ...
+            <sp-icon>${RedoIcon()}</sp-icon>
+            // ...
+        `;
+    }
+}
+```
+
+In this case the `RedoIcon()` literal will be wrapped in the `html` template literal tag as supplied by `@spectrum-web-components/icons-workflow`. After this update, the version of this tag will no longer be compatible with the `html` template literal tag applied in the `render()` method of your custom element. You can correct this by applying your local value for `html` to the icon literals:
```
hunterloftis, line 57:
```suggestion
Here, `RedoIcon()` returns a value built with the `html` dependency of `@spectrum-web-components/icons-workflow`. After this update, that value will no longer be compatible with the `html` tag in the `render()` function here, which is still built with the `lit-html` module.

There are two ways to correct this. We recommend replacing template-literal icons with icon elements, resulting in better performance and decoupled dependencies:
```

projects/documentation/content/guides/migration-guide.md
```diff
@@ -0,0 +1,132 @@
...
+
+```js
+import { html } from 'lit-html';
+import {
+    RedoIcon,
+    setCustomTemplateLiteralTag,
+} from '@spectrum-web-components/icons-workflow';
+
+setCustomTemplateLiteralTag(html);
+```
```
hunterloftis, line 67: If you accept the above suggestion, the examples will need to be flipped.

projects/documentation/content/guides/migration-guide.md
```diff
@@ -0,0 +1,132 @@
...
+
+Or, you can avoid running into issues like this in the future, while leveraging more performant icon delivery, by leveraging fully registered icon elements instead:
```
hunterloftis, line 69:
```suggestion
Alternatively, provide your local value for `html` to the icon templates via `setCustomTemplateLiteralTag`:
```

projects/documentation/content/guides/migration-guide.md
```diff
@@ -0,0 +1,132 @@
...
+export { repeat } from 'lit/directives/repeat.js';
+export { classMap } from 'lit/directives/class-map.js';
+export type { ClassInfo } from 'lit/directives/class-map.js';
+export { styleMap } from 'lit/directives/style-map.js';
+export type { StyleInfo } from 'lit/directives/style-map.js';
+export { until } from 'lit/directives/until.js';
+export { live } from 'lit/directives/live.js';
+```
+
+Additional `lit` exports can be acquired directly from that package and with the reduction of `instanceof`-centric checking within the `lit-html` library it is possible that overtime we lean towards that being the preferred path towards acquiring this values.
```
hunterloftis, line 132:
```suggestion
Additional `lit` exports can be acquired directly from the `lit` package.

With Lit 2.0 bringing same-document library-version flexibility, re-exporting from shared base libraries is less critical. We may eventually recommend direct imports from `lit` as the preferred path towards acquiring these values.
```

hunterloftis (dismissed):
Let's ship it!

I left several copy suggestions, nothing blocking.

PR #1914, opened 2021-11-06
fix: abstract "hasVisibleFocusInTree" functionality and return trigger focus after close

### LINKED ISSUE #1894: [Bug]: split-button stops showing its menu after an open->close
### Code of conduct

### Impacted component(s)

split-button

### Expected behavior

Clicking the down arrow on an `<sp-split-button>` repeatedly should open and close its menu.

### Actual behavior

Clicking the down arrow on an `<sp-split-button>` opens and closes its menu once, then never again.

Clicking the down arrow, then selecting a new option from the menu each time, works as expected.

### Screenshots

https://user-images.githubusercontent.com/364501/139466363-bcff516b-4c2e-487c-80ff-8890ed3b35ac.mov

_No response_

1. Go to https://opensource.adobe.com/spectrum-web-components/components/split-button/
2. Click on a split button's down arrow repeatedly

_No response_

_No response_

Westbrook:
## Description
- correct extension from Picker so that both `<sp-action-menu>` and `<sp-split-button>` can now toggle
- abstract "hasVisibleFocusInTree" functionality to base with extra protection
- update `focusElement` on `sp-split-button[left]` elements

## Related issue(s)

- fixes #1894

## Motivation and context
Was broken

## How has this been tested?
-   [ ] _Test case 1_
    1. Go to https://splitter-toggle--spectrum-web-components.netlify.app/components/action-menu/#sizes
    2. Toggle the menu open/closed a few times
-   [ ] _Test case 2_
    1. Go to https://splitter-toggle--spectrum-web-components.netlify.app/components/split-button/#sizes
    2. Toggle the menu open/closed a few times

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/base/src/Base.ts
+++ b/packages/base/src/Base.ts
@@ -25,6 +25,7 @@ type Constructor<T = Record<string, unknown>> = {
 export interface SpectrumInterface {
     shadowRoot: ShadowRoot;
     isLTR: boolean;
+    hasVisibleFocusInTree(): boolean;
     dir: 'ltr' | 'rtl';
 }
 
@@ -78,6 +79,26 @@ export function SpectrumMixin<T extends Constructor<UpdatingElement>>(
             return this.dir === 'ltr';
         }
 
+        public hasVisibleFocusInTree(): boolean {
+            const activeElement = (this.getRootNode() as Document)
+                .activeElement as HTMLElement;
+            if (!activeElement) {
+                return false;
+            }
+            // Browsers without support for the `:focus-visible`
+            // selector will throw on the following test (Safari, older things).
+            // Some won't throw, but will be focusing item rather than the menu and
+            // will rely on the polyfill to know whether focus is "visible" or not.
+            try {
+                return (
+                    activeElement.matches(':focus-visible') ||
+                    activeElement.matches('.focus-visible')
+                );
+            } catch (error) {
+                return activeElement.matches('.focus-visible');
+            }
+        }
+
         public connectedCallback(): void {
             if (!this.hasAttribute('dir')) {
                 let dirParent = ((this as HTMLElement).assignedSlot ||
@@ -90,7 +111,7 @@ export function SpectrumMixin<T extends Constructor<UpdatingElement>>(
                 ) {
                     dirParent = ((dirParent as HTMLElement).assignedSlot || // step into the shadow DOM of the parent of a slotted node
                         dirParent.parentNode || // DOM Element detected
-                        ((dirParent as unknown) as ShadowRoot)
+                        (dirParent as unknown as ShadowRoot)
                             .host) as HTMLElement;
                 }
                 this.dir =
@@ -104,9 +125,9 @@ export function SpectrumMixin<T extends Constructor<UpdatingElement>>(
                         !customElements.get(localName)
                     ) {
                         customElements.whenDefined(localName).then(() => {
-                            (dirParent as ThemeRoot).startManagingContentDirection(
-                                this
-                            );
+                            (
+                                dirParent as ThemeRoot
+                            ).startManagingContentDirection(this);
                         });
                     } else {
                         (dirParent as ThemeRoot).startManagingContentDirection(
--- a/packages/color-wheel/src/ColorWheel.ts
+++ b/packages/color-wheel/src/ColorWheel.ts
@@ -260,23 +260,7 @@ export class ColorWheel extends Focusable {
     }
 
     private forwardFocus(): void {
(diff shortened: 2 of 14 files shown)


packages/base/src/Base.ts
```diff
@@ -78,6 +79,27 @@ export function SpectrumMixin<T extends Constructor<UpdatingElement>>(
...
+                // will rely on the polyfill to know whether focus is "visible" or not.
+                hasVisibleFocusInTree =
+                    activeElement.matches(':focus-visible') ||
+                    activeElement.matches('.focus-visible');
+            } catch (error) {
+                hasVisibleFocusInTree = activeElement.matches('.focus-visible');
+            }
+            return hasVisibleFocusInTree;
+        }
+
```
hunterloftis, line 102: Nice! I like that outside of tests, this PR deletes more code than it adds, by consolidating this shared functionality.

Is there a benefit to storing the temporary result in a variable vs. returning from whichever case?

```suggestion
        public hasVisibleFocusInTree(): boolean {
            const activeElement = (this.getRootNode() as Document)
                .activeElement as HTMLElement;
            if (!activeElement) {
                return false;
            }
            // Browsers without support for the `:focus-visible`
            // selector will throw on the following test (Safari, older things).
            // Some won't throw, but will be focusing item rather than the menu and
            // will rely on the polyfill to know whether focus is "visible" or not.
            try {
                return activeElement.matches(':focus-visible') || activeElement.matches('.focus-visible');
            } catch (error) {
                return activeElement.matches('.focus-visible');
            }
        }
```

packages/color-wheel/src/ColorWheel.ts
```diff
@@ -442,16 +426,16 @@ export class ColorWheel extends Focusable {
...
-                    this.requestUpdate();
+            this.observer = new (
+                window as unknown as WithSWCResizeObserver
+            ).ResizeObserver((entries: SWCResizeObserverEntry[]) => {
+                for (const entry of entries) {
+                    this.boundingClientRect = entry.contentRect;
                 }
-            );
+                this.requestUpdate();
+            });
```
hunterloftis, line 438: After https://github.com/adobe/spectrum-web-components/pull/1401 we can extract this into a shared Controller.

packages/picker/src/Picker.ts
```diff
@@ -457,8 +443,8 @@ export class PickerBase extends SizedMixin(Focusable) {
                 <sp-menu
                     id="menu"
                     role="${this.listRole}"
-                    selects="single"
                     @change=${this.handleChange}
+                    .selects=${this.selects}
```
hunterloftis, line 447: It looks like this is intentionally setting a property rather than an attribute. Could you share the motivation?

packages/split-button/src/SplitButton.ts
```diff
@@ -72,16 +72,14 @@ export class SplitButton extends SizedMixin(PickerBase) {
...
-        if (this.disabled) {
-            return;
+    public get focusElement(): HTMLElement {
+        if (this.open) {
+            return this.optionsMenu;
         }
         if (this.left) {
-            this.trigger.focus();
-            return;
+            this.trigger;
```
hunterloftis, line 80: Is there meant to be a return here?

```suggestion
            return this.trigger;
```

hunterloftis (changes requested):
Thanks for figuring out this issue.

I made a few optional suggestions; the only thing to definitely resolve is the `left` branch of `get focusElement`.

packages/split-button/src/SplitButton.ts
Westbrook: Bwah! I was just like "yeah, these tests are supposed to change..." 😞

packages/color-wheel/src/ColorWheel.ts
Westbrook: Totally. It also doesn't "change" as much as the linter decided it was no long on style.

packages/picker/src/Picker.ts
Westbrook: In this case the property is used to overcome the typing. Typescript likes to point out that `this.selects` can be `undefined` which is not an acceptable type for the `selects` attribute, and would prefer `ifDefined` be used here to manage that. However, `Menu.selects` is a reflecting property, so it we know that it will do the work of `ifDefined` internally when supplied `undefined` as a property, so this intends to simplify the usage.

hunterloftis: Gotcha, thanks!

hunterloftis (approved):

PR #1931, opened 2021-11-12
Help text

### LINKED ISSUE #1704: Add Helptext pattern
https://spectrum.adobe.com/page/text-field/#Help-text-(description-and-error-message)
https://github.com/adobe/spectrum-css/issues/1156

### Expected Behaviour

The new helptext component should be available in the library.

### Actual Behaviour

It's missing.

Westbrook:
## Description
Add `<sp-help-text>` pattern.
- ship `HelpTextManager` helper class to support delivery of Help Text in other elements
- update `<sp-field-group>`, `<sp-number-field>`, `<sp-radio-group>`, `<sp-search>`, and `<sp-textfield>` to support this pattern
- add `findDescribedNode` helper to the unit testing suite to cover over eccentricities in WebKit support for described content

## Related issue(s)

- fixes #1704

## Motivation and context
If it's in Spectrum _AND_ Spectrum CSS it should likely be here! Also, we want to make your forms even nicer looking.

## How has this been tested?
-   [ ] _Test case 1_
    1. Go to:
      - https://help-text--spectrum-web-components.netlify.app/components/help-text/
      - https://help-text--spectrum-web-components.netlify.app/components/field-group/
      - https://help-text--spectrum-web-components.netlify.app/components/radio/
      - https://help-text--spectrum-web-components.netlify.app/components/textfield/
      - https://help-text--spectrum-web-components.netlify.app/components/textarea/
    2. Experience the new element and its alignment with https://opensource.adobe.com/spectrum-css/helptext.html and https://spectrum.adobe.com/page/help-text/
-   [ ] _Test case 2_
    1. Go to https://help-text--spectrum-web-components.netlify.app/components/help-text/
    2. With the screen reader, make sure that the Help Text is read in association with the form element that it is placed within AS WELL AS singularly
    3. Toggle the validity to ensure that this content live updates as well

## Types of changes

## Checklist

CC: @davidkudera

### DIFF
--- a/packages/help-text/README.md
+++ b/packages/help-text/README.md
@@ -1,0 +1,89 @@
+## Description
+
+An `<sp-help-text>` provides either an informative description or an error message that gives more context about what a user needs to input. It's commonly used in forms.
+
+### Usage
+
+[![See it on NPM!](https://img.shields.io/npm/v/@spectrum-web-components/help-text?style=for-the-badge)](https://www.npmjs.com/package/@spectrum-web-components/help-text)
+[![How big is this package in your project?](https://img.shields.io/bundlephobia/minzip/@spectrum-web-components/help-text?style=for-the-badge)](https://bundlephobia.com/result?p=@spectrum-web-components/help-text)
+
+```
+yarn add @spectrum-web-components/help-text
+```
+
+Import the side effectful registration of `<sp-help-text>` via:
+
+```
+import '@spectrum-web-components/help-text/sp-help-text.js';
+```
+
+When looking to leverage the `HelpText` base class as a type and/or for extension purposes, do so via:
+
+```
+import { HelpText } from '@spectrum-web-components/help-text';
+```
+
+## Sizes
+
+<sp-tabs selected="m" auto label="Size Attribute Options">
+<sp-tab value="s">Small</sp-tab>
+<sp-tab-panel value="s">
+
+```html
+<sp-help-text size="s">Passwords must be at least 8 characters.</sp-help-text>
+```
+
+</sp-tab-panel>
+<sp-tab value="m">Medium</sp-tab>
+<sp-tab-panel value="m">
+
+```html
+<sp-help-text size="m">Passwords must be at least 8 characters.</sp-help-text>
+```
+
+</sp-tab-panel>
+<sp-tab value="l">Large</sp-tab>
+<sp-tab-panel value="l">
+
+```html
+<sp-help-text size="l">Passwords must be at least 8 characters.</sp-help-text>
+```
+
+</sp-tab-panel>
+<sp-tab value="xl">Extra Large</sp-tab>
+<sp-tab-panel value="xl">
+
+```html
+<sp-help-text size="xl">Passwords must be at least 8 characters.</sp-help-text>
+```
+
+</sp-tab-panel>
+</sp-tabs>
+
+## Negative
+
+The negative variant of `<sp-help-text>` is used to convey error messages. An error message should be different than the informative message otherwise delivers to the visitor and should show a solution for correcting the error that has been encountered.
+
+```html
+<sp-help-text variant="negative">
+    Create a password with at least 8 characters.
+</sp-help-text>
+```
+
+### Icon
+
+When associated with content that does not supply an icon outlining the presence of an error, use the `icon` attribute to display one as part of the `<sp-help-text>` element.
+
+```html
+<sp-help-text variant="negative" icon>
+    Create a password with at least 8 characters.
+</sp-help-text>
+```
+
+## Disabled
+
+When associated to content the is disabled, use the `disabled` attribute to match the delivery of the `<sp-help-text>` element to that content.
+
+```html
+<sp-help-text disabled>Passwords must be at least 8 characters.</sp-help-text>
+```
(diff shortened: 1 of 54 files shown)


najikahalsema (approved):
LGTM!! Nice work on all of this.

jnurthen:
The Self-managed radio group doesn't seem to work correctly. Is that intentional or a bug?

Westbrook:
@jnurthen here? https://help-text--spectrum-web-components.netlify.app/components/radio/#help-text There is a bug! Is that what you're hitting? If so, great catch, will fix, if not, please share more. 🙏🏼

packages/help-text/README.md
```diff
@@ -0,0 +1,328 @@
...
+    value="Positive"
+    oninput='
+        const helpText = this.querySelector(`[slot="help-text"]`);
+        helpText.textContent = this.invalid ? `Please be "Positive".` : `Tell us how you are feeling today.`;
+        helpText.variant = this.invalid ? `negative` : `neutral`;
+    '
+>
+    <sp-help-text slot="neutral-text">
+        Tell us how you're feeling today.
+    </sp-help-text>
```
hunterloftis, line 322: Since `neutral-text` is nested within `help-text`, is this unnecessary?

packages/help-text/README.md
```diff
@@ -0,0 +1,328 @@
...
+
+The `createHelpTextManager()` is called with default options of `{ mode: 'internal' }` and should be leveraged to associate slotted `<sp-help-text>` elements with content within the shadow root of the parent element.
+
+```js
+import { SpectrumElement, html } from '@spectrum-web-components/base';
+import { createHelpTextManager } from '@spectrum-web-components/help-text/manager.js';
+
+export class ParentWithDescribedElement extends SpectrumElement {
+    helpTextManager = createHelpTextManager({ mode: 'internal' });
```
hunterloftis, line 102: I was going to ask why a helper method (`createHelpTextManager`) was needed, instead of just `new HelpTextManager(this)`... but I can't find an implementation for `createHelpTextManager` in this PR. Do I need more coffee?

packages/help-text/README.md
```diff
@@ -0,0 +1,328 @@
...
+        return html`
+            ${this.helpTextManager.render(this.invalid)}
+        `;
+    }
+}
+```
+
+### Usage with self managed validity
+
+When the parent element into which you are placing an `<sp-help-text>` manages its own validity in some way, you can leverage content in the `neutral-text` and the `negative-text` slots together and rely on the `helpTextManager()` supplied templating to surface the appropriate one based off of that internal state.
```
hunterloftis, line 136: I was trying to figure out of the `HelpTextManager` listened for changes in its parent's `.invalid` state or something - I think the reality is less general than currently described here. This won't automatically work on any component that manages its own validity in some way... that component needs to specifically be designed to interact with `<sp-help-text>`:

```suggestion
Spectrum Web Components that support automatic validation, like `<sp-fieldgroup>` and `<sp-textfield>`, automatically render either `neutral-text` or `negative-text` slots based on validity. Provide both, and the appropriate `<sp-help-text>` element will be surfaced:
```

hunterloftis (commented):
Since accepting slotted `<sp-help-text>` elements requires changes to the implementation of a component:

- instantiating a `HelpTextManager`
- rendering the manager's template as part of the owning component's `render()`

I like your idea of making that coupling explicit by moving these responsibilities into a Class mixin. It seems fair to say that a certain Class _is a_ `HelpTextManaged` class. The mixin could automatically:

- create a `HelpTextManager`
- provide a `renderHelpText()` method
- (in future) provide a way for `<sp-help-text>` to warn about incorrect usage (being slotted into an incompatible container) in dev-mode
- (in future) provide a way to warn if `render()` never uses the results from `renderHelpText()` in dev-mode

packages/help-text/README.md
Westbrook: Updated lated 🤦🏼

packages/help-text/help-text-mixin.md
```diff
@@ -0,0 +1,268 @@
...
+```
+import { HelpTextManagedElement } from '@spectrum-web-components/help-text/HelpTextManagedElement.js';
+```
+
+The `HelpTextManager` class is also available for import as:
+
+```
+import { HelpTextManager } from '@spectrum-web-components/help-text/HelpTextManager.js';
+```
+
```
hunterloftis, line 33:
```suggestion

## HelpTextMixin

The `HelpTextMixin` mixes two properties into your class:

- `helpTextId`: the `id` attribute of the associated `<sp-help-text>`
- `renderHelpText`: a function that renders a`TemplateResult` with the `help-text` and `negative-help-text` slots

```

packages/help-text/help-text-mixin.md
```diff
@@ -0,0 +1,268 @@
...
+        return html`
+            ${this.renderHelpText(this.invalid)}
+        `;
+    }
+}
+```
+
+This functionality is powered by the `HelpTextManager` class which is also exported from this package and can be leveraged directly. It accepts the root element on which within which it will manage help text and a options object that accepts the `mode` by which that help text will be managed at construction time. Leveraged at render time it surfaces an `id` property and a `render(invalid?: boolean)` method for use in your template.
+
+## Usage with self managed validity
```
hunterloftis, line 77:
```suggestion
## Usage with self-managed validity
```

hunterloftis (dismissed):
Thanks for making the change to mixin + class... the lazy developer in me will be happy to just `extends HelpTextManagedElement` and be on my way.

packages/help-text/src/helpTextMixin.ts
```diff
@@ -0,0 +1,41 @@
+/*
```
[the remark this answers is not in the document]
Westbrook: I do this because a `mixin` is not in and of itself a class. I actually do this as kabob case in other uses 😱 https://github.com/adobe/spectrum-web-components/blob/main/packages/shared/src/observe-slot-text.ts should I match that up/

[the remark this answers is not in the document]
Westbrook: I worked the naming here a little bit to be more normative to other mixins in the library.

PR #1934, opened 2021-11-16
fix(action-button): all "selected" Action Buttons should be "aria-pressed=true"

### LINKED ISSUE #756: should actionButton selected also set aria-pressed?
Perhaps this is just a doc issue where we should note that if managing the selected attribute the user should also manage the aria-pressed attribute so the button is read as a toggle button correctly.
I don't think we can do this for them if they aren't using the toggles attribute

Westbrook:
## Description
Apply `aria-pressed="true"` to all Action Buttons that `role="button"`. Avoid adding this attribute when other roles are applied as they are not part of the expected accessibility tree in that case.

## Related issue(s)

- fixes #756

## Motivation and context
More accessibility by default.

## How has this been tested?

-   [ ] _Test case 1_
    1. Go to https://action-button-selected--spectrum-web-components.netlify.app/components/action-button/#sizes
    2. Add the `selected` attribute to one of the Action Buttons there
    3. See that the `aria-pressed=true` attribute is also applied.
-   [ ] _Test case 2_
    1. Go to https://action-button-selected--spectrum-web-components.netlify.app/components/action-group/#single
    2. See that the Action Button there with the `selected` attribute does not have `aria-pressed=true`
-   [ ] _Test case 3_
    1. Go to https://action-button-selected--spectrum-web-components.netlify.app/components/action-button/#toggles
    2. See that the Action Button there without the `selected` attribute continue to have `aria-pressed=false`

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/action-button/src/ActionButton.ts
+++ b/packages/action-button/src/ActionButton.ts
@@ -64,9 +64,21 @@ export class ActionButton extends SizedMixin(ButtonBase) {
     @property({ type: Boolean, reflect: true })
     public quiet = false;
 
+    @property({ reflect: true })
+    public role = 'button';
+
+    /**
+     * Whether an Action Button with `role='button'`
+     * should also be `aria-pressed='true'`
+     */
     @property({ type: Boolean, reflect: true })
     public selected = false;
 
+    /**
+     * Whether to automatically manage the `selected`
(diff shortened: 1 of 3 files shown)


najikahalsema (approved):
Looks good to me. Thank you for making these fixes!

PR #1945, opened 2021-11-19
Update send mouse usage

### LINKED ISSUE #1784: [Feat]: sendMouse helper with clean up
### Code of conduct

### Impacted component(s)

testing

When writing test we can use `await executeServerCommand('send-mouse', {})` to have Playwright send an actual mouse interaction to the browser. This is powerful, but we've realized that if we don't clean up the mouse at the end of the test there might be overflow into the next test causing unexpected results.

We should add a `sendMouse()` helper that orchestrates the `executeServerCommand()` call internally while also attempting to attach clean up scripting so that test writers don't have to think about adding it manually the `afterEach` or similar in their tests.

### Mockups or screenshots

_No response_

The following is a riff off of similar work for the `fixture()` helper in OpenWC: https://github.com/open-wc/open-wc/blob/master/packages/testing-helpers/src/fixture.js#L8. Adding something like the this into the `sendMouse()` method should allow us to easily clean up this work.
```
/**
 * This registers the fixture cleanup as a side effect
 */
try {
  // we should not assume that our users load mocha types globally
  // @ts-ignore
  if ('afterEach' in window) {
    // @ts-ignore
    afterEach(() => {
      mouseCleanup();
    });
  }
  // @ts-ignore
  if ('teardown' in window) {
    // @ts-ignore
    teardown(() => {
      mouseCleanup();
    });
  }
} catch (error) {
  /* do nothing */
}
```
Things to keep in mind:
- updating the mouse is an asynchronous activity, we need to ensure that clean up is one time rather than additive
  - prevent multiple uses in a single `it()` from causing multiple clean ups
  - prevent multiple uses in a single `describe()` from causing multiple clean ups
- It may be useful to mouse up AND to move the mouse back to `[0,0]` to ensure that the mouse does not interact with the test context at all

Westbrook:
## Description
Create a `sendMouse` helper method that enacts the command while also submitting a clean up call to run after the test.
- apply this helper to all `send-mouse` plugin usage locations

## Related issue(s)

- fixes #1784 

## Motivation and context
Less likely to have surprising test errors based on mouse context and location.

## How has this been tested?
-   [ ] _Test case 1_
    1. Tests ran in CI
    2. Feel free to try them locally.

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/number-field/test/number-field.test.ts
+++ b/packages/number-field/test/number-field.test.ts
@@ -25,17 +25,14 @@ import {
 } from '../stories/number-field.stories.js';
 import '../sp-number-field.js';
 import { FRAMES_PER_CHANGE, indeterminatePlaceholder, NumberField } from '..';
-import {
-    executeServerCommand,
-    sendKeys,
-    setUserAgent,
-} from '@web/test-runner-commands';
+import { sendKeys, setUserAgent } from '@web/test-runner-commands';
 import { spy } from 'sinon';
 import {
     clickBySelector,
     createLanguageContext,
     getElFrom,
 } from './helpers.js';
+import { sendMouse } from '../../../test/plugins/browser.js';
 
 describe('NumberField', () => {
     before(async () => {
@@ -450,7 +447,7 @@ describe('NumberField', () => {
                 '.stepUp'
             ) as HTMLElement;
             const buttonUpRect = buttonUp.getBoundingClientRect();
-            const buttonUpPosition = [
+            const buttonUpPosition: [number, number] = [
                 buttonUpRect.x + buttonUpRect.width / 2,
                 buttonUpRect.y + buttonUpRect.height / 2,
             ];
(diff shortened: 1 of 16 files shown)


packages/number-field/test/number-field.test.ts
```diff
@@ -462,11 +459,11 @@ describe('NumberField', () => {
                 buttonDownRect.x + buttonDownRect.width / 2,
                 buttonDownRect.y + buttonDownRect.height / 2,
             ];
```
hunterloftis, line 461: GH won't let me suggest all of these (too far from touched code), but wherever we're responsible for creating the original value and for consuming it, we should specify the type on the value itself instead of asserting it later:

```suggestion
            ] as [number, number];
```

packages/number-field/test/number-field.test.ts
```diff
@@ -481,11 +478,11 @@ describe('NumberField', () => {
             expect(el.value).to.equal(52);
             expect(inputSpy.callCount).to.equal(2);
             expect(changeSpy.callCount).to.equal(0);
-            executeServerCommand('send-mouse', {
+            sendMouse({
                 steps: [
                     {
                         type: 'move',
-                        position: buttonDownPosition,
+                        position: buttonDownPosition as [number, number],
```
hunterloftis, line 485:
```suggestion
                        position: buttonDownPosition,
```

hunterloftis (approved):
Awesome, this will save future test authors some head-scratching.

packages/number-field/test/number-field.test.ts
Westbrook: Good call. Added on the `const variable: [number, number]` version wherever used here.

Also, got CI to pass, so 🥳

PR #1961, opened 2021-11-30
docs: include description and type for all event listing in the docs

### LINKED ISSUE #1176: [search] Fix "events" listing in docs.
The current events listing is as follows:

image

`keydown`, `change`, and `input` are likely a little much here, but `submit` should reflect some of the `submit` listing should likely take on some of the content from the "Examples" page:

> The submit event fires when the <sp-search> is submitted. This is a non-composed event inline with what you would expect a <form />{:target="_blank"} to fire. If you choose to prevent the default of this event, the default action for the underlying <form /> element will also be prevented, which will allow you to handle the search action in javascript.

`@fires` and `@private` will likely be needed in various places to clarify this for the analyzer.

Westbrook:
## Description
Ensure that all events have both a description and type in our docs site.

## Related issue(s)

- fixes #1176

## Motivation and context
Easier to consume docs.

## How has this been tested?

-   [ ] _Test case 1_
    1. Go to https://events-docs--spectrum-web-components.netlify.app/components/search/api/#events
    2. See that the docs have a nice description and type for each event
    3. See that other elements have the similar niceties...

New version of Search component events
vs
Old version of Search component events

## Types of changes

## Checklist

### DIFF
--- a/packages/accordion/src/AccordionItem.ts
+++ b/packages/accordion/src/AccordionItem.ts
@@ -26,6 +26,7 @@ import styles from './accordion-item.css.js';
 /**
  * @element sp-accordion-item
  * @slot - The content of the item that is hidden when the item is not open
+ * @fires sp-accordion-item-toggle - Announce that an accordion item has been toggled while allowing the event to be cancelled.
  */
 export class AccordionItem extends Focusable {
     public static get styles(): CSSResultArray {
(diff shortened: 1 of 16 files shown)


hunterloftis:
This change seems to add some events and remove others - for example, `sp-slider` [currently documents `keydown`](https://opensource.adobe.com/spectrum-web-components/components/slider/) (although the current docs miss its `input` and `change` events). This branch lists [no events for `sp-slider`](https://events-docs--spectrum-web-components.netlify.app/components/slider/api/).

Westbrook:
Good catch on the `input`/`change` for slider, I thought I added those, and will get them in. The `keyboard` event is not something that should publish, IMO, so it's been removed everywhere.

hunterloftis (approved):
Gotcha - in that case, let's merge this improvement, then add any others we notice missing to JSDoc `@fires` as we find them.

PR #1967, opened 2021-11-30
fix(slider): update "value" default to match browser native range input

### LINKED ISSUE #1788: [Slider]: why does value default to 10?
### Code of conduct

### Description of issue

The value of a slider [defaults to 10](https://github.com/adobe/spectrum-web-components/blob/de3151b9e940121f518f1d94d2601cb87ba95ddb/packages/slider/src/SliderHandle.ts#L88). Is there a reason for this, or is it arbitrary? It surprised me as a user, since I would expect the [defaultValue from input type="range"](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range#value).

Changing the default `value` of `sp-slider` and `sp-slider-handle` would be a breaking change. Is v1.0 a good opportunity for this?

Westbrook:
BREAKING CHANGE: When not supplied value will now be set to halfway between min and max.

## Description
Update default `value` when no applied by the application. Update tests to match.

## Related issue(s)

- fixes #1788

## Motivation and context
Be more like native elements.

## How has this been tested?

-   [ ] _Test case 1_
    1. With unit tests...

## Types of changes

## Checklist

### DIFF
--- a/packages/slider/src/SliderHandle.ts
+++ b/packages/slider/src/SliderHandle.ts
@@ -70,6 +70,9 @@ const MaxConverter = {
 
 /**
  * @element sp-slider-handle
+ *
+ * @fires input - The value of the element has changed.
+ * @fires change - An alteration to the value of the element has been committed by the user.
  */
 export class SliderHandle extends Focusable {
     public handleController?: Controller;
--- a/packages/slider/src/SliderHandle.ts
+++ b/packages/slider/src/SliderHandle.ts
@@ -88,8 +88,12 @@ export class SliderHandle extends Focusable {
 
     _forcedUnit = '';
 
+    /**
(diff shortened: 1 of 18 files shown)


packages/slider/src/SliderHandle.ts
```diff
@@ -131,6 +135,16 @@ export class SliderHandle extends Focusable {
         super.update(changes);
     }
 
+    protected firstUpdated(changedProperties: PropertyValues<this>): void {
+        super.firstUpdated(changedProperties);
+        const { max, min } = this as { max: number; min: number };
+        if (!this.value) {
```
hunterloftis, line 141: We should specifically check for missing value, since `!0` is `true`:

```suggestion
        if (this.value == null) {
```

hunterloftis (approved):

PR #1971, opened 2021-12-03
refactor(overlay): reduce paint/render work when opening a modal

### LINKED ISSUE #1910: [Bug]: Picker menu is very slow to open
### Code of conduct

### Impacted component(s)

Picker

### Expected behavior

When you click on a picker button, the menu opens pretty immediately.

### Actual behavior

With the latest release it's taking around a second to open.

### Screenshots

_No response_

_No response_

1. Go to https://opensource.adobe.com/spectrum-web-components/components/picker/
2. Click on a picker button (and observer the time it takes)
3. Repeat step 2 and notice that it's not faster on subsequent attempts.

_No response_

_No response_

Westbrook:
## Description
- correct application of theme and remove body margin in benchmarking
- expand `<sp-split-button>` benchmark to include 5 open/close actions on one element rather than init on 100
- update `<sp-picker>` and derivatives to gather menu items on a document fragment rather than in the DOM before projecting into an overlay
- reorder and conditionalize work within the overlay process to reduce unneeded work
- update CSS to reduce paints when opening overlay

We get a small, but clear, perf benefit in our automated benchmarking. Locally, under stress, I'm seeing even better perf benefits to this change. On the order of 2 - 3 frames (32 - 48ms) during open or close actions, and I expect to see that further scale up as this is leveraged in busier and busier code paths.

## Related issue(s)

- fixes #1910

## Motivation and context
Faster is better!

## How has this been tested?

-   [ ] _Test case 1_
    1. Go [here](https://modal-overay-refactor--spectrum-web-components.netlify.app/)
    2. Open the "Performance" tab in DevTools
    3. Start recording
    4. Open and close a theme Picker 2 or 3 times
-   [ ] _Test case 2_
    1. Go [here](https://opensource.adobe.com/spectrum-web-components/)
    2. Open the "Performance" tab in DevTools
    3. Start recording
    4. Open and close a theme Picker 2 or 3 times
-   [ ] _Test case 3_
    1. Compare the perf results of the above. Generally I get a 4ms, 14ms and 11ms series of tasks in the new code and a 5ms, 40ms, and 15ms series of tasks in the old code.

image
vs
image

## Types of changes
-   [x] Refactor

## Checklist

### DIFF
--- a/packages/picker/src/Picker.ts
+++ b/packages/picker/src/Picker.ts
@@ -16,6 +16,7 @@ import {
     html,
     nothing,
     PropertyValues,
+    render,
     SizedMixin,
     TemplateResult,
 } from '@spectrum-web-components/base';
@@ -132,7 +133,6 @@ export class PickerBase extends SizedMixin(Focusable) {
 
     private closeOverlay?: () => void;
 
-    @query('sp-popover')
     private popover!: Popover;
 
     protected listRole: 'listbox' | 'menu' = 'listbox';
@@ -265,11 +265,29 @@ export class PickerBase extends SizedMixin(Focusable) {
         this.menuStateResolver();
     }
 
+    private popoverFragment!: DocumentFragment;
+
+    private generatePopover(deprecatedMenu: Menu | null): void {
+        if (this.popoverFragment) return;
+
+        this.popoverFragment = document.createDocumentFragment();
+        render(this.renderPopover, this.popoverFragment, { host: this });
+        this.popover = this.popoverFragment.children[0] as Popover;
+        this.optionsMenu = this.popover.children[1] as Menu;
+
+        if (deprecatedMenu) {
+            console.warn(
+                `Deprecation Notice: You no longer need to provide an sp-menu child to ${this.tagName.toLowerCase()}. Any styling or attributes on the sp-menu will be ignored.`
+            );
+        }
+    }
+
     private async openMenu(): Promise<void> {
         /* c8 ignore next 9 */
         let reparentableChildren: Element[] = [];
-
         const deprecatedMenu = this.querySelector('sp-menu');
+
+        this.generatePopover(deprecatedMenu);
         if (deprecatedMenu) {
             reparentableChildren = Array.from(deprecatedMenu.children);
         } else {
@@ -296,7 +314,6 @@ export class PickerBase extends SizedMixin(Focusable) {
         });
 
         this.sizePopover(this.popover);
-        const { popover } = this;
         this.addEventListener(
             'sp-opened',
             async () => {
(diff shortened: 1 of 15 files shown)


packages/picker/src/Picker.ts
```diff
@@ -265,11 +265,29 @@ export class PickerBase extends SizedMixin(Focusable) {
...
+    private generatePopover(deprecatedMenu: Menu | null): void {
+        if (this.popoverFragment) return;
+
+        this.popoverFragment = document.createDocumentFragment();
+        render(this.renderPopover, this.popoverFragment, { host: this });
+        this.popover = this.popoverFragment.children[0] as Popover;
+        this.optionsMenu = this.popover.children[1] as Menu;
+
+        if (deprecatedMenu) {
+            console.warn(
```
hunterloftis, line 279: 👍 👍

packages/picker/src/Picker.ts
```diff
@@ -412,7 +432,7 @@ export class PickerBase extends SizedMixin(Focusable) {
 
     protected render(): TemplateResult {
         return html`
-            ${this.renderButton} ${this.renderPopover}
+            ${this.renderButton}
```
hunterloftis, line 435: Would it make sense to get rid of this function and rename `renderButton` to `render`?

packages/picker/src/Picker.ts
```diff
@@ -460,6 +480,7 @@ export class PickerBase extends SizedMixin(Focusable) {
     protected updateMenuItems(
         event?: MenuItemAddedOrUpdatedEvent | MenuItemRemovedEvent
     ): void {
+        if (this.open && event && event.type === 'sp-menu-item-removed') return;
```
hunterloftis, line 483:
```suggestion
        if (this.open && event?.type === 'sp-menu-item-removed') return;
```

test/benchmark/helpers.ts
```diff
@@ -13,7 +13,7 @@ governing permissions and limitations under the License.
 import { html, LitElement, render, TemplateResult } from 'lit';
 import { customElement, property } from 'lit/decorators.js';
 import '@spectrum-web-components/theme/sp-theme.js';
-import '@spectrum-web-components/theme/scale-medium.js';
+import '@spectrum-web-components/theme/scale-large.js';
```
hunterloftis, line 16: Just curious - why switch to large scale for benchmarking?

hunterloftis (dismissed):

test/benchmark/helpers.ts
Westbrook: It's not so much a switch as a correction. https://github.com/adobe/spectrum-web-components/pull/1971/files#diff-737633c0f2c82bcb5bc7453285cb9547b8d34ac0a4aae636c7ca02410ebb12d2R126 I could fix this separate from this work, but in debugging I found that they didn't match and wanted them to in order to better understand the test results.

hunterloftis: Gotcha! 👍

PR #1982, opened 2021-12-09
fix(overlay): reduce DOM and use of "display: contents" for simplicity and accessibility

Westbrook:
## Description
We'd originally moved to wrapped slots in a number of places due to support with polyfills, but as we don't specifically support polyfilled delivery anymore we can move away from this practice, especially when there are bugs that come out of doing so.

As per https://bugs.chromium.org/p/chromium/issues/detail?id=1278446 the use of `display: contents` here seems to confusing the accessibility tree starting in Chrome 96, this removed the need for such usage by pushing the events down onto the `<slot>`.

## Motivation and context
Accessibility. Also running perf tests to see if this counts as more performant.

## How has this been tested?

-   [ ] _Test case 1_
    1. Go [here](https://a11y-overlay-trigger--spectrum-web-components.netlify.app/storybook/index.html?path=/story/action-group-tooltips--selects-single)
    2. Interact the the buttons in the demo with the screen reader
    3. See that in Voice Over the buttons are visible with the VO cursor
    4. See that in VO and NVDA that the buttons register as "radio" buttons that can be selected or unselected.

## Types of changes

## Checklist

### DIFF
--- a/packages/overlay/src/OverlayTrigger.ts
+++ b/packages/overlay/src/OverlayTrigger.ts
@@ -96,7 +96,7 @@ export class OverlayTrigger extends LitElement {
         // Keyboard event availability documented in README.md
         /* eslint-disable lit-a11y/click-events-have-key-events */
         return html`
-            <div
+            <slot
                 id="trigger"
                 @click=${this.onTrigger}
                 @longpress=${this.onTrigger}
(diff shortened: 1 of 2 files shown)


Westbrook:
The underlying issue here has been corrected in Chrome 97, which should ship as stable today or tomorrow. While I like the reduction of DOM here, there's no clear benefit to do so, expect for screen reader users in Chrome 96. Happy to close this as "won't fix" if we choose.

hunterloftis (approved):

PR #1984, opened 2021-12-10
fix: don't expose global DOM type definitions

davidaurelio:
## Description

Removes public global TypeScript definitions of DOM / Web Platform features.

## Motivation and context

Spectrum Web Components are using some extensions to global interfaces,
in particular `DomElement` and `ShadowRoot`, to declare members that at
some point were not covered by the definitions shipping with Typescript.

In consequence, consumers of SWC may face issues when upgrading their
version of Typescript when the respective definitions differ and can’t
be merged automatically. An example error message is:

```
node_modules/typescript/lib/lib.dom.d.ts:4256:11 -
  error TS2430: Interface 'Document' incorrectly extends
      interface 'FontFaceSource'.
  Types of property 'fonts' are incompatible.
    Type '{ ready: Promise<void>; } | undefined' is not assignable to
        type 'FontFaceSet'.
      Type 'undefined' is not assignable to type 'FontFaceSet'.
```

This kind of error message is difficult to silence, or even impossible
if SWC are a transitive dependency.

As a solution, we propose to use _local_ definitions for the Web
Platform features that single components are using. None of these
definitions end up in `.d.ts` files. Consumers end up with un-augmented
type definitions, while component implementations get the definitions
they need.

## How has this been tested?

- [x] Run `tsc`
   1. `yarn build:ts:clean`
- [x] Verify Build Output
   1. manually inspected `.d.ts` files to verify that the newly defined types are not exported.

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/overlay/src/ActiveOverlay.ts
+++ b/packages/overlay/src/ActiveOverlay.ts
@@ -41,13 +41,11 @@ export interface PositionResult {
     positionTop: number;
 }
 
-declare global {
-    interface Document {
-        fonts?: {
-            ready: Promise<void>;
-        };
-    }
-}
+type DocumentWithFonts = typeof document & {
+    fonts?: {
+        ready: Promise<void>;
+    };
+};
 
 type OverlayStateType =
     | 'idle'
@@ -426,7 +424,8 @@ export class ActiveOverlay extends SpectrumElement {
     }
 
     public async updateOverlayPosition(): Promise<void> {
-        await (document.fonts ? document.fonts.ready : Promise.resolve());
+        const doc = document as DocumentWithFonts;
+        await (doc.fonts ? doc.fonts.ready : Promise.resolve());
         if (this.popper) {
             await this.popper.update();
         }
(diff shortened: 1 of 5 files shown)


Westbrook:
Thanks for getting into this, super helpful! Feel like there's always much to learn when it comes to types like this.

Definitely want to get this cleared up for you, but wanted your thoughts on this versus #1985. I wanted to make sure it worked before I asked you about it, but there I update Typescript so the `document.fonts` types are "just available" and follow your suggestion on `ShadowRoot.adoptedStyleSheets` with a small change in where the type is applied.

With your insight here, I'd happy to get one of these into the library and out in a release ASAP!

davidaurelio:
@Westbrook #1985 would absolutely work, too, and creates less technical debt with needless utility types. To me it only matters that no global type definitions leak out of SWC that can potentially conflict with TS’s builtins.

Insofar I would go with your approach. It’s simpler and does the job.

Westbrook:
Corrected via #1985

PR #2001, opened 2021-12-17
fix(tabs): error on click - undefined tab target

### LINKED ISSUE #2000: [Bug]: sp-tabs error on click
### Code of conduct

### Impacted component(s)

sp-tabs

### Expected behavior

A click inside sp-tabs but NOT on a sp-tab should not throw

### Actual behavior

Throws error because sp-tab target is undefined

### Screenshots

_No response_

_No response_

1. Go to https://opensource.adobe.com/spectrum-web-components/components/tabs/
2. Click inside sp-tabs but NOT on a sp-tab
3. Check console
5. See error

_No response_

_No response_

sorinsarca:
## Description

A click inside sp-tabs throws if outside of a sp-tab.

## Related issue(s)

- fixes #2000 

## Motivation and context

Bugfix

## How has this been tested?

Manually

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/tabs/src/Tabs.ts
+++ b/packages/tabs/src/Tabs.ts
@@ -269,10 +269,13 @@ export class Tabs extends SizedMixin(Focusable) {
     }
 
     private onClick = (event: Event): void => {
+        if (this.disabled) {
+            return;
+        }
         const target = event
             .composedPath()
             .find((el) => (el as Tab).parentElement === this) as Tab;
-        if (this.disabled || target.disabled) {
+        if (!target || target.disabled) {
             return;
         }
         this.shouldAnimate = true;
--- a/packages/tabs/test/tabs.test.ts
+++ b/packages/tabs/test/tabs.test.ts
@@ -656,4 +656,20 @@ describe('Tabs', () => {
         expect(initialWidth).to.be.greaterThan(shorterWidth);
         expect(longerWidth).to.be.greaterThan(shorterWidth);
     });
+    it('clicks on #list do not throw', async () => {
+        const tabs = await createTabs();
+        const tabList = (tabs.shadowRoot as ShadowRoot).querySelector(
+            '#list'
+        ) as HTMLDivElement;
+        // exceptions thrown in event listeners do not propagate to caller
+        // we must catch them with window.onerror
+        let hasError = false;
+        const oldOnerror = window.onerror;
+        window.onerror = () => {
+            hasError = true;
+        };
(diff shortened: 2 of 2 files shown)


Westbrook (commented):
Happy New Year! Thanks for the submission. Happy to get this into the library ASAP. To support that, would you be able to add a test for this? 🙏🏼

sorinsarca:
Happy New Year!
To be honest, I don't know how to add a test in this case, mostly due to this:

> Warning: Exceptions thrown by event handlers are reported as uncaught exceptions. The event handlers run on a nested callstack; they block the caller until they complete, but exceptions do not propagate to the caller.

https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/dispatchEvent


Here is a test example

```ts
it('should not throw on click', async () => {
    const tabs = await createTabs();
    const tabList = (tabs.shadowRoot as ShadowRoot).querySelector('#list') as HTMLDivElement;

    expect(() => {
        // dispatchEvent() doesn't propagate the exception
        // same story with .click()
        tabList.dispatchEvent(new MouseEvent('click'));
    }).to.not.throw();
});
```

Westbrook:
Your example test fails appropriately. How's it looking post the fix? There are certainly more complex recreations we could discuss (e.g. leveraging `sendMouse` or similar for a real pointer interaction), but if that works on both sides of the test, I'd be happy to approve.

sorinsarca:
The problem is that it also fails if you replace `.to.not.throw()` with `.to.throw()`

After the fix there is no exception thrown, so the test passes - but it has zero value because it doesn't really test anything.
Maybe the testing framework provides some tools to "catch" errors thrown by event listeners, but I'm not aware of any.

Westbrook:
Why would you want to use `expect().to.throw()`? It shouldn't throw, right?

The fact that the test fails in the old code and passes (presumably) with the new code, is the clarity that I think we want here while also adding coverage to see regressions if they were to arise.

Conversely, I could see this change being applied differently. If the `@click=${this.onClick}` binding were moved down from the `div` wrapper and down to the `slot` we should be able to get the same functionality without being able to trigger it on the "tab list" element, which would prevent the need to test it. Would need to confirm that the fix works as expected, but it would not require a test if it did indeed work.

sorinsarca:
I've added the proposed test.
Too bad the testing framework doesn't have something to handle these cases.

Westbrook (approved):
LGTM! Thanks for catching this and putting up with the less than optimal test case to ensure it doesn't come up again. 🙏🏼

PR #2017, opened 2022-01-04
Refactor Roving Tab Index Implementations

### LINKED ISSUE #1999: [Bug]: Accordion items in docs are not opening
### Code of conduct

### Impacted component(s)

Accordion

### Expected behavior

Clicking on one of the headings in the first example should open the corresponding item. The second example (Allow Multiple) works fine.

### Actual behavior

Clicking on headings does nothing.

### Screenshots

accordion

Firefox, Chrome, Microsoft Edge

1. Go to https://opensource.adobe.com/spectrum-web-components/components/accordion/#example
2. Scroll to "Example"
3. Click on the heading
4. See that nothing happens

_No response_

_No response_

Westbrook:
## Description
- create the `@spectrum-web-components/reactive-controllers` package. Does that sound OK?
- vend a Roving Tab Index Reactive Controller
- refactor Roving Tab Index usage to leverage the Reactive Controller
- remove `PageUp` and `PageDown` functionality from `sp-action-group` and `sp-radio-group` as it's not actually in the w3c or Spectrum specs for those patterns.

## Related issue(s)

- fixes #1999

## Motivation and context
Centralize, abstract, make less.

## How has this been tested?

-   [ ] _Test case 1_
    1. Go [here](https://roving-tab-index--spectrum-web-components.netlify.app/storybook/index.html?path=/story/accordion--default)
    2. Click an Accordion Item
    3. Use `ArrowUp`/`ArrowDown` to navigate to another Item and `Enter`/`Space` to toggle that item
    4. Use `Home`/`End` to jump to the first/last Item
-   [ ] _Test case 2_
    1. Go [here](https://roving-tab-index--spectrum-web-components.netlify.app/storybook/index.html?path=/story/action-group--default)
    2. Click an Action Button
    3. Use `ArrowUp`/`ArrowDown`/`ArrowLeft`/`ArrowRight` to navigate to another Button
    4. Use `Home`/`End` to jump to the first/last Button
-   [ ] _Test case 3_
    1. Go [here](https://roving-tab-index--spectrum-web-components.netlify.app/storybook/index.html?path=/story/radio--group-example)
    2. Click an Radio Button
    3. Use `ArrowUp`/`ArrowDown`/`ArrowLeft`/`ArrowRight` to navigate to AND select another Button
    4. Use `Home`/`End` to jump to AND select the first/last Button
-   [ ] _Test case 4_
    1. Go [here](https://roving-tab-index--spectrum-web-components.netlify.app/storybook/index.html?path=/story/sidenav--default): unmanaged tab index
    2. Click a Sidenav Item
    3. Use `Tab` to navigate through the various Sidenav Items
    4. Go [here](https://roving-tab-index--spectrum-web-components.netlify.app/storybook/index.html?path=/story/sidenav--manage-tab-index): managed index
    5. Click a Sidenav Item
    6. Use `ArrowUp`/`ArrowDown` to navigate to another Sidenav Item and `Enter` to toggle that item
    7. Use `Home`/`End` to jump to the first/last Item
-   [ ] _Test case 5_
    1. Go [here](https://roving-tab-index--spectrum-web-components.netlify.app/storybook/index.html?path=/story/tabs--default): horizontal
    2. Click a Tab Item
    3. Use `ArrowLeft`/`ArrowRight` to navigate to another Tab and `Enter`/`Space` to toggle that item
    4. Use `Home`/`End` to jump to the first/last Item
    5. Go [here](https://roving-tab-index--spectrum-web-components.netlify.app/storybook/index.html?path=/story/tabs--vertical): vertical
    6. Use `ArrowUp`/`ArrowDown` to navigate to another Tab and `Enter`/`Space` to toggle that item
    7. Use `Home`/`End` to jump to the first/last Item
-   [ ] _Test case 6_
    1. Go [here](https://roving-tab-index--spectrum-web-components.netlify.app/storybook/index.html?path=/story/tags--deletable)
    2. Click a Tag Item
    3. Use `ArrowUp`/`ArrowDown`/`ArrowLeft`/`ArrowRight` to navigate to another Tag and `Delete`/`Backspace`/`Space` to "delete" that item, show in the Actions pane of Storybook.
    4. Use `Home`/`End` to jump to the first/last Item
-   [ ] _Test case 7_
    1. [x] Integration testing required!

## Types of changes

## Checklist

### DIFF
--- a/tools/reactive-controllers/roving-tab-index.md
+++ b/tools/reactive-controllers/roving-tab-index.md
@@ -1,0 +1,92 @@
+## Description
+
+[Roving tabindex](https://www.w3.org/TR/wai-aria-practices-1.2/#kbd_roving_tabindex) in a pattern whereby multiple focusable elements are represented by a single `tabindex=0` element, while the individual elements maintain `tabindex=-1` and are made accessible via arrow keys after the entry element if focused. This allows keyboard users to quickly tab through a page without having to stop on every element in a large collection. Attaching a `RovingTabindexController` to your custom element will manage the supplied `elements` via this pattern.
+
+### Usage
+
+[![See it on NPM!](https://img.shields.io/npm/v/@spectrum-web-components/reactive-controllers?style=for-the-badge)](https://www.npmjs.com/package/@spectrum-web-components/reactive-controllers)
+[![How big is this package in your project?](https://img.shields.io/bundlephobia/minzip/@spectrum-web-components/reactive-controllers?style=for-the-badge)](https://bundlephobia.com/result?p=@spectrum-web-components/reactive-controllers)
+
+```
+yarn add @spectrum-web-components/reactive-controllers
+```
+
+Import the `RovingTabindexController` via:
+
+```
+import { RovingTabindexController } from '@spectrum-web-components/reactive-controllers/RovingTabindex.js';
+```
+
+## Example
+
+A `Container` element that manages a collection of `<sp-button>` elements that are slotted into it from outside might look like the following:
+
+```js
+import { html, LitElement } from 'lit';
+import { RovingTabindexController } from '@spectrum-web-components/reactive-controllers/RovingTabindex.js';
+import type { Button } from '@spectrum-web-components/button';
+
+class Container extends LitElement {
+    rovingTabindexController =
+        new RovingTabindexController() <
+        Button >
+        (this,
+        {
+            elements: () => [...this.querySelectorAll('sp-button')],
+        });
+
+    render() {
+        return html`
+            <slot></slot>
+        `;
+    }
+}
+```
+
+The above will default to entering the `Container` element via the first `<sp-button>` element every time while making all slotted `<sp-button>` elements accessible via the the arrow key (`ArrowLeft`, `ArrowRight`, `ArrowUp`, and `ArrowDown`) managed tab order.
+
+## Options
+
+A `Container` can further customize the implementation of the `RovingTabindexController` with the following options:
+
+-   `direction` to customize how and which arrow keys manage what element is to be focused and accepts a either a string of `both`, `vertical`, `horizontal`, or `grid` or a method returning one of those strings
+-   `elementEnterAction` enacts actions other than `focus` on the entered element which accepts a method with a signature of `(el: T) => void`
+-   `elements` provides the elements that will have their `tabindex` managed via a method with a signature of `() => T[]`
+-   `focusInIndex` to control what element will recieve `tabindex=0` while focus is outside of the `Container` and accepts a method with a signature of `(_elements: T[]) => number`
+-   `isFocusableElement` describes the state an element much be in to receive `focus` via a method with a signature of `(el: T) => boolean`
+-   `listenerScope` outlines which parts on a container's DOM when listening for arrow key presses via an element reference or a method returning an element reference with the signature `() => HTMLElement`
+
+## Advanced usage
+
+These options can be combined to form various interfaces from the more default that we saw above to the very complex. Below is another `Container` that manages slotted `<sp-button>` elements via the `RovingTabindexController`. The options provided ensure:
+
+-   the first focused `<sp-button>` is the one `selected` by the container
+-   the elements are only focused via the `ArrowLeft` and `ArrowRight` keys
+-   when an element is focused it becomes the `selected` element
+-   only enabled elements are focusable
+
+```js
+import { html, LitElement } from 'lit';
+import { RovingTabindexController } from '@spectrum-web-components/reactive-controllers/RovingTabindex.js';
+import type { Button } from '@spectrum-web-components/button';
+
+class Container extends LitElement {
+    rovingTabindexController = new RovingTabindexController<Button>(this, {
+        focusInIndex: (buttons) => return this.selected
+            ? buttons.indexOf(this.selected)
+            : 0,
+        direction: 'horizontal',
+        elementEnterAction: (button) => this.selected = button,
+        elements: () => [...this.querySelectorAll('sp-button')],
+        isFocusableElement: (button) => !button.disabled,
+    });
+
+    selected!: Button;
+
+    render() {
+        return html`<slot></slot>`;
+    }
+}
+```
+
+The above usage is very close to what can be seen in the [`<sp-radio-group>` element](../components/radio).
--- a/tools/reactive-controllers/src/RovingTabindex.ts
+++ b/tools/reactive-controllers/src/RovingTabindex.ts
@@ -1,0 +1,327 @@
+/*
+Copyright 2020 Adobe. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License. You may obtain a copy
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+import type { ReactiveController, ReactiveElement } from 'lit';
+
+type DirectionTypes = 'horizontal' | 'vertical' | 'both' | 'grid';
+export type RovingTabindexConfig<T> = {
+    focusInIndex?: (_elements: T[]) => number;
+    direction?: DirectionTypes | (() => DirectionTypes);
+    elementEnterAction?: (el: T) => void;
+    elements: () => T[];
+    isFocusableElement?: (el: T) => boolean;
+    listenerScope?: HTMLElement | (() => HTMLElement);
+};
+interface UpdateTabIndexes {
+    tabIndex: number;
+    removeTabIndex?: boolean;
+}
+
+export class RovingTabindexController<T extends HTMLElement>
+    implements ReactiveController
+{
+    private cachedElements?: T[];
+
+    get currentIndex(): number {
+        if (this._currentIndex === -1) {
+            this._currentIndex = this.focusInIndex;
+        }
+        return this._currentIndex;
+    }
+
+    set currentIndex(currentIndex) {
+        this._currentIndex = currentIndex;
+    }
+
+    private _currentIndex = -1;
+
+    get direction(): DirectionTypes {
+        return this._direction();
+    }
+
+    _direction = (): DirectionTypes => 'both';
+
+    directionLength = 5;
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    elementEnterAction = (_el: T): void => {
+        return;
+    };
+
+    get elements(): T[] {
+        if (!this.cachedElements) {
+            this.cachedElements = this._elements();
+        }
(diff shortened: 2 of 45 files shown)


hunterloftis:
In test case 2, is this expected behavior? It seems like the first element may be being treated differently from the other elements.

1. Click on middle button
2. Press [right arrow]
3. Focus moves to right button
4. Click on left button
5. Press [right arrow]
6. Focus moves to left button
7. (repeat with [down arrow])

https://user-images.githubusercontent.com/364501/150224302-35081e88-617f-4140-845d-1270b067f1b3.mov

hunterloftis:
In test case 5, arrow keys & home/end don't produce any visible change for me (neither does space). Tab & enter works as expected though:

https://user-images.githubusercontent.com/364501/150229446-d7ee8bfa-4f90-45db-ae7d-3e7eadf01637.mov

hunterloftis:
Test case 6 works as expected for tags, although space/enter doesn't toggle anything (I wouldn't expect it to - just making sure I understand the test instructions)

Westbrook:
@hunterloftis thanks for getting started here, updates as follows.

Test case 2:
- corrected  multiple uses of `0 || other` fall through in logic to `??` or other more reliable logic
- allowed `focusin` based index management while focus is held

Test case 4:
- linked to the wrong story; Sidenav has both managed and unmanaged versions
- prevented the unamanaged version from accidentally becoming managed so that all elements are `tabindex=0` to start

Test case 6:
- added `delete` action to Storybook so that the `Delete`/`Backspace`/`Space`  key will provide visible reaction

Westbrook:
Writing test cases in English is a very strange experience 🙃 I've expanded that Tabs context to include a link to the _vertical_ instance as well. Horizontal should accept `ArrowLeft` and `ArrowRight` while vertical accepts `ArrowUp` and `ArrowDown`...

tools/reactive-controllers/src/RovingTabindex.ts
```diff
@@ -0,0 +1,304 @@
...
+        const { length } = this.elements;
+        let nextIndex = (length + this.currentIndex + diff) % length;
+        while (
+            this.elements[nextIndex] &&
+            !this.isFocusableElement(this.elements[nextIndex])
+        ) {
+            nextIndex = (length + nextIndex + diff) % length;
+        }
+        this.currentIndex = nextIndex;
+    }
```
hunterloftis, line 137: `while`s always make me start thinking defensively about the potential for infinite loops.

What it ever be possible to enter this function with 10 elements that test negative for `isFocusableElement` (either by being disabled, or by the user providing a bad `isFocusableElement` like `return false`)?

To be extra safe, I might limit iterations to `length`, since we never want to loop in more than a circle:

```ts
for (let i = 0; i < length; i++) {
  const index = (length + i + diff) % length;
  if (this.elements[index] && this.isFocusableElement(this.elements[index])) {
    this.currentIndex = index;
    return;
  }
}
```

Less dangerous: could length ever be 0? (x % 0 = NaN)

tools/reactive-controllers/src/RovingTabindex.ts
```diff
@@ -0,0 +1,304 @@
...
+
+    hostContainsFocus(): void {
+        this.host.addEventListener('focusout', this.handleFocusout);
+        this.host.addEventListener('keydown', this.handleKeydown);
+        this.focused = true;
+        const currentElement = this.elements[this.currentIndex];
+        if (!currentElement) return;
```
hunterloftis, line 144: What is the purpose of this if statement?

tools/reactive-controllers/src/RovingTabindex.ts
```diff
@@ -0,0 +1,304 @@
...
+    }
+
+    hostNoLongerContainsFocus(): void {
+        this.host.addEventListener('focusin', this.handleFocusin);
+        this.host.removeEventListener('focusout', this.handleFocusout);
+        this.host.removeEventListener('keydown', this.handleKeydown);
+        this.currentIndex = this.focusInIndex;
+        this.focused = false;
+        const currentElement = this.elements[this.currentIndex];
+        if (!currentElement) return;
```
hunterloftis, line 154: Ditto here

tools/reactive-controllers/src/RovingTabindex.ts
```diff
@@ -0,0 +1,304 @@
...
+    }
+
+    handleFocusin = (event: FocusEvent): void => {
+        if (!this.isEventWithinListenerScope(event)) return;
+        if (this.isRelatedTargetAnElement(event)) {
+            this.hostContainsFocus();
+        }
+        const path = event.composedPath() as T[];
+        let targetIndex: number | undefined;
+        path.find((el) => {
```
hunterloftis, line 169: maybe another opportunity for `.findIndex`

tools/reactive-controllers/src/RovingTabindex.ts
```diff
@@ -0,0 +1,304 @@
...
+            targetIndex = this.elements.indexOf(el);
+            return targetIndex !== -1;
+        });
+        this.currentIndex = targetIndex ?? this.currentIndex;
```
hunterloftis, line 173: This may be doing exactly what you want as right now `targetIndex` can be `undefined`, but since indices are typically invalid at `-1` and valid at `0`, I'd recommend:

```suggestion
        this.currentIndex = targetIndex > -1 ? targetIndex : this.currentIndex;
```

tools/reactive-controllers/src/RovingTabindex.ts
```diff
@@ -0,0 +1,304 @@
...
+        switch (this.direction) {
+            case 'horizontal':
+                return code === 'ArrowLeft' || code === 'ArrowRight';
+            case 'vertical':
+                return code === 'ArrowUp' || code === 'ArrowDown';
+            case 'both':
+            case 'grid':
+                return code.startsWith('Arrow');
+        }
+    }
```
hunterloftis, line 195: I am guaranteed to typo these strings someday if we don't make them constants 😅

tools/reactive-controllers/src/RovingTabindex.ts
```diff
@@ -0,0 +1,304 @@
...
+                this.currentIndex = 0;
+                diff -= 1;
+                event.preventDefault();
+                break;
+            case 'Home':
+                this.currentIndex = this.elements.length - 1;
+                diff += 1;
+                event.preventDefault();
+                break;
+        }
```
hunterloftis, line 229: Would this accomplish the same thing, or is there a reason not to bail if none of our codes match?

```suggestion
        switch (event.code) {
            case 'ArrowRight':
                diff += 1;
                break;
            case 'ArrowDown':
                diff += this.direction === 'grid' ? this.directionLength : 1;
                break;
            case 'ArrowLeft':
                diff -= 1;
                break;
            case 'ArrowUp':
                diff -= this.direction === 'grid' ? this.directionLength : 1;
                break;
            case 'End':
                this.currentIndex = 0;
                diff -= 1;
                break;
            case 'Home':
                this.currentIndex = this.elements.length - 1;
                diff += 1;
                break;
            default:
                return;
        }
        event.preventDefault();
```

tools/reactive-controllers/src/RovingTabindex.ts
```diff
@@ -0,0 +1,304 @@
...
+        } else {
+            this.updateTabindexes((el: HTMLElement): number | false => {
+                const containsFocusInElement = el.contains(this.focusInElement);
+                if (el === this.focusInElement) {
+                    return 0;
+                } else if (containsFocusInElement) {
+                    return false;
+                }
+                return -1;
+            });
```
hunterloftis, line 249: This is a weird function signature (0, -1, or `false`). I don't think this is necessarily better, but wanted to think of an alternative:

```suggestion
            this.updateTabindexes((el: HTMLElement): number | undefined => {
                if (el === this.focusInElement) return 0;
                return el.contains(this.focusInElement) ? undefined : -1;
            });
```

Not really an improvement to `updateTabindexes` either; maybe we can simplify these later instead:

```ts
    updateTabindexes(getTabIndex: (el: HTMLElement) => number | undefined): void {
        this.elements.forEach((el) => {
            const tabIndex = getTabIndex(el);
            if ([0, -1].includes(tabIndex)) {
                el.tabIndex = tabIndex;
                return;
            }
            el.removeAttribute('tabindex');
            const updatable = el as unknown as {
                requestUpdate?: () => void;
            };
            if (updatable.requestUpdate) updatable.requestUpdate();
        });
    }
```

hunterloftis (approved):
image

I left a bunch of feedback to consider, but all the tests work as described now & there's nothing blocking here.

tools/reactive-controllers/src/RovingTabindex.ts
```diff
@@ -0,0 +1,304 @@
...
+    ) {
+        this.host = host;
+        this.host.addController(this);
+        this._elements = elements;
+        this.isFocusableElement = isFocusableElement || this.isFocusableElement;
+        if (typeof direction === 'string') {
+            this._direction = () => direction;
+        } else if (typeof direction === 'function') {
+            this._direction = direction;
+        }
```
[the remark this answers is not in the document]
Westbrook: `direction` isn't required in the `constructor` and defaults to `_direction` defaults to `() => 'both'` in the class.

tools/reactive-controllers/src/RovingTabindex.ts
```diff
@@ -0,0 +1,304 @@
...
+        this.calculateFocusInIndex =
+            calculateFocusInIndex || this.calculateFocusInIndex;
+        this.elementEnterAction = elementEnterAction || this.elementEnterAction;
```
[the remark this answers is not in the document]
Westbrook: This looks like it forces the `undefined` into this? Am I missing something?

tools/reactive-controllers/roving-tab-index.md
```diff
@@ -0,0 +1,92 @@
...
+```
+
+The above will default to entering the `Container` element via the first `<sp-button>` element every time while making all slotted `<sp-button>` elements accessible via the the arrow key (`ArrowLeft`, `ArrowRight`, `ArrowUp`, and `ArrowDown`) managed tab order.
+
+## Options
+
+A `Container` can further customize the implementation of the `RovingTabindexController` with the following options:
+
+-   `calculateFocusInIndex` to control what element will recieve `tabindex=0` while focus is outside of the `Container` and accepts a method with a signature of `(_elements: T[]) => number`
+-   `direction` to customize how and which arrow keys manage what element is to be focused and accepts a either a string of `both`, `vertical`, `horizontal`, or `grid` or a method returning one of those strings
```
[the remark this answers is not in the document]
Westbrook: Unimplemented..._but_ while "both" moves linear along the list regardless of input, "grid" will visually follow the layout of the elements, so `ArrowUp`/`ArrowDown` will "skip" the length of the row.

Westbrook (commented):
@hloftis made some changes on your great suggestions. Very useful! Let me know if this still looks good to you.

tools/reactive-controllers/src/RovingTabindex.ts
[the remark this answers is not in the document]
hunterloftis: 🤦  yep you're right, I was missing something!

hunterloftis (approved):

Issue #2035, opened 2022-01-07
[Bug]: sp-slider only adds one sp-slider-handle when multiple are defined


peterjanes:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-slider, sp-slider-handle

### Expected behavior

Adding an `sp-slider` with multiple handles from an HTML `template` should add all of the handles.

### Actual behavior

I haven't nailed down the exact conditions, but sometimes when adding `sp-slider`s dynamically from an HTML `template` only one instance of `sp-slider-handle` will be added because the handle map key is `undefined`. (I finally tracked down the behaviour when I started getting "First slider handle cannot have attribute min='previous'" after adding `max="next"` to the first `sp-slider-handle` and `min="previous"` to the second.)

### What browsers are you seeing the problem in?

Firefox, Chrome

### How can we reproduce this issue?

I haven't been able to come up with a reduced test case, but the issue is 100% reproducible in my (currently non-public) app and seems to be due to some instances of a field not being renamed. (`handle.name` is used instead of `handle.handleName` in `HandleController.ts`.) I'll create an associated pull request for review.

Westbrook:
@peterjanes thanks for sharing this issue!

Before we can fully review changes in support of what you're experiencing, we will need a reproduction case so that we can compare your findings with ours and ensure that all users of this element are being appropriately served. I'd suggest you start from something like https://webcomponents.dev/edit/collection/fO75441E1Q5ZlI0e9pgq/U7LQv7LsAVBwJayJXG3B/src/index.ts and see if you can isolate the issue you are experiencing.

peterjanes:
Yeah, I've been working on a reproduce case for a couple of days but can't get it to fail. Figured I'd submit the PR that fixes it (or appears to) as it seems like a pretty obvious update; `handle.name` is only set when `handle.handleName` isn't set (and only used in a log message), and the `undefined` `handle.handleName` is immediately used as the map key.

peterjanes:
Here's my current non-failing reproduce case, which mirrors what the live code does:

```html
<!DOCTYPE html>
<script src="https://jspm.dev/@spectrum-web-components/bundle/elements.js" type="module"></script>

<div>
<sp-theme scale="medium" color="light">
</sp-theme>
</div>

<template id="rangeTemplate">
  <sp-slider variant="range" step="1">
[...]

PR #2036, opened 2022-01-07
Use handleName field consistently in HandleController

peterjanes:
## Description

There are two instances of `handle.name` that should be `handle.handleName` so that the handle map is populated correctly.

## Related issue(s)

- #2035

## Motivation and context

I haven't nailed down the exact conditions, but sometimes when adding `sp-slider`s dynamically from an HTML template only one instance of `sp-slider-handle` will be added because the handle map key is undefined. (I finally tracked down the behaviour when I started getting `First slider handle cannot have attribute min="previous"` after adding `max="next"` to the first `sp-slider-handle` and `min="previous"` to the second.)

## How has this been tested?

I'm afraid I don't have a reduced test case, but I was able to track down the behaviour in my (currently non-public) app. The code does something like this (which doesn't reproduce the issue in isolation, but is fixed by the change in this PR):

```js
const sliders = document.createElement('div');
arr.forEach(parms => {
  const el = sliderTemplate.content.cloneNode(true);
  const slider = el.querySelector('sp-slider');
  slider.setAttribute('label', parms.description);
  slider.querySelector('[name=min]').setAttribute('value', parms.min);
  slider.querySelector('[name=max]').setAttribute('value', parms.max);
  sliders.appendChild(slider);
});
document.body.appendChild(sliders);
```

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/slider/src/HandleController.ts
+++ b/packages/slider/src/HandleController.ts
@@ -91,7 +91,7 @@ export class HandleController implements Controller {
             return input;
         }
         /* c8 ignore next 2 */
-        throw new Error(`No input for handle "${handle.name}"`);
+        throw new Error(`No input for handle "${handle.handleName}"`);
     }
 
     public requestUpdate(): void {
@@ -208,7 +208,7 @@ export class HandleController implements Controller {
         handles.forEach((handle, index) => {
             /* c8 ignore next */
             if (!handle.handleName?.length) {
-                handle.name = `handle${index + 1}`;
+                handle.handleName = `handle${index + 1}`;
             }
             this.handles.set(handle.handleName, handle);
             this.handleOrder.push(handle.handleName);


packages/slider/src/HandleController.ts
```diff
@@ -208,7 +208,7 @@ export class HandleController implements Controller {
         handles.forEach((handle, index) => {
             /* c8 ignore next */
             if (!handle.handleName?.length) {
-                handle.name = `handle${index + 1}`;
+                handle.handleName = `handle${index + 1}`;
```
Westbrook, line 211: This seems weird in that `handle.handleName` is _only_ a getter for `handle.name`...

Westbrook: https://github.com/adobe/spectrum-web-components/blob/d23f735db3cea01c2243e0485448d1598c6a8692/packages/slider/src/SliderHandle.ts#L80-L82

peterjanes: Should've said, my "fix" is wrong (in editing it while I was waiting for CLA approval I removed that minor detail) but I'm not sure why the handle's `name` property isn't being initialized properly.

Westbrook: Could you share into the issue more info around how you add the DOM to the page, etc.?

As this change isn't right, let's close this PR and discuss finding the source of this in the issue.

Westbrook:
Let's look more for a source before we review a fix.

Issue #2071, opened 2022-01-21
[Bug]: no visible focus on sp-top-nav-item


jnurthen:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-top-nav sp-top-nav-item

### Expected behavior

All components should have a visible focus when they receive focus

### Actual behavior

There is no visible focus

### What browsers are you seeing the problem in?

Firefox, Chrome, Microsoft Edge

### How can we reproduce this issue?

1. Go to https://opensource.adobe.com/spectrum-web-components/components/top-nav/
2. Tab through the page until you get to the top nav example
3. Note there is no visible focus

### Sample code that illustrates the problem

fonts.adobe.com

Westbrook:
That's awkward as it seems to work in _some_ contexts: https://opensource.adobe.com/spectrum-web-components/storybook/index.html?path=/story/top-nav--default This might be due to tricks of the `:focus-visible` polyfill that's hanging around the library for just a little longer. Thanks for the catch, we'll figure out what needs to change to guarantee this!

jnurthen:
@Westbrook didn't file yet but also saw some really weird behaviour with top-nav with Firefox

Westbrook:
Was it along the lines of it follow macOS rules and not tab stopping on `<a href="#">` elements, or something else?

PR #2073, opened 2022-01-22
feat: conditionally load focus-visible polyfill

Westbrook:
## Description
Test perf benefits of only loading the polyfill when needed.
- use dynamic import and capability test to prevent polyfill use in Chrome and Firefox
- ungated usage can cause edge case situations like in #2071 

## Motivation and context
Perf

## How has this been tested?

-   [ ] _Test case 1_
    1. Go [here](https://focus-visible--spectrum-web-components.netlify.app/)
    2. Navigate the site with the keyboard and see that focus visible states are triggered.

## Types of changes

## Checklist

### DIFF
--- a/packages/shared/src/focus-visible.ts
+++ b/packages/shared/src/focus-visible.ts
@@ -10,8 +10,6 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import 'focus-visible';
-
 declare global {
     interface Window {
         applyFocusVisiblePolyfill?: (scope: Document | ShadowRoot) => void;
@@ -40,6 +38,9 @@ try {
     document.body.querySelector(':focus-visible');
 } catch (error) {
     hasFocusVisible = false;
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    import('focus-visible');
 }
 
 /**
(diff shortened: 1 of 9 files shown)


benjamind:
Interesting how much this affects popover performance. I wonder if the polyfill was what was causing the excessive focus event interactions I was seeing on my FocusFollower experiments we were discussing yesterday?

Westbrook:
This includes `popover` in the perf tests only because `popover` depends of `overlay` which depends on `shared` where this code was changes, but doesn't include it itself... so the variance of `14.36ms - 17.44ms` for 100 instances of the `popover` element is likely more noise than anything else.

For the added focuses, I'd less think that was the polyfill that the handling for certain elements within the library. If the lack of the polyfill did remove them somehow, then winning!

packages/shared/src/focus-visible.ts
```diff
@@ -40,6 +38,9 @@ try {
     document.body.querySelector(':focus-visible');
 } catch (error) {
     hasFocusVisible = false;
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    import('focus-visible');
```
Westbrook, line 43: This is the guts of this PR. Everything else is test support for this change in browsers that do not require the polyfill, which is ALL of them now! 🥳

najikahalsema (approved):
yay!!!

PR #2082, opened 2022-01-28
feat(picker): support responsive delivery of menu

Westbrook:
## Description
Add support for delivering the options in a Picker, Action Menu, or Split Button in an `sp-tray` element when on "mobile".
- leverage `MatchMediaController` to centralize the usage/manage of `window.matchMedia` interactions
- choose between the `sp-popover` and `sp-tray` element as appropriate when opening the Menu element
  - currently DOES NOT change delivery while the Menu is open
- correct the open/close timing in Tray for use in an overlay
- surface additional style management in Menu

## Motivation and context
More complete coverage of the responsive delivery of the Spectrum spec.

## How has this been tested?

-   [ ] _Test case 1_
    1. Go [here](https://popover-tray--spectrum-web-components.netlify.app/)
    2. Size your browser at more that 700px wide
    3. Open a Picker element
    4. See it in a Popover
-   [ ] _Test case 2_
    1. Go [here](https://popover-tray--spectrum-web-components.netlify.app/)
    2. Size your browser at less that 700px wide
    3. Open a Picker element
    4. See it in a Popover
-   [ ] _Test case 3_
    1. Using a device that is less than 700px wide or tall and leverages a coarse pointer and does not allow hover, a phone.
    2. Go [here](https://popover-tray--spectrum-web-components.netlify.app/)
    3. Open a Picker element
    4. See it in a Popover

## Types of changes

## Checklist


davidyxu:
Couple minor things I noticed:

The tray vs popover doesn't switch forms if it's resized while open. This might be intended behavior but it's a little unclear? Feels not useful on desktop when shrinking your browser size, but might be more useful on phones when rotating between portrait and landscape. That said, my phones resolution is so small it triggers the tray in both scenarios, so unable to test on my end :(.

The tray overlay seems to cause scrolling (I'm guessing theres some 100vh usage somewhere which causes some annoyances on mobile due to the extra mobile browser chrome), see screenshot below

packages/overlay/src/ActiveOverlay.ts
```diff
@@ -245,18 +247,40 @@ export class ActiveOverlay extends SpectrumElement {
 
         this.state = 'active';
         this.feature();
-        if (this.placement && this.placement !== 'none') {
+        if (this.placement === 'none') {
```
Westbrook, line 250: @davidyxu The code here should address the `100vh` situation causing scroll in your previous review, can you confirm that this has corrected the issue on your end (Android, right?) as well?

tools/reactive-controllers/src/MatchMedia.ts
```diff
@@ -0,0 +1,48 @@
...
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+import type { ReactiveController, ReactiveElement } from 'lit';
+
+export const DARK_MODE = '(prefers-color-scheme: dark)';
+export const IS_MOBILE =
```
Westbrook, line 15: @davidyxu I've updated the query here to prevent the tray in desktop all together. This _should_ mitigate some of the question about wanting to switch from one to the other while open.

In general, I'd like not to switch between the two while open, do you have thoughts on how we could be more clear about that reality herein?

packages/overlay/src/ActiveOverlay.ts
davidyxu: 👍 yep looks good on my end

tools/reactive-controllers/src/MatchMedia.ts
davidyxu: 👍 seems reasonable, not sure what the spectrum standards are but I wonder if just always using tray on phone vs always using popover for desktop/ipad would be the simplest world to reason about.

Maybe the best of all worlds would be some configurable attribute on sp-picker which defaults to responsive, and the documentation can describe what responsive behavior means here.

davidyxu (approved):
lgtm

tools/reactive-controllers/src/MatchMedia.ts
Westbrook: Agree on this. Added to #2086 for tracking...

PR #2085, opened 2022-02-01
fix(number-field): validate value before dispatching "change" event

Westbrook:
## Description
Validate `value` before dispatching `change` so that restraints like `min` and `max` are followed before announcing a new `value`.

## Motivation and context
Correctness.

## How has this been tested?

-   [ ] _Test case 1_
    1. Go [here](https://number-field-change--spectrum-web-components.netlify.app/storybook/?path=/story/number-field--default)
    2. Apply a value for `max` in the "Controls" panel
    3. Switch to the "Actions" panel
    4. Type a value larger than what you applied for `max`
    5. Press `Enter`
    6. See that the final `input` and the `change` event immediately there after are bound by the value you applied to `max`.

## Types of changes

## Checklist

### DIFF
--- a/packages/number-field/src/NumberField.ts
+++ b/packages/number-field/src/NumberField.ts
@@ -122,7 +122,8 @@ export class NumberField extends TextfieldBase {
     public stepModifier = 10;
 
     @property({ type: Number })
-    public set value(value: number) {
+    public set value(rawValue: number) {
+        const value = this.validateInput(rawValue);
         if (value === this.value) {
             return;
         }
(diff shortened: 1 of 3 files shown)


hunterloftis (approved):
lgtm, just want to make sure that this behavior is intentional:

- unconstrained input is still visible after pressing enter, becomes constrained on blur
- onInput event sometimes passes a numeric value and sometimes an object

https://user-images.githubusercontent.com/364501/152038651-a301d9fb-cb9d-43ed-b639-a2e807033c4d.mov

Westbrook:
1. Intended, the current functionality chooses not to alter the content in the field while the focus is in the field, even if the surfaced `value` is altered under the covers. Could definitely take notes on that for the future, but that aligns with all other validation processes.
2. Interesting. It looks like Storybook protects logging `NaN` directly, even though that's the value of the field when there is not contents.

PR #2117, opened 2022-02-16
'sp-dialog': Accessibility Updates

### LINKED ISSUE #1283: [dialog] a11y review
### Feedback

❌ - focus not moved to dialog when opened
❌ - focus not trapped in the dialog once entered - can still move to base page
❌ Dialog should have role=dialog
❌ The rest of the page should have aria-hidden=true so it is not presented to screen readers - will need to move the dialog to be a sibling of page content
❌ Dialog should have aria-labelledby pointing to the dialog title
❌ Dialog contents may or may not be pointed to by aria-describedby on the dialog - this depends on the content of the dialog and where focus ends up.
❌ = focus does not move to the dialog and no indication that it is a dialog is read

### Path forward

1. Review why focus isn't being moved into overlay. Had been doing that recently.
2. Focus is now trapped in the overlay. Previously was not leveraging suggested API in docs site
3. Add `role='dialog'`. This should only be associated with overlay content that capture the tab order, correct?
4. This has been added since the review
5. Associate title via `aria-labelledby` 
6. What sort of heuristic can be used to decide whether content should be associated via `aria-describedby`? 
7. Solved by no. 1 above.

cc @jnurthen

Westbrook:
## Description
- ensure the use of `role="dialog"` on the `<sp-dialog>` element
- relate the `[slot="heading"]` content to the `<sp-dialog>` via `aria-labelledby`
- when three or less elements are supplied to the default slot, related that content to the `<sp-dialog>` via `aria-describedby`
- allow `[autofocus]` content to receive focus when not sending focus to the `<sp-dialog>` itself

## Related issue(s)
- fixes #1283

## Motivation and context
Content should inform screen readers of itself appropriately.

## How has this been tested?
-   [ ] _Test case 1_
    1. Turn on your screen reader
    2. Go [here](https://a11y-dialog--spectrum-web-components.netlify.app/storybook/?path=/story/dialog-wrapped--form)
    3. Close the dialog.
    4. Open the dialog.
    5. See that the screen reader is directed to the "Street" input automatically
-   [ ] _Test case 2_
    1. Turn on your screen reader
    2. Go [here](https://a11y-dialog--spectrum-web-components.netlify.app/storybook/?path=/story/dialog-wrapped--long-content)
    3. Close the dialog.
    4. Open the dialog.
    5. See that the screen reader is directed to the `<sp-dialog>` by default
    6. There what seems to be a bug here with Voice Over where you fill be forwarded to the `role='dialog'` element but in Chrome and Firefox you will need to use `shift + control + option + ArrowUp` to exit the group and the `shift + control + option + ArrowDown` to re-enter the group and in Safari you will need to use `shift + control + option + ArrowDown` to enter the group before you'll be allowed to navigate through all the items in the group.
-   [ ] _Test case 3_
    1. Repeat Test case 1 and Test case 2 across browsers Chrome, Firefox, Safari
    2. Repeat Test case 1 and Test case 2 in you mobile browser with the screen reader turned on
    3. Repeat Test case 1 and Test case 2 in various browsers in Windows with https://assistivlabs.com/dashboard

## Types of changes

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,7 +9,7 @@ executors:
 parameters:
     current_golden_images_hash:
         type: string
-        default: b48c59df4af377fcd40c6dbdd51a57e34192ce8c
+        default: 79a1b6971a96a61ddef0ac801035b74708b01430
 commands:
     downstream:
         steps:
--- a/packages/base/package.json
+++ b/packages/base/package.json
@@ -22,6 +22,8 @@
     "exports": {
         ".": "./src/index.js",
         "./src/*": "./src/*",
+        "./condition-attribute-with-id": "./src/condition-attribute-with-id.js",
+        "./condition-attribute-with-id.js": "./src/condition-attribute-with-id.js",
         "./decorators": "./src/decorators.js",
         "./decorators.js": "./src/decorators.js",
         "./directives": "./src/directives.js",
--- a/packages/base/src/condition-attribute-with-id.ts
+++ b/packages/base/src/condition-attribute-with-id.ts
@@ -1,0 +1,46 @@
+/*
+Copyright 2020 Adobe. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License. You may obtain a copy
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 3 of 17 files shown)


najikahalsema (dismissed):
lgtm! nice work.

PR #2121, opened 2022-02-22
fix(overlay): measure initial overlay data offscreen

Westbrook:
## Description
Measure initial overlay data off screen such that even when resource constrained we can position the overlay without it appearing in an incorrect location during janked frames.

## Types of changes

## Checklist


spdev3000 (approved):
LGTM, while I cannot guarantee we never need an overlay start position at (0,0).

PR #2152, opened 2022-03-07
Tsc order

Westbrook:
## Description
- ensure `shared` is built correctly before `overlay` is build 🤞🏼 
- test that all packages have an `src/index.js` before allowing publication
- depend directly on `reparentChildren` rather than through the package entry file.

## Motivation and context
More reliable build/release process

## How has this been tested?
Building a bunch of time in CI to ensure it doesn't go red there.
Run `yarn build:clean` and then `yarn build:confirm` to ensure you exit the task with errors.

## Types of changes

## Checklist


najikahalsema (approved):
lgtm but I'd like to get some other eyes on it too!

packages/shared/tsconfig.json
```diff
@@ -5,5 +5,6 @@
         "rootDir": "./"
     },
     "include": ["*.ts", "src/*.ts", "global.d.ts"],
-    "exclude": ["test/*.ts", "stories/*.ts"]
+    "exclude": ["test/*.ts", "stories/*.ts"],
+    "references": [{ "path": "../base" }]
```
Westbrook, line 9: Likely a red herring, but the only reason I can think of that `shared` wouldn't be available when `overlay` was built is that it somehow beat `base` and then had missing types.

packages/overlay/src/ActiveOverlay.ts
```diff
@@ -19,7 +19,7 @@ import {
 } from '@spectrum-web-components/base';
 import { ifDefined } from '@spectrum-web-components/base/src/directives.js';
 import { property } from '@spectrum-web-components/base/src/decorators.js';
-import { reparentChildren } from '@spectrum-web-components/shared';
+import { reparentChildren } from '@spectrum-web-components/shared/src/reparent-children.js';
```
Westbrook, line 22: This is where the issue arrises and _might_ point to types not being passed through `export * from 'x'` patterns in the `shared` entry point, but I can't 100% be sure.

package.json
```diff
@@ -16,6 +16,7 @@
         "build:clear-cache": "rimraf packages/*/tsconfig.tsbuildinfo && rimraf tools/*/tsconfig.tsbuildinfo",
         "build:compare": "tsc --build projects/vrt-compare/tsconfig.json",
         "build:component-inventory": "node ./tasks/build-component-inventory.js",
+        "build:confirm": "lerna exec --ignore \"{@spectrum-web-components/{base,bundle,close-button,clear-button,icons-ui,icons-workflow,iconset,modal,shared,styles},documentation,example-project-rollup,example-project-webpack,swc-templates}\" -- test -f src/index.js",
```
Westbrook, line 19: This will check for `src/index.js` files in all but the listed packages `base,bundle,close-button,clear-button,icons-ui,icons-workflow,iconset,modal,shared,styles` and then fail is the build artifact requires is missing.

Westbrook:
Through 270+ installs I've not caught the race condition on this version of the branch, so 🤞🏼 this is a change that holds water going forward. The GitHub Actions won't pass as they need the version on `latest` to work, which don't. That begs an interesting question as to whether we should run that post release so we can catch issues like this in the future, before consumers.

hunterloftis (approved):

PR #2160, opened 2022-03-10
One config

Westbrook:
## Description
Make build calls to TSC through one config rather than an array of smaller ones.
- surface the TSC call to the main command line context so that we are more likely to catch errors there.
- see up to 2 minute performance benefits in CI from the p95 on visual regression jobs.

## Motivation and context
Safer builds, faster builds.

## How has this been tested?

-   [ ] _Test case 1_
    1. Go [here](https://app.circleci.com/pipelines/github/adobe/spectrum-web-components/11261/workflows/2cf3316e-fe58-4baa-a4fe-b6d8571b822e)
    2. See success and timing, assuming it doesn't catch a flake after I sign off 🤞🏼 

## Types of changes

## Checklist


tsconfig-all.json
```diff
@@ -0,0 +1,86 @@
...
+        "tools/**/*.ts",
+        "projects/story-decorator/**/*.ts",
+        "projects/vrt-compare/**/*.ts",
+        "test/lit-helpers.ts"
+    ],
+    "exclude": [
+        "packages/*/node_modules/**/*.ts",
+        "tools/*/node_modules/**/*.ts"
+    ],
+    "references": [
```
castastrophe, line 18: Would it make sense to generate these automatically rather than manually curate them?

najikahalsema (approved):

tsconfig-all.json
Westbrook: This is a great idea. I've created https://github.com/adobe/spectrum-web-components/issues/2162 to track. With the relative stability of the library I don't think we should block on getting the perf benefits here.

PR #2164, opened 2022-03-15
fix: avoid registering lit-virtualizer globally

hunterloftis:
## Description

Avoid registering `lit-virtualizer` in the global custom elements namespace.

## Motivation and context

Grid depends on `LitVirtualizer`, but not on `<lit-virtualizer>`. Since other components may use the virtualizer at different versions, let's avoid registering in the global namespace which can create collisions.

## How has this been tested?

- Added unit tests

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/tools/grid/src/LitVirtualizer.ts
+++ b/tools/grid/src/LitVirtualizer.ts
@@ -1,0 +1,27 @@
+/*
+Copyright 2020 Adobe. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License. You may obtain a copy
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under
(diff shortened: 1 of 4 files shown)


tools/grid/src/LitVirtualizer.ts
```diff
@@ -0,0 +1,27 @@
...
+// Grid extends LitVirtualizer and doesn't need the tag, which can conflict with other uses of lit-virtualizer in the same document.
+// This is a hack to keep lit-virtualizer from registering globally.
+
+const oldDefine = customElements.define;
+customElements.define = (name, constructor) => {
+    if (name === 'lit-virtualizer') return;
+    oldDefine.call(customElements, name, constructor);
+};
+
+const { LitVirtualizer } = await import('@lit-labs/virtualizer');
```
benjamind, line 23: Top level await! Fun!

benjamind (approved):
I *think* this should be ok. I'm not entirely certain how toplevel await works across bundlers, but I suspect its fine.

This is definitely an interesting 'hack', definitely not a 1.0 release type of thing we should be doing but for now as a stopgap this should work until we can get the lit-virtualizer updated.

PR #2174, opened 2022-03-21
feat(theme): filter css variables

### LINKED ISSUE #2173: scale/themes package is very large due to many variables
### Code of conduct

### Description of issue

The `spectrum-scale-medium.css` and `spectrum-scale-large.css` files are very large due to a high number of spectrum varable definitions that map component specific names to global/alias variables.

These are largely unused by component implementations at this time.

Suggest we filter these out so we only include the variables actually used by components in the theme.

benjamind:
## Description

Attempts to fix #2173 by filtering all variables in the scales package to only those found in the css files of the components.

## Related issue(s)

#2173 

## Motivation and context

Bundle sizes have ballooned in recent versions due to the addition of many new variables.

## Types of changes

## Checklist

## Best practices


GarthDB:
Smart - the move to a more focused token model should help with this.

castastrophe:
Love this concept though a note that it could cause some issues for downstream projects like UEC.  I pull variables as defined by SWC as a source of truth.  I think ideally the source would be pulled from Spectrum as tokens but referencing the values available in sp-theme was the fastest path forward.

Not that this is a blocker by any means but could this minified set be pre or postfixed so as to prevent a possible breaking change for downstream projects until we can clean things up on our end? Might be good to test this in PSWeb as well to see if all the variables being used in their components are also still working.

scripts/spectrum-vars.js
```diff
@@ -17,11 +17,79 @@ import fs from 'fs-extra';
...
+    options = options || {};
+
+    var variables = options.variables;
+
+    function filterDeclarations(root) {
+        root.walk(function (rule) {
+            if (rule.type === 'rule') {
+                rule.each(function (decl, index) {
+                    if (decl.variable) {
+                        // always include global and alias vars
```
castastrophe, line 41: Love this - great inclusion.

hunterloftis (approved):

PR #2179, opened 2022-03-23
chore: update component inventory

hunterloftis:
## Description

Updates inventory.md.

@Westbrook I feel like I should connect this to `docs:build` to automate it - any objection?


Westbrook:
Yes. I’d almost say make it a page in the docs site that’s not _currently_ in the nav so it’s easier to link to for sharing. Then the artifact doesn’t need to be in the repo at all…

PR #2205, opened 2022-04-06
Spectrum CSS updates and tooling to manage

Westbrook:
## Description
- Add `yarn update:spectrum-css` to support moving all package with `@spectrum-css` dependencies to latest
  - might want to:
    - extend to the root package.json, too
    - add usage to README.md
    - support non-`latest` versioning...
- Update Spectrum CSS dependencies

## Motivation and context
Prep for CCX PR.

## How has this been tested?
CI was run and VRTs are reviewed.

## Types of changes

## Checklist


scripts/update-spectrum-css.js
```diff
@@ -0,0 +1,64 @@
...
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import { readFileSync, writeFileSync } from 'fs';
+import latestVersion from 'latest-version';
```
castastrophe, line 16: Rather than pulling in a new dependency (latestVersion is a great package though it also pulls in packageJson with it), what would you think of running exec: `yarn info ${packageName} dist-tags.latest --json` this can be easily parsed for the value of data which holds the latest version number.  It lets us leverage yarn and exec rather than any new packages.  Just an option - I do think latestVersion is a nice clean syntax though.

scripts/update-spectrum-css.js
```diff
@@ -0,0 +1,64 @@
...
+import fg from 'fast-glob';
+
+async function update() {
+    let updated = false;
+    for (const packageJSONPath of await fg([
+        `./package.json`,
+        `./packages/**/package.json`,
+        `./projects/**/package.json`,
+        `./tools/**/package.json`,
+    ])) {
```
castastrophe, line 26: Could this be similarly expressed as:
```suggestion
    for (const packageJSONPath of await fg([
        `**/package.json`
    ], { ignore: ['**/node_modules/**'] })) {
```

castastrophe: The benefit here being that it would capture packages even if new workspaces are added or folders renamed.

scripts/update-spectrum-css.js
```diff
@@ -0,0 +1,64 @@
...
+        let shouldUpdate = false;
+        console.log(packageJSONPath);
```
castastrophe, line 28: Is this line for debugging or do you want to print the package path as part of the cli output?

scripts/update-spectrum-css.js
```diff
@@ -0,0 +1,64 @@
...
+        }
+        for (const packageName in devDependencies) {
+            if (packageName.startsWith('@spectrum-css')) {
+                const version = await latestVersion(packageName);
+                devDependencies[packageName] = version;
+                shouldUpdate = true;
+            }
+        }
+        if (shouldUpdate) {
+            packageJSON.dependencies = dependencies;
```
castastrophe, line 46: Instead of manually writing the files here - would running `yarn upgrade` via an exec command be an option?

castastrophe (dismissed):
A few suggestions on the script and 1 potentially issue with when the consoles are being called (1 after the exit command) but otherwise this looks good.

scripts/update-spectrum-css.js
Westbrook: Initially, because I did not know this command. Currently, because I like not having to go back and forth between the subprocess and easier. I look at this more after your other notes.

scripts/update-spectrum-css.js
Westbrook: No matter where I run `yarn upgrade` I can't get this to actually do what I want here. If this worked from the outside, we _might_ be able to get away without a script all together. Maybe it works in a newer version of `yarn`, but I'm not ready to dive into that rabbit hole at this moment. Are you able to use this across monorepo packages on your end?

scripts/update-spectrum-css.js
```diff
@@ -0,0 +1,64 @@
...
+                shouldUpdate = true;
+            }
+        }
+        for (const packageName in devDependencies) {
+            if (packageName.startsWith('@spectrum-css')) {
+                const version = await latestVersion(packageName);
+                devDependencies[packageName] = version;
+                shouldUpdate = true;
+            }
+        }
```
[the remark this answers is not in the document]
Westbrook: Tweaked this a bit to manage the asynchrony and the way I was handling persistence through the method, but great call!

scripts/update-spectrum-css.js
castastrophe: I have used it in the past yes but that's fair enough.  What issue are you seeing with it?  Does it not run at all or does it run in an unexpected way?

Issue #2206, opened 2022-04-06
[Feat]: Ability to prevent keyboard driven interaction events on elements which are not focus-visible


benjamind:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Possibly all

### Description of the requested feature

In some applications, particularly those focused around design tools, we have often have a global keyboard shortcut for the 'Pan tool' to activate when holding down the `Space` key. This creates a problem from an a11y point of view since `Space` also commonly is used to 'activate' a control, e.g. a button, or a dropdown.

Currently when a button is focused via a mouse/pointer input modality it receives focus, but does not get granted focus-visible matching. This is desirable so that focus indication does not clutter the UI for a mouse driven experience. Our design system gives no other indication of focused element state, unless the user uses a keyboard input modality to focus the button.

Pressing `Space` however when a button is has focus, will result in the button both getting `focus-visible` matching AND activating. So there's a bit of an unexpected UX here, where the user has no idea what will happen when they press the `Space` key because there is no focus-visibility on the button, although this is what happens in native HTML controls if they do not have focus styling.

This also introduces the problem for our design tools, `keydown` events happen after `focus-visible` matching is applied (even in capturing phase), this means we can't detect if a `keydown` capturing phase event at the high level in the DOM is destined for a control that is `focus`/`focus-visible`, so we can't determine if we should intercept the `Space` key or not to handle our pan tool shortcut, i.e. we have no way of knowing if the event should be handled by the button or our global shortcut handler.

In discussion with our a11y team and @hunterloftis we came to the conclusion that a better approach would be for all our activatable controls to only activate on `Space` if they are already `focus-visible` matching. The keydown event will cause `focus-visible` to apply to the control, so the first press of `Space` with a button focused should make it `focus-visible` but not activate it, then the second press would activate it as normal and because it handled the event it should call `stopPropagation()` on the event. The simple rubric here is that keyboard activation events should not be enacted unless the control has visible focus indication, i.e. you know where your focus is as a user. This obviously runs counter to the native HTML elements which activate on `Space` if they have `focus` regardless of their `focus-visible` state, these controls also do not prevent propagation of the `keydown` event (since otherwise they would be useless as primitives).

### Implementation notes or ideas

To implement this behavior I propose the following:
* Modify `FocusVisiblePolyfillMixin` to implement tracking of `focus-visible` state on `focus`
* Add a new function `shouldAllowKeyboardActivation` to the mixin that will:
[...]

PR #2207, opened 2022-04-06
feat(shared): adds keyboard-activation event

benjamind:
## Description

An implementation of the approach outlined in #2206.

DO NOT MERGE - for discussion only.

## Related issue(s)

- #2206 

## Motivation and context

Attempt to provide better capability for design tools to support use of `Space` key as a global keyboard shortcut without sacrificing accessibility of keyboard navigation/activation of controls.

## How has this been tested?

- Unit tests added

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/shared/src/focus-visible.ts
+++ b/packages/shared/src/focus-visible.ts
@@ -30,6 +30,16 @@ interface OptionalLifecycleCallbacks {
     manageAutoFocus?(): void;
 }
 
+export interface FocusVisibleKeyboardActivation {
+    shouldAllowKeyboardActivation(event: KeyboardEvent): boolean;
+}
+
+type FVConstructor<T extends Constructor<MixableBaseClass>> = {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    new (...args: any[]): InstanceType<T> & FocusVisibleKeyboardActivation;
+    prototype: T['prototype'] & FocusVisibleKeyboardActivation;
+};
+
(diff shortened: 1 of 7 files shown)


packages/shared/src/focus-visible.ts
```diff
@@ -30,6 +30,16 @@ interface OptionalLifecycleCallbacks {
...
 
+export interface FocusVisibleKeyboardActivation {
+    shouldAllowKeyboardActivation(event: KeyboardEvent): boolean;
+}
+
+type FVConstructor<T extends Constructor<MixableBaseClass>> = {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    new (...args: any[]): InstanceType<T> & FocusVisibleKeyboardActivation;
+    prototype: T['prototype'] & FocusVisibleKeyboardActivation;
+};
```
benjamind, line 41: This type was nasty....I'm not happy about it.

packages/shared/src/focus-visible.ts
```diff
@@ -152,5 +229,5 @@ export const FocusVisiblePolyfillMixin = <
         }
     }
 
-    return FocusVisibleCoordinator;
+    return FocusVisibleCoordinator as unknown as FVConstructor<T>;
```
benjamind, line 232: I couldn't get the typing to work here so I just forced it, seems like its not straightforward to have the mixin constructor apply new properties to the incoming types prototype.

benjamind:
Closing this as I have concluded after further discussions that this isn't the right outcome.

PR #2211, opened 2022-04-07
fix: support darkest/lightest color in express theme

hunterloftis:
Potential implementation of express/darkest and express/lightest.


Westbrook:
Applied in #1905

PR #2217, opened 2022-04-11
Css Corrections

Westbrook:
## Description
- use latest Spectrum CSS
- correct processing of CSS Custom Properties defined off of `:host` in Picker
- correct CSS update script to install dependencies as a range
- precess unicode values in all CSS files at conversion time

## Motivation and context
Keeping up to date with Spectrum.

## How has this been tested?
Ci ran.

## Types of changes

## Checklist

### DIFF
--- a/packages/number-field/src/NumberField.ts
+++ b/packages/number-field/src/NumberField.ts
@@ -311,24 +311,24 @@ export class NumberField extends TextfieldBase {
     protected onFocus(): void {
         super.onFocus();
         this._trackingValue = this.inputValue;
-        this.keyboardFocused = true;
+        this.keyboardFocused = !this.readonly && true;
         this.addEventListener('wheel', this.onScroll);
     }
 
     protected onBlur(): void {
         super.onBlur();
-        this.keyboardFocused = false;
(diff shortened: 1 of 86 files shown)


packages/number-field/src/NumberField.ts
```diff
@@ -311,24 +311,24 @@ export class NumberField extends TextfieldBase {
     protected onFocus(): void {
         super.onFocus();
         this._trackingValue = this.inputValue;
-        this.keyboardFocused = true;
+        this.keyboardFocused = !this.readonly && true;
```
najikahalsema, line 314: Are these changes just a fix you're sticking in there? Or are they relevant to the rest of the work? Just curious :-)

najikahalsema (approved):
Looks good! Have a question, but it's not a merge blocker or anything. Nice work, thank you!

packages/number-field/src/NumberField.ts
Westbrook: This comes from a change in the `readonly` delivery in Spectrum. Text field centric UIs will no longer deliver the text field background when readonly and this prevents them from delivering the focus state in that context.

Issue #2197, opened 2022-03-31
[Bug]: Submenu doesn't open on hover or remains open after hovering out, overlapping with other submenus


loredanaspataru:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-menu

### Expected behavior

When having multiple menu items that have a submenu that opens on hover, submenu should close when hovering out.  

### Actual behavior

Submenu persists open after hovering out of their area. Because of this sometimes overlaps with other submenus or doesn't open the other submenu.

### Screenshots

![Screenshot 2022-03-31 at 10 45 20](https://user-images.githubusercontent.com/37333191/161007788-181e60da-c104-47cb-89cf-436257c8c3cf.png)
![Screenshot 2022-03-31 at 10 45 36](https://user-images.githubusercontent.com/37333191/161007811-edf93618-67e1-4460-9ebc-f83b7a053c85.png)

### What browsers are you seeing the problem in?

Firefox, Chrome, Safari

### How can we reproduce this issue?

1. Go to [https://git.corp.adobe.com/pages/DIWI/uec-previews/main/?path=/story/application-menu--default](https://git.corp.adobe.com/pages/DIWI/uec-previews/main/?path=/story/application-menu--default)
2. Open menu and hover over `Edit` and then `Transform` menu items to open their submenus
3. Hover out of the menu area 
4. Then hover again over `View` to open its submenu
5. Notice that sometimes `View` menu doesn't open and other times overlaps with the previously opened `Edit` submenu

PR #2218, opened 2022-04-12
fix(menu): submenu remains open

sorinsarca:
## Description

When mouse moves quickly over menu items that contain submenus, some submenus remain open.
This is because `pointerenter` directly opens the submenu, in contrast with `pointerleave` which waits 100ms
to see if pointer returns.
In this PR `pointerenter` delays submenu open to make sure the pointer is targeting the menu item and not just passing by.

## Related issue(s)

- #2197 

## How has this been tested?

Updated existing test

## Types of changes

## Checklist


Westbrook:
The 100ms here is a super naive low lift solution for ["the cone"](https://ambientimpact.com/web/snippets/amazons-menu-prediction-cone) due to timing and team availability for the initial release here. I think it would be better if we implemented actual support for this reality and allow these interactions to happen immediately rather than bind both sides to this 100ms "work around". In particular this allows for a user to "pass by" the submenu root on their way to making a submenu selection without having to think about pausing for effect.

Would you be interested in taking a look at doing so?

sorinsarca:
I don't think that "the cone" is the approach we should take here. It works when you show/hide the submenu using the css display block/none, but things start to go south when you reparent the submenu, add in-out animations or transitions, use touchpad or mouse whell, and on top of this use multiple nested submenus.

Here are some reasons:

**Handling the toucpad/mousewheel**

Some submenus can be vary tall, limiting the height results in a vertical scrollbar.
Using the mouse wheel to scroll over the items won't trigger `pointerenter/pointerleave` events ([see fiddle](https://jsfiddle.net/0jLte3s4/)),
so special handlers must be added. Depending if the user has inverted scroll or not, special logic must be added to those
special wheel handlers to handle the vertical axis.

**Controlling the overlay**

I have almost zero control over the Overlay. For example, there is no option to remove animations, and I have
to patch it somehow to be able to create a context-menu with a native feeling. 

```ts
ActiveOverlay.prototype.applyContentAnimation = function () {
    // skip animations
    return Promise.resolve(true);
}
```
I also injected `:host {transition: none!important}` into the Popover's stylesheet to disable the transitions.

Directly opening the submenu on `pointerenter` is bad idea until we can fully control the overlay, otherwise there
will be flashes of submenu opening and closing.

**Who's in control**

Currently menu-item handles the state of the submenu, so we can end up with multiple menu-item siblings having the submenu open. 
The "owner" of menu-item should be in control and not allow multiple submenus to be open. That owner
can only be menu because there are also menu-groups.
Because there are animations and transitions the menu must first wait for submenu to fully close before
opening other submenu.

> The 100ms here is a super naive low lift solution

Naive or not, the solution is a simple one that works. 
I do believe in keeping things simple, instead of predicting final pointer intent we should focus on user intent,
so instead of triangulating positions to find if the pointer is in a position where it can open the submenu
we can let user pause a little bit over the menu-item to see if the intent is to open the submenu.
Simple like that, without gps connection.

> Would you be interested in taking a look at doing so?

We can create a new issue to investigate if the cone can work with the current system and if it brings value, but until we have clear idea on how it will work we need a fix for current behavior.

Westbrook:
I will work to make time to do a more thorough review of this in the next day or two. Please rebase this to `main` to help get me started.

sorinsarca:
Rebased

Westbrook:
Upon closer inspection, I find that the issue can still be recreated with this change, which points to the root of the issue still not being clear. We'll need to dig deeper on this and see if we can find an alternative or additional change to fully remedy this situation.

I agree with the `disabled` commit, and would love to get that into the library. Would you be able to contribute that as a stand alone PR?

sorinsarca:
I've submitted a PR for disabled bug: #2256 
Please check the description.

PR #2232, opened 2022-04-18
fix: allow for Picker element to be reparented

### LINKED ISSUE #2175: Reparenting an sp-picker while open leads to its losing some state
### Code of conduct

### Impacted component(s)

sp-picker

### Expected behavior

Reparenting `<sp-picker>` maintains the list of options as well as the currently-active option.

### Actual behavior

Reparenting `<sp-picker>` while it's open loses the currently-active option.

Additionally, while writing the reproduction, I noticed that the picker's styling changes each time it's reparented (replicate by clicking "reparent before" repeatedly).

### Screenshots

https://user-images.githubusercontent.com/364501/159538574-12ff8ecb-0126-4a62-9fec-3ec5aca138b5.mov

_No response_

Reproduced here: https://webcomponents.dev/edit/NyO89LhEIfJnZJMNeR3m/src/index.ts?p=stories

1. Select an option from the picker
2. Click on "reparent after"
3. Click on the picker to open it
4. Observe that the selected option is cleared when the picker is reparented

_No response_

_No response_

Westbrook:
## Description
- ensure that the `dir` attribute is managed correctly in all elements when reparenting them
- allow for `value` property in `sp-picker` and derivative elements to be managed correctly when reparenting the element
- ensure Picker passes "selection" to the Menu Items it holds before reparenting them into a Menu, this allows no 0th indexed Menu Items to be "focused" when the overlay opens with a "selection"

## Related issue(s)
- fixes #2175

## Motivation and context
Elements should be resilient to varying use cases

## How has this been tested?
I've added unit tests for the use case in the issue, you could recreate this manually in the browser if you'd like, as well.

## Types of changes

## Checklist


hunterloftis:
Manually testing here: https://picker-reparenting-2--spectrum-web-components.netlify.app/storybook/?path=/story/picker--default

In the console, via:

```
document.getElementsByTagName('p')[2].append(document.getElementsByTagName('sp-picker')[0])
```

Works in this PR, fails in prod, so 👍

packages/split-button/src/SplitButton.ts
```diff
@@ -186,26 +186,26 @@ export class SplitButton extends SizedMixin(PickerBase) {
         }
     }
 
-    protected manageSelection(): void {
+    protected async manageSelection(): Promise<void> {
         super.manageSelection();
```
hunterloftis, line 190: If changing the signature to be async, should this also await its super call to be sure order is preserved? (1. whatever super does, 2. this.manageSplitButtonItems())

```suggestion
        await super.manageSelection();
```

hunterloftis (dismissed):
Thanks for sorting this out!

@spdev3000 once this is released, I think you'll be able to simplify/delete some picker refs & close handlers in your recent work.

packages/split-button/src/SplitButton.ts
Westbrook: Actually 👎🏼 ... I tried this and then it fails VRT because it updates the timing of the open/select/highlight code. It _likely_ points to some lucky timing vs intentional timing, but I'd like to skip digging into it while it doesn't show as broken in any of our tests/demos.

hunterloftis: Interesting, then I think we're running `super.manageSelection()` and `this.manageSplitButtonItems()` concurrently, and relying on some relative timing between the two.

Not changing while things pass works for me.

I'm curious what happens with this, just to complete all the different ways they could be ordered:

```ts
await this.manageSplitButtonItems()
await this.manageSelection()
```

Westbrook: Trying in CI now. Having `manageSplitButtonItems` should encapsulate it in the asynchrony of `manageSelection` 🤞🏼 Hopefully, there's not other side effects here...

PR #2256, opened 2022-04-27
fix(menu): disabled menu-item should not open submenu

sorinsarca:
## Description

Disabled menu item should not open submenu.

@Westbrook 
There are some issues with the tests on Chrome,
I suspect the cleanup doesn't properly work because the html is reused (on the next two tests) in a buggy way (keeping the disabled attribute).
Moving the test to the end of list resolves the issue.
This doesn't happen on Firefox or WebKit, only on Chrome.
Did this happened before? I don't want to "solve" this just by changing the test position.

## Related issue(s)

- #2218 

## Types of changes

## Checklist

### DIFF
--- a/packages/menu/test/submenu.test.ts
+++ b/packages/menu/test/submenu.test.ts
@@ -399,6 +399,51 @@ describe('Submenu', () => {
 
         expect(rootItem.open).to.be.false;
     });
+    it('not opens if disabled', async () => {
+        const el = await styledFixture<Menu>(
+            html`
+                <sp-menu>
+                    <sp-menu-item disabled class="root">
+                        Has submenu
+                        <sp-menu slot="submenu">
+                            <sp-menu-item class="submenu-item-1">
+                                One
+                            </sp-menu-item>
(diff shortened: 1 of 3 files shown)


packages/menu/test/submenu.test.ts
```diff
@@ -399,6 +399,51 @@ describe('Submenu', () => {
 
         expect(rootItem.open).to.be.false;
     });
+    it('not opens if disabled', async () => {
```
Westbrook, line 402: Finally got a chance to review these two test failures. It comes from a bug in `Focusable` that is unrelated to your change and that we will be tracking as https://github.com/adobe/spectrum-web-components/issues/2325

To allow this PR to move forward, can you move this test to the end of the test file so that it is not effected by the bug?

Westbrook (changes requested):
Address the code order question in the tests and rebase to `main` and this should be good to go!

Westbrook (approved):
LGTM. Thanks for working through this @sorinsarca 🙇🏽‍♂️

Issue #1170, opened 2021-02-04
Required attributes?


Westbrook:
When `sp-progress-bar` applied `role="progress"` to itself the implication is that `label` (which applies a visible label to the shadow DOM) or `aria-label/lledby` is required to ensure the accessibility of the element. We'd like to ensure that here and in other locations where element sprout roles onto themselves these sorts of requirements are maintained at more than just the documentation level.

Can TS or lit-plugin or something else support the inclusion of required attributes?

If not, can we do so with a SWC specific linting plugin?

If not, should we add runtime warnings (possibly a development build 🙈) to clarify the absence of these attribute?

hunterloftis:
For 1.0: at least one instance of the end-to-end workflow of a dev build, so we can add component support for it over time.

Issue #800, opened 2020-08-03
There's no error when using sp-theme without importing a color/scale


adixon-adobe:
### Expected Behaviour
It would be nice for newer users to add a console error when trying to use sp-theme and only importing '@spectrum-web-components/theme/sp-theme' (without removing the ability to delay-load or not load them).

### Actual Behaviour
There's no error to indicate you need to load more for it to properly show up. 

#### Sample Code that illustrates the problem
```
import '@spectrum-web-components/theme/sp-theme'
...
return html`
  <sp-theme color="light" scale="large">
  {...}
  </sp-theme>`
```

adixon-adobe:
Offhand I could see adding something like this to suppress errors: `<sp-theme async-load />`

This case is probably a little less important, but I'd guess new clients will reasonably expect the sample code to "just work". The fact that it doesn't is likely to be surprising, and adding some breadcrumbs about it seems reasonable.

Westbrook:
Gathering info in the same place, currently this is what the "easy" usage notes are:

![image](https://user-images.githubusercontent.com/1156657/89216847-25da5680-d599-11ea-8bb6-548cbacea09b.png)

I like how `<sp-theme async-load />` or `<sp-theme defer />` etc. works here. I was playing with similar in a different context that never worked out as I had hoped, but here it might be the best path forward.

Westbrook:
There's also docs that don't say you need both: 

```
<sp-theme color="light">
```

Westbrook:
Relates to #1170

Westbrook:
Looking through the right messaging here, and for each of `theme`, `color`, and `scale` it seems like we can cover this set of realities:
- there is no value for the attribute available
- the value of the attribute is resolving to the _default_ for the application (the first value of that attribute with a fully registered theme fragment, e.g. loaded and usable)
- the value supplies does not have a registered theme fragment
 
By publishing these messages in the Dev Build it seems like it makes sense to publish them all, to ensure that consumers are fully equipped to make decisions for their applications' success.

PR #2257, opened 2022-04-27
feat: build *.dev.js files from everything

Westbrook:
## Description
- add `.dev.js` exports via the the export condition `development` to all packages
  - empower by switching from `tsc` to `esbuild` and processing each file twice one into dev mode and one removing all dev mode relative code
- add documentation for consuming/leveraging this code: https://esbuild-2--spectrum-web-components.netlify.app/dev-mode/
- add messages to:
  - Bage, re: variants
  - Base, re: dev mode
  - Picker, re: sp-menu deprecation
  - Progress Bar/Circle, re: accessible consumption
  - Slider, re: min/max usage in handles
  - Theme, re: attribute and fragment presence 

## Related issue(s)
- fixes https://github.com/adobe/spectrum-web-components/issues/800
- fixes https://github.com/adobe/spectrum-web-components/issues/2320
- fixes https://github.com/adobe/spectrum-web-components/issues/1170

## How has this been tested?

-   [ ] _Test case 1_
    1. Added tests
-   [ ] _Test case 2_
    1. Built test apps

## Types of changes

## Checklist


tasks/esbuild-packages.js
```diff
@@ -0,0 +1,92 @@
...
+        (path) =>
+            path.search('/test/') === -1 && path.search('/stories/') === -1
+    );
+    build({
+        entryPoints: devPaths,
+        bundle: false,
+        outdir: '.',
+        sourcemap: true,
+        watch: options.watch,
+        define: { 'window.__SWC_DEBUG__': true },
```
castastrophe, line 51: What would you think of using __DEBUG__ instead so that Hera could use the same naming approach?

Westbrook: I could definitely be convinced. There's some level of copy pasta not making them the same _could_ prevent. This could be made an argument of the method for easy cross project usage, as well. My first goal was to make sure I could get the build not to blow up at types time. Next goal is to see if I can keep that happening while being able to remove `window.` as it bothers my and makes the placement code pretty verbose. If we can get that working, I'd love to get back to just `DEBUG` as it's simpler and less likely to cause issues, and solves your request, too!

package.json
```diff
@@ -64,14 +64,16 @@
         "spectrum-tokens": "node ./scripts/spectrum-tokens.js",
         "spectrum-vars": "node ./scripts/spectrum-vars.js",
         "start": "yarn storybook",
-        "storybook": "yarn build:ts && yarn build:css && run-p storybook:start watch:css build:watch",
+        "storybook": "yarn build && yarn storybook:quick",
```
castastrophe, line 67: Oh `storybook:quick` - clever naming idea

package.json
```diff
@@ -11,19 +11,17 @@
     "scripts": {
         "analyze": "lit-analyzer \"packages/*/src/**/!(*.css).ts\"",
         "analyze:quick": "lit-analyzer \"packages/*/src/!(*.css).ts\"",
-        "build": "node test/visual/create.js && yarn build:css && yarn build:ts",
+        "build": "run-p test:create build:css && run-p esbuild:ts build:ts",
```
castastrophe, line 14: Love the addition of the `test:create` command.

package.json
```diff
@@ -64,14 +64,16 @@
         "spectrum-tokens": "node ./scripts/spectrum-tokens.js",
         "spectrum-vars": "node ./scripts/spectrum-vars.js",
         "start": "yarn storybook",
-        "storybook": "yarn build:ts && yarn build:css && run-p storybook:start watch:css build:watch",
+        "storybook": "yarn build && yarn storybook:quick",
         "storybook:build": "build-storybook -o projects/documentation/dist/storybook",
-        "storybook:start": "web-dev-server --config wds-storybook.config.js",
+        "storybook:quick": "run-p watch:css build:watch storybook:run",
+        "storybook:run": "wds --config wds-storybook.config.js",
```
castastrophe, line 70: I saw some great advice the other day that recommended not using shorthand in npm scripts since they're more self-documenting when fully spelled out. Might be nice to do that here for `wds` and use `web-dev-server` instead so it's really clear what package is being called.

package.json
```diff
@@ -64,14 +64,16 @@
...
         "test": "yarn test:focus unit",
         "test:bench": "node test/benchmark/cli.cjs",
-        "test:build": "tsc --build test/tsconfig-plugins.json && tsc --build test/tsconfig-test.json",
+        "test:build": "yarn build",
```
castastrophe, line 73: Interesting - can you outline this change for me? Test is running the generic build - does that now include building test assets including the plugins?

package.json
```diff
@@ -80,7 +82,7 @@
         "test:visual:clean:baseline": "rimraf test/visual/screenshots-baseline",
         "test:visual:clean:current": "rimraf test/visual/screenshots-current",
         "test:watch": "yarn test:watch:focus unit",
-        "test:watch:focus": "yarn test:build && run-p watch:css \"test:build -w\" \"test:start --watch --group {1}\" --",
+        "test:watch:focus": "yarn test:build && run-p watch:css build:watch \"test:start --watch --group {1}\" --",
```
castastrophe, line 85: Curious what the {1} does here?  Is that referencing the $1 input variable for the command or is this something else?

packages/accordion/exports.json
```diff
@@ -0,0 +1,7 @@
+{
```
castastrophe, line 1: Note to self to ask about this in our 1:1

packages/accordion/package.json
```diff
@@ -20,13 +20,33 @@
     "module": "./src/index.js",
     "type": "module",
     "exports": {
-        ".": "./src/index.js",
-        "./src/*": "./src/*",
+        ".": {
```
castastrophe, line 23: LOVE THIS!

packages/action-menu/package.json
```diff
@@ -20,13 +20,28 @@
...
+        "./src/action-menu.css.js": "./src/action-menu.css.js",
+        "./src/index.js": {
+            "development": "./src/index.dev.js",
+            "default": "./src/index.js"
+        },
+        "./sp-action-menu.js": {
+            "development": "./sp-action-menu.dev.js",
+            "default": "./sp-action-menu.js"
+        },
+        "./sync/sp-action-menu.js": {
```
castastrophe, line 41: Note to ask about this sync directory

packages/badge/src/Badge.ts
```diff
@@ -51,6 +51,21 @@ export class Badge extends SizedMixin(ObserveSlotText(SpectrumElement, '')) {
     public variant: BadgeVariant = 'informative';
 
     protected override render(): TemplateResult {
+        if (window.__swc.DEBUG) {
```
castastrophe, line 54: What if we alias'ed this in SpectrumBase eventually so components could do a `this.DEBUG_MODE` or something? Then if you want to alter the way we set and fetch the state, we could do that at SpectrumBase without having to update every component file.

packages/badge/test/badge.test.ts
```diff
@@ -46,5 +48,40 @@ describe('Badge', () => {
         await elementUpdated(el);
 
         await expect(el).to.be.accessible();
+        expect(consoleWarnStub.called).to.be.false;
+        consoleWarnStub.restore();
+    });
+    it('warns in Dev Mode when sent an incorrect `variant`', async () => {
```
castastrophe, line 54: We could add a TODO to build a utility function to make setting up these dev mode tests more standardized for future components.

packages/base/src/Base.ts
```diff
@@ -163,3 +163,49 @@ export function SpectrumMixin<T extends Constructor<ReactiveElement>>(
 }
 
 export class SpectrumElement extends SpectrumMixin(LitElement) {}
+
+if (window.__swc.DEBUG) {
+    window.__swc = {
```
castastrophe, line 168: Curious why attach them to the window object and not make these available to call on the component.

packages/menu/src/Menu.ts
```diff
@@ -22,8 +22,8 @@ import {
     query,
 } from '@spectrum-web-components/base/src/decorators.js';
 
-import {
-    MenuItem,
+import { MenuItem } from './MenuItem.js';
+import type {
```
castastrophe, line 26: Love separating out the types like this

.storybook/manager.js
```diff
@@ -3,5 +3,7 @@ import yourTheme from './theme.js';
 
 addons.setConfig({
     theme: yourTheme,
-    showRoots: false,
+    sidebar: {
```
castastrophe, line 6: Nice error message squashing!  I should do the same in UEC

packages/styles/exports.json
```diff
@@ -0,0 +1,52 @@
+{
```
castastrophe, line 1: We should look into if these files can be automatically generated - it might be easy to miss an update here if files are changed upstream.

packages/theme/README.md
```diff
@@ -54,6 +54,23 @@ import '@spectrum-web-components/theme/express/scale-medium.js';
 import '@spectrum-web-components/theme/express/scale-large.js';
 ```
 
+## Example
+
+An `<sp-theme>` element expects a value for each of its `theme`, `color`, and `scale` attributes to either be provided on the element.
```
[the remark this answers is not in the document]
Westbrook: There's only a default _if_ fragments in those spaces have been loaded; default just means first loaded. This means that nested use doesn't need to be so strict, but it doesn't ensure a delivery for you. If you've not opted into one you may be getting functionality that on accident or different than what you expected. The structure here is about preventing the number of KBs that a product needs to ship over the wire.

In that way, you'll actually see the documentation site trigger this warning by default. However, in that case, we load the default styles in CSS so that they are not needed until a user chooses to customize their theme beyond size, prefers color, and Spectrum Classic.

packages/slider/src/HandleController.ts
```diff
@@ -606,11 +615,20 @@ export class HandleController implements Controller {
...
+                    if (!next) {
+                        window.__swc.issueWarning(
+                            this.host,
+                            {
+                                localName: this.host.localName,
+                                type: 'api',
+                                level: 'default',
+                            },
+                            '<sp-slider-handle> elements that are the last child of an <sp-slider> element cannot have attribute "max=\'next\'"',
+                            'This needs a URL!'
```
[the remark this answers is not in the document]
Westbrook: Oh no, it literally _needs_ a URL. HA! I'll create/apply this.

projects/types/global.d.ts
```diff
@@ -0,0 +1,49 @@
...
+         * @param id {string:string:string} - the id of the warning, a `:` joined string of the component directory, the warning type, and the warning level
+         * @param warning {string} - the primary message to delivery to the developer
+         * @param url {string} - a URL at which more infromation, or the standard documentation, can be found
+         * @param issues {string[]} - an optional array of issues to format into the message
+         */
+        issueWarning(
+            element: HTMLElement | undefined,
+            warningData: SWCWarningData,
+            warning: string,
+            url: string,
```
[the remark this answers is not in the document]
Westbrook: I think this IS required, if you can't point to why you issue a warning then there shouldn't really need to be a warning.

tasks/build-exports-list.js
```diff
@@ -0,0 +1,41 @@
...
+        return;
+    }
+    const packageJSON = JSON.parse(fs.readFileSync(packageJSONPath, 'utf8'));
+    const exports = packageJSON.exports;
+    if (!exports) {
+        return;
+    }
+    delete exports['.'];
+    delete exports['./package.json'];
+    fs.writeFileSync(exportsPath, JSON.stringify(exports, null, '    '), {
```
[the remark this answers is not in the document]
Westbrook: It can be edited, which is why we don't write it if it already exists. While _most_ packages won't need to do so, not being able to would prevent packages like [styles](https://github.com/adobe/spectrum-web-components/pull/2257/files#diff-04e371540a2335aafa408bf506165a4305a795efcda81bd5c604aaaf0c5534a0) or [theme](https://github.com/adobe/spectrum-web-components/pull/2257/files#diff-1d44a5f253c534f58a60f8319b939b023cf07622ea7763ef42757223e25a1cd4) from being able to build a complete exports map.

package.json
Westbrook: Good call. I was namely pruning commands back as I renamed them as things. Will clean this up.

package.json
Westbrook: It applies the argument from `test:watch:focus tooltip` to that specific command `test:start --watch --group tooltip`. So, maybe the same? I had never done it before this and the blog I read said jump, and I asked how high 😉

tasks/build-exports-list.js
```diff
@@ -0,0 +1,41 @@
...
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+import path from 'path';
+import fs from 'fs';
+import fg from 'fast-glob';
+
+const build = (packageJSONPath) => {
+    const exportsPath = path.join(packageJSONPath, '..', 'exports.json');
```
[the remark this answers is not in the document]
Westbrook: I mainly wrote this to help me create them the first time. It's not _really_ to maintain them dynamically. It would be rad if you had thoughts on how to do that, but with there not being a 100% homogenous approach to exports, I wasn't seeing a path.

packages/base/src/Base.ts
```diff
@@ -163,3 +163,49 @@ export function SpectrumMixin<T extends Constructor<ReactiveElement>>(
...
+        ...window.__swc,
```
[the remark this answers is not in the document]
Westbrook: I'm not sure we should be allowing customizations to those. As outlined in the docs, these are not API that should be included in your consumption. This is really just allowing the consuming developer to support values to options properties like `ignoreWarningLocalNames`, `ignoreWarningTypes`, `ignoreWarningLevels`, and `verbose`.

tasks/build-exports-list.js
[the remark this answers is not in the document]
castastrophe: Makes sense, do you mind if I open an issue as a placeholder to think on that in the future?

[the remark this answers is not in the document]
castastrophe: https://github.com/adobe/spectrum-web-components/issues/2463

packages/badge/src/Badge.ts
```diff
@@ -51,6 +51,21 @@ export class Badge extends SizedMixin(ObserveSlotText(SpectrumElement, '')) {
...
+            if (!BADGE_VARIANTS.includes(this.variant)) {
+                window.__swc.issueWarning(
+                    this,
+                    {
+                        localName: this.localName,
+                        type: 'api',
+                        level: 'default',
+                    },
+                    `<${this.localName}> element expect the "variant" attribute to be one of the following:`,
+                    'https://opensource.adobe.com/spectrum-web-components/components/badge/#variants',
```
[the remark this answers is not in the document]
hunterloftis: Since tooling takes care of quickly & reliably replacing strings, splitting a url into multiple parts like `${BASE}/badge/#variants` can make it harder to make updates when the url changes.

Another nice thing in vscode & other tools is that you can cmd+click on a full URL to open it.

packages/base/src/Base.ts
```diff
@@ -163,3 +163,49 @@ export function SpectrumMixin<T extends Constructor<ReactiveElement>>(
...
+        issuedWarnings: new Set(),
```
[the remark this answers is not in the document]
hunterloftis: If the goal is to shorten, then consider just `warnings`, since `emitted` is so general as to not be self-descriptive.

packages/base/src/Base.ts
```diff
@@ -163,3 +163,47 @@ export function SpectrumMixin<T extends Constructor<ReactiveElement>>(
...
+        issueWarning: (element, data, message, url, issues): void => {
```
hunterloftis, line 171: Perhaps consider the verb form `warn()` ala `console`.

It would also be 👍 if we could figure out a way to reduce the number of arguments required here.

```suggestion
        issueWarning: (element, data: SWCWarningData, message: string, url: string, issues: string[]): void => {
```

hunterloftis (dismissed):
imo the API is kind of big (it would take looking up the structure several different types in order to add a warning). Maybe that's ok, since we don't want to overwhelm users with warnings.

packages/base/src/Base.ts
Westbrook: `.warn()` sets a nice tone for expanding to `.log()` and `.error()` along the lines of @castastrophe's note later down the line if we need.

We type this method in `projects/types/global.d.ts` does it need to be typed here, too? Can add for coverage, however.

Would an options bag be better for this method?

hunterloftis: In that case it probably doesn't need to be typed (I didn't check in an editor).

If we used an options bag with some reasonable defaults, could we flatten out the nested props in `SWCWarningData`? Such that both of these would be valid uses:

`.warn(this, "Don't do that", 'https://example.com')`

`.warn(this, 'Or that either', 'https://adobe.com', {
  level: 'high',
  type: 'accessibility',
});`

Alternatively, with tags:

`.warn(this, "Don't do that", 'https://example.com', ['high', 'accessibility']);`

Westbrook: Updated with that. Good call on the defaults for simplicity.

projects/documentation/content/dev-mode.md
```diff
@@ -0,0 +1,94 @@
...
+
+type WarningType = 'default' | 'accessibility' | 'api';
+
+type WarningLevel = 'default' | 'low' | 'medium' | 'heigh' | 'deprecation';
+
+type SWCWarningData = {
+    localName: string;
+    type: WarningType;
+    level: WarningLevel;
+};
```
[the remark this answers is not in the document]
Westbrook: I think with the other API updates (optionality and defaults) this makes more sense as an object.

packages/badge/src/Badge.ts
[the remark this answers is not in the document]
Westbrook: VSCode click to open is a really powerful point. I do like _less_, but I don't think this is a place where that's possible right now.

packages/base/src/Base.ts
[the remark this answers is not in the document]
Westbrook: `issued*` or some other past tense verb is I think it important because this is where we track what _has_ happened. I also like the parallel to `console.*` in the management of the API names here. I could see it making sense to expanding into those other types as we find good use cases for them.

packages/base/src/Base.ts
```diff
@@ -163,3 +163,49 @@ export function SpectrumMixin<T extends Constructor<ReactiveElement>>(
 }
 
 export class SpectrumElement extends SpectrumMixin(LitElement) {}
+
+if (window.__swc.DEBUG) {
+    window.__swc = {
+        ...window.__swc,
+        issuedWarnings: new Set(),
+        issueWarning: (element, warningData, warning, url, issues): void => {
```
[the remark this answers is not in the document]
Westbrook: I _hope_ the typing making up for brevity. But, being able to include other `console.*` methods over time may prove useful!

package.json
Westbrook: If this is useful to people, sounds good! Updating.

PR #2272, opened 2022-05-05
fix: move property management into update

Westbrook:
## Description
Follow some Lit Dev mode warnings and reduce some of the double renders possible in the library.

## Motivation and context
Performance.

## How has this been tested?
-   [ ] _Test case 1_
    1. Run storybook locally, compare the Progress Bar console to that in a previous branch.

## Types of changes
-   [x] Perf refactor

## Checklist

### DIFF
--- a/packages/base/src/sizedMixin.ts
+++ b/packages/base/src/sizedMixin.ts
@@ -73,11 +73,11 @@ export function SizedMixin<T extends Constructor<ReactiveElement>>(
 
         private _size: ElementSize | null = defaultSize;
 
-        protected firstUpdated(changes: PropertyValues): void {
-            super.firstUpdated(changes);
+        protected update(changes: PropertyValues): void {
             if (!this.hasAttribute('size') && !noDefaultSize) {
                 this.setAttribute('size', this.size);
             }
+            super.update(changes);
         }
     }
     return SizedElement;
--- a/packages/field-label/src/FieldLabel.ts
+++ b/packages/field-label/src/FieldLabel.ts
@@ -134,17 +134,16 @@ export class FieldLabel extends SizedMixin(SpectrumElement) {
 
     protected firstUpdated(changes: PropertyValues): void {
         super.firstUpdated(changes);
+        this.addEventListener('click', this.handleClick);
+    }
+
+    protected willUpdate(changes: PropertyValues): void {
         if (!this.hasAttribute('id')) {
             this.setAttribute(
                 'id',
                 `${this.tagName.toLowerCase()}-${FieldLabel.instanceCount++}`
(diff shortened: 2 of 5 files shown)


packages/base/src/sizedMixin.ts
```diff
@@ -73,11 +73,11 @@ export function SizedMixin<T extends Constructor<ReactiveElement>>(
 
         private _size: ElementSize | null = defaultSize;
 
-        protected firstUpdated(changes: PropertyValues): void {
-            super.firstUpdated(changes);
+        protected update(changes: PropertyValues): void {
```
hunterloftis, line 76: Would `willUpdate` be a better fit here? It'd remove the need for `super` and sounds from the [docs](https://lit.dev/docs/components/lifecycle/#willUpdate) like it may exist for essentially this kind of use case.

packages/field-label/src/FieldLabel.ts
```diff
@@ -134,19 +134,19 @@ export class FieldLabel extends SizedMixin(SpectrumElement) {
 
     protected firstUpdated(changes: PropertyValues): void {
         super.firstUpdated(changes);
+        this.addEventListener('click', this.handleClick);
+    }
+
+    protected update(changes: PropertyValues): void {
```
hunterloftis, line 140: ditto re: `willUpdate`

hunterloftis (approved):

packages/field-label/src/FieldLabel.ts
Westbrook: Asking these two together is a great question. And, while I didn't realize that `willUpdate()` was an option, I have a follow up question. Because Field Label mixes the Sized functionality into itself, there would be a need to call `super` for our code, but not for the Lit code. Would the use of `willUpdate()` and the differences between Lit and SWC point to the need/benefit of more advanced lifecycle documentation on our part? Does using `update()` protect us from the need for that additional documentation? I was about to make this update, until I realized that already we'd be hitting this problem and thought it would be good to have a path forward marked out for this internally.

hunterloftis: Hm, that's a good question. Maybe things that are expected to be extended, like mixins and SpectrumElement, _should_ use update() since that parallels what Lit does & means the Lit lifecycle recommendations still apply to the final subclasses.

If we went that route, then we'd want to keep `update` on the sized mixin (mirroring Lit) and use `willUpdate` on FieldLabel (like a typical Lit consumer).

Westbrook: I think that makes good sense as a rule to support these contexts! Will update.

PR #2278, opened 2022-05-07
fix: correctly apply tab order to Accordion Items

### LINKED ISSUE #2273: [Bug]: Accordions are navigable by Up/Down keys instead of Tab key
### Code of conduct

### Impacted component(s)

sp-accordion

### Expected behavior

Accordion items are navigable only by pressing Tab or Shift-Tab.

### Actual behavior

Accordion items are navigable only by pressing Up or Down.

### Screenshots

Screen Shot 2022-05-05 at 1 43 28 PM

Chrome

1. Go to https://opensource.adobe.com/spectrum-web-components/components/accordion/
2. Tab to the accordion.
3. Press Tab to go to the next accordion item.
4. See that focus has actually exited the accordion instead of moving to the next accordion item.

Looks like this was caused by https://github.com/adobe/spectrum-web-components/pull/2017

_No response_

_No response_

Westbrook:
## Description
- removed Roving Tab Index management of Accordion Items
- expanded testing of natural Tab Order access of Accordion Items

## Related issue(s)

- fixes #2273

## Motivation and context
Appropriate accessibility models.

## How has this been tested?

-   [ ] _Test case 1_
    1. Go [here](https://accordion-keys--spectrum-web-components.netlify.app/components/accordion/#example)
    2. Tab into the example.
    3. See that you can only access Accordion Items via `Tab` and not `ArrowUp/Down`

## Types of changes

## Checklist

### DIFF
--- a/packages/accordion/test/accordion.test.ts
+++ b/packages/accordion/test/accordion.test.ts
@@ -15,11 +15,7 @@ import { elementUpdated, expect, fixture, html } from '@open-wc/testing';
 import '../sp-accordion.js';
 import { Default } from '../stories/accordion.stories.js';
 import { Accordion, AccordionItem } from '@spectrum-web-components/accordion';
-import {
-    arrowDownEvent,
-    arrowUpEvent,
-    shiftTabEvent,
-} from '../../../test/testing-helpers.js';
+import { sendKeys } from '@web/test-runner-commands';
 import { spy } from 'sinon';
 
 describe('Accordion', () => {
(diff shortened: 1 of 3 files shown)


najikahalsema (approved):
Nice work!

packages/accordion/test/accordion.test.ts
```diff
@@ -15,11 +15,7 @@ import { elementUpdated, expect, fixture, html } from '@open-wc/testing';
 import '../sp-accordion.js';
 import { Default } from '../stories/accordion.stories.js';
 import { Accordion, AccordionItem } from '@spectrum-web-components/accordion';
-import {
```
jnurthen, line 18: previously the accordion supported both Tab AND up/down arrows. I'm absolutely fine with just Tab/Shift-Tab being the end result but want to make sure that the change was intentional

jnurthen (approved):

packages/accordion/test/accordion.test.ts
hunterloftis: I think up/down, when pressed while an accordion header button is focused, makes a lot of sense for quickly navigating - especially navigating _past_ large sets of focusable content to a sibling header, as in [this example](https://www.w3.org/TR/wai-aria-practices-1.2/examples/accordion/accordion.html).

This PR already moves accordion into a more-accessible state, so I recommend merging it & then adding support for arrow keys if you decide to as a separate endeavor.

hunterloftis (approved):

PR #2300, opened 2022-05-18
Overlay fallbacks

### LINKED ISSUE #2198: [Bug]: VirtualTrigger: Overlay get cut off screen with placement "right-start"
### Code of conduct

### Impacted component(s)

Overlay, VirtualTrigger

### Expected behavior

Overlay should always be positioned and rendered at the screen that all content is visible.

### Actual behavior

When using a virtual trigger to open a context menu with placement="right-start" - it happens that the overlay content won't be correct positioned and always visible for small window sizes - and content will get cut off the screen.
This is mostly the case when the opened popover content width is >= half of the UI viewport. For example: If the popover width is 300px, this happens when viewport (body) is shrank to 600px and less. 

### Screenshots

overlay-wrong-position

Chrome Version 99.0.4844.84

1. Go to https://webcomponents.dev/edit/ohZswGWTi4aHOb1GUd4X
2. Click on "ContextMenu Trigger"
3. Shrink page width (to about 600px or less)
4. Click again on trigger button (around phrase "Context" of ContextMenu Trigger)
5. Notify context menu content cut off page

You may try to set placement to "left-start" and there won't be a problem.

_No response_

_No response_

Westbrook:
## Description
Allow content opened from a "virtual trigger" to "flip" on both axises when looking for a place to position itself.
  - don't use `auto` because it's not predictable enough for closely related content like Pickers
  - update `size` usage to always fallback to the initial size when measuring
  - moves the Picker Menu location to above the Picker button in some Large VRTs due to their not being enough space below the picker for the content

## Related issue(s)

- fixes #2198

## How has this been tested?

-   [ ] _Test case 1_
    1. Go [here](https://overlay-fallbacks--spectrum-web-components.netlify.app/storybook/?path=/story/overlay--virtual-element)
    2. Squeeze the preview window down to 400px wide or so
    3. Try opening the "context menu" from various places on the screen.
      - top left
      - top right
      - bottom left
      - bottom right
    4. See they the "context menu" opens in a related position that DOES NOT flow outside of the viewport.

## Types of changes

## Checklist

### DIFF
--- a/packages/overlay/src/ActiveOverlay.ts
+++ b/packages/overlay/src/ActiveOverlay.ts
@@ -38,6 +38,7 @@ import {
     arrow,
     computePosition,
     flip,
+    Placement as FloatingUIPlacement,
     offset,
     shift,
     size,
@@ -114,6 +115,26 @@ const parentOverlayOf = (el: Element): ActiveOverlay | null => {
         return parentOverlayOf(rootNode.host);
     }
     return null;
+};
+
+const getFallbackPlacements = (
+    placement: FloatingUIPlacement
+): FloatingUIPlacement[] => {
+    const fallbacks: Record<FloatingUIPlacement, FloatingUIPlacement[]> = {
+        left: ['right', 'bottom', 'top'],
+        'left-start': ['right-start', 'bottom', 'top'],
+        'left-end': ['right-end', 'bottom', 'top'],
+        right: ['left', 'bottom', 'top'],
+        'right-start': ['left-start', 'bottom', 'top'],
+        'right-end': ['left-end', 'bottom', 'top'],
+        top: ['bottom', 'left', 'right'],
+        'top-start': ['bottom-start', 'left', 'right'],
+        'top-end': ['bottom-end', 'left', 'right'],
+        bottom: ['top', 'left', 'right'],
+        'bottom-start': ['top-start', 'left', 'right'],
+        'bottom-end': ['top-end', 'left', 'right'],
+    };
+    return fallbacks[placement] ?? [placement];
 };
 
 /**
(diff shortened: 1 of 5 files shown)


packages/overlay/src/ActiveOverlay.ts
```diff
@@ -446,6 +492,34 @@ export class ActiveOverlay extends SpectrumElement {
...
+        //                     : '';
+        //                 Object.assign(this.style, {
+        //                     maxWidth: `${Math.floor(width)}px`,
+        //                     maxHeight: appliedHeight,
+        //                     height: appliedHeight,
+        //                 });
+        //             },
+        //         })
+        //     );
+        // }
```
hunterloftis, line 522:
```suggestion
```

hunterloftis (approved):

packages/overlay/src/ActiveOverlay.ts
```diff
@@ -116,6 +117,43 @@ const parentOverlayOf = (el: Element): ActiveOverlay | null => {
...
+    } else if (placement.startsWith('bottom')) {
+        fallbacks.push('left');
+        fallbacks.push('right');
+        fallbacks[0] = fallbacks[0].replace(
+            'bottom',
+            'top'
+        ) as FloatingUIPlacement;
+    }
+    return fallbacks;
+};
```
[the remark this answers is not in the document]
Westbrook: Great call on the dictionary. `auto` continues to not be an acceptable value for `FloatingUIPlacement` so while it might do what you are saying it _shouldn't_ ever do that...

packages/overlay/src/ActiveOverlay.ts
Westbrook: 😓

spdev3000:
Thx @Westbrook for that quick solution!

Issue #1643, opened 2021-07-23
Add Badge pattern


davidkudera:
Currently, there is no badge component.

Spectrum: https://spectrum.adobe.com/page/badge/
Spectrum CSS: https://opensource.adobe.com/spectrum-css/label.html

---

I'm not sure if the Spectrum CSS Label is compatible with the Spectrum Badge or if it even matters... I can see two possible ways of how to implement this:

1. Wait with this issue until https://github.com/adobe/spectrum-css/issues/598 is done
2. Implement it based on the current Label component and change it later to use the new Badge component

The name should be (I think) in both cases `<sp-badge>`. I could also probably help by preparing a PR with an initial implementation.

Westbrook:
Agreed, this will be a good add. We're working through the a pending release of Spectrum CSS output, and then if you interested in taking a swing at this, we'd love to see it. If not, I could see someone getting into it in the next couple of months. Thoughts?

davidkudera:
@Westbrook Cool, I'll send a PR. Is there some issue in Spectrum CSS I can subscribe to, to be notified when your work is done? Or is the issue I send above?

Westbrook:
I believe that is happening in https://github.com/adobe/spectrum-css/pull/1138 but I’ll double check with their team!

hunterloftis:
For visibility - we have a branch with a first draft here: https://hunterloftis-add-badge--spectrum-web-components.netlify.app/storybook/?path=/story/badge--default

The Badge design spec is going through some revisions right now; I'll integrate any changes into the above component and open a PR once the spec is stable.

castastrophe:
I was hoping to follow-up on this component - it's used in the Account Menu designs and I would rather pull it from SWC than create a UEC element for it.

hunterloftis:
@castastrophe sorry for the delay getting back to you; I'm catching up on GH notifications today.

As long as @westbrook and @najikahalsema are 👍 , I can update this branch and ship a web component that matches the current CSS implementation.

The spec question is basically: should (can) a badge be interactive? Or is it purely informational? The hover-state visual cue suggests interactivity, and the spec mentions interactivity, but it's not currently an accessibly interactive component (the CSS implementation is a [span](https://opensource.adobe.com/spectrum-css/badge.html)).

Along those same lines: if a badge can be interactive, is it sufficiently different from an action button?

action-button:
[...]

PR #2305, opened 2022-05-23
feat: add badge component

hunterloftis:
## Description

Implements the [Badge](https://spectrum.adobe.com/page/badge/) component.

## Related issue(s)

https://github.com/adobe/spectrum-web-components/issues/1643

## Motivation and context

Expanding component coverage

## How has this been tested?

1. Go to [the preview storybook](https://hunterloftis-badge--spectrum-web-components.netlify.app/storybook/?path=/story/badge--default)
2. Check out the various examples of badge usage
3. Go to [the preview docs site](https://hunterloftis-badge--spectrum-web-components.netlify.app/components/badge/)
4. Check out the documentation

## Types of changes

-   [X] New feature (non-breaking change which adds functionality)

## Checklist

## Best practices

### DIFF
--- a/packages/badge/README.md
+++ b/packages/badge/README.md
@@ -1,0 +1,113 @@
+## Description
+
+`<sp-badge>` elements display a small amount of color-categorized metadata. They're ideal for getting a user's attention.
+
+### Usage
+
+[![See it on NPM!](https://img.shields.io/npm/v/@spectrum-web-components/badge?style=for-the-badge)](https://www.npmjs.com/package/@spectrum-web-components/badge)
+[![How big is this package in your project?](https://img.shields.io/bundlephobia/minzip/@spectrum-web-components/badge?style=for-the-badge)](https://bundlephobia.com/result?p=@spectrum-web-components/badge)
+
+```
+yarn add @spectrum-web-components/badge
+```
+
+Import the side effectful registration of `<sp-badge>` via:
+
+```
+import '@spectrum-web-components/badge/sp-badge.js';
+```
+
+When looking to leverage the `Badge` base class as a type and/or for extension purposes, do so via:
+
+```
+import { Badge } from '@spectrum-web-components/badge';
+```
+
+## Sizes
+
+<sp-tabs selected="m" auto label="Size Attribute Options">
+<sp-tab value="s">Small</sp-tab>
+<sp-tab-panel value="s">
+
+```html demo
+<div style="display: flex; gap: var(--spectrum-global-dimension-size-50);">
+    <sp-badge size="s">Label</sp-badge>
+    <sp-badge size="s">
+        <sp-icon-checkmark-circle slot="icon"></sp-icon-checkmark-circle>
+        Icon and label
+    </sp-badge>
+</div>
+```
+
+</sp-tab-panel>
+<sp-tab value="m">Medium</sp-tab>
+<sp-tab-panel value="m">
+
+```html demo
+<div style="display: flex; gap: var(--spectrum-global-dimension-size-50);">
+    <sp-badge size="m">Label</sp-badge>
+    <sp-badge size="m">
+        <sp-icon-checkmark-circle slot="icon"></sp-icon-checkmark-circle>
+        Icon and label
+    </sp-badge>
+</div>
+```
+
+</sp-tab-panel>
+<sp-tab value="l">Large</sp-tab>
+<sp-tab-panel value="l">
+
+```html demo
+<div style="display: flex; gap: var(--spectrum-global-dimension-size-50);">
+    <sp-badge size="l">Label</sp-badge>
+    <sp-badge size="l">
+        <sp-icon-checkmark-circle slot="icon"></sp-icon-checkmark-circle>
+        Icon and label
+    </sp-badge>
+</div>
+```
+
+</sp-tab-panel>
+<sp-tab value="xl">Extra Large</sp-tab>
+<sp-tab-panel value="xl">
+
+```html demo
+<div style="display: flex; gap: var(--spectrum-global-dimension-size-50);">
+    <sp-badge size="xl">Label</sp-badge>
+    <sp-badge size="xl">
+        <sp-icon-checkmark-circle slot="icon"></sp-icon-checkmark-circle>
+        Icon and label
+    </sp-badge>
+</div>
+```
+
+</sp-tab-panel>
+</sp-tabs>
+
+## Variants
+
+The `<sp-badge>` can the customized with either semantic or non-semantic variants.
+
+### Semantic
+
+```html demo
+<div style="display: flex; gap: var(--spectrum-global-dimension-size-50);">
+    <sp-badge variant="neutral">Neutral</sp-badge>
+    <sp-badge variant="informative">Informative</sp-badge>
+    <sp-badge variant="positive">Positive</sp-badge>
+    <sp-badge variant="negative">Negative</sp-badge>
+</div>
+```
+
+### Non-Semantic
+
+```html demo
+<div style="display: flex; gap: var(--spectrum-global-dimension-size-50);">
+    <sp-badge variant="fuchsia">Fuchsia</sp-badge>
+    <sp-badge variant="indigo">Indigo</sp-badge>
+    <sp-badge variant="magenta">Magenta</sp-badge>
+    <sp-badge variant="purple">Purple</sp-badge>
+    <sp-badge variant="seafoam">Seafoam</sp-badge>
+    <sp-badge variant="yellow">Yellow</sp-badge>
+</div>
+```
--- a/packages/badge/src/Badge.ts
+++ b/packages/badge/src/Badge.ts
@@ -1,0 +1,61 @@
+/*
+Copyright 2020 Adobe. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License. You may obtain a copy
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import {
+    CSSResultArray,
+    html,
+    SizedMixin,
+    SpectrumElement,
+    TemplateResult,
+} from '@spectrum-web-components/base';
+import { property } from '@spectrum-web-components/base/src/decorators.js';
+
+import { ObserveSlotText } from '@spectrum-web-components/shared/src/observe-slot-text.js';
+import styles from './badge.css.js';
+
+export const BADGE_VARIANTS = [
+    'neutral',
+    'informative',
+    'positive',
+    'negative',
+    'fuchsia',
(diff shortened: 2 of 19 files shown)


package.json
```diff
@@ -178,7 +178,7 @@
         "typescript": "^4.5.3",
         "yargs": "^17.2.1"
     },
-    "customElements": ".storybook/custom-elements.json",
+    "customElements": "projects/documentation/custom-elements.json",
```
hunterloftis, line 181: This wasn't an intentional change, but rather a side effect of development (maybe a `yarn docs:something` script?). I can revert it if the .storybook version is canonical.

Westbrook: Related to https://github.com/open-wc/custom-elements-manifest/issues/121 don't worry about it here.

packages/badge/README.md
```diff
@@ -0,0 +1,49 @@
+## Description
+
+Badges show a small amount of color-categorized metadata. They're ideal for getting a user's attention.
```
Westbrook, line 3:
```suggestion
`<sp-badge>` elements display a small amount of color-categorized metadata. They're ideal for getting a user's attention.
```

packages/badge/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+### Usage
+
+[![See it on NPM!](https://img.shields.io/npm/v/@spectrum-web-components/badge?style=for-the-badge)](https://www.npmjs.com/package/@spectrum-web-components/badge)
+[![How big is this package in your project?](https://img.shields.io/bundlephobia/minzip/@spectrum-web-components/badge?style=for-the-badge)](https://bundlephobia.com/result?p=@spectrum-web-components/badge)
```
Westbrook, line 8: It'll be a follow up task (as it requires a release), but we should add a demo to webcomponents.dev. I can get you added to the Adobe Org so you can add it directly, once you set up an account there.

packages/badge/README.md
```diff
@@ -0,0 +1,49 @@
...
+import '@spectrum-web-components/badge/sp-badge.js';
+```
+
+When looking to leverage the `Badge` base class as a type and/or for extension purposes, do so via:
+
+```
+import { Badge } from '@spectrum-web-components/badge';
+```
+
+## Example
```
Westbrook, line 26: For element with t-shirt sizing, we use an `<sp-tabs>` patters to feature all of the sizes here. Can you review https://spectrum-css--spectrum-web-components.netlify.app/components/action-button/#sizes for that?

packages/badge/src/Badge.ts
```diff
@@ -0,0 +1,61 @@
...
+ * @slot - Text label of the badge
+ * @slot icon - Optional icon that appears to the left of the label
+ */
+export class Badge extends SizedMixin(ObserveSlotText(SpectrumElement, '')) {
+    public static get styles(): CSSResultArray {
+        return [styles];
+    }
+
+    @property({ type: String, reflect: true })
+    public variant: BadgeVariant = 'informative';
```
Westbrook, line 51: Will the badge break if someone supplies an invalid variant? If so, we may want to add some additional value management here.

packages/badge/src/Badge.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    protected render(): TemplateResult {
+        return html`
+            <slot name="icon" ?icon-only=${!this.slotHasContent}></slot>
+            <div id="label">
```
Westbrook, line 56: Buttons use `<div id="label" ?hidden=${!this.hasLabel}>` hidden to manage some style delivery here. Not sure if this needs to too.

packages/badge/src/badge.css
```diff
@@ -0,0 +1,139 @@
...
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+@import './spectrum-badge.css';
+
+/* center align icons and text */
+
+:host {
+    display: inline-flex;
```
Westbrook, line 18: This overrides https://github.com/adobe/spectrum-web-components/pull/2305/files#diff-5edf46dbb57c6fb3d202eb46df1dd1abd71d482a43702f41acc2478f3d765bfcR201 can you comment here why, if it's still needed.

packages/badge/src/badge.css
```diff
@@ -0,0 +1,139 @@
...
+}
+
+/* disable :hover until Spectrum CSS disables hover upstream */
+
+:host(:hover) {
+    background-color: var(--spectrum-badge-m-textonly-background-color);
+    color: var(--spectrum-badge-m-texticon-text-color);
+}
+
+/* implement fixed placement styling until Spectrum CSS adds it upstream */
```
Westbrook, line 29: Is there an issue we can track here?

packages/badge/src/badge.css
```diff
@@ -0,0 +1,139 @@
...
+    --spectrum-badge-neutral-textonly-border-bottom-left-radius: 0;
+    --spectrum-badge-neutral-textonly-border-bottom-right-radius: 0;
+}
+
+:host([fixed='left']) {
+    --spectrum-badge-neutral-textonly-border-top-left-radius: 0;
+    --spectrum-badge-neutral-textonly-border-bottom-left-radius: 0;
+}
+
+/* cascade badge's size to its icon */
```
Westbrook, line 51: Doesn't need to block, but we've added this in a couple of places, might be useful to share a single CSS import rather that copy/paste.

packages/badge/src/spectrum-badge.css
```diff
@@ -0,0 +1,398 @@
...
+    --spectrum-badge-m-textonly-background-color-hover: var(
+        --spectrum-badge-m-yellow-textonly-background-color-hover,
+        var(--spectrum-alias-background-color-yellow-hover)
+    );
+}
+:host {
+    background-color: var(--spectrum-badge-m-textonly-background-color);
+    color: var(--spectrum-badge-m-texticon-text-color); /* .spectrum-Badge */
+}
+:host(:hover) {
```
Westbrook, line 393: Did they say whether they would be removing hover with the move to a non-interactive model? If so, do we want to disabled that now so it doesn't become "expected" at the consumer level?

packages/badge/stories/badge.stories.ts
```diff
@@ -0,0 +1,146 @@
...
+            Icon and label
+        </sp-badge>
+
+        <sp-badge>
+            <sp-icon-checkmark-circle slot="icon"></sp-icon-checkmark-circle>
+        </sp-badge>
+    `;
+};
+
+export const Sizes = (): TemplateResult => {
```
Westbrook, line 43: Having multiple sizes in a single story can make VRTs difficult with changes, but this is pretty contained, so it might not be an issue.

packages/badge/test/benchmark/basic-test.ts
```diff
@@ -0,0 +1,19 @@
...
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import '@spectrum-web-components/badge/sp-badge.js';
+import { html } from '@spectrum-web-components/base';
+import { measureFixtureCreation } from '../../../../test/benchmark/helpers.js';
+
+measureFixtureCreation(html`
+    <sp-badge></sp-badge>
```
Westbrook, line 18: Might be useful to add content to this test. A label is enough, but an icon might be nice, too. Same with the above unit test.

packages/bundle/elements.ts
```diff
@@ -17,6 +17,7 @@ import '@spectrum-web-components/action-group/sp-action-group.js';
 import '@spectrum-web-components/action-menu/sp-action-menu.js';
 import '@spectrum-web-components/asset/sp-asset.js';
 import '@spectrum-web-components/avatar/sp-avatar.js';
+import '@spectrum-web-components/badge/sp-badge.js';
```
Westbrook, line 20: This needs to be in the `package.json` for `bundle`, too.

packages/badge/src/badge.css
```diff
@@ -0,0 +1,139 @@
...
+        --spectrum-alias-ui-icon-cornertriangle-size-300
+    );
+    --spectrum-ui-icon-tshirt-size-width: var(
+        --spectrum-alias-ui-icon-cornertriangle-size-300
+    );
+}
+
+/* don't shrink icons and ensure they're separated from labels */
+
+::slotted([slot='icon']) {
```
Westbrook, line 115: Oh, are there no rules for icons in this for icons in the actual CSS. Hrm. We may need to run this by Spectrum CSS for future additions on their end. In the interim, this likely wants to be more like Action Button, which is a little more complex. We might not have all of the custom properties to power it 🤔

packages/badge/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+Badges should always have a label for accessibility and clear comprehension.
```
Westbrook, line 28: Could a badge be simply an icon? In which case the Badge or the Icon would need `aria-label` or similar to also prove the accessibility of the delivery...

Westbrook (changes requested):
Super excited to see this _FINALLY_ landing!

packages/badge/README.md
hunterloftis: https://github.com/adobe/spectrum-web-components/issues/2319

packages/badge/src/Badge.ts
hunterloftis: https://github.com/adobe/spectrum-web-components/issues/2320

packages/badge/src/badge.css
hunterloftis: Yep, it's converting from `inline-block` to `inline-flex` in order to support the internal layout of `${icon} ${label}` (whereas the current upstream implementation only has labels, no icons).

For example, limiting the label to a max of 2 rows, allowing flexible shrinking & cropping of the label, while keeping the icon from shrinking, and maintaining a vertical center across both the icon and the label:

image

packages/badge/src/badge.css
hunterloftis: Good idea: https://github.com/adobe/spectrum-css/issues/1443

packages/badge/src/spectrum-badge.css
hunterloftis: Moved to `spectrum-config.js`:

```
/* disable :hover until Spectrum CSS disables hover upstream */
            exclude: [/\:hover/],
```

packages/badge/src/badge.css
hunterloftis: https://github.com/adobe/spectrum-css/issues/1443

packages/badge/README.md
hunterloftis: The [spec](https://spectrum.adobe.com/page/badge/#Label) says a badge "should always have a label," but also demonstrates icon-only badges.

I think specifying aria should probably be done by the user, if the user opts to omit a label - right? eg:

```html
<sp-badge aria-label="filters enabled">
  <sp-icon-checkmark-circle slot="icon"></sp-icon-checkmark-circle>
</sp-badge>
```

Westbrook: Yes. `sp-icon-*` also accept a `label` attribute.

najikahalsema (dismissed):
This looks great! Nice work, Hunter!

Westbrook (approved):
This looks great! 💯 Thanks for the slog with the various teams to bring this over the line. 🙇🏼‍♂️ With this I dare say that you onboarding is finally complete 😜

PR #2321, opened 2022-05-27
docs(badge): typo in readme

hunterloftis:
Was tempted to commit directly to main (any thoughts around that @Westbrook?)


Westbrook:
I'm not a fan on committing to `main`, though I can't say that I've not done it... In cases like this, using the PR to share that it happened with people and not needing a review (when an admin) also seems acceptable.

PR #2335, opened 2022-06-01
fix: prevent leaving multiple submenus open at a time

### LINKED ISSUE #2197: [Bug]: Submenu doesn't open on hover or remains open after hovering out, overlapping with other submenus

Westbrook:
## Description
Rather than hack the Menu/Menu Item relationship to enforce open/close rules on a Menu with multiple submenus, add a `root` concept to the overlay stack that prevents multiple overlays from the same root, even if they have different triggers.
- prevent leaving things open
- prevent passing focus back to the "root"
- reduce code needed to manage submenus
- doesn't have a test for duplicate submenus being open because while I'm "certain ™️" that the issue comes from the way `setCloseSelfAsSubmenu` and `setCloseOpenSubmenu` were being used, I could never isolate a reproduction case to base a new test on
- [x] needs a test for closing super decedent submenus when changing two+ levels up

## Related issue(s)
- fixes #2197
- closes #2218 

## Motivation and context
Things should "just work" 😉 

## How has this been tested?
-   [ ] _Test case 1_
    1. Go [here](https://submenu--spectrum-web-components.netlify.app/storybook/?path=/story/menu-submenu--submenu)
    2. Note: the menus open wonky by default.
    3. Mouse around in an attempt to break it and have a submenu not open or have more than one submenu for the same menu open or have sub-submenus not close, etc.
    4. Rejoice.

## Types of changes

## Checklist

### DIFF
--- a/packages/overlay/src/overlay-stack.ts
+++ b/packages/overlay/src/overlay-stack.ts
@@ -12,12 +12,15 @@ governing permissions and limitations under the License.
 
 import { ActiveOverlay } from './ActiveOverlay.js';
 import type {
-    OverlayCloseReasonDetail,
     OverlayOpenCloseDetail,
     OverlayOpenDetail,
 } from './overlay-types';
 import { OverlayTimer } from './overlay-timer.js';
 import '../active-overlay.js';
+import {
+    findOverlaysRootedInOverlay,
+    parentOverlayOf,
+} from './overlay-utils.js';
 
 function isLeftClick(event: MouseEvent): boolean {
     return event.button === 0;
@@ -268,6 +271,9 @@ export class OverlayStack {
             }
         }
 
+        if (details.root) {
+            this.closeOverlaysForRoot(details.root);
+        }
         if (details.interaction === 'click') {
             this.closeAllHoverOverlays();
         } else if (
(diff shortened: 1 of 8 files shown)


najikahalsema (approved):
This is so beautiful. I'm crying right now. Nice work!

Westbrook:
Nice catch @hunterloftis! It helped me to catch another issue, as well. Added tests for both of them now that they are fixed.

packages/overlay/src/overlay-stack.ts
```diff
@@ -478,7 +509,7 @@ export class OverlayStack {
     private async hideAndCloseOverlay(
         overlay?: ActiveOverlay,
         animated?: boolean,
-        clickAway?: boolean
+        shouldNotReceiveFocus?: boolean
```
hunterloftis, line 512: nit: could this name be more obvious / how could we avoid negating a boolean? What do you think of `skipFocus`?

hunterloftis (approved):

PR #2347, opened 2022-06-07
chore: create site-build UI in GitHub Actions; remove from CircleCI

castastrophe:
## Description

This PR creates a workflow_dispatch trigger for the site deploy commands. This allows the site to be deployed manually via the GitHub Actions UI.

## Motivation and context

CircleCI shows a failed status on most builds on the main branch because the approval to build the site is only "passed" on release branches.  This approach will clean up the status reporting for the main branch, making it easier to identify true failures.

## How has this been tested?

-   [ ] _gh workflow run publish.yml --ref chore-ci-build-site-on-release_
    - [docs](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow#running-a-workflow-using-github-cli)

Note: Cannot be run from the UI until the workflow exists on main.  Once on main you can use the UI or use gh cli or trigger it using a REST API call.

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,33 @@
+name: Site publish
+
+on:
+    check_suite:
+        types: [completed]
+    workflow_dispatch:
+jobs:
+    site-build:
+        name: Build & publish site
+        runs-on: ubuntu-latest
+        # Run the job if manually triggered or if the commit message includes '#publish' & the check suite has passed
+        if: github.event_name == "workflow_dispatch" || (contains(github.event.head_commit.message, '#publish') && github.event.check_suite.conclusion == "success")
+        steps:
+            - uses: actions/checkout@v2
+            - uses: actions/setup-node@v2
+              with:
+                  node-version: '16'
+                  cache: 'yarn'
+                  registry-url: 'https://registry.npmjs.org'
+
+            - name: Install dependencies
+              run: yarn --frozen-lockfile
+
+            - name: Generate Docs
+              run: yarn docs:ci
+
+            - name: Deploy to GitHub Pages
+              run: |
+                  git remote set-url origin https://git:${GITHUB_TOKEN}@github.com/${GITHUB_REPOSITORY}.git
+                  touch projects/documentation/dist/.nojekyll
+                  npx gh-pages -d projects/documentation/dist -m "[skip ci] update demonstration site" -t -u "github-actions-bot <support+actions@github.com>"
+              env:
+                  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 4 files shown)


Westbrook:
I'm not quite sure how this workflow would get triggered. Please make sure that information around that is updated in https://github.com/adobe/spectrum-web-components/blob/main/RELEASE_PROCESS.md so that we can accurately discuss this as a path forward.

castastrophe:
> I'm not quite sure how this workflow would get triggered. Please make sure that information around that is updated in https://github.com/adobe/spectrum-web-components/blob/main/RELEASE_PROCESS.md so that we can accurately discuss this as a path forward.

Can do!  Would you like me to link to the [workflow_dispatch docs](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow) (they have some nice screenshots in their walkthrough) or do up a summary in our docs?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,33 @@
+name: Site publish
+
+on:
+    check_suite:
```
castastrophe, line 4: @Westbrook I set this up to run when the check_suite completes, the goal being that if a commit message includes `#publish` and the check suite has passed, it will try to publish the docs site.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,33 @@
...
+        types: [completed]
+    workflow_dispatch:
```
castastrophe, line 6: Kept the manual workflow as well so you can always go trigger a build by hand.

.github/workflows/publish.yml
Westbrook: Nice: do we need to change something about the publish script? That string doesn't seem available to-date: https://github.com/adobe/spectrum-web-components/commit/3be62133721efba844cd7032566a2c49ed6d9875

Westbrook (commented):
This is looking great. Thanks for cleaning it up. Once the publish commit actually has the right strings in it, we should be good to merge.

castastrophe:
> This is looking great. Thanks for cleaning it up. Once the publish commit actually has the right strings in it, we should be good to merge.

I added it to the lerna publish message string - is that where you were thinking?

package.json
```diff
@@ -37,7 +37,7 @@
         "icons": "node ./scripts/process-icons.js && pretty-quick --pattern \"packages/**/*.svg.ts\" && eslint -f pretty --fix \"packages/**/*.svg.ts\"",
         "icons:ui": "yarn workspace @spectrum-web-components/icons-ui build",
         "icons:workflow": "yarn workspace @spectrum-web-components/icons-workflow build",
-        "lerna-publish": "lerna publish --message \"chore: release new versions\"",
+        "lerna-publish": "lerna publish --message \"chore: release new versions #publish\"",
```
Westbrook, line 40: 👍🏼

Westbrook (approved):
LGTM!

The real test is always the first _actual_ publish. We've got no new changes since the last publication, so I'll test a manual one once this has finished CI!

PR #2366, opened 2022-06-15
Ensure the Picker label is managed by its Menu Item children

### LINKED ISSUE #2158: [Bug]: When sp-menu-item changes its label while selected, picker label does not update
### Code of conduct

### Impacted component(s)

sp-picker

### Expected behavior

When the label of an item changes while selected, the picker should reflect that label

### Actual behavior

The old label remains

### Screenshots

_No response_

_No response_

3. Scroll to '....'
4. Check console
5. See error

https://webcomponents.dev/edit/msStOl89omAAl2v2k3VR/src/index.ts?p=stories

In the above demo, select "Soon" from the picker (ensure "Soon shows as the label in the picker)

Next click the button - this button changes the label "Soon" to "My Item Changed"

Note that the picker label field does not get updated.

My workaround currently is to target the selected sp-menu-item and do `.breakItemChildrenCache();` on it

Westbrook:
## Description
Update Menu Item to leverage a Mutation Observer as opposed to `slotchange` events when updating its content to work around https://github.com/lit/lit/issues/2629
- apply Mutation Observe with mutation controller
- update other usages of Mutation Observer to leverage mutation controller

## Related issue(s)

- fixes #2158

## Motivation and context
Components should "just work ™️" regardless of the tools you leverage them in.

## How has this been tested?

-   [ ] _Test case 1_
    1. Go [here](https://picker-label--spectrum-web-components.netlify.app/components/picker/#matching-value)
    2. Find the `sp-menu-item` element in the inspector with "Select inverse" as its text content
    3. Select the *Text Node* that represents that content
    4. In the console use `$0.textContent = 'Something else'` to update the active label of the Picker
    5. See that the "label" is updated
    6. Open the Picker and do the above using another text value
    7. See that the "label" is updated

## Types of changes

## Checklist

### DIFF
--- a/packages/shared/src/observe-slot-text.ts
+++ b/packages/shared/src/observe-slot-text.ts
@@ -14,11 +14,9 @@ import {
     property,
     queryAssignedNodes,
 } from '@spectrum-web-components/base/src/decorators.js';
+import { MutationController } from '@lit-labs/observers/mutation_controller.js';
 
-const slotElementObserver = Symbol('slotElementObserver');
-// Fix needed for: https://github.com/lit/lit/issues/1789
 const assignedNodesList = Symbol('assignedNodes');
-const startObserving = Symbol('startObserving');
 
 type Constructor<T = Record<string, unknown>> = {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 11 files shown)


packages/shared/src/observe-slot-text.ts
```diff
@@ -39,7 +37,25 @@ export function ObserveSlotText<T extends Constructor<ReactiveElement>>(
...
+
+            new MutationController(this, {
+                config: {
+                    characterData: true,
+                    subtree: true,
+                },
+                callback: (mutationsList: Array<MutationRecord>): boolean => {
+                    for (const mutation of mutationsList) {
+                        if (mutation.type === 'characterData') {
+                            this.manageTextObservedSlot();
```
hunterloftis, line 52: Would it make sense to early return here, so `this.manageTextObservedSlot()` is guaranteed to only be called once?

```suggestion
                            this.manageTextObservedSlot();
                            return true;
```

hunterloftis (approved):

PR #2367, opened 2022-06-15
Address various warnings raised in Lit Dev Mode.

Westbrook:
## Description
- general: clean up multiple `lit-html` usage in local tooling
- shared: move initial Observer Slot Text work into `update`, as a mixin having it there will ensure calls to `super` as the same is required by `LitElement`
- color-area: use private properties when validating state from form elements in the UI so there's not additional render required
- use `willUpdate`
  - Tabs
  - Active Overlay
  - Split View
  - Radio Group
  - Number Field
  - Search
- Update Benchmarking to register perf only once there are no additional renders queued
- use `update`
  - Picker
  - Action Menu
  - Split Button
  - Slider Handle
  - Textfield
- Track changes in Observe Slot Presence

## Related issue(s)
- refs #2257 

## Motivation and context
Clearing out upstream warnings make room for #2308 

## How has this been tested?

-   [ ] _Test case 1_
    1. Everything works as it did before... 😏 

## Types of changes
-   [x] Refactor

## Checklist

### DIFF
--- a/packages/radio/src/RadioGroup.ts
+++ b/packages/radio/src/RadioGroup.ts
@@ -90,41 +90,46 @@ export class RadioGroup extends FocusVisiblePolyfillMixin(FieldGroup) {
     @property({ reflect: true })
     public selected = '';
 
-    protected override firstUpdated(changes: PropertyValues<this>): void {
-        super.firstUpdated(changes);
-        this.setAttribute('role', 'radiogroup');
-        const checkedRadio = this.querySelector('sp-radio[checked]') as Radio;
-        const checkedRadioValue = checkedRadio ? checkedRadio.value : '';
-        // Prefer the checked item over the selected value
-        this.selected = checkedRadioValue || this.selected;
-        // Validate the selected value is actual a radio option
-        if (this.selected && this.selected !== checkedRadioValue) {
-            const selectedRadio = this.querySelector(
-                `sp-radio[value="${this.selected}"]`
+    protected override willUpdate(changes: PropertyValues<this>): void {
+        if (!this.hasUpdated) {
+            this.setAttribute('role', 'radiogroup');
+            const checkedRadio = this.querySelector(
+                'sp-radio[checked]'
             ) as Radio;
-            if (!selectedRadio) {
-                this.selected = '';
-            } else {
-                selectedRadio.checked = true;
+            const checkedRadioValue = checkedRadio ? checkedRadio.value : '';
+            // Prefer the checked item over the selected value
+            this.selected = checkedRadioValue || this.selected;
+            // Validate the selected value is actual a radio option
+            if (this.selected && this.selected !== checkedRadioValue) {
+                const selectedRadio = this.querySelector(
+                    `sp-radio[value="${this.selected}"]`
+                ) as Radio;
+                if (!selectedRadio) {
+                    this.selected = '';
+                } else {
+                    selectedRadio.checked = true;
+                }
             }
+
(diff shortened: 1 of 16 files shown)


packages/radio/src/RadioGroup.ts
```diff
@@ -90,41 +90,46 @@ export class RadioGroup extends FocusVisiblePolyfillMixin(FieldGroup) {
...
-        this.setAttribute('role', 'radiogroup');
-        const checkedRadio = this.querySelector('sp-radio[checked]') as Radio;
-        const checkedRadioValue = checkedRadio ? checkedRadio.value : '';
-        // Prefer the checked item over the selected value
-        this.selected = checkedRadioValue || this.selected;
-        // Validate the selected value is actual a radio option
-        if (this.selected && this.selected !== checkedRadioValue) {
-            const selectedRadio = this.querySelector(
-                `sp-radio[value="${this.selected}"]`
+    protected override willUpdate(changes: PropertyValues<this>): void {
```
castastrophe, line 93: Awesome - love the migration from firstUpdated to willUpdate

packages/radio/src/RadioGroup.ts
```diff
@@ -90,41 +90,46 @@ export class RadioGroup extends FocusVisiblePolyfillMixin(FieldGroup) {
...
+                event.stopPropagation();
+                const target = event.target as Radio;
+                this._setSelected(target.value);
+            });
+        }
 
-    protected override updated(changes: PropertyValues<this>): void {
-        super.updated(changes);
         if (changes.has('selected')) {
             this.validateRadios();
```
castastrophe, line 122: I suspect no but just double checking - does `this.validateRadios` need an await here?

yarn.lock
```diff
@@ -13956,25 +13956,34 @@ lit-analyzer@^1.2.1:
     vscode-html-languageservice "3.1.0"
     web-component-analyzer "~1.1.1"
 
-lit-element@^3.2.0:
+lit-element@^3.1.0, lit-element@^3.2.0:
```
castastrophe, line 13959: Interesting - why was lit-element@3.1.0 added here when before it was all aligned on 3.2.0?

yarn.lock
```diff
@@ -13956,25 +13956,34 @@ lit-analyzer@^1.2.1:
...
   version "3.2.0"
   resolved "https://registry.yarnpkg.com/lit-element/-/lit-element-3.2.0.tgz#9c981c55dfd9a8f124dc863edb62cc529d434db7"
   integrity sha512-HbE7yt2SnUtg5DCrWt028oaU4D5F4k/1cntAFHTkzY8ZIa8N0Wmu92PxSxucsQSOXlODFrICkQ5x/tEshKi13g==
   dependencies:
     "@lit/reactive-element" "^1.3.0"
     lit-html "^2.2.0"
 
-lit-html@^2.0.0, lit-html@^2.2.0:
+lit-html@^2.0.0, lit-html@^2.1.0, lit-html@^2.2.0:
```
castastrophe, line 13967: Same here, looks like several packages for lit are now bringing in multiple versions where before it only brought in one.

castastrophe (approved):
Approved but made note of some additional lit versions now showing up in the yarn.lock file.

packages/radio/src/RadioGroup.ts
Westbrook: Good check, and you're right that the work is (or rather can be) asynchronous, but it's not blocking work, so no await.

PR #2378, opened 2022-06-23
refactor: abstract "getDeepElementFromPoint" into shared package

### LINKED ISSUE #1798: [feat] Abstracted shadow root piecing method
> Is this something other components might need to do? If so, we could extract it into a shared function, since it already has no dependencies on Overlay and it's a pure function.

_Originally posted by @hunterloftis in https://github.com/adobe/spectrum-web-components/pull/1797#discussion_r709367903_

Westbrook:
## Description
Abstract `getDeepElementFromPoint` into `@spectrum-web-components/shared` so it can be used by other things when needed.

## Related issue(s)

- fixes #1798

## How has this been tested?
-   [ ] _Test case 1_
    1. Doesn't break anything. Works the same. Exported so other parts of the library can consume it as needed.\

## Types of changes
-   [x] Refactor

## Checklist

### DIFF
--- a/packages/overlay/src/overlay-stack.ts
+++ b/packages/overlay/src/overlay-stack.ts
@@ -22,6 +22,7 @@ import {
     parentOverlayOf,
 } from './overlay-utils.js';
 import { OverlayCloseEvent } from './overlay-events.js';
+import { getDeepElementFromPoint } from '@spectrum-web-components/shared/src/get-deep-element-from-point.js';
 
 function isLeftClick(event: MouseEvent): boolean {
     return event.button === 0;
@@ -184,19 +185,9 @@ export class OverlayStack {
         event.stopPropagation();
         event.preventDefault();
         await this.closeTopOverlay();
-        let target = document.elementFromPoint(event.clientX, event.clientY);
-        while (target?.shadowRoot) {
-            const innerTarget = (
-                target.shadowRoot as unknown as {
-                    elementFromPoint: (x: number, y: number) => Element | null;
-                }
-            ).elementFromPoint(event.clientX, event.clientY);
-            if (!innerTarget || innerTarget === target) {
-                break;
-            }
-            target = innerTarget;
-        }
-        target?.dispatchEvent(new MouseEvent('contextmenu', event));
+        getDeepElementFromPoint(event.clientX, event.clientY)?.dispatchEvent(
+            new MouseEvent('contextmenu', event)
+        );
     };
 
     private get document(): Document {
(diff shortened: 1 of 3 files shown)


packages/shared/README.md
```diff
@@ -23,6 +23,10 @@ import {
 } from '@spectrum-web-components/shared';
 ```
 
+### getDeepElementFromPoint
+
+The `getDeepElementFromPoint` method allows you to obtain the deepest possible element at a given coordinates on the current page. The method will step into any available `shadowRoot`s until it reaches the first element with no `shadowRoot` or no children available at the given coordinates.
```
castastrophe, line 28: Wow! Very cool.

packages/shared/src/get-deep-element-from-point.ts
```diff
@@ -0,0 +1,30 @@
...
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+export const getDeepElementFromPoint = (
+    x: number,
+    y: number
+): Element | null => {
+    let target = document.elementFromPoint(x, y);
+    while (target?.shadowRoot) {
```
castastrophe, line 18: Would it be valuable to include an escape hatch via a maximum depth? Could even put the maximum depth at something like 100 but it seems like a nice protection.

packages/overlay/src/overlay-stack.ts
```diff
@@ -184,19 +185,9 @@ export class OverlayStack {
...
-                    elementFromPoint: (x: number, y: number) => Element | null;
-                }
-            ).elementFromPoint(event.clientX, event.clientY);
-            if (!innerTarget || innerTarget === target) {
-                break;
-            }
-            target = innerTarget;
-        }
-        target?.dispatchEvent(new MouseEvent('contextmenu', event));
+        getDeepElementFromPoint(event.clientX, event.clientY)?.dispatchEvent(
```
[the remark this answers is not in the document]
Westbrook: This callback for a mouse click that is using the “right” button (pretty sure you can’t do that and scroll at the same time), and has yet to surface a need for a debounce. Could definitely see that being useful
In other cases, however.

packages/shared/src/get-deep-element-from-point.ts
Westbrook: I’m not sure of a lot of cases where this is actually functionality that a developer actually means to be leveraging and without more than the one we currently support I don’t want to over optimize here.

Issue #2086, opened 2022-02-01
[Feat]: Abstract Overlay API Review


Westbrook:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

overlay

## openOverlay currently accepts:
```
{
    target: HTMLElement,
    interaction:
      | 'click'
      | 'longpress'
      | 'hover'
      | 'custom'
      | 'replace'
      | 'inline'
      | 'modal',
    content: HTMLElement,
    options: {
      delayed?: boolean;
      placement?: Placement;
      offset?: number;
      receivesFocus?: 'auto';
      notImmediatelyClosable?: boolean;
      abortPromise?: Promise<boolean>;
      virtualTrigger?: VirtualTrigger;
    }
}
```

A content element managed for overlay looks like:
```
interface ManagedOverlayContent {
    open: boolean;
    overlayWillOpenCallback?: (args: { trigger: HTMLElement }) => void;
    overlayOpenCallback?: (args: { trigger: HTMLElement }) => void;
    overlayOpenCancelledCallback?: (args: { trigger: HTMLElement }) => void;
    overlayCloseCallback?: (args: { trigger: HTMLElement }) => void;
}
```

## Important question
If we created a new API could we smooth transition by wrapping it with the old API for the initial term? 

## Things we might want instead:
`content` is required but `trigger` is not
  - points to a usage of the "Active Stack" concept where the "trigger" was the previous "active element" and that content would manage things like theme, focus return, etc.

tabOrder: 'none' | 'inline' | 'replace' | 'trap'
  - 'none' = tooltip and similar that are given `aria-*` based content relations
  - 'inline' = before trigger <=> trigger <=> overlay <=> after overlay
  - 'replace' = before trigger <=> trigger when closed, overlay when open <=> after overlay
  - 'trap' = modal
This should absolve the need for `receivedFocus` as all overlays need focus for the screen reader to step into them when `aria-*` attributes are not managing the content relations.

interaction:
  - likely the only interaction not managed by `tabOrder` is `longpress` which really only manages the first click event. 

backdrop management:
  - not all "modals" have a backdrop and we need to manage it somehow
  - it may be that backdrops need to be handled 100% of the time by the overlaid content and not by the overlay system

stack management:
  - overlays need access to the stack to prevent DOM based state resolution
  - adding/removing something from the stack should be synchronous
[...]

PR #2395, opened 2022-06-30
fix(overlay): export OverlayTriggerInteractions type

beeduul:
Expose overlay trigger interactions as a type, so UE account menu can provide user-customizable overlay trigger behaviors while providing type safety of compatible options.

## Related issue(s)

-

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/overlay/src/OverlayTrigger.ts
+++ b/packages/overlay/src/OverlayTrigger.ts
@@ -31,6 +31,7 @@ import {
 import {
     OverlayOpenCloseDetail,
     OverlayOptions,
+    OverlayTriggerInteractions,
     Placement,
     TriggerInteractions,
 } from './overlay-types';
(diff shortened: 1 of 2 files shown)


beeduul:
Good point about pre-1.0 semver behavior.  This is not a priority fix as there is a local patch in the UEC account menu, so we could wait for #2086 and even 1.0 before merging this in (and then updating UEC account menu).

castastrophe:
@beeduul Alternatively, you can capture the type for Account Menu this way:

```
import type { OverlayTrigger } ...

export class AccountMenu extends SpectrumElement {
...
   public overlayTriggerInteraction: OverlayTrigger['type'] = 'modal';
```

Westbrook (approved):
Thanks for cleaning up these points here! LGTM. 🚢 it!

PR #2407, opened 2022-07-11
chore: icon workflow processing script

castastrophe:
## Description

To gain a better view of assets being generated, I updated the icon workflow build script to leverage templates instead of writing to a file line-by-line.

## Motivation and context

- Leverage async where possible
- Fetch data from the provided icon package's manifest
- Resolve pathing using require instead of pointing to node_modules
- Utilize ejs templating to make updates/maintenance easier in future

## How has this been tested?

-   [x] `yarn workspace @spectrum-web-components/icons-workflow build`
    1. Same assets generated in this branch as in `main`
    2. Caveat: added an underscore prefix on numeric-leading variable names; results in 2 additional icons being added.

## Checklist

<!--- Go over all the following points, and put an `x` in all the boxes that apply.  If you're unsure about any of these, don't hesitate to ask. We're here to help! —>

-   [N/A] I have added tests to cover my changes.

## Best practices


castastrophe:
I think there's a definite opportunity here to abstract this into a tool for folks to use to dynamically generate web components from SVG assets. 🤔

Westbrook (commented):
Starting to come together here. I'd love to see this in the `icons-ui` package as well, and then we could do some perf comparison between the old and the new approaches here.

castastrophe:
@Westbrook would love to tackle that as well but since they run separately, I thought it prudent to present them in different PRs. You can see running the script solo before & after that they're both quite quick. The benefit of this approach is in maintainability (easier to update templates in future) and in leveraging the provided manifest from Spectrum rather than grepping for *.svg assets.

PR #2439, opened 2022-07-22
chore: clean up folders leftover during post-checkout

castastrophe:
## Description

When a new component is created and then the developer moves on to a new branch, sometimes a `leftover` component folder remains behind due to the untracked node_modules folder which prevents git from fully cleaning up after the branch change.

## Motivation and context

Prevents ghost folders from staying behind as you develop across multiple branches.

## How has this been tested?

-   [x] _Validate identification & clean-up of leftover folders_
    1. (from root) `mkdir packages/foo/node_modules`
    2. `sh .husky/post-checkout` - should remove the `packages/foo` folder

## Types of changes

## Checklist

-   [x] I have read the **[CONTRIBUTING](<(https://git.corp.adobe.com/DIWI/unified-experience-components/blob/main/CONTRIBUTING.md)>)**
        document.

### DIFF
--- a/.husky/post-checkout
+++ b/.husky/post-checkout
@@ -1,0 +1,18 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+# If this is a file checkout (== 0) exit.
+if [ "$3" == "0" ]; then exit; fi
+
+# If this is a branch checkout (== 1) continue...
+
+# For each element, check for a node_modules dir without a corresponding
+# package.json.  That indicates the element does not exist in this branch.
+for e in {packages,projects,tools}/*; do
+  if [[ -d $e/node_modules && !( -f $e/package.json ) ]]; then
+    # Safe to delete this folder
+    rm -rf $e
+  fi
+done
+
+exit 0


.husky/post-checkout
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+# For each element, check for a node_modules dir without a corresponding
+# package.json.  That indicates the element does not exist in this branch.
+for e in {packages,experiments}/*; do
```
Westbrook, line 6: We use `packages,projects,tools`.

Westbrook (commented):
I think I like this, but I don't quite get this test:
```
if [[ -d $e/node_modules && !( -f $e/package.json ) ]]; then
```
Does that mean the delete only happens when there is a `node_modules`? Or something.

.husky/post-checkout
castastrophe: doh! I'll fix.  Just did a mindless copy-pasta.

castastrophe:
> Does that mean the delete only happens when there is a node_modules? Or something.

It checks a folder to see if it has a `node_modules` folder but not a package.json file.  If that is the case, it's 99% likely it's a leftover.

castastrophe:
@Westbrook Fixed the folder list

Westbrook:
In these directories, should any subdirectory ever NOT have a `package.json`? Wondering if we could be more broad and still be safe being I feel like `node_modules` directories are much less common in specific package directories here than in other of our projects.

castastrophe:
> In these directories, should any subdirectory ever NOT have a package.json? Wondering if we could be more broad and still be safe being I feel like node_modules directories are much less common in specific package directories here than in other of our projects.

I think as long as you're checking that the `node_modules` and `package.json` do or do not exist at the same depth, it's safe. Since it's doing a recursive delete, I think it makes sense to keep it pretty tightly scoped to these workspaces.

Westbrook:
Maybe it's not possible, but I'm trying to see if this would cover a more common use case in this repo, but I may also not quite understand this workflow. Is there a way to have this support the following?

```bash
git checkout chore-leftover-folder-check
yarn # sort of a noop
git checkout table
yarn # creates ignored artifacts in /packages/table, but not a node_modules directory
git checkout chore-leftover-folder-check
yarn
# /packages/table continues to exist because it never had a node_modules directory
```

Westbrook:
The main thing here is that the `table` branch, and really any branch that is ahead of `main` that features new packages will get a package directory, but no `node_module` directory preventing it from being deleted. Simply taking out `-d $e/node_modules && ` is a a little greedy, I tried it out locally, but we could shore that up by giving `projects/vrt-quick-link` a `package.json`. There's likely a better contract to ensure the deletion of these directories, do you have any other suggestions there? It it possible to check for git tracked files in some way, or maybe we could just for `*.md`, `*.ts`, and `*.html`? I do really like what this looks to bring to the table, but it's not quite doing it at a regular enough basis, yet.

castastrophe:
Let me make sure I understand the workflow you're describing:
1. git checkout -b new-branch
2. build new component, do development, etc.
3. git checkout some-other-branch

As I see when I test this, as long as development has happened in step 2, a node_modules directory will be created in the table folder but the package.json will disappear when switching to the new branch because it is a tracked file. The only reason node_modules hangs out is because it is untracked.

Is the workflow you're describing one in which an install has not been done and thus there has not been a node_modules folder created in the folder? Without the untracked node_modules folder nested, I think the entire folder goes away when you switch branches because it gets cleaned up by git.

PR #2442, opened 2022-07-25
test: ensure Lit Dev Mode warnings are not emitted

Westbrook:
## Description
Test default package consumption patterns to ensure they do not emit Lit Dev Mode warnings.

## Related issue(s)
- refs #2257 

## Motivation and context
Safe guard the return of possibly perf negative patterns.

## How has this been tested?

-   [ ] _Test case 1_
    1. Tests run
    2. No failures
-   [ ] _Test case 2_
    1. Undo the changes in https://github.com/adobe/spectrum-web-components/commit/b2c393ef5da68cb832cbf6cddb1b96c67423fb80
    2. Run the `action-button` units tests
    3. See them fail

## Types of changes
-   [ ] Tests

## Checklist


packages/underlay/test/underlay.test.ts
```diff
@@ -14,8 +14,17 @@ import { elementUpdated, expect, fixture, html } from '@open-wc/testing';
...
+import { testForLitDevWarnings } from '../../../test/testing-helpers.js';
 
 describe('Underlay', () => {
+    testForLitDevWarnings(
+        async () =>
+            await fixture<Underlay>(
+                html`
+                    <sp-underlay></sp-underlay>
+                `
+            )
```
hunterloftis, line 26: For all of these calls, could you get away with omitting the wrapped `async` & `await`, since fixture is already returning a Promise? eg:

```suggestion
        () => fixture<Underlay>(
                html`
                    <sp-underlay></sp-underlay>
                `
            )
```

hunterloftis (approved):
lgtm! Left a small suggestion that might shorten these but it isn't important.

packages/underlay/test/underlay.test.ts
Westbrook: Any chance that would be faster? I was going for uniformity, and by that I mean "copy and paste"-ability.

hunterloftis: I'm pretty sure it wouldn't be measurably faster. Just changes where the await happens (inside the fn or outside).

PR #2448, opened 2022-07-26
chore: alias storybook command to 'start'

castastrophe:
## Description

Adding an alias for the `yarn storybook` command: `yarn start`

## Motivation and context

A `start` command is a nice feature because it describes what the developer is doing rather than the tool they are using.  We are using storybook for viewing development but the task we are performing is development. Just a personal preference but hoping it makes it easier to standardize commands across projects when different development tools might be in use.

## How has this been tested?

-   [x] _yarn start_
    1. Calls & runs `yarn storybook`

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,6 +63,7 @@
         "spectrum-css": "node ./scripts/process-spectrum-css.js",
         "spectrum-tokens": "node ./scripts/spectrum-tokens.js",
         "spectrum-vars": "node ./scripts/spectrum-vars.js",
+        "start": "yarn storybook",
         "storybook": "yarn build:ts && yarn build:css && run-p storybook:start watch:css build:watch",
         "storybook:build": "build-storybook -o projects/documentation/dist/storybook",
         "storybook:start": "web-dev-server --config wds-storybook.config.js",


Westbrook:
Seems like a decent idea. In a realm of abstraction and normalization would it be beneficial to pivot to the `start` script that is one of those prioritized by the `npm` ecosystem as not needing `run` and as such has a good amount of head space there? `dev` is interesting, however in a world of of "*DD" techniques, Storybook Driven Development, Test Driven Development, Resume Driven Development, Documentation Driven Development, etc. `start` may adjust better to changes experience/preferences of the team involved.

castastrophe:
@Westbrook I like it. I went with `dev` originally over `start` because I had this feeling like `start` was purely for the server but not necessarily includes `watch`. That is pretty typical of `start` commands but I'm not sure it should be!

package.json
```diff
@@ -24,6 +24,7 @@
         "build:ts:clean": "yarn build:ts --clean",
         "build:watch": "yarn build:ts --watch",
         "custom-element-json": "lerna exec --ignore \"{@spectrum-web-components/{base,bundle,modal,iconset,shared,styles},documentation,example-project-rollup,example-project-webpack,swc-templates}\" -- cem analyze --config ../../custom-elements-manifest.config.js --packagejson",
+        "dev": "yarn storybook",
```
castastrophe, line 27:
```suggestion
        "start": "yarn storybook",
```

castastrophe:
I'll make a note to update UEC as well if this merges so we'll be aligned on the script name

Westbrook (approved):
LGTM

Issue #2480, opened 2022-08-04
[Bug]: submenu items within sp-action-menu lose their `selected` states


hunterloftis:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

action-menu, picker, menu-item

### Expected behavior

An `<sp-menu-item selected>` remains selected when part of a `slot="submenu"` menu within an `<sp-action-menu>`.

### Actual behavior

`<sp-menu-item selected>` has its selected state stripped when rendered as a child of an `<sp-action-menu>`.

### Screenshots

![image](https://user-images.githubusercontent.com/364501/182910468-d8a41298-c375-4364-a5bb-e8a872dac83d.png)

### How can we reproduce this issue?

https://studio.webcomponents.dev/edit/3Aq0lreH6YVKOfdUNcmx/src/index.ts?p=stories

hunterloftis:
Closed by https://github.com/adobe/spectrum-web-components/pull/2481

PR #2481, opened 2022-08-04
fix(action-menu): stop stripping selected state from submenu items

hunterloftis:
## Description

New behavior: Omit `submenu` items from the `menuItems` list managed by `PickerBase`.

## Related issue(s)

https://github.com/adobe/spectrum-web-components/issues/2480

## Motivation and context

A side-effect of the existing behavior (managing _all_ menu items) is that nested submenus have their selection states removed.

## How has this been tested?

1. Go to [Storybook](https://62ec14b045349f19bd39b6b0--spectrum-web-components.netlify.app/storybook/?path=/story/action-menu--submenu)
2. Toggle the selections in the submenu
---
1. Regression test added to suite 
2. All existing tests pass

image

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/action-menu/stories/action-menu.stories.ts
+++ b/packages/action-menu/stories/action-menu.stories.ts
@@ -99,3 +99,20 @@ customIcon.args = {
     `,
     visibleLabel: '',
 };
+
+export const submenu = (): TemplateResult => {
+    return html`
+        <sp-action-menu label="More Actions">
+            <sp-menu-item>One</sp-menu-item>
+            <sp-menu-item>Two</sp-menu-item>
+            <sp-menu-item>
+                Select some items
+                <sp-menu slot="submenu" selects="multiple">
+                    <sp-menu-item>A</sp-menu-item>
+                    <sp-menu-item selected>B</sp-menu-item>
+                    <sp-menu-item>C</sp-menu-item>
+                </sp-menu>
+            </sp-menu-item>
+        </sp-action-menu>
+    `;
+};
--- a/packages/action-menu/test/action-menu.test.ts
+++ b/packages/action-menu/test/action-menu.test.ts
@@ -24,6 +24,8 @@ import {
     oneEvent,
 } from '@open-wc/testing';
 import { testForLitDevWarnings } from '../../../test/testing-helpers';
+import type { MenuItem } from '@spectrum-web-components/menu/src/MenuItem.js';
+import type { Menu } from '@spectrum-web-components/menu';
 
 const deprecatedActionMenuFixture = async (): Promise<ActionMenu> =>
     await fixture<ActionMenu>(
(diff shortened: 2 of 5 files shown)


packages/action-menu/stories/action-menu.stories.ts
```diff
@@ -99,3 +99,20 @@ customIcon.args = {
...
+    return html`
+        <sp-action-menu label="More Actions">
+            <sp-menu-item>One</sp-menu-item>
+            <sp-menu-item>Two</sp-menu-item>
+            <sp-menu-item>
+                Select some items
+                <sp-menu slot="submenu" selects="multiple">
+                    <sp-menu-item>A</sp-menu-item>
+                    <sp-menu-item>B</sp-menu-item>
+                    <sp-menu-item>C</sp-menu-item>
```
Westbrook, line 113: Should one of these be selected by default...as proof?

packages/action-menu/test/action-menu.test.ts
```diff
@@ -24,6 +24,8 @@ import {
     oneEvent,
 } from '@open-wc/testing';
 import { testForLitDevWarnings } from '../../../test/testing-helpers';
+import { MenuItem } from '@spectrum-web-components/menu/src/MenuItem.js';
+import { Menu } from '@spectrum-web-components/menu';
```
Westbrook, line 28:
```suggestion
import type { MenuItem } from '@spectrum-web-components/menu/src/MenuItem.js';
import type { Menu } from '@spectrum-web-components/menu';
```

packages/picker/src/Picker.ts
```diff
@@ -539,7 +539,9 @@ export class PickerBase extends SizedMixin(Focusable) {
                 this.menuItems = this.optionsMenu.childItems;
             } else {
                 this.menuItems = [
-                    ...this.querySelectorAll('sp-menu-item'),
+                    ...this.querySelectorAll(
```
Westbrook, line 542: Did you end up liking this more than the gating? There's _something_ to be said about a single selection Picker that has submenus...but I'm not sure that was accurately covered previously, either.

hunterloftis: Personally I like the gating more (simpler conceptually - something with no `selects` behavior doesn't run `manageSelection`), but at the root level, items are selectable by default. With the gate:

image

Westbrook: With the selector approach, can you confirm that this works there, too? https://studio.webcomponents.dev/edit/guTpKeAjgecibF150Qbg/src/index.ts?branch=group%40wfVpXdsTYhf9mY2slxW4t2Y2SjC3&p=stories Might just need some massage to get it working, but want to ensure we're not regressing on anything that might not be accurately covered in the tests.

hunterloftis: Good call, take a look at the latest commit - I added that sketch as a test so we can be sure to not regress it, & loosened up the query a little to support it.

Westbrook: That's a fancy selection. I like it. Don't usually use `:not()` in that way, but looks powerful. :clap:

Westbrook (approved):
Nice catch here!

PR #2508, opened 2022-08-11
test(table): fix a11y error by adding tab stop on tablebody

najikahalsema:
## Description

Added a tab stop to the TableBody upon rendering the Virtualised Table so that the a11y test no longer flakes. 

## Motivation and context

Moving it closer to going into main :) 

## How has this been tested?
Visually, and with the a11y test passing.

-   [ ] _Test case 1_
    1. Go to storybook
    2. Tab into the page
    3. See that you can't tab into the tableBody
-   [ ] _Test case 2_
    1. Go to Tachometer Storybook
    2. Tab into the page
    3. See that you CAN, in fact, tab into the tableBody! 

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/table/src/Table.ts
+++ b/packages/table/src/Table.ts
@@ -453,7 +453,7 @@ export class Table extends SizedMixin(SpectrumElement, {
             this.tableBody = this.querySelector('sp-table-body') as TableBody;
             if (!this.tableBody) {
                 this.tableBody = document.createElement('sp-table-body');
-                this.tableBody.tabIndex = 0;
+                if (this.scroller) this.tableBody.tabIndex = 0;
                 this.append(this.tableBody);
             }
             this.tableBody.addEventListener(
--- a/packages/table/src/Table.ts
+++ b/packages/table/src/Table.ts
@@ -453,6 +453,7 @@ export class Table extends SizedMixin(SpectrumElement, {
             this.tableBody = this.querySelector('sp-table-body') as TableBody;
             if (!this.tableBody) {
                 this.tableBody = document.createElement('sp-table-body');
+                this.tableBody.tabIndex = 0;
                 this.append(this.tableBody);
             }
             this.tableBody.addEventListener(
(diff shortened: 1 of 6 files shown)


packages/table/src/Table.ts
```diff
@@ -447,6 +447,7 @@ export class Table extends SizedMixin(SpectrumElement, {
             this.tableBody = this.querySelector('sp-table-body') as TableBody;
             if (!this.tableBody) {
                 this.tableBody = document.createElement('sp-table-body');
+                this.tableBody.tabIndex = 0;
```
Westbrook, line 450: Is this needed in the case that the items do not force the Table Body to scroll?

packages/table/src/Table.ts
```diff
@@ -447,7 +447,7 @@ export class Table extends SizedMixin(SpectrumElement, {
             this.tableBody = this.querySelector('sp-table-body') as TableBody;
             if (!this.tableBody) {
                 this.tableBody = document.createElement('sp-table-body');
-                this.tableBody.tabIndex = 0;
+                if (this.scroller) this.tableBody.tabIndex = 0;
```
Westbrook, line 450: This is closer, but if you have a Table that is 1000px high and items that only add up to 500px there will be no scrollbar. We can confirm that with something like `element.offsetHeight < element.scrollHeight`. It's sadly something that likely needs an observer as it'll be something that could change over time... 🤔

packages/table/test/virtualized-table.test.ts
```diff
@@ -59,6 +59,9 @@ describe('Virtualized Table', () => {
 
     it('loads virtualized table accessibly', async () => {
         const el = await fixture<Table>(virtualized());
+        await nextFrame();
+        await nextFrame();
+        await nextFrame();
```
Westbrook, line 64: Can we add this to the non-virtualized tests as well?

Issue #2456, opened 2022-07-27
[Feat]: Table Phase 2


Westbrook:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Table

### Description of the requested feature

- [x] clarify "Selection" documentation by removing the non-selectable table from that section
- [ ] add Shift + Click selection to add "many" items at once
  - review the keyboard accessible equivalent to this in Tree View
- [x] review scroll stopping experienced in https://github.com/adobe/spectrum-web-components/pull/2365#issuecomment-1161836202 likely related to a bug specifically in Chrome 102
- [x] correct the item shape in the Virtualized demos
- [ ] review keyboard navigation technique and whether we want to use a Roving Tab index or something else...
- [x] surface `table-elements.js` export
- [ ] document drag and drop options
- [ ] better support for non-virtual rows across the lifecycle of a Table (adding/removing rows, etc)
- [ ] better documentation around sorting interface

najikahalsema:
Looks good! could you tlel me more about the table-elements.js export?

Westbrook:
Currently to build a table you need to import all of these things:
```
import '@spectrum-web-components/table/sp-table.js';
import '@spectrum-web-components/table/sp-table-body.js';
import '@spectrum-web-components/table/sp-table-cell.js';
import '@spectrum-web-components/table/sp-table-checkbox-cell.js';
import '@spectrum-web-components/table/sp-table-head.js';
import '@spectrum-web-components/table/sp-table-head-cell.js';
import '@spectrum-web-components/table/sp-table-row.js';
```
So many imports are easy to mess up. I _think_ we should export a helper file that registers all of those to simplify consumption by client developers.
```
import '@spectrum-web-components/table/table-elements.js';
```

SavoInfostream:
Also add the `@spectrum-web-components/table` to the `index.d.ts`?

castastrophe:
We have this work tracked now in Jira so I'm going to close. Please do reach out if you still need a status update on this feature.

PR #2509, opened 2022-08-11
Table Phase 2 Additions

Westbrook:
## Description
- toggle selection on row clicks
- expose `elements.js` as an export
- expand docs
- bind the virtualized list in a way that doesn't create run away listeners
- test coverage and clean up post refactor

## Related issue(s)
- refs #2456 

## How has this been tested?
-   [ ] _Test case 1_
    1. Go [here](https://table--spectrum-web-components.netlify.app/components/table/)
    2. Read the docs
-   [ ] _Test case 2_
    1. Go [here](https://app.circleci.com/pipelines/github/adobe/spectrum-web-components/12737/workflows/635f986d-8cd1-47d5-b47a-d9d58b38ea41)
    2. See that the unit tests are passing
    3. VRTs will come later

## Types of changes\
-   [x] Feature work

## Checklist

### DIFF
--- a/packages/table/README.md
+++ b/packages/table/README.md
@@ -14,6 +14,12 @@ yarn add @spectrum-web-components/table
 Import the side effectful registration of `<sp-table>`, `<sp-table-body>`, `<sp-table-cell>`, `<sp-table-checkbox-cell>`, `<sp-table-head>`, `<sp-table-head-cell>`. and `<sp-table-row>` via:
 
 ```
+import '@spectrum-web-components/table/elements.js';
+```
+
+Or individually via:
+
+```
 import '@spectrum-web-components/table/sp-table.js';
 import '@spectrum-web-components/table/sp-table-body.js';
 import '@spectrum-web-components/table/sp-table-cell.js';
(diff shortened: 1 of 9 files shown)


packages/table/src/Table.ts
```diff
@@ -443,6 +446,9 @@ export class Table extends SizedMixin(SpectrumElement, {
     }
 
     protected renderVirtualizedItems(index?: number): void {
+        // Rendering updates into the table while disconnected can
+        // cause runaway event binding in ancestor elements.
```
najikahalsema, line 450: I'd love to hear more about this disconnected callback addition and how it adds to the robustness of the code! Is this related to the event binding we talked about before?

packages/table/src/TableHeadCell.ts
```diff
@@ -95,7 +95,7 @@ export class TableHeadCell extends SpectrumElement {
     }
 
     protected override update(changes: PropertyValues): void {
-        if (changes.has('sorted')) {
+        if (changes.has('sortDirection')) {
```
najikahalsema, line 98: Nice catch!

packages/table/README.md
```diff
@@ -166,7 +137,7 @@ To manage selection on an `<sp-table>`, utilise the `selects` attribute on `<sp-
 
 ### `selects="multiple"`
 
-When `selects` is set to "multiple", the `<sp-table-checkbox-cell>` in `<sp-table-head>` acts as the select/deselect all button.
+When `selects="multiple"` the `<sp-table>` will manage a selection in the array value of `selected` in via a presence toggle. Further, an `<sp-table-checkbox-cell>` will be made available in the `<sp-table-head>` in order to select/deselect all items in the `<sp-table>`.
```
najikahalsema, line 140:
```suggestion
When `selects="multiple"`, the `<sp-table>` will manage a selection in the array value of `selected` via a presence toggle. Additionally, an `<sp-table-checkbox-cell>` will be made available in the `<sp-table-head>` in order to select/deselect all items in the `<sp-table>`.
```

The first sentence of this statement is a little bit confusing. What does "via a presence toggle" mean?

packages/table/README.md
```diff
@@ -329,6 +300,207 @@ const renderItem = (item: Item, index: number): TemplateResult => {
 
 Please note that there is a bug when attempting to select all virtualised elements. The items are selected programatically, it's just not reflected visually.
 
+### Selection
+
+When making a selection on a virtualized table, it can sometimes be useful to track that selection as something other than an array of indexes. To make this possible the `itemValue` property will accept a method who argument is an item object, return a computed string value to track selection with this value rather than the item's index:
```
najikahalsema, line 305:
```suggestion
When making a selection on a virtualized table, it can sometimes be useful to track that selection as something other than an array of indexes. To make this possible, the `itemValue` property accepts a method whose argument is an`Item` object. It then returns a computed string value to track selection with this value instead of the item's index:
```

I think this wording is clearer. Not sure about the change from item to `Item`, but I wanted to check if that's what you meant. In that case, this styling provides more clarity for me.

packages/table/README.md
```diff
@@ -329,6 +300,207 @@ const renderItem = (item: Item, index: number): TemplateResult => {
...
+        });
+    };
+    customElements.whenDefined('sp-table').then(() => {
+        initTable();
+    });
+</script>
+
+### Row Types
+
+All values in the item array are assumed to be homogenous by default. This means all of the rendered rows are either delivered as provided, or, in the case you are leveraging `selects`, rendered with an `<sp-table-checkbox-cell>`. However, when virtualizing a table with selection it can sometimes be useful to surface rows with additional interactions, e.g. "Load more data" links. To support that you can optionally include the `_$rowType$` brand in your item. The values for this are outlined by the `RowType` enum and include `ITEM` (0) and `INFORMATION` (1). When `_$rowType$: RowType.INFORMATION` is provided it informs the `<sp-table>` element not to deliver an `<sp-table-checkbox-cell>` with that row.
```
najikahalsema, line 403:
```suggestion
All values in the item array are assumed to be homogenous by default. This means all of the rendered rows are either delivered as provided, or, in the case you are leveraging `selects`, rendered with an `<sp-table-checkbox-cell>`. However, when virtualizing a table with selection, it can sometimes be useful to surface rows with additional interactions, e.g. "Load more data" links. To support that, you can optionally include the `_$rowType$` brand in your item. The values for this are outlined by the `RowType` enum and include `ITEM` (0) and `INFORMATION` (1). When `_$rowType$: RowType.INFORMATION` is provided, it informs the `<sp-table>` element that it should not deliver an `<sp-table-checkbox-cell>` in that row.
```

Suggestions for clarity

najikahalsema (changes requested):
Looks good! :D I would make some changes to the wording in the documentation, as it was a little difficult for me to understand at times. The actual code itself looks great. Really appreciate your awesome additions to the element!

packages/table/src/Table.ts
Westbrook: This is not exactly as I expected, _but_ when you render this list while the element is disconnected, which is not something we were planning for initially, it will render a list into the ether that adds event listeners, but can never haven them removed. Remove and attach the Table a bunch of times and you end up with a lot of those!

This gets worse when you don't include the code in `disconnectedCallback()`. There we render to "nothing" which is needed to get actual event cleanup.

So now, instead of adding two extra listeners every time the Table is removed and added back, we clean up the first event listeners, don't add uncontrolled listeners and then add back the needed listers when reconnecting.

Westbrook:
@najikahalsema took you notes, expanded it to some similar contexts, and in some cases hopefully found a middle ground that's understandable to the reader in https://github.com/adobe/spectrum-web-components/pull/2509/commits/872dde57788ea6107804901eb039e04dfb881b08

najikahalsema (approved):

PR #2545, opened 2022-08-26
fix(overlay): move "escape" listener to "keydown"

### LINKED ISSUE #1844: [Bug]: SWC uses keyup handlers more often than it should
### Code of conduct

### Impacted component(s)

Overlay and many more

### Expected behavior

SWC should default to using keydown events unless there's a specific reason to use keyup.

### Actual behavior

There's many places where SWC handles events with keyup. This makes it difficult if you have another handler for that key registered to keydown. Rather than being able to use event bubbling you'd need to add specific logic based on the handling SWC is doing.

The example we hit was the escape key handling in the overlay implementation. We have a keydown handler that was triggering for a component under the overlay.

### Screenshots

_No response_

_No response_

3. Scroll to '....'
4. Check console
5. See error

_No response_

_No response_

Westbrook:
## Description
Move `Escape` listener to `keydown` to match native `<modal>` interaction.

## Related issue(s)

- fixes #1844

## How has this been tested?
-   [ ] _Test case 1_
    1. With updated unit tests.

## Types of changes

## Checklist

### DIFF
--- a/packages/overlay/src/ActiveOverlay.ts
+++ b/packages/overlay/src/ActiveOverlay.ts
@@ -223,7 +223,10 @@ export class ActiveOverlay extends SpectrumElement {
     }
 
     public feature(): void {
-        this.tabIndex = -1;
+        // eslint-disable-next-line spectrum-web-components/document-active-element
+        if (!this.contains(document.activeElement)) {
+            this.tabIndex = -1;
+        }
         const parentOverlay = parentOverlayOf(this.trigger);
         const parentIsModal = parentOverlay && parentOverlay.slot === 'open';
         if (parentIsModal) {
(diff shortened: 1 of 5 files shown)


packages/overlay/src/ActiveOverlay.ts
```diff
@@ -223,7 +223,10 @@ export class ActiveOverlay extends SpectrumElement {
     }
 
     public feature(): void {
-        this.tabIndex = -1;
+        // eslint-disable-next-line spectrum-web-components/document-active-element
+        if (!this.contains(document.activeElement)) {
```
Westbrook, line 227: Allows you to tab through https://overlay-escape--spectrum-web-components.netlify.app/storybook/?path=/story/dialog-wrapped--tooltips in both forward and backward directions without getting blocked out of the Dialog after leaving the iframe...

najikahalsema (approved):

PR #2569, opened 2022-09-08
refactor(badge): use core tokens

lunarfusion:
This updates the Badge component to use core tokens

Uses beta release. Hold for full release.

## Related issue(s)

-

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/badge/README.md
+++ b/packages/badge/README.md
@@ -102,13 +102,81 @@ The `<sp-badge>` can be customized with either semantic or non-semantic variants
 
 ### Non-Semantic
 
+Non-semantic badge colors are no longer supported directly by Spectrum and Spectrum Web Components. You can mimic their delivery via the following CSS Custom Properties. These values for the `variant` attribute/property have not been marked as deprecated, but will no longer achieve the results you may have relied on in the past.
+
 ```html demo
 <div style="display: flex; gap: var(--spectrum-global-dimension-size-50);">
-    <sp-badge variant="fuchsia">Fuchsia</sp-badge>
-    <sp-badge variant="indigo">Indigo</sp-badge>
-    <sp-badge variant="magenta">Magenta</sp-badge>
-    <sp-badge variant="purple">Purple</sp-badge>
-    <sp-badge variant="seafoam">Seafoam</sp-badge>
-    <sp-badge variant="yellow">Yellow</sp-badge>
+    <sp-badge
+        variant="seafoam"
+        style="--mod-badge-background-color-default: var(--spectrum-global-color-static-seafoam-600)"
+    >
+        Seafoam
+    </sp-badge>
+    <sp-badge
+        variant="indigo"
+        style="--mod-badge-background-color-default: var(--spectrum-global-color-static-indigo-600)"
+    >
+        Indigo
+    </sp-badge>
+    <sp-badge
+        variant="purple"
+        style="--mod-badge-background-color-default: var(--spectrum-global-color-static-purple-600)"
+    >
+        Purple
+    </sp-badge>
+    <sp-badge
+        variant="fuchsia"
+        style="--mod-badge-background-color-default: var(--spectrum-global-color-static-fuchsia-600)"
+    >
+        Fuchsia
+    </sp-badge>
+    <sp-badge
+        variant="magenta"
+        style="--mod-badge-background-color-default: var(--spectrum-global-color-static-magenta-600)"
+    >
+        Magenta
+    </sp-badge>
+    <sp-badge
+        variant="yellow"
+        style="--mod-badge-background-color-default: var(--spectrum-alias-background-color-yellow-default); --mod-badge-label-icon-color-white: var(--spectrum-global-color-static-black);"
+    >
+        Yellow
+    </sp-badge>
+</div>
+```
+
+## Fixed
+
+When you'd like to have the `<sp-badge>` display as if "fixed" to the edge of a UI, the `fixed` attribute/property can be leveraged to alter the border rounding based on the position you would like to achieve:
+
+```html
+<div
+    style="position: relative; width: 400px; height: 200px; background: #eee; max-width: 100%"
+>
+    <sp-badge>None</sp-badge>
+    <sp-badge
+        fixed="block-start"
+        style="position: absolute; top: 0; left: 200px;"
+    >
+        block-start
+    </sp-badge>
+    <sp-badge
+        fixed="inline-end"
+        style="position: absolute; right: 0; top: 100px;"
+    >
+        inline-end
+    </sp-badge>
+    <sp-badge
+        fixed="block-end"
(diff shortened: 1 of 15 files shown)


lunarfusion:
It appears that all the failures are likely due to changed markup that adds a "spectrum-Badge-label" element to control the spacing required by the addition of the icon variant. 

Screen Shot 2022-09-12 at 10 32 39 AM

The markup of web-components will need to be updated to reflect the correct styles. Please advise if I should make any changes.

@Westbrook

Westbrook:
There are a lot of DOM changes here, yes. @pfulton this makes the Badge even more like the Action Button, I know we don't usually share code this way, but should we share code here?

With the differences, would it be better for us to take this one on, or would it make sense to pair a little on it or?

pfulton:
> There are a lot of DOM changes here, yes. @pfulton this makes the Badge even more like the Action Button, I know we don't usually share code this way, but should we share code here?

😬 I'm not sure! My gut reaction is _no, don't share_ - I'm more worried about how consumers would understand the differences between the two than I am about the code side of it. Do you think it'd be difficult to advise folks on what to use and when (especially if there would come a time when folks would need to customize either of these?).

That is, unless you're talking about having some "base" level component that both of these would inherit from. Maybe then I could see some sharing.

> With the differences, would it be better for us to take this one on, or would it make sense to pair a little on it or?

Yeah, I think given our workload, we might need to rely on you all for the integration work. We're happy to support/fix any bugs with the CSS, but for this one, we might need y'all to carry it across the finish line.

Westbrook:
@hunterloftis you helped us bring this to life to begin with, and there are a bunch of changes included to adopted the new CSS dependency, so I was hoping you could take a closer look and make sure this aligns with anything that might remain in your head regarding the implementation of this.

hunterloftis:
Sorry, I missed this GH ping @Westbrook.

Re: the similarity w/ action button, I think the guidance for users boils down to: Is this an interactive component?
- Yes: Action Button
- No: Badge

@lunarfusion where can I learn more about the 'fixed' variant being deprecated? That variant is used in several places in production & doesn't appear deprecated in [the design spec.](https://spectrum.adobe.com/page/badge/#Fixed)

hunterloftis:
@Westbrook I must be missing something; the updated values l[ook like colors](https://github.com/adobe/spectrum-web-components/pull/2569/files#diff-0fd280552765430481a6f68aaf85bfdf3390a26d7b7455aa1f8cc21c1bd5bcf2R39-R50)? I'll take a closer look later.

packages/badge/src/Badge.ts
```diff
@@ -35,21 +36,70 @@ export const BADGE_VARIANTS = [
...
 export type BadgeVariant = typeof BADGE_VARIANTS[number];
+export const FIXED_VALUES_DEPRECATED = ['top', 'bottom', 'left', 'right'];
+export const FIXED_VALUES = [
+    'fuchsia',
+    'indigo',
+    'magenta',
+    'purple',
+    'seafoam',
+    'yellow',
+] as const;
```
Westbrook, line 47: 🤣 that's definitely a mistake, haha! Code review for the win. I'll get this corrected later today. 🙈

packages/badge/src/Badge.ts
```diff
@@ -35,21 +37,68 @@ export const BADGE_VARIANTS = [
...
+                window.__swc.warn(
+                    this,
+                    `The following values for "fixed" in <${this.localName}> have been deprecated. They will be removed in a future release.`,
+                    'https://opensource.adobe.com/spectrum-web-components/components/badge/#fixed',
+                    {
+                        issues: [...FIXED_VALUES_DEPRECATED],
+                    }
+                );
+            }
+        }
```
hunterloftis, line 83: 👍 on warning, I need to start looking for places to provide dev mode hints myself.

Once we start sprinkling in more than one or two, it might be nice to follow a pattern like:

```ts
this.warnIfDeprecatedFixed(fixed);

// ...

protected warnIfDeprecatedFixed(fixed: FixedValues | undefined) {
    if (!window.__swc.DEBUG) return;
    if (!(fixed && FIXED_VALUES_DEPRECATED.includes(fixed))) return;

    window.__swc.warn(this, '...', { issues: [...FIXED_VALUES_DEPRECATED] });
}
```

packages/badge/README.md
```diff
@@ -102,13 +102,81 @@ The `<sp-badge>` can be customized with either semantic or non-semantic variants
...
+        fixed="block-end"
+        style="position: absolute; bottom: 0; left: 200px;"
+    >
+        Bottom
+    </sp-badge>
+    <sp-badge
+        fixed="inline-start"
+        style="position: absolute; left: 0; top: 100px;"
+    >
+        Left
```
hunterloftis, line 179: Wondering if this would make it less-likely for people to use deprecated values:

```suggestion
        block-start
    </sp-badge>
    <sp-badge
        fixed="inline-end"
        style="position: absolute; right: 0; top: 100px;"
    >
        inline-end
    </sp-badge>
    <sp-badge
        fixed="block-end"
        style="position: absolute; bottom: 0; left: 200px;"
    >
        block-end
    </sp-badge>
    <sp-badge
        fixed="inline-start"
        style="position: absolute; left: 0; top: 100px;"
    >
        inline-start
```

hunterloftis (approved):
Thanks for handling this integration!

Now that "magenta" and "purple" don't work, should we update [the inline story](https://lunarfusion-badge-core-tokens--spectrum-web-components.netlify.app/storybook/?path=/story/badge--inline) to use semantic colors instead?

packages/badge/src/Badge.ts
Westbrook: This feels like this nicer way to do this, _BUT_ if you write it like this you'd have to write custom tooling to get this to be pulled out of the "for production" files. Wrapping all warn code in `if (window.__swc.DEBUG) {` means that when we convert `window.__swc.DEBUG` to `false` in esbuild the subsequent terser passes will eliminate the dead code for us.

packages/badge/README.md
Westbrook: Updated here and in the README.md. Great catch!

packages/badge/src/Badge.ts
hunterloftis: Aha! Good to know. I may try something like `if (window.__swc.DEBUG) this.warnIfDeprecatedFixed(fixed);` then.

hunterloftis (approved):

packages/badge/src/Badge.ts
Westbrook: Interesting, whether it pulls out the unused `private` would be a good test. Look forward to hearing what you find!

Issue #2510, opened 2022-08-12
[Bug]: action-menu can't act as a controlled component in non-submenus


hunterloftis:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

action-menu

### Expected behavior

With `selection=undefined`, `manageSelection` never runs at any level of the menu

### Actual behavior

`manageSelection` always runs on the root level, due to inherited behaviors. Reference:

https://github.com/adobe/spectrum-web-components/pull/2481#discussion_r938129114

### How can we reproduce this issue?

3. Scroll to '....'
4. Check console
5. See error

PR #2587, opened 2022-09-14
Hloftis/fix 2510 v2

hunterloftis:
## Description

`ActionMenu` inherits from `PickerBase`, which currently manages selections in two places even when `selects` is undefined:

- `restoreChildren` (via `openMenu`)
- `manageSelection`

This change guards against managing child `selected` states when `selects` is undefined.

I also found that the action-menu tests were frequently timing out during development, so I increased the timeout limit from 30s to 60s.

## Related issue(s)

Resolves https://github.com/adobe/spectrum-web-components/issues/2510

## Motivation and context

Consumers should be able to control top-level menu-item selection states.

## How has this been tested?

1. Added an automated regression test in the action-menu suite
2. Added an [interactive story](https://hloftis-fix-2510-v2--spectrum-web-components.netlify.app/storybook/?path=/story/action-menu--controlled) where you can test that:
3. "Snap" and "Guides" are selected by default
4. All selectable items can be toggled on/off by the user
5. The selection state is accurately reflected into the logged application state object
6. The non-selectable action triggers an alert, but never is selected

image

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/action-menu/stories/action-menu.stories.ts
+++ b/packages/action-menu/stories/action-menu.stories.ts
@@ -16,6 +16,7 @@ import '@spectrum-web-components/menu/sp-menu-item.js';
 import { ActionMenuMarkup } from './';
 
 import '@spectrum-web-components/icons-workflow/icons/sp-icon-settings.js';
+import { MenuItem } from '@spectrum-web-components/menu/src/MenuItem.js';
 
 export default {
     component: 'sp-action-menu',
@@ -114,5 +115,54 @@ export const submenu = (): TemplateResult => {
                 </sp-menu>
             </sp-menu-item>
         </sp-action-menu>
+    `;
+};
+
+export const controlled = (): TemplateResult => {
+    const state = {
+        snap: true,
+        grid: false,
+        guides: true,
+    };
+    function toggle(prop: keyof typeof state) {
+        return (event: Event): void => {
+            const item = event.target as MenuItem;
+            state[prop] = !state[prop];
+            // in Lit-based usage, this would be handled via render():
+            // <sp-menu-item ?selected=${this.isSomethingSelected}>
+            item.selected = state[prop];
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+            document.getElementById(
+                'state-json'
+            )!.textContent = `application state: ${JSON.stringify(state)}`;
+        };
+    }
+    return html`
+        <sp-action-menu label="View">
+            <sp-menu-item @click=${() => alert('action')}>
+                Non-selectable action
+            </sp-menu-item>
+            <sp-menu-item ?selected=${state.snap} @click=${toggle('snap')}>
+                Snap
+            </sp-menu-item>
+            <sp-menu-item>
+                Show
+                <sp-menu slot="submenu">
+                    <sp-menu-item
+                        ?selected=${state.grid}
+                        @click=${toggle('grid')}
+                    >
+                        Grid
+                    </sp-menu-item>
+                    <sp-menu-item
+                        ?selected=${state.guides}
+                        @click=${toggle('guides')}
+                    >
+                        Guides
+                    </sp-menu-item>
+                </sp-menu>
+            </sp-menu-item>
+        </sp-action-menu>
+        <span id="state-json"></span>
     `;
 };
--- a/packages/action-menu/stories/action-menu.stories.ts
+++ b/packages/action-menu/stories/action-menu.stories.ts
@@ -16,7 +16,7 @@ import '@spectrum-web-components/menu/sp-menu-item.js';
 import { ActionMenuMarkup } from './';
 
 import '@spectrum-web-components/icons-workflow/icons/sp-icon-settings.js';
-import { MenuItem } from '@spectrum-web-components/menu/src/MenuItem.js';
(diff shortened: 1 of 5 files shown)


packages/action-menu/stories/action-menu.stories.ts
```diff
@@ -16,6 +16,7 @@ import '@spectrum-web-components/menu/sp-menu-item.js';
 import { ActionMenuMarkup } from './';
 
 import '@spectrum-web-components/icons-workflow/icons/sp-icon-settings.js';
+import { MenuItem } from '@spectrum-web-components/menu/src/MenuItem.js';
```
Westbrook, line 19: Can we import this as a type here?

packages/action-menu/stories/action-menu.stories.ts
```diff
@@ -116,3 +117,52 @@ export const submenu = (): TemplateResult => {
...
+            // <sp-menu-item ?selected=${this.isSomethingSelected}>
+            item.selected = state[prop];
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+            document.getElementById(
+                'state-json'
+            )!.textContent = `application state: ${JSON.stringify(state)}`;
+        };
+    }
+    return html`
+        <sp-action-menu label="View">
```
Westbrook, line 141: Is it possible to toggle from the `change` event instead of from each Menu Item?

packages/action-menu/test/action-menu.test.ts
```diff
@@ -242,10 +242,10 @@ describe('Action menu', () => {
             'sp-menu[slot="submenu"]'
         ) as Menu;
         const selectedItem = submenu.querySelector(
-            '#selected-item'
+            '.selected-item'
```
Westbrook, line 245: This points to two items now...should be be more explicit to support reading this in the future?

packages/picker/src/Picker.ts
```diff
@@ -319,7 +319,7 @@ export class PickerBase extends SizedMixin(Focusable) {
                     selected?: boolean;
                 }
             ) => {
-                if (this.value === el.value) {
+                if (this.selects != null && this.value === el.value) {
```
Westbrook, line 322: Does the amount to the same as testing for truthiness? A `!=` always feels surprising, so it might be good just to add a quick comment for the future. Applies to the below usage as well.

hunterloftis:
This appears to fix the first state change, but subsequent state changes still get lost. I'll have to debug more...

web-test-runner.config.js
```diff
@@ -84,7 +84,7 @@ export default {
     protocol: 'https:',
     concurrency: 4,
     concurrentBrowsers: 1,
-    testsFinishTimeout: 30000,
+    testsFinishTimeout: 60000,
```
Westbrook, line 87: Does seem to effect actual runtime, but we should try to look into this when next we have a lot of extra time available... 😏

Westbrook (commented):
All except for being able to use the `change` event vs the `click` event in the story are nits. Looking good!

packages/action-menu/stories/action-menu.stories.ts
hunterloftis: It's possible, but I think for this story we should keep it as per-item events. The point of the story is to show that a consumer can lift all control out of the component, both for listening to inputs and for setting selected state.

Not a strong opinion though, so I'm happy to migrate to `change` if you still think we should.

packages/picker/src/Picker.ts
hunterloftis: `== null` tests for `null | undefined` (https://stackabuse.com/javascript-check-if-variable-is-a-undefined-or-null/)

It's a pattern I picked up on my last team where it also stood out to me because of the two-char check.

I encapsulated this check everywhere selected is set, & added a comment there.

packages/action-menu/stories/action-menu.stories.ts
Westbrook: Maybe it needs both then. I think of it as a `<select>` which prevent you from setting `click` events on its `<option>` elements, but I know a lot of teams already consume this element a la this demo, which I think is a less optimal consumption. Ensuring success for everyone in doing things _their_ way is that name of the game though, I guess.

hunterloftis: Cool, I've added state tracking to the last-changed item as well.

Westbrook (approved):

packages/action-menu/stories/action-menu.stories.ts
```diff
@@ -116,3 +117,65 @@ export const submenu = (): TemplateResult => {
...
+        return (event: Event): void => {
+            const item = event.target as MenuItem;
+            state[prop] = !state[prop];
+            // in Lit-based usage, this would be handled via render():
+            // <sp-menu-item ?selected=${this.isSomethingSelected}>
+            item.selected = state[prop];
+        };
+    }
+    function onChange(event: Event): void {
+        state.latestChange = (event.target as MenuItem).value;
```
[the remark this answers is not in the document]
Westbrook: This is just a demo, so we can refactor this later.

I'd expect `change` to be able manage this similar to https://github.com/adobe/spectrum-web-components/blob/main/packages/menu/stories/submenu.stories.ts#L145-L210

PR #2603, opened 2022-09-20
fix: convert the langage resolution workflow to a Reactive Controller

Westbrook:
## Description
- centralize language resolution tooling for reuse
- ensure the use of `unsubscribe` rather than event based contract nullification so that memory can be correctly released

## How has this been tested?

-   [ ] _Test case 1_
    1. The tests pass
-   [ ] _Test case 2_
    1. Go [here](https://resolved-language--spectrum-web-components.netlify.app/storybook/index.html?path=/story/slider--default)
    2. Duplicate the Slider in the story a bunch of times within the `<sp-story-decorator>` element
    3. Find the `<sp-theme>` element within the Decorator
    4. Check the value of `_contextConsumers.size` on the Theme element
    5. Delete some of the `<sp-slider>` elements. 
    6. See that `_contextConsumers.size` has decreased by the number of Slider you have deleted

## Types of changes

## Checklist

### DIFF
--- a/packages/number-field/src/NumberField.ts
+++ b/packages/number-field/src/NumberField.ts
@@ -20,7 +20,7 @@ import {
     property,
     query,
 } from '@spectrum-web-components/base/src/decorators.js';
-import { ProvideLang } from '@spectrum-web-components/theme';
+import { LanguageResolutionController } from '@spectrum-web-components/reactive-controllers/src/LanguageResolution.js';
 import { streamingListener } from '@spectrum-web-components/base/src/streaming-listener.js';
 import { NumberFormatter, NumberParser } from '@internationalized/number';
 
@@ -103,10 +103,6 @@ export class NumberField extends TextfieldBase {
     @property({ type: Number })
     public min?: number;
 
-    @property({ attribute: false })
-    private resolvedLanguage =
-        document.documentElement.lang || navigator.language;
-
     /**
      * The distance by which to alter the value of the element when taking a "step".
      *
@@ -183,6 +179,7 @@ export class NumberField extends TextfieldBase {
     private findChange!: (event: PointerEvent) => void;
     private change!: (event: PointerEvent) => void;
     private safty!: number;
+    private languageResolver = new LanguageResolutionController(this);
 
     private handlePointerdown(event: PointerEvent): void {
         if (event.button !== 0) {
@@ -440,12 +437,12 @@ export class NumberField extends TextfieldBase {
                 (formatOptionsNoUnit as Intl.NumberFormatOptions).style = style;
             }
             this._numberFormatterFocused = new NumberFormatter(
-                this.resolvedLanguage,
+                this.languageResolver.language,
                 formatOptionsNoUnit
             );
             try {
                 this._numberFormatter = new NumberFormatter(
-                    this.resolvedLanguage,
+                    this.languageResolver.language,
                     this.formatOptions
                 );
                 this._forcedUnit = '';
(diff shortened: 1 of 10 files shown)


packages/number-field/src/NumberField.ts
```diff
@@ -184,6 +184,13 @@ export class NumberField extends TextfieldBase {
     private change!: (event: PointerEvent) => void;
     private safty!: number;
 
+    constructor() {
+        super();
+        new LanguageResolutionControler(this, (language: string) => {
+            this.resolvedLanguage = language;
+        });
```
hunterloftis, line 191: Not a criticism of this PR, but creating "anonymous" controllers that reference the host, that the host has no reference to, feels weird to me. I'm not sure how I feel about that part of the ReactiveController API.

packages/theme/src/Theme.ts
```diff
@@ -508,24 +508,34 @@ export class Theme extends HTMLElement implements ThemeKindProvider {
...
+        this._contextConsumers.forEach(([callback, unsubscribe]) =>
+            callback(this.lang, unsubscribe)
+        );
     }
 
     private _handleContextPresence(event: CustomEvent<ProvideLang>): void {
         const target = event.composedPath()[0] as HTMLElement;
         if (this._contextConsumers.has(target)) {
-            this._contextConsumers.delete(target);
+            return;
```
hunterloftis, line 525: Because of this early return, you could drop the subsequent `else`

tools/reactive-controllers/src/LanguageResolution.ts
```diff
@@ -0,0 +1,54 @@
...
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import type { ReactiveController, ReactiveElement } from 'lit';
+import { ProvideLang } from '@spectrum-web-components/theme';
+
+export class LanguageResolutionControler implements ReactiveController {
+    host: ReactiveElement;
+    language!: string;
```
hunterloftis, line 18: Since this is public, it may be safer to type it as nullable, since it starts undefined. That way, a consumer querying `this.languageController.language` will know to handle the null-or-undefined case.

```suggestion
    language?: string;
```

hunterloftis (commented):
Is it expected that .size is 0 before duplicating the slider? (I would have expected 1)

image

packages/number-field/src/NumberField.ts
Westbrook: I think the question is, do we think a callback API or a property API is better to consume here. I'm looking at taking `language` off of the API because it's not currently "reactive", but maybe we should lean more into that workflow for consuming _this_ Reactive Controller?

hunterloftis: That's a good point - I don't think I have a strong opinion about which direction the API guides users in, but ideally it would be one obvious way.

I have a _slight_ preference towards making `this.langController.language` reactive and getting rid of the callback, because:

- that seems to align with the emerging field of reactive controllers generally
- it's harder for a user to fail / puts less on the user (they don't need to understand what to put into the callback & why)

But if you'd prefer dropping `language` I'd be fine with that too, at least then the callback requirement would be clear from the shape of the API.

Westbrook: Very helpful! I've updated to a reactive `this.languageResolver.language` value for this API. Stones and birds. 😉

Westbrook:
> Is it expected that .size is 0 before duplicating the slider? (I would have expected 1)

This is a timing situation related to Storybook, which is not new in the context of this issue. There might be benefit to looking deeper into it, e.g. https://github.com/lit/lit/tree/main/packages/labs/context#late-upgraded-context-providers but I am not surprised by it today.

tools/reactive-controllers/src/LanguageResolution.ts
```diff
@@ -0,0 +1,52 @@
...
+        this.host = host;
+        this.host.addController(this);
+    }
+
+    public hostConnected(): void {
+        this.resolveLanguage();
+    }
+
+    public hostDisconnected(): void {
+        this.unsubscribe?.();
```
hunterloftis, line 31: 👀 TIL

hunterloftis (dismissed):
Given how important these foundational utilities are, at some point we should add some tests for them!

Nit: in [the spec](https://github.com/webcomponents-cg/community-protocols/blob/main/proposals/context.md) `unsubscribe` is called `dispose`.

PR #2614, opened 2022-09-26
fix: prevent mobile interactions from triggering the virtual keyboard

### LINKED ISSUE #2588: [Bug]: Slider, with editable, using a mobile device, shows virtual keyboard for a short time
### Code of conduct

### Impacted component(s)

Slider

### Expected behavior

Slider with `editable` attribute, on mobile devices, should not trigger virtual keyboard when dragging to change.

### Actual behavior

Slider with `editable` attribute, on mobile devices, is opening virtual keyboard, although input is not focused

### Screenshots

_No response_

Chrome

Open https://studio.webcomponents.dev/preview/a5T5HC4etLGmfUD750PF on a mobile device, using Chrome and try to use the first slider

`<sp-slider editable/>`

_No response_

Westbrook:
## Description
Only focus the Number Field after interacting with the Slider when the interaction is from a `mouse` so as to prevent the focus action from triggering the Virtual Keyboard.

To support setting up the tests here, I merged the a/sync tests for this into a helper method so that they could share the same code.

## Related issue(s)
- fixes #2588 

## How has this been tested?
@alincc can you help us confirm this perform in a way that you would expect

-   [x] _Test case 1_
    1. Go [here](https://slider-editable-mobile--spectrum-web-components.netlify.app/components/slider/#editable) in a mobile device
    2. Interact with one of the "editable" Sliders
    3. See that the Virtual Keyboard is not triggered

## Types of changes

## Checklist


alincc:
The virtual keyboard no longer appears when using the slider with simple input (hide-stepper). However I found there is an issue when slider is editable with numeric stepper enabled. When I press up or down the keyboard appears for a short amount of time.
Tested on Android (chrome) and iPhone (safari)

Westbrook:
@alincc when you put it that way, I worry that this intervention isn't the right path forward. I don't want to prevent a mobile context from using an editable Slider, but it might not be in the best interests of your visitors for a mobile application to do so. I'm going to cycle back with design and accessibility on this as see if there's a more clear expectation here, and you might want to do the same with your team to ensure you're targeting the highest quality experience for you users by shipping Sliders that are editable in the mobile context.

alincc:
@Westbrook I suggest to move forward with this change since it fixes the Slider in the way the design required (editable, quiet with hide-stepper) then I can create a new bug related to the numeric stepper issue I identified.

alincc (approved):

alincc (approved):

PR #2663, opened 2022-10-17
fix: correct the relationship between overlayWillCloseCallback and phased animations

### LINKED ISSUE #2646: [Bug]: dismissing an overlay twice, quickly, breaks the overlay in question
### Code of conduct

### Impacted component(s)

overlay,dialog

### Expected behavior

Pressing escape N times, or clicking on an underlay N times, no matter how fast, does not impact the future behavior of overlays.

### Actual behavior

Pressing escape twice or clicking on an underlay twice, quickly, can prevent future overlays from opening.

### Screenshots

https://user-images.githubusercontent.com/364501/194432655-23f81f5f-b8a4-457a-a716-554c877a4849.mov

_No response_

1. Go to https://opensource.adobe.com/spectrum-web-components/components/dialog-wrapper/
2. Click "toggle dialog"
3. Double-click on its underlay quickly
4. Click "toggle dialog"
5. Notice that no dialog opens

_No response_

_No response_

Westbrook:
## Description
- prevent N number of `Escape` pressed or "close" button clicks from breaking the overlay that is being managed by those interactions
- ensure "open" animations are being run in Firefox
  - turn tests in this are back on for Firefox
- refactor work workflows for simplicity
- remove test time workaround for multiple overlay open/closes

## Related issue(s)
- fixes #2646

## How has this been tested?
-   [ ] _Test case 1_
    1. Go [here](https://double-close--spectrum-web-components.netlify.app/storybook/?path=/story/dialog-base--slotted)
    2. Once the dialog is open, quickly press `Escape` N number of times.
    3. See that the dialog is closed
    4. Click the "Toggle Dialog" button
    5. See that the dialog is opened.
    6. Repeat 2-5 N number of times.
-   [ ] _Test case 2_
    1. Go [here](https://double-close--spectrum-web-components.netlify.app/storybook/?path=/story/dialog-base--slotted)
    2. Once the dialog is open, quickly click "Cancel" N number of times
    3. See that the dialog is closed
    4. Click the "Toggle Dialog" button
    5. See that the dialog is opened.
    6. Repeat 2-5 N number of times.
-   [ ] _Test case 3_
    1. Go [here](https://double-close--spectrum-web-components.netlify.app/storybook/?path=/story/dialog-base--slotted) in FireFox
    2. Open and close the dialog N number of times by which ever method you would like
    3. See that while _opening_ there is an animated transition between the closed and opened state

## Types of changes

## Checklist

### DIFF
--- a/packages/dialog/src/DialogBase.ts
+++ b/packages/dialog/src/DialogBase.ts
@@ -80,18 +80,14 @@ export class DialogBase extends FocusVisiblePolyfillMixin(SpectrumElement) {
         return dialog || this;
     }
 
-    public override focus(): void {
+    public override async focus(): Promise<void> {
         if (this.shadowRoot) {
             const firstFocusable = firstFocusableIn(this.dialog);
             if (firstFocusable) {
-                if (firstFocusable.updateComplete) {
-                    firstFocusable.updateComplete.then(() =>
-                        firstFocusable.focus()
-                    );
-                    /* c8 ignore next 3 */
-                } else {
-                    firstFocusable.focus();
+                if ((firstFocusable as SpectrumElement).updateComplete) {
+                    await firstFocusable.updateComplete;
                 }
+                firstFocusable.focus();
                 this.removeAttribute('tabindex');
             } else {
                 this.dialog.focus();
@@ -102,8 +98,10 @@ export class DialogBase extends FocusVisiblePolyfillMixin(SpectrumElement) {
         }
     }
 
+    private animating = false;
+
     public overlayWillCloseCallback(): boolean {
-        if (!this.open) return false;
+        if (!this.open) return this.animating;
         this.close();
         return true;
     }
@@ -134,8 +132,8 @@ export class DialogBase extends FocusVisiblePolyfillMixin(SpectrumElement) {
 
     protected handleUnderlayTransitionend(event: TransitionEvent): void {
         if (!this.open && event.propertyName === 'visibility') {
-            this.dispatchClosed();
             this.resolveTransitionPromise();
+            this.dispatchClosed();
         }
     }
 
@@ -150,9 +148,13 @@ export class DialogBase extends FocusVisiblePolyfillMixin(SpectrumElement) {
 
     protected override update(changes: PropertyValues<this>): void {
         if (changes.has('open') && changes.get('open') !== undefined) {
-            this.transitionPromise = new Promise(
-                (res) => (this.resolveTransitionPromise = res)
-            );
+            this.animating = true;
+            this.transitionPromise = new Promise((res) => {
+                this.resolveTransitionPromise = () => {
+                    this.animating = false;
+                    res();
+                };
+            });
         }
         super.update(changes);
     }
--- a/packages/dialog/test/dialog-base.test.ts
+++ b/packages/dialog/test/dialog-base.test.ts
@@ -29,7 +29,6 @@ import '@spectrum-web-components/overlay/overlay-trigger.js';
 import { alertDestructive } from '../stories/dialog.stories.js';
 import { Button } from '@spectrum-web-components/button/src/Button.js';
 import { DialogBase } from '@spectrum-web-components/dialog';
-import { sendMouse } from '../../../test/plugins/browser.js';
 
 async function styledFixture<T extends Element>(
     story: TemplateResult
@@ -50,18 +49,6 @@ const overlayTrigger = (story: () => TemplateResult): TemplateResult => html`
 `;
 
 describe('dialog base', () => {
-    beforeEach(async () => {
-        // Something about this prevents Chromium from swallowing the CSS transitions
-        // to "open" so that test timing can be properly acquired below.
-        await sendMouse({
-            steps: [
-                {
-                    type: 'move',
-                    position: [0, 0],
-                },
-            ],
-        });
-    });
     it('does not close by default with interacting with buttons', async () => {
         const el = await styledFixture<OverlayTrigger>(
             overlayTrigger(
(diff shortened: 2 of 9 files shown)


packages/dialog/src/DialogBase.ts
```diff
@@ -150,9 +148,13 @@ export class DialogBase extends FocusVisiblePolyfillMixin(SpectrumElement) {
 
     protected override update(changes: PropertyValues<this>): void {
         if (changes.has('open') && changes.get('open') !== undefined) {
-            this.transitionPromise = new Promise(
-                (res) => (this.resolveTransitionPromise = res)
-            );
+            this.animating = true;
```
hunterloftis, line 151: This sort of async bookkeeping & synchronization is what suggests that perhaps we should consider a [state machine](https://www.youtube.com/watch?v=RqTxtOXcv8Y) pattern. (/cc @najikahalsema)

packages/dialog/test/dialog-base.test.ts
```diff
@@ -115,10 +102,6 @@ describe('dialog base', () => {
         expect(dialog.parentElement?.localName).to.equal('overlay-trigger');
     });
     it('does not close by default with interacting with buttons when recycled', async () => {
-        // There is an `sp-dialog-base` recyling issue in Firefox
```
hunterloftis, line 118: 🎉 also love to see test skips deleted!

packages/dialog/test/dialog-wrapper.test.ts
```diff
@@ -105,7 +105,7 @@ describe('Dialog Wrapper', () => {
         expect(el.open).to.be.false;
         expect(closeSpy.callCount).to.equal(1);
     });
-    xit('opens and closes when element is recycled', async () => {
+    it('opens and closes when element is recycled', async () => {
         // Skip while there is an `sp-dialog-base` recyling issue in Chromium and Firefox,
         // this test passes outside of the testing context and in Safari...
```
hunterloftis, line 110:
```suggestion
```

packages/overlay/src/overlay-stack.ts
```diff
@@ -48,6 +48,10 @@ interface ManagedOverlayContent {
     updateComplete?: Promise<boolean>;
 }
 
+function nextFrame(): Promise<void> {
```
hunterloftis, line 51: Any reason not to import `nextFrame` from '@open-wc/testing'?

packages/overlay/test/overlay-trigger-hover.test.ts
```diff
@@ -53,32 +52,20 @@ async function styledFixture<T extends Element>(
...
+    //     const promises = triggers.map((trigger) => {
+    //         if (trigger.open) {
+    //             const closed = oneEvent(trigger, 'sp-closed');
+    //             trigger.open = undefined;
+    //             return closed;
+    //         }
+    //         return Promise.resolve();
+    //     });
+    //     await Promise.all(promises);
+    // });
```
hunterloftis, line 68:
```suggestion
```

packages/shared/src/focusable.ts
```diff
@@ -16,6 +16,10 @@ import { FocusVisiblePolyfillMixin } from './focus-visible.js';
 
 type DisableableElement = HTMLElement & { disabled?: boolean };
 
+function nextFrame(): Promise<void> {
+    return new Promise((res) => requestAnimationFrame(() => res()));
+}
```
hunterloftis, line 21: Same Q here

packages/shared/src/focusable.ts
```diff
@@ -259,12 +263,31 @@ export class Focusable extends FocusVisiblePolyfillMixin(SpectrumElement) {
...
+            // If at connect time the [autofocus] content is placed within
+            // content that needs to be "hidden" by default, it would need to wait
+            // two rAFs for animations to be triggered on that content in
+            // order for the [autofocus] to become "visisble" and have its
+            // focus() capabilities enabled.
+            //
+            // Await this with `getUpdateComplete` so that the element cannot
+            // become "ready" until `manageFocus` has occured.
+            await nextFrame();
+            await nextFrame();
```
hunterloftis, line 279: Good encapsulation of this requirement & documentation, but painful that we're manually counting frames.

hunterloftis (approved):

packages/shared/src/focusable.ts
Westbrook: This definitely should become a platform API, but we're not specifically "counting" frames as much as ensuring that a full paint cycle has gone by before this happens. Cycle is one `rAF` to get to the end of _this_ fame (which is pre-paint) and one `rAF` to get to the end of the _next_ frame (which is post paint).

Maybe the abstraction shouldn't be "next frame" and should be "paintComplete" or something?

packages/overlay/src/overlay-stack.ts
Westbrook: I didn't want to set a standard for importing "test" code into the production library...but maybe that's being overly strict?

packages/dialog/src/DialogBase.ts
Westbrook: An this state management might not "be cleaned up" but a new overlay approach, as it's bound deeply into the content that is being overlaid, which means we might still benefit from it...

packages/overlay/test/overlay-trigger-hover.test.ts
Westbrook: Whoops, I got too excited when that actually fixed the issue I was running into.

packages/overlay/src/overlay-stack.ts
hunterloftis: 👍 no worries, they're identical today but we can't expect a testing helper to always be production-ready. And it's 1 line.

packages/shared/src/focusable.ts
hunterloftis: Understood, I'm just wary of linking rafs to any kind of logic. Perhaps there aren't extreme cases here where a particular event might be skipped due to raf not firing when we expect (or not including a particular change that we expect), but I've added bugs before by relying on ${something} happening before an raf.

packages/dialog/src/DialogBase.ts
nickschaap: For purposes of extensibility it might be nice to consider moving away from this lifecycle approach where the consumer is having to manage transitions or consolidating the logic into an OverlayBase element so that if a client wants to plugin to the Overlay API with their own custom component they can do so rather easily without have to rewrite a lot of this animation state management logic. #2086

PR #2694, opened 2022-10-27
fix: correct the origin on "maximumFractionDigits" when deciding "inputMode"

### LINKED ISSUE #2674: [Bug]: missing the decimal point on iOS keyboard if enter the editable <sp-slider>
### Code of conduct

### Impacted component(s)

<sp-slider>

### Expected behavior

the iOS keyboard should have "." (the decimal point)

### Actual behavior

no decimal point on iOS keyboard

### Screenshots

image

_No response_

missing the decimal point on iOS keyboard if enter the editable <sp-slider>.

I tried with my iPhone to open the spectrum component website https://opensource.adobe.com/spectrum-web-components/components/slider/ , and found the keyboard doesn’t show the point there even the numeric input box is for floating number. see screenshot.
Steps:
1. with iPhone, 
2. open spectrum component website https://opensource.adobe.com/spectrum-web-components/components/slider/
2. tap on the numeric input of the slider
4. Check the keyboard
5. Missing decimal point

_No response_

_No response_

Westbrook:
## Description
Allow `.` button in iPhone virtual keyboard.

### Question
This leveraged the `resolvedOptions` of the Number Formatter, as seen in https://github.com/adobe/react-spectrum/blob/7cebb4eea31f383f1eec06cc53a7909f8736f66a/packages/%40react-aria/numberfield/src/useNumberField.ts#L148-L149, which seems to default to `3` if not set. Does this precipitate adding information around this to the documentation so that consumers know how to actively manage this keyboard input? cc: @wyanghz1

## Related issue(s)
- fixes #2674

## How has this been tested?
-   [ ] _Test case 1_
    1. On an iPhone
    2. Go [here](https://number-field-decimal--spectrum-web-components.netlify.app/components/slider/#editable)
    3. Attempt to edit the text of the slider input
    4. See that the `.` key is present

Slider Spectrum Web Components

## Types of changes

## Checklist

### DIFF
--- a/packages/number-field/src/NumberField.ts
+++ b/packages/number-field/src/NumberField.ts
@@ -161,6 +161,9 @@ export class NumberField extends TextfieldBase {
     }
 
     private convertValueToNumber(value: string): number {
+        if (isIPhone() && this.inputElement.inputMode === 'decimal') {
+            value = value.replace(',', '.');
+        }
         return this.numberParser.parse(value);
     }
 
@@ -590,9 +593,9 @@ export class NumberField extends TextfieldBase {
         if (changes.has('min') || changes.has('formatOptions')) {
             let inputMode = 'numeric';
             const hasNegative = typeof this.min !== 'undefined' && this.min < 0;
-            const { maximumFractionDigits } = this.formatOptions;
-            const hasDecimals =
-                maximumFractionDigits && maximumFractionDigits > 0;
+            const { maximumFractionDigits } =
+                this.numberFormatter.resolvedOptions();
+            const hasDecimals = maximumFractionDigits > 0;
             /* c8 ignore next 18 */
             if (isIPhone()) {
                 // iPhone doesn't have a minus sign in either numeric or decimal.


najikahalsema:
So, I just messed around with this on my iPhone and get the following behaviour: 

https://user-images.githubusercontent.com/44980010/199116610-46a8317c-dbfc-49fe-8344-390ae4cdcf3d.mov

My iphone is set to German units as I changed it when I moved there and never switched it back. ^^; 

I thought we already handled i18n behaviour for the number field, but with the number resetting to the highest input when I try to use a decimal... I'm wondering if we should double check on this. Is using the comma something we didn't account for during the i18n phase? Or does the docs site specifically not expect this type of input? 

The other thing is that I would expect the keyboard to go away when I click "Done", but since Apple is constantly changing how they want to display this type of interaction, I don't know how much that matters.... I haven't upgraded to iOS16 fwiw.

Westbrook:
@najikahalsema that's an interestingly complex set of realities. While i18n support is present, it requires that the site language be set, as per https://opensource.adobe.com/spectrum-web-components/components/slider/#labels-and-formatting. At current, the site language _is not_ set to German, so I wouldn't expect it to be ready to support German input in that context.

Westbrook:
@najikahalsema, took a deeper look at this, and there is a setting in the OS that is not clearly surfaced to the browser, whether to use `1,000,000.00` or `1.000.000,00` in the keyboard. I've added some extra checking to _possibly_ handle this, and it's looking good on my end, but I had to mess with my device settings a whole bunch and I'd prefer to see it working in a slightly more "normative" context. Can you confirm on your end when this build goes up?

najikahalsema (approved):
Looks good!

PR #2716, opened 2022-11-08
fix(dialog): don't show DialogWrapper divider when there's no headline

### LINKED ISSUE #2699: [Bug]: Dialog Wrapper without headline value displays strangely
### Code of conduct

### Impacted component(s)

Dialog

### Expected behavior

You should be able to display a Dialog Wrapper without a header attribute/property.

### Actual behavior

There's an empty headline with a divider...

### Screenshots

_No response_

_No response_

_No response_

_No response_

_No response_

beeduul:
DialogWrapper shouldn't show a divider if there's no headline

## Related issue(s)

fixes #2699

## How has this been tested?

3 new unit tests. visual confirmation of expected behavior using storybook.

See Storybook cases:
DialogWrapper:Wrapper With Headline
DialogWrapper:Wrapper With Headline No Divider
DialogWrapper:Wrapper No Headline

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/dialog/src/DialogWrapper.ts
+++ b/packages/dialog/src/DialogWrapper.ts
@@ -69,6 +69,9 @@ export class DialogWrapper extends DialogBase {
     @property()
     public headline = '';
 
+    @property({ type: String, attribute: 'headline-visibility' })
+    public headlineVisibility: 'none' | undefined;
+
     protected override get dialog(): Dialog {
         return this.shadowRoot.querySelector('sp-dialog') as Dialog;
     }
@@ -98,10 +101,14 @@ export class DialogWrapper extends DialogBase {
     }
 
     protected override renderDialog(): TemplateResult {
+        const hideDivider =
+            this.noDivider ||
+            !this.headline ||
+            this.headlineVisibility === 'none';
         return html`
             <sp-dialog
                 ?dismissable=${this.dismissable}
-                ?no-divider=${this.noDivider || !this.headline}
+                ?no-divider=${hideDivider}
                 ?error=${this.error}
                 mode=${ifDefined(this.mode)}
                 size=${ifDefined(this.size)}
@@ -122,7 +129,12 @@ export class DialogWrapper extends DialogBase {
                     : html``}
                 ${this.headline
                     ? html`
-                          <h2 slot="heading">${this.headline}</h2>
+                          <h2
+                              slot="heading"
+                              ?hidden=${this.headlineVisibility === 'none'}
+                          >
+                              ${this.headline}
+                          </h2>
                       `
                     : html``}
                 <slot></slot>
--- a/packages/dialog/src/DialogWrapper.ts
+++ b/packages/dialog/src/DialogWrapper.ts
@@ -105,6 +105,20 @@ export class DialogWrapper extends DialogBase {
             this.noDivider ||
             !this.headline ||
             this.headlineVisibility === 'none';
+
+        if (window.__swc.DEBUG) {
+            if (!this.headline) {
+                window.__swc.warn(
+                    this,
+                    `<${this.localName}> elements will not be accessible to screen readers without a "headline" attribute or property.`,
+                    'https://opensource.adobe.com/spectrum-web-components/components/dialog-wrapper/#accessibility',
+                    {
+                        type: 'accessibility',
+                    }
+                );
+            }
+        }
+
         return html`
             <sp-dialog
                 ?dismissable=${this.dismissable}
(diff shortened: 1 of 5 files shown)


packages/dialog/test/dialog-wrapper.test.ts
```diff
@@ -127,6 +131,43 @@ describe('Dialog Wrapper', () => {
...
+
+        const dialog = wrapper.shadowRoot.querySelector('sp-dialog') as Dialog;
+        const divider = dialog.shadowRoot.querySelector(
+            'sp-divider.divider'
+        ) as Divider;
+
+        expect(divider).to.be.null;
+    });
+    it("hides header divider when there's no header", async () => {
+        const wrapper = await styledFixture<DialogWrapper>(wrapperNoHeadline());
```
Westbrook, line 161: Can you add a test of this story with `await expect(el).to.be.accessible();` as well? Want to confirm there are no accessibility issues with a no-headline delivery that we'd need to handle.

packages/dialog/stories/dialog-wrapper.stories.ts
```diff
@@ -473,6 +473,51 @@ export const wrapperFullscreen = (
     `;
 };
 
+export const wrapperWithHeadline = (
```
Westbrook, line 476: Story additions are going to bump into this requirement: https://github.com/adobe/spectrum-web-components#keeping-ci-assets-updated

packages/dialog/src/DialogWrapper.ts
```diff
@@ -101,7 +101,7 @@ export class DialogWrapper extends DialogBase {
         return html`
             <sp-dialog
                 ?dismissable=${this.dismissable}
-                ?no-divider=${this.noDivider}
+                ?no-divider=${this.noDivider || !this.headline}
```
Westbrook, line 104: This is much nicer than I thought it was going to have to be when I looked at this quickly the other week!

Westbrook (changes requested):
Thanks for diving into this! Two quick changes and we should be on our way here. 🤞🏼

packages/dialog/test/dialog-wrapper.test.ts
beeduul: your intuition was correct!  adding the test results in the below error.  any guidance on how to fix?

```
 ❌ Dialog Wrapper > hides header divider when there's no header
      Error: Accessibility Violations
      ---
      Rule: aria-dialog-name
      Impact: serious
      ARIA dialog and alertdialog nodes should have an accessible name (https://dequeuniversity.com/rules/axe/4.4/aria-dialog-name?application=axeAPI)
      
      Issue target: sp-dialog-wrapper,sp-dialog
      Context: <sp-dialog no-divider="" dir="ltr" tabindex="0" role="dialog" aria-describedby="sp-dialog-description-55-17">
      Fix any of the following:
        aria-label attribute does not exist or is empty
        aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty
        Element has no title attribute
      ---
        aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty
        Element has no title attribute
        ---
        at processResults (node_modules/chai-a11y-axe/src/accessible.js:87:15)
        at node_modules/chai-a11y-axe/src/accessible.js:117:7
        at async o.<anonymous> (packages/dialog/test/dialog-wrapper.test.ts:166:8)
```

Westbrook: @jnurthen if we have a dialog without a visible headline, can we promote the body content to the `aria-labelledby` content to appropriately fulfil the accessibility contract here? Visually, imagine a prompt or alert.

Westbrook: @beeduul FYI the Spectrum spec lists the [title as "required"](https://spectrum.adobe.com/page/alert-dialog/#Title) but I'll reach out to the design team as that feels somewhat constrictive. In the meantime, hopefully we'll hear from the accessibility team on the feasibility of a Dialog without a title in that context.

Westbrook: @beeduul with this information the pattern we've namely been using would be to add a `label` attribute to the `<sp-dialog-wrapper>` and `<sp-dialog>` elements. The Wrapper would namely pass it down to the Dialog, and the Dialog would apply it to itself as an `aria-label` attribute. This would be a good place to install a Dev Mode warning to support consumers in accurately delivering this pattern to their visitors, similar to what can be found at https://github.com/adobe/spectrum-web-components/blob/main/packages/progress-circle/src/ProgressCircle.ts#L122-L142

PR #2732, opened 2022-11-14
Add element resolver for keeping a live reference to an element in the same DOM tree

Westbrook:
## Description
- add some test coverage ignore markers
- add the Element Resolution Controller
  - resolves an element in the same DOM tree from a selector
  - always resolve the first element of that selector regardless of DOM/attribute mutations
- applies the controller to Field Label which previously has a one shot element resolution method

## Motivation and context
Prepares this functionality to be leveraged in overlay trigger interactions

## How has this been tested?
-   [ ] _Test case 1_
    1. Go [here](https://element-resolver--spectrum-web-components.netlify.app/storybook/?path=/story/field-label--standard)
    2. Futz with the DOM and attributes of the story dom
    3. Change the `for` value on the Field Label
    4. Change the `id` value on the Textfield
    5. Move the Textfield around
    6. Add/remove a different Textfield with the same ID
    7. See that the element resolves as expected.\

## Types of changes
-   [x] refactor

## Checklist

### DIFF
--- a/packages/field-label/src/FieldLabel.ts
+++ b/packages/field-label/src/FieldLabel.ts
@@ -25,6 +25,14 @@ import {
 import type { Focusable } from '@spectrum-web-components/shared';
 import '@spectrum-web-components/icons-ui/icons/sp-icon-asterisk100.js';
 import asteriskIconStyles from '@spectrum-web-components/icon/src/spectrum-icon-asterisk.css.js';
+import {
+    conditionAttributeWithId,
+    conditionAttributeWithoutId,
+} from '@spectrum-web-components/base/src/condition-attribute-with-id.js';
+import {
+    ElementResolutionController,
+    elementResolverUpdatedSymbol,
+} from '@spectrum-web-components/reactive-controllers/src/ElementResolution.js';
 
 import styles from './field-label.css.js';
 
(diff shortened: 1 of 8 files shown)


tools/reactive-controllers/src/ElementResolution.ts
```diff
@@ -0,0 +1,122 @@
...
+                            ([...mutation.addedNodes] as HTMLElement[]).some(
+                                (el) => el?.matches?.(this.selector)
+                            );
+                        needsResolution =
+                            needsResolution ||
+                            currentElementRemoved ||
+                            matchingElementAdded;
+                        break;
+                    case 'attributes':
+                        const attributeChangedOnCurrentElement =
```
Westbrook, line 74: Sometime this talks about `selectors` sometimes it talks about `ids` which are not 100% the same, which is the right one, generic resolution or ID resolution.

packages/field-label/src/FieldLabel.ts
```diff
@@ -79,13 +87,24 @@ export class FieldLabel extends SizedMixin(SpectrumElement) {
...
-            return;
-        }
-        const parent = this.getRootNode() as HTMLElement;
-        const target = parent.querySelector(`#${this.for}`) as Focusable;
+    private resolvedElement = new ElementResolutionController(this);
+
+    private async manageTarget(): Promise<void> {
+        const target = this.resolvedElement.element as Focusable;
         if (!target) {
+            if (this.target) {
```
Westbrook, line 95: Should the previous target be cleaned regardless? Check for identity change instead?

tools/reactive-controllers/element-resolution.md
```diff
@@ -0,0 +1,82 @@
...
+
+The resolved reference will always be the first element matching the selector applied, so in the following example the element with content "First!" will be the reference:
+
+```html-no-demo
+<root-el></root-el>
+<div class="other-element">First!</div>
+<div class="other-element">Last.</div>
+```
+
+A [`MutationObserver`](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) it leveraged to track mutations to the DOM tree in which the host element resides in order to update the element reference on any changes to the content therin that could change the resolved element.
```
nickschaap, line 55: typo: A MutationObserver _is_

tools/reactive-controllers/element-resolution.md
```diff
@@ -0,0 +1,82 @@
+## Description
+
+An `ElementResolutionController` keeps an active reference to another element in the same DOM tree. Supply the controller with the a selector by which it can locate the element to reference and it will manage observing the DOM tree in question to ensure that the reference it holds is always the first element that matches that selector or `null`.
```
nickschaap, line 3:
```suggestion
An `ElementResolutionController` keeps an active reference to another element in the same DOM tree. Supply the controller with a selector that matches the element and it will manage observing the DOM tree to ensure that the reference it holds is always the first matched element or `null`.
```

tools/reactive-controllers/element-resolution.md
Westbrook:
```suggestion
A [`MutationObserver`](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) is leveraged to track mutations to the DOM tree in which the host element resides in order to update the element reference on any changes to the content therein that could change the resolved element.
```

PR #2734, opened 2022-11-14
refactor(tags): use core tokens

lunarfusion:
## Description

Migrates tag to spectrum tokens. This is a beta release. Hold for full release.

## Related issue(s)

-

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices


lunarfusion:
# Unexpected VRT Failure - Express border radius

The border radius on Express is expected to be fully rounded (equal to the height of the component) as seen here in my local dev of spectrum-css:

Screen Shot 2022-11-15 at 12 48 21 PM

SWC VRTS (example: https://0493193c21494deb0d937a3a4e3c5487--spectrum-web-components.netlify.app/review/) show no rounding on Express.

Screen Shot 2022-11-15 at 12 45 40 PM

Any ideas? @Westbrook

lunarfusion:
# VRT Failure Assessment

**We are expecting to see this:**
Screen Shot 2022-11-30 at 4 10 19 PM

**But instead we see this:**
Screen Shot 2022-11-30 at 4 10 26 PM

It appears that the width style on the clear button fill is causing this. In spectrum-css, I've overridden the clear button fill width to neutralize the extra spacing like so

```
.spectrum-ClearButton,
.spectrum-Tag-clearButton {
  .spectrum-ClearButton-fill {
    width: auto;
  }
}
```
But this is failing to target the fill in spectrum web components. Would removing the nesting shown above be the correct approach, or should I target the fill with a different selector?

@Westbrook ?

lunarfusion:
# VRT Failure Assessment - Revised

## Changes expected to impact VRT Failures

- Block size of Tag has decreased by 2 pixels.
- Spacing from clear button to edge has increased.

# VRT Failure Pattern

After a lot of corrections, the failures appear to be consistent and expected. 

Visual of pattern:

Screen Shot 2022-12-01 at 5 08 42 PM

Differences seen here appear to be due to the height increase, as seen here:
Screen Shot 2022-12-01 at 5 09 55 PM

And due to the increase in inline-end spacing of clear button, seen here:
Screen Shot 2022-12-01 at 5 12 12 PM

There is also a slight difference in text size of the label, which appears to be slightly larger. The text size has not been altered (was and is 14px on medium size tag, for example). This _may_ be due to an existing error in SWC.

Also noticeable is a slight change in color of the disabled icon. Disabled content color has not changed. Previously it was --spectrum-gray-400 = rgb(177, 177, 177); and it remains the same. The difference in color of the magnifying glass in disabled state _may_ be due to a previous error.

@Westbrook this is ready for your review.

Westbrook:
This is looking much better, want to make sure the spacing around the `X` in the following screen shot it correct:
image

Visually it feels like it's larger than it should be at the end of the Tag, but if it's "as designed" let's get a full version of this package out, today!

pfulton:
Merged in Spectrum CSS and graduated! The full version is: `@spectrum-css/tag@4.0.0`

PR #2746, opened 2022-11-18
fix: ensure that an overlay can be released even if it does not complete its fade in animation

Westbrook:
## Description
Previously opening and closing an overlay in short succession could trick the state of the Active Overlay container into "never being ready". This was because the `sp-overlay-fade-in` animation would not complete before the `sp-overlay-fade-out` animation was applied, leaving the Active Overlay in an incomplete state. From this state the overlay could not be fully returned to its place of origin, which would prevent it from being opened again in the future. The timing hole for this is 160ms long, and seemingly not guaranteed, so it is quite difficult to actually fall into, however not so difficult as to prevent visitors from falling into it. This could be encountered with the most regularity in a Menu with Flyout Submenus that would open/close in succession.

The work here captures the "fade in" promise and resolves it manually in the case that a "fade out" animation is applied before the initial animation is complete. A test is added to cover this situation. The test itself only failed in Chrome...but I feel that is a clear enough signal that we're moving in the right direction here.

## How has this been tested?

-   [ ] _Test case 1_
    1. Go [here](https://opensource.adobe.com/spectrum-web-components/storybook/index.html?path=/story/menu-submenu--submenu)
    2. Fiddle around with the submenus
    3. Open and close them quickly, over and over again
    4. At some point, assuming your system isn't too strong, or is possibly over taxed in other ways, you'll end up with a submenu that will no longer open.
    5. This is a great sign that you can test our fix.
-   [ ] _Test case 2_
    1. Go [here](https://test-overlay-timing--spectrum-web-components.netlify.app/storybook/?path=/story/menu-submenu--submenu)
    2. If you were able to recreate the issue above, attempt to do so again in this new context
    3. See that the submenus are always available no matter how much you futz with them.
-   [ ] _Test case 2_
    1. Open the branch locally
    2. Run the tests `yarn test:watch:focus menu`
    3. See that tests pass
    4. Comment out these lines of TS: https://github.com/adobe/spectrum-web-components/blob/test-overlay-timing/packages/overlay/src/ActiveOverlay.ts#L580-L582
    5. See that a test fails in at least Chromium

## Types of changes

## Checklist

### DIFF
--- a/packages/menu/test/submenu.test.ts
+++ b/packages/menu/test/submenu.test.ts
@@ -32,6 +32,7 @@ import { ActionMenu } from '@spectrum-web-components/action-menu';
 import '@spectrum-web-components/action-menu/sp-action-menu.js';
 import '@spectrum-web-components/menu/sp-menu-group.js';
 import '@spectrum-web-components/icons-workflow/icons/sp-icon-show-menu.js';
+import { ActiveOverlay } from '@spectrum-web-components/overlay';
 
 async function styledFixture<T extends Element>(
     story: TemplateResult,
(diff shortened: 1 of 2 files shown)


hunterloftis:
Testing now... I wasn't able to repro without throttling my CPU down:

image

najikahalsema (approved):
Ran through the test cases and they all work as expected! :^)

hunterloftis (approved):
More async bookkeeping!

PR #2799, opened 2022-12-13
fix(documentation): add base search url

nickschaap:
## Description

Currently clicking on search results in production will lead you to 404 due to the fact that the base-url or SWC_DIR is not correctly set on production environments.

Carve out the build:search-index step from build:assets to conditionally apply build:search-index or build:search-index:prod based off the environment.

## Related issue(s)

-

## How has this been tested?

-   [ ] _Test case 1_
    1. Run `yarn docs:production`
    2. Inspect searchIndex.json in the generated output for item refs to be prefixed with /spectrum-web-components/

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/projects/documentation/package.json
+++ b/projects/documentation/package.json
@@ -33,6 +33,7 @@
         "@rollup/plugin-commonjs": "^21.0.1",
         "@rollup/plugin-json": "^5.0.0",
         "@rollup/plugin-node-resolve": "^13.1.3",
+        "@rollup/plugin-replace": "^5.0.1",
         "@toycode/markdown-it-class": "^1.2.4",
         "@types/lunr": "^2.3.3",
         "@types/prismjs": "^1.16.6",
--- a/projects/documentation/package.json
+++ b/projects/documentation/package.json
@@ -213,7 +213,8 @@
                     "cascade": false
                 },
                 {
-                    "script": "build:search-index"
+                    "script": "build:search-index",
+                    "cascade": false
                 }
             ]
         }
--- a/projects/documentation/package.json
+++ b/projects/documentation/package.json
@@ -15,6 +15,7 @@
         "build:postcss": "wireit",
         "build:production": "wireit",
         "build:search-index": "wireit",
+        "build:search-index:prod": "wireit",
         "build:tsc": "wireit",
         "copy-docs": "wireit",
         "serve": "wireit",
--- a/projects/documentation/package.json
+++ b/projects/documentation/package.json
@@ -140,7 +140,7 @@
                     "script": "build:assets"
                 },
                 {
-                    "script": "build:search-index"
+                    "script": "build:search-index:prod"
                 }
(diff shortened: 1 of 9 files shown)


projects/documentation/src/components/search-index.ts
```diff
@@ -72,7 +74,7 @@ export async function search(value: string): Promise<ResultGroup[]> {
             catagoryData.results.push({
                 name,
                 label: label(name),
-                url: item.ref,
+                url: BASE_URL + item.ref,
```
Westbrook, line 77: This work was previously being managed [here](https://github.com/adobe/spectrum-web-components/blob/main/projects/documentation/scripts/build-search-index.js#L58-L61) and [here](https://github.com/adobe/spectrum-web-components/blob/main/projects/documentation/scripts/build-search-index.js#L77-L80). I'd definitely like to continue to process this at index build time, rather than site build time. This would likely involve a deeper understanding of how we leveraging our scripting in association with [wireit](https://github.com/google/wireit) to build and run the docs site. Something along the lines of breaking `"build:search-index"` out of the `"build:assets"` script and running a context specific version in each of `"build"`/`"serve"`, which will likely be the same for the two contexts, and `"build:production"`, which would need a value set for `process.eng.SWC_DIR` or possibly update the underlying script to accept a flag in the CLI instead...

projects/documentation/scripts/build-ts.js
```diff
@@ -13,8 +13,6 @@ governing permissions and limitations under the License.
 import { build } from 'esbuild';
 import { litCssPlugin } from 'esbuild-plugin-lit-css';
 import postcss from 'postcss';
-// import { postCSSPlugins } from '../../../scripts/css-processing.cjs';
-import purgecss from '@fullhuman/postcss-purgecss';
```
nickschaap, line 17: Remove unused import

projects/documentation/src/components/styles.css
```diff
@@ -192,7 +192,7 @@ body {
     width: 36px;
     height: 31px;
     display: block;
-    margin-right: var(--spectrum-global-dimension-size-200);
```
nickschaap, line 195: Using logical operators for RTL margins of the sidenav header

projects/documentation/package.json
```diff
@@ -196,6 +211,9 @@
                 {
                     "script": "build:assets",
                     "cascade": false
+                },
+                {
+                    "script": "build:search-index"
```
Westbrook, line 216:
```suggestion
                    "script": "build:search-index",
                    "cascade": false
```

Westbrook (approved):
LGTM! Thanks for working through this. Really points out that we need a more reliable testing context for the code that actually makes it out to the live site for this, or we need to work on getting that context more like the review context. I've some ideas that I hopefully have some time for during the holiday lull this week and after the New Year.

PR #2803, opened 2022-12-15
fix(overlay): allow overlays to be positioned 0px from their targets

nickschaap:
## Description

While working on a project I was not able to open a popover if the computedPosition x position returned by floating-ui was 0 due to the way [we round numbers by device pixel ratio](https://github.com/adobe/spectrum-web-components/blob/main/packages/overlay/src/ActiveOverlay.ts#L437). It seems like this change was intentional #2121.

image

## Related issue(s)

-

## Motivation and context

Allows popovers to be positioned correctly if their computed x position is 0.

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/overlay/src/ActiveOverlay.ts
+++ b/packages/overlay/src/ActiveOverlay.ts
@@ -433,6 +433,9 @@ export class ActiveOverlay extends SpectrumElement {
         await (document.fonts ? document.fonts.ready : Promise.resolve());
 
         function roundByDPR(num: number): number {
+            if (num === 0) {
+                return num;
+            }
             const dpr = window.devicePixelRatio || 1;
             return Math.round(num * dpr) / dpr || -10000;
         }


packages/overlay/src/ActiveOverlay.ts
```diff
@@ -433,6 +433,9 @@ export class ActiveOverlay extends SpectrumElement {
         await (document.fonts ? document.fonts.ready : Promise.resolve());
 
         function roundByDPR(num: number): number {
+            if (num === 0) {
+                return num;
+            }
             const dpr = window.devicePixelRatio || 1;
             return Math.round(num * dpr) / dpr || -10000;
```
nickschaap, line 440: Not sure what returning -10000 is accomplishing here. Don't have full context

Westbrook:
This was specifically set to prevent erroneous placement to the top/left corner based on a timing issue. The reason it worked is this isn't calculating "distance from trigger", persay, as much as distance from top/left corner. In this way the number should _never_ be `0px`, particularly taking into account the Spectrum rules available here: https://spectrum.adobe.com/page/popover/#Container-padding

Would you be able to add a version of the issue that you're seeing to this branch or a demo site so I can take a closer look at why you are getting a `0px` response here?

nickschaap:
Here is a demo which may be helpful. Let me know if you need more context.

https://user-images.githubusercontent.com/4684894/208347986-e1e80029-8e67-4ae0-8b82-45bdb2762cf7.mov

It looks like the active-overlay offsetParent is already 8px inset resulting in the x position returned from computePosition to still be 0.

Westbrook:
This demo is showing the page without removing the `margin` applied to `body`, not explicitly due to the overlay's relationship to the trigger. That means it _shouldn't_ happen in a more common page delivery context, but it should be something that we can handle appropriately. It's difficult to test the context that we were working around, so I'm a little leery about changing things that could cause the issue to reemerge. In that way, I wonder if we move from `position: absolute` to `fixed` in both the CSS for `active-overlay` and the configuration for `floating-ui` if this situation would persist or not. 🤔

PR #2822, opened 2023-01-02
build: happy 2023!

Westbrook:
Tooling will bump the copyright year in the CSS files, so let's handle that in it's own PR so it doesn't dirty up any other changes.


Westbrook:
Merging to unblock downstream branches.

Issue #1207, opened 2021-02-23
[Tabs] Support overflow pattern


Westbrook:
Ensure responsive delivery of `sp-tab` elements by supporting the "with overflow" pattern of delivery outlines here: https://opensource.adobe.com/spectrum-css/tabs.html#basictabswithoverflow

Westbrook:
Spectrum also support a "scrolling" pattern than delivers some use controls for moving to see all of the available Tabs:
[
<img width="345" alt="image" src="https://user-images.githubusercontent.com/1156657/201188961-2050b95a-61cf-4415-bec0-b60e1d588164.png">
](https://spectrum.adobe.com/page/tabs/#Too-many-tabs)

Delivering this interface as a decorator for the `<sp-tabs>` element should make it relatively easy for users to opt-in to leverage this capability. This could take the shape of:

```
<sp-tabs-overflow>
    <sp-tabs>
        <sp-tab>1<sp-tab>
        <sp-tab>2<sp-tab>
        <sp-tab>...<sp-tab>
        <sp-tab>N<sp-tab>
        <sp-tab-panel>1<sp-tab-panel>
        <sp-tab-panel>2<sp-tab-panel>
        <sp-tab-panel>...<sp-tab-panel>
        <sp-tab-panel>N<sp-tab-panel>
    </sp-tabs>
</sp-tabs-overflow>
```

Things that we should take into account for an initial release:
- `<sp-tabs>` usage with and without `<sp-tab-panel>` elements
- Resize observing for overflow, depending on how this is done, there may be benefit to sharing this work between this and an updated version of https://github.com/adobe/spectrum-web-components/blob/main/packages/tabs/src/Tabs.ts#L200
- these styles are not currently provided by Spectrum CSS, so maybe we could get them added to Core Tokens for the future @pfulton?

Things that could be included, but are acceptable for future scope:
- major visual differentiation between Express and Classic delivery (likely expressed in the form of fully opaque buttons vs ones that are placed on a gradient of transparency)
- managing overflow of `<sp-tabs direction="vertical">` elements
- properties to customize scroll distance
- access to managing scroll behavior beyond directly addressing `sp-tabs::part(tablist)`, though maybe that's enough

pfulton:
Pinging @GarthDB for visibility on the tokens side of things here.

PR #2833, opened 2023-01-05
feat(tabs-overflow): sp-tabs-overflow component

monkeyjabs:
## Description
This issue implements #1207, adding a support Tabs overflow. This is a decorator component for `<sp-tabs>` which enable scrolling of the horizontal tabs when there is not enough window's width.

## Related issue(s)
- implements #1207 

## Motivation and context
When a horizontal tabs component does not have enough width space to render all the tabs it gets clipped. We want to enable a way for the component automatically show a horizontal scroller when there is not enough space. This use case common for mobile screen.

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

*Usage with Tabs and Tab Panel*
PixelSnap 2023-01-05 at 11 04 51

*Usage with Tabs*
PixelSnap 2023-01-05 at 11 05 32

## Types of changes

## Checklist

## Best practices


packages/tabs-overflow/src/spectrum-config.js
```diff
@@ -0,0 +1,24 @@
+/*
```
Westbrook, line 1: This pattern does not currently have a corollary in Spectrum CSS, so we should remove this file so all of the tooling can run appropriately...

packages/tabs-overflow/src/TabsOverflow.ts
```diff
@@ -0,0 +1,153 @@
...
+    query,
+    queryAssignedElements,
+    state,
+} from '@spectrum-web-components/base/src/decorators.js';
+import { classMap } from '@spectrum-web-components/base/src/directives.js';
+import { ResizeController } from '@lit-labs/observers/resize_controller.js';
+import { Tabs } from '@spectrum-web-components/tabs';
+import '@spectrum-web-components/action-button/sp-action-button.js';
+import '@spectrum-web-components/icons-workflow/icons/sp-icon-chevron-left.js';
+import '@spectrum-web-components/icons-workflow/icons/sp-icon-chevron-right.js';
```
nickschaap, line 29: You can probably get away with just importing on of these and using CSS transforms to flip. Also not sure if these should be coming from `@spectrum-web-components/icons-ui` following practice from`sp-accordion`. You'll also want to list the package as a dependency

packages/tabs-overflow/package.json
```diff
@@ -0,0 +1,70 @@
...
+    },
+    "description": "Web component implementation of a Spectrum design TabsOverflow",
+    "license": "Apache-2.0",
+    "repository": {
+        "type": "git",
+        "url": "https://github.com/adobe/spectrum-web-components.git",
+        "directory": "packages/tabs-overflow"
+    },
+    "author": "",
+    "homepage": "https://adobe.github.io/spectrum-web-components/components/tabs-overflow",
```
nickschaap, line 15: Not sure this the right homepage.

packages/tabs-overflow/src/tabs-overflow.css
```diff
@@ -0,0 +1,116 @@
...
+        --spectrum-global-dimension-static-size-500
+    );
+    --sp-tabs-overflow-button-height: var(--spectrum-global-dimension-size-450);
+    --sp-tabs-overflow-icon-color: var(--spectrum-gray-800);
+    --sp-tabs-overflow-shadow-color: var(--spectrum-gray-100);
+    --sp-tabs-overflow-shadow-width: 50px;
+}
+
+:host([size='s']) {
+    --sp-tabs-overflow-button-height: var(--spectrum-global-dimension-size-450);
```
Westbrook, line 33: I don't remember anywhere where I found these numbers, but it looks like using `500`, `600`, etc matches the Tabs height more exactly...

packages/tabs-overflow/README.md
```diff
@@ -0,0 +1,45 @@
...
+```
+import { TabsOverflow } from '@spectrum-web-components/tabs-overflow';
+```
+
+## Example
+
+To use the `<sp-tabs-overflow>` component, simply wrap it around the `<sp-tabs>` element as a child element:
+
+```html
+<sp-tabs-overflow>
```
Westbrook, line 33: We likely want to do something with this demo that will _always_ trigger the overflow UI, maybe a `max-width` or something?

packages/tabs/src/Tabs.ts
```diff
@@ -101,6 +104,9 @@ export class Tabs extends SizedMixin(Focusable) {
     @property()
     public label = '';
 
+    @property({ type: Boolean, reflect: true })
```
Westbrook, line 107: It doesn't look like we are styling or querying anything off of this attribute, does it need to reflect?

packages/tabs-overflow/src/TabsOverflow.ts
```diff
@@ -0,0 +1,158 @@
...
+    protected override render(): TemplateResult {
+        const { canScrollRight, canScrollLeft } = this.overflowState;
+        return html`
+            <div
+                class=${classMap({
+                    'tabs-overflow-container': true,
+                    'left-shadow': canScrollLeft,
+                    'right-shadow': canScrollRight,
+                })}
+                @sp-tabs-scroll=${this._updateScrollState}
```
Westbrook, line 125: Can we move this down to the `<slot>` element instead?

Westbrook (commented):
Please rebase this branch to `main` so we can get that one CI light to go green for you 🤞🏼 

After that, one (or two) last big things:

1. Do we want this to be its own package 100%? @najikahalsema @hunterloftis any thoughts on what you would prefer? This could be a completely opt-in component registration _within_ the Tabs package, rather than its own standalone package. Side effects being:
  - you need less dependencies in-sync to leverage both the `<sp-tabs>` and `<sp-tabs-overflow>` pattern
  - you get one less "Components" navigation item in the docs site, and you move it to the "Tabs" group, assuming you rename the new `README.md` to `tabs-overflow.md` when bringing all of this into that directory
2. Whether we leave the above as is, or not, these files will need to be included in the `bundle` package in order to be fully a part of the docs site and part of that installation option. You can test the docs site with `yarn docs:start`, but to get the component definition included you'll need to add it to `https://github.com/adobe/spectrum-web-components/blob/overlay-mixins/tools/bundle/package.json#L123`. There are four places within that package that `tabs-overflow` would need to be added, please follow how the tabs package is included therein.

packages/tabs/test/tabs-overflow.test.ts
```diff
@@ -0,0 +1,161 @@
...
+import { TabsOverflow } from '../src/TabsOverflow';
+import { Tab } from '../src/Tab';
+import { ActionButton } from '@spectrum-web-components/action-button';
+import '@spectrum-web-components/theme/sp-theme.js';
+import '@spectrum-web-components/theme/scale-medium.js';
+import '@spectrum-web-components/theme/theme-light.js';
+import '../sp-tabs-overflow.js';
+import '../sp-tabs.js';
+import '../sp-tab.js';
+import '../sp-tab-panel.js';
```
Westbrook, line 25: Try updating these to `@spectrum-web-components/tabs/...` rather than `../` the relative path doesn't register the dev vs prod build separations and might be messing up the code coverage.

Westbrook (approved):
LGTM!

https://sp-tabs-overlow--spectrum-web-components.netlify.app/components/tabs-overflow/

Thanks for pushing through the coverage, VRT and reorg requests. 🙇🏼 Not sure if we'll have another release this week, but look for this to be part of the next SWC release!

PR #2835, opened 2023-01-09
test: unskip smoke tests

Westbrook:
Unskip smoke tests now that all recent updates to the Documentation site live.


nickschaap (approved):
LGTM. Nit: it makes sense to me to move the search field focus code in with the rest of the searching logic instead of splitting it up between beforeEach and searchFor.

PR #2850, opened 2023-01-17
Make "sp-update-overlays" work from a clean render

### LINKED ISSUE #2813: [Bug]: Overlay.update() should release previously used measures before updating
### Code of conduct

### Impacted component(s)

Overlay

### Expected behavior

`Overlay.update()` should update not only the position of an overlay, but also it's available size, when restricted by the view port.

### Actual behavior

If the overlay has previously been sized, then the previous size will override the requirements of the view port.

### Screenshots

_No response_

_No response_

1. Go to [here](https://studio.webcomponents.dev/edit/1YZCcOq4PMMLIJVekDBk/src/index.ts?branch=overlay%40wfVpXdsTYhf9mY2slxW4t2Y2SjC3&p=stories)
2. Click on "Picker"
3. See that `height` and `max-height` have been applied to the `active-overlay`.
4. Click on "More things"
5. See that content becomes available on scroll, but the overlay not resize to the available size of the view port.

https://studio.webcomponents.dev/edit/1YZCcOq4PMMLIJVekDBk/src/index.ts?branch=overlay%40wfVpXdsTYhf9mY2slxW4t2Y2SjC3&p=stories

_No response_

Westbrook:
## Description
When dispatching `sp-update-overlays` the expectation is the the size AND position of Overlays is updated. This change ensures that size rules applied as part of previous layout application passes is _removed_ before the update workflow is reentered.

## Related issue(s)
- fixes #2813

## How has this been tested?
-   [ ] _Test case 1_
    1. Go [here](https://sp-update-overlays--spectrum-web-components.netlify.app/storybook/?path=/story/overlay--accordion)
    2. Open the Overlay
    3. Toggle open Accordion Items until the overlay fills the vertical height of the viewport
    4. See that it is updated to fill the viewport but never overflows the viewport.
-  [ ] _Test case 2_
    1. Go [here](https://studio.webcomponents.dev/edit/1YZCcOq4PMMLIJVekDBk/src/index.ts?branch=overlay%40wfVpXdsTYhf9mY2slxW4t2Y2SjC3&p=stories)
    2. See how the size wasn't appropriately updated before.

## Types of changes

## Checklist

### DIFF
--- a/packages/overlay/src/ActiveOverlay.ts
+++ b/packages/overlay/src/ActiveOverlay.ts
@@ -427,8 +427,8 @@ export class ActiveOverlay extends SpectrumElement {
     };
 
     public resetOverlayPosition = (): void => {
-        this.style.setProperty('max-height', '');
-        this.style.setProperty('height', '');
+        this.style.removeProperty('max-height');
+        this.style.removeProperty('height');
         this.initialHeight = undefined;
         this.isConstrained = false;
         // force paint
--- a/packages/overlay/src/ActiveOverlay.ts
+++ b/packages/overlay/src/ActiveOverlay.ts
@@ -288,7 +288,7 @@ export class ActiveOverlay extends SpectrumElement {
             await this.updateOverlayPosition();
             document.addEventListener(
                 'sp-update-overlays',
-                this.setOverlayPosition
+                this.resetOverlayPosition
             );
         }
         if (this.placement && this.placement !== 'none') {
@@ -400,7 +400,7 @@ export class ActiveOverlay extends SpectrumElement {
         }
     }
 
-    private initialHeight!: number;
+    private initialHeight: number | undefined;
     private isConstrained = false;
 
     public async placeOverlay(): Promise<void> {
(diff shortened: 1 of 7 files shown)


najikahalsema (commented):
It looks good to me, but should we be testing this in our test suite somehow? 🤔

Westbrook:
@najikahalsema good call, added the equivalent of Test Case 1 as a unit test in 849a9dbe399

packages/overlay/src/ActiveOverlay.ts
```diff
@@ -426,6 +426,16 @@ export class ActiveOverlay extends SpectrumElement {
         this.setOverlayPosition();
     };
 
+    public resetOverlayPosition = (): void => {
+        this.style.setProperty('max-height', '');
```
hunterloftis, line 430: Are `style.setProperty('max-height', '')` and `style.removeProperty('max-height')` equivalent?

hunterloftis (dismissed):

packages/overlay/src/ActiveOverlay.ts
Westbrook: Oooh, I've not used that API, it seems like it should be, so I'll test that as it seems like a nicer future to maintain.

PR #2868, opened 2023-01-23
fix: ensure "wheel" interactions lead to a "change" event

### LINKED ISSUE #2867: [Bug]:  Change event is not getting triggered at the end of mouse scroll on sp-number-field
### Code of conduct

### Impacted component(s)

sp-number-field

### Expected behavior

Mouse scroll  on <sp-number-field> changes its value. It emits input events when the value is changing but at at the end of the scroll, no change event is triggered. 

### Actual behavior

It is expected to emit change event at the end of scroll operation. 

### Screenshots

_No response_

1. Use the below code sample to create `<sp-number-field>`
2. Click and scroll.
3. Observe the values getting increase/decreased.
4. Stop scrolling and move the focus out if need.
5. Observe the change event not getting triggered at the end of scroll operation.

```
 <sp-number-field 
                hide-stepper
                title="sample"
                value=720
                @change=<some function which should get triggered on change event>
></sp-number-field>
```

_No response_

Westbrook:
## Description
When interacting with a Number Field via the `wheel` event, ensure that a `change` event is dispatched, to ensure batched listening is successful. Do so by waiting 100ms after the last `input` event is fired to fire a `change` event. Timing here seems pretty flexible, and I could be convinced of a shorter debounce, but this feels like a good balance between firing the `change` and not firing it all the time, as the `wheel` event could be stop and start and allows for breaks in the interaction.

## Related issue(s)

- fixes #2867 

## How has this been tested?

-   [ ] _Test case 1_
    1. Go [here](https://number-field-change--spectrum-web-components.netlify.app/storybook/index.html?path=/story/number-field--default)
    2. Interact with the Number Field with the `wheel` event
    3. Check the "Actions" Tab in the Storybook interface
    4. See that a large number of `onInput` events are being handled and a smaller number of `onChange` events are being handled.

## Types of changes

## Checklist

### DIFF
--- a/packages/number-field/src/NumberField.ts
+++ b/packages/number-field/src/NumberField.ts
@@ -38,6 +38,8 @@ import chevronStyles from '@spectrum-web-components/icon/src/spectrum-icon-chevr
 import styles from './number-field.css.js';
 
 export const FRAMES_PER_CHANGE = 5;
+// Debounce duration for inserting a `change` event after a batch of `wheel` originating `input` events.
+export const CHANGE_DEBOUNCE_MS = 100;
 export const indeterminatePlaceholder = '-';
 export const remapMultiByteCharacters: Record<string, string> = {
     '１': '1',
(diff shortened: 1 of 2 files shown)


packages/number-field/src/NumberField.ts
```diff
@@ -318,6 +320,13 @@ export class NumberField extends TextfieldBase {
             : event.deltaY / Math.abs(event.deltaY);
         if (direction !== 0 && !isNaN(direction)) {
             this.stepBy(direction * (event.shiftKey ? this.stepModifier : 1));
+            // Dispatch a change event after no new `input` events are queued by `wheel` interactions for 100ms.
+            clearTimeout(this.queuedChangeEvent);
+            this.queuedChangeEvent = setTimeout(() => {
+                this.dispatchEvent(
+                    new Event('change', { bubbles: true, composed: true })
+                );
+            }, 100) as unknown as number;
```
hunterloftis, line 329: The comment above addresses this number, but just to present an alternative, we could also consider something like:

```ts
const CHANGE_DEBOUNCE_MS = 100;
```

hunterloftis (dismissed):

packages/number-field/src/NumberField.ts
Westbrook: Kind of like _both_ as it allows me to export this and leverage it in the test as well...

Issue #2807, opened 2022-12-20
[Bug]: When using Switch or Checkbox component in a table row, exception occur when the Switch or Checkbox is clicked


monkeyjabs:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Table, TableRow

### Expected behavior

No exception should occur when the Switch or Checkbox component is clicked. The TableRow component should account for other components that could trigger the `change` event in the `handleChange()` method.

### Actual behavior

`TypeError: Cannot read properties of undefined (reading 'checked')`

### Screenshots

Example
![PixelSnap 2022-12-20 at 11 49 54](https://user-images.githubusercontent.com/4380525/208754128-5c1a052c-7140-4d33-b89d-4fac302b56f4.png)

error
![PixelSnap 2022-12-20 at 11 49 05](https://user-images.githubusercontent.com/4380525/208753966-01b1dff5-2254-48f7-bfa7-8079b2fb9bdd.png)

### What browsers are you seeing the problem in?
Any Browser

### How can we reproduce this issue?

1. Go to this playground example: https://studio.webcomponents.dev/edit/U6z97bJLW2cZahcDZ4Mh/src/index.ts?p=stories
2. Click on Switch toggle
3. Observe the error on screen and console

### Sample code that illustrates the problem

```
<sp-table>
  <sp-table-head>
      <sp-table-head-cell>Flag</sp-table-head-cell>
      <sp-table-head-cell>Value</sp-table-head-cell>
  </sp-table-head>
  <sp-table-body>
    <sp-table-row>
        <sp-table-cell>Enable</sp-table-cell>
        <sp-table-cell>
            <sp-switch>
                Value
            </sp-switch>
        </sp-table-cell>
    </sp-table-row>
  </sp-table-body>
</sp-table>
```

bangank36:
Hello @monkeyjabs, I'm new contributor here and would like to chime in
I think the issue is caused on this line https://github.com/adobe/spectrum-web-components/blob/a227f511afef4e23d1a328465d7c5e6da4ae0199/packages/table/src/TableRow.ts#L57

As soon as a `sp-table-row` found a `sp-checkbox` on its cells, it considers the cell as `sp-table-checkbox-cell` right away which will lead to the error above

Westbrook:
[...]

PR #2919, opened 2023-02-09
fix(table): checking for checkbox in Table handleChange

bangank36:
## Related issue(s)

- #2807 

## How has this been tested?

- Create a template which has a `sp-checkbox` wrapped inside a `sp-table-cell`
- Confirm the checkboxes are working properly without the error
```
<sp-table>
  <sp-table-head>
      <sp-table-head-cell>Flag</sp-table-head-cell>
      <sp-table-head-cell>Value</sp-table-head-cell>
  </sp-table-head>
  <sp-table-body>
    <sp-table-row>
        <sp-table-cell>Enable</sp-table-cell>
        <sp-table-cell>
            <sp-switch>
                Value
            </sp-switch>
        </sp-table-cell>
    </sp-table-row>
  </sp-table-body>
</sp-table>
```

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/table/src/Table.ts
+++ b/packages/table/src/Table.ts
@@ -363,11 +363,15 @@ export class Table extends SizedMixin(SpectrumElement, {
         };
         if (!rowItem.value) {
             const { checkbox } = target as TableCheckboxCell;
-            if (checkbox.checked || checkbox.indeterminate) {
-                this.selectAllRows();
-            } else {
-                this.deselectAllRows();
+            if (checkbox) {
+                if (checkbox.checked || checkbox.indeterminate) {
+                    this.selectAllRows();
+                } else {
+                    this.deselectAllRows();
+                }
             }
+
+            return;
         } else {
             switch (this.selects) {
                 case 'single': {
--- a/packages/table/src/Table.ts
+++ b/packages/table/src/Table.ts
@@ -363,15 +363,13 @@ export class Table extends SizedMixin(SpectrumElement, {
         };
         if (!rowItem.value) {
             const { checkbox } = target as TableCheckboxCell;
-            if (checkbox) {
-                if (checkbox.checked || checkbox.indeterminate) {
-                    this.selectAllRows();
-                } else {
-                    this.deselectAllRows();
-                }
-            }
+            if (!checkbox) return;
 
-            return;
+            if (checkbox.checked || checkbox.indeterminate) {
+                this.selectAllRows();
+            } else {
+                this.deselectAllRows();
+            }
         } else {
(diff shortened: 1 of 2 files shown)


packages/table/src/Table.ts
```diff
@@ -363,11 +363,15 @@ export class Table extends SizedMixin(SpectrumElement, {
...
-                this.deselectAllRows();
+            if (checkbox) {
+                if (checkbox.checked || checkbox.indeterminate) {
+                    this.selectAllRows();
+                } else {
+                    this.deselectAllRows();
+                }
             }
+
+            return;
```
Westbrook, line 374: Not 100%, but pretty sure you want something closer to:
```
if (!checkbox) return;

// ... rest of code here
```
But, I'd have to dig deeper and run the workflow to confirm.

bangank36: Working well

packages/table/src/TableRow.ts
```diff
@@ -80,17 +84,14 @@ export class TableRow extends SpectrumElement {
 
     protected handleClick(event: Event): void {
         if (
-            event
-                .composedPath()
-                .find(
-                    (node) => (node as HTMLElement).localName === 'sp-checkbox'
-                )
+            ['sp-table-cell'].includes((event.target as HTMLElement).localName)
```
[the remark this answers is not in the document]
bangank36: hmm, things're about to gettings weird here, the reason for this change is for the bug I found: when there is a `selectable component` ( like switch ) on a `selectable` row, clicking on the component may mess up the selection of the row
Video in [the comment](https://github.com/adobe/spectrum-web-components/issues/2807#issuecomment-1421950981)
After refactoring and retest I still see some hiccups as below
- If we only check for `sp-table-cell` like I am suggesting
   - I will solve the issue when clicking on switch on a selectable row, but since it only expects `sp-table-cell` in the logic, when we try to click on the `sp-table-checkbox-cell` ( on the cell itself, not the checkbox ), then the event will be skipped
- If we check for both local names `sp-checkbox` AND `sp-table-checkbox-cell`, the issue still be there, since not only the `sp-checkbox` might be added into the cell, but also others like `sp-color-area` or `sp-switch`

Westbrook:
This change is causing tests to fail and likely would be benefitted by tests preventing it from future regressions.

Westbrook (commented):
Sorry for not getting back to these changes sooner.

I think they're starting to line up appropriately. Would you be able to structure a couple of tests to support these changes in order to prevent regression in the future?

PR #2801, opened 2022-12-14
refactor(textfield): migrate spectrum tokens

lunarfusion:
## Description
This is a draft PR of the beta release of textfield. Hold for full release.

## Related issue(s)

-

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

-   [ ] Bug fix (non-breaking change which fixes an issue)
-   [ ] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to change)
-   [ ] Chore (minor updates related to the tooling or maintenance of the repository, does not impact compiled assets)

## Checklist

-   [ ] I have signed the [Adobe Open Source CLA](http://opensource.adobe.com/cla.html).
-   [ ] My code follows the code style of this project.
-   [ ] If my change required a change to the documentation, I have updated the documentation in this pull request.
-   [ ] I have read the **[CONTRIBUTING](<(https://github.com/adobe/spectrum-web-components/blob/main/CONTRIBUTING.md)>)** document.
-   [ ] I have added tests to cover my changes.
-   [ ] All new and existing tests passed.
-   [ ] I have reviewed at the Accessibility Practices for this feature, see: [Aria Practices](https://www.w3.org/TR/wai-aria-practices/)

## Best practices

This repository uses conventional commit syntax for each commit message; note that the GitHub UI does not use this by default so be cautious when accepting suggested changes. Avoid the "Update branch" button on the pull request and opt instead for rebasing your branch against `main`.

PR #2920, opened 2023-02-09
chore(textfield,number-field): use core tokens

### LINKED ISSUE #2931: [Bug]: Inconsistent multiline textfield focus state for Express theme
### Code of conduct

### Impacted component(s)

sp-textfield

### Expected behavior

The focus ring for a multiline textfield should be visually consistent with the focus ring for single line.

### Actual behavior

The focus ring for a multiline textfield is not consistent with the focus ring for single line, as we can see [here](https://j9g3m1.csb.app/) or in the attached screenshots.

### Screenshots

image

1. Go to https://j9g3m1.csb.app/ 

_No response_

_No response_

lunarfusion:
## Description

This is a draft PR of the beta release of textfield and the beta release of stepper (numberfield). Hold for full release.

Jira issue CSS-153 https://jira.corp.adobe.com/browse/CSS-153
Jira issue CSS-185 https://jira.corp.adobe.com/browse/CSS-185
Jira issue CSS-161 https://jira.corp.adobe.com/browse/CSS-161
Jira issue CSS-388 https://jira.corp.adobe.com/browse/CSS-388

Previous PR (closed): https://github.com/adobe/spectrum-web-components/pull/2801

## Related issue(s)
- fixes #2931

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices


mlogsdon18:
All failures are as expected. Updated design changes that are causing these include:

- Updated placeholder style from italics to regular to match search 
- Border color updated to meet accessibility contrast requirements 
- Both themes' error state border color updated
- Alert icon and validation icon colors updated
- Keyboard focus states updated to have a gap
- TextField minimum spacing from text to icon updated to be consistent with TextArea
- Character count alignment updated
- Field label and Quiet field alignment fixed to align with the baseline of the field label
- Express border width updated 

These changes are all causing slight differences in appearance.

Westbrook:
Can you look specifically at this one where the Search icon is now missing when disabled? https://185cdd14841c2b3204c02f4510e5170b--spectrum-web-components.netlify.app/review/#SearchStories/Default.png

mlogsdon18:
@Westbrook this has been updated with the new release of Textfield and Stepper and it looks to me like the VRTs are good except for the hideStepper option. It looks like the changes you all are doing to the class names are causing the styling to not apply properly, specifically this 
```
.spectrum-Stepper:not(.hide-stepper) .spectrum-Stepper-input {
  min-inline-size: 0;
  border-start-end-radius: 0;
  border-end-end-radius: 0;
  border-inline-end-width: 0;
}
```
When it gets compiled in SWC it becomes 
```
#textfield:not(.hide-stepper) .input {
    border-end-end-radius: 0;
    border-inline-end-width: 0;
    border-start-end-radius: 0;
    min-inline-size: 0;
}
```
`#textfield` doesn't have the `hide-stepper` class so this CSS no longer works and therefore breaks the hideStepper variant. I looked through the config to see if we could change how it compiled `.spectrum-Stepper` and I'm not totally sure how to update it without breaking the component.

mlogsdon18:
@Westbrook I actually just went ahead and added a few lines to the custom CSS that was already in place for the `number-field` component to allow the hide stepper styles to work. I think this should finally be good to go!

If you could take a look as soon as you could that would be awesome, as this is blocking some other components.

packages/textfield/src/textfield.css
```diff
@@ -56,7 +56,25 @@ textarea {
...
+                var(--spectrum-text-area-min-block-size)
+            ) +
+            var(
+                --mod-textfield-focus-indicator-gap,
+                var(--spectrum-textfield-focus-indicator-gap)
+            ) * 2
+    );
+}
+
 :host([grows]) #sizer {
```
Westbrook, line 73: The padding rules for this need to be updated to match those applied to the `<textarea>` roughly:
```
padding-block-end: var(--mod-textfield-spacing-block-end,var(--spectrum-textfield-spacing-block-end) );
    padding-block-start: var(--mod-textfield-spacing-block-start,var(--spectrum-textfield-spacing-block-start) );
    padding-inline: var(--mod-textfield-spacing-inline,var(--spectrum-textfield-spacing-inline) );
```

packages/textfield/src/textfield.css
```diff
@@ -56,7 +56,25 @@ textarea {
     overflow: hidden;
 }
 
+:host([grows]) #textfield:after {
```
Westbrook, line 59: If we fully move to the Spectrum CSS implementation here, Textfield can be delivered like this: 
image


I'm not sure you can achieve this layout with `box-shadow`. It may be necessary to move to `outline` for the focus loupe and get over the fact that older versions of Safari will receive square corners.

Westbrook (commented):
Feels like we're moving in a good direction. We're gonna have a fix for the VRT previews in `main` at some point today that will be good to rebase onto. For the Express focus loupe, let me know if it would be more productive to discuss that in person. We have Office Hours on Wednesday at 8am ET, if you were able to join, but could make other time as well if that was needed.

Westbrook:
There's also a regression in https://327025922f3f81f02c6b86a8a4b36cbc--spectrum-web-components.netlify.app/review/#SearchStories/focusedOverflowing.png

Westbrook:
I wonder if the weirdness in the rounded corners [here](https://327025922f3f81f02c6b86a8a4b36cbc--spectrum-web-components.netlify.app/review/#NumberFieldStories/Default.png) comes from the fact that the buttons containers is (in medium scale) 23px wide with 1px border and `box-sizing: border-box;` while the buttons inside of it are also 23px wide, which means they technically overflow the now 21px available horizontally within the container? 💭

image

Westbrook:
I think [this](https://327025922f3f81f02c6b86a8a4b36cbc--spectrum-web-components.netlify.app/review/#TextareaStories/growsWithLargeWords.png) is just the padding that you added back later, but it may be that the [merge artifact](https://app.circleci.com/pipelines/github/adobe/spectrum-web-components/15651/workflows/a181b17d-3d3c-492b-9b87-3819565247cb/jobs/323098) that's currently in the branch is preventing the correct VRT from being published.

mlogsdon18:
Ok I've made some updates that should get us closer to what we currently have. This is just a partial migration of stepper, the full one will be coming later once we get designs and tokens for that, so that's why this is proving a bit difficult

packages/textfield/src/textfield.css
Westbrook: @pfulton can you help us on a path forward here. This issue needs to be addressed in some way. `outline` would work and aligns with my preference for progressive enhancement. In a way it could begin a shift towards a simpler focus loupe across the whole library, but it or something like this is a requirement here.

packages/textfield/src/textfield.css
Westbrook: Looks so much nicer, here!

Westbrook (commented):
Not sure the Express Number Field is shippable in this state: https://feat-textfield-core-tokens--spectrum-web-components.netlify.app/storybook/?path=/story/number-field--default&sp_theme=express 

image

packages/textfield/src/textfield.css
pfulton: I think a move toward `outline` could be ok here, but let me double-check with design folks before I make a final call.

pfulton: By the way, current MacOS Safari is ok with the `border-radius` values. iOS Safari, however, is not.

**MacOS:**

Screen Shot 2023-05-11 at 1 42 12 PM

---

**iOS:**

Simulator Screen Shot - iPhone 14 Pro - 2023-05-11 at 13 41 58

Westbrook: Which version are you on? 
IMG_5195

mlogsdon18:
> Not sure the Express Number Field is shippable in this state: https://feat-textfield-core-tokens--spectrum-web-components.netlify.app/storybook/?path=/story/number-field--default&sp_theme=express
> 
> image

Whoa yeah don't know what happened there. I may have to tweak the padding and do one more release to make sure that icon doesn't shrink. Is that all that you were referring to one this one? I know the design is different but that's the direction PJ gave us for this one.

packages/textfield/src/textfield.css
pfulton: 😱 hrmm `16.2` according to the simulator.

mlogsdon18:
@Westbrook Ok I've updated the Textfield and Stepper releases. The Express button issue should be fixed and I've updated the Textfield to use an outline which has fixed the resize issue with the focus indicator. Let me know if you find anything else but we should be close!

Westbrook:
This [URL](https://feat-textfield-core-tokens--spectrum-web-components.netlify.app/storybook/?path=/story/number-field--default&sp_theme=express) still seems to exhibit the issues I questioned before. What I'm seeing is the buttons pressed against the right edge of the Textfield. It was my understanding that these sorts of "infield" buttons where supposed to have the same distance from the border on all three sides...

Also, if you zoom in you can see that the border radius is not as expected. The up button currently has two rounded corners while the down button only has one... 🤔 

image

Some of this may be related to tokens, which has been updated a good bit in `main`, if you were to catch up to that branch.

mlogsdon18:
@Westbrook I have a fix up in Spectrum CSS for the button spacing. In terms of the border radius, it looks like when the CSS is getting compiled for SWC, the order of the properties is switching but only for this one section of CSS. The CSS we have written for the `.stepUp` button is 
```

**Comment by Westbrook** [comment]
@mlogsdon18 this seems to be a fun bug with one of our dependencies 🤦🏼‍♂️  https://github.com/cssnano/cssnano/issues/1502

Would it be possible for CSS to output something like the following for the near future?

```css
.selector {
  border-start-start-radius: 0;
  border-start-end-radius: 0;
  border-end-start-radius: 0;
  border-end-end-radius: 0;
}
```

mlogsdon18:
@Westbrook Ok I made the border-radius properties more specific so the issue doesn't look like it's there anymore!

Westbrook (approved):
Thanks for taking the reigns on this @mlogsdon18. LGTM!

PR #2934, opened 2023-02-15
fix: #2933 by adding optional variant property to SideNav

### LINKED ISSUE #2933: [Bug]: Sidenav variant property is missing in typescript type definition
### Code of conduct

### Impacted component(s)

@spectrum-web-components/sidenav

### Expected behavior

According to the [Sidenav](https://opensource.adobe.com/spectrum-web-components/components/sidenav/#multi-level) documentation, users can this property to render multi-layers Side navigation. This property should be usable in a typescript project. 

### Actual behavior

The typescript type definition of this property is missing. https://unpkg.com/browse/@spectrum-web-components/sidenav@0.13.9/src/Sidenav.d.ts. It will lead to ts compile error in a typescript project if I use it.

### Screenshots

image

_No response_

1. `npm i @spectrum-web-components/sidenav@latest`
2. `cd node_modules/@spectrum-web-components/sidenav`
3. `code src/sp-sidenav.d.ts`

or just go to https://unpkg.com/browse/@spectrum-web-components/sidenav@0.13.9/src/Sidenav.d.ts

The latest type definition for `SideNav` looks like:
```
/**
 * @element sp-sidenav
 *
 * @slot - the Sidenav Items to display
 * @fires change - Announces a change in the `value` property of the navigation element.
 * This change can be "canceled" via `event.preventDefault()`.
 */
export declare class SideNav extends Focusable {
    static get styles(): CSSResultArray;
    private items;
    startTrackingSelectionForItem(item: SideNavItem): void;
    stopTrackingSelectionForItem(item: SideNavItem): void;
    rovingTabindexController: RovingTabindexController<SideNavItem>;
    manageTabIndex: boolean;
    value: string | undefined;
    private handleSelect;
    focus(): void;
    blur(): void;
    click(): void;
    get focusElement(): SideNavItem | SideNav;
    private isDisabledChild;
    private handleSlotchange;
    protected render(): TemplateResult;
    protected willUpdate(): void;
    protected updated(changes: PropertyValues): void;
}
```

_No response_

jianliao:
## Description

Fixed #2933 

## Related issue(s)
#2933 

-

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/sidenav/src/Sidenav.ts
+++ b/packages/sidenav/src/Sidenav.ts
@@ -74,6 +74,9 @@ export class SideNav extends Focusable {
     @property({ reflect: true })
     public value: string | undefined = undefined;
 
+    @property({ reflect: true })
+    public variant?: 'multilevel';
+
     private handleSelect(
         event: CustomEvent<SidenavSelectDetail> & { target: SideNavItem }
     ): void {
--- a/packages/sidenav/src/Sidenav.ts
+++ b/packages/sidenav/src/Sidenav.ts
@@ -74,8 +74,11 @@ export class SideNav extends Focusable {
     @property({ reflect: true })
     public value: string | undefined = undefined;
 
+    /**
+     * The multilevel variant will have multiple layers of hierarchical navigation items.
+     */
     @property({ reflect: true })
-    public variant?: 'multilevel';
+    public variant?: 'multilevel' = undefined;
 
     private handleSelect(
         event: CustomEvent<SidenavSelectDetail> & { target: SideNavItem }


packages/sidenav/src/Sidenav.ts
```diff
@@ -74,6 +74,9 @@ export class SideNav extends Focusable {
     @property({ reflect: true })
     public value: string | undefined = undefined;
 
+    @property({ reflect: true })
```
Westbrook, line 77: Can you add a description as to what this does so that it populates to: https://63ed2ff266722307437d351a--spectrum-web-components.netlify.app/components/sidenav/api/#attributes-and-properties

image

Local preview available via `yarn docs:start`. 

Here's an example of a property with a description: https://github.com/adobe/spectrum-web-components/blob/main/packages/accordion/src/Accordion.ts#L42

As seen in: https://63ed2ff266722307437d351a--spectrum-web-components.netlify.app/components/accordion/api/

image

jianliao: Done.
image

jianliao: I also set the default value as `undefined`.

Westbrook: Interesting, does it not set a default as `undefined` by way of the `?` usage? TIL!

Thanks for this.

Westbrook (approved):
LGTM! Thanks for the docs addition, too.

packages/sidenav/src/Sidenav.ts
jianliao: The default column will be empty if I don't explicitly set it as undefined.

Issue #1458, opened 2021-05-13
[Documentation] Add system level CSS Custom Property Documentation


Westbrook:
### Expected Behaviour
It's easy to understand and leverage system level CSS Custom Properties in an application that is composed of SWC elements.

### Actual Behaviour
There is _some_ component level documentation that is not the most productive as it doesn't map to static values at current.

### Prior art
React Spectrum does a version of this in https://react-spectrum.adobe.com/react-spectrum/styling.html#dimension-values  and https://react-spectrum.adobe.com/react-spectrum/styling.html#color-values that could stand as a reference.

mephraim:
In particular, documentation for the color related variables would be useful for our team. We're building components that we'd like to be themable within Spectrum, so most of our colors are specified using the Spectrum CSS color variables.

Westbrook:
Early thoughts lean toward building off of something like https://webcomponents.dev/edit/EIWbVqsSKu4TtpbPQm7b/src/index.ts where you can click each row to capture the CSS Custom Property (appropriate UX affordance, etc. TK, hand waves). The Spectrum docs use a UI like https://spectrum.adobe.com/page/color/ but the token can be quite long, so we'd likely want to rely on fewer columns for ease of consumption.

Questions:
- This only shows the token and the displayed color in each theme, how would you see wanting to use the actual color value in this experience? Spectrum switches between hex and RGB, but on a purely visual standpoint. Interested to hear if/how that was useful.
  - Kind of implies a done level 2 where you can do some sort of searching, but would want it to be useful if we were to invest in that way.
- This relies on the page level theme to switch the visible values, would you want to be able to manage that locally to the documentation, e.g. a picker that only effecting this list of colors?
- @mephraim is your team leveraging the scale tokens at all? And/or, would they want to?

Westbrook:
Initial draft of what this might become: https://custom-props-docs--spectrum-web-components.netlify.app/custom-properties/

First thoughts:
- expand to all color themes (via `sp-tabs` or an `sp-picker`)
- we should probably stick to `global` properties for a first pass
- match this with a scales version?
- investigate the right way to surface `alias` colors. they likely need more description to clarify their usage.
- we know that Spectrum will be expanding the colors available in `global` in the next quarter or so. should we wait for that to publish something like this?

Westbrook:
[...]

PR #2954, opened 2023-02-22
docs: Add custom vars table to documentation site

najikahalsema:
## Description

Created a custom variable table that allows users to view colours. This lives as its own app within the projects folder, but is also live on the documentation for `sp-theme`. I chose to add it to the Theme documentation since we link there directly from the "Getting started" page, and I felt that adding a new page would lead to less visibility. 

## Related issue(s)

fixes https://github.com/adobe/spectrum-web-components/issues/1458

## Motivation and context

Improving onboarding experience for new users. Improve understanding of how sp-theme works. Documenting custom properties somewhere more accessible. 
## How has this been tested?

- Go to documentation for sp-theme
- Scroll down to "What it's doing"
- See that the table loads. 
- Change theme via picker and see that colour values change.
- Scroll back up, change the documentation site's theme
- See that the theme of the colour values also changes (unless you have wireframe selected)

## Types of changes

## Checklist

## Best practices


projects/css-custom-vars-viewer/scripts/parse-json.js
```diff
@@ -0,0 +1,81 @@
...
+        const colourSubstring = key.search(colour);
+
+        if (colourSubstring !== -1) {
+            const infoValues = info.value;
+            if (infoValues !== undefined) {
+                // eslint-disable-next-line no-console
+                console.log('invalid format!');
+            } else {
+                colorArray.push({
+                    customVar: `--spectrum-global-color-${key}`,
```
nickschaap, line 52:
```suggestion
                    customVar: `--spectrum-${key}`,
```
Should these be pointing to core color tokens?

projects/css-custom-vars-viewer/package.json
```diff
@@ -0,0 +1,63 @@
...
+        "./sp-css-table.js": "./dist/src/sp-css-table.js",
+        "./custom-vars-viewer.js": "./dist/src/custom-vars-viewer.js"
+    },
+    "engines": {
+        "node": ">=16.14.2",
+        "yarn": ">=1.16.0"
+    },
+    "scripts": {
+        "analyze": "cem analyze --litelement",
+        "build": "tsc && yarn parse-json && npm run analyze -- --exclude dist",
```
Westbrook, line 22:
```suggestion
        "build": "yarn parse-json && tsc && yarn parse-json && npm run analyze -- --exclude dist",
```

I _think_ this is what is blocking the build...but, I've not been right in many tries so far 😬

projects/css-custom-vars-viewer/scripts/parse-json.js
najikahalsema: Yes. This is mainly just for an example project so I'm not too pressed about using the up-to-date tokens since they'll be deprecated anyway.

Westbrook (commented):
Love to see this in the actual docs site: https://halsema-custom-vars-viewer-2--spectrum-web-components.netlify.app/tools/theme/

Or when the ID request is added: https://halsema-custom-vars-viewer-2--spectrum-web-components.netlify.app/tools/theme/#color-tokens

The two nits in specific comments and then we should be good to merge.

Notes for future work/ideas:
- when you have a long list and scroll to the bottom and then select a single color, the Table doesn't recycle correctly and move scroll back to the top, not sure that's the Color Viewer, the Table, or the Virtualizer that's at fault
- we should make the clicking anywhere on the whole table row copy the property
- we should update and/or add an option for the property that is copied to be wrapped by `var(${clickedProperty})` so that it can go right into CSS later.
- these root colors don't sway Classic vs Express, but we'll need to include that, if future versions point into colors that change between the two (do they change between the two?)
- we continue to treat this new package as _separate_, which is fine, but we don't treat any of the other packages as _separate_, so we should confirm the work to make it more inline with the others as it could apply to adopting features from other teams in the future

Westbrook (approved):
LGTM! Let's get this shipped and slip those other notes/follow up tasks into our upcoming Q2 planning.

Great work!

Issue #2956, opened 2023-02-23
docs: update docs site to use core tokens in html examples


nickschaap:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Description of issue

We should update the docs site examples to use spectrum core tokens to encourage their use and avoid advertising the use of legacy Spectrum tokens

PR #2957, opened 2023-02-23
docs: use core tokens

nickschaap:
## Description

This change updates several component README files to use core tokens in their demos to encourage their use as we prepare to fully migrate over to Spectrum core tokens.

## Related issue(s)

#2956

-

## How has this been tested?

-   [ ] _Manual review of the preview docs site for changed README files_

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/badge/README.md
+++ b/packages/badge/README.md
@@ -108,31 +108,31 @@ Non-semantic badge colors are no longer supported directly by Spectrum and Spect
 <div style="display: flex; gap: var(--spectrum-spacing-75);">
     <sp-badge
         variant="seafoam"
-        style="--mod-badge-background-color-default: var(--spectrum-seafoam-600)"
+        style="--mod-badge-background-color-default: var(--spectrum-seafoam-background-color-default)"
     >
         Seafoam
     </sp-badge>
     <sp-badge
         variant="indigo"
-        style="--mod-badge-background-color-default: var(--spectrum-indigo-600)"
+        style="--mod-badge-background-color-default: var(--spectrum-indigo-background-color-default)"
     >
         Indigo
     </sp-badge>
     <sp-badge
         variant="purple"
-        style="--mod-badge-background-color-default: var(--spectrum-purple-600)"
+        style="--mod-badge-background-color-default: var(--spectrum-purple-background-color-default)"
     >
         Purple
     </sp-badge>
     <sp-badge
         variant="fuchsia"
-        style="--mod-badge-background-color-default: var(--spectrum-fuchsia-600)"
+        style="--mod-badge-background-color-default: var(--spectrum-fuchsia-background-color-default)"
     >
         Fuchsia
     </sp-badge>
     <sp-badge
         variant="magenta"
-        style="--mod-badge-background-color-default: var(--spectrum-magenta-600)"
+        style="--mod-badge-background-color-default: var(--spectrum-magenta-background-color-default)"
     >
         Magenta
     </sp-badge>
--- a/packages/badge/README.md
+++ b/packages/badge/README.md
@@ -108,30 +108,35 @@ Non-semantic badge colors are no longer supported directly by Spectrum and Spect
 <div style="display: flex; gap: var(--spectrum-spacing-75);">
     <sp-badge
         variant="seafoam"
+        static="black"
         style="--mod-badge-background-color-default: var(--spectrum-seafoam-background-color-default)"
     >
         Seafoam
     </sp-badge>
     <sp-badge
         variant="indigo"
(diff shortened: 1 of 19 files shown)


packages/overlay/overlay-trigger.md
```diff
@@ -49,8 +49,8 @@ Here a default `<overlay-trigger>` manages content that is triggered by click an
     }
 
     .tooltip {
-        background-color: var(--spectrum-global-color-gray-800);
-        color: var(--spectrum-global-color-gray-50);
+        background-color: var(--spectrum-gray-800);
```
nickschaap, line 52: Not sure this rule set is being used in the demo?

Westbrook (commented):
@nickschaap thanks for bringing this up, this looks like a nice and contained test case for the alias system that is suggested in Spectrum Tokens.

Some of these values, e.g. `--spectrum-gray-50` et al, feel like they fall into the pending breaking change from Spectrum CSS. @pfulton can you confirm? If so, would it be better to pair this change with that update?

Similarly, in cases where you've moved to raw `px` values, would the Spectrum direction point to those being things that should be orchestrated into an application specific set of tokens, for normalization across an application? If so, we should use this change to start roughing out a couple of places where that is true, to better align out usage to the usage we should be preaching to others.

nickschaap:
@Westbrook You're right I think it makes sense to hold on this change until we pull the latest from Spectrum CSS which will include the new color usage pattern.

I switched to raw px values for instances where we were previously using `--spectrum-global-dimension-size-*` variables. I believe it was brought up once before that Spectrum was removing these tokens in favor of having client applications define their own values (cc @GarthDB)

Westbrook:
I agree with the move away from `--spectrum-global-dimension-size-*`, and maybe the way you've done it is correct to the expectations of the ecosystem. However, my question is: does token consumption benefit from an app local level of tokens so that there is a system that specifically reflects the needs of the application (in this case the docs site) when those needs are intentionally left out of the Spectrum Token system? E.G. should be suggesting that apps locally create a system of `--${app}-alias-*` style tokens so that end of devs never write raw values into the actual UI defintions?

pfulton:
> Some of these values, e.g. --spectrum-gray-50 et al, feel like they fall into the pending breaking change from Spectrum CSS. @pfulton can you confirm? If so, would it be better to pair this change with that update?

_Yes_ - but let's briefly pause on this decision. We're going to take a quick look to see if there's a way for us to work with and preserve the current global colors without requiring those to be wrapped in the `rgba()` function. More info to come.

nickschaap:
@pfulton Would love to see these preserved!

nickschaap:
@pfulton Are there any updates here regarding color tokens?

pfulton:
@nickschaap Apologies for not responding sooner to this. I missed the notification and then I was on PTO.

The answer is: YES! We've managed to preserve the tokens, so downstream folks should not need to make any adjustments. We're putting a few final touches on the PR now, and hope to release next week. We'll make an announcement when it goes out.

Westbrook:
It looks like these color align with the future plans for Spectrum (@pfulton correct me if I'm wrong about this weeks announcement). @nickschaap if you could rebase this branch to `main`, I think it'll be ready to merge now.

pfulton:
Y'all should be good to go.

packages/overlay/overlay-trigger.md
Westbrook: Removed!

packages/badge/README.md
```diff
@@ -105,41 +105,48 @@ The `<sp-badge>` can be customized with either semantic or non-semantic variants
 Non-semantic badge colors are no longer supported directly by Spectrum and Spectrum Web Components. You can mimic their delivery via the following CSS Custom Properties. These values for the `variant` attribute/property have not been marked as deprecated, but will no longer achieve the results you may have relied on in the past.
 
 ```html demo
-<div style="display: flex; gap: var(--spectrum-global-dimension-size-50);">
+<div style="display: flex; gap: var(--spectrum-spacing-75);">
     <sp-badge
         variant="seafoam"
-        style="--mod-badge-background-color-default: var(--spectrum-global-color-static-seafoam-600)"
+        static="black"
```
Westbrook, line 111: Set these all to `static="black"` as the `-background-color-default` versions of all these colors are quite light.

Westbrook (approved):
@nickschaap thanks for kicking off this conversation! Looking forward to having you and your team continue to push in this are so that we all can make more _commonly_ themable UIs. Tweaked a couple of small things at rebase time. Otherwise, great work. 🙇🏼

PR #2983, opened 2023-03-05
fix(@swc-react/overlay): cem file glob cause one customer element info were omitted

jianliao:
## Description
Fixed two issues in this PR.
1. React wrapper generation should exclude the new project - `custom-vars-viewer`
2. [overlay](https://github.com/adobe/spectrum-web-components/tree/main/packages/overlay) component is special.
Two customer element declaration file names do not follow the naming convention. [overlay-trigger.ts](https://github.com/adobe/spectrum-web-components/blob/main/packages/overlay/overlay-trigger.ts) and [active-overlay.ts](https://github.com/adobe/spectrum-web-components/blob/main/packages/overlay/active-overlay.ts). They should both have prefix `sp-`. This will cause the generated customer-elements.json miss two customer element declaration information and the react wrapper miss two file import statements.

## Related issue(s)

-

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/cem-react-wrapper.config.js
+++ b/cem-react-wrapper.config.js
@@ -16,12 +16,7 @@ import { resolve } from 'path';
 import yaml from 'js-yaml';
 
 export default {
-    globs: [
-        '**/sp-*.ts',
-        '**/active-overlay.ts',
-        '**/overlay-trigger.ts',
-        '**/src/[A-Z]*.ts',
-    ],
+    globs: ['**/sp-*.ts', '**/overlay-trigger.ts', '**/src/[A-Z]*.ts'],
     exclude: [
         '**/sp-icon-*.ts',
         '**/*.d.ts',
--- a/cem-react-wrapper.config.js
+++ b/cem-react-wrapper.config.js
@@ -16,7 +16,12 @@ import { resolve } from 'path';
 import yaml from 'js-yaml';
 
 export default {
-    globs: ['**/sp-*.ts', '**/src/[A-Z]*.ts'],
+    globs: [
+        '**/sp-*.ts',
+        '**/active-overlay.ts',
+        '**/overlay-trigger.ts',
+        '**/src/[A-Z]*.ts',
+    ],
     exclude: [
         '**/sp-icon-*.ts',
         '**/*.d.ts',
--- a/scripts/cem-plugin-react-wrapper.js
+++ b/scripts/cem-plugin-react-wrapper.js
@@ -176,9 +176,9 @@ export default function genReactWrapper({
                 .filter(
                     (m) =>
                         m.exports.length === 1 &&
-                        m.exports.filter(
+                        m.exports.some(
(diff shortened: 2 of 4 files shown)


jianliao:
I think a better solution shall be rename [overlay-trigger.ts](https://github.com/adobe/spectrum-web-components/blob/main/packages/overlay/overlay-trigger.ts) and [active-overlay.ts](https://github.com/adobe/spectrum-web-components/blob/main/packages/overlay/active-overlay.ts) to `sp-overlay-trigger.ts` and `sp-active-overlay.ts`. Another component refactor is [tooltip](https://github.com/adobe/spectrum-web-components/tree/main/packages/tooltip), customer element declaration [TooltipProxy](https://github.com/adobe/spectrum-web-components/blob/2123c57f15d3239e3c8fb348309658c33e010a4d/packages/tooltip/src/Tooltip.ts#L38) should be move to a file named `sp-tooltip-proxy.ts` in the component root folder. 

But all these are breaking changes.

cem-react-wrapper.config.js
```diff
@@ -16,7 +16,12 @@ import { resolve } from 'path';
 import yaml from 'js-yaml';
 
 export default {
-    globs: ['**/sp-*.ts', '**/src/[A-Z]*.ts'],
+    globs: [
+        '**/sp-*.ts',
+        '**/active-overlay.ts',
```
Westbrook, line 21: `<active-overlay>` isn't an element I would expect a consumer to ever leverage directly. I'm not sure we should include it here. Have you run into a specific case where you wanted to use it?

scripts/cem-plugin-react-wrapper.js
```diff
@@ -175,8 +175,10 @@ export default function genReactWrapper({
             const fileImports = modules
                 .filter(
                     (m) =>
-                        m.declarations.length === 0 &&
-                        m.path.indexOf('sync/') === -1
+                        m.exports.length === 1 &&
```
Westbrook, line 178: If we used [some](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) would we be able to simplify to a single condition?

Westbrook (approved):
LGTM! I'll try to work out a release on these by EOD tomorrow.

scripts/cem-plugin-react-wrapper.js
```diff
@@ -175,8 +175,10 @@ export default function genReactWrapper({
...
+                        m.exports.some(
```
jianliao, line 179: I changed this line to Array.some(...), but I have to keep L178 as it is now because Tooltip component has a special customer element declaration - TooltipProxy in `src/Tooltip.ts`.

Issue #2961, opened 2023-02-26
[Bug]: sp-tabs-overflow next and previous buttons position is not align then sp-tabs is in compact mode


monkeyjabs:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-tabs-overflow

### Expected behavior

The next/previous scroll button should auto align when sp-tabs is in compact mode

### Actual behavior

The next/previous buttons are lower than the tab label

### Screenshots

![PixelSnap 2023-02-26 at 10 48 26](https://user-images.githubusercontent.com/4380525/221430468-a0b8c5dc-ad4a-4070-a6a2-8965223fe66f.png)

### How can we reproduce this issue?

1. Go to: https://studio.webcomponents.dev/edit/OmrIqazVRHvZAJAvJn4D/src/index.ts?p=stories
2. Observe the Next/Previous button

Westbrook:
closed by #3012

PR #3012, opened 2023-03-14
fix(tabs): compact and normal sp-tab-overflow styling

nickschaap:
## Description

This changes add a compact attribute to `sp-tabs-overflow`. It also updates spectrum config to pull custom properties from Spectrum CSS to ensure alignment in how sp-tabs-overflow buttons height is being calculated.

## Related issue(s)

- #2961

## Motivation and context

To support compact delivery of `sp-tabs-overflow`

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices


packages/tabs/src/spectrum-config.v2.js
```diff
@@ -248,6 +248,29 @@ const config = {
...
+                    'size'
+                ),
+            ],
+            requireComponentPresence: [
+                {
+                    type: 'class',
+                    name: 'regex',
+                    regex: /spectrum-Tabs--size/,
+                },
+            ],
```
nickschaap, line 272: Is it possible to pull just the few custom properties we are interested in?

Westbrook: Not at current, as the tooling here is meant to have a one for one conversion between Spectrum CSS and SWC.

If the goal is to reduce duplication, you could do similar by making a `tabs-sizes` CSS file that is shared between the `overflow` and `tabs` elements: essentially, what you've done already, but take these values out of the conversions for Tabs.

packages/tabs/src/tabs-overflow.css
```diff
@@ -10,6 +10,8 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
+@import './spectrum-tabs-overflow.css';
```
Westbrook, line 13: If you end up sharing, you'll want to import this in the JS instead of the CSS so that it doesn't get compiled into two different files at delivery time.

packages/tabs/src/spectrum-config.v2.js
nickschaap: Good idea I think I'll go with that

nickschaap:
@Westbrook Do you think its worth binding the `size` and `compact` attributes of `sp-tabs-overflow` to the underlying `sp-tabs` element or should we expect consumers to keep these in sync manually?

Westbrook:
We may have jammed you up a bit here with our late stage file rename, but hopefully a rebase to `main` for this goes pretty easily in the terminal. If not, happy to support getting this re-caught up and merged, so it can be included in today( or tomorrow)'s release.

nickschaap:
@Westbrook I'm not sure why some Color Wheel VRTs are failing here

Westbrook:
Looks like you accepted your golden image cache as opposed to the upstream cache, where that file was changed.

Westbrook (approved):
Thanks for working through this @nickschaap! I did a reorg of the stories over the top to ensure that `compact` was included in our VRTs and that more of the settings for the Overflow component where "configurable" there. We're racing towards a release, and hope to have this out to you, soon!

packages/tabs/stories/tabs-overflow-panel-sizes.stories.ts
```diff
@@ -0,0 +1,47 @@
...
+m.args = {
+    size: 'm',
+    includeTabPanel: true,
+};
+export const l = (args: OverflowProperties): TemplateResult => {
+    return renderTabsOverflowExample(args);
+};
+l.args = {
+    size: 'l',
+    pincludeTabPanelanel: true,
```
nickschaap, line 39: @Westbrook typo

PR #3015, opened 2023-03-14
fix: prevent default hoisting of custom pseudo elements

Westbrook:
- revisit default hoisting
- realize that selector is no longer functional
- remove non-functional selectors

### DIFF
--- a/tasks/process-spectrum.js
+++ b/tasks/process-spectrum.js
@@ -206,14 +206,6 @@ async function processComponent(componentPath) {
                         },
                     });
                     matched = true;
-                } else if (isHoistedPseudoClass(component)) {
-                    match.push({
-                        hoist: true,
-                        find: { ...component },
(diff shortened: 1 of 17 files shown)


tasks/process-spectrum.js
```diff
@@ -214,21 +207,13 @@ async function processComponent(componentPath) {
...
-                    });
-                    matched = true;
-                    log = true;
+                    // match.push({
+                    //     hoist: true,
+                    //     find: { ...component },
+                    //     replace: { ...component },
+                    // });
+                    // matched = true;
+                    // log = true;
```
hunterloftis, line 216:
```suggestion
```

hunterloftis (approved):

tasks/process-spectrum.js
Westbrook: 🙈 🔪

PR #3023, opened 2023-03-15
feat(table): support quiet style

jianliao:
## Description

Spectrum CSS Table component supports the `quiet` [mode](https://opensource.adobe.com/spectrum-css/table.html#divsquiet). SWC should unleash it, too.
image

## Related issue(s)

-

## Motivation and context

All the Spectrum tables on Spectrum documentation website are in quiet mode. If I want to replace them with SWC table component, it should support quiet table mode. 

For example:
1. Example 1
image
2. Example 2
image

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

image
image
image

## Types of changes

## Checklist

## Best practices


jianliao:
Hi @najikahalsema,
Here is what I've done so far.
1. Change the [spectrum-config.js](https://github.com/adobe/spectrum-web-components/pull/3023/files#diff-2670c93c7cef0cde1f407ab2752127eaecc371c58105f074af99af62e072211b) to add quiet attribute.
2. Run `yarn process-spectrum` at the repo root folder to get the `spectrum-table.css` updated.
3. Add quiet property in [Table.ts](https://github.com/adobe/spectrum-web-components/blob/1a2e31ca90bae8b301ceabb102692261e94f79ed/packages/table/src/Table.ts) for typescript type definition.
4. Test the `quiet` mode by adding a new storybook [file](https://github.com/adobe/spectrum-web-components/blob/1a2e31ca90bae8b301ceabb102692261e94f79ed/packages/table/stories/table-elements-quiet.stories.ts).

Two issues:
1. The new added storybook page does not show up in storybook. See [Storybook](https://support-quiet-table--spectrum-web-components.netlify.app/storybook/).
2. The quiet style does not applied when I enable it through the Chrome console programmatically. I compared my change with other components that support `quiet` like ActionButton etc. I couldn't find much difference except the table html markup is more complex. I also reviewed the generated `spectrum-table.css`, the change introduced for `quiet` make sense to me. I don't know why it doesn't work.

packages/table/src/Table.ts
```diff
@@ -157,6 +157,12 @@ export class Table extends SizedMixin(SpectrumElement, {
     @property({ type: Boolean, reflect: true })
     public scroller = false;
 
+    /**
+     * Uses quiet styles or not
+     */
+    @property({ type: Boolean, reflect: true, attribute: 'quiet' })
```
najikahalsema, line 163:
```suggestion
    @property({ type: Boolean, reflect: true })
```

packages/table/src/spectrum-table.css
```diff
@@ -1199,7 +1199,7 @@ governing permissions and limitations under the License.
     ); /* [dir=ltr] .spectrum-Table--compact .spectrum-Table:not(.spectrum-Table--quiet) tbody.spectrum-Table-body .spectrum-Table-row:first-child .spectrum-Table-cell:last-child */
 }
 :host([dir='rtl'][density='compact'])
-    .spectrum-Table:not(.spectrum-Table--quiet)
+    .spectrum-Table:not([quiet])
```
najikahalsema, line 1202: This being applied as `.spectrum-Table:not([quiet])` instead of being added as an attribute to the host element is certainly interesting. 

@Westbrook do you think this would require some kind of complex selector syntax in the config to handle this? This looks odd to me. 🤔

najikahalsema (commented):
There are a couple things I noticed and left comments on, but when I pulled your branch, I wasn't seeing the `quiet` styles being applied. Do you see them on your end? 
I've been looking at the `spectrum-config` and the generated css output, and I'm wondering if there aren't enough complex selectors that would allow the CSS to be applied properly. That's just a hypothesis. @Westbrook, does anything here stand out to you?

packages/table/src/spectrum-table.css
jianliao: Here is the CSS quiet table:
image
Not sure why this style is missing in SWC quiet table:
```
.spectrum-Table--quiet .spectrum-Table-body {
    border-width: 1px 0;
    background-color: var(--spectrum-table-m-regular-quiet-cell-background-color, var(--spectrum-alias-background-color-transparent));
}
```

packages/table/src/spectrum-table.css
```diff
@@ -2038,135 +2030,127 @@ governing permissions and limitations under the License.
         var(--spectrum-alias-border-color-mid)
     ); /* [dir=rtl] .spectrum-Table-cell--divider */
 }
-.spectrum-Table--quiet .spectrum-Table-body {
+:host([quiet]) .spectrum-Table-body {
```
jianliao, line 2033: Why this style would not apply in runtime?

packages/table/src/spectrum-config.js
jianliao, line 1: Hi @najikahalsema,
I redo it based on the new spectrum-config format. But after `yarn run process-spectrum`, no css files got changed. Did I miss anything?

najikahalsema: Hi Jian, thanks for pinging me on this. 

The way the CSS is written for Table isn't very compatible with our API, so it requires a couple "find/replace" selectors to get it going--around 16 of them, I think. It's not a quick fix like we originally thought, but I'll explore this path on Monday and see if I can get an example to get you started. We can sync up on Monday afternoon, and you can decide based on the work if this is a path you want to pursue. The other option is to wait for core tokens, I believe, which will take longer to get through, but will be a more straightforward path for you.

jianliao:
Hi Najika,
After comparing the Table component with the Table component in the Spectrum design page, I noticed that there are quite a few differences beyond just the quiet mode. For example, the table header capitalization, font weight, and style all vary. Given these discrepancies, I've decided to wait until the Table component is updated to use the latest core tokens. Thank you for your assistance on this matter, and I appreciate your willingness to explore a solution for me.

PR #3029, opened 2023-03-16
refactor(quick-actions): process CSS with lightning css

Westbrook:
Create a V2 config for Quick Actions CSS processing.

The main diff below is simplifying `slot[name='action']` to `[name='action']`.


najikahalsema (approved):
nice!

PR #3041, opened 2023-03-17
fix: sets action-menu quiet to false by default, fixes #3040

### LINKED ISSUE #3040: [Bug]: sp-action-menu forces quiet mode
### Code of conduct

### Impacted component(s)

sp-action-menu

### Expected behavior

`quiet` defaults to false

### Actual behavior

`quiet` is forced to true

### Screenshots

_No response_

1. Go to ['...'](https://opensource.adobe.com/spectrum-web-components/components/action-menu/) and notice that even though quiet is not set in the examples it is defaulting to true

_No response_

_No response_

benjamind:
## Description

* Modified `PickerBase` to no longer force divergent behavior on `quiet` attribute in `sizePopover`, moved this logic down into `Picker`.
* Modified `ActionMenu` to reflect its `quiet` property down onto the internal action-button, and no longer force its value to true.

NOTE: This is a breaking change, as the default behavior for `sp-action-menu` will now be noisy rathern than quiet.

## Related issue(s)

Fixes #3040 

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ executors:
 parameters:
     current_golden_images_hash:
         type: string
-        default: 2dbdd52d58095177457928aa96a394428c980acd
+        default: 46c860cf036161f43b5726702bd831fce66cbf74
 commands:
     downstream:
         steps:
--- a/packages/action-menu/src/ActionMenu.ts
+++ b/packages/action-menu/src/ActionMenu.ts
@@ -62,7 +62,7 @@ export class ActionMenu extends ObserveSlotText(PickerBase, 'label') {
     protected override render(): TemplateResult {
         return html`
             <sp-action-button
-                quiet
+                ?quiet=${this.quiet}
                 ?selected=${this.open}
                 aria-haspopup="true"
                 aria-controls="popover"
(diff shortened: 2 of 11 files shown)


Westbrook (commented):
This looks great. Can you do the quick favor of editing the commit from `fix:` to `feat:` to it amounts to the _breaking_ change it is meant to be when publication time comes next week?

Westbrook:
Look CI is also telling us it's a breaking change. It's so nice when computers actually works in ways that prevent humans from making mistakes. Would be nice if you were to drop in some a new `quiet` demo or two for Storybook/CI, but we can always follow up on that later. We will need the VRTs bumped, as outlined in https://github.com/adobe/spectrum-web-components/tree/lightningcss#keeping-ci-assets-updated

Westbrook (approved):
Thanks for getting into all the documentation and story changes for this! Cleaned up the one test that no longer applied and cycled the VRTs based on this change. Going out later today, or so 🤞🏼

Issue #2359, opened 2022-06-14
[Bug]: Memory Leak when we remove sp-tooltip(self-managed) from dom but not its parent


maheshwari-ashutosh:
### Code of conduct

- [x] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-tooltip

### Expected behavior

When client conditionally render the `sp-tooltip`, when its removed from DOM it should remove the event listener from the parent. so that when we hover on the parent the active-overlay should not be added on the DOM.
See : [https://studio.webcomponents.dev/edit/dcmsShnRpGqlmmaBvPWy/src/index.ts?p=stories](https://studio.webcomponents.dev/edit/dcmsShnRpGqlmmaBvPWy/src/index.ts?p=stories)

**When the second time user hovers for 1.5secs on the card, it should look like this**
<img width="306" alt="Screenshot 2022-06-14 at 9 51 23 PM" src="https://user-images.githubusercontent.com/18366310/173627446-b30e6b65-7aea-4234-8611-98a487b9d032.png">

### Actual behavior

**When the second time user hovers for 1.5secs on the card, it looks like this**
<img width="314" alt="Screenshot 2022-06-14 at 9 49 13 PM" src="https://user-images.githubusercontent.com/18366310/173627814-1ebe9f27-c55d-4737-8765-621d1eed1ab7.png">

### What browsers are you seeing the problem in?

Chrome, Safari

### How can we reproduce this issue?

3. Scroll to '....'
4. Check console
5. See error

### Sample code that illustrates the problem

Please check the link in expected behaviour section

Westbrook:
Does it work to leverage the tooltip like this? https://studio.webcomponents.dev/edit/zig2Krhi5Ovi6vSrUMcz/src/index.ts?p=stories

maheshwari-ashutosh:
It works functionality wise but lets say we have many cards in a single page, then it will always be in the DOM for each and every card element even when its not visible thus increasing the DOM size.

Westbrook:
I'd love to see the stats at which point you actually see a perf drop off here. Very interesting area, but not something that benefits from optimizations that aren't supported by measurement.

[...]

PR #3046, opened 2023-03-21
feat(tooltip): fixing the crash if tooltip parent is already removed from dom

anandpal-adobe:
## Related issue(s)

https://github.com/adobe/spectrum-web-components/issues/2359

-

## Motivation and context

This is needed if the parent view of tooltip is already removed, there was no check present which was causing the crash

## How has this been tested?

The issue was tested in venus app where crash was happening 

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/tooltip/src/Tooltip.ts
+++ b/packages/tooltip/src/Tooltip.ts
@@ -242,23 +242,27 @@ export class Tooltip extends SpectrumElement {
                 this.previousSlot = this.slot;
             }
             this.slot = 'self-managed-tooltip';
-            parentElement.addEventListener('pointerenter', this.openOverlay);
-            parentElement.addEventListener('focusin', this.openOverlay);
-            parentElement.addEventListener('pointerleave', this.closeOverlay);
-            parentElement.addEventListener('focusout', this.closeOverlay);
+            if (parentElement) {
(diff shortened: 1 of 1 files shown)


Westbrook (changes requested):
Can you help us with a test case here? There's no clear situation where this should be needed, and may just be preventing functionality if there were.

Westbrook:
We'd also prefer that you use a branch over a fork as outlined for you in Slack.

ananpal:
@Westbrook the crash is happening in venus, with some button elements with elements; which are being created and removed just after that; as manageTooltip is being called asynchronously by the time manageTooltip run, parent element is already deleted. I am not sure how to add testcase for this.
I have added the check before adding or removing event from parent element only if parentElement is present. I dont see how it will prevent any functionality, rather it will just prevent application from crashing.
I am waiting for access request to be approved then I will create a new pr from branch.

Westbrook:
Closing for lack of a test case. If you're able to supply one at a later date, we'd be happy to revisit this PR with you.

PR #3061, opened 2023-03-27
chore(checkbox): declare change event in customer-elements.json

jianliao:
## Description

It looks like the customer-elements.json for `Checkbox` component hasn't mentioned  its `change` event.
 
## Related issue(s)

N/A

## Motivation and context

There is no `change` event declaration in customer-elements.json which cause @swc-react/checkbox component is unable to support `change` event callback.

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

image

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/checkbox/src/Checkbox.ts
+++ b/packages/checkbox/src/Checkbox.ts
@@ -89,6 +89,7 @@ const dashIcon = {
 /**
  * @element sp-checkbox
  * @slot - content to display as the label for the Checkbox
+ * @fires change - Announces a change in the `checked` property of a Checkbox
  */
 export class Checkbox extends SizedMixin(CheckboxBase) {
     @property({ type: Boolean, reflect: true })
--- a/packages/split-view/src/SplitView.ts
+++ b/packages/split-view/src/SplitView.ts
@@ -47,6 +47,7 @@ const COLLAPSE_THREASHOLD = 50;
  * @element sp-split-view
  *
  * @slot Two sibling elements to be sized by the element attritubes
+ * @fires change - Announces the new position of the splitter
  */
 export class SplitView extends SpectrumElement {
     public static override get styles(): CSSResultArray {


Westbrook (dismissed):
Looks good to me, but could you help me understand the underlying situation a bit better here?

Specifically, `change` is a native event, does React really not allow users to bing `onChange` to any element? Does it have something to do with `<SpCheckbox>` not actually being the `<sp-checkbox>` element so it gets confused from an event bubbling perspective?  Or maybe because we dispatch a Custom Event _called_ `change`?

As we develop a better root understanding of the relationship between our HTML elements and the React components that wrap them we can do a better job of developing, reviewing, or tooling the code to ensure things like this are included system wide and not just each time we run into something that you or another client specifically needs in the moment.

jianliao:
According to the React doc([old](https://legacy.reactjs.org/docs/events.html#form-events) and [new](https://react.dev/reference/react-dom/components/input#input)), `onChange` synthetic event is only supported on HTML form field native element like `input`, `select` and `textarea`. But `sp-checkbox` is a customer element that wrapped an input element.

jianliao:
> As we develop a better root understanding of the relationship between our HTML elements and the React components that wrap them we can do a better job of developing, reviewing, or tooling the code to ensure things like this are included system wide and not just each time we run into something that you or another client specifically needs in the moment.

I agree. All the SWC components that wrapped a form element should all be made sure they declare the re-dispatch native events.

jianliao:
Hi @Westbrook, I'm afraid I'm confusing the typescript type issue between swc react wrapper and my own nextjs wrapper.  React Synthetic event should be able to apply to all the `HTMLElment` type react components. Closed the issue.

jianliao:
Hi @Westbrook,
I found this [comment](https://github.com/facebook/react/issues/19846#issuecomment-827550050) from [React#19846](https://github.com/facebook/react/issues/19846), React would not dispatch `onChange` synthetic event for `change` native event to none native browser component. The following react code would compile, but onChange call back handler would never be called.

```
const handleChange = (e: FormEvent) => {
  console.log((e.currentTarget as CheckboxType).checked);
};
return (
    <Theme theme="spectrum" scale="medium" color="light">
      <div className="App">
        <Checkbox emphasized onChange={handleChange}>It is a SWC checkbox</Checkbox>
      </div>
    </Theme>
  )
```
For react wrapper code generation to work, we have to explicit declare it as an event.

I also audited the whole code base and found another missing declaration in [SplitView] (https://github.com/adobe/spectrum-web-components/blob/d4e49f581d41720d92e9f786cda4af56dbafc8c5/packages/split-view/src/SplitView.ts#L403-L409) component. Let me know if you want me to add the event declaration.

Westbrook (approved):
Got it! LGTM. 🙇🏼 

Next release may not be till next week, but you'll see it in the channel when it comes.

PR #3062, opened 2023-03-27
perf: reduce render cycles when managing "dir" attribute

Westbrook:
## Description
Review the binding relationship between `dir` and `SpectrumElement`. @chrisbank pointed out that we were getting a detach time render call based on the previous relationship and this removes that, unless reactivity to `dir` is required for some part of the elements functionality (see Tabs, Tabs Overflow, and Top Nav). Instead, this relies on the native reflection of the `dir` property to ensure that the JS and HTML values are kept in sync rather than the reflection provided by Lit.

Local testing shows this gives us a little performance bump at connection time, too, but we'll know more when the full perf test cycle runs below.

## How has this been tested?

-   [ ] _Test case 1_
    1. CI runs
    2. Unit tests pass
    3. VRTs pass
-   [ ] _Test case 2_
    1. Go [here](https://content-dir--spectrum-web-components.netlify.app/storybook/?path=/story/tabs--autofocus)
    2. Tinker with content direction and various components
    3. Focus on Tabs, Tabs Overlay, and Top Nav, but include a couple others for good measure.

## Types of changes
-   [x] Performance refactor

## Checklist


pinkhominid (approved):
LGTM! Thanks for picking this up.

hunterloftis (approved):

Issue #3076, opened 2023-04-03
Move Picker's state icon in a method, so that it could be overridden from outside


razvancir96:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Description of issue

We need to add a loading spinner inside the picker's button, which is not supported right now. There is an invalid icon supported and we can extract that logic in a separate method, which can be overridden from outside. This way, the component's API and DOM structure isn't changed, but the consumers gain more flexibility.

What we have now (in render):
```
${this.invalid
        ? html`
              <sp-icon-alert
                  class="validation-icon"
              ></sp-icon-alert>
          `
        : nothing}
```

The more flexible version:
```
protected renderStateIcon(): TemplateResult | typeof nothing {
    if (!this.invalid) {
        return nothing;
    }
    return html`
        <sp-icon-alert class="validation-icon"></sp-icon-alert>
    `;
}
```
In render:
`${this.renderStateIcon()}`

Westbrook:
Loading UI has recently been added to the [CSS implementation](https://opensource.adobe.com/spectrum-css/picker.html#:~:text=Select%20a%20Country-,Closed%20and%20Loading,-Loading...). It would be preferrable that we add support for this directly to the element. Thoughts on how you would approach that here?

razvancir96:
> Loading UI has recently been added to the [CSS implementation](https://opensource.adobe.com/spectrum-css/picker.html#:~:text=Select%20a%20Country-,Closed%20and%20Loading,-Loading...). It would be preferrable that we add support for this directly to the element. Thoughts on how you would approach that here?

Thanks for pointing it out! Sounds good to implement it directly in this repo. There are 2 things that I want to point out:
- [Spectrum's Picker page](https://spectrum.adobe.com/page/picker/) has no specifications about the loading state (also checked the XD file)
- The CSS implementation, for the `is-loading` class, it only applies a [disabled style to the arrow down icon](https://github.com/adobe/spectrum-css/blob/main/components/picker/index.css#L344). But with the CSS approach, the components can be easily composed.

Those being said, we can definitely add a new property to the picker, called `loading`, which decides to show/hide the loading spinner. But I'm not sure if we should do some other things when setting the property, or let the user decide if the picker is disabled or other things. Thoughts on this?

Westbrook:
That's a great question!

[...]

PR #3093, opened 2023-04-06
feat(picker): add loading state to the picker

razvancir96:
## Description

We need to add a loading spinner inside the picker's button.

## Related issue(s)

- https://github.com/adobe/spectrum-web-components/issues/3076

## Motivation and context

In one of Adobe's projects, we need to support the loading spinner inside the picker's button.

## How has this been tested?

-   [x] _Test case 1_
    1. Run `yarn storybook`
    2. Go to Picker -> Sizes stories: https://localhost:8001/?path=/story/picker-sizes--s 
    3. Change the sizes and from the controls, set the `loading` property on true. Also, change the scale.

Size S:
Screenshot 2023-04-06 at 18 11 22

Size M:
Screenshot 2023-04-06 at 18 11 37

Size L:
Screenshot 2023-04-06 at 18 11 50

Size XL:
Screenshot 2023-04-06 at 18 12 01

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/picker/README.md
+++ b/packages/picker/README.md
@@ -376,12 +376,50 @@ When the `value` of an `<sp-picker>` matches the `value` attribute or the trimme
 </sp-picker>
 <br />
 <br />
-<sp-field-label for="picker-disabled">Quiet:</sp-field-label>
+<sp-field-label for="picker-disabled-quiet">Quiet:</sp-field-label>
 <sp-picker
     label="Select a Country with a very long label, too long in fact"
     disabled
     quiet
     id="picker-disabled-quiet"
+>
+    <sp-menu-item>Deselect</sp-menu-item>
+    <sp-menu-item>Select inverse</sp-menu-item>
+    <sp-menu-item>Feather...</sp-menu-item>
+    <sp-menu-item>Select and mask...</sp-menu-item>
+    <sp-menu-divider></sp-menu-divider>
+    <sp-menu-item>Save selection</sp-menu-item>
+    <sp-menu-item disabled>Make work path</sp-menu-item>
+</sp-picker>
+```
+
+### Loading
+
+When given a `loading` attribute, an `<sp-picker>` will be delivered with an `<sp-progress-circle>` to visually outline that it is `loading` and it will not toggle open or display its `<sp-menu-item>` descendants until the attribute is removed.
+
+```html
+<sp-field-label for="picker-loading">Standard:</sp-field-label>
+<sp-picker
+    label="Select a Country with a very long label, too long in fact"
+    loading
+    id="picker-loading"
+>
+    <sp-menu-item>Deselect</sp-menu-item>
+    <sp-menu-item>Select inverse</sp-menu-item>
+    <sp-menu-item>Feather...</sp-menu-item>
+    <sp-menu-item>Select and mask...</sp-menu-item>
+    <sp-menu-divider></sp-menu-divider>
+    <sp-menu-item>Save selection</sp-menu-item>
+    <sp-menu-item disabled>Make work path</sp-menu-item>
+</sp-picker>
+<br />
+<br />
+<sp-field-label for="picker-loading-quiet">Quiet:</sp-field-label>
+<sp-picker
+    label="Select a Country with a very long label, too long in fact"
+    loading
+    quiet
+    id="picker-loading-quiet"
 >
     <sp-menu-item>Deselect</sp-menu-item>
     <sp-menu-item>Select inverse</sp-menu-item>
--- a/packages/picker/src/Picker.ts
+++ b/packages/picker/src/Picker.ts
@@ -43,6 +43,8 @@ import type {
 } from '@spectrum-web-components/menu';
 import '@spectrum-web-components/tray/sp-tray.js';
 import '@spectrum-web-components/popover/sp-popover.js';
+import '@spectrum-web-components/progress-circle/sp-progress-circle.js';
+
 import type { Popover } from '@spectrum-web-components/popover';
 import {
     openOverlay,
@@ -104,6 +106,9 @@ export class PickerBase extends SizedMixin(Focusable) {
 
     @property({ type: Boolean, reflect: true })
     public invalid = false;
+
+    @property({ type: Boolean, reflect: true })
+    public loading = false;
 
     @property()
     public label?: string;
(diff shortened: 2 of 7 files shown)


packages/picker/README.md
```diff
@@ -376,7 +376,7 @@ When the `value` of an `<sp-picker>` matches the `value` attribute or the trimme
 </sp-picker>
 <br />
 <br />
-<sp-field-label for="picker-disabled">Quiet:</sp-field-label>
+<sp-field-label for="picker-disabled-quiet">Quiet:</sp-field-label>
```
razvancir96, line 379: This was a previous bug that I've identified 😄

packages/picker/src/Picker.ts
```diff
@@ -406,6 +411,16 @@ export class PickerBase extends SizedMixin(Focusable) {
                           ></sp-icon-alert>
                       `
                     : nothing}
+                ${this.loading
+                    ? html`
+                          <sp-progress-circle
+                              label="Loading picker"
```
razvancir96, line 417: How do you localize stuff in this project? I to search some examples, but found some unlocalized strings 😅 
Screenshot 2023-04-06 at 18 16 19

packages/picker/src/picker.css
```diff
@@ -35,6 +35,41 @@ governing permissions and limitations under the License.
...
+        var(--spectrum-picker-spacing-text-to-alert-icon-inline-start)
+    );
+}
+
+:host([loading]) .sp-picker-loading-s {
+    --spectrum-progress-circle-size: var(--spectrum-workflow-icon-size-75);
+    --spectrum-progress-circle-thickness: var(
+        --spectrum-progress-circle-thickness-small
+    );
+}
```
razvancir96, line 50: Should I pick some other variables? Or make my own ones? More context: the picker has 4 dimensions: s/m/l/xl and the loading spinner has 3 dimensions: s/m/l, and do not match together. So I had to adapt its dimensions. Also, should we validate with design those dimensions?

packages/picker/test/index.ts
```diff
@@ -286,6 +286,13 @@ export function runPickerTests(): void {
             expect(el.invalid).to.be.true;
             await expect(el).to.be.accessible();
         });
+        it('renders loading accessibly', async () => {
+            el.loading = true;
+            await elementUpdated(el);
+
+            expect(el.loading).to.be.true;
+            await expect(el).to.be.accessible();
+        });
```
razvancir96, line 295: I copied the pattern from the invalid property, but I feel like some more tests should be added. Are there any guidelines about writing tests?

packages/picker/src/picker.css
Westbrook: @pfulton is there any token support for these variation?

Westbrook: From a selector standpoint, we can use `:host([size="s"]) sp-progress-circle` instead of needing to add a dynamic class to this element.

packages/picker/stories/picker.stories.ts
```diff
@@ -405,6 +416,29 @@ export const custom = (args: StoryArgs): TemplateResult => {
...
+            <sp-menu-item>Select Inverse</sp-menu-item>
+            <sp-menu-item>Feather...</sp-menu-item>
+            <sp-menu-item>Select and Mask...</sp-menu-item>
+            <sp-menu-divider></sp-menu-divider>
+            <sp-menu-item>Save Selection</sp-menu-item>
+            <sp-menu-item disabled>Make Work Path</sp-menu-item>
+        </sp-picker>
+    `;
+};
+
```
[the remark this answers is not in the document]
Westbrook: VRTs _should_ use `prefers-reduced-motion` which would alleviate any worries here, but you're still on a fork and not a branch so the full CI isn't running to show us these sorts of things by default.

packages/picker/test/index.ts
Westbrook: Two things we'll want to test for:
- is the listbox disabled when `loading`: load, apply loading, try to click, see that `open === false` and the Menu Items cannot be clicked with `sendMouse`
- is the final "loading" label applied to the accessibility tree: load, apply loading, get an accessibility snapshot, ensure there is a node with the expected label

packages/picker/src/Picker.ts
```diff
@@ -406,6 +411,16 @@ export class PickerBase extends SizedMixin(Focusable) {
...
+                              indeterminate
+                              class="sp-picker-loading sp-picker-loading-${this
```
Westbrook, line 419: This class could just be `loading`, or similar, as we're scoped to this shadow root.
```suggestion
                              class="loading sp-picker-loading-${this
```
The sizing can also be interpreted from the `:host` as I commented in the CSS, simplifying the render here.

packages/picker/README.md
```diff
@@ -393,6 +393,42 @@ When the `value` of an `<sp-picker>` matches the `value` attribute or the trimme
 </sp-picker>
 ```
 
+### Loading
```
Westbrook, line 396: We should outline here that when `[loading]` a Pickers listbox cannot be opened, maybe:

> When given a `loading` attribute, an `<sp-picker>` will be delivered with an `<sp-progress-circle>` to visually outline that it is "loading" and it will not toggle open or display its `<sp-menu-item>` descendants until the attribute is removed.

packages/picker/test/index.ts
razvancir96: Thanks! One additional question about testing: is there any command to run only the tests from one file? I tried some combinations, but failed 😅

Westbrook: `yarn test:watch:focus picker`

packages/picker/src/Picker.ts
razvancir96: Yes, I see that `File` is the default, but if the user doesn't set another value and relies on the default, the default is not localized 😅

If I remove the aria-label, the a11y test fails, with the following message:
```
Error: Accessibility Violations
      ---
      Rule: aria-progressbar-name
      Impact: serious
      ARIA progressbar nodes must have an accessible name (https://dequeuniversity.com/rules/axe/4.4/aria-progressbar-name?application=axeAPI)
      
      Issue target: #picker,sp-progress-circle
      Context: <sp-progress-circle indeterminate="" dir="ltr" size="m" role="progressbar"></sp-progress-circle>
      Fix any of the following:
        aria-label attribute does not exist or is empty
        aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty
        Element has no title attribute
      ---
```

So if we do not want to localize something, the only solution would we to use `aria-labelledby` and tie it with the button's content and hope that the user updates the button with a suggestive message when loading. WDYT?

PR #3102, opened 2023-04-11
feat(styles): bump tokens to latest release

castastrophe:
## Description

Bumping @spectrum-css/tokens to latest.
  - Adds the font-weight fix for --spectrum-black-font-weight which is not a CSS-safe value.
  - Includes rgb parsed token values with original fallbacks (no code changes required).

`scripts/spectrum-tokens.js`: Adding fix for `.spectrum--light, .spectrum--lightest`.

## Motivation and context

- Aligns with latest released tokens package, enables upgrade of newly migrated assets

## How has this been tested?

-   [x] _Check for unexpected token changes in diff_
-   [x] _Validate VRT results_
  -  I'm seeing subtle changes to a couple of icons, color change seems imperceptible (see Folder2ColorIcon)
  - HCM checkbox offset seems altered but tokens appear unchanged, CSS team is investigating
  - Avatar seems to have a subtle image diff on the disabled avatar; [imperceptible](https://4bcfa84fe27df28571516d5d8f1360d9--spectrum-web-components.netlify.app/review/#TagsStories/deletable.png)

## Types of changes

## Checklist

-   [N/A] I have reviewed at the Accessibility Practices for this feature, see: [Aria Practices](https://www.w3.org/TR/wai-aria-practices/)

## Best practices


tools/styles/tokens/express/custom-large-vars.css
```diff
@@ -10,10 +10,26 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-/* This file contains overrides and additions to core tokens */
+/*!
```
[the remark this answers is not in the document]
Westbrook: Thanks for pointing this out!

The `/*!` preventing existing tooling from removing that means that this is gonna get piled on a number of times across the project (a growing number as this decision makes its way across the CSS repo, the actually JS this specific file ends up in has three already just from this PR) and not wash out in the final production code. If Spectrum CSS feels it needs to ship this way, we'll likely build tooling to manually remove it so that our end consumers don't need to eat any size difference (regardless of how small the copyright is or what gzip might be able to do in any one file).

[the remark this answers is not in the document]
castastrophe: Pretty standard for copyrights to use the important flag but also ultra easy to strip all comments using existing tooling: https://www.npmjs.com/package/postcss-discard-comments#user-content-removeall

My point however was not regarding this change - the Spectrum Web Components was already adding the copyright comment via tooling, I'm pointing out that the logic seems not to be catching that the copyright was _already_ there in the supplied file because it's perhaps not designed to handle the important flag.  The copyright has always been in these files, the only change was flagging it important.

linters/stylelint-header/index.js
```diff
@@ -71,9 +71,11 @@ const rule = createPlugin(
             // Walk comments on root to find if header exists
             let found = false;
             root.walkComments((comment, _idx) => {
+                // Remove any asterisks and whitespace from the texts before comparing
+                const clean = (text) => text.replace(/(\*|\n|\s)/g, '').trim();
                 // If the two strings are at least 98% alike, it's a match
                 if (
-                    compareTwoStrings(comment.text, header) >=
+                    compareTwoStrings(clean(comment.text), clean(header)) >=
```
castastrophe, line 78: @Westbrook This fix to your stylelint-header tool allows the comparator to match the two strings. Just needed a little pre-cleansing to both strings so it could focus on diffs in the content rather than the formatting.  It didn't have anything to do with the ! important added by Spectrum CSS.

tools/styles/tokens/express/custom-large-vars.css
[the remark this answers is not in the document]
castastrophe: See above: https://github.com/adobe/spectrum-web-components/pull/3102/files#r1165738007.  This is now fixed.

pfulton (approved):
This looks like what I'd expect after our RGB fix went in.

scripts/spectrum-tokens.js
```diff
@@ -29,6 +29,7 @@ const tokensRoot = path.join(
     '*.css'
 );
 
+/** @todo Could generate this from CSS packages that have @spectrum-css/tokens as a dependency */
```
Westbrook, line 32: The goal will be to not need an allow list here at some point, but would certainly be an interesting way to manage this data in the interim. We have used it as a human reabable collection of packages that have made the Core Tokens conversion to date, but hopefully we'll be through this process before too much longer!

Westbrook (commented):
Not sure of any reason why this should throw errors in jut a few packages' unit test 🤔... Those Safari failures _might_ just be flakes, and the numebr of unreported failures in Firefox make me think there was something up on the infrastructure level.

Could you rebase onto `main` to see if that clears up?

Westbrook:
Ci still is unhappy. 😞

Taking a close look at the changes here, it seems that there are some unexpected updates in the `yarn.lock`. You mentioned having some network issues the other day, I wonder if some intersection of things in that area tripped up the lock management 🤔 Not a thing that's _supposed_ to happen, but 🤷🏼. Worse, it seems that it was a bumpy lock break and the intersecting packages are non-too happy with each other. I did a clean update to the effect packages in #3119, and they seem to be working, so the options for next steps seem to be:

- recheckout the lock from `main` and recreate it with just the changes that are specific to this PR
- wait for #3119 to land and then rebase to that.

castastrophe:
@Westbrook Rebased after your @web version bump and checked out the yarn.lock from main and re-ran yarn on a good wifi and now all I'm seeing is the token bump! Thanks for catching that. I ended up having to remove the playwright dependency in order to get the install to finish at the hospital.

.circleci/config.yml
```diff
@@ -10,7 +10,7 @@ executors:
 parameters:
     current_golden_images_hash:
         type: string
-        default: 10500edb351f4db403b24e1305a959910ef80e12
+        default: ad6a2c7402744826239beac9f89e022a588da9f0
```
Westbrook, line 13: The CI failures point to what I think is a more expected _lack_ of changs from `main`. E.G. when I parse changes in the `*medium-vars.css` files there do not seem to be any changes for "checkbox", but I 'd love you insight on what this update was _supposed_ to do.

The previously captured changes may have been part of the dirty `yarn.lock`? 😬

castastrophe, line 13:
```suggestion
        default: 10500edb351f4db403b24e1305a959910ef80e12
```

Westbrook (approved):
Thanks for working through the bumps and bruises here. LGTM!

Issue #1975, opened 2021-12-07
[RESEARCH] Internationalize small content pieces within elements.


Westbrook:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Description of issue

Smaller tools like https://github.com/shoelace-style/localize#readme allow for customization of small pieces of a component library.

Larger tools like https://github.com/lit/lit/tree/main/packages/localize might imply that we should take a position on internationalization for our consumers as well.

Find a solution to covering the few cases where we're shipping content within our components that needs to be translated without surfacing superfluous properties for this action.

jnurthen:
Any update on thoughts for this? It seems like one of the major obstacles to supporting accessibility well in languages other than English.

jnurthen:
The following components would currently be broken when translated (probably not exhausative):
* Asset (default hardcoded but can be overridden)
* SplitButton (aria-label of split hardcoded to More)
* Dialog (hardcoded close button label)
* NumberField (hardcoded increase/decrease buttons)
* Picker (hardcoded visually hidden dismiss button)
* Search (hardcoded Reset button)
* Tag (hardcoded Remove button)
* Toast (hardcoded Information/Error/Success Icon & Close button)

majornista:
Color Slider currently has the string "hue", which can be overwritten using the `label` prop: https://github.com/adobe/spectrum-web-components/blob/a447ba2714483619a17ab73e11efc79f9f815c84/packages/color-slider/src/ColorSlider.ts#L57

Color Area currently uses "saturation" and "luminosity", which can be overwritten using the `labelX` and `labelY` props respectively: https://github.com/adobe/spectrum-web-components/blob/a447ba2714483619a17ab73e11efc79f9f815c84/packages/color-area/src/ColorArea.ts#L53-L60

For Color Area, it may be better to include the color channel names as part of the `aria-valuetext` for each of the inputs, so that the user aware of which channel their interaction changed, and we should include additional strings:

- 'Color Picker' as the default `aria-label` for each of the `input[type="range"]` elements, which should be concatenated with whatever `label` the user provides, so that the control announces something like "Foreground Color Picker".
- '2d slider' as the `aria-roledescription` for each of the `input[type="range"]` elements, so that he user gets some indication of the keyboard behavior for the control and understands that keyboard interaction changes values in two dimensions.

[...]

PR #3110, opened 2023-04-12
fix(picker): add loading state to the picker

razvancir96:
## Description

We need to add a loading spinner inside the picker's button.
[Spectrum CSS](https://opensource.adobe.com/spectrum-css/preview/?path=/story/components-picker--default&args=size:xl;isLoading:!true)

## Related issue(s)

- https://github.com/adobe/spectrum-web-components/issues/3076

## Motivation and context

In one of Adobe's projects, we need to support the loading spinner inside the picker's button.

## How has this been tested?

-   [x] _Test case 1_
    1. Run `yarn storybook`
    2. Go to Picker -> Sizes stories: https://localhost:8001/?path=/story/picker-sizes--s 
    3. Change the sizes and from the controls, set the `pending` property on true. Also, change the scale.

Size S:
Screenshot 2024-02-27 at 17 43 21

Size M:
Screenshot 2024-02-27 at 17 43 29

Size L:
Screenshot 2024-02-27 at 17 43 35

Size XL:
Screenshot 2024-02-27 at 17 43 41

## Types of changes

## Checklist

## Best practices


packages/picker/test/index.ts
```diff
@@ -286,6 +286,13 @@ export function runPickerTests(): void {
             expect(el.invalid).to.be.true;
             await expect(el).to.be.accessible();
         });
+        it('renders loading accessibly', async () => {
+            el.loading = true;
+            await elementUpdated(el);
+
+            expect(el.loading).to.be.true;
+            await expect(el).to.be.accessible();
```
Westbrook, line 294: This test "is this generally accessible" which is great, but it doesn't test for a specific accessibility tree output. I think we want to do something like https://github.com/adobe/spectrum-web-components/blob/main/packages/picker/test/index.ts#L1136-L1147 and test for "Picker in loading state Loading...". This will ensure the `<button>` isn't `disabled` and that the expected labels are being applied.

packages/picker/src/Picker.ts
```diff
@@ -406,6 +412,15 @@ export class PickerBase extends SizedMixin(Focusable) {
                           ></sp-icon-alert>
                       `
                     : nothing}
+                ${this.loading
+                    ? html`
+                          <sp-progress-circle
+                              indeterminate
+                              aria-labelledby="label"
```
Westbrook, line 419: `aria-hidden` below cancels this out, so we can exclude this line.

Westbrook (changes requested):
Great start. Mainly want to make sure we're shoring up the accessibility tree that this delivers and clarifying how we're preventing interaction. @jnurthen if you have any suggestions beyond what's here, would love your insight.

packages/picker/src/Picker.ts
```diff
@@ -433,7 +448,7 @@ export class PickerBase extends SizedMixin(Focusable) {
                 @blur=${this.onButtonBlur}
                 @click=${this.onButtonClick}
                 @focus=${this.onButtonFocus}
-                ?disabled=${this.disabled}
+                ?disabled=${this.disabled || this.loading}
```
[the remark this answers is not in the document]
Westbrook: Agreed, we shouldn't use `disabled` directly on the `<button>` for this. We should do that functionally within the component, but the screen reader/tab order should still see the element as _active_, with the "Loading"-centric labelling implying that nothing will happen when you actually interacti with it, yet.

packages/picker/src/Picker.ts
```diff
@@ -406,6 +412,15 @@ export class PickerBase extends SizedMixin(Focusable) {
...
                     : nothing}
+                ${this.loading
+                    ? html`
+                          <sp-progress-circle
+                              indeterminate
+                              aria-labelledby="label"
+                              aria-hidden="true"
+                          ></sp-progress-circle>
+                      `
+                    : nothing}
```
paulodetarsofm, line 423: Just a suggestion, instead of using ternary operators in which we need to inform `nothing` as "else", we could use the `when` directive:
```html
                ${when(
                    this.loading,
                    () => html`
                        <sp-progress-circle
                            indeterminate
                            aria-hidden="true"
                        ></sp-progress-circle>
                    `
                )}
```
😁

packages/picker/src/Picker.ts
```diff
@@ -105,6 +107,9 @@ export class PickerBase extends SizedMixin(Focusable) {
     @property({ type: Boolean, reflect: true })
     public invalid = false;
 
+    @property({ type: Boolean, reflect: true })
+    public loading = false;
```
Westbrook, line 111: I think we should normalize this to the API recently added to `<sp-button>`, [pending](https://fix-switch-onchange-event--spectrum-web-components.netlify.app/components/button/#pending).

Westbrook (commented):
@razvancir96 could you sync up with @TarunAdobe about normalizing this with the work in https://github.com/adobe/spectrum-web-components/pull/3163 so we could maybe get this over the line?

Rocss:
After a discussion with @razvancir96 we agreed that our team will take over this PR from next week.
@Westbrook are there any concerns about this functionality besides the code review on this PR?

Westbrook:
@Rocss please sync up with @TarunAdobe about normalizing this PR and the API applied to the Picker herein with the work done in https://github.com/adobe/spectrum-web-components/pull/3163.

Rocss:
Given that the progress spinner is spinning inside the picker, VRT will fail because of different states of the spinner. Was there anything similar being done in the repo so far?

packages/picker/src/Picker.ts
Rocss: `sp-progress-circle` throws a11y warnings if label is not provided, even when `aria-hidden` is present. Should we add a check for `aria-hidden` in this case too?

packages/picker/src/Picker.ts
```diff
@@ -405,13 +410,33 @@ export class PickerBase extends SizedMixin(Focusable, { noDefaultSize: true }) {
...
                               class="validation-icon"
                           ></sp-icon-alert>
                       `
                     : nothing}
+                ${when(this.pending, () => {
+                    import(
+                        '@spectrum-web-components/progress-circle/sp-progress-circle.js'
+                    );
+                    return html`
+                        <span
```
Rocss, line 425: Why this `span` you may ask:

If you want the picker to automagically add "loading" (or any other synonym) to its accessible name when `pending` is true, you need to add either this visually hidden span, or to add a label on the `sp-progress-circle` and link it with the button through `aria-labelledby`. 

When the second option is added, the screen reader actually reads the percentage of the "progress-circle" as well ("Loading 50 Choose an option.... Country") 🤦‍♀️ .

Or we can not automagically add the "loading" part to the name and let the user handle that through the picker's label.

Westbrook: @TarunAdobe in a "pending button" we obsolete the (previously) visible test in the accessibility tree for the provided "Loading" text. Note that that content is _provided_ so that i18n can happen on it.

Do you think we want to dump all of the label data in this case as well?
Should we dump all of the value data and only use `${labelData} ${loadingLabel}`? 
Can you sync with the accessibility team to get their thoughts on what a screen reader should be departing to a visitor in this context?

TarunAdobe: `When the second option is added, the screen reader actually reads the percentage of the "progress-circle" as well ("Loading 50 Choose an option.... Country") 🤦‍♀️ .`
In our case the progress-circle would be indeterminate so the screen reader should just read "Loading Choose an Option.... Country"? Am I missing something here?

Westbrook:
For stability in the VRTs, checkout https://github.com/adobe/spectrum-web-components/pull/4055. It should "just work ™️", 🤞🏼, after that PR lands.

packages/picker/src/Picker.ts
Rocss: @Westbrook asked the accessibility team for some feedback and they said it's okay. The computed accessibility name for the Picker’s rendered button element will be “_Loading Select an option… Country._”
However, "Loading" is not localized. I am not aware if there's something for i18n in SWC, or else we should let the user provide a localized pending label, as the Button does.

@TarunAdobe I had the same expectation. However, unit tests fail with:
```
Browser logs on Firefox:
          {
            role: 'buttonmenu',
            name: 'Loading 50 Select an option... Country',
            haspopup: 'true'
          }
```
when using this approach:
```
<button
  .....
  aria-labelledby="loader-label icon label applied-label"
  ....
/>

......

  <sp-progress-circle
      id="loader-label"
      label="Loading"
      size="s"
      indeterminate
      class="progress-circle"
  ></sp-progress-circle>

```
Chromium and Webkit do not have this issue.

Westbrook: An external label is best for right now. While we have[ this issue](https://github.com/adobe/spectrum-web-components/issues/1975), not telling when it will be prioritized/addressed.

For the Progress Circle in Firefox... it may be that that is a "situation" (neither a bug or a feature) in Firefox specifically for the Progress Circle.
- @TarunAdobe can you help confirm the actual label delivered in that case, separate of this Picker change?
  - if it's something about Progress Circle, we can make an issue and come back to it separately
  - if it's not, it may be a timing question in that Picker could be `updateComplete` before Progress Circle is, which means a couple of `nextFrame()`s might tip the scale.

Rocss: I tested the same thing in the `progress-circle.test.ts`:
```
const el = await fixture<ProgressCircle>(html`
            <input aria-labelledby="test" />
            <sp-progress-circle
                indeterminate
                label="Loading"
                id="test"
            ></sp-progress-circle>
        `);
```

result is 
```
 🚧 Browser logs on Firefox:
      {
        role: 'document',
        name: '',
        children: [
          { role: 'textbox', name: 'Loading 50' },
          {
            role: 'progressbar',
            name: 'Loading',
            valuetext: '50',
            checked: 'mixed',
            value: '50'
          }
        ]
      }
```

One way to bypass that "50" value is by using `aria-valuetext=""` although it seems hackish.

Rocss: @TarunAdobe would you please check this ^ out? I addressed all the other comments in this PR.

packages/picker/src/Picker.ts
```diff
@@ -405,13 +415,33 @@ export class PickerBase extends SizedMixin(Focusable, { noDefaultSize: true }) {
...
+                            aria-hidden="true"
+                            class="visually-hidden"
+                            id="loader-label"
+                        >
+                            ${this.pendingLabel}
+                        </span>
+                        <sp-progress-circle
```
Westbrook, line 437: Button uses `static=white`, does design have direction on this?

packages/picker/src/picker.css
```diff
@@ -36,6 +36,39 @@ governing permissions and limitations under the License.
...
+        --spectrum-progress-circle-thickness-medium
+    );
+}
+
+:host([size='xl']) sp-progress-circle {
+    --spectrum-progress-circle-size: var(--spectrum-workflow-icon-size-300);
+    --spectrum-progress-circle-thickness: var(
+        --spectrum-progress-circle-thickness-large
+    );
+}
```
Westbrook, line 70: These values appear slightly different than those in Button, please make sure they align. We could even go so far as to externalize these styles to Progress Circle and leverage them from there in both places if we wanted.

packages/picker/src/spectrum-picker.css
```diff
@@ -687,7 +687,7 @@ governing permissions and limitations under the License.
         )
     );
 }
-#button.is-loading .picker {
+:host([pending]) #button .picker {
```
Westbrook, line 690: `pending` in Button delays, which makes sense for Button, we're attempting to cede user interface to processing. Here, the `pending` would likely come from an unrelated interaction, so does it make any sense to delay here as well? @TarunAdobe

packages/picker/stories/picker.stories.ts
```diff
@@ -34,8 +34,10 @@ export default {
         invalid: false,
         open: false,
         quiet: false,
+        pending: false,
     },
     argTypes: {
+        ...argTypes,
```
Westbrook, line 40: Would you do us the favor while you're touching Picker and add an `options` array to https://github.com/adobe/spectrum-web-components/blob/razvancir96/add-picker-loading-state/packages/picker/stories/args.ts#L29 as a sibling to the `controls` property that listed `options: [ 'Small', 'Medium', 'Large', 'Extra Large' ]`?

packages/picker/stories/picker-sizes.stories.ts
```diff
@@ -22,21 +22,49 @@ export default {
...
+                type: 'boolean',
+            },
+        },
     },
 };
 
 type StoryArgs = {
     onChange: (val: string) => void;
+    invalid: boolean;
+    pending: boolean;
```
Westbrook, line 53: @TarunAdobe we broken Button into a while list of sized stories, particularly due to the custom sizing of the Progress Circle. Have you found value in that? Should we do it here, too?

packages/picker/README.md
```diff
@@ -376,7 +376,7 @@ When the `value` of an `<sp-picker>` matches the `value` attribute or the trimme
 </sp-picker>
 <br />
 <br />
-<sp-field-label for="picker-disabled">Quiet:</sp-field-label>
+<sp-field-label for="picker-disabled-quiet">Quiet:</sp-field-label>
```
Westbrook, line 379: 🙇🏼

packages/picker/src/Picker.ts
Rocss: I could not find anything around this in https://spectrum.adobe.com/page/picker/, so I implemented it the same way it is in spectrum-css. But let me ping design and ask about this 👍

packages/picker/src/picker.css
Rocss: Yep, these were here since the PR was first opened, I was not sure if you guys already discussed around them.
Wouldn't this `--spectrum-progress-circle-size` make more sense in [spectrum-css](https://github.com/adobe/spectrum-css/blob/main/components/picker/index.css#L88)? I observed that in the spectrum-css storybook the spinner has the same dimension regardless of the Picker's size. Pinging again @pfulton about this 😄

packages/picker/src/spectrum-picker.css
TarunAdobe: I agree Picker should doesn't need to delay the `pending` effect.

packages/picker/src/Picker.ts
Rocss: @Westbrook but why is it `static=white` in Button? In https://spectrum.adobe.com/page/button/#Pending it is not static, and in light theme the color contrast is quite low:

Screenshot 2024-02-27 at 17 18 50

packages/picker/src/Picker.ts
TarunAdobe: Hey @Rocss... I've confirmed this behaviour in firefox and for us right now the `hackish` way seems like a slightly better option than to have a separate `<span/>` tag. So maybe we should move forward in that direction unless you have different thoughts?

Rocss: Another option would be to completely replace the label inside the picker with the `pendingLabel`, this way it would be read by screen readers automatically like: "Pending... Country"

Rocss:
I have no more things to add to this PR. Happy to implement any other feedback if there is.
cc: @Westbrook @TarunAdobe

TarunAdobe:
Hey @Rocss I just have a final comment before we can land this... we'd want to split up the Picker into a list of sized stories so that we can track down how the pending progress-circle looks in different sizes just like we do in button.

TarunAdobe (approved):
LGTM! Tagging @Westbrook for one final pass before we land this.

packages/overlay/stories/overlay.stories.ts
```diff
@@ -162,11 +162,7 @@ const template = ({
             type=${ifDefined(type)}
         >
             <sp-button variant="primary" slot="trigger">Show Popover</sp-button>
-            <sp-popover
```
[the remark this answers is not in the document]
Westbrook: I wonder if there's a loose config somewhere. Not germain to this PR, indeed.

Westbrook (approved):
Thanks for adopting this lost child of a PR and giving it a good home! :shipit:

PR #3141, opened 2023-04-21
feat(icon): image src invalid error api

vikalpgupta:
Adding support for error API within the icon component

## Description

Added error API support for icon component having invalid src passed, in such case the component will announce an `Event("error")` which can be handled by parent component

## Related issue(s)

-

## How has this been tested?

-   [x] _Test case 1_
    Added a test case

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/icon/src/Icon.ts
+++ b/packages/icon/src/Icon.ts
@@ -73,9 +73,9 @@ export class Icon extends IconBase {
     private announceIconImageSrcError(): void {
         this.dispatchEvent(
             new Event('error', {
-                cancelable: true,
-                bubbles: true,
-                composed: true,
+                cancelable: false,
+                bubbles: false,
+                composed: false,
             })
         );
     }
--- a/packages/icon/src/Icon.ts
+++ b/packages/icon/src/Icon.ts
@@ -70,6 +70,16 @@ export class Icon extends IconBase {
         }
     };
 
+    private announceIconImageSrcError(): void {
+        this.dispatchEvent(
+            new Event('error', {
+                cancelable: true,
+                bubbles: true,
+                composed: true,
+            })
+        );
+    }
+
     protected override render(): TemplateResult {
         if (this.name) {
             return html`
@@ -77,7 +87,11 @@ export class Icon extends IconBase {
             `;
         } else if (this.src) {
             return html`
-                <img src="${this.src}" alt=${ifDefined(this.label)} />
+                <img
+                    src="${this.src}"
+                    alt=${ifDefined(this.label)}
+                    @error=${this.announceIconImageSrcError}
+                />
             `;
         }
         return super.render();
--- a/packages/icon/stories/icon.stories.ts
+++ b/packages/icon/stories/icon.stories.ts
@@ -57,6 +57,27 @@ export const imageIcon = (): TemplateResult => {
 
 imageIcon.storyName = 'Image Icon';
 
+export const imageIconSrcError = (): TemplateResult => {
+    const error = (): void => {
+        console.error('Invalid sp-icon src provided');
+    };
+
+    return html`
+        ${sizes.map(
+            (size) => html`
+                <sp-icon
+                    label="Back"
+                    size=${size}
+                    src=${back}
+                    @error=${error}
+                ></sp-icon>
+            `
+        )}
+    `;
+};
+
+imageIcon.storyName = 'Image Icon src invalid error';
+
 export const svgIcon = (): TemplateResult => {
     return html`
         ${sizes.map(
--- a/packages/icon/stories/icon.stories.ts
+++ b/packages/icon/stories/icon.stories.ts
@@ -77,7 +77,11 @@ export const imageIconSrcError = (): TemplateResult => {
     `;
 };
 
-imageIcon.storyName = 'Image Icon src invalid error';
+imageIconSrcError.storyName = 'Image Icon src invalid error';
+
+imageIconSrcError.swc_vrt = {
+    skip: true,
+};
 
 export const svgIcon = (): TemplateResult => {
     return html`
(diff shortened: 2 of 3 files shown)


packages/icon/src/Icon.ts
```diff
@@ -70,14 +70,28 @@ export class Icon extends IconBase {
         }
     };
 
+    private announceIconImageSrcError(): void {
+        this.dispatchEvent(
+            new Event('error', {
+                cancelable: true,
+                bubbles: true,
+                composed: true,
```
Westbrook, line 78: The default values for the event options in the event that you are redispatching here are all `false`. Can you share a little what you expect to do with these value as `true`?

Specifically, you've not added any code to support an application interacting with the `cancelability` of the event, but I'm not sure even what you could do there is you decided to. Taking it out might be the right call.

For `bubbles` and `composed`, is being able to hear this message outside of the shadow root where the `<sp-icon>` element is that useful to a consuming developer? Answers here can help form the basis for a little documentation around this addition so that this feature can be well leveraged by others.

vikalpgupta: I made the silly mistake of copy-pasting this code and setting these back to default again, let me know if you have any more suggestions, Thanks

packages/icon/test/icon.test.ts
```diff
@@ -69,6 +70,25 @@ describe('Icon', () => {
...
+                    @error=${error}
+                ></sp-icon>
+            `
+        );
+
+        await elementUpdated(el);
+
+        await expect(el).to.be.accessible();
+
+        expect(error).to.be.calledWithExactly(new Event('error'));
```
Westbrook, line 89: Does WebKit pass on this test for you locally? CI is showing `expected spy to have been called with exact arguments { isTrusted: [Getter] }`. If you thought there was something about the event construction you thought could match that up, that would be great. However, a generic `error.called` or `error.calledOnce` would also be enough here as there's no real information in the error itself.

vikalpgupta: Seems the spy is not getting called at all, looking for a way to check whether the same is supported via Webkit

Westbrook: You may also try changing the order of the bindings here. While it would be a bit unexpected, the error could be dispatched synchronously with the erroneous `src` application, which currently happens just before the event binding.

vikalpgupta: Found a bug logged for the same on WebKit's end https://github.com/microsoft/playwright/issues/22332

vikalpgupta: @Westbrook Any suggestion on this, what could be done next if this is only failing for playwright webkit type ones

Westbrook: Like I mentioned above, `error.called` or `error.calledOnce` or `error.callCount` are all fine alternatives to `calledWithExactly` In my opinion.

vikalpgupta: But for the playwright webkit, the above suggestions are also failing, I hope we are good without this feature for the said type browser for now

Westbrook: I see, I didn't quite catch that. You can leverage the helpers available here: https://github.com/adobe/spectrum-web-components/blob/main/tools/shared/src/platform.ts#L52 to query the browser and `skip` the test just for WebKit. Please be sure to add a comment as to why and link back to that issue.

packages/icon/stories/icon.stories.ts
```diff
@@ -57,6 +57,28 @@ export const imageIcon = (): TemplateResult => {
...
+                    size=${size}
+                    src=${invalidImgSrc}
+                    @error=${error}
+                ></sp-icon>
+            `
+        )}
+    `;
+};
+
+imageIcon.storyName = 'Image Icon src invalid error';
```
Westbrook, line 80: Please include the following to exclude this story from our VRT suite.
```
imageIcon.swc_vrt = {
    skip: true,
};
```

Westbrook (changes requested):
This is looking good! Thanks for the contribution.

If you can get that one change to the story and then rebase the branch to `main`, the only thing left will be signing our CLA and this will be good to ship.

vikalpgupta:
Closing the pull request to reflect the CLA signed check

vikalpgupta:
Closing the pull request to reflect the CLA signed check

packages/icon/test/icon.test.ts
```diff
@@ -69,6 +71,28 @@ describe('Icon', () => {
...
+            `
+        );
+
+        await elementUpdated(el);
+
+        await expect(el).to.be.accessible();
+
+        // Skipping the test case expectation for webkit because of error event not dispatching bug for the same, https://github.com/microsoft/playwright/issues/22332
+        if (!isWebKit()) {
+            expect(error).to.be.calledOnce;
```
Westbrook, line 92: [This is why we wait for CI to run before merging](https://app.circleci.com/pipelines/github/adobe/spectrum-web-components/15630/workflows/576533f7-9d4d-428b-80a2-6c0c0ac71011/jobs/322293)... It looks like the timing on this test can be flaky (at least in Firefox). Hopefully, it's just timing an not something more complicated. Can you try this?
```suggestion
            await waitUntil(() => error.calledOnce, 'The error was thrown.');
```
This will give all of the browsers as much as two seconds for the error to be thrown. It _should_ be OK, as in some passes Firefox has passed, but it's hard to tell 100% what's going on from those logs.

Westbrook (changes requested):
Looking good with the CLA signed. I've rebased to `main` and assuming everything runs fine we'll merge this in after that. Thanks for helping us work through the tests here so we can include this additional feature on the Icon pattern. 🙇🏼

packages/icon/test/icon.test.ts
Westbrook: You need to import this method.

Westbrook (approved):
I think we're good to go!! Thanks for the contribution.

PR #3163, opened 2023-04-28
fix(button): adds pending button, fixes #3162

### LINKED ISSUE #3162: [Feat]: Support 'pending' button states
### Code of conduct

### Impacted component(s)

button

We should implement the 'pending' button state as described here: https://spectrum.adobe.com/page/button/#Pending

### Mockups or screenshots

_No response_

_No response_

benjamind:
## Description

Adds the 'pending' button state to the `<sp-button>` element.

The `pending` attribute drives the state, and internally starts the 1s cooldown timer to set the button state to `disabled`. The spinner will then render when the button is both `disabled` and `pending`.

Main question outstanding I think is how to manage the import here, I don't think we want all buttons automatically requiring `sp-progress-circle` its unnecessary page weight. My suggestion is we leave including the spinner up to the end user and maybe in dev we add a warning if its not been registered when using pending state?

## Related issue(s)

- fixes #3162 

## Motivation and context

Pending buttons are part of the Spectrum design spec, though they should be used sparingly they are occasionally useful.

## How has this been tested?

Added a unit test to confirm the button is being driven into the disabled state at the right time. I assume VRT would confirm the rest (though I have not yet updated VRT).

Screenshot 2023-04-27 at 6 01 32 PM

## Types of changes

## Checklist

## Best practices


packages/button/src/Button.ts
```diff
@@ -51,6 +53,45 @@ export class Button extends SizedMixin(ButtonBase) {
...
+                    this.disabledByPending = true;
+                }
+                this.disabled = true;
+            }, 1000);
+        } else {
+            window.clearTimeout(this.pendingCooldown);
+            if (this.disabledByPending) {
+                this.disabled = false;
+            }
+        }
```
Westbrook, line 92: Disabling the button does slightly different things to the accessibility tree delivered than I think that we want to here. Name you get "Button, dimmed". We likely want to surface a secondary label as an attribute that we can apply when when `pending`. We're looking at this in #3110 for Picker, so we should make sure the approaches are aligned.

packages/button/src/Button.ts
```diff
@@ -136,4 +177,15 @@ export class Button extends SizedMixin(ButtonBase) {
...
     }
+
+    protected override renderButton(): TemplateResult {
+        return html`
+            ${this.buttonContent}
+            <sp-progress-circle
+                indeterminate
+                static="white"
+                size="s"
+            ></sp-progress-circle>
```
Westbrook, line 188: Should this always be available in the doc? If that's the approach we want, then we'll need something like `aria-hidden=${ifDefined(!this.pending ? 'true' : undefined)}` to keep it off of the accessibility tree.

Also, it doesn't look like this registration is ever loaded. Should it be static, dynamic, in the docs?

packages/button/src/button.css
```diff
@@ -25,3 +25,24 @@ governing permissions and limitations under the License.
...
+:host([pending][disabled]) sp-progress-circle {
+    display: block;
+    position: absolute;
+    left: 50%;
+    transform: translate(-50%, 0);
+}
+
+:host([pending][disabled]) slot[name='icon'],
+:host([pending][disabled]) #label {
+    visibility: hidden;
```
Westbrook, line 47: Can't remember what visibly hiding this content does to the accessibility tree right off, so making a note to confirm.

packages/button/src/Button.ts
```diff
@@ -136,4 +177,15 @@ export class Button extends SizedMixin(ButtonBase) {
...
         }
     }
+
+    protected override renderButton(): TemplateResult {
+        return html`
+            ${this.buttonContent}
+            <sp-progress-circle
+                indeterminate
+                static="white"
+                size="s"
```
Westbrook, line 187: We likely want to match the t-shirt size approach to that of #3110 as well.

packages/button/test/button.test.ts
```diff
@@ -513,6 +513,32 @@ describe('Button', () => {
...
+                el.disabled;
+            },
+            'button is disabled',
+            { interval: 200, timeout: 1200 }
+        );
+
+        await elementUpdated(el);
+
+        expect(el.disabled).to.be.true;
+        expect(el.pending).to.be.true;
```
Westbrook, line 540: Using an accessibility snap shot here could support us in guaranteeing a specific entry in the a11y tree.

Westbrook (commented):
Those helpers in the buttons do help a good deal of boilerplate, but there is a lot left there 😓 

Good first pass, a couple of areas to confirm what we want to offer in the API and then away we go!

packages/button/src/Button.ts
benjamind: yeah my feeling is it should be in the docs only (you don't want to lazy load this it might not show up for long as you're only supposed to use these pending states for < 5s loads). Maybe a dev mode warning if its used and pending-circle isn't registered?

packages/button/src/Button.ts
benjamind: Yeah i had a play with this, was 'tricky', one step down in size was not always the rule, and the sizes didn't seem to align well.

benjamind:
> @benjamind are you able to look into getting this over the line?

Only a month later.....(my SWC notifications are a mess)...yes I can try and find some time now to push this over.

packages/button/src/Button.ts
benjamind: Switch to using `is-pending` state, and modified the spectrum-css conversion config to add the right `is-pending` attribute where it finds `disabled`. This is working with a bit of a hack, and I'm not sure of the 'right' fix here, though I think the hack has a low chance of failing?

See here https://github.com/adobe/spectrum-web-components/pull/3163/files#diff-68a09e148bb1fa4125cc905579fa756d38725e86244b153a1ec4c819eaffed5fR44

packages/button/src/Button.ts
benjamind: Modified this to only add the element when needed, this avoids the aria issue. Have still not made a decision on importing, I'm leaning towards just documentation of the need for `sp-progress-circle` being a required dependency when using `pending` state though? This would then avoid unnecessary payload. Maybe add a `dev` warning if the element is not registered?

packages/button/src/button.css
benjamind: Checked in voiceover and the label seems to be hidden? Is there some other way I can test?

packages/button/test/button.test.ts
benjamind: Updated to use the a11y snapshot testing for the `aria-label` changing to the `pending-label` value.

packages/button/src/Button.ts
benjamind: Did some more testing here and there's no good sizes here that work. `s` size progress-circle on `s` sized button is too big, `s` on `m` button is about right, but `m` on `l` button is too big.

Not sure what to do here, we could apply `scale` on the transform for the progress-circle to tweak it at each t-shirt size perhaps? But it'll likely be a custom value per button size.

packages/radio/src/radio.css
```diff
@@ -23,7 +23,7 @@ governing permissions and limitations under the License.
 /**
  * Work around for https://github.com/adobe/spectrum-css/issues/2251
  **/
-:host([dir="rtl"]) #button:after {
+:host([dir='rtl']) #button:after {
```
Westbrook, line 26: I'm not a fan of the lint changes falling into unrelated PRs...

packages/button/src/Button.ts
```diff
@@ -51,6 +54,42 @@ export class Button extends SizedMixin(ButtonBase, { noDefaultSize: true }) {
...
+    }
+
+    @property({ type: Boolean, reflect: true, attribute: 'is-pending' })
+    public isPending = false;
+
+    @property({ type: String, attribute: 'pending-label' })
+    public pendingLabel = 'Pending';
+
+    @property({ type: Boolean, reflect: true, attribute: true })
+    public pending = false;
```
Westbrook, line 84: I don't really understand the use of `pending` and `isPending`. Can you add some comments to clarify the workflow that's being acheived?

benjamind: Yeah `pending` is the indication that this is a pending state button, while `is-pending` is added/removed during the pending state. Agree needs comments.

packages/button/src/Button.ts
```diff
@@ -140,4 +198,22 @@ export class Button extends SizedMixin(ButtonBase, { noDefaultSize: true }) {
...
+            ${this.buttonContent}
+            ${when(this.isPending, () => {
+                import(
+                    '@spectrum-web-components/progress-circle/sp-progress-circle.js'
+                );
+                return html`
+                    <sp-progress-circle
+                        indeterminate
+                        static="white"
+                        size="s"
```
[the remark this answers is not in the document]
benjamind: In internal discussion with the Spectrum-CSS team, specifically @KayiuCarlson we concluded that there are no good t-shirt sizes right now to use for the progress circle inside of a button in these cases. So we concluded that we should fudge the sizes manually here to match the workflow-icon sizes.

packages/button/src/Button.ts
Westbrook: That timing seems weird, though it does make the code make so much more sense.

Based on this [documentation](https://spectrum.adobe.com/page/button/#Pending), it feels like `pending` should be an immediate state. Was there something specific to your application requirements that lead to this relationship?

packages/radio/src/radio.css
TarunAdobe: agreed! I should be more careful with this 💯

packages/button/src/Button.ts
```diff
@@ -51,6 +54,42 @@ export class Button extends SizedMixin(ButtonBase, { noDefaultSize: true }) {
...
+                    event.preventDefault();
+                    event.stopImmediatePropagation();
+                    event.stopPropagation();
+                    return false;
+                }
+            },
+            {
+                capture: true,
+            }
+        );
```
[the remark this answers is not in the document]
TarunAdobe: `pointer-events:none` should be better in this case so as to avoid any event propagation at all during pending state. 
Note: we are overriding the click function to avoid clicking on pending either way so even with `pointer-events:none`  we won't run into cases where users tries to click the button programatically while in pending state!

packages/button/src/button.css
```diff
@@ -21,3 +21,32 @@ governing permissions and limitations under the License.
...
 }
+
+::slotted([slot='icon']) {
+    inset: unset;
+}
+
+#container {
+    position: relative;
+    display: flex;
+}
```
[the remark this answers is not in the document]
TarunAdobe: I don't think so... maybe it did during the initial phase of this PR however now it's useless. so im removing it.

packages/button/src/button.css
```diff
@@ -21,3 +21,32 @@ governing permissions and limitations under the License.
         border-color: highlight;
     }
 }
+
+::slotted([slot='icon']) {
+    inset: unset;
+}
```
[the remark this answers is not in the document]
TarunAdobe: same for this one... It doesn't seem to do anything in context of what we are trying to achieve.

packages/button/src/button.css
```diff
@@ -21,3 +21,32 @@ governing permissions and limitations under the License.
...
+
+#container {
+    position: relative;
+    display: flex;
+}
+
+sp-progress-circle {
+    display: none;
+}
```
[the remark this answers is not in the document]
TarunAdobe: checked and found this is redundant!

packages/button/src/Button.ts
TarunAdobe: `pending` is the property that the user supplies to the component upon which we update the `isPending` state variable to true and do the necessary css updates. `isPending` represents the state of the component and can be accessed through DOM Apis.

benjamind: Well it was more a matter of convenience at point of use. If the `pending` state is added as an indicator of future behavior, then the click handler behaves like a normal button, with the click event firing on first click. The button then adds its own `isPending` state to track its timeout period and disable further clicks during that time.

If the consumer has to add the `pending` state when click occurs, thats a little more work for the consumer to deal with? Not sure, I could go both ways on this, but it felt convenient to have it this way, and it marks the button in the consumers render template as a pending state button so its quite clear to see that the behavior will be distinct here, rather than a side effect of the buttons click handler?

Westbrook: This is an interesting approach, but not seemingly what the code says?
```
        if (changedProperties.has('pending')) {
            if (this.pending) {
                this.pendingCooldown = window.setTimeout(() => {
                    if (!this.disabled) {
                        this.isPending = true;
                        this.setAttribute('aria-label', this.pendingLabel);
                    }
                }, 1000);
            } else if (this.pending === false && this.isPending) {
                window.clearTimeout(this.pendingCooldown);
                this.isPending = false;
                this.setAttribute('aria-label', this.label || '');
            }
        }
```
This may be updated functionality, but this says that as soon as you set `pending` you start the clock for `isPending`. 

I discussed why this might be in with the design team, and they outlined a _seemingly_ valid use case of no wanting to go to "visually pending" and then "processed" in flicker speed progression. It's valid, but it's valid at any point in the process where you might switch to `pending`, which makes me wonder if it. would be better to require an application manage any staged loading they do local to where there can be some understanding of the time required to do the actual task in question.

I can be convinced of either way, with clear comments that outline a workflow that we all see the same way, but right now I'm leaning towards it being better to only offer `pending` and it being immediate.

benjamind: Ah yes sorry forgot that nuance (its been a while since i started this one)....

I think if we leave management of the visually pending state to the consumer, like as not it will get implemented poorly in many places and result in a multitude of flickering bugs in a variety of places with solutions to it varying across the board.

But yes the state flow as I recall now is:

* Enabled button
* Click
* Click handler - consumer assigns `pending` state
* `isPending` timeout begins
* `isPending` && `pending` causes display of 'visually pending' state
* Time goes by
* Consumer removes `pending` state
* `isPending` is removed
* Enabled button.

TarunAdobe:
@Westbrook I have added the tshirt sizing for the progress-circle... Does this PR need more work?

packages/button/src/Button.ts
```diff
@@ -140,4 +183,22 @@ export class Button extends SizedMixin(ButtonBase, { noDefaultSize: true }) {
...
+            ${this.buttonContent}
+            ${when(this.isPending, () => {
+                import(
+                    '@spectrum-web-components/progress-circle/sp-progress-circle.js'
+                );
+                return html`
+                    <sp-progress-circle
+                        indeterminate
+                        static="white"
+                        label=${this.pendingLabel + 'progress circle'}
```
Westbrook, line 198: Do we need the `progress circle` part? That forces un-translated content into the label.

Separately, is it even possible to get to the Progress Circle in the accessibility tree? Would it be better to use `aria-hidden="true"`, as this is namely a visible affordance for the change in label on the Button?

packages/button/src/Button.ts
Westbrook: Yeah, this is ugly pretty much regardless... @TarunAdobe can you test this approach which would simplify to a single state variable.

- Enable button
- Click
- Consumer assignes `pending`
- The 'visually pending' state is applied via an immediate transition or animation that has a delay of `1000ms` but is customizable via a CSS Custom Property
- Time goes by
- Consumer removes `pending`
- The 'visually pending' state is removed via an immediate transition or animation that has no delay
- Button is enable

packages/button/src/button.css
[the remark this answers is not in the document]
TarunAdobe: With the removal of isPending variable in the button. this is now back and useful again as we want to show the progress-circle only when the is-pending attribute is set after the animation

packages/button/src/Button.ts
TarunAdobe: done

packages/button/src/Button.ts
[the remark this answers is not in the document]
TarunAdobe: now using tshirt sizing for progress-circle.

packages/button/src/Button.ts
TarunAdobe: you're right! this is just a visible affordance so it makes much more sense to have aria-hidden="true"

packages/button/src/Button.ts
```diff
@@ -127,11 +144,25 @@ export class Button extends SizedMixin(ButtonBase, { noDefaultSize: true }) {
...
 
+    protected override willUpdate(
+        changedProperties: PropertyValues<this>
+    ): void {
+        super.willUpdate(changedProperties);
+        if (changedProperties.has('pending')) {
+            if (this.pending) {
+                if (!this.disabled) {
+                    this.setAttribute('aria-label', this.pendingLabel);
+                }
```
Westbrook, line 159: Good move here. We do need coverage for when `this.disabled` moves to `false` and `this.pending` is already `true` to ensure the `aria-label` is appropriately applied in that case.

packages/button/src/button.css
```diff
@@ -21,3 +21,72 @@ governing permissions and limitations under the License.
...
+    0% {
+        visibility: visible;
+    }
+
+    100% {
+        visibility: hidden;
+    }
+}
+
+@keyframes update-button-styles {
```
Westbrook, line 45: Let's name this something about `pending`, so it's more clear what it is for.

packages/button/src/button.css
```diff
@@ -21,3 +21,72 @@ governing permissions and limitations under the License.
...
+
+@keyframes hide-icons-label {
+    0% {
+        visibility: visible;
+    }
+
+    100% {
+        visibility: hidden;
+    }
+}
```
Westbrook, line 43: Does https://developer.mozilla.org/en-US/docs/Web/CSS/animation-direction make it so we only need one of these?

TarunAdobe: Nope... Since on page load the animation state for everything is 0. and by applying `backwards` or `reverse` to the animation property it takes the current animation state and pushes it to the respective (0 or 100 state). Since at the start the animation was at 0 for icons (say) then even if I do animation backwards on icons with the `show-progress-circle` animation it will not set them to hidden because it is already in 0% state.

packages/button/src/button.css
TarunAdobe: done

packages/button/src/button.css
Westbrook: Really wanted that to work nicely 😞 Seems like it's _possible_, but the actual complexities of it out weight the benefits. Thanks for checking!

packages/button/src/Button.ts
```diff
@@ -127,11 +144,33 @@ export class Button extends SizedMixin(ButtonBase, { noDefaultSize: true }) {
...
+            } else if (
+                changedProperties.get('pending') !== undefined &&
+                !this.pending
```
Westbrook, line 162: What do you see us addressing by the more complex condition here?

packages/button/src/Button.ts
```diff
@@ -127,11 +144,33 @@ export class Button extends SizedMixin(ButtonBase, { noDefaultSize: true }) {
...
+            ) {
+                this.setAttribute('aria-label', this.label || '');
+            }
+        }
+        if (changedProperties.has('disabled')) {
+            if (this.pending && !this.disabled) {
+                this.setAttribute('aria-label', this.pendingLabel);
+            }
```
Westbrook, line 170: `if (this.pending && this.disabled)` should we remove the `aria-label` in this case?

packages/button/src/Button.ts
TarunAdobe: If a consumer only provides the `aria-label` property for a button and doesn't set the label then this condition would make sure we won't override the `aria-label` value to `this.label` which is undefined in this case. 
Note: `sp-split-button` element currently does the same thing... we provide the `aria-label` to the buttons but not the `label`.

packages/button/src/Button.ts
TarunAdobe: right yes!

packages/button/src/Button.ts
Westbrook: Help me understand. How I currently understand the code as written:
```
if ("pending has changed") {
    if ("is pending") {
        if ("is not disabled") {
            // Apply pending label without regard for whether it is present or not
        }
    } else if ("is NOT pending AND NOT previously `undefined`, which happens at construction time when creating the `pending` property to begin with") {
        // Apply `label` or ''
    }
}
```
I like the idea of covering for when the author has applied `aria-label`, but I'm not sure that this code does. I _think_ doing so would look more like:
```
if ("pending has changed") {
    if ("is pending") {
        if ("is not disabled") {
            // Assume the applied `aria-label`, no matter who set it, is correct (this _might_ hit some issues if a Button was created as `pending` to start with, depending on code order)
            this.cachedLabel = this.getAttribute('aria-label');
            this.setAttribute('aria-label', this.pendingLabel);
        } else if (this.cachedLabel) {
            // If there was previously an applied `aria-label` re-apply it, or do nothing
            this.setAttribute('aria-label', this.cachedLabel);
        }
    }
}
```
This logic then duplicates some complexity with the change logic for `disabled`, which might benefit from being merged in, but is more annoying.

Either way, this got complex fast, it will be important to confirm the value of the `arial-label` attribute across these various states in the unit tests.
- pending===false => pending===true
- disabled===true && pending===false => disabled===true && pending===false
- disabled===false && pending===true => disabled===true && pending===true
- disabled===true && pending===true => disabled===false && pending===true
- All the same again with `aria-label` applied from the outside
- All the same again with `label` applied from the outside

There may be others, so think through the logic your creating and make sure the various workflows are fully managed so that the output you're expecting at the various points is specifically under test in the tests you add.

TarunAdobe: I've updated the logic and added some new tests. I think they cover all of the possibilities (?) However the logic is now a bit complicated so if you need help in understanding anything or if I missed some edge-cases do let me know.

TarunAdobe: @Westbrook  I've updated the code based on our discussion. Have a look whenever you get time

Westbrook (commented):
Last ones a nit but if we can get that back in, let's ship it!

packages/button/src/Button.ts
```diff
@@ -127,7 +146,6 @@ export class Button extends SizedMixin(ButtonBase, { noDefaultSize: true }) {
     public set quiet(quiet: boolean) {
         this.treatment = quiet ? 'outline' : 'fill';
     }
-
```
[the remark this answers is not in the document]
TarunAdobe: done hehe

Westbrook (approved):
Awesome! Thanks for driving this home. :shipit:

benjamind:
Amazing work @TarunAdobe thanks for making this a real thing!

Issue #2845, opened 2023-01-11
[Feat]: Add support for Express Workflow Icons


nickschaap:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Icons Workflow

### Description of the requested feature

We should add built-in support for [Spectrum Express Workflow Icons](https://github.com/adobe/spectrum-css-workflow-icons/tree/main/dist/spectrum-css-ccx-workflow-icons). In this way we can continue to close the gap on full support for the Express variant of the Spectrum design language.

### Implementation notes or ideas

It was suggested by @Westbrook that this can be achieved by adding a [conditional export](https://nodejs.org/api/packages.html#conditional-exports) to the icons-workflow [package.json](https://github.com/adobe/spectrum-web-components/blob/main/packages/icons-workflow/package.json) which will have to be dynamically generated pre-build. We will need some sort of resolver system for determining exactly which icons are exported when. In cases for when there is no replacement for the Spectrum Classic workflow icon in the Express icon set we should fallback to the Classic workflow icon. For now, no additional icons that do not have a parity with the Classic set will be exported.

castastrophe:
This has since been resolved so I'm going to close this one out.

PR #3169, opened 2023-05-02
feat: Add support for express icons

nickschaap:
## Description

This PR modifies the build process for the icons-workflow package to also process express icons and then in a post build step uses Node conditional exports to map exports to a particular icon based off the current export conditions.

## Related issue(s)

- #2845

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/package.json
+++ b/package.json
@@ -310,7 +310,9 @@
             "command": "yarn workspace @spectrum-web-components/icons-workflow build",
             "files": [
                 "packages/icons-workflow/bin/build.js",
-                "node_modules/@adobe/spectrum-css-workflow-icons/dist/18/**.svg"
+                "packages/icons-workflow/bin/postbuild.js",
+                "node_modules/@adobe/spectrum-css-workflow-icons/dist/18/**.svg",
+                "node_modules/@adobe/spectrum-css-ccx-workflow-icons/dist/18/**.svg"
             ],
             "output": [
                 "packages/icons-workflow/**/*.ts",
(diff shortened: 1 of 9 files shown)


nickschaap:
It looks like there is a discrepancy in the naming conventions for the `<sp-icon-fullscreen>` and `<sp-icon-fullscreen-exit>` icons. In the Spectrum set the SVG is named `FullScreen.svg` and in the Express set is named `Fullscreen.svg`

Westbrook:
Closing as stale. Happy to work with you or the team in yall generating an updated and complete version of this whenever you're ready!

AndreiBaicu26:
Coming back here after some reasearch
TL;DR I still could not make this work with conditional imports. 

As far as i understand from the [docs](https://nodejs.org/api/packages.html#conditional-exports) and what i read online, NodeJs only accepts those options as conditional imports "node-addons", "node"... and are only used for deciding how to import a file depending on the environment. 

One more issue I encountered was with the naming of the icons. Since we would want a component to only use the format `<sp-my-icon>`, typescript would not allow 2 different files to create the same definition (one in spectrum icons and one in express icons): 

```ts
declare global {
    interface HTMLElementTagNameMap {
        `sp-my-icon': MyIcon;
    }
}
```

I cound not really find another way without maybe linking somehow the icon imports and the theme itself and using dynamic imports :/ 

If you have any other tips or leads that I could follow please do let me know :D

PR #3177, opened 2023-05-03
chore: wind example project back to 0-version for unified publishing

hunterloftis:
## Description

In order to publish all the packages at 0.x, none of the packages can be 1.x


Westbrook (approved):
This looks like it does the trick in a test.

PR #3178, opened 2023-05-03
fix: generate react/picker and pass react TS checks

hunterloftis:
## Description

Publication currently fails at `yarn build:react` with two errors: missing picker/tsconfig and a TS error in Sidenav. This gets `yarn build:react` to pass.

I don't know why `PickerBase` is originally part of the exclusion list; someone who has some context about the react wrapper & how it works should probably validate this.


Westbrook:
With builds `Picker` as `PickerBase` which is incorrect... 😞 

```ts
import * as React from 'react';
import { createComponent } from '@lit-labs/react';
import type { EventName } from '@lit-labs/react';
import { PickerBase as SpPickerBase } from '@spectrum-web-components/picker';

export const PickerBase = createComponent({
    displayName: 'PickerBase',
    elementClass: SpPickerBase,
    react: React,
    tagName: 'sp-picker',
    events: {
        change: 'change' as EventName<Event>, // Announces that the `value` of the element has changed
        spOpened: 'sp-opened' as EventName<Event>, // Announces that the overlay has been opened
        spClosed: 'sp-closed' as EventName<Event>, // Announces that the overlay has been closed
    },
});

export type PickerBaseType = EventTarget & SpPickerBase;

```

Westbrook:
Maybe [this](https://github.com/adobe/spectrum-web-components/blob/main/packages/picker/src/Picker.ts#L65-L73) goes on the other class in this file?

hunterloftis:
@Westbrook updated!

jianliao (approved):

Westbrook (approved):

PR #3191, opened 2023-05-09
refactor: prepare label/tooltip styles for Overlay refactor

Westbrook:
## Description
Can we make buttons smarter about consuming Tooltips in advance of the new Overlay API?

## How has this been tested?
-   [ ] _Test case 1_
    1. Go [here](https://button-slots--spectrum-web-components.netlify.app/storybook/?path=/story/tooltip--self-managed-icon-only)
    2. Ensure the top button (with Tooltip) and bottom button (without Tooltip) display the same visually
-   [ ] _Test case 2_
    1. Go [here](https://3f739bdefe5cf6e781e59c351b81bef0--spectrum-web-components.netlify.app/review/)
    2. See that the VRTs only fail on the addition of a new story.

## Types of changes
-   [x] Refactor

## Checklist

### DIFF
--- a/tools/shared/src/observe-slot-text.ts
+++ b/tools/shared/src/observe-slot-text.ts
@@ -31,8 +31,13 @@ export interface SlotTextObservingInterface {
 
 export function ObserveSlotText<T extends Constructor<ReactiveElement>>(
     constructor: T,
-    slotName?: string
+    slotName?: string,
+    excludedSelectors: string[] = []
 ): T & Constructor<SlotTextObservingInterface> {
+    const notExcluded = (el: HTMLElement) => (selector: string) => {
+        return el.matches(selector);
+    };
+
     class SlotTextObservingElement
         extends constructor
         implements SlotTextObservingInterface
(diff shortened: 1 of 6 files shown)


tools/shared/src/observe-slot-text.ts
```diff
@@ -68,6 +69,12 @@ export function ObserveSlotText<T extends Constructor<ReactiveElement>>(
             const assignedNodes = [...this[assignedNodesList]].filter(
                 (node) => {
                     if ((node as HTMLElement).tagName) {
+                        if (exclude?.length) {
+                            const shouldExclude = exclude.some((selector) =>
+                                (node as HTMLElement).matches(selector)
+                            );
+                            return !shouldExclude;
+                        }
```
hunterloftis, line 77: Not a blocker, but I'm curious if defaulting `exclude` to `[]` and then simplifying the branching here would be equally-fast, eg:

```suggestion
                        return exclude.every((selector) => !(node as HTMLElement).matches(selector));
```

tools/shared/src/observe-slot-text.ts
```diff
@@ -81,6 +88,12 @@ export function ObserveSlotText<T extends Constructor<ReactiveElement>>(
                 const { childNodes } = this;
                 const textNodes = [...childNodes].filter((node) => {
                     if ((node as HTMLElement).tagName) {
+                        if (exclude?.length) {
+                            const shouldExclude = exclude.some((selector) =>
+                                (node as HTMLElement).matches(selector)
+                            );
+                            return !shouldExclude;
+                        }
```
hunterloftis, line 96: Consider pulling this (plus a few lines above/below) into a function that could be used like `textNodes = [...childNodes].filter(notExcluded)`?

hunterloftis (approved):

tools/shared/src/observe-slot-text.ts
Westbrook: I worried about a conditional argument with a default, but that's a good idea!

tools/shared/src/observe-slot-text.ts
Westbrook: Yeah, then it can be used in both tests. Good call!

PR #3203, opened 2023-05-16
Fix(overlay): stop the tab trapping if shadow root is not found

deepthi-majji:
In the code, `trappingInited` is set to true in the `initTabTrapping` method but when shadow root is not found on the `body` we should ideally stop the trapping. If we don't set this flag to false, it will result in an exception [here](https://github.com/adobe/spectrum-web-components/blob/main/packages/overlay/src/overlay-stack.ts#L189) in the above scenario. 

The overlay code is being revamped and this bug fix might not make sense in the new code but this will be helpful for integrating SWC with UXP. 

## Related issue(s)

-

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/overlay/src/overlay-stack.ts
+++ b/packages/overlay/src/overlay-stack.ts
@@ -90,8 +90,9 @@ export class OverlayStack {
             return;
         }
         this.document.body.attachShadow({ mode: 'open' });
-        /* c8 ignore next 3 */
+        /* c8 ignore next 4 */
         if (!this.document.body.shadowRoot) {
+            this.trappingInited = false;
             return;
         }
         const root = this.document.body.shadowRoot as ShadowRoot;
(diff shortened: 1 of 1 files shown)


packages/overlay/src/overlay-stack.ts
```diff
@@ -90,8 +90,9 @@ export class OverlayStack {
             return;
         }
         this.document.body.attachShadow({ mode: 'open' });
-        /* c8 ignore next 3 */
+        /* c8 ignore next 4 */
         if (!this.document.body.shadowRoot) {
+            this.trappingInited = false;
```
Westbrook, line 95: Instead, can we move line 86 to after this block? Seems like a reasonable change, even if short lived.

deepthi-majji: yes makes sense.

Westbrook (approved):
LGTM. Thanks for catching this!

PR #3223, opened 2023-05-22
chore: use latest Lit syntax and labs tools

Westbrook:
## Description
Fast forward all non-breaking changes in support of the move to Lit@3.0

## Types of changes

## Checklist

### DIFF
--- a/packages/table/src/Table.ts
+++ b/packages/table/src/Table.ts
@@ -35,11 +35,10 @@ import {
     virtualizerRef,
 } from '@lit-labs/virtualizer/virtualize.js';
 import { Virtualizer } from '@lit-labs/virtualizer/Virtualizer.js';
-
-interface Range {
-    first: number;
-    last: number;
-}
+import {
+    RangeChangedEvent,
+    VisibilityChangedEvent,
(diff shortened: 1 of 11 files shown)


packages/table/src/Table.ts
```diff
@@ -475,6 +461,17 @@ export class Table extends SizedMixin(SpectrumElement, {
                     );
                 }
             );
+            this.tableBody.addEventListener(
```
hunterloftis, line 464: Just for understanding - why is this event re-dispatched instead of bubbling?

hunterloftis (dismissed):

packages/table/src/Table.ts
Westbrook: This "normalizes" to what Lit is doing in their element: https://github.com/lit/lit/blob/main/packages/labs/virtualizer/src/events.ts Previously, they dispatched it bubbling, but not in the new version.

By default, we create the `sp-table-body` where this event is actually dispatched, which makes listening on it precarious, so we mimic the event on the `sp-table` that the consuming dev does own to make it as much like consuming a `lit-virtualizer` directly.

PR #3225, opened 2023-05-22
updating paths to be package imports

inasirhussain:
We are from UXP team, We are working on [swc-wrappers ](https://git.corp.adobe.com/torq/swc-uxp-wrappers/, ) UXP.
We are facing issue while writing UXP wrapper for DialogBase, DialogWrapper, MenuGroup and Table.

Please find below findings for DialogBase Example.
```
import {
    CSSResultArray,
    html,
    PropertyValues,
    SpectrumElement,
    TemplateResult,
} from '@spectrum-web-components/base';
import { property } from '@spectrum-web-components/base/src/decorators.js';

import '@spectrum-web-components/underlay/sp-underlay.js';
import '@spectrum-web-components/button/sp-button.js';

import '../sp-dialog.js';  -> sp-dialog.js importing from local path
```

In DialogBase.ts code is importing sp-dialog.js from local path. This is causing reregistration sp-dialog us while implementing UXP Wrapper for DialogBase.
'CustomElementRegistry': the name "sp-dialog" has already been used with this registry
    at eval (webpack-internal:///../../node_modules/@swc-uxp-internal/dialog/sp-dialog.dev.js:5:16)
    at ../../node_modules/@swc-uxp-internal/dialog/sp-dialog.dev.js (file:///Users/asshukla/swc-wrappers/projects/swc-starter-webpack/dist/main.bundle.js:1629:1)
    at __webpack_require__ (file:///Users/asshukla/swc-wrappers/projects/swc-starter-webpack/dist/main.bundle.js:2673:42)
    at eval (webpack-internal:///../../node_modules/@swc-uxp-internal/dialog/src/DialogBase.dev.js:9:75)
    at ../../node_modules/@swc-uxp-internal/dialog/src/DialogBase.dev.js (file:///Users/asshukla/swc-wrappers/projects/swc-starter-webpack/dist/main.bundle.js:1649:1)
    at __webpack_require__ (file:///Users/asshukla/swc-wrappers/projects/swc-starter-webpack/dist/main.bundle.js:2673:42)
    at eval (webpack-internal:///../../packages/dialog/src/DialogBase.js:5:100)
    at ../../packages/dialog/src/DialogBase.js (file:///Users/asshukla/swc-wrappers/projects/swc-starter-webpack/dist/main.bundle.js:2439:1)
    at __webpack_require__ (file:///Users/asshukla/swc-wrappers/projects/swc-starter-webpack/dist/main.bundle.js:2673:42)
    at eval (webpack-internal:///../../packages/dialog/sp-dialog-base.js:2:76)

Instead of internal dependencies we tried in DialogBase.ts and rest of the component to make package imports, things are working fine.
import '@spectrum-web-components/dialog/sp-dialog.js'

### DIFF
--- a/packages/dialog/src/DialogBase.ts
+++ b/packages/dialog/src/DialogBase.ts
@@ -25,7 +25,7 @@ import '@spectrum-web-components/button/sp-button.js';
 import '@spectrum-web-components/dialog/sp-dialog.js'
 import modalWrapperStyles from '@spectrum-web-components/modal/src/modal-wrapper.css.js';
 import modalStyles from '@spectrum-web-components/modal/src/modal.css.js';
-import { Dialog } from  '@spectrum-web-components/dialog/src/Dialog.js'
+import { Dialog } from './Dialog.js';
 import { FocusVisiblePolyfillMixin } from '@spectrum-web-components/shared';
 import { firstFocusableIn } from '@spectrum-web-components/shared/src/first-focusable-in.js';
 
--- a/packages/dialog/src/DialogBase.ts
+++ b/packages/dialog/src/DialogBase.ts
@@ -22,10 +22,10 @@ import { property } from '@spectrum-web-components/base/src/decorators.js';
 import '@spectrum-web-components/underlay/sp-underlay.js';
 import '@spectrum-web-components/button/sp-button.js';
 
-import '../sp-dialog.js';
+import '@spectrum-web-components/dialog/sp-dialog.js'
 import modalWrapperStyles from '@spectrum-web-components/modal/src/modal-wrapper.css.js';
 import modalStyles from '@spectrum-web-components/modal/src/modal.css.js';
-import { Dialog } from './Dialog.js';
+import { Dialog } from  '@spectrum-web-components/dialog/src/Dialog.js'
 import { FocusVisiblePolyfillMixin } from '@spectrum-web-components/shared';
 import { firstFocusableIn } from '@spectrum-web-components/shared/src/first-focusable-in.js';
 
--- a/packages/dialog/src/DialogBase.ts
+++ b/packages/dialog/src/DialogBase.ts
@@ -22,6 +22,7 @@ import { property } from '@spectrum-web-components/base/src/decorators.js';
 import '@spectrum-web-components/underlay/sp-underlay.js';
 import '@spectrum-web-components/button/sp-button.js';
 
+// Leveraged in build systems that use aliasing to prevent multiple registrations: https://github.com/adobe/spectrum-web-components/pull/3225
 import '@spectrum-web-components/dialog/sp-dialog.js'
 import modalWrapperStyles from '@spectrum-web-components/modal/src/modal-wrapper.css.js';
 import modalStyles from '@spectrum-web-components/modal/src/modal.css.js';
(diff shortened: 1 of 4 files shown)


packages/dialog/src/DialogBase.ts
```diff
@@ -22,10 +22,10 @@ import { property } from '@spectrum-web-components/base/src/decorators.js';
 import '@spectrum-web-components/underlay/sp-underlay.js';
 import '@spectrum-web-components/button/sp-button.js';
 
-import '../sp-dialog.js';
+import '@spectrum-web-components/dialog/sp-dialog.js'
 import modalWrapperStyles from '@spectrum-web-components/modal/src/modal-wrapper.css.js';
 import modalStyles from '@spectrum-web-components/modal/src/modal.css.js';
-import { Dialog } from './Dialog.js';
+import { Dialog } from  '@spectrum-web-components/dialog/src/Dialog.js'
```
Westbrook, line 28: Let's stick to moving registering imports, only, to bare module specifiers, for now.

packages/dialog/src/DialogBase.ts
```diff
@@ -22,10 +22,10 @@ import { property } from '@spectrum-web-components/base/src/decorators.js';
 import '@spectrum-web-components/underlay/sp-underlay.js';
 import '@spectrum-web-components/button/sp-button.js';
 
-import '../sp-dialog.js';
+import '@spectrum-web-components/dialog/sp-dialog.js'
```
Westbrook, line 25: Please comment the use case (and a little about why it'll actually work, to save future confusion) for each of the changes registering paths.

Westbrook (changes requested):
Looks reasonable, let's reduce the path changes to just registering imports and include some comments and it should be good to go!

inasirhussain:
Use case : As part of leveraging support of SWC in UXP we are writing a wrapper layer which will mostly be taking care of filling the gaps in UXP platform from CSS properties perspective.

We are extending from SWC's component's source file and appending our styles to it and eventually registering the tag with the same name with our extended class.
```
import { MenuGroup } from '@swc-uxp-internal/menu/src/MenuGroup.js';

import styles from './uxp-menu-group.css.js';

class UxpMenuGroup extends MenuGroup {
    static get styles() {
        // We are combining our styles to make all super class styles available along with the transitive dependendent classes styles.
        return [super.styles, styles];
    }
}

export { UxpMenuGroup as MenuGroup };

```
```
import { MenuGroup } from './src/MenuGroup.js';

customElements.define('sp-menu-group', MenuGroup);
```

With above example in picture, MenuGroup.ts have internal dependency specified on sp-menu and import causes double registration since we alias `'@spectrum-web-components/menu': '@swc-uxp-wrappers/menu'` 
With this PR changes, we are moving from internal dependency to package imports which will eventually help with aliasing using webpack, avoiding double registration.

packages/dialog/src/DialogBase.ts
Westbrook: Can you add something small inline. Something like `// Leveraged in build systems that use aliasing to prevent multiple registrations: https://github.com/adobe/spectrum-web-components/pull/3225` or so?

inasirhussain: Updated.

Issue #3186, opened 2023-05-09
[Bug]: Slider performance issues on mobile, low-end devices or complex apps.


moisadoru:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-slider

### Expected behavior

Dragging the slider should be smooth, even on low-end devices.

### Actual behavior

When using the slider on mobile devices or low-end laptops, dragging the handle lags, as seen in the attached video. Setting the slider handle's absolute position via left/right CSS seems to cause style recalculations and layout changes.

### Screenshots

https://user-images.githubusercontent.com/109328/237030810-945f160d-f946-439f-8f34-a579e25b0ea7.mp4

### What browsers are you seeing the problem in?

Chrome, Microsoft Edge

### How can we reproduce this issue?

1. If using a powerful machine (eg. M1), configure hardware concurrency to 4 and 4x CPU throttling to simulate a low-end/mobile device, or use a real consumer device (phone, non-pro laptop)
2. Go to any page that renders an SWC slider, eg. https://opensource.adobe.com/spectrum-web-components/components/slider
3. Drag the slider
4. Observe the slider handle lagging behind the cursor

### Logs taken while reproducing problem

[Trace-20230509T101433.json.zip](https://github.com/adobe/spectrum-web-components/files/11428706/Trace-20230509T101433.json.zip)

Westbrook:
If you continue to see issues with this post #3243 please open a new issue with updated reproduction information.

PR #3226, opened 2023-05-22
[CCEX-45622] fix: small performance improvements

AndreiBaicu26:
When CPU is throttled to 4x and Hardware concurrency is set to 4 the sp-slider is laggy when moving the slider faster. 

## Description
* Cancel updates on pointerMove is the new position of the handle does not match any step. 
* Use transform instead of left positioning 

## Related issue(s)
 https://github.com/adobe/spectrum-web-components/issues/3186

-

## Motivation and context
Slightly improve the performance for sp-slider

## How has this been tested?
Manual testing for multiple types of sliders: 
* One handle slider
* Two handles slider
* Gradient slider
* Default slider
* Tried with different types of tickStep 1, 1.5, 1.15

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/slider/src/HandleController.ts
+++ b/packages/slider/src/HandleController.ts
@@ -361,6 +361,41 @@ export class HandleController implements Controller {
         this.requestUpdate();
     }
 
+    // Check if the new handle position matches any step value
+    private _hasHandleLocationChanged(
+        input: InputWithModel,
+        handlePosition: number,
+        model: ModelValue
+    ): boolean {
+        // Calculate the number of steps from the minimum value
+        const handleStep = Math.round(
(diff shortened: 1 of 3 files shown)


Westbrook:
@AndreiBaicu26 I _think_ [this](https://github.com/adobe/spectrum-web-components/pull/3243) is more what we'd want to see here. This solution ensure that all of the component using the feature that Slider is leveraging are bound to the animation frame, rather than forcing a unique limiting solution into point of consumption.

Can you help take a look at that branch and see if it's getting the sort of results you were hoping for with this change?

PR #3229, opened 2023-05-22
docs: prune release process

Westbrook:
## Description
- `git clean -dfx` removes the need to `rm -rf` anything.
- most actions work even outside of the Node version requirements, but make sure you're on something expected before running the release.


jianliao (approved):
LGTM

Issue #2254, opened 2022-04-27
[Feat]: Move checkmark of selected item in Picker/Menu to left side 


spdev3000:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Picker, Menu

### Description of the requested feature

Pls align Spectrum css of Picker/Menu of the selected state (checkmark) so that the checkmark is on the _left_ of menu item. see https://spectrum.adobe.com/page/picker/ for reference

### Mockups or screenshots

![](https://spectrum.adobe.com/static/images1x/picker_hero_1638572223403.png)

PR #3254, opened 2023-05-30
feat(menu): convert to core tokens

### LINKED ISSUE #2984: [Feat]: For menu-items, we have the option of specifying a shortcut by using the "value" slot. the gap between the menu-item text and the value slot is very less.
### Code of conduct

### Impacted component(s)

Venus

For menu-items, we have the option of specifying a shortcut by using the "value" slot. the gap between the menu-item text and the value slot is very less.
It would be good to have a configuration where we can specify the minimum gap between text and shortcut label in menu-item. 

### Mockups or screenshots

image

_No response_

najikahalsema:
## Description

Convert to core tokens. Things that still need to be done/looked at:

* [x] checkmark margin via -spectrum-text-to-visual-100 is 8 when it should be 10 to keep the selected item from narrowing
* [x] Menu divider- should it touch the sides in this new version? 
* [x] Menu margins for popover is pretty narrow— but putting a popover on the styling for the css makes it look nicer. can we get clarification on whether or not the sp-menu element should have margins?
* [x] Focus/hover state isn’t getting applied to the menu item background. I think this is due to the “custom vars” 
* [x] add "Sizes" stories to the Menu package
* [x] pass `this.size` down from `Picker.ts` into the `<sp-menu>` child in its shadow DOM
* [x] add a set of "Sizes Open" stories to Picker so that the above change is included in VRTs
* [x] add a set of "Sizes Open" stories to Action Menu and Split Button
* [x] update the VRT cache with the updated and new stories
* [x] get a stable release from the CSS team

## Related issue(s)

- fixes https://github.com/adobe/spectrum-web-components/issues/2254 
- fixes #2984

## Motivation and context

upgrading to core tkoens :) 

## How has this been tested?

-   [ ] _Test case 1_
    1. Go [here](https://d95f1d9aab4dd919907c3cee1e51bd9b--spectrum-web-components.netlify.app/review/#ActionMenuStories/selects.png)
    2. Compare the changes between the old and new versions. Note the checkmark is now on the left, and that the focus-visible is... well, visible. 

-   [ ] _Test case 2_
    1. Go [here](https://9905080f40f3bb038d2df931c70e6c4c--spectrum-web-components.netlify.app/review/#ActionMenuStories/selects.png) and see that the WHCM doesn't have that strange black line when a menu item is focused 
    2. Compare the changes between the old and new versions
   
-   [ ] _Test case 3_
    1. Go [here](https://d95f1d9aab4dd919907c3cee1e51bd9b--spectrum-web-components.netlify.app/review/#MenuStories/MenuGroupSelectsMultiple.png) 
    2. See that the menu divider is visible 
    
- [ ] _Test Case 4_
     1. Go [here](https://d95f1d9aab4dd919907c3cee1e51bd9b--spectrum-web-components.netlify.app/review/#SubmenuStories/submenu.png) 
     2. Confirm that the space after the chevron is equal to the previous version
## Types of changes

## Checklist

## Best practices


packages/menu/src/Menu.ts
```diff
@@ -95,6 +96,10 @@ export class Menu extends SpectrumElement {
     @query('slot:not([name])')
     public menuSlot!: HTMLSlotElement;
 
+    public override get focusElement(): HTMLElement {
+        return this.menuSlot;
```
Westbrook, line 100: This breaks the focus passing into Pickers and Action Menu's, etc. (see [this VRT](https://9905080f40f3bb038d2df931c70e6c4c--spectrum-web-components.netlify.app/review/#ActionMenuStories/selects.png)). Let's chat about what the goal was for this in Office Hours tomorrow and see if we can't find a better path forward here!

najikahalsema: I was trying to add a focusElement getter because I was getting an error in the console, which I think you pointed out some time ago. I wasn't sure exactly how to accomplish it though 😬

Westbrook:
FYI: this did a weird thing: https://9905080f40f3bb038d2df931c70e6c4c--spectrum-web-components.netlify.app/review/#ActionMenuStories/selects.png 

image

packages/menu/src/menu-group.css
```diff
@@ -19,6 +19,35 @@ governing permissions and limitations under the License.
...
+}
+
+:host(:last-child) sp-menu {
+    margin-bottom: 0;
+}
+
+:host(:first-child) .header[hidden] + sp-menu {
+    margin-top: 0;
+}
+
```
Westbrook: I've learned from experience that it's much harder to remove a workaround if you don't mark where it ends
```suggestion
/**
 * End workaround for https://github.com/adobe/spectrum-css/issues/794
 */
```

packages/menu/src/spectrum-config.js
```diff
@@ -62,10 +79,38 @@ const config = {
...
+                ),
                 converter.classToClass('spectrum-Menu-checkmark', 'checkmark'),
+                converter.classToClass(
+                    'spectrum-Menu-checkmark--withAdjacentIcon',
+                    'checkmark--withAdjacentIcon'
+                ),
+                converter.classToAttribute(
+                    'spectrum-Menu-item--drillIn',
+                    'has-submenu'
+                ),
```
Westbrook, line 95: If they call these `drillIn` and we call them `submenu`, will there be any naming issues when we move to support the "expanded menu"?

Westbrook (commented):
https://dd8f58a27a69ec29ce5a86faac0d3141--spectrum-web-components.netlify.app/review/#SubmenuStories/submenu.png

"Submenu chevrons" are getting `margin-inline-end` from what I think is a rule meant for "collapsable chevrons" in the Spectrum CSS source. Can you confirm with their team that this is being applied appropriately.

Current:
image

Without the added margin:
image

packages/menu/src/spectrum-config.js
najikahalsema: I was wondering that at first, as well, but I made the decision not to bring it up with the team 🤔 This is because it wouldn't affect any of the other work that needs to happen down the line since it's really just for the submenu and that's it. The other work that needs to happen is from the collapsible stuff, which uses a different class name, and the descriptive text below the main menuitem label.

Westbrook: Sounds good!

Westbrook (commented):
Looking real nice to me! Looking good to you?

Final practical steps: 🤞🏼
- [ ] get a stable release from the CSS team
- [x] add ["Sizes" stories](https://menu-core-tokens--spectrum-web-components.netlify.app/storybook/?path=/story/picker-sizes--s) to the Menu package
- [ ] pass `this.size` down from `Picker.ts` into the `<sp-menu>` child in its shadow DOM
- [ ] add a set of "Sizes Open" stories to Picker so that the above change is included in VRTs
- [ ] (possible overkill, but nice to have for sure) add a set of "Sizes Open" stories to Action Menu and Split Button
- [ ] update the VRT cache with the updated and new stories

Very exciting!!

Westbrook (approved):
LGTM!!! 🚀

PR #3267, opened 2023-06-02
chore: use latest axe-core

### LINKED ISSUE #3266: [Bug][a11y]: indeterminate checkbox input use indeterminate property rather than aria-checked="mixed"
### Code of conduct

### Impacted component(s)

Checkbox

### Expected behavior

Automated testing tools like aXe throw a conformance error when an `aria-checked` attribute is assigned to an HTMLInputElement element, because we should be using the native HTML `checked` attribute or the `indeterminate` property.

We should set the [HTMLInputElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement) object's `indeterminate` property via JavaScript (it cannot be set using an HTML attribute).

### Actual behavior

Automated testing tools like aXe throw a conformance error when an `aria-checked` attribute is assigned to an  [HTMLInputElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement)  element, because we should be using the native HTML `checked` attribute or the `indeterminate` property.

### Screenshots

_No response_

1. Go to https://opensource.adobe.com/spectrum-web-components/components/checkbox/
2. Open aXe DevTools Extension in the Developer Tools panel (I can be acquired here: https://docs.deque.com/devtools-html/4.0.0/en/devtools-extension#installation)
3. Run "Scan ALL my page"
4. Note that  aXe DevTools reports 7 "Elements must only use allowed ARIA attributes" errors for the indeterminate checkboxes that include `aria-checked="mixed"` attribute.

_No response_

_No response_

Westbrook:
## Description
- update axe-core version and respond to test failures
- correct `indeterminate` values for Checkboxes
- correct role/aria attribute application in Menu Groups
- apply `valuenow`, thought I'm not 100% sure why, to SplitView

## Related issue(s)
- fixes #3266

## How has this been tested?
-   [ ] _Test case 1_
    1. Go [here](https://app.circleci.com/pipelines/github/adobe/spectrum-web-components/16000/workflows/27506cb2-99dc-49d9-9b54-22cf02164c6c/jobs/336229/tests#failed-test-0)
    2. See test failures
-   [ ] _Test case 2_
    1. Go [here](https://app.circleci.com/pipelines/github/adobe/spectrum-web-components/16002/workflows/8c63a347-64f5-4fc8-89c6-fc9ec94600a8/jobs/336327)
    2. See them not fail

## Types of changes

## Checklist

### DIFF
--- a/packages/checkbox/src/Checkbox.ts
+++ b/packages/checkbox/src/Checkbox.ts
@@ -105,6 +105,11 @@ export class Checkbox extends SizedMixin(CheckboxBase) {
         return [checkboxStyles, checkmarkSmallStyles, dashSmallStyles];
     }
 
+    public override handleChange(): void {
+        this.indeterminate = false;
+        super.handleChange();
+    }
+
     protected override render(): TemplateResult {
         return html`
             ${super.render()}
@@ -126,11 +131,7 @@ export class Checkbox extends SizedMixin(CheckboxBase) {
             }
         }
         if (changes.has('indeterminate')) {
-            if (this.indeterminate) {
-                this.inputElement.setAttribute('aria-checked', 'mixed');
-            } else {
-                this.inputElement.removeAttribute('aria-checked');
-            }
+            this.inputElement.indeterminate = this.indeterminate;
         }
     }
 }
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ executors:
 parameters:
     current_golden_images_hash:
         type: string
-        default: 80a8f8e80ec5375753c239b5b96151079c6ed793
+        default: 12a0924e2cb935952dd29ba42145901d7269bc89
     wireit_cache_name:
         type: string
         default: wireit
--- a/packages/checkbox/stories/checkbox.stories.ts
+++ b/packages/checkbox/stories/checkbox.stories.ts
@@ -32,7 +32,13 @@ export const readonly = (): TemplateResult => {
 
 export const Indeterminate = (): TemplateResult => {
     return html`
-        <sp-checkbox indeterminate>Checkbox</sp-checkbox>
+        <sp-checkbox indeterminate>
+            Checkbox, indeterminate, not checked
+        </sp-checkbox>
+        <br />
+        <sp-checkbox indeterminate checked>
+            Checkbox, indeterminate, checked
+        </sp-checkbox>
     `;
 };
 
--- a/packages/checkbox/test/checkbox.test.ts
+++ b/packages/checkbox/test/checkbox.test.ts
@@ -204,4 +204,36 @@ describe('Checkbox', () => {
 
         expect(el.checked).to.be.true;
     });
+
+    it('`indeterminate, checked` becomes `not checked` on click', async () => {
+        const el = await fixture<Checkbox>(html`
+            <sp-checkbox checked .indeterminate=${true}>
+                indeterminate, checked
+            </sp-checkbox>
+        `);
+        expect(el.checked).to.be.true;
+        expect(el.indeterminate).to.be.true;
+
+        el.click();
+        await elementUpdated(el);
+
+        expect(el.checked).to.be.false;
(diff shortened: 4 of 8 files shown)


najikahalsema (approved):
Looks good !

packages/checkbox/src/Checkbox.ts
```diff
@@ -126,11 +126,7 @@ export class Checkbox extends SizedMixin(CheckboxBase) {
             }
         }
         if (changes.has('indeterminate')) {
-            if (this.indeterminate) {
-                this.inputElement.setAttribute('aria-checked', 'mixed');
-            } else {
-                this.inputElement.removeAttribute('aria-checked');
-            }
+            this.inputElement.indeterminate = this.indeterminate;
```
majornista, line 129: I think this needs to be a bit more robust. Otherwise, an indeterminate checkbox will announce the checked/unchecked state on click, even while `this.indeterminate` remains `true`.

Here is a suggestion:

```ts
        if (changes.has('indeterminate') || this.indeterminate) {
            this.inputElement.indeterminate = this.indeterminate;
            if (this.indeterminate) {
                this.checked = false;
            } else {
                this.inputElement.checked = this.checked;
            }
        }
```

majornista (changes requested):

packages/checkbox/src/Checkbox.ts
majornista: See https://github.com/adobe/spectrum-web-components/pull/3268

Westbrook: Agree that there's something more here, but the interaction with `this.checked` begs some more thought on other things that might not have been covered previously.

1. Is an `indeterminate` Checkbox also `checked === false`?
2. When an `indeterminate` Checkbox is clicked, should it move to a state and remove `indeterminate`?
3. When an `indeterminate` checkbox is clicked, if it changes state, should it _always_ go to `checked === true` or should it go to the opposite of the last state before it became `interderminate`?
4. The suggestion above outlines a `checked === true` Checkbox that is moved to `indeterminate` and then later leaves `indeterminate` (without use interaction) will end up as `checked === false`. Is that correct?

Westbrook: Looking more closely at a native `<input type="checkbox" />` I don't think we want `indeterminate` to effect `checked`. https://codepen.io/Westbrook/pen/poxMVWE

Here we see that whatever the value of `checked`, clicking and `indeterminate` `<input type="checkbox" />` will switch the value of `checked` and set `indeterminate = false`.

Is there some non-platform advice beyond that (there often is, but this is clearly a depth on this pattern deeper that I've researched before) to point to doing anything differently here?

Westbrook: For now, I've added some tests and expanded the story for `indeterminate` to back us on up this stance. We can chat more next week, if there's more info we should take into account.

majornista: If take the stand that the value of the input's checked state takes precedence, we should remove the indeterminate style when `checked === true`.

Westbrook: Value over indeterminate seems to be the order of operations for native Checkboxes, so it stands that we should do the same. I've included the removal of `indeterminate` on any change (whether to `true` or to `false`) of `checked`. Let me know if you think that covers the intersection of realities here enough to move forward in an accessible way.

majornista: This seems fine now.

majornista: Adding this makes the difference:
https://github.com/adobe/spectrum-web-components/pull/3267/files#diff-f726e9e9adb1c2e8f11bc0649f910bf51ec08b9fc4ab3e3e528a46a78e8f3fc6R108-R112

Westbrook: Thanks for helping us clear that up!

PR #3268, opened 2023-06-02
fix(checkbox): [a11y] use indeterminate property rather than aria-checked="mixed" #3266

majornista:
## Description

Automated testing tools like aXe throw a conformance error when an aria-checked attribute is assigned to an HTMLInputElement element, because we should be using the native HTML checked attribute or the indeterminate property.

We should set the [HTMLInputElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement) object's indeterminate property via JavaScript (it cannot be set using an HTML attribute).

## Related issue(s)

#3266 

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

-   [ x I have reviewed at the Accessibility Practices for this feature, see: [Aria Practices](https://www.w3.org/TR/wai-aria-practices/)

## Best practices

### DIFF
--- a/packages/checkbox/src/Checkbox.ts
+++ b/packages/checkbox/src/Checkbox.ts
@@ -112,7 +112,7 @@ export class Checkbox extends SizedMixin(CheckboxBase) {
                 ${checkmarkIcon[this.size as DefaultElementSize]}
                 ${dashIcon[this.size as DefaultElementSize]}
             </span>
-            <label id="label"><slot></slot></label>
+            <label id="label" for="input"><slot></slot></label>
         `;
     }
 
@@ -125,11 +125,12 @@ export class Checkbox extends SizedMixin(CheckboxBase) {
                 this.inputElement.removeAttribute('aria-invalid');
             }
         }
-        if (changes.has('indeterminate')) {
+        if (changes.has('indeterminate') || this.indeterminate) {
+            this.inputElement.indeterminate = this.indeterminate;
             if (this.indeterminate) {
-                this.inputElement.setAttribute('aria-checked', 'mixed');
+                this.checked = false;
             } else {
-                this.inputElement.removeAttribute('aria-checked');
+                this.inputElement.checked = this.checked;
             }
         }
     }
--- a/packages/checkbox/src/Checkbox.ts
+++ b/packages/checkbox/src/Checkbox.ts
@@ -105,6 +105,11 @@ export class Checkbox extends SizedMixin(CheckboxBase) {
         return [checkboxStyles, checkmarkSmallStyles, dashSmallStyles];
     }
 
+    public override handleChange(): void {
+        this.indeterminate = false;
+        super.handleChange();
+    }
+
     protected override render(): TemplateResult {
         return html`
             ${super.render()}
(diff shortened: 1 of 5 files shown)


packages/checkbox/src/Checkbox.ts
```diff
@@ -112,7 +112,7 @@ export class Checkbox extends SizedMixin(CheckboxBase) {
                 ${checkmarkIcon[this.size as DefaultElementSize]}
                 ${dashIcon[this.size as DefaultElementSize]}
             </span>
-            <label id="label"><slot></slot></label>
+            <label id="label" for="input"><slot></slot></label>
```
Westbrook, line 115: This `<input>` already leverages `aria-labelledby="label"`, is one better than the other?

packages/checkbox/src/Checkbox.ts
```diff
@@ -125,11 +125,12 @@ export class Checkbox extends SizedMixin(CheckboxBase) {
                 this.inputElement.removeAttribute('aria-invalid');
             }
         }
-        if (changes.has('indeterminate')) {
+        if (changes.has('indeterminate') || this.indeterminate) {
```
Westbrook, line 128: There's further question in https://github.com/adobe/spectrum-web-components/pull/3267#discussion_r1214795811 before we can move forward with this...

packages/checkbox/test/checkbox.test.ts
```diff
@@ -118,6 +137,67 @@ describe('Checkbox', () => {
...
         await expect(el).to.be.accessible();
+
+        const labelEl = labelForCheckbox(el);
+        const inputEl = inputForCheckbox(el);
+
+        expect(labelEl.getAttribute('for')).to.equal(inputEl.id);
+        expect(inputEl).to.have.attribute('aria-invalid', 'true');
+    });
+
+    it.only('is `indeterminate` checkbox accessibly', async () => {
```
Westbrook, line 148: Leverages `.only`.

packages/checkbox/test/checkbox.test.ts
```diff
@@ -28,6 +28,11 @@ function inputForCheckbox(checkbox: Checkbox): HTMLInputElement {
     return checkbox.shadowRoot.querySelector('#input') as HTMLInputElement;
 }
 
+function labelForCheckbox(checkbox: Checkbox): HTMLLabelElement {
```
Westbrook, line 31: This approach expands on reaching into the Shadow DOM from a very early history on the library, would it be less invasive to use the accessibility snap shot were we can ensure the relationships between these elements, rather than just the presence of attributes? Checkout https://github.com/adobe/spectrum-web-components/blob/main/packages/radio/test/radio-group.test.ts#L301 for an example of this.

packages/checkbox/src/Checkbox.ts
majornista: `for` makes it so that clicking the label activates focuses and clicks the input, which provides a greater affordance. It's a bit odd to use CSS to make the input the size of the label.

Westbrook: Yeah, Spectrum is gonna Spectrum, I guess. But, there is something nice about how the `<input>` prevents double clicks from selecting text without preventing text selection in these controls.

`for` makes sense, and will be a good coverage in case the upstream CSS moves away from that pattern in the future!

Issue #3278, opened 2023-06-05
[Bug]: sp-color-slider and sp-color-wheel fail to render focused style when first receiving focus


majornista:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

ColorSlider, ColorWheel

### Expected behavior

`<sp-color-slider>` and `<sp-color-wheel>` should render the `focused` style when they first receive focus, tabbing into the document of iframe.

### Actual behavior

For ColorSlider:

1. Open https://opensource.adobe.com/spectrum-web-components/components/color-slider/#default 
2. Hover over and mouse click to set focus on the "#" sign next to the heading reading "Default."
3. Press Tab key once to set focus to the "#" link with href="#default".
4. Then press Tab a second time to move focus to the ColorSlider example rendered below.
5. Notice that the ColorSlider handle does not change, by growing in size, to display focus.
6. If you press Esc or an Arrow key to change the value, the focused style will render.

Repeat for ColorWheel:
1. Open https://opensource.adobe.com/spectrum-web-components/components/color-wheel/#example 
2. Hover over and mouse click to set focus on the "#" sign next to the heading reading "Example."
3. Press Tab key once to set focus to the "#" link with href="#default".
4. Then press Tab a second time to move focus to the ColorWheel example rendered below.
5. Notice that the ColorWheel handle does not change, by growing in size, to display focus.
6. If you press Esc or an Arrow key to change the value, the focused style will render.

### How can we reproduce this issue?

3. Scroll to '....'
4. Check console
5. See error

### Sample code that illustrates the problem

In order to account for the focused element in the shadowRoot matching  `:focus-visible` or `.focus-visible`, change  this:
https://github.com/adobe/spectrum-web-components/blob/2682d937d392f9858ea757e42acfb8b4ca738753/tools/base/src/Base.ts#L81-L100

To this:
```ts
        public hasVisibleFocusInTree(): boolean {
            const activeElement = (this.getRootNode() as Document)
                .activeElement as HTMLElement;
            if (!activeElement) {
                return false;
            }

            const shadowRootActiveElement = activeElement.shadowRoot
                ?.activeElement as HTMLElement;

            // Browsers without support for the `:focus-visible`
            // selector will throw on the following test (Safari, older things).
            // Some won't throw, but will be focusing item rather than the menu and
            // will rely on the polyfill to know whether focus is "visible" or not.
            try {
                return (
                    activeElement.matches(':focus-visible') ||
[...]

PR #3279, opened 2023-06-05
fix(color-slider,color-wheel): fail to render focused style when first receiving focus #3278

majornista:
## Description

`<sp-color-slider>` and `<sp-color-wheel>` should render the focused style when they first receive focus, tabbing into the document of iframe.

## Related issue(s)

#3278 

## How has this been tested?

For ColorSlider:

1. Open https://opensource.adobe.com/spectrum-web-components/components/color-slider/#default
2. Hover over and mouse click to set focus on the "#" sign next to the heading reading "Default."
3. Press Tab key once to set focus to the "#" link with href="#default".
4. Then press Tab a second time to move focus to the ColorSlider example rendered below.
5. Notice that the ColorSlider handle does not change, by growing in size, to display focus.
6. If you press Esc or an Arrow key to change the value, the focused style will render.

Repeat for ColorWheel:

1. Open https://opensource.adobe.com/spectrum-web-components/components/color-wheel/#example
2. Hover over and mouse click to set focus on the "#" sign next to the heading reading "Example."
3. Press Tab key once to set focus to the "#" link with href="#default".
4. Then press Tab a second time to move focus to the ColorWheel example rendered below.
5. Notice that the ColorWheel handle does not change, by growing in size, to display focus.
6. If you press Esc or an Arrow key to change the value, the focused style will render.

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/tools/base/src/Base.ts
+++ b/tools/base/src/Base.ts
@@ -84,6 +84,10 @@ export function SpectrumMixin<T extends Constructor<ReactiveElement>>(
             if (!activeElement) {
                 return false;
             }
+
+            const shadowRootActiveElement = activeElement.shadowRoot
+                ?.activeElement as HTMLElement;
+
             // Browsers without support for the `:focus-visible`
             // selector will throw on the following test (Safari, older things).
             // Some won't throw, but will be focusing item rather than the menu and
@@ -91,11 +95,16 @@ export function SpectrumMixin<T extends Constructor<ReactiveElement>>(
             try {
                 return (
                     activeElement.matches(':focus-visible') ||
-                    activeElement.matches('.focus-visible')
+                    activeElement.matches('.focus-visible') ||
+                    shadowRootActiveElement?.matches(':focus-visible') ||
+                    shadowRootActiveElement?.matches('.focus-visible')
                 );
                 /* c8 ignore next 3 */
             } catch (error) {
-                return activeElement.matches('.focus-visible');
+                return (
+                    activeElement.matches('.focus-visible') ||
+                    shadowRootActiveElement?.matches('.focus-visible')
+                );
             }
         }
 
--- a/tools/base/src/Base.ts
+++ b/tools/base/src/Base.ts
@@ -90,24 +90,13 @@ export function SpectrumMixin<T extends Constructor<ReactiveElement>>(
             // Some won't throw, but will be focusing item rather than the menu and
             // will rely on the polyfill to know whether focus is "visible" or not.
             try {
-                return !!(
+                return (
                     activeElement.matches(':focus-visible') ||
-                    activeElement.matches('.focus-visible') ||
-                    activeElement.shadowRoot?.activeElement?.matches(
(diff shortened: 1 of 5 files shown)


Westbrook:
You should already be added to the project, would you be able to make this as a branch so that it is easier to review you changes via the preview URLs that come with that?

tools/base/src/Base.ts
```diff
@@ -84,18 +84,30 @@ export function SpectrumMixin<T extends Constructor<ReactiveElement>>(
...
                 return false;
             }
+
             // Browsers without support for the `:focus-visible`
             // selector will throw on the following test (Safari, older things).
             // Some won't throw, but will be focusing item rather than the menu and
             // will rely on the polyfill to know whether focus is "visible" or not.
             try {
-                return (
+                return !!(
```
Westbrook, line 93: `.matches` returns a boolean, I don't think we should need coercion here.

tools/base/src/Base.ts
```diff
@@ -84,18 +84,30 @@ export function SpectrumMixin<T extends Constructor<ReactiveElement>>(
...
                     activeElement.matches(':focus-visible') ||
-                    activeElement.matches('.focus-visible')
+                    activeElement.matches('.focus-visible') ||
+                    activeElement.shadowRoot?.activeElement?.matches(
+                        ':focus-visible'
+                    ) ||
+                    activeElement.shadowRoot?.activeElement?.matches(
+                        '.focus-visible'
+                    )
```
Westbrook, line 101: This doesn't _feel_ like the right fix. Namely in that it changes the meaning of the method from it's current name, but also because this move to encapsulation breaking which isn't something we should do lightly.

I'd want to look into why the visible focus is resolved on `Shift + Tab` entrance of the element and not on the `Tab` entrance before committing to something like this. Hopefully there's some less invasive insight to be gained from that...

tools/base/src/Base.ts
majornista: Coercion is is necessary for `activeElement.shadowRoot?.activeElement?`, which could return `undefined`.

tools/base/src/Base.ts
majornista: When tabbing forward, the `activeElement` will be the `sp-color-slider` or `sp-color-wheel`, but that element does not match `:focus-visible` or `.focus-visible`, because it is not the actual element within its shadowRoot that has focus.

Westbrook:
Looking at this more closely, it's probably best if we solve this close to the source rather than extend library methods beyond their scope. Please take a look at the changes in https://github.com/adobe/spectrum-web-components/commit/8c81282e230e0ed7fd482a2736cbf893c8438bb1. As a branch, you can preview how that works [here](https://color-slider-focus--spectrum-web-components.netlify.app/components/color-slider/#default). Each of the color elements maintains a similar pattern and should be amenable to the same refactoring.

majornista:
Close in favor of #3283

Issue #3164, opened 2023-04-28
[Bug]: Memory Leak in Menu


joekukish:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Menu

### Actual behavior

MenuItemRemovedEvent keeps an instance to an object, which is never cleared. This introduces a leak since the object cannot be garbage collected.

https://studio.webcomponents.dev/edit/C5gyfKCjjMMv14UmjQzr/src/index.ts?p=stories

### Screenshots
<img width="1049" alt="Screenshot 2023-04-28 at 16 38 02" src="https://user-images.githubusercontent.com/1413153/235177422-1f6e4e9e-21b5-4cc0-ac8d-b409b481d70f.png">

_No response_

### How can we reproduce this issue?

1. Go to https://studio.webcomponents.dev/edit/C5gyfKCjjMMv14UmjQzr/src/index.ts?p=stories
2. Open Chrome DevTools and go to the Memory tab.
3. Click on Profiles > Head snapshot > Take snapshot
4. Click the Show / Hide button (10-20 times)
5. Go to Performance Tab in DevTools, and click on the Trash Icon (garbage collection).
6. Go to Memory Tab, and take another snapshot
7. Next to the search field, "All Objects", select Objects allocated between Snapshot 1 and Snapshot 2.
<img width="892" alt="image" src="https://user-images.githubusercontent.com/1413153/235186117-acbbf57e-e108-495d-b8d0-51828ac6ae32.png">
8. Search for Detached
<img width="887" alt="image" src="https://user-images.githubusercontent.com/1413153/235186039-5099513d-0df6-4304-9df2-20a0d850bc84.png">

`_item` is the MenuItem held inside the event

<img width="694" alt="image" src="https://user-images.githubusercontent.com/1413153/235186362-1635acce-3f39-4204-b0cb-5665a4a81af8.png">

Westbrook:
Would be great to have a slice for slice test scenario for this to compare fixed against. You wouldn't happen to have set something like that up in order to track this down?

joekukish:
@Westbrook I've added a playground where it can be reproduced and steps of how to find the leak.

nawazmobusss:
Hi @Westbrook , Can I kindly ask for an update on this ticket?  One of our components is using Menu and we see memory leaks leading to memory crashes in mobile app & safari browser. It would be highly appreciated if this could be addressed soon as we are looking for a public release in 2 weeks.

[...]

PR #3189, opened 2023-05-09
fix(menu): #3164 plug memory leak with gobal events

chrisbank:
Global Events were holding onto DOM trees and showing up in chromes DetachedElement retainer chains. Use per instance events instead of reusing the 

- Fixes #3164 

## How has this been tested?

Tested the fix locally in a project using SWC.

## Types of changes

-   [x] Bug fix (non-breaking change which fixes an issue)
-   [ ] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to change)
-   [ ] Chore (minor updates related to the tooling or maintenance of the repository, does not impact compiled assets)

## Checklist

-   [x] I have signed the [Adobe Open Source CLA](http://opensource.adobe.com/cla.html).
-   [x] My code follows the code style of this project.
-   [ ] If my change required a change to the documentation, I have updated the documentation in this pull request.
-   [x] I have read the **[CONTRIBUTING](<(https://github.com/adobe/spectrum-web-components/blob/main/CONTRIBUTING.md)>)** document.
-   [ ] I have added tests to cover my changes.
-   [ ] All new and existing tests passed.
-   [ ] I have reviewed at the Accessibility Practices for this feature, see: [Aria Practices](https://www.w3.org/TR/wai-aria-practices/)

## Best practices

This repository uses conventional commit syntax for each commit message; note that the GitHub UI does not use this by default so be cautious when accepting suggested changes. Avoid the "Update branch" button on the pull request and opt instead for rebasing your branch against `main`.

PR #3289, opened 2023-06-06
fix(menu): #3164 plug memory leak with gobal events

chrisbank:
## Related issue(s)

Global Events were holding onto DOM trees and showing up in chromes DetachedElement retainer chains. Use per instance events instead of reusing the

Fixes https://github.com/adobe/spectrum-web-components/issues/3164

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/menu/src/Menu.ts
+++ b/packages/menu/src/Menu.ts
@@ -700,6 +700,10 @@ export class Menu extends SpectrumElement {
         }
         this.updateComplete.then(() => this.updateItemFocus());
     }
+    public override disconnectedCallback(): void {
+        this.cachedChildItems = undefined;
+        super.disconnectedCallback();
+    }
 
     protected childItemsUpdated!: Promise<unknown[]>;
     protected cacheUpdated = Promise.resolve();
--- a/packages/menu/src/MenuItem.ts
+++ b/packages/menu/src/MenuItem.ts
@@ -96,8 +96,29 @@ export class MenuItemAddedOrUpdatedEvent extends Event {
 
 export type MenuItemChildren = { icon: Element[]; content: Node[] };
 
-const addOrUpdateEvent = new MenuItemAddedOrUpdatedEvent();
-const removeEvent = new MenuItemRemovedEvent();
+let addOrUpdateEvent = new MenuItemAddedOrUpdatedEvent();
+let removeEvent = new MenuItemRemovedEvent();
+/**
+ * Code to cleanup these global events async in batches
+ */
+let addOrUpdateEventRafId = 0;
+function resetAddOrUpdateEvent(): void {
+    if (addOrUpdateEventRafId === 0) {
+        addOrUpdateEventRafId = requestAnimationFrame(() => {
+            addOrUpdateEvent = new MenuItemAddedOrUpdatedEvent();
+            addOrUpdateEventRafId = 0;
+        });
+    }
+}
+let removeEventEventtRafId = 0;
+function resetRemoveEvent(): void {
+    if (removeEventEventtRafId === 0) {
+        removeEventEventtRafId = requestAnimationFrame(() => {
+            removeEvent = new MenuItemRemovedEvent();
+            removeEventEventtRafId = 0;
+        });
+    }
+}
 
 /**
  * @element sp-menu-item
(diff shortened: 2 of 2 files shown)


benjamind:
Curious for the need for all this complexity? Isn't the result here that on each new frame you will instantiate and dispatch a new menu event? Why not simply instantiate the new event at the point its needed rather than managing the reset?

najikahalsema:
If you could make a story that replicates what's done [here](https://studio.webcomponents.dev/edit/C5gyfKCjjMMv14UmjQzr/src/index.ts?p=stories) so we can test this in the browser and compare to the current version using the steps outlined in #3164, that would be really helpful for moving this PR along!

najikahalsema:
> Curious for the need for all this complexity? Isn't the result here that on each new frame you will instantiate and dispatch a new menu event? Why not simply instantiate the new event at the point its needed rather than managing the reset?

@benjamind This might answer your question ??

hunterloftis:
I'm glad @najikahalsema and @benjamind are looking at this so it won't stall out while I'm on PTO. To pull a few threads together:

- @benjamind the driver for not instantiating new events is a previous performance issue. I don't have a lot of context here but Najika linked to Westbrook's thread about it.
- I hope this is possible to resolve in a simple way. @chrisbank did you test `reset()`ing the singleton events once they reach the top of the DOM? Did that fail?

chrisbank:
@benjamind I tried just using events in my original PR but it got rejected https://github.com/adobe/spectrum-web-components/pull/3189

@hunterloftis I did lots of testing on reset and looking for other ways to do this which is why it took so long. We think it's properties in the Native Event that are also holding onto stuff. Possibly clearing 
```
this.srcElement = undefined;
this..target = undefined;
```
in the reset might work, but typescript correctly complains that those are specs as Readonly properties so it might work in chrome but maybe not other browsers. If this way is acceptable I'll go with it. If the reset is required then I'll do a lot of any and unknowns to see if it will work

chrisbank:
@najikahalsema you can already see the leak in that story. Open and close with the button a few times, then end with it closed. Take a keep snapshot and look for Detached Element and the thing the event is holding onto should be near the top of the retainers list. The hard thing is the source there is minified, so in my screen shot the global event is m$6 but you can see the `_item` in the retainer above it as that's what's holding onto the menu. 

image

najikahalsema:
> @najikahalsema you can already see the leak in that story. Open and close with the button a few times, then end with it closed. Take a keep snapshot and look for Detached Element and the thing the event is holding onto should be near the top of the retainers list. The hard thing is the source there is minified, so in my screen shot the global event is m$6 but you can see the `_item` in the retainer above it as that's what's holding onto the menu.
> 
> image

@chrisbank I meant replicate that here in this branch.

chrisbank:
> > @najikahalsema you can already see the leak in that story. Open and close with the button a few times, then end with it closed. Take a keep snapshot and look for Detached Element and the thing the event is holding onto should be near the top of the retainers list. The hard thing is the source there is minified, so in my screen shot the global event is m$6 but you can see the `_item` in the retainer above it as that's what's holding onto the menu.
> > image
> 
> @chrisbank I meant replicate that here in this branch.

@najikahalsema This is my first PR to SWC. How do I get a storybook to do that and what else is required as part of the submission process?

najikahalsema:
> > > @najikahalsema you can already see the leak in that story. Open and close with the button a few times, then end with it closed. Take a keep snapshot and look for Detached Element and the thing the event is holding onto should be near the top of the retainers list. The hard thing is the source there is minified, so in my screen shot the global event is m$6 but you can see the `_item` in the retainer above it as that's what's holding onto the menu.
> > > image
> > 
> > 
> > @chrisbank I meant replicate that here in this branch.
> 
> @najikahalsema This is my first PR to SWC. How do I get a storybook to do that and what else is required as part of the submission process?

In each component, there is a folder called `stories`. Menu has multiple story files, so I'd use `menu-item.stories.ts` for this example. The way it works is very similar to the example on webcomponents.dev, but I would probably consolidate all the functions in that file to be within one method. Each method in the story file corresponds to the stories in the storybook. You can run the storybook by using `yarn storybook` in your console to see a local copy. 

You won't need to have these changes in the final PR, but since you're not familiar with this workflow, I may just pull your PR and do it myself to test your changes.

najikahalsema:
I tested the changes locally, but I found out with @chrisbank's help that Lit needs to be updated as there is a bug preventing garbage collection on elements. I wasn't seeing the garbage collection because of that. Once I updated lit to 2.7.5, all of the detached HTMLElements are gone. This is the head snapshot comparison in the new version: 
image Looks much better.

najikahalsema (approved):
Based on my previous comment, I think this looks good. We'll have this out for you later in the week. Thank you for your patience.

nawazmobusss:
Thanks a lot, @najikahalsema ! When can we start consuming this latest version or is it published to artifactory already?

najikahalsema:
@nawazmobusss You should be able to use it tomorrow once I get a release out. There are some other things that need to be updated before we can publish the release. Typically, we release on Thursday, but since this is high priority for you, I will have it ready a day early.

Issue #3048, opened 2023-03-21
[Bug]: Clicking action button not selecting action button when action button is inside action group.


sonirajan:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-action-group, sp-action-button

### Expected behavior

Clicking any action button should select it when they are child elements of action group.

### Actual behavior

I’m trying to use sp-action-group component with sp-action-button as child components. When I try to select/click one of the sp-action-button, it calls [handleClick](https://github.com/adobe/spectrum-web-components/blob/main/packages/action-group/src/ActionGroup.ts#L171) event. Here, event.target is always span instead of parent sp-action-button. And, because of that target.value is always "undefined" and returns and does nothing. Is this the bug on SWC side? I have sp-action-group v0.12.9 and sp-action-button v0.10.12 installed. Slack conversion is [here](https://creativecloud.slack.com/archives/CESK60MQD/p1678898210005339).

### What browsers are you seeing the problem in?

Chrome

Westbrook:
Indeed. Seems like we could update
```
const target = event.target as ActionButton;
```
To something like:
```
const target = event.composedPath().find(el => this.buttons.includes(el));
```
But, performant and typescripty and supported by a test.

We'd appreciate a contribution on this!

Westbrook:
fixed by #3292

PR #3292, opened 2023-06-07
fix(action-group): fixes action group click event which was not selecting action button when they are child elements of action group

sonirajan:
## Motivation and context

I’m trying to use `sp-action-group` component with `sp-action-button` as child components. When I try to select/click one of the `sp-action-button`, it calls [handleClick](https://github.com/adobe/spectrum-web-components/blob/main/packages/action-group/src/ActionGroup.ts#L171) event. Here, `event.target` is always `span` instead of parent `sp-action-button`. And, because of that `target.value` is always `undefined` and returns and does nothing. 

Technical detail: Retargeting is not occuring because the event occurs on a slotted element inside action button, that physically lives in the light DOM. And because of that event.target is always `span` instead of `sp-action-button`. To fix that, we used [event.composedPath()](https://javascript.info/shadow-dom-events#bubbling-event-composedpath) and compares if any html element in path is `action button` that is part of `this.buttons`.

## Related issue(s)

[#3048](https://github.com/adobe/spectrum-web-components/issues/3048)
[Slack discussion](https://creativecloud.slack.com/archives/CESK60MQD/p1678898210005339)

## How has this been tested?

ccx-commenting uses filter popover to filter comments where we have multiple action groups(as various filters) with action buttons as child elements. When we tried to use SWC (instead of RS2), clicking various filters (clicking action button) was not selecting action button(filter not being applied). I tested fix directly in commenting component.

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/action-button/src/action-button.css
+++ b/packages/action-button/src/action-button.css
@@ -19,6 +19,7 @@ governing permissions and limitations under the License.
 #label {
     flex-grow: var(--spectrum-actionbutton-label-flex-grow);
     text-align: var(--spectrum-actionbutton-label-text-align);
+    pointer-events: none !important;
 }
 
 :host([size='xs']) {
--- a/packages/action-group/src/ActionGroup.ts
+++ b/packages/action-group/src/ActionGroup.ts
@@ -171,8 +171,12 @@ export class ActionGroup extends SizedMixin(SpectrumElement, {
     }
 
     private handleClick(event: Event): void {
-        const target = event.target as ActionButton;
-        if (typeof target.value === 'undefined') {
+        const target = event.composedPath().find((el) => {
+            const btnTarget = el as ActionButton;
+            return this.buttons.includes(btnTarget);
+        }) as ActionButton;
+
+        if (typeof target?.value === 'undefined') {
             return;
         }
         switch (this.selects) {
--- a/packages/action-group/src/ActionGroup.ts
+++ b/packages/action-group/src/ActionGroup.ts
@@ -172,7 +172,6 @@ export class ActionGroup extends SizedMixin(SpectrumElement, {
 
     private handleClick(event: Event): void {
         const target = event.target as ActionButton;
-
         if (typeof target.value === 'undefined') {
             return;
         }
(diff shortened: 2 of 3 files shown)


Westbrook:
I wonder if there is a CSS only solution to this? Possibly:
```
::slotted(*) {
    pointer-events: none !important;
}
```
🤔...

sonirajan:
> I wonder if there is a CSS only solution to this? Possibly:
> 
> ```
> ::slotted(*) {
>     pointer-events: none !important;
> }
> ```
> 
> 🤔...

Should I use it [here](https://github.com/adobe/spectrum-web-components/blob/main/packages/action-group/src/spectrum-action-group.css#L58) and test it out?

Westbrook:
That file is mechanically generated, so you'll need to include it [here](https://github.com/adobe/spectrum-web-components/blob/main/packages/action-group/src/action-group.css), to test. 🤞🏼

sonirajan:
@Westbook Hope you had nice days off :)
Regarding your last comment, CSS solution is not working. `event.target` is `sp-action-group` and `target.value` is undefined, so `handleClick` just returns.  
image

sonirajan:
I did more research and found why above CSS won't work _I think_. The `::slotted()` represents any element that has been placed into a slot inside an HTML template. In ActionGroup, all action buttons are placed inside one [slot](https://github.com/adobe/spectrum-web-components/blob/main/packages/action-group/src/ActionGroup.ts#L316). So, no pointer event is attached to any action button when we go with CSS solution above. And so, when we click any button of Action Group, `event.target` is `ActionGroup` itself in `handleClick` and it will just return and do nothing. 
image

Westbrook:
I see. My intent was for this CSS rule to be added to the Action Button, not the Group. The hope there being that no matter what was slotted into the Action Button the host would receive all events. If that doesn’t work, we can move forward with you approach in JS.

sonirajan:
Hi @west

> I see. My intent was for this CSS rule to be added to the Action Button, not the Group. The hope there being that no matter what was slotted into the Action Button the host would receive all events. If that doesn’t work, we can move forward with you approach in JS.

CSS fix on Action Button is working fine. Thanks for the hint.

PR #3296, opened 2023-06-07
fix(number-field): simplify width management

Westbrook:
## Description
Resolve a couple of incongruities between Spectrum CSS and SWC width application in Number Field, Text Field, and editable Slider elements.

## How has this been tested?
-   [ ] _Test case 1_
    1. Go [here](https://number-field-width--spectrum-web-components.netlify.app/storybook/?path=/story/slider--editable)
    2. Compare to [here](https://opensource.adobe.com/spectrum-web-components/storybook/index.html?path=/story/slider--editable)

## Types of changes

## Checklist

### DIFF
--- a/packages/number-field/src/number-field.css
+++ b/packages/number-field/src/number-field.css
@@ -12,13 +12,12 @@ governing permissions and limitations under the License.
 
 @import './spectrum-number-field.css';
 
-:host([quiet]) {
-    width: var(--spectrum-stepper-quiet-width);
+:host {
+    inline-size: var(--mod-stepper-width, var(--spectrum-stepper-width));
 }
 
-#textfield,
-:host([quiet]) #textfield {
-    width: 100%;
+#textfield {
+    inline-size: 100%;
 }
 
 sp-field-button {
--- a/packages/textfield/test/textfield.test.ts
+++ b/packages/textfield/test/textfield.test.ts
@@ -127,13 +127,15 @@ describe('Textfield', () => {
                 ></sp-textfield>
             `
         );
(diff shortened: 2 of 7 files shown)


packages/number-field/src/number-field.css
```diff
@@ -12,12 +12,11 @@ governing permissions and limitations under the License.
...
-:host([quiet]) {
-    width: var(--spectrum-stepper-quiet-width);
+:host {
+    inline-size: var(--mod-stepper-width, var(--spectrum-stepper-width));
 }
 
-#textfield,
-:host([quiet]) #textfield {
+#textfield {
     width: 100%;
```
Rajdeepc, line 20: We should use logical-properties like inline-size instead of width

Rajdeepc (changes requested):

packages/number-field/src/number-field.css
Westbrook: Great catch!

packages/textfield/test/textfield.test.ts
```diff
@@ -127,13 +127,15 @@ describe('Textfield', () => {
                 ></sp-textfield>
             `
         );
-        const startBounds = el.getBoundingClientRect();
-
+        // Resizing only effect block size of host rect, so measure the `focusElement` when resizing.
```
hunterloftis, line 130:
```suggestion
        // Resizing only effects the block size of the host rect, so measure the `focusElement` when resizing.
```

hunterloftis (dismissed):

PR #3297, opened 2023-06-07
fix(number-field,search,textfield): add t-shirt sizes

Westbrook:
## Description
Add t-shirt sizing to Number Field, Search, Textfield and subsequently "text area".

Normally this would be "breaking", but in this instance the Spectrum CSS source allows for delivering the pattern _without_ a t-shirt size listing meaning that all consumers should be able to update and continue to deliver their content as they had previously.

### To Do:
- [ ] complete design review
- [ ] update VRT cache

## Types of changes

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ executors:
 parameters:
     current_golden_images_hash:
         type: string
-        default: 0c939bbe6b6fe9440a5273d073e014e361354361
+        default: ff19056ab0be11ab2a412e204e7fddf8c92e1818
     wireit_cache_name:
         type: string
         default: wireit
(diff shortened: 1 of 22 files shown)


pfulton (dismissed):
Nice! This looks good to me. IMO it's nice to see this "medium as default" pattern playing out and it's definitely something to discuss with the larger CSS squad.

hunterloftis (commented):
Is the width of the number field in the docs being constrained?

https://tshirt-sizes--spectrum-web-components.netlify.app/components/number-field/#sizes

image

hunterloftis (dismissed):
Gotcha, thanks for the explanation. 

It could be useful to include an example of using an XL number field in the docs and increasing its width via CSS, in order for the docs to be more helpful to folks who want to use it in L/XL mode.

hunterloftis (dismissed):

PR #3300, opened 2023-06-08
feat(accordion): core token migration

vjosyula:
## Description
This change covers accordion core token migration.
with this t-shirt and densities are supported on accordion.

## Related issue(s)

-

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

accordion

## Types of changes

## Checklist

### DIFF
--- a/packages/accordion/src/Accordion.ts
+++ b/packages/accordion/src/Accordion.ts
@@ -41,6 +41,12 @@ export class Accordion extends SpectrumElement {
     @property({ type: Boolean, reflect: true, attribute: 'allow-multiple' })
     public allowMultiple = false;
 
+    @property({ type: String, reflect: true })
+    public density?: 'compact' | 'regular' | 'spacious' = 'regular';
+
+    @property({ type: String, reflect: true })
+    public size?: 's' | 'm' | 'l' | 'xl' = 'm';
+
     @queryAssignedNodes()
     private defaultNodes!: NodeListOf<AccordionItem>;
 
@@ -87,11 +93,20 @@ export class Accordion extends SpectrumElement {
         this.focusGroupController.clearElementCache();
     }
 
+    private setAccordionProperties(): void {
+        this.items.forEach((item) => {
+            item.size = this.size;
+            item.density = this.density;
+        });
+    }
+
     protected override render(): TemplateResult {
+        this.setAccordionProperties();
         return html`
             <slot
                 @slotchange=${this.handleSlotchange}
                 @sp-accordion-item-toggle=${this.onToggle}
+                @sp-accordion-item-added=${this.setAccordionProperties}
             ></slot>
         `;
     }
--- a/packages/accordion/src/Accordion.ts
+++ b/packages/accordion/src/Accordion.ts
@@ -32,7 +32,9 @@ import styles from './accordion.css.js';
  * @element sp-accordion
  * @slot - The sp-accordion-item children to display.
  */
-export class Accordion extends SizedMixin(SpectrumElement) {
+export class Accordion extends SizedMixin(SpectrumElement, {
+    noDefaultSize: true,
+}) {
     public static override get styles(): CSSResultArray {
         return [styles];
     }
@@ -44,7 +46,7 @@ export class Accordion extends SizedMixin(SpectrumElement) {
     public allowMultiple = false;
 
     @property({ type: String, reflect: true })
-    public density?: 'compact' | 'regular' | 'spacious' = 'regular';
+    public density: 'compact' | 'regular' | 'spacious' = 'regular';
 
     @queryAssignedNodes()
     private defaultNodes!: NodeListOf<AccordionItem>;
@@ -88,13 +90,9 @@ export class Accordion extends SizedMixin(SpectrumElement) {
         });
     }
 
-    private handleSlotchange({
-        target: slot,
-    }: Event & { target: HTMLSlotElement }): void {
+    private handleSlotchange({}: Event & { target: HTMLSlotElement }): void {
         this.focusGroupController.clearElementCache();
-
-        const assignedElements = slot.assignedElements() as AccordionItem[];
-        assignedElements.forEach((item) => {
+        this.items.forEach((item) => {
             item.size = this.size;
         });
     }
@@ -102,8 +100,7 @@ export class Accordion extends SizedMixin(SpectrumElement) {
     protected override updated(changed: PropertyValues<this>): void {
         super.updated(changed);
         if (changed.has('size')) {
-            const assignedElements = [...this.items] as AccordionItem[];
-            assignedElements.forEach((item) => {
+            this.items.forEach((item) => {
                 item.size = this.size;
             });
         }
--- a/packages/accordion/src/Accordion.ts
+++ b/packages/accordion/src/Accordion.ts
@@ -49,7 +49,7 @@ export class Accordion extends SizedMixin(SpectrumElement, {
      * Sets the spacing between the content to borders of an accordion item
      */
     @property({ type: String, reflect: true })
-    public density: 'compact' | 'spacious' | undefined;
+    public density?: 'compact' | 'spacious';
 
     @queryAssignedNodes()
     private defaultNodes!: NodeListOf<AccordionItem>;
@@ -93,7 +93,7 @@ export class Accordion extends SizedMixin(SpectrumElement, {
         });
     }
 
-    private handleSlotchange({}: Event & { target: HTMLSlotElement }): void {
+    private handleSlotchange(): void {
         this.focusGroupController.clearElementCache();
         this.items.forEach((item) => {
             item.size = this.size;
@@ -102,7 +102,10 @@ export class Accordion extends SizedMixin(SpectrumElement, {
 
     protected override updated(changed: PropertyValues<this>): void {
         super.updated(changed);
-        if (changed.has('size')) {
+        if (
+            changed.has('size') &&
+            (!!changed.get('size') || this.size !== 'm')
+        ) {
             this.items.forEach((item) => {
                 item.size = this.size;
             });
--- a/packages/accordion/src/Accordion.ts
+++ b/packages/accordion/src/Accordion.ts
@@ -13,6 +13,7 @@ governing permissions and limitations under the License.
 import {
     CSSResultArray,
     html,
+    PropertyValues,
     SizedMixin,
     SpectrumElement,
     TemplateResult,
@@ -96,6 +97,16 @@ export class Accordion extends SizedMixin(SpectrumElement) {
         assignedElements.forEach((item) => {
             item.size = this.size;
         });
+    }
+
+    protected override updated(changed: PropertyValues<this>): void {
+        super.updated(changed);
+        if (changed.has('size')) {
+            const assignedElements = [...this.items] as AccordionItem[];
+            assignedElements.forEach((item) => {
+                item.size = this.size;
+            });
+        }
     }
 
     protected override render(): TemplateResult {
--- a/packages/accordion/src/Accordion.ts
+++ b/packages/accordion/src/Accordion.ts
@@ -45,8 +45,11 @@ export class Accordion extends SizedMixin(SpectrumElement, {
     @property({ type: Boolean, reflect: true, attribute: 'allow-multiple' })
     public allowMultiple = false;
 
+    /**
+     * Sets the spacing between the content to borders of an accordion item
+     */
     @property({ type: String, reflect: true })
-    public density: 'compact' | 'regular' | 'spacious' = 'regular';
+    public density: 'compact' | 'spacious' | undefined;
 
     @queryAssignedNodes()
     private defaultNodes!: NodeListOf<AccordionItem>;
--- a/packages/accordion/src/Accordion.ts
+++ b/packages/accordion/src/Accordion.ts
@@ -13,6 +13,7 @@ governing permissions and limitations under the License.
 import {
     CSSResultArray,
     html,
+    SizedMixin,
     SpectrumElement,
     TemplateResult,
 } from '@spectrum-web-components/base';
@@ -30,12 +31,7 @@ import styles from './accordion.css.js';
  * @element sp-accordion
  * @slot - The sp-accordion-item children to display.
  */
-
-import { SizedMixin } from '@spectrum-web-components/base';
-
-export class Accordion extends SizedMixin(SpectrumElement, {
-    validSizes: ['s', 'm', 'l', 'xl'],
-}) {
+export class Accordion extends SizedMixin(SpectrumElement) {
     public static override get styles(): CSSResultArray {
         return [styles];
     }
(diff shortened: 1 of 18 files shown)


pfulton (dismissed):
Thanks for getting started with this @vjosyula.

I think the SWC team would like to see the t-shirt sizes added to this component, so if you're able to add those to the component as well as the Storybook, that would be excellent.

There are several other components in the repo that have t-shirt sizes, so you can probably take a look at those as a guide.

Westbrook:
Thanks for pointing to the t-shirt size additions @pfulton that will be a very welcome addition to this PR.

Right now the VRTs are pretty unhappy, but with it still in "draft", I'm sure you're taking a look at those now that the branch is on remote.

Looking forward to seeing this come together!

vjosyula:
Hi @pfulton and @Westbrook,
Working on t-shirt sizes and densities. For now, I created this PR to understand the failures.

packages/accordion/src/Accordion.ts
```diff
@@ -41,6 +41,12 @@ export class Accordion extends SpectrumElement {
     @property({ type: Boolean, reflect: true, attribute: 'allow-multiple' })
     public allowMultiple = false;
 
+    @property({ type: String, reflect: true })
+    public density?: 'compact' | 'regular' | 'spacious' = 'regular';
+
+    @property({ type: String, reflect: true })
+    public size?: 's' | 'm' | 'l' | 'xl' = 'm';
```
najikahalsema, line 48: We no longer size our components with an attribute like this. Rather, we use the `SizedMixin` helper as an extension (see a usage of this [here](https://github.com/adobe/spectrum-web-components/blob/6d5d6b3c31d92a058202a242a283125d25131f81/packages/tags/src/Tag.ts#L35)). This should simplify the implementation for you as you won't need the `setAccordionProperties()` method anymore.

packages/accordion/src/AccordionItem.ts
```diff
@@ -42,6 +50,9 @@ export class AccordionItem extends Focusable {
     @property({ type: Boolean, reflect: true })
     public override disabled = false;
 
+    @property({ type: String, reflect: true })
```
najikahalsema, line 53: Because you're handling the sizing of the component from `Accordion.ts` via `SizedMixin`, you shouldn't need to worry about sizing in `AccordionItem.ts`.

packages/accordion/src/AccordionItem.ts
```diff
@@ -23,6 +24,13 @@ import chevronIconStyles from '@spectrum-web-components/icon/src/spectrum-icon-c
 
 import styles from './accordion-item.css.js';
 
+const chevronClass = {
```
najikahalsema, line 27: This is what's stumping me about whether or not we'd want to use a helper extension. If we don't need `SizedMixin` here, but we have sized _icons_, will that cause a problem?

packages/accordion/src/AccordionItem.ts
```diff
@@ -80,8 +91,10 @@ export class AccordionItem extends Focusable {
                 >
                     ${this.label}
                 </button>
-                <sp-icon-chevron100
-                    class="indicator spectrum-UIIcon-ChevronRight100"
+                <span class="iconContainer">
+                    <sp-icon-chevron100 class="indicator ${
```
najikahalsema, line 95: If you use `SizedMixin` as I indicated in my previous comments, you might be able to remove `chevronClass[this.size as DefaultElementSize]` and just have `indicator` be the only class. If that doesn't work, I would maybe try to also use the class `spectrum-UIIcon-ChevronRight100`.

packages/accordion/src/spectrum-config.js
```diff
@@ -39,17 +40,34 @@ const config = {
...
+                        ['spectrum-Accordion--compact', 'compact'],
+                        ['spectrum-Accordion--spacious', 'spacious'],
+                    ],
+                    'density'
+                ),
+                ...converter.enumerateAttributes(
+                    [
+                        ['spectrum-Accordion--sizeS', 's'],
+                        ['spectrum-Accordion--sizeL', 'l'],
+                        ['spectrum-Accordion--sizeXL', 'xl'],
```
najikahalsema, line 54: I believe size medium needs to be added to this list, as well.

packages/accordion/stories/accordion-densities.stories.ts
```diff
@@ -0,0 +1,29 @@
...
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import { TemplateResult } from '@spectrum-web-components/base';
+
+import '@spectrum-web-components/menu/sp-menu.js';
+import '@spectrum-web-components/menu/sp-menu-item.js';
```
najikahalsema, line 16: Why is `sp-menu` and `sp-menu-item` being imported?

najikahalsema (changes requested):
The changes look good visually, and almost everything else seems good. The only thing is the way the size attribute is handled. While you're doing that, please rebase your branch with main. That will get the github tachometer to stop failing for you on Chrome :) Keep up the great work and let me know if you have questions about anything I've written!

packages/accordion/stories/accordion-densities.stories.ts
vjosyula: removed unused imports.

packages/accordion/src/AccordionItem.ts
najikahalsema: That surprises me. Even with all of the changes I recommended?

packages/accordion/src/Accordion.ts
```diff
@@ -30,7 +32,7 @@ import styles from './accordion.css.js';
  * @element sp-accordion
  * @slot - The sp-accordion-item children to display.
  */
-export class Accordion extends SpectrumElement {
+export class Accordion extends SizedMixin(SpectrumElement) {
```
najikahalsema, line 35: We don't want to apply a size by default, so let's use the second argument noDefaultSize so it doesn't assume it needs to apply size m by default. This will improve the performance of the component.

packages/accordion/src/Accordion.ts
```diff
@@ -41,6 +43,9 @@ export class Accordion extends SpectrumElement {
     @property({ type: Boolean, reflect: true, attribute: 'allow-multiple' })
     public allowMultiple = false;
 
+    @property({ type: String, reflect: true })
+    public density?: 'compact' | 'regular' | 'spacious' = 'regular';
```
najikahalsema, line 47:
```suggestion
    public density?: 'compact' | 'spacious';
```

We shouldn't need to manage an attribute for 'regular', since this density property is optional and "regular" is the default, anyway. Please remove this and its usages.

packages/accordion/src/Accordion.ts
```diff
@@ -83,8 +88,25 @@ export class Accordion extends SpectrumElement {
         });
     }
 
-    private handleSlotchange(): void {
+    private handleSlotchange({
+        target: slot,
+    }: Event & { target: HTMLSlotElement }): void {
         this.focusGroupController.clearElementCache();
+
+        const assignedElements = slot.assignedElements() as AccordionItem[];
```
najikahalsema, line 96: `assignedElements` is the same as `this.items`, so we don't need to set it as a variable here. There may some typing nuances that need to be taken care of.

packages/accordion/src/Accordion.ts
```diff
@@ -83,8 +88,25 @@ export class Accordion extends SpectrumElement {
...
+        assignedElements.forEach((item) => {
+            item.size = this.size;
+        });
+    }
+
+    protected override updated(changed: PropertyValues<this>): void {
+        super.updated(changed);
+        if (changed.has('size')) {
+            const assignedElements = [...this.items] as AccordionItem[];
```
najikahalsema, line 105: Here, too, `assignedElements` can just be `this.items`

packages/accordion/src/AccordionItem.ts
```diff
@@ -23,12 +25,19 @@ import chevronIconStyles from '@spectrum-web-components/icon/src/spectrum-icon-c
 
 import styles from './accordion-item.css.js';
 
+const chevronClass = {
```
najikahalsema, line 28:
```suggestion
const chevronIcon: Record<string, () => TemplateResult> = {```

Refer to the code for Swatch [here](https://github.com/adobe/spectrum-web-components/blob/a17290b3e1b6b0b99d68f4a77c42e9694da82982/packages/swatch/src/Swatch.ts#L34) so we can render the icon for the appropriate size up here, as opposed to in the render method.

File: packages/accordion/src/AccordionItem.ts [hunk 12]
```diff
@@ -23,12 +25,19 @@ import chevronIconStyles from '@spectrum-web-components/icon/src/spectrum-icon-c
...
+    s: 'spectrum-UIIcon-ChevronRight75',
```
najikahalsema, line 29:
```suggestion
    s: () => html`
    <sp-icon-chevron100
            class="indicator spectrum-UIIcon-Chevron75"
        ></sp-icon-chevron100>
    ````

Here's an example of how this would look.

File: packages/accordion/src/AccordionItem.ts [hunk 13]
```diff
@@ -23,12 +25,19 @@ import chevronIconStyles from '@spectrum-web-components/icon/src/spectrum-icon-c
...
+    xl: 'spectrum-UIIcon-ChevronRight300',
+};
+
 /**
  * @element sp-accordion-item
  * @slot - The content of the item that is hidden when the item is not open
  * @fires sp-accordion-item-toggle - Announce that an accordion item has been toggled while allowing the event to be cancelled.
  */
-export class AccordionItem extends Focusable {
+export class AccordionItem extends SizedMixin(Focusable) {
```
najikahalsema, line 40: Again here, let's use the argument noDefaultSize within SizedMixin so it doesn't assume it needs to apply size m by default.

pfulton:
This has been merged in Spectrum CSS, and the pre-release has been graduated to the full version. The full version is: `@spectrum-css/accordion@4.0.0`.

packages/accordion/src/AccordionItem.ts
```diff
@@ -80,9 +89,14 @@ export class AccordionItem extends Focusable {
                 >
                     ${this.label}
                 </button>
-                <sp-icon-chevron100
-                    class="indicator spectrum-UIIcon-ChevronRight100"
-                ></sp-icon-chevron100>
+                <span class="iconContainer">
+                    <sp-icon-chevron100
```
najikahalsema, line 93: Replace this with the `${chevronIcon[this.size]()}` method.

najikahalsema (changes requested):
Just a few more things.

packages/accordion/src/Accordion.ts
vjosyula: If we remove regular from here, documentation may not have this option.

packages/accordion/src/Accordion.ts
vjosyula: updated

packages/accordion/src/AccordionItem.ts
vjosyula: updated

packages/accordion/src/Accordion.ts
najikahalsema: That's okay. It's not needed. Please refer to some of [our](https://opensource.adobe.com/spectrum-web-components/components/action-button/) [other](https://opensource.adobe.com/spectrum-web-components/components/toast/) [APIs](https://opensource.adobe.com/spectrum-web-components/components/button/) to see that we don't document the "standard" option. Just make sure to fill out the JS Doc for this component so that it shows up in the API tab on the docs site.

While you're at it, please also add a segment to the documentation for the different densities, as those are new to the API.

vjosyula: updated the examples and JS Doc string.

packages/accordion/stories/accordion.stories.ts
```diff
@@ -35,66 +38,60 @@ export default {
...
+            },
+        },
     },
 };
 
 type Properties = {
     allowMultiple?: boolean;
     disabled?: boolean;
     open?: boolean;
+    density?: 'compact' | 'regular' | 'spacious';
```
najikahalsema, line 72: and here...

najikahalsema (changes requested):
So close. Just a few teeny changes.

packages/accordion/src/Accordion.ts
```diff
@@ -41,6 +45,12 @@ export class Accordion extends SpectrumElement {
     @property({ type: Boolean, reflect: true, attribute: 'allow-multiple' })
     public allowMultiple = false;
 
+    /**
+     * Sets the spacing between the content to borders of an accordion item
+     */
+    @property({ type: String, reflect: true })
+    public density: 'compact' | 'spacious' | undefined;
```
Westbrook, line 52:
```suggestion
    public density?: 'compact' | 'spacious';
```

packages/accordion/src/Accordion.ts
```diff
@@ -83,8 +93,20 @@ export class Accordion extends SpectrumElement {
         });
     }
 
-    private handleSlotchange(): void {
+    private handleSlotchange({}: Event & { target: HTMLSlotElement }): void {
```
Westbrook, line 96:
```suggestion
    private handleSlotchange(): void {
```

packages/accordion/src/Accordion.ts
```diff
@@ -83,8 +93,20 @@ export class Accordion extends SpectrumElement {
...
         this.focusGroupController.clearElementCache();
+        this.items.forEach((item) => {
+            item.size = this.size;
+        });
+    }
+
+    protected override updated(changed: PropertyValues<this>): void {
+        super.updated(changed);
+        if (changed.has('size')) {
```
Westbrook, line 105: If we don't apply a size when `m` (unless removing another value) then the attribute shouldn't be sprouted either, which is better for performance.
```suggestion
        if (changed.has('size') && (!!changed.get('size') || this.size !== 'm')) {
```

packages/accordion/src/AccordionItem.ts
```diff
@@ -68,6 +107,10 @@ export class AccordionItem extends Focusable {
         }
     }
 
+    protected renderChevronIcon = (): TemplateResult => {
+        return chevronIcon[this.size]();
```
Westbrook, line 111: If we can have no size (because we stop passing it from the parent), then we likely need a fallback here.
```suggestion
        return chevronIcon[this.size || 'm']();
```

packages/accordion/stories/accordion-densities.stories.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export default {
+    title: 'Accordion/Densities',
+    component: 'sp-accordion',
+};
+
+export const compact = (): TemplateResult =>
+    AccordionMarkup({ density: 'compact' });
+export const spacious = (): TemplateResult =>
+    AccordionMarkup({ density: 'spacious' });
```
Westbrook, line 24: Please include sized versions of each of these densities.

Westbrook (commented):
Coming together quite nicely!

Westbrook (approved):
LGTM! Thanks @venkateshjo 🙇🏼

najikahalsema (approved):
LGTM!

Issue #3313, opened 2023-06-13
[Bug][a11y]: ColorArea labelling, RTL support, vertical slider orientation


majornista:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

ColorArea, ColorSlider

### Expected behavior

ColorArea should provide a clearer indication of the component's behavior for screen reader users.

#### Labeling
By default, ColorArea should have an `aria-label` for the localized string "Color Picker", which labels the visually hidden `<input>` elements for the two color channels, or on mobile devices, the group containing them. If you wish to override this with a more specific label, a `label` attribute may be specified to further to identify the element to assistive technologies. For example, for a color area that adjusts a text color you might pass the `label` attribute, "Text color", which ColorArea will return as the `aria-label` prop "Text color, Color picker". When a custom `label` is provided, it should be localized accordingly. The internal string "Color picker" should also be localized.

ColorArea currently support `labelX` and `labelY` properties to label each of the color channel inputs as either `saturation` or `luminosity`, which is useful for passing localized strings, but it may make sense for SWC to provide the localization.

#### Role description
In order to communicate to a screen reader user that the color area adjusts in two dimensions, ColorArea should use an `aria-roledescription`, using the localized string "2D Slider", on each of the visually hidden `<input>` elements.

#### Value formatting
The `aria-valuetext` of each `<input>` element should be formatted according to the user's locale automatically. It should include the localized color channel name and current value for each channel, with the channel name and value that the `<input>` element controls coming before the channel name and value for the adjacent `<input>` element. For example, for an HSV color area where the xChannel is "saturation", the yChannel is "luminosity", when the current selected color is red (rgb(255, 0, 0)), the `<input>` representing the saturation channel will have `aria-valuetext` to announce as "100%, saturation, 100%, luminosity", and the `<input>` representing the luminosity channel will have `aria-valuetext` to announce as "100%, luminosity, 100%, saturation". 

#### RTL
In right-to-left languages, color areas should be mirrored. Orientation of the gradient background, positioning of the thumb, and dragging behavior should be automatic.

#### Vertical Slider orientation
The `input` that controls the y value, or luminosity, should be oriented so that a mobile screen reader's increment or decrement gestures behave as expected.

### Actual behavior

ColorArea labelling is ambiguous and confusing for screen reader users.

RTL orientation is not properly supported.

[...]

PR #3315, opened 2023-06-13
fix(color-area,color-slider): color-area labeling, RTL support, vertical slider orientation #3313

majornista:
## Related issue(s)

- #3313

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices


packages/color-area/src/ColorArea.ts
```diff
@@ -171,15 +183,18 @@ export class ColorArea extends SpectrumElement {
...
     }
 
-    private handleFocusin(): void {
+    private handleFocus(): void {
         this.focused = true;
+        this._valueChanged = false;
     }
 
-    private handleFocusout(): void {
+    private handleBlur(): void {
```
najikahalsema, line 191: I like the name change of these functions!

packages/color-area/src/ColorArea.ts
```diff
@@ -414,45 +475,69 @@ export class ColorArea extends SpectrumElement {
...
     }
 
     protected override firstUpdated(changed: PropertyValues): void {
         super.firstUpdated(changed);
         this.boundingClientRect = this.getBoundingClientRect();
 
-        this.addEventListener('focusin', this.handleFocusin);
-        this.addEventListener('focusout', this.handleFocusout);
+        this.addEventListener('focus', this.handleFocus);
+        this.addEventListener('blur', this.handleBlur);
```
najikahalsema, line 540: I do worry about the API name change not being consistent with our other components, though. I think this is going in the right direction, but I feel like it would be good for me to confer with the rest of the team.

packages/color-area/test/color-area.test.ts
```diff
@@ -323,9 +339,9 @@ describe('ColorArea', () => {
         });
         await elementUpdated(el);
 
-        expect(el.color).to.equal('hsl(100, 69%, 52%)');
-        expect(el.x).to.equal(0.77);
-        expect(el.y).to.equal(0.15);
+        expect(el.color).to.equal('hsl(100, 66%, 56%)');
+        expect(el.x).to.equal(0.69);
```
najikahalsema, line 343: Is there a reason why this value is different from before?

najikahalsema (commented):
I'm curious about the change in values for some of the tests. If you could explain those to me, I'd really appreciate it!

packages/color-area/test/color-area.test.ts
majornista: The RTL orientation change, causes values to be different.

majornista: `this.x` is a position that is always left-to-right. When `dir="rtl"`, we the values will change, because the orientation is reversed.

najikahalsema: Right, thank you. I actually was more so wondering why the value went from 77 to 69 here, but I _also_ remember when I wrote these tests a few years ago that getting the numbers to match is very difficult to do :) 

So that + the orientation change leading to slightly different numbers seems plausible if you also feel like those changes made sense to you as you were writing the test.

majornista: You can get a better understanding of why the tests changed by looking at the Storybook example: https://issue-3313-color-area-a11y--spectrum-web-components.netlify.app/storybook/?path=/story/color-area--default

If in LTR, you move the handle somewhere within the upper right portion of the Color Area gradient, use ArrowRight and observe the value change. Then switch to RTL. The Color Area gradient will be reversed, and the handle will now be in the upper left portion of the gradient, but the value remains unchanged. Now tab back to restore focus to the Color Area and press ArrowRight again, notice that the value changes back to where it was originally, because the orientation of the horizontal slider is now reversed.

najikahalsema (dismissed):

packages/color-area/src/ColorArea.ts
majornista: The function name for these private methods can remain `handleFocusin`/`handleFocusout`, but we should listen for the `focus` and `blur` events in order to set the `focused` state properly.

najikahalsema (approved):
LGTM! Just resolve the conflicts and it should be good to go.

Westbrook (approved):
That flakey test is on our end, :shipit:!

Thanks @majornista 🙇🏼

Issue #2560, opened 2022-09-01
[Feat]: Add Calendar pattern


Westbrook:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Calendar, Date Picker

### Description of the requested feature

Please add the Calendar pattern

cc: @najikahalsema @Jabark 

### Mockups or screenshots

https://opensource.adobe.com/spectrum-css/calendar.html

### Implementation notes or ideas

There's readable copy in this pattern. Whether we externalize that or not will be an important question. If not, we'll need to leverage https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat to hopefully get translations for free. This may point to the need to resolve language data similar to how Slider or Number Field do.

For Date management constructs it would be nice to lean closer to the pending Temporal API, but if we need to go into the user-space library field for this, definitely take a look at https://moment.github.io/luxon/#/ before something like moment.js.

PR #3327, opened 2023-06-19
feat(date-time-picker): add calendar component (WIP)

paulodetarsofm:
## Description

Adding the first component that will be part of the date/time picker: the calendar
image

## Related issue(s)

## Motivation and context

Allow choosing a date using a calendar instead of the user having to type the date

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/calendar/README.md
+++ b/packages/calendar/README.md
@@ -1,0 +1,28 @@
+## Description
+
+### Usage
+
+[![See it on NPM!](https://img.shields.io/npm/v/@spectrum-web-components/calendar?style=for-the-badge)](https://www.npmjs.com/package/@spectrum-web-components/calendar)
+[![How big is this package in your project?](https://img.shields.io/bundlephobia/minzip/@spectrum-web-components/calendar?style=for-the-badge)](https://bundlephobia.com/result?p=@spectrum-web-components/calendar)
+
+```
+yarn add @spectrum-web-components/calendar
+```
+
+Import the side effectful registration of `<sp-calendar>` via:
+
+```
+import '@spectrum-web-components/calendar/sp-calendar.js';
+```
+
+When looking to leverage the `Calendar` base class as a type and/or for extension purposes, do so via:
+
+```
+import { Calendar } from '@spectrum-web-components/calendar';
+```
+
+## Example
+
+```html
+<sp-calendar></sp-calendar>
+```
--- a/packages/calendar/package.json
+++ b/packages/calendar/package.json
@@ -34,6 +34,10 @@
             "development": "./src/index.dev.js",
             "default": "./src/index.js"
         },
+        "./src/types.js": {
+            "development": "./src/types.dev.js",
+            "default": "./src/types.js"
+        },
         "./sp-calendar.js": {
             "development": "./sp-calendar.dev.js",
             "default": "./sp-calendar.js"
@@ -57,9 +61,11 @@
         "lit-html"
     ],
     "dependencies": {
+        "@internationalized/date": "^3.2.0",
         "@spectrum-web-components/action-button": "^0.32.0",
         "@spectrum-web-components/base": "^0.32.0",
-        "@spectrum-web-components/icons-workflow": "^0.32.0"
+        "@spectrum-web-components/icons-workflow": "^0.32.0",
+        "@spectrum-web-components/reactive-controllers": "^0.32.0"
     },
     "devDependencies": {
         "@spectrum-css/calendar": "^3.1.23"
--- a/packages/calendar/package.json
+++ b/packages/calendar/package.json
@@ -1,0 +1,74 @@
+{
+    "name": "@spectrum-web-components/calendar",
+    "version": "0.0.1",
+    "publishConfig": {
+        "access": "public"
+    },
+    "description": "Web component implementation of a Spectrum design Calendar",
+    "license": "Apache-2.0",
+    "repository": {
+        "type": "git",
+        "url": "https://github.com/adobe/spectrum-web-components.git",
+        "directory": "packages/calendar"
+    },
+    "author": "",
+    "homepage": "https://adobe.github.io/spectrum-web-components/components/calendar",
+    "bugs": {
+        "url": "https://github.com/adobe/spectrum-web-components/issues"
+    },
+    "main": "src/index.js",
+    "module": "src/index.js",
+    "type": "module",
+    "exports": {
+        ".": {
+            "development": "./src/index.dev.js",
+            "default": "./src/index.js"
+        },
+        "./package.json": "./package.json",
+        "./src/Calendar.js": {
+            "development": "./src/Calendar.dev.js",
+            "default": "./src/Calendar.js"
+        },
+        "./src/calendar.css.js": "./src/calendar.css.js",
+        "./src/index.js": {
+            "development": "./src/index.dev.js",
+            "default": "./src/index.js"
+        },
+        "./sp-calendar.js": {
+            "development": "./sp-calendar.dev.js",
+            "default": "./sp-calendar.js"
+        }
+    },
+    "scripts": {
+        "test": "echo \"Error: run tests from mono-repo root.\" && exit 1"
+    },
+    "files": [
+        "**/*.d.ts",
+        "**/*.js",
+        "**/*.js.map",
+        "custom-elements.json",
+        "!stories/",
+        "!test/"
+    ],
+    "keywords": [
+        "spectrum css",
+        "web components",
+        "lit-element",
+        "lit-html"
+    ],
+    "dependencies": {
+        "@spectrum-web-components/action-button": "^0.32.0",
+        "@spectrum-web-components/base": "^0.32.0",
+        "@spectrum-web-components/icons-workflow": "^0.32.0"
+    },
+    "devDependencies": {
+        "@spectrum-css/calendar": "^3.1.23"
+    },
+    "types": "./src/index.d.ts",
+    "customElements": "custom-elements.json",
+    "sideEffects": [
+        "./sp-*.js",
+        "./**/*.dev.js",
+        "./**/*.dev.js"
+    ]
+}
--- a/packages/calendar/package.json
+++ b/packages/calendar/package.json
@@ -68,7 +68,7 @@
         "@spectrum-web-components/reactive-controllers": "^0.32.0"
     },
     "devDependencies": {
-        "@spectrum-css/calendar": "^3.1.23"
+        "@spectrum-css/calendar": "^3.1.24"
     },
     "types": "./src/index.d.ts",
     "customElements": "custom-elements.json",
--- a/packages/calendar/src/Calendar.ts
+++ b/packages/calendar/src/Calendar.ts
@@ -53,13 +53,13 @@ export class Calendar extends SpectrumElement {
     @property({ type: Boolean, reflect: true })
     disabled = false;
 
-    @property({ reflect: true })
+    @property({ reflect: true, attribute: false })
     selectedDate!: Date | undefined;
 
-    @property({ reflect: true })
+    @property({ reflect: true, attribute: false })
     min!: Date | undefined;
 
-    @property({ reflect: true })
+    @property({ reflect: true, attribute: false })
     max!: Date | undefined;
 
     @state()
--- a/packages/calendar/src/Calendar.ts
+++ b/packages/calendar/src/Calendar.ts
@@ -108,25 +108,7 @@ export class Calendar extends SpectrumElement {
         return when(
             this._currentDate,
             () => html`
-                ${this.renderCalendarHeader()}
-
-                <div
-                    class="spectrum-Calendar-body"
-                    role="grid"
-                    tabindex=${ifDefined(!this.disabled ? '0' : undefined)}
-                    aria-readonly="true"
-                    aria-disabled=${this.disabled}
-                >
-                    <table role="presentation" class="spectrum-Calendar-table">
-                        <thead role="presentation">
-                            <tr role="row">${this.renderWeekDays()}</tr>
-                        </thead>
-
-                        <tbody role="presentation">
-                            ${this.renderWeekRows()}
-                        </tbody>
-                    </table>
-                </div>
+                ${this.renderCalendarHeader()}${this.renderCalendarGrid()}
             `
         );
     }
@@ -182,38 +164,70 @@ export class Calendar extends SpectrumElement {
         `;
     }
 
-    public renderWeekDays(): TemplateResult[] {
-        return this._getWeekdays().map(
-            (weekDay) => html`
-                <th
-                    role="columnheader"
-                    scope="col"
-                    class="spectrum-Calendar-tableCell"
-                >
-                    <abbr
-                        class="spectrum-Calendar-dayOfWeek"
-                        title=${weekDay.long}
-                    >
-                        ${weekDay.narrow}
-                    </abbr>
-                </th>
-            `
-        );
+    public renderCalendarGrid(): TemplateResult {
+        return html`
+            <div
+                class="spectrum-Calendar-body"
+                role="grid"
+                tabindex=${ifDefined(!this.disabled ? '0' : undefined)}
+                aria-readonly="true"
+                aria-disabled=${this.disabled}
+            >
+                <table role="presentation" class="spectrum-Calendar-table">
+                    ${this.renderCalendarTableHead()}
+                    ${this.renderCalendarTableBody()}
+                </table>
+            </div>
+        `;
     }
 
-    public renderWeekRows(): TemplateResult[] {
-        return [...new Array(this._getWeeksInCurrentMonth()).keys()].map(
-            (weekIndex) => html`
+    public renderCalendarTableHead(): TemplateResult {
+        return html`
+            <thead role="presentation">
                 <tr role="row">
-                    ${this._getDatesInWeek(weekIndex).map((calendarDate) =>
-                        this.renderDay(calendarDate)
+                    ${this._getWeekdays().map((weekday) =>
+                        this.renderWeekdayColumn(weekday)
                     )}
                 </tr>
-            `
-        );
+            </thead>
+        `;
     }
 
-    public renderDay(calendarDate: CalendarDate): TemplateResult {
+    public renderWeekdayColumn(weekday: CalendarWeekday): TemplateResult {
+        return html`
+            <th
+                role="columnheader"
+                scope="col"
+                class="spectrum-Calendar-tableCell"
+            >
+                <abbr class="spectrum-Calendar-dayOfWeek" title=${weekday.long}>
+                    ${weekday.narrow}
+                </abbr>
+            </th>
+        `;
+    }
+
+    public renderCalendarTableBody(): TemplateResult {
+        return html`
+            <tbody role="presentation">
+                ${[...new Array(this._getWeeksInCurrentMonth()).keys()].map(
+                    (weekIndex) => this.renderCalendarTableRow(weekIndex)
+                )}
+            </tbody>
+        `;
+    }
+
+    public renderCalendarTableRow(weekIndex: number): TemplateResult {
+        return html`
+            <tr role="row">
+                ${this._getDatesInWeek(weekIndex).map((calendarDate) =>
+                    this.renderCalendarTableCell(calendarDate)
+                )}
+            </tr>
+        `;
+    }
+
+    public renderCalendarTableCell(calendarDate: CalendarDate): TemplateResult {
         const isOutsideMonth = calendarDate.month !== this._currentDate.month;
 
         const isSelected = Boolean(
--- a/packages/calendar/src/Calendar.ts
+++ b/packages/calendar/src/Calendar.ts
@@ -56,8 +56,20 @@ export class Calendar extends SpectrumElement {
     @property({ reflect: true })
     selectedDate!: Date | undefined;
 
+    @property({ reflect: true })
+    min!: Date | undefined;
+
+    @property({ reflect: true })
+    max!: Date | undefined;
+
     @state()
     private _currentDate!: CalendarDate;
+
+    @state()
+    private _minDate!: CalendarDate;
+
+    @state()
+    private _maxDate!: CalendarDate;
 
     private _languageResolver = new LanguageResolutionController(this);
     private _locale!: string;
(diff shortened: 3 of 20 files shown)


packages/calendar/package.json
```diff
@@ -62,10 +62,10 @@
     ],
     "dependencies": {
         "@internationalized/date": "^3.2.0",
-        "@spectrum-web-components/action-button": "^0.32.0",
-        "@spectrum-web-components/base": "^0.32.0",
-        "@spectrum-web-components/icons-workflow": "^0.32.0",
-        "@spectrum-web-components/reactive-controllers": "^0.32.0"
+        "@spectrum-web-components/action-button": "^0.33.2",
```
Westbrook, line 65: 👍🏼

packages/calendar/README.md
```diff
@@ -0,0 +1,35 @@
+## Description
```
Westbrook, line 1: We can add more robust docs later...but know it'll be required at some point before a full release.

packages/calendar/README.md
```diff
@@ -0,0 +1,35 @@
...
+```html
+<sp-calendar></sp-calendar>
+```
+
+## To-do list
+
+-   Attribute `role="heading"` removed from `.spectrum-Calendar-title`, due to error _'The "heading" role requires the attribute "aria-level"'_
+-   Translate _"Previous"_ and _"Next"_ used in the `title` and `aria-label` of buttons displayed in the header of the calendar
+-   Translate _"Today"_ and _"selected"_ used in the `title` of the day when it is today
+-   Implement keyboard navigation
```
Westbrook, line 35: Are these still "to do"?

packages/calendar/src/Calendar.ts
```diff
@@ -0,0 +1,449 @@
...
+    PropertyValueMap,
+    SpectrumElement,
+    TemplateResult,
+} from '@spectrum-web-components/base';
+import { LanguageResolutionController } from '@spectrum-web-components/reactive-controllers/src/LanguageResolution.js';
+
+import { property, state } from 'lit/decorators.js';
+import { classMap } from 'lit/directives/class-map.js';
+import { ifDefined } from 'lit/directives/if-defined.js';
+import { when } from 'lit/directives/when.js';
```
Westbrook, line 33: Can you take these from `@spectrum-web-components/base` in some way?

packages/calendar/src/Calendar.ts
```diff
@@ -0,0 +1,449 @@
...
+    disabled = false;
+
+    @property({ reflect: true, attribute: false })
+    selectedDate?: Date;
+
+    @property({ reflect: true, attribute: false })
+    min?: Date;
+
+    @property({ reflect: true, attribute: false })
+    max?: Date;
```
Westbrook, line 68: If you use JSDoc comments on these properties, they will appear in the API docs for this components: https://paulodetarsofm-calendar--spectrum-web-components.netlify.app/components/calendar/api/

packages/calendar/src/Calendar.ts
```diff
@@ -0,0 +1,449 @@
...
+    @property({ type: Boolean, reflect: true })
+    disabled = false;
+
+    @property({ reflect: true, attribute: false })
+    selectedDate?: Date;
+
+    @property({ reflect: true, attribute: false })
+    min?: Date;
+
+    @property({ reflect: true, attribute: false })
```
Westbrook, line 67: Reflecting without an attribute is...unexpected.

Does this value get provided from the outside? Then it likely deserves an attribute.

Does this value effect the visual delivery of the element? If not, then likely it doesn't reflect.

packages/calendar/src/Calendar.ts
```diff
@@ -0,0 +1,449 @@
...
+
+    @state()
+    private _minDate!: CalendarDate;
+
+    @state()
+    private _maxDate!: CalendarDate;
+
+    private _languageResolver = new LanguageResolutionController(this);
+    private _locale!: string;
+    private _timeZone: string = getLocalTimeZone();
```
Westbrook, line 81: We don't actually have a fixed pattern here, but TS private is better than "implied JS private". We might be better off without the `_`s here.

packages/calendar/src/Calendar.ts
```diff
@@ -0,0 +1,449 @@
...
+
+    public get today(): CalendarDate {
+        return today(this._timeZone);
+    }
+
+    constructor() {
+        super();
+
+        this._setLocale();
+        this._setInitialCalendarDate();
```
Westbrook, line 91: We definitely don't need `_` on methods...

packages/calendar/src/Calendar.ts
```diff
@@ -0,0 +1,449 @@
...
+    }
+
+    protected override willUpdate(
+        changedProperties: PropertyValueMap<this>
+    ): void {
+        this._setLocale();
```
Westbrook, line 97: Should local get set _every time_ there's an update?

packages/calendar/src/Calendar.ts
```diff
@@ -0,0 +1,449 @@
...
+        }
+
+        if (changedProperties.has('max')) {
+            this._setMaxCalendarDate();
+        }
+    }
+
+    protected override render(): TemplateResult {
+        return when(
+            this._currentDate,
```
Westbrook, line 114: How does `currentDate` get set, and why would it ever _not_ be set?

packages/calendar/src/Calendar.ts
```diff
@@ -0,0 +1,449 @@
...
+                year: 'numeric',
+            })
+        );
+
+        return html`
+            <div class="spectrum-Calendar-header">
+                <!--
+                 * TODO: Attribute 'role="heading"' removed, due to error 'The "heading" role requires the attribute
+                 * "aria-level"'
+                -->
```
Westbrook, line 134: Is this still current?

packages/calendar/src/Calendar.ts
```diff
@@ -0,0 +1,449 @@
...
+            </th>
+        `;
+    }
+
+    public renderCalendarTableBody(): TemplateResult {
+        return html`
+            <tbody role="presentation">
+                ${[...new Array(this._getWeeksInCurrentMonth()).keys()].map(
+                    (weekIndex) => this.renderCalendarTableRow(weekIndex)
+                )}
```
Westbrook, line 229: `new Array` here will likely thrash the render. Can you confirm?

If so, we should memoize these keys so that the days get rendered less often.

packages/calendar/src/Calendar.ts
```diff
@@ -0,0 +1,449 @@
...
+                        </slot>
+                    </div>
+                </sp-action-button>
+            </div>
+        `;
+    }
+
+    public renderCalendarGrid(): TemplateResult {
+        return html`
+            <!-- TODO: Implement keyboard navigation -->
```
Westbrook, line 182: See https://opensource.adobe.com/spectrum-web-components/tools/roving-tab-index/

Westbrook: Might be some things to add beyond that to really fill out the expectations here, but a good start.

packages/calendar/src/Calendar.ts
```diff
@@ -0,0 +1,449 @@
...
+
+    public renderCalendarTableCell(calendarDate: CalendarDate): TemplateResult {
+        const isOutsideMonth = calendarDate.month !== this._currentDate.month;
+
+        const isSelected = Boolean(
+            this.selectedDate &&
+                isSameDay(this._toCalendarDate(this.selectedDate), calendarDate)
+        );
+
+        const isToday = Boolean(isSameDay(calendarDate, this.today));
```
Westbrook, line 252: `isSameDay` already returns a Boolean...

packages/calendar/src/Calendar.ts
```diff
@@ -0,0 +1,449 @@
...
+    }
+
+    public handleDayClick(calendarDate: CalendarDate): void {
+        this.selectedDate = calendarDate.toDate(this._timeZone);
+        this._setCurrentCalendarDate();
+
+        this.dispatchEvent(
+            new CustomEvent('change', {
+                bubbles: true,
+                cancelable: true,
```
Westbrook, line 319: You've made this `cancelable` but don't seem to handle the case where the consumer `preventDefault()`s the event.

packages/calendar/src/Calendar.ts
```diff
@@ -0,0 +1,449 @@
...
+    private _setLocale(): void {
+        this._locale = this._languageResolver.language;
+    }
+
+    private _setInitialCalendarDate(): void {
+        this._currentDate = this.today;
+    }
+
+    private _setCurrentCalendarDate(): void {
+        if (this.selectedDate) {
```
Westbrook, line 357: Please you use an early return pattern in these methods.

Westbrook (commented):
Initial notes.

packages/calendar/src/Calendar.ts
```diff
@@ -0,0 +1,449 @@
...
+        `;
+    }
+
+    public renderCalendarGrid(): TemplateResult {
+        return html`
+            <!-- TODO: Implement keyboard navigation -->
+            <div
+                class="spectrum-Calendar-body"
+                role="grid"
+                tabindex=${ifDefined(!this.disabled ? '0' : undefined)}
```
Rajdeepc, line 186: undefined is not a valid value for the tabindex attribute

packages/calendar/src/Calendar.ts
```diff
@@ -0,0 +1,449 @@
...
+                </table>
+            </div>
+        `;
+    }
+
+    public renderCalendarTableHead(): TemplateResult {
+        return html`
+            <thead role="presentation">
+                <tr role="row">
+                    ${this._getWeekdays().map((weekday) =>
```
Rajdeepc, line 202: can you please make sure this._getWeekdays() should not return undefined

packages/calendar/src/Calendar.ts
```diff
@@ -0,0 +1,449 @@
...
+        };
+
+        const currentDayTitle = this._formatDate(calendarDate, {
+            weekday: 'long',
+            year: 'numeric',
+            month: 'long',
+            day: 'numeric',
+        });
+
+        // TODO: Translate "Today" and "selected"
```
Rajdeepc, line 274: Something you can try here. I created a pen for you for the translation logic
https://codesandbox.io/s/infallible-khayyam-cnymcs?file=/src/index.js
library used i18next

packages/calendar/src/Calendar.ts
```diff
@@ -0,0 +1,449 @@
...
+        const title: string = this._capitalFirstLetter(
+            `${todayTitle}${currentDayTitle}${selectedTitle}`
+        );
+
+        return html`
+            <td
+                role="gridcell"
+                class="spectrum-Calendar-tableCell"
+                title=${title}
+                tabindex=${ifDefined(!isOutsideMonth ? '-1' : undefined)}
```
Rajdeepc, line 287: undefined shouldn't be a fallback for tabIndex

packages/calendar/src/Calendar.ts
```diff
@@ -0,0 +1,449 @@
...
+            <tr role="row">
+                ${this._getDatesInWeek(weekIndex).map((calendarDate) =>
+                    this.renderCalendarTableCell(calendarDate)
+                )}
+            </tr>
+        `;
+    }
+
+    public renderCalendarTableCell(calendarDate: CalendarDate): TemplateResult {
+        const isOutsideMonth = calendarDate.month !== this._currentDate.month;
```
Rajdeepc, line 245: Can you specify  the type of isOutsideMonth here as Boolean?

packages/calendar/src/Calendar.ts
```diff
@@ -0,0 +1,449 @@
...
+    private _toCalendarDate(date: Date): CalendarDate {
+        return new CalendarDate(
+            date.getFullYear(),
+            date.getMonth() + 1, // The month to create a new `CalendarDate` cannot be a zero-based index, unlike `Date`
+            date.getDate()
+        );
+    }
+
+    /**
+     * Checks if the date is valid by parsing the time. Invalid dates return `NaN` for times of invalid dates
```
Rajdeepc, line 425: Can you add code comments to other methods too?

Rajdeepc (changes requested):
Thanks for bringing this to life! I have added few comments nothing major!

packages/calendar/README.md
paulodetarsofm: Yes. Most of the things involving ARIA attributes and other things related to accessibility have not been included yet, due to lack of knowledge.

packages/calendar/src/Calendar.ts
paulodetarsofm: As I mentioned [earlier](https://github.com/adobe/spectrum-web-components/pull/3327#discussion_r1267816937), with `ifDefined` Lit will not render the attribute if the value is `undefined` 😁

packages/calendar/src/Calendar.ts
paulodetarsofm: That's an interesting question. Although the locale is not changed via `@property`, it is changed when some parent element with the `lang` attribute changes the value of that attribute, and if I don't include the locale assignment here, the calendar is not automatically updated when selecting a different locale in the Storybook, for example. Do you have any suggestion?

packages/calendar/src/Calendar.ts
paulodetarsofm: Yes, all TODOs are listed in the README and where they occur

packages/calendar/src/Calendar.ts
paulodetarsofm: I didn't understand. Could you explain to me better what you mean? Are there any examples I could use as a reference?

packages/calendar/src/Calendar.ts
paulodetarsofm: But here I don't return anything, I only do something if there really is a `selectedDate`. How do you think I could improve this part?

packages/calendar/src/Calendar.ts
Rajdeepc: if the user does event.defaultPrevented how are you handling this?

packages/calendar/src/Calendar.ts
Rajdeepc: Something like with an early return pattern would be.
```
  if (!this.selectedDate) {
    return; // Early return if selectedDate is falsy (null, undefined, etc.)
  }

  this.selectedDate = new Date(this.selectedDate);

  if (!this._isValidDate(this.selectedDate)) {
    this.selectedDate = undefined;
    return; // Early return if the selectedDate is not a valid date
  }

  this._currentDate = this._toCalendarDate(this.selectedDate);
```

packages/calendar/src/Calendar.ts
```diff
@@ -0,0 +1,449 @@
...
+            if (!this._isValidDate(this.selectedDate)) {
+                this.selectedDate = undefined;
+            } else {
+                this._currentDate = this._toCalendarDate(this.selectedDate);
+            }
+        }
+    }
+
+    private _setMinCalendarDate(): void {
+        if (this.min) {
```
Rajdeepc, line 369: You can use the early return pattern here which is good in catching early errors into your methods

packages/calendar/src/Calendar.ts
```diff
@@ -0,0 +1,449 @@
...
+            if (!this._isValidDate(this.min)) {
+                this.min = undefined;
+            } else {
+                this._minDate = this._toCalendarDate(this.min);
+            }
+        }
+    }
+
+    private _setMaxCalendarDate(): void {
+        if (this.max) {
```
Rajdeepc, line 381: You can use the early return pattern here too

packages/calendar/src/Calendar.ts
paulodetarsofm: This check here is not needed

packages/calendar/src/Calendar.ts
paulodetarsofm: As I mentioned earlier, all TODOs will be resolved when the new date/time picker (which includes this calendar component) is ready to be merged into the main branch. All pending issues are being listed in the respective READMEs of each component

packages/calendar/src/Calendar.ts
paulodetarsofm: This method will not return `undefined` because the number of days in a week does not change, we will always have 7 days and, consequently, an array with 7 objects with the details of each of these days

paulodetarsofm: However, I will run this method elsewhere and use a property here instead of running the function directly.

packages/calendar/src/Calendar.ts
paulodetarsofm: I created a private property to store this information

packages/calendar/src/Calendar.ts
paulodetarsofm: I created a private property to store this information

packages/calendar/src/Calendar.ts
paulodetarsofm: Thanks for the example, Raj. At this time I will not include the translations, we will do this when the component is finalised and then all TODOs will be resolved so we can then merge the feature branch into main

packages/calendar/src/Calendar.ts
paulodetarsofm: Type boolean is inferred, ESLint would tell me to remove the type annotation
image

packages/textfield/stories/textarea.stories.ts
```diff
@@ -181,7 +181,11 @@ export const with5Rows = (): TemplateResult => html`
     <sp-textfield
         multiline
         id="predefinedRows"
-        value="Line 1\nLine 2\nLine 3\nLine 4\nLine 5"
+        value="Line 1
+Line 2
+Line 3
+Line 4
+Line 5"
```
paulodetarsofm, line 188: I don't know why some files were submitted together with my PR, they seem to be changes after linting...

packages/textfield/test/textfield.test.ts
```diff
@@ -168,7 +168,9 @@ describe('Textfield', () => {
 
         const boundsDefaultElement = defaultTextarea?.getBoundingClientRect();
         const boundsOneRowElement = oneRowTextarea?.getBoundingClientRect();
-        expect(boundsDefaultElement?.height).to.be.greaterThan(boundsOneRowElement?.height ?? 0);
+        expect(boundsDefaultElement?.height).to.be.greaterThan(
+            boundsOneRowElement?.height ?? 0
+        );
```
paulodetarsofm, line 173: I don't know why some files were submitted together with my PR, they seem to be changes after linting...

lerna.json
```diff
@@ -1,11 +1,5 @@
 {
-    "packages": [
-        "linters/*",
-        "packages/*",
-        "projects/*",
-        "tools/*",
-        "react/*"
-    ],
+    "packages": ["linters/*", "packages/*", "projects/*", "tools/*", "react/*"],
```
paulodetarsofm, line 2: I don't know why some files were submitted together with my PR, they seem to be changes after linting...

packages/checkbox/test/checkbox.test.ts
```diff
@@ -297,7 +297,6 @@ describe('Checkbox', () => {
         expect(el.indeterminate).to.be.false;
         expect(inputEl.checked).to.be.false;
         expect(inputEl.indeterminate).to.be.false;
-
```
paulodetarsofm, line 300: I don't know why some files were submitted together with my PR, they seem to be changes after linting...

packages/color-slider/test/color-slider.test.ts
```diff
@@ -229,7 +229,6 @@ describe('ColorSlider', () => {
 
         await elementUpdated(el);
 
-
```
paulodetarsofm, line 232: I don't know why some files were submitted together with my PR, they seem to be changes after linting...

packages/calendar/src/Calendar.ts
```diff
@@ -0,0 +1,511 @@
...
+                role="gridcell"
+                class="spectrum-Calendar-tableCell"
+                title=${currentDayTitle}
+                tabindex=${ifDefined(!isOutsideMonth ? '-1' : undefined)}
+                aria-disabled=${isOutsideMonth || this.disabled}
+                aria-selected=${isSelected}
+            >
+                <span
+                    role="presentation"
+                    class="spectrum-Calendar-date ${classMap(dayClasses)}"
```
Westbrook, line 316: We may want to wrap up the `spectrum-Calendar-date` in the `classMap()` application for uniformity here.

packages/calendar/src/Calendar.ts
```diff
@@ -0,0 +1,511 @@
...
+                    @click=${() => this.handleDayClick(calendarDate)}
```
Westbrook, line 317: There is a small perf win to be had to cache this call to `handleDayClick()` in a way that would allow us to pass the same method here each time. The anonymous method will be added and removed as a listener in each render.

packages/calendar/src/Calendar.ts
```diff
@@ -0,0 +1,511 @@
...
+        const numberOfWeeks = getWeeksInMonth(this.currentDate, this.locale);
+
+        this.weeksInCurrentMonth = [...new Array(numberOfWeeks).keys()];
+    }
+
+    /**
+     * Defines the array with data for the days of the week, starting on the first day of the week according to the
+     * defined location (Sunday, Monday, etc.)
+     */
+    private setWeekdays(): void {
```
Westbrook, line 361: This happens once a render, it might be clearer to make `this.weekdays` a getter that returns the work here, rather than needing to manually call `setWeekdays` each render pass. If there were an alternative path to being able to cache this so it doesn't need to be calculated so often, that would be cool, too.

packages/calendar/src/Calendar.ts
```diff
@@ -0,0 +1,511 @@
...
+                narrow: this.formatDate(date, { weekday: 'narrow' }),
+                long: this.formatDate(date, { weekday: 'long' }),
+            };
+        });
+    }
+
+    /**
+     * Defines the initial date that will be used to render the calendar, if no specific date is provided
+     */
+    private setInitialCalendarDate(): void {
```
Westbrook, line 376: Generally, I wonder if there's a way to do less imperative setting of values like this...

packages/calendar/src/Calendar.ts
```diff
@@ -0,0 +1,511 @@
...
+
+        return dates;
+    }
+
+    /**
+     * Converts a `Date` object to a `CalendarDate`
+     *
+     * @param date - `Date` object to be converted
+     */
+    private toCalendarDate(date: Date): CalendarDate {
```
Westbrook, line 471: From here down, are there any other locations you might want to use similar methods that would make sense for them to be utility methods rather than class methods?

Westbrook (commented):
Another handful of nits and notes...

Other than rebasing to try and clean up some of these unrelated changes, I'm OK with moving anything that doesn't seem like a small change here. If you can do what you're comfortable with, and move the other things to notes, we can get this merged to the project branch tomorrow at Office Hours.

Westbrook (approved):
This is good to go for the project branch. Thanks for collecting all of these to dos on top of a great foundation here!

Issue #3298, opened 2023-06-08
[Bug][a11y]: mixed-value state of swatches must be conveyed to screen readers using ARIA


dineshy87:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-swatch

### Expected behavior

Currently, mixed state is not conveyed to screen reader users. Please use aria-checked="mixed" when mixed-value is used. 

Please also note that mixed value is supported by only certain tri-state inputs such as a [checkbox](https://www.w3.org/TR/wai-aria-1.0/roles#checkbox) or [menuitemcheckbox](https://www.w3.org/TR/wai-aria-1.0/roles#menuitemcheckbox).

The mixed value is not supported on [radio](https://www.w3.org/TR/wai-aria-1.0/roles#radio) or [menuitemradio](https://www.w3.org/TR/wai-aria-1.0/roles#menuitemradio)

### Actual behavior

mixed-value is not communicated to screen reader users

najikahalsema:
Do we only want this for the `mixed-value` attribute or would the `nothing` attribute benefit from this usage, too? https://opensource.adobe.com/spectrum-web-components/components/swatch/

dineshy87:
Only mixed-value. 

nothing means no color. And i think that should be communicated by the label. That said, i have a question about labelling these swatches. Documentation doesn't have labels for these swatches even though the api allows label attribute. Does it make sense to use `color` value as the value for aria-label on swatches conditionally if `label` attribute is absent? If yes, let me know, I'll create a separate issue for that.

najikahalsema:
Good question. I think what's become clear through this audit, as well as comments we've had in the past from other consumers, is that our `label` attribute is inconsistent between components. I'm currently taking inventory of all of the differences, so it's good to know that this one will be included. I think your logic makes sense though, so I would appreciate a new issue for that :) Thank you!

Rajdeepc:
PR fix: https://github.com/adobe/spectrum-web-components/pull/3330

najikahalsema:
I was looking at Rajdeep's PR, and I have some questions about fixing the issue:

[...]

PR #3330, opened 2023-06-20
fix(swatch): mixed-value state must be conveyed to screen readers using ARIA

Rajdeepc:
## Description
Added a logic to check for mixedValue property and add aria-checked="mixed"

## Related issue(s)

- [#3298](https://github.com/adobe/spectrum-web-components/issues/3298)

## Motivation and context

Currently, mixed state is not conveyed to screen reader users. Please use aria-checked="mixed" when mixed-value is used.

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

Screenshot 2023-06-20 at 2 08 44 PM

## Types of changes

## Checklist

## Best practices


packages/swatch/README.md
```diff
@@ -116,6 +118,7 @@ import { Swatch } from '@spectrum-web-components/swatch';
         rounding="full"
         shape="rectangle"
         mixed-value
+        aria-checked="mixed"
```
Westbrook, line 121: The internal logic should be enough to cover this delivery. Adding it manually here could confuse consumers as to their responsibilities with this pattern. If for some reason these are not getting the attribute at runtime after the below change, then that’s a sign there’s more work to do.

Rajdeepc: Make sense. Thank you

packages/swatch/src/Swatch.ts
```diff
@@ -229,6 +229,11 @@ export class Swatch extends SizedMixin(Focusable, {
                 this.removeAttribute('aria-label');
             }
         }
+        if (changes.has('mixedValue')) {
+            if (this.mixedValue) {
+                this.setAttribute('aria-checked', 'mixed');
+            }
+        }
```
najikahalsema, line 236: Hey Rajdeep! Just following up on our conversation earlier today. 

To add a label to the `Swatch` that matches the `color` attribute, you can add the following in the Swatch code's willUpdate function: 

```        
if (changes.has('label')) {
            // if there is a colour set,
            if (this.color !== '') { 
                this.setAttribute('aria-label', this.color); // set the aria-label to be this.color
           // otherwise, if the label is different from the colour and is a length longer than 0,
            } else if (this.label !== this.color && this.label.length) { 
                // override the label with this.label
                this.setAttribute('aria-label', this.label); 
            }
             else {
                this.removeAttribute('aria-label');
            }
        }
```

This obviously doesn't handle situations where our value would be 'mixed' or 'nothing' (since `color=''` in that situation), but it's better than no label at all. Hopefully, we can get some guidance on @dineshy87 's side as to what kind of aria-label would be best for those situations, too. 😄 Depending on how many different types of labels we need, we may want to abstract this out of `willUpdate()`.

najikahalsema (changes requested):
Behold, the fruits of my labour!

packages/swatch/src/Swatch.ts
Rajdeepc: @najikahalsema Thanks!!

Screenshot 2023-06-23 at 3 41 23 PM

packages/swatch/README.md
```diff
@@ -202,9 +202,9 @@ The `mixed-value` attribute and `mixedValue` property outline when an `<sp-swatc
 
 ```html
 <sp-swatch-group>
-    <sp-swatch mixed-value></sp-swatch>
-    <sp-swatch mixed-value rounding="full"></sp-swatch>
-    <sp-swatch mixed-value shape="rectangle"></sp-swatch>
+    <sp-swatch mixed-value aria-checked="mixed"></sp-swatch>
```
najikahalsema, line 205: I believe that westbrook touched on this before, but you shouldn't need to add aria-checked="mixed" here since it's doing it now (you would have to add selects="multiple" to the swatch group, though)? Or maybe you could say something like "Please note that the `aria-checked="mixed"` value only applies when the swatch is in a group with `selects="multiple"`.

I believe that even if you add it manually, like you have here, it still won't display on the screen reader because the role is still `button`.

najikahalsema (changes requested):
Almost there!!!

packages/swatch/README.md
Rajdeepc: Thanks

packages/swatch/src/Swatch.ts
```diff
@@ -223,12 +223,19 @@ export class Swatch extends SizedMixin(Focusable, {
             );
         }
         if (changes.has('label')) {
-            if (this.label) {
+            if (this.color !== '') {
```
Westbrook, line 226: I _think_ you placed this logic in the wrong order and it's what's causing the test failures in CI right now. Can you take a look at those?

Rajdeepc: Should be good now except for the fact that few NumberFields tests are failing. Switching the order of if else did the trick!
Thanks

Rajdeepc:
@Westbrook @najikahalsema Let me know if this PR looks good to merge?

najikahalsema (approved):
LGTM! Thank you for your hard work 🙇

Issue #3255, opened 2023-05-31
[Bug][a11y]: In Windows High Contrast Mode, the tablist is missing its border.


dineshy87:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Tabs

### Expected behavior

When Windows High Contrast Mode is enabled, the tablist should have a bottom border when tabs are in a horizontal orientation, and a left border when tabs are in a vertical orientation, just like how it behaves in non-HCM settings.

![image](https://github.com/adobe/spectrum-web-components/assets/16927765/286c31c2-e254-4937-8cd0-f0d3e79f7e8b)

![image](https://github.com/adobe/spectrum-web-components/assets/16927765/8af6c1c0-efe0-48bf-a755-49b3c883a8b0)

### Actual behavior

When using Windows High Contrast Mode, the tablist does not display any border, making it challenging to distinguish what component it is.

![image](https://github.com/adobe/spectrum-web-components/assets/16927765/4027d863-948d-4622-a336-4b383d1aafe4)

![image](https://github.com/adobe/spectrum-web-components/assets/16927765/7cc50352-c280-434d-967b-a82bb87938c2)

### How can we reproduce this issue?

1. Go to https://opensource.adobe.com/spectrum-web-components/components/tabs/
2. Open the Chrome browser and navigate to the website you want to test.
3. Open the Chrome DevTools by pressing Ctrl+Shift+I (Windows) or Cmd+Option+I (Mac).
4. Click on the "Rendering" tab.
5. In the Rendering tab, scroll down and find the "Emulate CSS media feature forced-colors" option.
6. Select the "active" option from the drop-down menu.

pfulton:
This is great timing for us: we're currently migrating the Tabs component, and we'll be sure that this gets taken care of during that work. (cc: @mlogsdon18)

Westbrook:
fixed by https://github.com/adobe/spectrum-web-components/pull/3333

PR #3333, opened 2023-06-20
feat(tabs): update to use core tokens

mlogsdon18:
## Description

This updates the Tabs component to use the new version of Spectrum tokens. 

## Related issue(s)
- #3255 

## Motivation and context

Part of our effort to migrate all Spectrum components to the new tokens.

## Known Design Changes
- Horizontal Tabs height has decreased by 2px
- Icon size went down by 2px in width (ex: 18px to 16px for Medium)

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/top-nav/src/top-nav-item.css
+++ b/packages/top-nav/src/top-nav-item.css
@@ -20,21 +20,30 @@ a:focus {
 
 :host(:focus-within) {
     color: var(
-        --spectrum-tabs-m-text-color-key-focus,
-        var(--spectrum-alias-text-color-hover)
+        --mod-tabs-color-key-focus,
+        var(--spectrum-tabs-color-key-focus)
     ); /* .spectrum-Tabs-item.focus-ring */
 }
 
 :host(:focus-within):before {
     border-color: var(
-        --spectrum-tabs-m-focus-ring-color,
-        var(--spectrum-alias-border-color-key-focus)
+        --highcontrast-tabs-focus-indicator-color,
+        var(
+            --mod-tabs-focus-indicator-color,
+            var(--spectrum-tabs-focus-indicator-color)
+        )
     ); /* .spectrum-Tabs-item.focus-ring:before */
 }
 
 :host(:focus-within) ::slotted([slot='icon']) {
     color: var(
-        --spectrum-tabs-m-icon-color-key-focus,
-        var(--spectrum-alias-icon-color-key-focus)
+        --mod-tabs-color-key-focus,
+        var(--spectrum-tabs-color-key-focus)
     ); /* .spectrum-Tabs-item.focus-ring .spectrum-Icon */
+}
+
+@media (forced-colors: active) {
+    :host {
+        --highcontrast-tabs-focus-indicator-color: canvastext;
+    }
 }
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ executors:
 parameters:
     current_golden_images_hash:
         type: string
-        default: 5f1c3b5d2ecc0c44683517e3148819b6c93b1805
+        default: d9aee952df2073fc366c8f9673ae761da6c196a5
     wireit_cache_name:
         type: string
         default: wireit
--- a/packages/tabs/package.json
+++ b/packages/tabs/package.json
@@ -93,7 +93,7 @@
         "@spectrum-web-components/shared": "^0.33.2"
     },
     "devDependencies": {
-        "@spectrum-css/tabs": "^3.2.60"
+        "@spectrum-css/tabs": "^4.0.0"
     },
     "types": "./src/index.d.ts",
     "customElements": "custom-elements.json",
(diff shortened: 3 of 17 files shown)


mlogsdon18:
@Westbrook this one is ready for your review! I believe that all of the VRT failures are expected and mostly due to some height changes in the tabs.

Westbrook (commented):
Be sure to add `tabs` to https://github.com/adobe/spectrum-web-components/blob/main/scripts/spectrum-tokens.js#L33...

Westbrook:
@dineshy87 as this is looking to address #3255, should we be adding borders in WHCM to Quiet Tabs, even though they do not have borders to begin with?

Westbrook:
@mlogsdon18 this one looks...not right? https://db7b4652508f151a03b04517a6e1a796--spectrum-web-components.netlify.app/review/#TabsStories/verticalQuietCompact.png

Westbrook:
In other contexts, we do not use "blue" for focus loops in WHCM: https://db7b4652508f151a03b04517a6e1a796--spectrum-web-components.netlify.app/review/#TopNavStories/autofocus.png Should this be updated?

dineshy87:
> @dineshy87 as this is looking to address #3255, should we be adding borders in WHCM to Quiet Tabs, even though they do not have borders to begin with?

Yes. Adding border to quiet tabs as well in high contrast mode is a good thing. In my opinion.

mlogsdon18:
@Westbrook I updated the Top Nav to use `canvastext` for the focus indicator color like we're doing in other places for WHCM. We'll need to do a follow-up to this PR once a new release of @spectrum-web-components/tabs goes out to update the Top Nav css to use the new variables, which I can't do now since we need a new release. 

I also updated the Quiet variant to have a bottom border in WHCM. Was there anything else that needed to be updated? I know you mentioned the vertical compact tabs which I am going to have Patrick do one more release to make the selected background area a bit larger to match the previous.

packages/top-nav/src/top-nav-item.css
```diff
@@ -38,3 +41,9 @@ a:focus {
         var(--spectrum-alias-icon-color-key-focus)
     ); /* .spectrum-Tabs-item.focus-ring .spectrum-Icon */
 }
+
+@media (forced-colors: active) {
+    :host {
+        --highcontrast-tabs-focus-indicator-color: canvastext;
```
Westbrook, line 47: 🔥

packages/top-nav/src/top-nav-item.css
```diff
@@ -27,8 +27,11 @@ a:focus {
 
 :host(:focus-within):before {
     border-color: var(
-        --spectrum-tabs-m-focus-ring-color,
-        var(--spectrum-alias-border-color-key-focus)
+        --highcontrast-tabs-focus-indicator-color,
+        var(
+            --spectrum-tabs-m-focus-ring-color,
+            var(--spectrum-alias-border-color-key-focus)
```
Westbrook, line 33: Are these current to the new Tabs values?

Westbrook: I mis-read! Internal to the repo, the dependencies always resolve to what's in the current branch, so you shouldn't need a new release to make this update. If it gives you trouble, that'll be important to know as it points to other issues.

mlogsdon18: Ok this has been updated! For some reason I tried this initially and it didn't work but looks like it's good now

najikahalsema (commented):
Looks good to me! 

The one thing I did notice is that it looks like the text is slightly high up in the element by one pixel (9 at the top, 11 at the bottom). Before, it was evenly 9 px on both sides. I measured it on [this screenshot](https://db7b4652508f151a03b04517a6e1a796--spectrum-web-components.netlify.app/review/#TabsVerticalSizesStories/m.png), as well as in a non-WHCM story in the storybook. It looks like it occurs in both horizontal and vertical cases... I don't know if this is something we want to bring back to the Spectrum team?

mlogsdon18:
@najikahalsema Adding this in here too so it doesn't get lost in slack. 

Could you specify what you're measuring when you say "slightly high up in the element by one pixel". What is the "element" that you're basing the measurements on? I ask because depending on the size, according to the design, the spacing below the text is supposed to be one pixel greater than the spacing above (including the selection indicator)
Screenshot 2023-06-30 at 2 34 23 PM

najikahalsema:
> @najikahalsema Adding this in here too so it doesn't get lost in slack.
> 
> Could you specify what you're measuring when you say "slightly high up in the element by one pixel". What is the "element" that you're basing the measurements on? I ask because depending on the size, according to the design, the spacing below the text is supposed to be one pixel greater than the spacing above (including the selection indicator) Screenshot 2023-06-30 at 2 34 23 PM

I wasn't aware of this design spec, so if that's the case, it's good with me 👍

pfulton:
We've merged in the CSS PR, and graduated the release. The full release is: `@spectrum-css/tabs@4.0.0`

Westbrook (approved):
Thanks for bringing this one home for us @mlogsdon18! :shipit:

PR #3345, opened 2023-06-21
chore: chevron sizing approach [accordion fork]

castastrophe:
@vjosyula I left some inline notes to clarify the approaches, let me know if this works for you & @najikahalsema.

### DIFF
--- a/packages/accordion/src/Accordion.ts
+++ b/packages/accordion/src/Accordion.ts
@@ -13,6 +13,7 @@ governing permissions and limitations under the License.
 import {
     CSSResultArray,
     html,
+    SizedMixin,
     SpectrumElement,
     TemplateResult,
 } from '@spectrum-web-components/base';
@@ -30,12 +31,7 @@ import styles from './accordion.css.js';
  * @element sp-accordion
  * @slot - The sp-accordion-item children to display.
  */
-
-import { SizedMixin } from '@spectrum-web-components/base';
-
-export class Accordion extends SizedMixin(SpectrumElement, {
-    validSizes: ['s', 'm', 'l', 'xl'],
-}) {
+export class Accordion extends SizedMixin(SpectrumElement) {
     public static override get styles(): CSSResultArray {
         return [styles];
     }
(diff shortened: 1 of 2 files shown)


packages/accordion/src/Accordion.ts
```diff
@@ -30,12 +31,7 @@ import styles from './accordion.css.js';
  * @element sp-accordion
  * @slot - The sp-accordion-item children to display.
  */
-
-import { SizedMixin } from '@spectrum-web-components/base';
```
castastrophe, line 34: @vjosyula I moved this import up with the others sourced from the base.

packages/accordion/src/Accordion.ts
```diff
@@ -91,23 +87,22 @@ export class Accordion extends SizedMixin(SpectrumElement, {
         });
     }
 
-    private handleSlotchange(): void {
+    private handleSlotchange({
```
castastrophe, line 90: @vjosyula @najikahalsema I sourced this technique from the button group which has a similar parent-child relationship with button. It passes the size value down to it's children during slot changes.

packages/accordion/src/Accordion.ts
```diff
@@ -91,23 +87,22 @@ export class Accordion extends SizedMixin(SpectrumElement, {
...
-    private setAccordionProperties(): void {
-        this.items.forEach((item) => {
+        const assignedElements = slot.assignedElements() as AccordionItem[];
+        assignedElements.forEach((item) => {
             item.size = this.size;
         });
     }
 
     protected override render(): TemplateResult {
-        this.setAccordionProperties();
```
castastrophe, line 105: @vjosyula Setting property values during render can cause cycles so it is typically avoided or not allowed.  In this case, the setAccordionProperties would be setting properties on the slotted items at render time which is not ideal. I moved this work up into the slot change handler so that it can process at the time the slotted item is registered. Let me know if this works for you @najikahalsema!

packages/accordion/src/AccordionItem.ts
```diff
@@ -92,10 +90,13 @@ export class AccordionItem extends Focusable {
...
-                        chevronClass[this.size as DefaultElementSize]
-                    }"
-                ></sp-icon-chevron100>
+                    <sp-icon-chevron100
+                        class="indicator ${chevronClass[
+                            this.size as DefaultElementSize
+                        ]}"
+                        slot="icon"
+                    ></sp-icon-chevron100>
+                </span>
```
castastrophe, line 99: Was missing a closing span tag.

najikahalsema (approved):
Looks great. Thanks for this, Cassondra. Once this gets merged to the other branch, it'd be great to get it rebased with main.

packages/accordion/src/AccordionItem.ts
vjosyula: thank you for pointing it out

vjosyula (approved):
Thanks @castastrophe , changes looks good to me.

Issue #3348, opened 2023-06-21
[Bug][a11y]: SideNav should express hierarchy using list and listitem


majornista:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

SideNav, SideNavHeading, SideNavItem

### Expected behavior

1. SideNav should support a `label` and `labelledby` so that the `<nav>` element in the shadowDom can be distinguished from any other `nav` elements in a web page.
2. SideNav should wrap its `slot[name="descendant"]` element in a `div[role="list"]`, so that SideNav items will be contained within a hierarchical list.
3. SideNavHeading should have a default `role` property of `listitem` :
```ts
    @property({ reflect: true })
    public role = 'listitem';
```
4. SideNavHeading shadowDom element with `id="list"` should also have `role="list"`, so that it the list hierarchy will be appropriately communicated to assistive technology.
```ts
    protected override render(): TemplateResult {
        return html`
            <h2 id="heading">${this.label}</h2>
            <div id="list" aria-labelledby="heading" role="list">
                <slot name="descendant"></slot>
            </div>
        `;
    }
```
5. SideNavItem should have a default `role` property of `listitem` :
```ts
    @property({ reflect: true })
    public role = 'listitem';
```
6. SideNavItem should wrap its `slot[name="descendant"]` element in a `div[role="list"]`, so that descendant SideNavItems will be contained within a hierarchical list, and the SideNavItem link in shadowDom should include `aria-expanded` and `aria-controls` when `this.hasChildren === true` to express the current state of the expandable node:
```ts
    protected override render(): TemplateResult {
        return html`
            <a
                href=${this.href || '#'}
                target=${ifDefined(this.target)}
                download=${ifDefined(this.download)}
                rel=${ifDefined(this.rel)}
                data-level="${this.depth}"
                @click="${this.handleClick}"
                id="item-link"
                aria-current=${ifDefined(
                    this.selected && this.href ? 'page' : undefined
                )}
                aria-expanded=${ifDefined(
                    this.hasChildren ? this.expanded : undefined
                )}
                aria-controls=${ifDefined(
                    this.hasChildren && this.expanded ? 'list' : undefined
                )}
            >
                <slot name="icon"></slot>
                ${this.label}
                <slot></slot>
            </a>
            ${this.expanded
                ? html`
                    <div id="list" aria-labelledby="item-link" role="list">
                        <slot name="descendant"></slot>
                    </div>
                  `
                : html``}
[...]

PR #3352, opened 2023-06-22
fix(sidenav): [Bug][a11y]: sidenav should express hierarchy using list and listitem #3348

majornista:
## Description

1. SideNav should support a `label` and `labelledby` so that the `<nav>` element in the shadowDom can be distinguished from any other `nav` elements in a web page.
2. SideNav should wrap its `slot[name="descendant"]` element in a `div[role="list"]`, so that SideNav items will be contained within a hierarchical list.
3. SideNavHeading should have a default `role` property of `listitem` :
```ts
    @property({ reflect: true })
    public role = 'listitem';
```
4. SideNavHeading shadowDom element with `id="list"` should also have `role="list"`, so that it the list hierarchy will be appropriately communicated to assistive technology.
```ts
        return html`
            <h2 id="heading">${this.label}</h2>
            <div id="list" aria-labelledby="heading" role="list">
                <slot name="descendant"></slot>
            </div>
        `;
    }
```
5. SideNavItem should have a default `role` property of `listitem` :
```ts
    @property({ reflect: true })
    public role = 'listitem';
```
6. SideNavItem should wrap its `slot[name="descendant"]` element in a `div[role="list"]`, so that descendant SideNavItems will be contained within a hierarchical list, and the SideNavItem link in shadowDom should include `aria-expanded` and `aria-controls` when `this.hasChildren === true` to express the current state of the expandable node:
```ts
        return html`
            <a
                href=${this.href || '#'}
                target=${ifDefined(this.target)}
                download=${ifDefined(this.download)}
                rel=${ifDefined(this.rel)}
                data-level="${this.depth}"
                @click="${this.handleClick}"
                id="item-link"
                aria-current=${ifDefined(
                    this.selected && this.href ? 'page' : undefined
                )}
                aria-expanded=${ifDefined(
                    this.hasChildren ? this.expanded : undefined
                )}
                aria-controls=${ifDefined(
                    this.hasChildren && this.expanded ? 'list' : undefined
                )}
            >
                <slot name="icon"></slot>
                ${this.label}
                <slot></slot>
            </a>
            ${this.expanded
                ? html`
                    <div id="list" aria-labelledby="item-link" role="list">
                        <slot name="descendant"></slot>
                    </div>
                  `
                : html``}
        `;
    }
```
 7. The SideNavItem link should remain tabbable/focusable when it is collapsed, so that a user can toggle the expanded/collapsed state of descendant list.
 ```ts
        if (this.hasChildren && this.expanded && !this.selected) {
            this.focusElement.tabIndex = 0;
        } else {
            this.focusElement.removeAttribute('tabindex');
        }
        super.updated(changes);
    }
 ```

## Related issue(s)

-  #3348

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices


packages/sidenav/src/Sidenav.ts
```diff
@@ -162,11 +181,17 @@ export class SideNav extends Focusable {
...
         return html`
-            <nav @sidenav-select=${this.handleSelect}>
-                <slot
-                    name="descendant"
-                    @slotchange=${this.handleSlotchange}
-                ></slot>
+            <nav
+                @sidenav-select=${this.handleSelect}
+                aria-label=${ifDefined(this.label)}
+                aria-labelledby=${ifDefined(this.labelledby)}
```
Westbrook, line 187: This won't point through the shadow boundary and isn't worth applying at this time. Future specs may open the door for this type of label application, at which point we can revisit these sorts of patterns.

packages/sidenav/src/SidenavItem.ts
```diff
@@ -44,6 +44,9 @@ export class SideNavItem extends LikeAnchor(Focusable) {
     @property({ type: Boolean, reflect: true })
     public expanded = false;
 
+    @property({ reflect: true })
+    public role = 'listitem';
```
Westbrook, line 48: Can you see a path towards this ever _not_ being "listitem"? If not, we should probably cover over it by not making it a public property. If so, should we document cases where a user would override this?

packages/sidenav/src/SidenavItem.ts
```diff
@@ -129,22 +132,32 @@ export class SideNavItem extends LikeAnchor(Focusable) {
...
+                      </div>
                   `
                 : html``}
         `;
     }
 
     protected override updated(changes: PropertyValues): void {
         if (this.hasChildren && this.expanded && !this.selected) {
-            this.focusElement.tabIndex = -1;
+            this.focusElement.tabIndex = 0;
```
Westbrook, line 158: This allows you to tab into any item, rather the leverage roving tab index. Was that indended?

projects/documentation/content/_includes/partials/sidenav.njk
```diff
@@ -2,7 +2,7 @@
 {% if componentName %}
   {% set value = componentName %}
 {% endif %}
-<sp-sidenav manage-tab-index slot="side-nav" value="{{ value }}">
+<sp-sidenav manage-tab-index slot="side-nav" value="{{ value }}" label="Spectrum Web Components">
```
Westbrook, line 5: 👍🏼

packages/sidenav/src/Sidenav.ts
```diff
@@ -81,6 +82,24 @@ export class SideNav extends Focusable {
...
+    public label?: string | undefined = undefined;
+
+    /**
+     * Identifies the element (or elements) that labels the component,
+     * so that the side navigation can be distinguished
+     * from other navigation by screen reader users.
+     * It will be applied to aria-labelledby, but not visually rendered.
+     */
+    @property({ reflect: true })
+    public labelledby?: string | undefined = undefined;
```
Westbrook, line 101: If we absolutely what this type of accessor, we'd need to resolve the text content of any relating element and then apply that string to label, do the connections across the shadow boundary being blocked. Do you think that's necessary, or could we leave this property out of the API?

packages/sidenav/src/SidenavItem.ts
majornista: It should probably be private, but if it is, can it still be reflected?

packages/sidenav/src/SidenavItem.ts
majornista: We want to be able to focus the item that toggles expand/collapse of children.

packages/sidenav/src/Sidenav.ts
majornista: We can omit `labelledby`. I actually had trouble associating the side nav on the docs with the Spectrum Web Components logo, because everything was contained within a web component.

packages/sidenav/src/SidenavItem.ts
Westbrook: In `firstUpdated` we can do `this.setAttribute(‘role', ‘listitem’)`, if there _shouldn’t_ be user overrides.

packages/sidenav/src/SidenavItem.ts
Westbrook: Do we want it _in_ the tab order? Or just about to do it programmatically through the Roving Tab Stop pattern? Making sure I understand the appropriate accessibility model here.

packages/sidenav/src/SidenavItem.ts
majornista: That's what I've done.

packages/sidenav/src/SidenavItem.ts
majornista: I think it's more complicated than that. When `manageTabIndex` is `false`, we want to be able to navigate to the parent item when the children are expanded, so that it can be collapsed. But, for the `manageTabIndex` use case, we need to account for when the selected item is a child of a collapsed item, which I address below:

https://github.com/adobe/spectrum-web-components/pull/3352/files#diff-0343271c00070ee3f19ceebd95378dd6c0d7c2a115c3bceb3cb082721ea6f226R56-R77
```ts
    rovingTabindexController = new RovingTabindexController<SideNavItem>(this, {
        focusInIndex: (elements: SideNavItem[]) => {
            let parentSideNavItem: SideNavItem | undefined;
            let index = elements.findIndex((el) => {
                // If the selected item's parent is collapsed, save it for later.
                if (el.value === this.value && this.isDisabledChild(el)) {
                    parentSideNavItem = el.closest(
                        'sp-sidenav-item:not([expanded])'
                    ) as SideNavItem;
                }
                return this.value
                    ? !el.disabled &&
                          !this.isDisabledChild(el) &&
                          el.value === this.value
                    : !el.disabled && !this.isDisabledChild(el);
            });
            // If the selected item's parent is collapsed, focus the collapsed parent.
            if (index === -1 && parentSideNavItem) {
                index = elements.findIndex((el) => el === parentSideNavItem);
            }
            return index;
        },
        direction: 'vertical',
        elements: () =>
            [...this.querySelectorAll('sp-sidenav-item')] as SideNavItem[],
        isFocusableElement: (el: SideNavItem) =>
            !el.disabled && !this.isDisabledChild(el),
    });
```

Westbrook (approved):
LGTM! Thanks @majornista 🙇🏼

PR #3369, opened 2023-06-27
fix(action-group): bug fixed where it causes revert selection of button on click 

sonirajan:
Creating Draft PR just to review changes. If everything looks good, I'll update description of this PR:

Consider the case where `ActionGroup` has multiple `ActionButton` and `ActionGroup` is controlled component. When I click one of the `ActionButton`, in [onClick](https://github.com/adobe/spectrum-web-components/blob/main/packages/action-button/src/ActionButton.ts#L123) event, ActionButton already changes `selected` property of ActionButton. Later, when `handleClick` event is fired on `ActionGroup`, we unnecessarily changes ActionButton's `selected` property to opposite of current state which is already changed before in ActionButton. This causes issue and later ActionButton's `value` property's value [is not added](https://github.com/adobe/spectrum-web-components/pull/3369/files#diff-b86e1ad24ea1fd0c7ca6cc46a20dd4f7e8626d26f8aa77fccbeb388cf42c9d1dR195) in `this.selected` array and thus ActionButton selection UI is not working. 

I tested this changes in ccx-commenting filter component and it works fine. But, that breaks action group's 3 unit tests. I was wondering if this change is valid or not. Or, if there is any better solution. Please have a look when you get a chance. Thanks a lot.

image

## Related issue(s)

-

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/action-group/src/ActionGroup.ts
+++ b/packages/action-group/src/ActionGroup.ts
@@ -187,7 +187,6 @@ export class ActionGroup extends SizedMixin(SpectrumElement, {
             }
             case 'multiple': {
                 const selected = [...this.selected];
-                target.selected = !target.selected;
                 target.setAttribute(
                     'aria-checked',
                     target.selected ? 'true' : 'false'


sonirajan:
I added `toggles` attribute to `ActionButton` even though `ActionGroup` is controlled component. Once I removed `toggles`, ActionButton `onClick` event [returned](https://github.com/adobe/spectrum-web-components/blob/main/packages/action-button/src/ActionButton.ts#L121) and won't change `selected` property of `ActionButton`, rather selected property was changed in ActionGroup `handleClick` [event](https://github.com/adobe/spectrum-web-components/blob/main/packages/action-group/src/ActionGroup.ts#L190).

Issue #3364, opened 2023-06-26
[Bug][a11y]: deletable tags must support deleting a tag using delete or backspace key


dineshy87:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-tags

### Expected behavior

Spectrum design specifies that users should be able to press delete or backspace button to delete a deletable tag. Once implemented, make sure the keyboard focus moves on to the next non-disabled tag within the sp-tags.

For reference, check [react-spectrum implementation](https://react-spectrum.adobe.com/react-spectrum/TagGroup.html#contextual-help)

### Actual behavior

The above behavior is currently not supported. 

### What browsers are you seeing the problem in?

All

### How can we reproduce this issue?

1. Go to https://opensource.adobe.com/spectrum-web-components/components/tags/
2. Try deleting tags using keyboard.

Related issue:
https://jira.corp.adobe.com/browse/CSS-552

Rajdeepc:
PR: https://github.com/adobe/spectrum-web-components/pull/3391

blunteshwar:
The `mutationController` is calling the callback fucntion `handleItemMutaion` whenever there is a mutation in the host element whether the mutation is of the type `delete`, `add` or `update`. Thus some tests of sp-grid are failing

Issue #3091, opened 2023-04-06
[Bug]: Handle changing availability of "default" item in Focus Group


Westbrook:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Reactive Controllers

### Expected behavior

If I change the availability (add `disabled` or `tabindex="-1"`) of the "default" item (first or selected), a subsequent item should become default allowing the Focus Group to continue to exist within the tab order.

### Actual behavior

Making the "default" item unavailable removed the Focus Group from the tab order complete.

castastrophe:
We have this work tracked now in Jira so I'm going to close. Please do reach out if you still need a status update on this feature.

PR #3391, opened 2023-06-29
fix(tag): add delete functionality in sp-tag

Rajdeepc:
## Description
Added delete functionality on sp-tag on click and also on backspace/delete

## Related issue(s)

- https://github.com/adobe/spectrum-web-components/issues/3364

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/tags/src/Tag.ts
+++ b/packages/tags/src/Tag.ts
@@ -99,7 +99,6 @@ export class Tag extends SizedMixin(SpectrumElement, {
         if (this.readonly) {
             return;
         }
-        this.shiftFocusOnEvent();
         const deleteEvent = new Event('delete', {
             bubbles: true,
             composed: true,
@@ -109,11 +108,8 @@ export class Tag extends SizedMixin(SpectrumElement, {
         if (deleteEvent.defaultPrevented) {
             return;
         }
-
-        const parent = this.parentNode as HTMLElement | null;
-        if (parent) {
-            this.remove();
-        }
+        this.shiftFocusOnEvent();
+        this.remove();
     }
 
     protected override render(): TemplateResult {
--- a/packages/tags/src/Tag.ts
+++ b/packages/tags/src/Tag.ts
@@ -75,8 +75,8 @@ export class Tag extends SizedMixin(SpectrumElement, {
             return;
         }
         const { code } = event;
-
-        const tags = Array.from(this.parentNode.querySelectorAll('sp-tag'));
+        const parent = this.parentNode as HTMLElement | null;
+        const tags = Array.from(parent?.querySelectorAll('sp-tag') ?? []);
         const currentIndex = tags.indexOf(this);
 
         switch (code) {
--- a/packages/tags/src/Tag.ts
+++ b/packages/tags/src/Tag.ts
@@ -132,7 +132,7 @@ export class Tag extends SizedMixin(SpectrumElement, {
 
         const parent = this.parentNode as HTMLElement | null;
         if (parent) {
-            parent.removeChild(this);
+            this.remove();
             this.focusAfterDelete(parent);
         }
     }
--- a/packages/tags/src/Tag.ts
+++ b/packages/tags/src/Tag.ts
@@ -70,36 +70,6 @@ export class Tag extends SizedMixin(SpectrumElement, {
         this.removeEventListener('focusout', this.handleFocusout);
     };
 
-    /**
-     * This function, named 'focusAfterDelete', is used to handle the focus behavior after
-     * deleting an element within a parent element.
-     * @param parent
-     */
-    private focusAfterDelete = (parent: HTMLElement): void => {
-        const tags = Array.from(parent?.querySelectorAll('sp-tag') ?? []);
-        const currentIndex = tags.indexOf(this);
-        let focusIndex = currentIndex;
-        if (currentIndex < tags.length - 1) {
-            for (let i = currentIndex + 1; i < tags.length; i++) {
-                if (!tags[i].disabled) {
-                    focusIndex = i;
-                    break;
-                }
-            }
-        } else if (currentIndex > 0) {
-            for (let i = currentIndex - 1; i >= 0; i--) {
-                if (!tags[i].disabled) {
-                    focusIndex = i;
-                    break;
-                }
-            }
-        }
-        if (focusIndex !== currentIndex) {
-            const focusTag = tags[focusIndex];
-            focusTag.focus();
-        }
-    };
-
     private handleKeydown = (event: KeyboardEvent): void => {
         if (!this.deletable || this.disabled) {
             return;
@@ -116,10 +86,20 @@ export class Tag extends SizedMixin(SpectrumElement, {
         }
     };
 
+    private shiftFocusOnEvent(): void {
+        const event = new CustomEvent('sp-tag-focus', {
+            bubbles: true,
+            composed: true,
+            cancelable: true,
+        });
+        this.dispatchEvent(event);
+    }
+
     private delete(): void {
         if (this.readonly) {
             return;
         }
+        this.shiftFocusOnEvent();
         const deleteEvent = new Event('delete', {
             bubbles: true,
             composed: true,
@@ -133,7 +113,6 @@ export class Tag extends SizedMixin(SpectrumElement, {
         const parent = this.parentNode as HTMLElement | null;
         if (parent) {
             this.remove();
-            this.focusAfterDelete(parent);
         }
     }
 
--- a/packages/tags/src/Tag.ts
+++ b/packages/tags/src/Tag.ts
@@ -31,6 +31,7 @@ import styles from './tag.css.js';
  * @slot avatar - an avatar element to display within the Tag
  * @slot icon - an icon element to display within the Tag
  */
+export let nextSibling: HTMLElement;
 export class Tag extends SizedMixin(SpectrumElement, {
     validSizes: ['s', 'm', 'l'],
 }) {
@@ -99,6 +100,9 @@ export class Tag extends SizedMixin(SpectrumElement, {
         if (deleteEvent.defaultPrevented) {
             return;
         }
+        nextSibling =
+            (this.nextElementSibling as HTMLElement) ||
+            (this.previousElementSibling as HTMLElement);
         this.remove();
     }
 
--- a/packages/tags/src/Tag.ts
+++ b/packages/tags/src/Tag.ts
@@ -108,12 +108,9 @@ export class Tag extends SizedMixin(SpectrumElement, {
 
         switch (code) {
             case 'Backspace':
+            case 'Space':
             case 'Delete':
                 this.delete();
-                break;
-            case 'Space':
-            case 'Tab':
-                break;
             default:
                 return;
         }
--- a/packages/tags/src/Tag.ts
+++ b/packages/tags/src/Tag.ts
@@ -86,15 +86,6 @@ export class Tag extends SizedMixin(SpectrumElement, {
         }
     };
 
-    private shiftFocusOnEvent(): void {
-        const event = new CustomEvent('sp-tag-focus', {
-            bubbles: true,
-            composed: true,
-            cancelable: true,
-        });
-        this.dispatchEvent(event);
-    }
-
     private delete(): void {
         if (this.readonly) {
             return;
@@ -108,7 +99,6 @@ export class Tag extends SizedMixin(SpectrumElement, {
         if (deleteEvent.defaultPrevented) {
             return;
         }
-        this.shiftFocusOnEvent();
         this.remove();
     }
 
--- a/packages/tags/src/Tag.ts
+++ b/packages/tags/src/Tag.ts
@@ -31,7 +31,6 @@ import styles from './tag.css.js';
  * @slot avatar - an avatar element to display within the Tag
  * @slot icon - an icon element to display within the Tag
  */
-export let nextSibling: HTMLElement;
 export class Tag extends SizedMixin(SpectrumElement, {
     validSizes: ['s', 'm', 'l'],
 }) {
@@ -91,18 +90,15 @@ export class Tag extends SizedMixin(SpectrumElement, {
         if (this.readonly) {
             return;
         }
-        const deleteEvent = new Event('delete', {
-            bubbles: true,
-            composed: true,
-        });
-        this.dispatchEvent(deleteEvent);
-
-        if (deleteEvent.defaultPrevented) {
+        const applyDefault = this.dispatchEvent(
+            new Event('delete', {
+                bubbles: true,
+                composed: true,
+            })
+        );
+        if (!applyDefault) {
             return;
         }
-        nextSibling =
-            (this.nextElementSibling as HTMLElement) ||
-            (this.previousElementSibling as HTMLElement);
         this.remove();
     }
 
--- a/packages/tags/src/Tag.ts
+++ b/packages/tags/src/Tag.ts
@@ -90,11 +90,22 @@ export class Tag extends SizedMixin(SpectrumElement, {
         if (this.readonly) {
             return;
         }
-        this.dispatchEvent(
-            new Event('delete', {
-                bubbles: true,
-            })
-        );
+        const deleteEvent = new Event('delete', {
+            bubbles: true,
+            composed: true, // Allow the event to propagate across the shadow boundary
+        });
+        this.dispatchEvent(deleteEvent);
+
+        if (deleteEvent.defaultPrevented) {
+            // The event handler canceled the deletion, so return
+            return;
+        }
+
+        // Remove the Tag element from its parent
+        const parent = this.parentNode;
+        if (parent) {
+            parent.removeChild(this);
+        }
     }
 
     protected override render(): TemplateResult {
--- a/packages/tags/src/Tag.ts
+++ b/packages/tags/src/Tag.ts
@@ -70,41 +70,46 @@ export class Tag extends SizedMixin(SpectrumElement, {
         this.removeEventListener('focusout', this.handleFocusout);
     };
 
+    /**
+     * This function, named 'focusAfterDelete', is used to handle the focus behavior after
+     * deleting an element within a parent element.
+     * @param parent
+     */
+    private focusAfterDelete = (parent: HTMLElement): void => {
+        const tags = Array.from(parent?.querySelectorAll('sp-tag') ?? []);
+        const currentIndex = tags.indexOf(this);
+        let focusIndex = currentIndex;
+        if (currentIndex < tags.length - 1) {
+            for (let i = currentIndex + 1; i < tags.length; i++) {
+                if (!tags[i].disabled) {
+                    focusIndex = i;
+                    break;
+                }
+            }
+        } else if (currentIndex > 0) {
+            for (let i = currentIndex - 1; i >= 0; i--) {
+                if (!tags[i].disabled) {
+                    focusIndex = i;
+                    break;
+                }
+            }
+        }
+        if (focusIndex !== currentIndex) {
+            const focusTag = tags[focusIndex];
+            focusTag.focus();
+        }
+    };
+
     private handleKeydown = (event: KeyboardEvent): void => {
         if (!this.deletable || this.disabled) {
             return;
         }
         const { code } = event;
-        const parent = this.parentNode as HTMLElement | null;
-        const tags = Array.from(parent?.querySelectorAll('sp-tag') ?? []);
-        const currentIndex = tags.indexOf(this);
 
         switch (code) {
             case 'Backspace':
             case 'Delete':
                 this.delete();
-
-                let focusIndex = currentIndex;
-                if (currentIndex < tags.length - 1) {
-                    for (let i = currentIndex + 1; i < tags.length; i++) {
-                        if (!tags[i].disabled) {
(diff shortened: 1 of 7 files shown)


packages/tags/src/Tag.ts
```diff
@@ -70,17 +70,47 @@ export class Tag extends SizedMixin(SpectrumElement, {
         this.removeEventListener('focusout', this.handleFocusout);
     };
 
+    /**
+     * This function, named 'focusAfterDelete', is used to handle the focus behavior after
+     * deleting an element within a parent element.
+     * @param parent
+     */
+    private focusAfterDelete = (parent: HTMLElement): void => {
```
Westbrook, line 78: This feels pretty similar to what we might need in https://github.com/adobe/spectrum-web-components/issues/3091. Could we delegate this work to Focus Group rather than building it into Tag?

packages/tags/src/Tag.ts
```diff
@@ -90,11 +120,21 @@ export class Tag extends SizedMixin(SpectrumElement, {
...
+            bubbles: true,
+            composed: true,
+        });
+        this.dispatchEvent(deleteEvent);
+
+        if (deleteEvent.defaultPrevented) {
+            return;
+        }
+
+        const parent = this.parentNode as HTMLElement | null;
```
Westbrook, line 133: It’s possible that this causes issues when the Tag is a direct child of a shadow root. We likely want `.remove()` instead.

packages/tags/src/Tag.ts
Rajdeepc: Make sense.

packages/tags/src/Tag.ts
Rajdeepc: Done

tools/reactive-controllers/src/FocusGroup.ts
```diff
@@ -163,6 +163,26 @@ export class FocusGroupController<T extends HTMLElement>
         }
     }
 
+    changeDefaultItemFocus(): void {
```
Westbrook, line 166: I think it would make better sense for this to be a reaction to something we can observe rather than something we call imperatively. Can you think of paths to leverage a [Mutation Observer](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) here? A great way to do so in this context is via [this Lit Labs package](https://github.com/lit/lit/blob/main/packages/labs/observers/src/mutation-controller.ts).

packages/tags/src/Tags.ts
```diff
@@ -104,9 +110,27 @@ export class Tags extends FocusVisiblePolyfillMixin(SpectrumElement) {
         this.rovingTabindexController.clearElementCache();
     }
 
+    private async changeDefaultFocus(event: Event): Promise<void> {
+        await 0;
```
Westbrook, line 114: What's this `await` for?

packages/tags/src/Tags.ts
```diff
@@ -104,9 +110,27 @@ export class Tags extends FocusVisiblePolyfillMixin(SpectrumElement) {
...
+        if (event.defaultPrevented) {
+            // No deleting when user prevents it.
+            return;
+        }
+        const items = [...this.tags] as Tag[];
+        /* c8 ignore next 3 */
+        if (items && !items.length) {
+            // no deleting when there aren't items.
+            return;
+        }
```
Westbrook, line 124: This implies we're only reacting to the element removing itself, but there are other contexts that we should take into account here.
- when a user "prevents default" that will likely be in conjunction with the Tags component being controlled, so the element _may_ still be removed as part of the workflow
- the element becomes `disabled`
- etc.

packages/tags/src/Tags.ts
```diff
@@ -104,9 +110,27 @@ export class Tags extends FocusVisiblePolyfillMixin(SpectrumElement) {
         this.rovingTabindexController.clearElementCache();
     }
 
+    private async changeDefaultFocus(event: Event): Promise<void> {
+        await 0;
+        if (event.defaultPrevented) {
+            // No deleting when user prevents it.
+            return;
+        }
+        const items = [...this.tags] as Tag[];
```
Westbrook, line 119: This should already be an array. Is there a special benefit for copying it like this?

packages/tags/src/Tag.ts
```diff
@@ -71,30 +71,49 @@ export class Tag extends SizedMixin(SpectrumElement, {
...
     private delete(): void {
         if (this.readonly) {
             return;
         }
-        this.dispatchEvent(
-            new Event('delete', {
-                bubbles: true,
-            })
-        );
+        this.shiftFocusOnEvent();
```
Westbrook, line 102: _If_ the right decision is to do this imperatively, shouldn't it go after the remove? Or at least after default isn't prevented?

tools/reactive-controllers/src/RovingTabindex.ts
```diff
@@ -74,6 +74,26 @@ export class RovingTabindexController<
...
+            for (let i = currentIndex + 1; i < this.elements.length; i++) {
+                focusIndex = i;
+                break;
+            }
+        } else if (currentIndex > 0) {
+            for (let i = currentIndex - 1; i >= 0; i--) {
+                focusIndex = i;
+                break;
+            }
+        }
```
Westbrook, line 90: Without logic inside of the `for` statements, won't this always get simply the next or previous item, regardless of whether those items are focusable or not?

Also, if `(currentIndex < this.elements.length - 1)` and all subsequent items are NOT focusable, what element should become focusable?

packages/tags/src/Tag.ts
```diff
@@ -71,30 +71,45 @@ export class Tag extends SizedMixin(SpectrumElement, {
...
     };
 
+    private shiftFocusOnEvent(): void {
+        const event = new CustomEvent('sp-tag-focus', {
+            bubbles: true,
+            composed: true,
+            cancelable: true,
+        });
+        this.dispatchEvent(event);
+    }
```
Westbrook, line 96: This tightly binds the workflow to Tag/Tags, is there a more generic observation method that would apply to all consumers of FocusGroup and RovingTabIndex?

packages/tags/src/Tags.ts
```diff
@@ -104,9 +104,29 @@ export class Tags extends FocusVisiblePolyfillMixin(SpectrumElement) {
...
+            return;
+        }
+        // check if tag item is disabled
+        this.tags.forEach((item) => {
+            if (item.disabled) {
+                return;
+            }
+        });
+        this.rovingTabindexController.changeDefaultItemFocus();
+    }
```
Westbrook, line 122: Again, this workflow works when the element is removed. Please also test if the element becomes `disabled`.

tools/reactive-controllers/src/RovingTabindex.ts
Rajdeepc: 1. No the next or previous item doesn't get the focus without the logic inside `for` statements
2. If the current index is not the last element and if all the next items are not focusable then the focus should remain on the last element if its not disabled?

packages/tags/src/Tag.ts
Rajdeepc: Right so do you want this method to go inside RovingTabIndex so that we can leverage it as `this.rovingTabIndexController.shiftFocus()`.

tools/reactive-controllers/src/FocusGroup.ts
Rajdeepc: Yes now we are observing this method instead of calling it imperatively.
```
            config: {
                childList: true,
                subtree: true,
            },
            callback: () => {
                this.rovingTabindexController.changeDefaultItemFocus();
            },
        });
```

packages/tags/src/Tags.ts
Rajdeepc: Also this is not needed as of now. Shifted to Mutation Observer

tools/reactive-controllers/src/RovingTabindex.ts
Westbrook: What I thinking, and maybe this is just a side effect of me only reviewing this first thing in the morning, is that if `currentIndex === 5` and `this.elements.length === 10`, the logic above will also set `focusIndex` as `6` whether it's focusable or not?

Similarly, `currentIndex === 5` and `this.elements.length === 6`, this will also set `focusIndex` as `4`... 

Maybe resolving past that is managed elsewhere?

packages/tags/src/Tags.ts
```diff
@@ -55,6 +56,15 @@ export class Tags extends FocusVisiblePolyfillMixin(SpectrumElement) {
...
         super();
+        new MutationController(this, {
+            config: {
+                childList: true,
+                subtree: true,
+            },
+            callback: () => {
+                this.rovingTabindexController.changeDefaultItemFocus();
+            },
+        });
```
Westbrook: This is definitely the direction that I'm thinking!

One step further, if this was a part of `FocusGroup` would it "just work" for every pattern leveraging this, e.g. the Action Group listed in the related issue?

packages/tags/src/Tag.ts
```diff
@@ -90,11 +90,16 @@ export class Tag extends SizedMixin(SpectrumElement, {
...
+        const deleteEvent = new Event('delete', {
+            bubbles: true,
+            composed: true,
+        });
+        this.dispatchEvent(deleteEvent);
+
+        if (deleteEvent.defaultPrevented) {
+            return;
+        }
+        this.remove();
```
Westbrook:
```suggestion
        const applyDefault = this.dispatchEvent(new Event('delete', {
            bubbles: true,
            composed: true,
        }));

        if (!applyDefault) {
            return;
        }
        this.remove();
```
This normalizes this pattern to other usage in the library.

packages/tags/src/Tag.ts
```diff
@@ -71,16 +71,16 @@ export class Tag extends SizedMixin(SpectrumElement, {
     };
 
     private handleKeydown = (event: KeyboardEvent): void => {
-        if (!this.deletable) {
+        if (!this.deletable || this.disabled) {
```
Westbrook, line 75: 👍🏼

blunteshwar (dismissed):

packages/tags/src/Tag.ts
```diff
@@ -31,6 +31,7 @@ import styles from './tag.css.js';
  * @slot avatar - an avatar element to display within the Tag
  * @slot icon - an icon element to display within the Tag
  */
+export let nextSibling: HTMLElement;
```
Westbrook, line 34: We shouldn't rely on this shared state.

packages/tags/src/Tags.ts
```diff
@@ -70,7 +70,6 @@ export class Tags extends FocusVisiblePolyfillMixin(SpectrumElement) {
     private handleKeydown = (event: KeyboardEvent): void => {
         const { code } = event;
         if (code !== 'PageUp' && code !== 'PageDown') return;
-
```
Westbrook, line 73: Can we clean up this white space change?

tools/reactive-controllers/src/FocusGroup.ts
```diff
@@ -145,6 +155,24 @@ export class FocusGroupController<T extends HTMLElement>
         );
     }
 
+    changeDefaultItemFocus(): void {
```
Westbrook, line 158: Please test that deleting something that is _NOT_ currently focused doesn't have negative effects on the focus here.

If would be great if we could do some work with the `disabled` attribute, but it is fine if we hold that change for a follow up PR.

tools/reactive-controllers/src/FocusGroup.ts
```diff
@@ -145,6 +155,24 @@ export class FocusGroupController<T extends HTMLElement>
...
+        this.clearElementCache();
+        let diff = 0;
+        if (this.currentIndex === this.elements.length) {
+            diff = -1;
+        }
```
Westbrook, line 163:
```suggestion
        const diff = this.currentIndex === this.elements.length ? -1 : 1;
```
Does this achieve the same?

tools/reactive-controllers/src/FocusGroup.ts
Westbrook, line 158: Naming is hard, does this feel like it's clear about what it actually does? Is it `shouldUpdateActiveFocus`? I'm not sure.

A JSdoc comment would also clarify, if a good name isn't immediately apparent.

tools/reactive-controllers/src/FocusGroup.ts
blunteshwar: nope, it should be 
`const diff = this.currentIndex === this.elements.length ? -1 : 0;`

tools/reactive-controllers/src/FocusGroup.ts
```diff
@@ -145,6 +155,24 @@ export class FocusGroupController<T extends HTMLElement>
...
+        this.setCurrentIndexCircularly(diff);
+        let steps = this.elements.length;
+        while (
+            steps > 0 &&
+            !this.isFocusableElement(this.elements[this.currentIndex])
+        ) {
+            this.setCurrentIndexCircularly(1);
+            steps--;
+        }
+        this?.elements[this?.currentIndex]?.focus();
```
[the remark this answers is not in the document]
blunteshwar: yes this is same as this.focus, it calls focus on the immediate sibling of the deleted element

packages/tags/src/Tags.ts
blunteshwar: done

tools/reactive-controllers/src/FocusGroup.ts
blunteshwar: can you elaborate this? I didn't understand it

blunteshwar: done

tools/reactive-controllers/src/FocusGroup.ts
blunteshwar: done

tools/reactive-controllers/src/FocusGroup.ts
Rajdeepc: This make sense we can add an attribute property in Mutation Observer config to check for updates which have disabled property and not do this on the method level. I can write a new ticket for this.

blunteshwar: Great!

tools/reactive-controllers/src/FocusGroup.ts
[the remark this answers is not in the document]
Rajdeepc: Instead can you use this.elements to be always truthy to go inside this.elements[this.currentIndex].focus()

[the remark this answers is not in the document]
Westbrook: Is all of the the same as?
```
        this.clearElementCache();
        cost moveToNextElement = this.currentIndex !== this.elements.length;
        const diff = moveToNextElement ? 1 : -1;
        if (moveToNextElement) {
            // Account for the item at the current index being removed.
            this.currentIndex -= 1;
        }
        this.setCurrentIndexCircularly(diff);
        this.focus();
```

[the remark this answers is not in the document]
Westbrook: I'm not sure we see `this?.elements[this?.currentIndex]?.focus();` processing the same way.

I see it as the equivalent to:
```
if (this) {
   if (this && this.elements[this.currentIndex]) {
       this.element[this.currentIndex].focus();
   }
}
```


That's different than `this.element?.[this.currentIndex].focus()` or `this.element?.[this.currentIndex]?.focus()`, which would be more like:
```
if (this.elements) {
    this.elements[this.currentIndex].focus();
// or the latter
    if (this.elements[this.currentIndex]) {
        this.elements[this.currentIndex].focus();
    }
}
```

I don't use optional chaining all that much, but it's good for us to ensure we all understand it the same way.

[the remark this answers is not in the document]
blunteshwar: Nope, let me document the code here with comments

        `this.clearElementCache(); ` // clear chace for the olde elemets bcz one element is deleted so new cache has to be created

        `const diff = this.currentIndex === this.elements.length ? -1 : 0; `// if the deleted element was the last one set dfiff =-1 else diff =0. We are doing this because if deleted element is the last element we have to decrease the currIndex by 1 but if it is not the last one we dont have to explicitly increment it.

      `  this.setCurrentIndexCircularly(diff);` // we set the currIndex using the function and diff as argument

      ```
  let steps = this.elements.length;
        while (
            steps > 0 &&
            !this.isFocusableElement(this.elements[this.currentIndex])
        ) {
            this.setCurrentIndexCircularly(1);
            steps--;
        }
```

once the element is deleted the focus gets shift to the next successor or to the predecessor (if deleted element was the last one), in the above while loop we are checking that the element on which the focus has shifted, weather its focusable or not. I its not focusable we shift the focus to its successor

        `this?.elements[this?.currentIndex]?.focus();` focusing on the corresponding element

**Comment by Rajdeepc** [comment]
@Westbrook The mutationController is calling the callback function `handleItemMutaion` whenever there is a mutation in the host element whether the mutation is of the type delete, add or update. Thus some tests of `sp-grid` are failing. Please advise what would be the best foot forward. Though I think keeping it in FocusGroup is a future proof solution

**Comment by blunteshwar** [comment]
@Westbrook  I have pushed some changes and tried to generalise the "mutation controller".
Can you have a look at that

File: tools/reactive-controllers/src/FocusGroup.ts [hunk 19]
```diff
@@ -144,6 +154,27 @@ export class FocusGroupController<T extends HTMLElement>
             this._listenerScope
         );
     }
+    /* In  handleItemMutation() method the first if condition is checking if the element is not focused or if the element's children's length is not decreasing then it means no element has been deleted and we must return.
```
Rajdeepc, line 157: Needs some formatting here

tools/reactive-controllers/src/FocusGroup.ts
```diff
@@ -153,10 +184,13 @@ export class FocusGroupController<T extends HTMLElement>
     }
 
     focus(options?: FocusOptions): void {
-        let focusElement = this.elements[this.currentIndex];
+        const elements = this.elements;
+        if (!elements.length) return;
+        let focusElement = elements[this.currentIndex];
+        //console.log(this.currentIndex);
```
Rajdeepc, line 190: remove this

tools/reactive-controllers/src/FocusGroup.ts
```diff
@@ -144,6 +154,27 @@ export class FocusGroupController<T extends HTMLElement>
             this._listenerScope
         );
     }
+    /* In  handleItemMutation() method the first if condition is checking if the element is not focused or if the element's children's length is not decreasing then it means no element has been deleted and we must return.
+    Then we are checking if the deleted element was the focused one before the deletion if so then we need to proceed else we casn return;
+     */
+    handleItemMutation(): void {
+        if (
+            this._currentIndex == -1 ||
+            this.elements.length <= this._elements.length
```
Westbrook, line 163: `_elements` is always a method `private _elements!: () => T[];`, it must be called for `.length` to appropriately be compared here.

blunteshwar: I didn't exactly understand this?

tools/reactive-controllers/src/FocusGroup.ts
blunteshwar: which line?

tools/reactive-controllers/src/FocusGroup.ts
Westbrook: The types and values of `this.elements` and `this._elements` are different:
```
console.log(this.elements); // [element, element]
console.log(this.elements.length); // 2
console.log(this._elements); // undefined
console.log(this._elements.length); // 0
```
So the length of `this.elements` would never (in normal use) be less than the length of `this._elements`.

blunteshwar: done

tools/reactive-controllers/src/FocusGroup.ts
```diff
@@ -144,6 +154,24 @@ export class FocusGroupController<T extends HTMLElement>
...
+    /*  In  handleItemMutation() method the first if condition is checking if the element is not focused or if the element's children's length is not decreasing then it means no element has been deleted and we must return.
+        Then we are checking if the deleted element was the focused one before the deletion if so then we need to proceed else we casn return;
+    */
+    handleItemMutation(): void {
+        if (this._currentIndex == -1) return;
+        const lengthOfElementsBeforeMutation = this.elements.length;
+        const focusedElement = this.elements[this.currentIndex];
+        this.clearElementCache();
+        if (lengthOfElementsBeforeMutation <= this.elements.length) return;
+        if (this.elements.includes(focusedElement)) return;
```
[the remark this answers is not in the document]
blunteshwar: done!

tools/reactive-controllers/src/FocusGroup.ts
Westbrook: The comment, I think.

blunteshwar: done!

Westbrook (approved):
LGTM! Thanks for working through all the nuance here. 🙇🏼

PR #3392, opened 2023-06-29
fix(picker): correct label application for screen readers

### LINKED ISSUE #3238: [Bug][a11y]:  Picker: VoiceOver fails to announce label in Chrome
### Code of conduct

### Impacted component(s)

Picker

### Expected behavior

The button element in the sp-picker component's shadow DOM is labeled internally using aria-labelledby:
1. by the button itself, which has an aria-label that comes from the textContent of the associated sp-field-label component,
2. by the icon slot within the shadow DOM,
3. and by the label slot within the shadow DOM, that will contain either the selected value or the assigned placeholder text.

VoiceOver should announce the menu button using the accessibility name assembled from elements referenced in the aria-labelledby attribute, but instead it only announces the component description, "menu popup, collapsed button."

### Actual behavior

Steps to reproduce the problem:
1. With VoiceOver on, open https://opensource.adobe.com/spectrum-web-components/components/picker/#sizes.
2. Navigate to focus the Dropdown menu button, or Picker, labelled "Selection type:" with a placeholder value "Selection type" under the heading Sizes and within the Medium tab panel.
3. VoiceOver announces the Dropdown menu button as "menu popup, collapsed button," without including the accessibility name, which, if you look at the Accessibility properties panel in the Dev tools, does seem to be properly assembled and communicated as "Selection type: Selection type." 

The issue does not reproduced using Safari or Webkit.

### Related Chromium bug

[Issue 1448092: VoiceOver fails to announce label for web component menu button](https://bugs.chromium.org/p/chromium/issues/detail?id=1448092)

Chrome, Microsoft Edge

1. With VoiceOver on, open https://opensource.adobe.com/spectrum-web-components/components/picker/#sizes.
2. Navigate to focus the Dropdown menu button, or Picker, labelled "Selection type:" with a placeholder value "Selection type" under the heading Sizes and within the Medium tab panel.
3. VoiceOver announces the Dropdown menu button as "menu popup, collapsed button," without including the accessibility name, which, if you look at the Accessibility properties panel in the Dev tools, does seem to be properly assembled and communicated as "Selection type: Selection type." 

The issue does not reproduced using Safari or Webkit.

See CodePen example: https://codepen.io/majornista/pen/BaqMQJg

Westbrook:
## Description
At some point Chrome apparently decided that you can use IDREFs to elements who's parent has `aria-label`. This updates the labelling patter for Picker to account for this. I didn't do a larger refactor of Field Label application into custom elements, because I'd like to believe that we can wait for up coming browser APIs before we do that. It may be appropriate at a intermediate date to move for a well known property to a well known method and then allow the labelled element to _always_ decide how to apply the label, which would prevent the need to hijack the native DOM method from the outside like this does.

## Related issue(s)
- fixes #3238

## Motivation and context
#a11y

## How has this been tested?

-   [ ] _Test case 1_
    1. Go [here](https://picker-label--spectrum-web-components.netlify.app/storybook/?path=/story/picker--default) in Chrome in macOS
    2. Turn on Voice Over
    3. See that the Picker is appropriately announced\

## Types of changes

## Checklist


packages/picker/src/Picker.ts
```diff
@@ -406,6 +411,17 @@ export class PickerBase extends SizedMixin(Focusable) {
...
 
+    customSetAttribute = (attribute: string, value: string): void => {
+        if (attribute !== 'aria-label') {
+            HTMLElement.prototype.setAttribute.apply(this.button, [
+                attribute,
+                value,
+            ]);
+            return;
+        }
+        this.appliedLabel = value;
```
hunterloftis, line 422: Could `this.appliedLabel` be assigned alongside `this.value` whenever the selected value changes, instead of overriding `setAttribute`?

If not, like maybe a timing issue for VO to pick up the new value, a comment here explaining this approach would probably be useful down the line.

hunterloftis (approved):
Works in chrome + macos

packages/picker/src/Picker.ts
Westbrook: @hunterloftis I swapped the approaches here to something that I think is less abusive.

packages/field-label/src/FieldLabel.ts
```diff
@@ -89,25 +92,33 @@ export class FieldLabel extends SizedMixin(SpectrumElement) {
...
+            if (typeof applyLabel !== 'undefined') {
+                applyLabel();
+            } else if (targetParent === (this.getRootNode() as HTMLElement)) {
+                conditionAttributeWithoutId(focusable, 'aria-labelledby', [
                     this.id,
                 ]);
             } else {
-                this.target.removeAttribute('aria-label');
+                focusable.removeAttribute('aria-label');
             }
```
hunterloftis, line 122: Is this identical (minus `this.labelText -> undefined`) to `addTarget`? If so, you could extract this to something like `setTarget(value: string)`, then `addTarget` could call `setTarget(this.labelText)` and `removeTarget` could call `setTarget(undefined)`. Unnecessary, just something to consider.

hunterloftis (dismissed):

PR #3410, opened 2023-07-04
fix(tabs): allow bi-directional arrow key navigation in both orientations

### LINKED ISSUE #3263: [Bug][a11y]: Screen reader users perceive no distinction between horizontal and vertical orientation
### Code of conduct

### Impacted component(s)

Tabs

### Expected behavior

Since we can't rely on screen readers announcing the tablist orientation to determine whether to use ArrowUp/ArrowDown instead of ArrowLeft/ArrowRight. Both ArrowUp and ArrowLeft should work to navigate to the previous tab, and both ArrowDown and ArrowRight should work to navigate to the next tab.

### Actual behavior

With focus on a Tab, a screen reader user would have to guess the orientation to figure out which arrow keys will work to navigate to the next or previous Tab.

### Screenshots

_No response_

_No response_

_No response_

_No response_

_No response_

TarunAdobe:
## Description

Updated the logic to accept event code on arrow key press in TabGroup.ts allowing users to use both ArrowLeft and ArrowUp keys interchangeably (same for ArrowDown and ArrowRight)

## Related issue(s)

fixes #3263 

## Motivation and context

Currently a screen reader user would have to guess the orientation to figure out which arrow keys will work to navigate to the next or previous Tab since we can't rely on screen readers announcing the tablist orientation.

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/tools/reactive-controllers/src/FocusGroup.ts
+++ b/tools/reactive-controllers/src/FocusGroup.ts
@@ -230,9 +230,9 @@ export class FocusGroupController<T extends HTMLElement>
         }
         switch (this.direction) {
             case 'horizontal':
-                return code.startsWith('Arrow');
+                return code === 'ArrowLeft' || code === 'ArrowRight';
             case 'vertical':
-                return code.startsWith('Arrow');
+                return code === 'ArrowUp' || code === 'ArrowDown';
             case 'both':
             case 'grid':
                 return code.startsWith('Arrow');
--- a/tools/reactive-controllers/src/FocusGroup.ts
+++ b/tools/reactive-controllers/src/FocusGroup.ts
@@ -230,9 +230,9 @@ export class FocusGroupController<T extends HTMLElement>
         }
         switch (this.direction) {
             case 'horizontal':
-                return code === 'ArrowLeft' || code === 'ArrowRight';
+                return code.startsWith('Arrow');
             case 'vertical':
-                return code === 'ArrowUp' || code === 'ArrowDown';
+                return code.startsWith('Arrow');
             case 'both':
             case 'grid':
                 return code.startsWith('Arrow');
--- a/packages/tabs/src/Tabs.ts
+++ b/packages/tabs/src/Tabs.ts
@@ -183,8 +183,7 @@ export class Tabs extends SizedMixin(Focusable) {
             });
             return firstFocusableElement ? focusInIndex : -1;
         },
-        direction: () =>
-            this.direction === 'horizontal' ? 'horizontal' : 'vertical',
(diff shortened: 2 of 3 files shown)


Rajdeepc:
@TarunAdobe Please check the tests on Tabs. I think these are failing because of your changes.

tools/reactive-controllers/src/FocusGroup.ts
```diff
@@ -230,9 +230,9 @@ export class FocusGroupController<T extends HTMLElement>
         }
         switch (this.direction) {
             case 'horizontal':
-                return code === 'ArrowLeft' || code === 'ArrowRight';
+                return code.startsWith('Arrow');
             case 'vertical':
-                return code === 'ArrowUp' || code === 'ArrowDown';
+                return code.startsWith('Arrow');
```
Westbrook, line 235: Based on the conversation in the issue this stems from, I'm still not sure we want to flatten _ALL_ usage of this controller or if we want to just leverage `both` in Tabs, so that it can be accurately supported. Can you take a look at leveraging that instead?

If really _nothing_ should be `horizontal` or `vertical`, then that's another question. Do we auto fallback to both like this? And then use Dev Mode warnings to clarify why the functionality isn't what might be expected by the input, or do we leave `horizontal` and `vertical` as is, change all of our consumption to `both`, and then Dev Mode warn that `horizontal` and `vertical` may not deliver the expected screen reader/keyboard experience.

Westbrook (commented):
Which ever path we take, you additional tests should help support that decision, so nice addition there!

tools/reactive-controllers/src/FocusGroup.ts
TarunAdobe: That is actually a good point to consider here... I feel though that this should apply to pretty much all the components leveraging the controller right (not just Tabs) as the functionality we are expecting should be consistent? But sure I can try to leverage the both functionality in just the tabs.

TarunAdobe: @Westbrook  Based on the conversation in issue should I make the changes such that this functionality of using both Down/RightArrow and Up/LeftArrow works specifically with tablist only... I'd start by removing the changes from FocusGroup.ts. Let me know if that makes sense?

Westbrook (approved):
Thanks for simplifying this! We'll get down to the real guts of this change with the a11y team later.

PR #3416, opened 2023-07-05
feat(opacity-checkerboard): adding new component

vjosyula:
## Related issue(s)

-

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices


Westbrook:
Is there an issue where you’ve outlined the goals, API, and usage of an addition like this?

Westbrook:
There's still no issue or consumption instructions in this change. Please add that somewhere before I can productively review these changes.

jenndiaz:
I worked on the CSS for this component, just wanted to share the context I have. It is not in the design docs site but is in contributions (https://spectrum-contributions.corp.adobe.com/page/opacity-checkerboard-beta/) 

This component is used by Swatch, ColorSlider, Thumbnail, and ColorHandle to display alpha transparency. 
Here is the PR on spectrum-css to add and implement the component: https://github.com/adobe/spectrum-css/pull/1916

Westbrook:
@jenndiaz based off of the info in that PR, I'm not sure that building this as a custom element is the right call here. This seems like it is more of a shared CSS rule(s) rather than shared functionality.

@vjosyula could you look into leveraging this similar to how the [`clear-button` package](https://github.com/adobe/spectrum-web-components/tree/main/packages/clear-button) or the [various icon styles](https://github.com/adobe/spectrum-web-components/blob/main/packages/icon/src/spectrum-icon-arrow.css)? We've got Office Hours this Wednesday at 8am ET, if you wanted to discuss that path more closely.

packages/opacity-checkerboard/package.json
```diff
@@ -0,0 +1,54 @@
...
+    ],
+    "keywords": [
+        "spectrum css",
+        "web components",
+        "lit-element",
+        "lit-html"
+    ],
+    "dependencies": {},
+    "devDependencies": {
+        "@spectrum-css/opacitycheckerboard": "^1.0.0-alpha.0"
```
Westbrook, line 46: Is this package scheduled for a stable release?

packages/opacity-checkerboard/src/spectrum-config.js
```diff
@@ -0,0 +1,37 @@
...
+const config = {
+    conversions: [
+        {
+            inPackage: '@spectrum-css/opacitycheckerboard',
+            outPackage: 'opacity-checkerboard',
+            fileName: 'opacity-checkerboard',
+            components: [
+                converter.classToClass(
+                    'spectrum-OpacityCheckerboard',
+                    'opacityCheckerboard'
```
Westbrook, line 30: Please convert this to kabob case.

Westbrook (commented):
Great start. If we had docs, I'd be happy to merge to `main` with the couple of notes here. A project branch path with actual implementation to educate the documentation process also works great here. Feel free to take which ever path you're more comfortable with here.

packages/opacity-checkerboard/src/spectrum-config.js
vjosyula: updated

Rajdeepc:
@Westbrook I have updated the docs with the usage part to the actual documentation. Let me know if you are expecting any more changes on this.

Rajdeepc (approved):

Westbrook:
My question here is do we find [this page](https://venkateshjo-css-547--spectrum-web-components.netlify.app/components/opacity-checkerboard/) helpful to a developer looking to leverage this new feature?

For instance, in #3507 it's noted that this feature is not currently leveraged in that pattern. If we were to ask the contributor there to leverage this, would the above link be enough for them to do so?

Rajdeepc:
I noticed clear button is not added as a part of the SWC components list, is there any specific reason we are adding opacity Checkerboard?
If we want to add this in the component list, we can create an example like Clear button inside either of ColorHandle, ColorSlider, Swatch and Thumbnail or in all of them and add it as usage examples. 
Some specs will be required to achieve this.
cc @jnjosh

Westbrook:
Most times when there is documentation missing (which is too often), it happens because no one ever asked for it loud enough. I'd much rather documentation be missing than not really usable, but the line there is much more personal that I can really define. If we chose to exclude this from the docs, we could remove the README.md, and then the contents are no longer at question.

I'd like to say that I specifically chose not to include Clear Button in the docs because "it's only ever used in the Clear Button element", but I'd be lying if I said I had thought that far ahead. The CSS values are not documented, and if we intended for them to be used in apps, then we should absolutely get an issue open to expand the docs to include that. What I can say, the Clear Button and Close Button package are primarily left overs from a time when we couldn't consume more than one CSS package in a single SWC package (previously they were all part of the Button package from CSS, which is why the custom element definitions are where they are). It might be time to deprecate those packages and resume consume the CSS for them directly in Button, where [Clear Button and Close Button are documented as custom elements.](https://opensource.adobe.com/spectrum-web-components/components/button/)

As a consumer of SWC, I do know that we've copy/pasted the previous opacity CSS into other projects in the past, which points to me wanting the contents of this package documented at some point. I don't want to block this from moving forward, so I'll follow your lead here, either remove the README.md so the package is not included in the docs or expand the README.md so that usage instructions and the like are included, and we can get this merged. Sound good?

packages/opacity-checkerboard/src/spectrum-config.js
```diff
@@ -0,0 +1,37 @@
...
+const config = {
+    conversions: [
+        {
+            inPackage: '@spectrum-css/opacitycheckerboard',
+            outPackage: 'opacity-checkerboard',
+            fileName: 'opacity-checkerboard',
+            components: [
+                converter.classToClass(
+                    'spectrum-OpacityCheckerboard',
+                    'OpacityCheckerboard'
```
Westbrook, line 30:
```suggestion
                    'opacity-checkerboard'
```

packages/opacity-checkerboard/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+ColorHandle, ColorSlider, Swatch and Thumbnail
+
+## Usage
+
+Import the styles from `opacity-checkerboard` css
+
+```
+import opacityCheckeryBoardStyles from '@spectrum-web-components/opacity-checkerboard/src/opacity-checkerboard.css.js';
+
```
Westbrook, line 15:
```suggestion
```

packages/opacity-checkerboard/README.md
```diff
@@ -0,0 +1,33 @@
+## Description
+
+This package is part of the [Spectrum CSS project](https://github.com/adobe/spectrum-css).
+
+Opacity checkerboard is intended to be used within other components, including:
+
+ColorHandle, ColorSlider, Swatch and Thumbnail
```
Westbrook, line 7:
```suggestion
The `opacity-checkerboard` class is used to highlight opacity. Leverage these styles in your component as outlined below to unify you application's visuals with those delivered by various Spectrum Web Components.
```

packages/opacity-checkerboard/package.json
```diff
@@ -0,0 +1,54 @@
+{
+    "name": "@spectrum-web-components/opacity-checkerboard",
+    "version": "1.0.0",
```
Westbrook, line 3:
```suggestion
    "version": "0.0.1",
```

Westbrook (commented):
Please move this from the packages directory to the tools directory.

Westbrook (approved):
Made some nuanced changes to the docs site to better outline what this does for others.

Also, it seems like you never ran `yarn` to install the package and build the `spectrum-*.css` file, but I'm not sure.

See the changes [here](https://github.com/adobe/spectrum-web-components/pull/3416/commits/a7cfc75abfefc2d632be3f4e51f23dd028d4ad2d#diff-dc09b5e88e95d6a927a867b1dc7adad674590255539760056a5a7215eccb71b5).

Thanks for getting this together! Hopefully this means that we can see this leveraged in some patterns, soon?

PR #3461, opened 2023-07-18
fix(textfield): add support for [grows] when [multiline][quiet]

Westbrook:
## Description
Allow `<sp-textfield multiline quiet>` to actively accept the `grows` attribute. At some point in the past, style delivery in this area blocked this from delivering correctly, but some combination of time and/or Core Tokens have unlocked this feature.

## How has this been tested?
-   [ ] _Test case 1_
    1. Go [here](https://textarea-quiet-grows--spectrum-web-components.netlify.app/components/textarea/#grows)
    2. See that the `quiet` Textfield (the furthest right) now takes the full height of its content.
    3. Type additional content and see that the visible area grows
    4. Delete content and see that the visible area shrinks
-   [ ] _Test case 2_
    1. Go [here](https://textarea-quiet-grows--spectrum-web-components.netlify.app/storybook/index.html?path=/story/textarea--grows-quiet)
    2. See that the Textfield now takes the full height of its content.
    3. Type additional content and see that the visible area grows
    4. Delete content and see that the visible area shrinks

## Types of changes

## Checklist

### DIFF
--- a/packages/textfield/src/Textfield.ts
+++ b/packages/textfield/src/Textfield.ts
@@ -58,7 +58,7 @@ export class TextfieldBase extends ManageHelpText(
     @property({ type: Boolean, reflect: true })
     public focused = false;
 
-    @query('.input')
+    @query('.input:not(#sizer)')
     protected inputElement!: HTMLInputElement | HTMLTextAreaElement;
 
     @property({ type: Boolean, reflect: true })
@@ -225,9 +225,11 @@ export class TextfieldBase extends ManageHelpText(
 
     private get renderMultiline(): TemplateResult {
         return html`
-            ${this.grows && !this.quiet && this.rows === -1
+            ${this.grows && this.rows === -1
                 ? html`
-                      <div id="sizer">${this.value}&#8203;</div>
+                      <div id="sizer" class="input" aria-hidden="true">
+                          ${this.value}&#8203;
+                      </div>
                   `
                 : nothing}
             <!-- @ts-ignore -->
--- a/packages/textfield/src/textfield.css
+++ b/packages/textfield/src/textfield.css
@@ -44,15 +44,6 @@ textarea {
     forced-color-adjust: var(--swc-test-forced-color-adjust);
 }
 
-:host([grows]:not([quiet])) .input {
-    position: absolute;
-    top: 0;
-    left: 0;
-    height: 100%;
-    resize: none;
-    overflow: hidden;
-}
-
 :host([grows]:not([quiet])) #textfield:after {
     grid-area: unset;
     min-block-size: calc(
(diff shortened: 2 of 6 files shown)


packages/textfield/stories/textarea.stories.ts
```diff
@@ -94,6 +94,18 @@ export const grows = (): TemplateResult => html`
     ></sp-textfield>
 `;
 
+export const growsQuiet = (): TemplateResult => html`
+    <sp-field-label for="story">Enter your life story...</sp-field-label>
+    <sp-textfield
+        multiline
+        id="story"
+        value="Sed ut perspiciatis unde omnis iste natus error sit voluptatem accusantium doloremque laudantium, totam rem aperiam, eaque ipsa quae ab illo inventore veritatis et quasi architecto beatae vitae dicta sunt explicabo. Nemo enim ipsam voluptatem quia voluptas sit aspernatur aut odit aut fugit, sed quia consequuntur magni dolores eos qui ratione voluptatem sequi nesciunt."
```
Rajdeepc, line 102: Any reason we have not used English translation here?

packages/textfield/stories/textarea.stories.ts
```diff
@@ -94,6 +94,18 @@ export const grows = (): TemplateResult => html`
     ></sp-textfield>
 `;
 
+export const growsQuiet = (): TemplateResult => html`
+    <sp-field-label for="story">Enter your life story...</sp-field-label>
```
Rajdeepc, line 98: Do we need ... here?  Something not blocking but curious to know

Rajdeepc (changes requested):

packages/textfield/src/Textfield.ts
```diff
@@ -225,7 +225,7 @@ export class TextfieldBase extends ManageHelpText(
 
     private get renderMultiline(): TemplateResult {
         return html`
-            ${this.grows && !this.quiet && this.rows === -1
+            ${this.grows && this.rows === -1
```
blunteshwar, line 228: Why are you checking "this.rows === -1"?

blunteshwar, line 228: Why are you checking "this.rows === -1"?

blunteshwar (changes requested):

packages/textfield/src/textfield.css
```diff
@@ -68,21 +68,49 @@ textarea {
 }
 
 :host([grows]) #sizer {
+    appearance: none;
+    appearance: textfield;
```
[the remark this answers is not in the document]
Westbrook: I was hard copying this from the Spectrum source, and didn't realize the linter was altering this. I've changed the approach here and we should have a much simpler time keeping these in sync, now.

packages/textfield/src/Textfield.ts
Westbrook: When added, we decided that `rows` was more important in the attribute application workflow than `grows`, the implication being, if you set a specific number of `rows`, it should stay that many.

This reminded me to check if that was true/worked, and it appears not to... I'll do some tweaking of the CSS and add a new story to confirm that going forward.

blunteshwar: Okay!
So basically by "rows===-1" the code is implying that this textarea does not have fixed number of rows.
Right?

packages/textfield/stories/textarea.stories.ts
Westbrook: Namely https://www.lipsum.com/...but always happy for a "humanification" of the docs, if you want to make a pass on things like this.

packages/textfield/src/Textfield.ts
Westbrook: Yep!

blunteshwar (approved):

Rajdeepc (approved):

PR #3474, opened 2023-07-19
fix(number-field): update button label to use number-field-labels as part of the text

### LINKED ISSUE #3273: [Bug]: sp-number-field Increment/Decrement button labels
### Code of conduct

### Impacted component(s)

sp-number-field

### Expected behavior

The Number field increase/decrease buttons should use less technical and more specific text

We should use Increase/Decrease NOT Increment/Decrement 
We should also use the number fields label as part of the text

So For
```
<sp-field-label for="units">Package width</sp-field-label>
<sp-number-field
    id="units"
    style="width: 200px"
    value="4"
    format-options='{
        "style": "unit",
        "unit": "inch",
        "unitDisplay": "long"
    }'
></sp-number-field>
```

The buttons should be "Increase Package width" and "Decrease Package Width"

### Actual behavior

Currently the buttons are labeled Increment/Decrement

### Screenshots

_No response_

1. Go to https://opensource.adobe.com/spectrum-web-components/components/number-field/
2. Right Click on an increase/decrease button
3. Inspect the Accessible Name in the Accessibility tab of Dev Tools

_No response_

_No response_

TarunAdobe:
Updated the Number field increase/decrease buttons to use the number fields label as part of their label text

## Description

Previously the number field increase/decrease buttons were using hardcoded labels as Increment/Decrement but now with this update buttons label text would use the number-field-label text as a part of it to give more context and clarity.

So for
```
<sp-field-label for="units">Package width</sp-field-label>
<sp-number-field
    id="units"
    style="width: 200px"
    value="4"
    format-options='{
        "style": "unit",
        "unit": "inch",
        "unitDisplay": "long"
    }'
></sp-number-field>
````

The buttons would have their labels as "Increase Package width" and "Decrease Package Width".

## Related issue(s)

- fixes #3273 

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

-   [x ] Bug fix (non-breaking change which fixes an issue)

## Checklist

## Best practices

### DIFF
--- a/packages/field-label/src/FieldLabel.ts
+++ b/packages/field-label/src/FieldLabel.ts
@@ -101,9 +101,11 @@ export class FieldLabel extends SizedMixin(SpectrumElement) {
             const applyLabel = this.target.applyFocusElementLabel;
             const focusable = this.target.focusElement || this.target;
             const targetParent = focusable.getRootNode() as HTMLElement;
+
             if (typeof applyLabel !== 'undefined') {
                 applyLabel(this.labelText);
-            } else if (targetParent === (this.getRootNode() as HTMLElement)) {
+            }
+            if (targetParent === (this.getRootNode() as HTMLElement)) {
                 const conditionAttribute = target
                     ? conditionAttributeWithId
                     : conditionAttributeWithoutId;
--- a/packages/field-label/src/FieldLabel.ts
+++ b/packages/field-label/src/FieldLabel.ts
@@ -104,8 +104,7 @@ export class FieldLabel extends SizedMixin(SpectrumElement) {
 
             if (typeof applyLabel !== 'undefined') {
                 applyLabel(this.labelText);
-            }
-            if (targetParent === (this.getRootNode() as HTMLElement)) {
+            } else if (targetParent === (this.getRootNode() as HTMLElement)) {
                 const conditionAttribute = target
                     ? conditionAttributeWithId
                     : conditionAttributeWithoutId;
--- a/packages/field-label/src/FieldLabel.ts
+++ b/packages/field-label/src/FieldLabel.ts
@@ -101,7 +101,6 @@ export class FieldLabel extends SizedMixin(SpectrumElement) {
             const applyLabel = this.target.applyFocusElementLabel;
             const focusable = this.target.focusElement || this.target;
             const targetParent = focusable.getRootNode() as HTMLElement;
-
             if (typeof applyLabel !== 'undefined') {
                 applyLabel(this.labelText);
             } else if (targetParent === (this.getRootNode() as HTMLElement)) {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Review by Rajdeepc — APPROVED** [review]
LGTM!

**Comment by Westbrook** [comment]
Please review the failing tests in this branch.

**Comment by TarunAdobe** [comment]
@Westbrook  Here's some explanation for these changes... I'd love to get your opinion on this. 
So in the FieldLabel we have this function applyTargetLabel which we use to apply target label to other components. Now let's say we talk about number-field component. In that we have the focusable set as the input so the label of input is applied automatically from the applyTargetLabel function. We also wanted to have labels on the buttons but the buttons are not focusable (the input is). Thus is we use a separate helper function that is provided namely applyFocusElementLabel then according to the older conditions we will never apply the label to the focusable element. Thus I modified the if/else such that it would always run the applyLabel function if it is provided by the component and on top of that it will do other house keeping tasks like applying label to the focus element. I wanna understand if this is the correct way to do this? Thnx.
(Refer to latest changes for better understanding)

File: packages/field-label/src/FieldLabel.ts [hunk 1]
```diff
@@ -101,9 +101,11 @@ export class FieldLabel extends SizedMixin(SpectrumElement) {
             const applyLabel = this.target.applyFocusElementLabel;
             const focusable = this.target.focusElement || this.target;
             const targetParent = focusable.getRootNode() as HTMLElement;
+
             if (typeof applyLabel !== 'undefined') {
                 applyLabel(this.labelText);
-            } else if (targetParent === (this.getRootNode() as HTMLElement)) {
+            }
+            if (targetParent === (this.getRootNode() as HTMLElement)) {
```
Westbrook, line 108: While I admire the simplicity of this change, allowing multiple paths to be taken in this logic can lead to unexpected outcomes.

In specific, while it is not explicitly negative, you can see when comparing this branch to `main` that when this is leveraged [in conjunction with `<sp-picker>`](https://issue-number-field-labels--spectrum-web-components.netlify.app/storybook/?path=/story/picker--default) this change applies an `aria-label` to the `<button>` within the Picker's shadow root, that is not currently being applied in `main`. 

I _think_ the path would be to either imperatively apply `aria-label` to the `<input>` in the Number Field as part of its `applyFocusElementLabel` method, or to hand state in the root Textfield component that could be adjusted to accept this value into its `render` method.

TarunAdobe:
updated the logic such that textField takes care of setting up the labels on the input properly

TarunAdobe:
updated the code to add a test and fixed the code styling issues @Westbrook

Westbrook (approved):
This looks great. Thanks for adding the test.

PR #3477, opened 2023-07-19
chore: take latest breaking Spectrum CSS updates

Westbrook:
## Description
Take breaking changes as time of PR creation and updates to Color Area and Checkbox to address prior comments.

There are _newer_ versions at Spectrum CSS latest, but many of those versions require the changes in https://github.com/adobe/spectrum-web-components/pull/3416 which are not yet ready to merge.

## How has this been tested?
-   [ ] _Test case 1_
    1. Go [here](https://c40bbfdd62fa1c608457ddf71f72102e--spectrum-web-components.netlify.app/review/)
    2. Review the VRT changes for expected outcomes

## Types of changes

## Checklist

### DIFF
--- a/packages/action-button/src/spectrum-action-button.css
+++ b/packages/action-button/src/spectrum-action-button.css
@@ -82,6 +82,18 @@ governing permissions and limitations under the License.
     );
     --spectrum-actionbutton-border-radius: var(--spectrum-corner-radius-100);
     --spectrum-actionbutton-border-width: var(--spectrum-border-width-100);
+    --spectrum-actionbutton-content-color-default: var(
+        --spectrum-neutral-content-color-default
+    );
+    --spectrum-actionbutton-content-color-hover: var(
+        --spectrum-neutral-content-color-hover
+    );
+    --spectrum-actionbutton-content-color-down: var(
+        --spectrum-neutral-content-color-down
+    );
+    --spectrum-actionbutton-content-color-focus: var(
+        --spectrum-neutral-content-color-key-focus
+    );
     --spectrum-actionbutton-focus-indicator-gap: var(
         --spectrum-focus-indicator-gap
     );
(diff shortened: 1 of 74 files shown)


packages/action-button/src/spectrum-action-button.css
```diff
@@ -96,6 +108,74 @@ governing permissions and limitations under the License.
             var(--spectrum-actionbutton-focus-indicator-gap)
     );
 }
+:host([selected]) {
+    --mod-actionbutton-background-color-default: var(
```
Westbrook, line 112: @pfulton the component specifically giving a value to the `--mod-` custom properties seems like it would break the contract that these can be customized from the outside by consumers. While this could be customized via `<sp-action-button style="--mod-actionbutton-background-color-default: ...">` the previously available approach of `<div style="--mod-actionbutton-background-color-default: ..."><sp-action-button>` to customize a while batch of Buttons is no longer possible.

packages/color-area/src/spectrum-color-area.css
```diff
@@ -72,15 +72,6 @@ governing permissions and limitations under the License.
     -webkit-user-select: none;
     user-select: none;
 }
-:host([focused]) {
-    z-index: 2;
-}
```
Westbrook, line 77: @pfulton some of the Color components remove this, and others do not. Which is correct?

packages/checkbox/src/spectrum-checkbox.css
```diff
@@ -727,8 +727,8 @@ governing permissions and limitations under the License.
             )
             ease-out;
 }
-#checkmark,
-#partialCheckmark {
+.spectrum-Icon#checkmark,
```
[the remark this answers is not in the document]
Westbrook: I worked around this internal to the `<sp-checkbox>` element, but it feels like the style of change we shouldn't need to manage between our teams.

pfulton:
@mlogsdon18 mind taking a look at this?

packages/action-button/src/spectrum-action-button.css
mlogsdon18: @pfulton correct me if I'm wrong here, but this is still possible, a different mod variable would just be used. This is for when the action button is selected, so you'll see that this variable is able to be overwritten by using `--mod-actionbutton-background-color-default-selected` and if that variable is not set it falls back to the selected value. This is our new approach to allow for some more variant/state specific mods while still keeping out CSS clean. Many of our recent components have implemented this approach

packages/checkbox/src/spectrum-checkbox.css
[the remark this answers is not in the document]
mlogsdon18: This was added in order to increase specificity, as depending on how the CSS was compiled, these styles were being overwritten by default `.spectrum-Icon` styling. I agree that these should be updated to be specific to the Checkbox (inside `.spectrum-Checkbox`) but unfortunately that `.spectrum-Icon` identifier is going to need to still be there

packages/color-area/src/spectrum-color-area.css
mlogsdon18: I believe @jenndiaz would know the answer to this better than I do!

jenndiaz: Are we referencing the z-index or the color handle focused styling? 

The color handle focused styling was all moved to be within the ColorHandle. 
For the z-index It looks like this one was removed by mistake, I've put up a PR to put this instance of it back.

Westbrook: Specifically this `z-index`. If you're adding it back, we'll get that when it comes, just wanted to make sure whether this approach or the others was the right one.

packages/checkbox/src/spectrum-checkbox.css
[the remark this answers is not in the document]
mlogsdon18: Yes I got this figured out and there's a PR up in Spectrum CSS so we should be able to get that added and updated for this

packages/action-button/src/spectrum-action-button.css
castastrophe: @mlogsdon18 Well said. Another way to put this is that the default background color mod should now be leveraging the default-selected variant of that same mod. It's a way to cut down on unnecessary selectors when state changes.

castastrophe: One last thought on this to help clarify the approach, the contract is maintained for sp-action-button to set --mod-actionbutton-background-color-default. The only time that variable will not work is when the selected state is active, in which case we expect to use --mod-actionbutton-background-color-default-selected instead anyway. This is how it works today except that css property is overridden and that property is being remapped to a different mod. This instead leverages 1 css property definition and just updates which variable it calls based on state. Let us know if there are any questions or clarifications we can make beyond this!

Westbrook: Thanks for the clarification!

packages/checkbox/src/spectrum-checkbox.css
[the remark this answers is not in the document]
Westbrook: Awesome! When that goes out, can you let us know here? Not needing the in component change will make this PR much cleaner, and hopefully ready to ship 🤞🏼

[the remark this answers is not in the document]
mlogsdon18: @pfulton I know that change got merged. Has that gone out in a release yet?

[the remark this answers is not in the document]
mlogsdon18: @Westbrook Patrick just got the change merged and released

Westbrook:
:shipit:

Issue #3146, opened 2023-04-25
[Bug]: textContent should be used as aria-label when no label attribute is provided in meter and progressbar components


dineshy87:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

ProgressBar, Meter

### Expected behavior

To improve accessibility, the component should be updated to set the aria-label attribute based on the following conditions:

1. If only label is provided, use the label attribute value as both the aria-label and the visible label (For ProgressBar and Meter).
2. If only textContent is provided, use the textContent as both the aria-label and the visible label for progress bar and Meter. For ProgressCircle use it to update just aria-label since there is no visible label.
3. If both label and textContent are provided, use the label attribute value as the aria-label, and display textContent as the visible label (in the sp-field-label element).

### Actual behavior

Currently, the <sp-meter> component sets the aria-label attribute only if the label attribute is provided. When only textContent is provided, the component doesn't set the aria-label, leading to an accessibility issue flagged by Axe DevTools.

For example, try: 

`<sp-meter size="m" progress="71">Tasks Completed</sp-meter>`

The same issue can be observed in examples provided in docs for Meter component: https://opensource.adobe.com/spectrum-web-components/components/meter/

I also created similar combinations with ProgressBar and ProgressCircle here: The above issue can be observed at https://studio.webcomponents.dev/edit/H0daZwrqfYXamkVzquST/src/index.stories.js?p=stories 

Please run Axe DevTools, and observe the errors.

### How can we reproduce this issue?

3. Scroll to '....'
4. Check console
5. See error

Rajdeepc:
@TarunAdobe Please add [this](https://chrome.google.com/webstore/detail/axe-devtools-web-accessib/lhdoppojpmngadmnindnejefpokejbdd) chrome extension to verify the above scenarios. FYI

TarunAdobe:
Yus Yus I have done it already. Thnx

TarunAdobe:
PR -> #3483

PR #3483, opened 2023-07-21
Issue(3146)-Use innerText when label is not provided (meter, progress-bar, progress-circle)

TarunAdobe:
Use innerText when label is not provided (meter, progress-bar, progress-circle)
## Description

To improve accessibility, the component should be updated to set the aria-label attribute based on the following conditions:

- If only label is provided, use the label attribute value as both the aria-label and the visible label (For ProgressBar and Meter).
- If only textContent is provided, use the textContent as both the aria-label and the visible label for progress bar and Meter. For ProgressCircle use it to update just aria-label since there is no visible label.
- If both label and textContent are provided, use the label attribute value as the aria-label, and display textContent as the visible label (in the sp-field-label element).

## Related issue(s)

- #3146 

## Motivation and context

Currently, the component sets the aria-label attribute only if the label attribute is provided. When only textContent is provided, the component doesn't set the aria-label, leading to an accessibility issue flagged by Axe DevTools.

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/meter/src/Meter.ts
+++ b/packages/meter/src/Meter.ts
@@ -18,7 +18,10 @@ import {
     SpectrumElement,
     TemplateResult,
 } from '@spectrum-web-components/base';
-import { property } from '@spectrum-web-components/base/src/decorators.js';
+import {
+    property,
+    query,
+} from '@spectrum-web-components/base/src/decorators.js';
 
 import { ObserveSlotText } from '@spectrum-web-components/shared/src/observe-slot-text.js';
 import { LanguageResolutionController } from '@spectrum-web-components/reactive-controllers/src/LanguageResolution.js';
@@ -53,6 +56,9 @@ export class Meter extends SizedMixin(ObserveSlotText(SpectrumElement, '')) {
     @property({ type: String, reflect: true })
     public label = '';
 
+    @query('slot')
+    private slotEl!: HTMLSlotElement;
+
     private languageResolver = new LanguageResolutionController(this);
 
     @property({ type: Boolean, reflect: true, attribute: 'side-label' })
@@ -66,7 +72,7 @@ export class Meter extends SizedMixin(ObserveSlotText(SpectrumElement, '')) {
         return html`
             <sp-field-label size=${this.size} class="label">
                 ${this.slotHasContent ? html`` : this.label}
-                <slot>${this.label}</slot>
+                <slot @slotchange=${this.onSlotChange}>${this.label}</slot>
             </sp-field-label>
             <sp-field-label size=${this.size} class="percentage">
                 ${new Intl.NumberFormat(this.languageResolver.language, {
@@ -81,6 +87,12 @@ export class Meter extends SizedMixin(ObserveSlotText(SpectrumElement, '')) {
                 ></div>
             </div>
         `;
+    }
+
+    private onSlotChange(): void {
+        if (!this.label && this.slotHasContent) {
+            this.label = this.slotEl.assignedNodes()[0].textContent || '';
+        }
     }
 
     protected override firstUpdated(changes: PropertyValues): void {
--- a/packages/progress-bar/src/ProgressBar.ts
+++ b/packages/progress-bar/src/ProgressBar.ts
@@ -18,8 +18,12 @@ import {
     SpectrumElement,
     TemplateResult,
 } from '@spectrum-web-components/base';
-import { property } from '@spectrum-web-components/base/src/decorators.js';
+import {
+    property,
+    query,
+} from '@spectrum-web-components/base/src/decorators.js';
 
+import { ObserveSlotText } from '@spectrum-web-components/shared/src/observe-slot-text.js';
 import { LanguageResolutionController } from '@spectrum-web-components/reactive-controllers/src/LanguageResolution.js';
 import '@spectrum-web-components/field-label/sp-field-label.js';
 import styles from './progress-bar.css.js';
@@ -27,7 +31,9 @@ import styles from './progress-bar.css.js';
 /**
  * @element sp-progress-bar
  */
-export class ProgressBar extends SizedMixin(SpectrumElement) {
+export class ProgressBar extends SizedMixin(
+    ObserveSlotText(SpectrumElement, '')
+) {
     public static override get styles(): CSSResultArray {
         return [styles];
     }
@@ -35,7 +41,7 @@ export class ProgressBar extends SizedMixin(SpectrumElement) {
     @property({ type: Boolean, reflect: true })
     public indeterminate = false;
 
-    @property({ type: String })
+    @property({ type: String, reflect: true })
     public label = '';
 
     private languageResolver = new LanguageResolutionController(this);
@@ -52,13 +58,17 @@ export class ProgressBar extends SizedMixin(SpectrumElement) {
     @property({ type: String, reflect: true })
     public static: 'white' | undefined;
 
+    @query('slot')
+    private slotEl!: HTMLSlotElement;
+
     protected override render(): TemplateResult {
         return html`
+            <sp-field-label size=${this.size} class="label">
+                ${this.slotHasContent ? html`` : this.label}
+                <slot @slotchange=${this.onSlotChange}>${this.label}</slot>
+            </sp-field-label>
             ${this.label
                 ? html`
-                      <sp-field-label size=${this.size} class="label">
-                          ${this.label}
-                      </sp-field-label>
                       ${this.indeterminate
                           ? html``
                           : html`
@@ -84,6 +94,12 @@ export class ProgressBar extends SizedMixin(SpectrumElement) {
                 ></div>
             </div>
         `;
+    }
+
+    private onSlotChange(): void {
+        if (!this.label && this.slotHasContent) {
+            this.label = this.slotEl.assignedNodes()[0].textContent || '';
+        }
     }
 
     protected override firstUpdated(changes: PropertyValues): void {
--- a/packages/progress-bar/src/ProgressBar.ts
+++ b/packages/progress-bar/src/ProgressBar.ts
@@ -115,6 +115,7 @@ export class ProgressBar extends SizedMixin(
             if (this.indeterminate) {
                 this.removeAttribute('aria-valuemin');
                 this.removeAttribute('aria-valuemax');
+                this.removeAttribute('aria-valuenow');
             } else {
                 this.setAttribute('aria-valuemin', '0');
                 this.setAttribute('aria-valuemax', '100');
@@ -122,8 +123,6 @@ export class ProgressBar extends SizedMixin(
         }
         if (!this.indeterminate && changes.has('progress')) {
             this.setAttribute('aria-valuenow', '' + this.progress);
-        } else if (this.hasAttribute('aria-valuenow')) {
-            this.removeAttribute('aria-valuenow');
         }
         if (this.label && changes.has('label')) {
             this.setAttribute('aria-label', this.label);
--- a/packages/progress-circle/src/ProgressCircle.ts
+++ b/packages/progress-circle/src/ProgressCircle.ts
@@ -18,7 +18,11 @@ import {
     SpectrumElement,
     TemplateResult,
 } from '@spectrum-web-components/base';
-import { property } from '@spectrum-web-components/base/src/decorators.js';
+import {
+    property,
+    query,
+} from '@spectrum-web-components/base/src/decorators.js';
+import { ObserveSlotText } from '@spectrum-web-components/shared/src/observe-slot-text.js';
 import { ifDefined } from '@spectrum-web-components/base/src/directives.js';
 
 import progressCircleStyles from './progress-circle.css.js';
@@ -26,9 +30,12 @@ import progressCircleStyles from './progress-circle.css.js';
 /**
  * @element sp-progress-circle
  */
-export class ProgressCircle extends SizedMixin(SpectrumElement, {
-    validSizes: ['s', 'm', 'l'],
-}) {
+export class ProgressCircle extends SizedMixin(
+    ObserveSlotText(SpectrumElement, ''),
+    {
+        validSizes: ['s', 'm', 'l'],
+    }
+) {
     public static override get styles(): CSSResultArray {
         return [progressCircleStyles];
     }
@@ -47,6 +54,9 @@ export class ProgressCircle extends SizedMixin(SpectrumElement, {
 
     @property({ type: Number })
     public progress = 0;
+
+    @query('slot')
+    private slotEl!: HTMLSlotElement;
 
     private makeRotation(rotation: number): string | undefined {
         return this.indeterminate
@@ -83,6 +93,9 @@ export class ProgressCircle extends SizedMixin(SpectrumElement, {
         ];
         const masks = ['Mask1', 'Mask2'];
         return html`
+            <div style="display:none">
+                <slot @slotchange=${this.onSlotChange}></slot>
+            </div>
             <div class="track"></div>
             <div class="fills">
                 ${masks.map(
@@ -99,6 +112,12 @@ export class ProgressCircle extends SizedMixin(SpectrumElement, {
                 )}
             </div>
         `;
+    }
+
+    private onSlotChange(): void {
+        if (!this.label && this.slotHasContent) {
+            this.label = this.slotEl.assignedNodes()[0].textContent || '';
+        }
     }
 
     protected override firstUpdated(changes: PropertyValues): void {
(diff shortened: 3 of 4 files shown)


packages/progress-bar/src/ProgressBar.ts
```diff
@@ -52,13 +58,17 @@ export class ProgressBar extends SizedMixin(SpectrumElement) {
     @property({ type: String, reflect: true })
     public static: 'white' | undefined;
 
+    @query('slot')
+    private slotEl!: HTMLSlotElement;
+
     protected override render(): TemplateResult {
         return html`
+            <sp-field-label size=${this.size} class="label">
+                ${this.slotHasContent ? html`` : this.label}
```
Rajdeepc, line 67: why an html construct is needed?

packages/progress-bar/src/ProgressBar.ts
```diff
@@ -86,6 +96,12 @@ export class ProgressBar extends SizedMixin(SpectrumElement) {
         `;
     }
 
+    private onSlotChange(): void {
+        if (!this.label && this.slotHasContent) {
+            this.label = this.slotEl.assignedNodes()[0].textContent || '';
```
Rajdeepc, line 101: can you keep a null check on textContent?

packages/progress-bar/src/ProgressBar.ts
```diff
@@ -99,15 +115,14 @@ export class ProgressBar extends SizedMixin(SpectrumElement) {
             if (this.indeterminate) {
                 this.removeAttribute('aria-valuemin');
                 this.removeAttribute('aria-valuemax');
+                this.removeAttribute('aria-valuenow');
```
Rajdeepc, line 118: where are  you adding this attribute?

packages/progress-circle/src/ProgressCircle.ts
```diff
@@ -18,17 +18,24 @@ import {
...
 /**
  * @element sp-progress-circle
  */
-export class ProgressCircle extends SizedMixin(SpectrumElement, {
-    validSizes: ['s', 'm', 'l'],
-}) {
+export class ProgressCircle extends SizedMixin(
+    ObserveSlotText(SpectrumElement, ''),
+    {
+        validSizes: ['s', 'm', 'l'],
```
Rajdeepc, line 36: does this sizes doesnt come from the defaults? You might not need to explicitly specify the sizes here again

packages/progress-circle/src/ProgressCircle.ts
```diff
@@ -101,6 +114,12 @@ export class ProgressCircle extends SizedMixin(SpectrumElement, {
         `;
     }
 
+    private onSlotChange(): void {
```
Rajdeepc, line 117: Think of a way to make this a helper method. This is repeated many number of times.

Rajdeepc (changes requested):

Rajdeepc:
Good Start!! Also please make sure the tests are updated accordingly if needed.

packages/progress-bar/src/ProgressBar.ts
TarunAdobe: right below these lines... when the changes include progress and indeterminate is false

packages/progress-circle/src/ProgressCircle.ts
```diff
@@ -83,6 +93,9 @@ export class ProgressCircle extends SizedMixin(SpectrumElement, {
         ];
         const masks = ['Mask1', 'Mask2'];
         return html`
+            <div style="display:none">
+                <slot @slotchange=${this.onSlotChange}></slot>
+            </div>
```
Westbrook, line 98: Interesting approach. You can `display: none` the `slot` rather than providing extra wrapping DOM. However, all styling should be applied via the associated CSS files when possible.

packages/progress-circle/src/ProgressCircle.ts
```diff
@@ -101,6 +114,12 @@ export class ProgressCircle extends SizedMixin(SpectrumElement, {
         `;
     }
 
+    private onSlotChange(): void {
+        if (!this.label && this.slotHasContent) {
+            this.label = this.slotEl.assignedNodes()[0].textContent || '';
+        }
```
Westbrook, line 120: This pattern can effect a consumers ability to leverage the `label` attribute/property themselves, maybe we don't use it as an interchange. Something like:
```
if (this.label) return;
const textContent = this.slotEl.assignedNodes()[0].textContent;
if (textContent) {
   // set attribute
} else {
  // remove attribute
}
```
Also, you likely want to do a `reduce` across all the assigned nodes to acquire this value.

packages/progress-circle/src/ProgressCircle.ts
```diff
@@ -101,6 +114,12 @@ export class ProgressCircle extends SizedMixin(SpectrumElement, {
...
+    }
+
```
Westbrook, line 122: Below in line 137, we were previously only setting `aria-label` that's technically incorrect. Can you set or remove the attribute based on the presence of `this.label` instead?

packages/progress-circle/src/ProgressCircle.ts
Westbrook: This is a great question! It may be a good idea to go this PR and then follow up with a learning spike to look at how not just this, but a good amount of functionality shared between these three components could be abstracted for simpler maintenance.

packages/progress-circle/src/ProgressCircle.ts
```diff
@@ -18,17 +18,24 @@ import {
...
 import progressCircleStyles from './progress-circle.css.js';
 
 /**
  * @element sp-progress-circle
  */
-export class ProgressCircle extends SizedMixin(SpectrumElement, {
-    validSizes: ['s', 'm', 'l'],
-}) {
+export class ProgressCircle extends SizedMixin(
+    ObserveSlotText(SpectrumElement, ''),
```
Westbrook, line 34: In that Progress Circle doesn't show the label visibly, it might not need this feature addition. It's [not "free" from a performance standpoint](https://github.com/adobe/spectrum-web-components/pull/3483#user-content-progress-circle), so we should be sure that it is 100% needed.

Progress Bar _may_ need it, as the label is visible there.

TarunAdobe: what could be the other way of allowing the innerText to act as a label in case label is not provided without using a slot?

packages/progress-circle/stories/progress-circle.stories.ts
```diff
@@ -80,7 +80,8 @@ export const staticWhite = ({
                 size="s"
                 ?indeterminate=${indeterminate}
                 label="Loading progess demo"
```
Rajdeepc, line 82: spell check on progess?

tools/shared/src/get-label-from-slot.ts
```diff
@@ -0,0 +1,33 @@
...
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+export const getLabelFromSlot = (
+    label: any,
```
Rajdeepc, line 14: why label is of type any?

tools/shared/src/get-label-from-slot.ts
```diff
@@ -0,0 +1,33 @@
...
+    slotHasContent: any,
```
Rajdeepc, line 15: please dont put type any everywhere . Lets be specific of the type

packages/progress-circle/src/ProgressCircle.ts
```diff
@@ -83,6 +91,19 @@ export class ProgressCircle extends SizedMixin(SpectrumElement, {
...
+                @slotchange=${() => {
+                    const labelFromSlot = getLabelFromSlot(
+                        this.label,
+                        this.slotHasContent,
+                        this.slotEl
+                    );
+                    if (labelFromSlot) {
+                        this.label = labelFromSlot;
+                    }
+                }}
```
Westbrook, line 104: Anonymous methods in the render template will unbind and rebind every time the element is rendered, please move this to a class method.

packages/progress-circle/src/ProgressCircle.ts
```diff
@@ -83,6 +91,19 @@ export class ProgressCircle extends SizedMixin(SpectrumElement, {
...
+                style="display:none"
```
Westbrook, line 105: Please move this to the CSS file for this pattern.

packages/meter/src/Meter.ts
```diff
@@ -65,8 +72,20 @@ export class Meter extends SizedMixin(ObserveSlotText(SpectrumElement, '')) {
     protected override render(): TemplateResult {
         return html`
             <sp-field-label size=${this.size} class="label">
-                ${this.slotHasContent ? html`` : this.label}
-                <slot>${this.label}</slot>
```
Westbrook, line 69: For reference, this code pattern supports the following...

By default it applies `this.label` as the default content of the slot, meaning the following will have a visible label of `This is a meter`.

```html
<sp-meter label="This is a meter"></sp-meter>
```

However, a `<slot>` will be filled _ANY_ nodes that are available as a child of the element, so _NONE_ of these will be labeled similarly, if ONLY a `<slot>` is leveraged:

```html
<sp-meter label="This is a meter"> </sp-meter>
<sp-meter label="This is a meter">
</sp-meter>
<sp-meter label="This is a meter">

</sp-meter>
```

This is why we use:

```html
${this.slotHasContent ? html`` : this.label}
<slot>${this.label}</slot>
```

In the case that there is no actual content in the `<slot>` but there is an empty node in the light DOM that would prevent the default `<slot>` content, this will ensure that the visible label is present for all above cases. When there is actual content in the light DOM:

```html
<sp-meter label="This is a meter">
    This has a special label.
</sp-meter>
```

Neither the default `<slot>` content, or the fallback helper will be available, so the only visible content will be `This has a special label.`

That means Meter and Progress Bar likely need to persist this pattern, while Progress Circle, which does not currently support a visible label, likely does not need it.

packages/progress-bar/src/ProgressBar.ts
```diff
@@ -121,13 +143,12 @@ export class ProgressBar extends SizedMixin(SpectrumElement) {
             ) {
```
Westbrook, line 143: The actual test here likely needs to change in order to support the new path to labeling this element with the `textContent`. Tests for the various contexts that we'd expect this to warn user with should look like https://github.com/adobe/spectrum-web-components/blob/main/packages/progress-circle/test/progress-circle.test.ts#L81-L103 or so.

Rajdeepc:
@TarunAdobe If you see some build failing on one test suite. Please re run the workflow. It should pass

packages/progress-circle/src/ProgressCircle.ts
```diff
@@ -123,17 +143,17 @@ export class ProgressCircle extends SizedMixin(SpectrumElement, {
...
                     this,
-                    '<sp-progress-circle> elements will not be accessible to screen readers without one of the following:',
+                    '<sp-progress-circle> elements will not be accessible to screen readers in the following situations:',
                     'https://opensource.adobe.com/spectrum-web-components/components/progress-circle/#accessibility',
                     {
                         type: 'accessibility',
                         issues: [
-                            'value supplied to the "label" attribute, which will be displayed visually as part of the element, or',
-                            'value supplied to the "aria-label" attribute, which will only be provided to screen readers, or',
+                            'if the value is not supplied to "label" attribute and the "content" is also not set for the component, or',
```
Westbrook, line 156:
```suggestion
                            'value supplied to the "label" attribute, which will be displayed visually as part of the element, or',
                            'text content supplied directly to the <sp-progress-circle> element, or',
                            'value supplied to the "aria-label" attribute, which will only be provided to screen readers, or',
```

packages/progress-circle/src/ProgressCircle.ts
```diff
@@ -18,17 +18,22 @@ import {
...
 
 import progressCircleStyles from './progress-circle.css.js';
 
 /**
  * @element sp-progress-circle
  */
-export class ProgressCircle extends SizedMixin(SpectrumElement, {
-    validSizes: ['s', 'm', 'l'],
-}) {
+export class ProgressCircle extends SizedMixin(
```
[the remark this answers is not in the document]
TarunAdobe: Yus sure, I'll take care of that in a new PR

packages/progress-circle/src/ProgressCircle.ts
```diff
@@ -123,17 +143,17 @@ export class ProgressCircle extends SizedMixin(SpectrumElement, {
...
                 !this.label &&
                 !this.getAttribute('aria-label') &&
-                !this.getAttribute('aria-labelledby')
+                !this.getAttribute('aria-labelledby') &&
+                !this.slotHasContent
             ) {
                 window.__swc.warn(
                     this,
-                    '<sp-progress-circle> elements will not be accessible to screen readers without one of the following:',
+                    '<sp-progress-circle> elements will not be accessible to screen readers in the following situations:',
```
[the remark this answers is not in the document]
TarunAdobe: @Westbrook I have modified the warning text to -> "<sp-progress-circle> elements need one of the following to be accessible:" because we have tests written specifically to check if the warning message has the word "accessible" in it. Is this good enough?

packages/progress-circle/src/ProgressCircle.ts
Westbrook: `ObserveSlotText` will only let you know if there is text there or not in a reactive way. You do much the same in `getLabelFromSlot` right now.

packages/progress-circle/src/ProgressCircle.ts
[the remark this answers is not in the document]
Westbrook: Much better, thanks!

Rajdeepc (dismissed):

Westbrook (commented):
The changes here sags us below test coverage requirements: https://app.circleci.com/pipelines/github/adobe/spectrum-web-components/16799/workflows/e30e81dd-a6e1-47d5-b054-8a5415fcbe3a/jobs/366581?invite=true#step-108-2421 Please add or update some tests to ensure the stability of this change. See https://output.circle-artifacts.com/output/job/fe48a7c5-01cb-4294-a7d0-8f777099416d/artifacts/0/coverage/lcov-report/tools/shared/src/get-label-from-slot.ts.html for some uncovered contexts.

Rajdeepc (dismissed):

packages/meter/src/Meter.ts
```diff
@@ -83,6 +90,17 @@ export class Meter extends SizedMixin(ObserveSlotText(SpectrumElement, '')) {
         `;
     }
 
+    protected onSlotChange(): void {
```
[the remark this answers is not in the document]
TarunAdobe: We are using onSlotChange naming convention in multiple components already.... so this feels more consistent. for instance checkout -> ActiveOverlay.ts, Tabs.ts and TopNav.ts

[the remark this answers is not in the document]
Westbrook: Thanks for digging in on normalization! That's a great trait, and something we should do more of as a project. However, we have not normalized here very well, see [handleSlotchange](https://github.com/search?q=repo%3Aadobe%2Fspectrum-web-components+handleSlotchange&type=code) usage... 😞 

`on*` is a common practice in frameworks like React, and surfacing methods on our custom elements with a similar name can cause issues when working in that context without a wrapping React Component. We see this locally in our Storybook from time to time, and while we've recently started shipping [React Wrappers](https://opensource.adobe.com/spectrum-web-components/using-swc-react/) to reduce the changes this happens for our consumers that leverage React, at some point in the future React will _finally_ ship framework native support for custom events and setting properties on DOM elements and users will be freed to return to direct consumption of SWC in that context, so we want to be prepared for that.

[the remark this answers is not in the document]
TarunAdobe: alright got it. I'll change it then

packages/progress-circle/src/ProgressCircle.ts
```diff
@@ -101,11 +112,23 @@ export class ProgressCircle extends SizedMixin(SpectrumElement, {
...
+            this.label = labelFromSlot;
+        }
+    }
+
     protected override firstUpdated(changes: PropertyValues): void {
         super.firstUpdated(changes);
         if (!this.hasAttribute('role')) {
             this.setAttribute('role', 'progressbar');
         }
+        this.slotHasContent = this.slotEl.assignedNodes().length > 0;
```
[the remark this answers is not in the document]
Westbrook: If the functionality continues as expected without it, yes.

Westbrook (approved):
LGTM!

Thanks for working through all the nuance on this. 👏🏼 👏🏼 👏🏼

PR #3496, opened 2023-07-25
fix(textfield): set "title" when textfield is invalid

### LINKED ISSUE #3099: [Bug]: Text Field type default error text cannot be configured
### Code of conduct

### Impacted component(s)

sp-textfield

### Expected behavior

Textfield accepts a `type` attribute, as well as a `pattern` attribute, that cascades to the native HTML element. Those attributes provide helpers for on-screen keyboards and some native validation functionality. One aspect of the native validation is that it silently creates a mouseover effect when the input doesn't match the native validation. One way to configure that mouseover text to something custom (or translated) is to use the `title` attribute. When the title attribute is set the mouseover text will match the value specified in the `title` attribute.

### Actual behavior

The `title` attribute does not cascade to the native `input` and so any specified text is ignored and the mouseover always matches the default text. I believe the default text matches your OS language settings but even when the browser settings are changed it ignores the setting and shows the text in English.

### Screenshots

Screenshot 2023-04-11 at 10 45 43 AM

#### option 1

1. Go to the storybook for textfield
2. Hover over the invalid input in the default stories area
3. Observe system level `mouseover` reading: "Please match the requested format."

#### Option 2

1. Go to https://opensource.adobe.com/spectrum-web-components/components/textfield/
1. Edit the invalid field and add a `type` attribute. I'd recommend using `url`
1. Enter an invalid url `sdfsadfs` will work.
1. Hover over the alert icon
1. Observe the system level text reading: "Please enter a URL."

```ts
... 
return html`
  <sp-textfield
    type="url"
    title="My translated error"
  >
  </sp-textfield>
`

```

_No response_

TheMcMurder:
## Description

When a textfield is marked as invalid, we should mark the title as an empty string to prevent non-intended text from appearing in the the DOM

## Related issue(s)

- fixes #3099

## Motivation and context

Fixes Adobe Express internal issue

## How has this been tested?

Changes were tested in the storybook environment

-   [ ] _Test case 1_
    1. Go to the default page for TextField in storybook
    2. Hover over the invalid field
    3. Notice that you no longer get the system level text: "Please match the requested format"

### Before

Screenshot 2023-07-25 at 2 09 32 PM

### After

Screenshot 2023-07-25 at 2 09 50 PM

## Types of changes

## Checklist

## Best practices

## Why this approach?

See discussion on [various approaches](https://github.com/adobe/spectrum-web-components/issues/3099#issuecomment-1506859616)

### DIFF
--- a/packages/textfield/test/textfield.test.ts
+++ b/packages/textfield/test/textfield.test.ts
@@ -168,7 +168,9 @@ describe('Textfield', () => {
 
         const boundsDefaultElement = defaultTextarea?.getBoundingClientRect();
         const boundsOneRowElement = oneRowTextarea?.getBoundingClientRect();
-        expect(boundsDefaultElement?.height).to.be.greaterThan(boundsOneRowElement?.height ?? 0);
+        expect(boundsDefaultElement?.height).to.be.greaterThan(
+            boundsOneRowElement?.height ?? 0
+        );
     });
     it('multiline with rows does not resize', async () => {
         const el = await litFixture<Textfield>(
@@ -429,6 +431,41 @@ describe('Textfield', () => {
             : null;
         expect(input).to.not.be.null;
     });
+    it('valid - boundary-type assertions and title', async () => {
+        const el = await litFixture<Textfield>(
+            html`
+                <sp-textfield
+                    placeholder="Enter your number"
+                    pattern="^[\\d]+$"
+                    value="123"
+                ></sp-textfield>
+            `
+        );
+        await elementUpdated(el);
+
+        expect(el).to.not.equal(undefined);
+        const input = el.shadowRoot.querySelector('#valid');
+        expect(input).to.not.be.null;
+        expect(el.focusElement).to.not.have.attribute('title');
+    });
+    it('valid - multiline - boundary-type assertions and title', async () => {
+        const el = await litFixture<Textfield>(
+            html`
+                <sp-textfield
+                    multiline
+                    placeholder="Enter your number"
+                    pattern="^[\\d]+$"
+                    value="123"
+                ></sp-textfield>
+            `
+        );
+        await elementUpdated(el);
+
+        expect(el).to.not.equal(undefined);
+        const input = el.shadowRoot.querySelector('#valid');
+        expect(input).to.not.be.null;
+        expect(el.focusElement).to.not.have.attribute('title');
+    });
     it('valid - unicode', async () => {
         const el = await litFixture<Textfield>(
             html`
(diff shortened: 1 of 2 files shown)


packages/textfield/test/textfield.test.ts
```diff
@@ -429,6 +431,49 @@ describe('Textfield', () => {
...
+                    value="123"
+                ></sp-textfield>
+            `
+        );
+        await elementUpdated(el);
+
+        expect(el).to.not.equal(undefined);
+        const input = el.shadowRoot
+            ? el.shadowRoot.querySelector('#valid')
+            : null;
```
Westbrook, line 449:
```suggestion
        const input = el.shadowRoot.querySelector('#valid');
```

We likely still have this pattern in a number of places, however the `shadowRoot` should be guaranteed by TS and no longer need this pattern.

packages/textfield/test/textfield.test.ts
```diff
@@ -429,6 +431,49 @@ describe('Textfield', () => {
...
+        expect(input).to.not.be.null;
+        expect(el.shadowRoot.querySelector('input')).to.not.have.attribute(
```
Westbrook, line 451:
```suggestion
        expect(el.focusElement).to.not.have.attribute(
```
While we can reach into the shadow root, it's better not to when there are other paths to acquiring an element.

packages/textfield/test/textfield.test.ts
```diff
@@ -602,6 +647,49 @@ describe('Textfield', () => {
...
         expect(input).to.not.be.null;
     });
+    it('invalid - boundary-type assertions and title', async () => {
+        const el = await litFixture<Textfield>(
+            html`
+                <sp-textfield
+                    placeholder="Enter your number"
+                    type="url"
+                    value="invalid-email"
+                    invalid="true"
```
Westbrook, line 657:
```suggestion
                    invalid
```
As a boolean attribute, `invalid` doesn't need to be provided a "value".

Westbrook (commented):
Thanks for updating this to the new pattern! :bow:

Just some nits here in the syntax used for the tests (which are great to see included here!!) and we're ready to go.

TheMcMurder:
Thanks for the review @Westbrook, I've addressed your feedback

Westbrook (commented):
Sorry, I put feedback in one of your tests, but you leverage the same pattern in each of the tests. Can you do the same across all of them.

I'm also going to set up a landing branch for this, just in case this tweaks the VRTs in anyway. No action needed on your part there, just letting you know

Westbrook (approved):
Awesome! Thanks for guiding this through. 🙇🏼 

LGTM.

PR #3501, opened 2023-07-31
fix(alert-dialog): add Alert Dialog package

Rajdeepc:
## Description
New Alert Dialog Component

## Related issue(s)

- https://github.com/adobe/spectrum-css/pull/1862

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices


Rajdeepc:
Opening up for initial review. Though some css issues are WIP

packages/alert-dialog/src/AlertDialog.ts
```diff
@@ -0,0 +1,209 @@
...
+
+        return [buttonVariant, buttonTreatment];
+    }
+
+    protected override renderDialog(): TemplateResult {
+        const [buttonVariant, buttonTreatment] =
+            this.getButtonVariantAndTreatment(this.variant);
+        return html`
+            <div class="alertDialog">
+                <sp-dialog>
```
jenndiaz, line 159: I am still concerned with how we are using <dialog /> with in alert dialog, They should not rely on each other and it is also causing the CSS for dialog to be applied to alert-dialog when it shouldn't. You can see in the screen shot that  .grid  from Dialog is being applied. 

Screenshot 2023-08-14 at 4 21 16 PM

Rajdeepc: Yes @jenndiaz This is because I have extended the DialogBase in AlertDialog. Is there any specific reasons that Alert Dialog CSS attributes for Divider is different from Dialog? Can we consider DialogBase as the foundation for dialog and built extended dialogs like Alert on top of it?

Rajdeepc: So previously I was re-using the dialog component to pull most of the UI into the alert dialog. To answer your question, I have created the UI in a new base alert component now so that it can be extended in future if any UI / CSS changes comes in only for alert dialog.

Rajdeepc:
@Westbrook @najikahalsema Would like to have an initial round of review. Let me know if you need any changes.

packages/alert-dialog/README.md
```diff
@@ -0,0 +1,377 @@
...
+        "
+    >
+        Toggle Dialog
+    </sp-button>
+</overlay-trigger>
+```
+
+## Accessibility
+
+An `<sp-toast>` element is by default rendered with a `role` of `alert`. When rendering the `<sp-toast>` to a page, it should be place in a container with a `role` of `region`.
```
jenndiaz, line 377: should this info about toast be removed from here?

jenndiaz (commented):
The CSS all looks good! 
Small comment on the docs site. 

Also, I am not sure what the intended scroll behavior should be when the dialog is open. On  the spectrum css docs site you can't scroll behind the dialog when it is open but I noticed you can here.

TarunAdobe (commented):
On the documentation site... Noone of the alert dialog component closes on clicking the continue button. For cases where there is a cancel button in the alert dialog this is still fine but in other variants like `error` there is absolutely no way to close the dialog once its open without reloading the site!

Rajdeepc:
> On the documentation site... Noone of the alert dialog component closes on clicking the continue button. For cases where there is a cancel button in the alert dialog this is still fine but in other variants like `error` there is absolutely no way to close the dialog once its open without reloading the site!

Yes there is a bug for this. Thanks for noticing this. I will pull back this PR for now since we need to discuss on the approach to this alert dialog in a more simpler way leveraging DialogBase as the base which includes the modals styles and attributes. This will make the alert dialog much more simpler.

Rajdeepc:
@Westbrook Would love to have your thoughts in this PR. I have abstracted the alert dialog logic to only surface the contents of the alert dialog.

packages/alert-dialog/README.md
```diff
@@ -0,0 +1,308 @@
+## Description
+
+`sp-alert-dialog` supplies an attribute based interface for the managed custmization of an `sp-alert-dialog` element and the light DOM supplied to it.
```
Westbrook, line 3: Is this still true?

packages/alert-dialog/README.md
```diff
@@ -0,0 +1,308 @@
...
+    </sp-button>
+</sp-alert-dialog>
+```
+
+### Warning
+
+Warning alert dialogs communicate important information to users in relation to an issue that needs to be acknowledged, but does not block the user from moving forward.
+
+```html
+<sp-alert-dialog variant="warning" headline="Unverified format">
```
Westbrook, line 92: Seems like `headline` is no longer part of the API?

packages/alert-dialog/README.md
```diff
@@ -0,0 +1,308 @@
...
+    >
+        Rate now
+    </sp-button>
+</sp-alert-dialog>
+```
+
+### Scroll
+
+```html
+<sp-alert-dialog variant="scroll">
```
Westbrook, line 208: Is `scroll` a variant, or a thing that just happens?

packages/alert-dialog/package.json
```diff
@@ -0,0 +1,77 @@
...
+        "lit-element",
+        "lit-html"
+    ],
+    "dependencies": {
+        "@spectrum-web-components/base": "^0.37.0",
+        "@spectrum-web-components/button": "^0.37.0",
+        "@spectrum-web-components/button-group": "^0.37.0",
+        "@spectrum-web-components/divider": "^0.37.0",
+        "@spectrum-web-components/icons-workflow": "^0.37.0",
+        "@spectrum-web-components/shared": "^0.37.0"
```
Westbrook, line 65: Needs updating after Tuesday's release.

packages/alert-dialog/package.json
```diff
@@ -0,0 +1,77 @@
...
+    },
+    "devDependencies": {
+        "@spectrum-css/alertdialog": "1.0.4"
```
Westbrook, line 68: There is one newer if we want before merging this.

packages/alert-dialog/src/AlertDialog.ts
```diff
@@ -0,0 +1,92 @@
...
+            this._variant = '';
+        }
+        this.requestUpdate('variant', oldValue);
+    }
+
+    public get variant(): AlertDialogVariants {
+        return this._variant;
+    }
+
+    public _variant: AlertDialogVariants = 'confirmation';
```
Westbrook, line 67: This initial value doesn't seem to be sprouted, does it need to be?

packages/alert-dialog/src/alert-dialog.css
```diff
@@ -0,0 +1,17 @@
...
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+@import url('./spectrum-alert-dialog.css');
+
+:host([variant='scroll']) {
+    max-height: 90vh;
+}
```
Westbrook, line 17: Is this really a style of the dialog, or is this something that is applied by the context in which an Alert Dialog is delivered?

packages/alert-dialog/src/AlertDialog.ts
```diff
@@ -0,0 +1,92 @@
...
+    'confirmation',
+    'information',
+    'warning',
+    'error',
+    'destructive',
+    'secondary',
+    'scroll',
+];
+
+export class AlertDialog extends Dialog {
```
Westbrook, line 42: One side effect of extending from Dialog is its attribute/properties are included in the API table: https://feature-new-alertdialog--spectrum-web-components.netlify.app/components/alert-dialog/api/

packages/alert-dialog/stories/alert-dialog.stories.ts
```diff
@@ -0,0 +1,300 @@
...
+            >
+                Continue
+            </sp-button>
+        </sp-alert-dialog>
+    `;
+};
+
+export const warning = (): TemplateResult => {
+    return html`
+        <sp-alert-dialog variant="warning" headline="Unverified format">
```
Westbrook, line 80: Is this still part of the API?

Westbrook (commented):
Some smaller things for clean up.

Only big question is re: the API table that is build by extending Dialog. We may be able to suppress some thing there via:

```
/**
 * @private
 */
public override noDivider = false;
```

But, I'm not sure we've tried that in an extension before....

packages/alert-dialog/README.md
Rajdeepc: yes scroll is a variant: https://opensource.adobe.com/spectrum-css/alertdialog.html#scroll

packages/alert-dialog/src/AlertDialog.ts
Rajdeepc: Nope i dont think this is required after our arch changes

packages/alert-dialog/stories/alert-dialog.stories.ts
Rajdeepc: Nope headline is not an attribute anymore.

packages/alert-dialog/README.md
Westbrook: Help me understand what's different in that "variant". It appears to be an example of a feature, which points to why an override was necessary to deliver it.

packages/alert-dialog/src/AlertDialog.ts
Rajdeepc: True if I create the Alert Dialog from scratch there will be a lot of repetition of code from Dialog. Either we make it DRY or we can abstract it and make it a standalone component detached from Dialog.

Westbrook: I'm less afraid of code duplication than I am of bad docs. Hopefully the `@private` note allows us to get the best of both. 🤞🏼

packages/alert-dialog/src/alert-dialog.css
Rajdeepc: Nope this was coming from spectrum-Modal class as in the image below. Currently we are detached from the modal so to achieve the scroll variant I have added the max-height.
One catch I think is if this alert dialog is added in a modal then we might have two scrolls??

Screenshot 2023-09-07 at 5 49 04 PM

packages/alert-dialog/src/AlertDialog.ts
Rajdeepc: @private or @ignore doesnt work when are extending classes. I think we might need to fallback to use SpectrumElement and define our properties for Alert Dialog

packages/alert-dialog/README.md
Rajdeepc: Updated

packages/alert-dialog/package.json
Rajdeepc: Updated

packages/alert-dialog/package.json
Rajdeepc: Updated

packages/alert-dialog/src/alert-dialog.css
Rajdeepc: Removed this

packages/alert-dialog/src/AlertDialog.ts
Westbrook: Another option here is to invert the class hierarchy. Alert Dialog is the less featured element, copy the things from it that are share to it and then extend it in Dialog?

Rajdeepc: So this way Dialog will extend these Alert Dialog variants. Is this acceptable? This way we are not creating sp-alert-dialog.

Westbrook: Just inverting the class hierarchy, not changing what elements are being defined. Roughly:
```
class AlertDialog extends SpectrumElement {
  // Add functionality shared between AlertDialog and Dialog here...
}

customElements.define('sp-alert-dialog', AlertDialog);

// ... other package

class Dialog extends AlertDialog {
  // Additional and cusotmized Dialog functionality here...
}

customElements.define('sp-dialog', Dialog);
```

Is that more clear?

Rajdeepc: Updated. Let me know how this looks now.!

packages/alert-dialog/test/benchmark/basic-test.ts
```diff
@@ -0,0 +1,18 @@
...
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+import '@spectrum-web-components/alert-dialog/sp-alert-dialog.js';
+import { html } from '@spectrum-web-components/base';
+import { measureFixtureCreation } from '../../../../test/benchmark/helpers.js';
+
+measureFixtureCreation(html`
+    <sp-alert-dialog></sp-alert-dialog>
```
Westbrook, line 17: Please expand this to a more representative test case.

packages/dialog/src/Dialog.ts
```diff
@@ -67,23 +43,18 @@ function gatherAppliedIdsFromSlottedChildren(
...
  * @fires close - Announces that the dialog has been closed.
  */
-export class Dialog extends FocusVisiblePolyfillMixin(
-    ObserveSlotPresence(SpectrumElement, [
-        '[slot="hero"]',
-        '[slot="footer"]',
-        '[slot="button"]',
-    ])
-) {
+export class Dialog extends ObserveSlotPresence(AlertDialog, [
```
Westbrook, line 46: Nice! Make sure we add `@spectrum-web-components/alert-dialog` to the dependencies here.

tools/bundle/elements.ts
```diff
@@ -94,3 +94,4 @@ import '@spectrum-web-components/top-nav/sp-top-nav.js';
 import '@spectrum-web-components/top-nav/sp-top-nav-item.js';
 import '@spectrum-web-components/tray/sp-tray.js';
 import '@spectrum-web-components/underlay/sp-underlay.js';
+import '@spectrum-web-components/alert-dialog/sp-alert-dialog.js';
```
Westbrook, line 97: Please inset this alphabetically.

packages/alert-dialog/src/AlertDialog.ts
```diff
@@ -0,0 +1,230 @@
...
+        );
+    }
+
+    public override disconnectedCallback(): void {
+        window.removeEventListener(
+            'resize',
+            this.shouldManageTabOrderForScrolling
+        );
+        super.disconnectedCallback();
+    }
```
Westbrook, line 229: Could I get you to volunteer for a follow up change that moved this to a Resize Observer so that we can also fix https://github.com/adobe/spectrum-web-components/issues/3613?

Westbrook (commented):
Couple of small polish points and then I think we're ready here.

Great work on the inversion!

packages/dialog/src/Dialog.ts
Rajdeepc: Will add once alert-dialog is published to the npm registry.

packages/alert-dialog/src/AlertDialog.ts
Rajdeepc: Sure!

packages/dialog/src/Dialog.ts
Westbrook: That's the wrong order of operations. All packages within the monorepo are dependable today, so there is no need to wait.. When we make a release with this change, were we not to include this dependency, users of the Dialog package couldn't update for the fact that the released version of Dialog would technically be broken unless they were to manually add a dependency on Alert Dialog locally.

packages/dialog/package.json
```diff
@@ -73,6 +73,7 @@
         "lit-html"
     ],
     "dependencies": {
+        "@spectrum-web-components/alert-dialog": "0.0.1",
```
Westbrook, line 76:
```suggestion
        "@spectrum-web-components/alert-dialog": "^0.0.1",
```
That or it won't get managed correctly at release time.

Issue #3510, opened 2023-08-03
[Bug]: Toast does not open in the correct position when used with overlay-trigger


benjamind:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

toast

### Expected behavior

`sp-toast` used in an overlay trigger should open in the correct position.

### Actual behavior

toast always appears centered on the trigger

### What browsers are you seeing the problem in?

Firefox, Chrome, Safari, Microsoft Edge

### How can we reproduce this issue?

1. Create an `sp-toast` inside an overlay trigger with a specified placement.
2. Trigger the overlay

### Sample code that illustrates the problem

https://studio.webcomponents.dev/edit/lzXQXJSMsytEpRIwIft3/src/index.ts?p=stories

Westbrook:
Closed by https://github.com/adobe/spectrum-web-components/pull/3511

PR #3511, opened 2023-08-03
fix(toast): switches toast[open] to use visibility hidden to fix overlay handling

benjamind:
## Description

Switches `sp-toast` to use the same behavior as `sp-tooltip` for controlling visibility with the `[open]` attribute. This switches from `display: none` to `visibility:hidden` which means the element now has size when hidden allowing the overlay behavior to properly measure it and therefore position it.

## Related issue(s)

#3510 

-

## Motivation and context

Fixes the placement in overlays.

## How has this been tested?

Added storybook to cover this case, and tested locally in Chrome.

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/toast/stories/toast.stories.ts
+++ b/packages/toast/stories/toast.stories.ts
@@ -14,6 +14,10 @@ import { html, TemplateResult } from '@spectrum-web-components/base';
 import '@spectrum-web-components/toast/sp-toast.js';
 import '@spectrum-web-components/button/sp-button.js';
 
+import { Placement } from '@spectrum-web-components/overlay';
+import '@spectrum-web-components/overlay/overlay-trigger.js';
+import { ifDefined } from '@spectrum-web-components/base/src/directives.js';
+
 const toast = ({
     variant = '',
     open = true,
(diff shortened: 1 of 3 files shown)


packages/toast/stories/toast.stories.ts
```diff
@@ -104,3 +108,74 @@ export const Negative = (args: Properties): TemplateResult =>
...
+        ${(
+            [
+                ['bottom', ''],
+                ['left', 'negative'],
+                ['right', 'positive'],
+                ['top', 'info'],
+            ] as [Placement, string][]
+        ).map(([placement, variant]) => {
+            return html`
+                <overlay-trigger
```
Westbrook, line 159: We probably should not insinuate that Toast elements can/should be opened on `hover`. I think if this was set up at `type="click"` and the slot was `slot="click-content"` that we'd be pretty good to go here!

benjamind: haha yes good point.

benjamind: addressed in [592bdbd](https://github.com/adobe/spectrum-web-components/pull/3511/commits/592bdbdf7b59f9b27fade2d47a56e1f63dba3f6f)

Westbrook (commented):
Looks a little _unique_, but otherwise looks pretty good to go!

If you could [cycle the VRTs](https://github.com/adobe/spectrum-web-components#keeping-ci-assets-updated), we'll be able to merge this and have it out to you sometime next week.

Westbrook (approved):
LGTM! Thanks @benjamind :shipit:

PR #3515, opened 2023-08-07
Unresolved dependency on buttonBase by sp-picker-button

inasirhussain:
This issue is happening when only `sp-picker-button` is being used or along with any other control which is not dependent on `sp-button` somewhow.  As <sp-picker-button> is dependent on <sp-button> for buttonBase and sp-button is not getting pulled as transitive dependency, this case arises. We caught this issue while working on SWC support in UXP.


Westbrook (approved):
LGTM! Thanks for catching this. 🙇🏼

Issue #478, opened 2020-02-06
Add Spectrum Combobox Pattern


Westbrook:
Track progress on this component [here](https://github.com/orgs/adobe/projects/48/views/1).

### Expected Behaviour
https://opensource.adobe.com/spectrum-css/combobox.html should be available in the library.

### Actual Behaviour
Missing

### The future
![image](https://user-images.githubusercontent.com/1156657/73966484-1331de00-48e4-11ea-8843-02f166852075.png)

Westbrook:
The hope would be that this could be an extension of the `sp-dropdown` pattern in some way.

CoreyDotCom:
We could definitely use this component, specifically the "quiet" variant and ability to enter value that differs from the stock values (e.g. XD's zoom level combo for example).

Westbrook:
Thanks for the info @CoreyDotCom! Would you be able to share any screenshots of what that looks like? As a "contribution" even the newest version of this pattern has only the smallest amount of direction in the Spectrum documentation, so this information would be invaluable.

jnurthen:
Accessibility of combo boxes is really hard to get right and hard to retrofit if not considered upfront. Please ping me when this component starts to move along.

CoreyDotCom:
Basically just the quiet combo as found here https://opensource.adobe.com/spectrum-css/components/combobox/ 

![image](https://user-images.githubusercontent.com/701828/83896480-3fd6f480-a722-11ea-9bd7-32069e016841.png)

Westbrook:
Revisit this comment https://github.com/adobe/spectrum-web-components/pull/1661#discussion_r679336866 while shipping this pattern.

hunterloftis:
Review pre-1.0; ship only if meeting quality bar.

SavoInfostream:
Any ETA for this component?

Westbrook:
Out current target is end of Q2.

Westbrook:
FYI: Currently, the work on this in blocked by the work going into the Overlay v2 release. Any team that was interested in leveraging a Combobox from SWC soon could smooth the process by beginning to test the [RC release](https://github.com/adobe/spectrum-web-components/discussions/2764#discussioncomment-6450198) of that code in their application.

After that, it would be great if we could collect your specific use cases in [the issue tracking this addition.](https://github.com/adobe/spectrum-web-components/issues/478) Specifically, it would be useful to know which of the various Combobox type you are looking to leverage to support scoping an initial release:
[...]

PR #3552, opened 2023-08-17
feat(textfield): improved functionality in combobox

anshuma:
Updated combobox to include the functionality of both number-field and textfield
Issue linked : https://github.com/adobe/spectrum-web-components/issues/478


Westbrook:
Please submit an issue with feature requests before taking on new work so we can agree on a productive way forward before you begin to address actual work. I'm not sure that this is an acceptable path toward the goals you may have here.

Westbrook:
Whether this feature addition is an expected part of the functionality of the library or not is certainly up for healthy debate, but I doubt that this change will land in this format. Closing for now.

Issue #3258, opened 2023-05-31
[Feat][Menu]: Add support for menu item description


najikahalsema:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-menu, sp-menu-item

### Description of the requested feature

Menu items should support a description to match what currently exists in core tokens work. 

### Mockups or screenshots

![image](https://github.com/adobe/spectrum-web-components/assets/44980010/4edaf680-ce0f-411d-9113-ddc9a430141a)

### Implementation notes or ideas

Add a div with `id="description"` that mimics the behaviour of the existing `label` attribute (which uses an id to style it and has a slot within in order for it to be accessed by the user). If this isn't the right approach, a regular slot element will do.

TarunAdobe:
created a pr for this - #3559

PR #3559, opened 2023-08-21
fix(menu): added support for menu item description

TarunAdobe:
Added support for menu-item description

## Description

Added a description slot in the menu-item that can be used to show description of the corresponding menu-item. With this update the menu-item would match to the existing core tokens work.

## Related issue(s)

-[3258](https://github.com/adobe/spectrum-web-components/issues/3258)

## Motivation and context

Menu items should support a description to match what currently exists in core tokens work.

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

image
image

## Types of changes

## Checklist

## Best practices


Rajdeepc (dismissed):

najikahalsema (changes requested):
Looks good, but I think we need to add at least a baseline test to account for these new changes. 

- I'd like to see a baseline test that renders the menu item as expected; similar to other tests for menu elements.

I would also recommend adding some menu items with descriptions to the different stories like disabled, selection states, etc, to test the visual changes.

packages/menu/stories/menu.stories.ts
```diff
@@ -107,6 +107,46 @@ export const multipleSelect = (): TemplateResult => {
     `;
 };
 
+export const menuItemHasDescription = (): TemplateResult => {
```
najikahalsema, line 110:
```suggestion
export const menuItemWithDescription = (): TemplateResult => {
```

najikahalsema (changes requested):
Thanks for addressing the changes! One little thing and we're good to go.

Rajdeepc (approved):
LGTM!!

blunteshwar (dismissed):

packages/menu/src/MenuItem.ts
```diff
@@ -315,6 +316,9 @@ export class MenuItem extends LikeAnchor(
             <div id="label">
                 <slot id="slot"></slot>
             </div>
+            <span class="description">
```
Westbrook, line 319: The way this is leveraged in concert with the CSS makes the Menu Items each an extra 1px tall. Seems to come from it having a `1px` top margin...

Westbrook: Looks like _maybe_ this could leverage `[name=description]::slotted(*)` rather than the `.description` element.

packages/menu/test/menu-item.test.ts
```diff
@@ -169,6 +169,24 @@ describe('Menu item', () => {
...
+                    </span>
+                </sp-menu-item>
+            `
+        );
+
+        const descriptionSlot = el.shadowRoot.querySelector(
+            'slot[name="description"]'
+        );
+
+        expect(descriptionSlot).to.not.be.null;
```
Westbrook, line 188: Testing wise, this tests that the code you wrote above went unchanged when rendered to the page. In the case that you are testing a renderer, great, in the case that you're testing a custom element, maybe not the most productive. The assertion to guarantee the contract here is more likely:
```suggestion
        expect(descriptionElement.assignedSlot).to.not.be.null;
```
I might change the title of the test to `'assigns content to the "description" slot'`, and change the name of the variable used to `descriptionElement` as the "description slot" is the `<slot name="description">` element within the shadow DOM of the Menu Item.

Westbrook (commented):
Those selectors and the visual output look much better. One other tweak to the test and I'm good to go, but I'll confirm with @najikahalsema once she's online before we merge.

packages/menu/test/menu-item.test.ts
```diff
@@ -169,6 +169,23 @@ describe('Menu item', () => {
...
+                        Description for the Menu-Item
+                    </span>
+                </sp-menu-item>
+            `
+        );
+
+        const descriptionElement = el.shadowRoot.querySelector(
+            'slot[name="description"]'
+        );
+        expect(descriptionElement).to.not.be.null;
```
najikahalsema, line 187: Agree with Westbrook that this should be updated to `descriptionElement.assignedSlot` as he noted.

najikahalsema (changes requested):
Please review the change requested and it should be good to go.

Westbrook (approved):
LGTM! Thanks for working through the testing context here, lots of nuance there but having unit test support for this going forward is gonna be super helpful.

Westbrook:
Oof, there's a conflict that I missed. Clean that up and we're good to go here!

Westbrook (approved):
Noice! Let's ship it. Great new feature for this element. 🥳

PR #3573, opened 2023-08-23
feat(action-menu): added static attribute support

blunteshwar:
## Description
Added `static` attribute support for action-menu 

## Related issue(s)

-

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices


packages/action-menu/src/ActionMenu.ts
```diff
@@ -43,6 +43,9 @@ export class ActionMenu extends ObserveSlotText(PickerBase, 'label') {
     @property({ type: String })
     public override selects: undefined | 'single' = undefined;
 
+    @property({ type: String, reflect: true })
+    public static: 'white' | 'black' | undefined = undefined;
```
Rajdeepc, line 47: default fallback should be 'none' instead of undefined. We should not show user undefined in the select dropdown in storybook

packages/action-menu/src/ActionMenu.ts
```diff
@@ -66,6 +69,7 @@ export class ActionMenu extends ObserveSlotText(PickerBase, 'label') {
             <sp-action-button
                 ?quiet=${this.quiet}
                 ?selected=${this.open}
+                static=${ifDefined(this.static || undefined)}
```
Rajdeepc, line 72: you can remove undefined from here

packages/action-menu/stories/action-menu.stories.ts
```diff
@@ -112,6 +114,19 @@ export default {
                 type: 'boolean',
             },
         },
+        staticValue: {
+            name: 'static',
+            type: { name: 'string', required: false },
+            description: 'The visual static variant to apply to the button.',
+            table: {
+                type: { summary: 'string' },
+                defaultValue: { summary: undefined },
```
Rajdeepc, line 123: defaultValue should be 'none'

packages/action-menu/stories/action-menu.stories.ts
```diff
@@ -112,6 +114,19 @@ export default {
...
+            },
+            control: {
+                type: 'select',
+                options: ['white', 'black', undefined],
```
Rajdeepc, line 127: add 'none' instead of undefined

packages/action-menu/stories/action-menu.stories.ts
```diff
@@ -120,6 +135,7 @@ export default {
         quiet: false,
         tooltipDescription: '',
         tooltipPlacement: 'bottom',
+        static: undefined,
```
Rajdeepc, line 138: static: 'none'

Rajdeepc (changes requested):

packages/action-menu/src/ActionMenu.ts
blunteshwar: done!

packages/action-menu/stories/action-menu.stories.ts
blunteshwar: done!

packages/action-menu/src/ActionMenu.ts
blunteshwar: done!

packages/action-menu/src/ActionMenu.ts
blunteshwar: done

packages/action-menu/stories/action-menu.stories.ts
blunteshwar: done!

packages/action-menu/stories/action-menu.stories.ts
blunteshwar: done!

Rajdeepc (approved):
LGTM!!

packages/action-menu/src/ActionMenu.ts
```diff
@@ -43,6 +43,9 @@ export class ActionMenu extends ObserveSlotText(PickerBase, 'label') {
     @property({ type: String })
     public override selects: undefined | 'single' = undefined;
 
+    @property({ type: String, reflect: true })
+    public static: 'white' | 'black' | 'none' = 'none';
```
Westbrook, line 47: This will reflect `static="none"` at runtime, which has negative performance implications. We should use `undefined` as the default. 

@Rajdeepc makes a good point about Storybook, but we can manage the experience delivered in that context without needing to surface `'none'` as an acceptable value for this property.

packages/action-menu/src/ActionMenu.ts
```diff
@@ -66,6 +69,7 @@ export class ActionMenu extends ObserveSlotText(PickerBase, 'label') {
             <sp-action-button
                 ?quiet=${this.quiet}
                 ?selected=${this.open}
+                static=${ifDefined(this.static)}
```
Westbrook, line 72: Nice use of `ifDefined()` 👏🏼 👏🏼 👏🏼

packages/action-menu/stories/action-menu.stories.ts
```diff
@@ -112,6 +114,19 @@ export default {
...
+            name: 'static',
+            type: { name: 'string', required: false },
+            description: 'The visual static variant to apply to the button.',
+            table: {
+                type: { summary: 'string' },
+                defaultValue: { summary: 'none' },
+            },
+            control: {
+                type: 'select',
+                options: ['white', 'black', 'none'],
```
Westbrook, line 127: I think we cover `undefined` with `"none"` if we use:
```
labels: {
          // 'labels' maps option values to string labels
          white: 'white',
          black: 'black',
          none: undefined,
        },
```
But, I've just read it straight out of the [docs](https://storybook.js.org/docs/react/essentials/controls#dealing-with-complex-values) and not used it before, so double check.

Westbrook (changes requested):

packages/action-menu/src/ActionMenu.ts
blunteshwar: OK!

packages/action-menu/src/ActionMenu.ts
blunteshwar: Thanks!

packages/action-menu/stories/action-menu.stories.ts
blunteshwar: ok

packages/action-menu/stories/action-menu.stories.ts
```diff
@@ -112,6 +114,24 @@ export default {
...
+                defaultValue: { summary: 'none' },
+            },
+            control: {
+                type: 'select',
+                labels: {
+                    white: 'white',
+                    black: 'black',
+                    none: undefined,
+                },
+                options: ['white', 'black', 'none'],
```
Westbrook, line 132: I think `option` overrides `labels` in this case. I would expect choosing `none` not to publish into the DOM: 
image

Rajdeepc: Right so the value of the selection will take from options and also if we add undefined in options it will override the labels. So I think there are two options we have:
1) We can add 'none' as a static variant and check for 'none' value in the update method and make it as empty. 
```
if (changedProperties.has('static') && this.static === 'none') {
            this.static ='';
}
```
2) add undefined as a variant like we have for other components in the code and proceed.

Westbrook: I don't think we should be weighing down runtime code with work arounds for Storybook.

I got the way the `labels` field works quite wrong... 😞 

I think we should interpret `"none"` as `undefined` in the story, something like this:
```
        <sp-action-menu
            label=${ariaLabel}
            ?disabled=${disabled}
            ?open=${open}
            ?quiet=${quiet}
            static=${ifDefined(staticValue === 'none' ? undefined : staticValue)}
            size=${size}
            @change="${changeHandler}"
            .selects=${selects ? selects : undefined}
            value=${selected ? 'Select Inverse' : ''}
        >
```

Westbrook (commented):
Sorry, one last thing!

Catch this up to `main` and decorate more selectively and let's land this baby. 🛬

Rajdeepc:
> Sorry, one last thing!
> 
> Catch this up to `main` and decorate more selectively and let's land this baby. 🛬

Changes updated. Let me know if you see any more issues here.

Westbrook (approved):
LGTM. Thanks for working through the Storybook delivery, makes understanding this addition a whole lot easier!

Issue #3358, opened 2023-06-23
[Bug]: Slider: css style issues after update to v0.32


spdev3000:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Slider, SliderHandle

### Expected behavior

1. ~Slider label should align with value (centered) in non-editable version~
<img width="213" alt="Screenshot 2023-06-23 at 11 52 22" src="https://github.com/adobe/spectrum-web-components/assets/29090947/462d2ffc-5e16-4398-97b7-53c2c7363841">

2. ~Slider variant=tick should not have any extra margin-block-end~
<img width="235" alt="Screenshot 2023-06-23 at 11 52 55" src="https://github.com/adobe/spectrum-web-components/assets/29090947/f0353b03-9a74-4253-8d86-a42ffb65f5e0">

3. Slider with multiple slider handles: slider handles should stay on-top of track (for example by using higher z-index to overlay track)
<img width="342" alt="Screenshot 2023-06-23 at 11 53 10" src="https://github.com/adobe/spectrum-web-components/assets/29090947/b9f2d972-6291-4d11-9538-336f117dec30">

### Actual behavior

After last update to v0.32 we experienced the following issues with slider component.
1. Slider label is not centered with value in non-editable version
<img width="214" alt="Screenshot 2023-06-23 at 11 59 09" src="https://github.com/adobe/spectrum-web-components/assets/29090947/4fb76972-df23-4aa6-abe3-1ef884b542f2">

2. Slider variant=tick have a large extra margin-block-end
<img width="233" alt="Screenshot 2023-06-23 at 11 59 27" src="https://github.com/adobe/spectrum-web-components/assets/29090947/f7ed3c9a-03ca-4e2a-ad25-f6018ca78e00">

3. Slider with multiple slider handles: (first) left most slider handle is overlaid by track as it has z-index: 1 instead of 2 at least.
<img width="395" alt="Screenshot 2023-06-23 at 11 59 43" src="https://github.com/adobe/spectrum-web-components/assets/29090947/b16b89a5-28cb-4ed1-b64d-f3fa06f94bcb">
<img width="32" alt="Screenshot 2023-06-23 at 11 59 51" src="https://github.com/adobe/spectrum-web-components/assets/29090947/93202b50-b553-4bb2-a32e-74a5b430cb8a">

### How can we reproduce this issue?

3. Scroll to '....'
4. Check console
5. See error

[...]

PR #3581, opened 2023-08-23
fix(slider): update CSS conversion for more correct visual delivery

Westbrook:
## Description
- update space between label and track
- update spacing around `<output>` field
- update spacing after a tick slider that does not have tick labels
- update storybook to better capture some of these changes

## Related issue(s)
- refs #3358

## How has this been tested?
-   [ ] _Test case 1_
    1. Go [here](https://slider-padding--spectrum-web-components.netlify.app/storybook/?path=/story/slider--default)
    2. See that the labels are closer to the track
-   [ ] _Test case 2_
    1. Go [here](https://slider-padding--spectrum-web-components.netlify.app/storybook/?path=/story/slider--tick) and [here](https://slider-padding--spectrum-web-components.netlify.app/storybook/?path=/story/slider--tick-labels)
    2. See that the vertical rhythm is different (or the same? specifically the second slider should not be in the same place vertically between the two views) based on the presence of labels

## Types of changes

## Checklist


packages/slider/src/spectrum-config.js
```diff
@@ -51,6 +69,24 @@ const config = {
                     ],
                     'variant'
                 ),
+                {
```
Rajdeepc, line 72: Can this change come from spectrum-css? My only goal here is to keep spectrum-config as light as possible

packages/slider/src/spectrum-config.js
```diff
@@ -51,6 +69,24 @@ const config = {
...
+                    ],
+                    replace: [
+                        {
+                            replace: builder.id('label-container'),
+                        },
+                        {
+                            replace: builder.combinator('+'),
+                        },
+                        {
+                            replace: builder.id('track'),
```
Rajdeepc, line 86: Can this change come from spectrum-css? My only goal here is to keep spectrum-config as light as possible

Westbrook: I'll make sure to get this in the conversation with their team:
1. ticks vs tick labels rules is 100% a bug on their end (but it is breaking our users, and causing them to leverage tools like `patch-pachage` to work around, so fast forwarding until they've been able to fix is important)
2. we've discussed the `<output>` delivery before, they called it "correct", so I'm not sure we'll get tracking there
3. the use of the `#track` element is something we added to complete the functional contract of the track. The actual CSS element don't go all the way to the left/right edges of the element, which means they have less fidelity in the pointerdown/move/up events than is needed
4. it's not in this PR but it's always a good idea to remind them that the ticks and the min/max position for the handle do not line up with the current visual delivery.
While they cycle on their end, I think we should move forward with these changes. We can revert/match to their final decisions later, but right now there are low quality visual deliveries hitting visitors today that we should clear up.

Rajdeepc (approved):

Issue #3574, opened 2023-08-23
[Feat]: Introduce publicly-accessible version to SWC components


crav710:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Base component

### Description of the requested feature

This relates to #3556 discussion started by @rickharris. 
Adding static version property for components classes so that component libraries can handle scenarios where two different versions of the same component are loaded, resulting into a conflict if they try to register the same custom element name.

### Mockups or screenshots

The feature would give ability to get version somewhat like below 
`const { version } = customElements.get('sp-action-bar') as SpectrumElement`

### Implementation notes or ideas

The proposed solution uses [genversion](https://github.com/axelpale/genversion) to write the current version to a file before committing a new version.  It writes one version file in `tools/base` and set it on `SpectrumElement` so that each component inherits the version number.

PR #3582, opened 2023-08-24
fix(base): introduce static version property for component class

crav710:
## Description
Adding Static property for component classes which can be useful for component libraries to handle scenarios like version conflict. 

## Related issue(s)

- fixes https://github.com/adobe/spectrum-web-components/issues/3574
- Discussion :: https://github.com/adobe/spectrum-web-components/discussions/3556

## Motivation and context

Adding version to SWC components would help component library to make decision in below scenarios ::

- If the registered component's version is semver compatible with the component libraries' dependency, there's no need to register the library's version of the component
- If the registered component's version is not semver compatible the library might register and use the component using a custom tag name.
- The component library could possibly smooth over backwards-incompatible changes by knowing what the registered version is and work around compatibility issues.

Versioning would also help in long term approach for custom scoped registries as discussed in #3556  

## How has this been tested?

- Unit test for base.test has been written to check the `version` is present in `SpectrumElement`
- Tested via storybook to check if the version is set correctly. 

## Types of changes

## Checklist

## Best practices


.husky/pre-commit
```diff
@@ -2,7 +2,10 @@
...
 
 STAGED_FILES_TO_LINT=$(git diff --name-only --cached --diff-filter=d -- "*.ts" "*.js")
+VERSION_FILE=$(dirname "$0")/../tools/base/src/version.js
 yarn eslint -f pretty $STAGED_FILES_TO_LINT
 yarn analyze
 yarn lint:css
-yarn pretty-quick --staged
\ No newline at end of file
+yarn pretty-quick --staged
+yarn genversion --es6 --semi $VERSION_FILE
```
crav710, line 10: @Westbrook  I have verified that upon running the `lerna publish` command locally, it automatically makes a commit. During this process, the above-mentioned workflow is triggered, which updates the version file to 0.38.0.

tools/base/src/Base.ts
```diff
@@ -11,6 +11,7 @@ governing permissions and limitations under the License.
 */
 
 import { LitElement, ReactiveElement } from 'lit';
+import { version } from './version';
```
Westbrook, line 14: Please use file extensions.

tools/.eslintrc.json
```diff
@@ -105,6 +105,12 @@
                     }
                 ]
             }
+        },
+        {
+            "files": ["./base/src/version.js"],
+            "rules": {
+                "notice/notice": "off"
+            }
```
Westbrook, line 113: Why was this needed?

tools/base/test/base.test.ts
```diff
@@ -10,6 +10,7 @@ governing permissions and limitations under the License.
 */
 import { SpectrumElement } from '@spectrum-web-components/base';
 import { elementUpdated, expect, fixture, html } from '@open-wc/testing';
+import { version } from '../src/version';
```
Westbrook, line 13: File extension.

tools/.eslintrc.json
Westbrook: If the generator can't add this information, should we add this file to `.eslintignore` instead?

tools/base/src/Base.ts
```diff
@@ -183,7 +184,9 @@ export function SpectrumMixin<T extends Constructor<ReactiveElement>>(
     return SpectrumMixinElement;
 }
 
-export class SpectrumElement extends SpectrumMixin(LitElement) {}
+export class SpectrumElement extends SpectrumMixin(LitElement) {
+    static VERSION = version;
```
Westbrook, line 188: Do we want this on any files built from the mixin, as well?

How do you think this location will effect projects that leverage SpectrumElement as a base class but aren't SWC?

tools/.eslintrc.json
Westbrook: Or I guess the question is why isn't listing it there _enough_?

Westbrook (commented):
Some nitty things, but really close. Seems like we'd be in a good place to test this out on the next release after those small points of clean up.

tools/.eslintrc.json
crav710: I think the intension here is to allow the linter to run on both the files (d.ts & .js) as `tools/**/*.d.ts
` excludes `tools/base/src/version.d.ts` file from tools in .eslintignore 
Then to ignore the header error for version.js we had to add the above override rule inside .eslintrc.json.

I believe we can safely remove the last line `!tools/base/src/version.js` as it's included by default. Although if we just add `tools/base/src/version.js` to .eslintignore we would not require the above change in .eslintrc.json. 

@rickharris why is it necessary to add version.d.ts & version.js files to linter ?

tools/base/src/Base.ts
crav710: Please correct me If I am wrong ,upon checking in the repository for [projects](https://git.corp.adobe.com/CCX/cloud-shared-components/tree/develop/packages/projects) I couldn't find an implementation which uses SpectrumElement as a base class.  I think best person to answer this would be @rickharris related to projects query.

tools/.eslintrc.json
rickharris: Yeah I think that we can eslintignore the files if we'd like to. I also think you're right that we can remove `!tools/base/src/version.js` from the eslintignore file. Either option should be fine, just a preference of if we want to ignore the whole file or just the header rule.

tools/base/src/Base.ts
rickharris: Personally I don't have enough info to know where the right place to add the version is. Hopefully the answer to 

> How do you think this location will effect projects that leverage SpectrumElement as a base class but aren't SWC?

is "minorly to not at all", but I really am not confident about that at all. I suppose that in that case it would still represent the version of `SpectrumElement` being used, and an extending component that's not SWC could override it they needed to.

tools/.eslintrc.json
crav710: @Westbrook  added tools/base/src/version.js to .eslintignore & removed the header override.

tools/base/test/base.test.ts
crav710: @Westbrook After adding the file extensions I see that all tests are failing reason being the version.js is being created/updated using genversion & it's excluded from build so version.dev.js is not created & all the test's fail. 
Is it okay to roll back to removing extension from version as only version.js file is being created ?

Westbrook (commented):
CI is failing because by default our tooling expects there to be both a `version.js` file AND a `version.dev.js` file. The laziest answer is to give it what it wants and build both files in the `pre-commit`. Alternatively, we could add specific support for `version.js` in [this script](https://github.com/adobe/spectrum-web-components/blob/main/tasks/hydrate-export-maps.js#L85-L93) so that the expectation of the `version.dev.js` file is not created. Both are a tiny bit awkward, so I'm fine with either.

tools/base/test/base.test.ts
Westbrook: File extensions are required. Checkout my other note re `*.dev.js` files.

crav710:
@Westbrook Added the support in [script](https://github.com/adobe/spectrum-web-components/blob/main/tasks/hydrate-export-maps.js#L85-L93) by removing the development condition similar to .css.

tools/.eslintrc.json
Westbrook: Are we able to remove the updates to this file with that change?

tools/base/src/Base.ts
Westbrook: We can probably work that out over time.

tools/.eslintrc.json
crav710: Yes Removed it.

Westbrook (approved):
Let's try it. If it does work for the team, I'd love for us to go back and document it's usage, and how other teams might override it, etc. but for now we can keep it more on the quiet side. 

:shipit:

Issue #2064, opened 2022-01-20
Investigate Library Wide Constant Usage


Westbrook:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Description of issue

Let's share constants so strings are easier on the eyes.

castastrophe:
There's not enough information to move forward on this with more action items so I'm going to close it out for now. If this needs to be reopened, please let us know!

PR #3586, opened 2023-08-24
feat:(picker): add aria describedby action menu, picker, split button

blunteshwar:
## Description
 Apply aria-describedby content to Action Button, Picker, Split Button]

## Related issue(s)

-

## How has this been tested?
Using accessibility and inspect element

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices


packages/split-button/stories/index.ts
```diff
@@ -156,6 +157,7 @@ export const splitbutton = (
         type=${properties.type || 'field'}
         ?open=${!!properties.open}
         ?disabled=${properties.disabled}
+        description=${properties.description}
```
Rajdeepc, line 160: please use a ifDefined here too

Rajdeepc (changes requested):

packages/split-button/stories/index.ts
blunteshwar: here description is an inherited property, it's inherited from `picker` hence I believe we don't explicitly need `ifDefined`

Rajdeepc (approved):

packages/action-button/stories/index.ts
```diff
@@ -42,6 +43,7 @@ export function renderButton(properties: Properties): TemplateResult {
             ?disabled=${!!properties.disabled}
             ?selected=${!!properties.selected}
             ?toggles=${!!properties.toggles}
+            description="${ifDefined(properties.description)}"
```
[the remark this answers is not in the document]
blunteshwar: I have tried to correct this according to my understanding
Please have a look

Westbrook:
Looks like this branch is pretty far behind `main`. Please catch it up and ensure that all VRTs are passing before we dive into code review.

blunteshwar:
@Westbrook the code coverage on my local is passing but here its not passing.

Westbrook:
It's not "failing", it's telling you that you added more code than you tested and falling below coverage levels: https://app.circleci.com/pipelines/github/adobe/spectrum-web-components/17756/workflows/04b029be-f42c-4014-8ca2-c8a7363c8a86/jobs/404448

packages/picker/src/Picker.ts
```diff
@@ -60,6 +60,18 @@ const chevronClass = {
...
+    static instanceCount = 0;
+    instanceCount: number;
+
+    @property({ type: String })
+    public description = '';
+
+    constructor() {
+        super();
+        this.instanceCount = PickerBase.instanceCount++;
+        this.description = `sp-picker-${this.instanceCount}`;
```
Westbrook, line 72: This description is being completed within the shadow root, we should be able to use a static value here, rather than constructing something with the instance count.

packages/color-area/test/color-area.test.ts
```diff
@@ -547,6 +547,8 @@ describe('ColorArea', () => {
         );
 
         await elementUpdated(el);
+        const Xvalue = Number(Number(el.inputX.value).toFixed(2));
+        const Yvalue = Number(Number(el.inputY.value).toFixed(2));
```
Westbrook, line 551: Can we fast forward unrelated changes into a standalone PR that we can merge first?

packages/split-button/src/SplitButton.ts
```diff
@@ -139,6 +140,7 @@ export class SplitButton extends SizedMixin(PickerBase) {
             `,
             html`
                 <sp-button
+                    aria-describedby=${this.description}
```
Westbrook, line 143: This ones hard to make decisions about, as it's deprecated. Hard to say whether we should be describing both buttons in this way or not.

packages/action-menu/src/ActionMenu.ts
```diff
@@ -87,6 +87,7 @@ export class ActionMenu extends ObserveSlotPresence(
     protected override render(): TemplateResult {
         return html`
             <sp-action-button
+                aria-describedby=${this.description}
```
Westbrook, line 90: You don't seem to be making this same associating in the `Picker.ts` code...

packages/action-button/stories/index.ts
```diff
@@ -42,6 +42,9 @@ export function renderButton(properties: Properties): TemplateResult {
             ?disabled=${!!properties.disabled}
             ?selected=${!!properties.selected}
             ?toggles=${!!properties.toggles}
+
+            description="${properties.description}"
+
```
Westbrook, line 47: Based on current usage, e.g. the component builds its own value for this property, do we want to surface it in the stories?

packages/color-area/test/color-area.test.ts
blunteshwar: These are just some tests for color-area that I added to increase the code coverage. Other than this I didn't make any changes in the color-area component!

Rajdeepc: @blunteshwar I think Westbrook is right! lets not pollute this PR with color-area changes.

packages/action-button/stories/index.ts
blunteshwar: removed from stories
Done!

packages/picker/test/index.ts
```diff
@@ -248,6 +248,11 @@ export function runPickerTests(): void {
         it('loads accessibly', async () => {
             await expect(el).to.be.accessible();
         });
+        it('checks for elements id and  description', async () => {
+            el = await pickerFixture();
+            expect(el).to.have.attribute('id');
+            expect(el.description).to.not.be.null;
```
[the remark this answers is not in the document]
blunteshwar: done!

packages/action-menu/src/ActionMenu.ts
blunteshwar: done!

packages/picker/src/Picker.ts
blunteshwar: done!

packages/split-button/src/SplitButton.ts
blunteshwar: @Rajdeepc can you help me out with this?

packages/picker/src/Picker.ts
```diff
@@ -417,6 +425,13 @@ export class PickerBase extends SizedMixin(Focusable, { noDefaultSize: true }) {
         `;
     }
 
+    public renderDescriptionSlot(): TemplateResult {
+        return html`
+            <div id=${ifDefined(this.description)}>
```
Westbrook, line 430: This value is internal, please make it static. Doing so should allow us to remove the `description` property from this class as well.

packages/picker/src/Picker.ts
```diff
@@ -417,6 +425,13 @@ export class PickerBase extends SizedMixin(Focusable, { noDefaultSize: true }) {
         `;
     }
 
+    public renderDescriptionSlot(): TemplateResult {
```
Westbrook, line 428: Please make this a getter to match with other sub-templates in this file. It should also be `private` (if only used in this file) or `protected` (if leveraged in other class extensions).

packages/action-menu/test/index.ts
```diff
@@ -304,6 +304,15 @@ export const testActionMenu = (mode: 'sync' | 'async'): void => {
             expect(button).to.have.attribute('aria-expanded', 'true');
             expect(button).to.have.attribute('aria-controls', 'menu');
         });
+        it('has attribute aria-describedby', async () => {
```
Westbrook, line 307: Great first step, but what we really need here is a test leveraged [findDescribedNode](https://github.com/adobe/spectrum-web-components/blob/main/test/testing-helpers-a11y.ts#L21) to confirm the expected association is being created here.

packages/picker/src/Picker.ts
Westbrook: Once this value is static, please add documentation for this slot similar to [this](https://github.com/adobe/spectrum-web-components/blob/main/packages/picker/src/Picker.ts#L689).

packages/picker/src/Picker.ts
blunteshwar: Done!

packages/picker/src/Picker.ts
blunteshwar: Done!

packages/action-menu/test/index.ts
blunteshwar: Done!

packages/picker/src/Picker.ts
```diff
@@ -60,6 +60,13 @@ const chevronClass = {
 export class PickerBase extends SizedMixin(Focusable, { noDefaultSize: true }) {
     protected isMobile = new MatchMediaController(this, IS_MOBILE);
 
+    @property({ type: String })
+    public description = 'sp-picker';
+
+    constructor() {
+        super();
+    }
```
Westbrook, line 68: Please no empty methods.

packages/picker/src/Picker.ts
```diff
@@ -417,6 +424,13 @@ export class PickerBase extends SizedMixin(Focusable, { noDefaultSize: true }) {
         `;
     }
 
+    protected get renderDescriptionSlot(): TemplateResult {
+        return html`
+            <div id="sp-picker">
```
Westbrook, line 429: I think we should use `description` or similar here, so that when it is leveraged in Action Menu or Split Button, no one is surprised.

packages/picker/src/Picker.ts
```diff
@@ -60,6 +60,13 @@ const chevronClass = {
 export class PickerBase extends SizedMixin(Focusable, { noDefaultSize: true }) {
     protected isMobile = new MatchMediaController(this, IS_MOBILE);
 
+    @property({ type: String })
+    public description = 'sp-picker';
```
Westbrook, line 64: I'm not sure we need to cary this as a property being we're leveraging a static value in 429, below.

packages/action-menu/src/ActionMenu.ts
```diff
@@ -87,6 +87,7 @@ export class ActionMenu extends ObserveSlotPresence(
     protected override render(): TemplateResult {
         return html`
             <sp-action-button
+                aria-describedby=${ifDefined(this.description)}
```
Westbrook, line 90: This should be statically applied, `ifDefined()` shouldn't be needed.

Westbrook (commented):
Please associate this content in Picker and Split Button, too.

packages/picker/src/Picker.ts
blunteshwar: Done!

packages/picker/src/Picker.ts
blunteshwar: Done!

packages/picker/src/Picker.ts
blunteshwar: I used it as a property because action-menu is leveraging this by inheriting it from picker.
If we declare description merely as a static value then I believe action-menu won't be able to inherit it and leverage it

packages/action-menu/src/ActionMenu.ts
blunteshwar: Done!

packages/picker/src/Picker.ts
Westbrook: Could you expand on why you want this value settable from the outside?

blunteshwar: I don't want this value to be settable from outside.
Shall I do like this

```
protected _description= 'sp-picker';

  static get description(): string {
    return this._description;
  }
```

Westbrook: This works, you can also hand it directly on the class for a little less code.
```
protected description = 'sp-picker';
```

Historically, the library has simply leveraged these values as magic strings, which would be acceptable here. This means we "know" via convention that the "description" of an element will be id'd as `description` rather than marking that out in sharable code.

However, we do have [this issue](https://github.com/adobe/spectrum-web-components/issues/2064) open to revisit that.

As a "constant", it may be of use to create and export one from this file, so it's not fixedly related to the class:
```
export const DESCRIPTION_ID = 'description';
```

Related, I wanted to work out how we hold that value before questioning the value itself, but the element is _sometimes_ called `sp-picker`, I'm not sure the description should be id'd as such...something a little more clear to what it does would be nice.

PR #3589, opened 2023-08-24
chore: enhanced robustness of table component in react runtime

jianliao:
## Description

Implemented TypeScript's optional chaining (question mark operator) for safer access to the **tableHead** property. This ensures that nullish values are gracefully handled, especially since the Lit Element's async update doesn't guarantee its presence in React runtime.

## Related issue(s)
[DTT-1780](https://jira.corp.adobe.com/browse/DTT-1780)

-

## Motivation and context

The change was necessitated by observations where the asynchronous updates in Lit Element did not guarantee the tableHead property's presence in React runtime, leading to potential runtime errors.

## How has this been tested?

Unable to provide a test case because it is race condition issue which was only reproducible in a special env(My local dev instance and sometimes during the VRT testing) 

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

image

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/table/src/Table.ts
+++ b/packages/table/src/Table.ts
@@ -238,7 +238,7 @@ export class Table extends SizedMixin(SpectrumElement, {
                 this.tableHeadCheckboxCell = document.createElement(
                     'sp-table-checkbox-cell'
                 ) as TableCheckboxCell;
-                this.tableHead.insertAdjacentElement(
+                this.tableHead?.insertAdjacentElement(
                     'afterbegin',
                     this.tableHeadCheckboxCell
                 );
@@ -306,7 +306,7 @@ export class Table extends SizedMixin(SpectrumElement, {
 
             this.manageHeadCheckbox(allSelected);
 
-            this.tableHead.insertAdjacentElement(
+            this.tableHead?.insertAdjacentElement(
                 'afterbegin',
                 this.tableHeadCheckboxCell
             );
@@ -320,7 +320,7 @@ export class Table extends SizedMixin(SpectrumElement, {
                 checkbox.checked = row.selected;
             });
         } else {
-            this.tableHead.querySelector('sp-table-checkbox-cell')?.remove();
+            this.tableHead?.querySelector('sp-table-checkbox-cell')?.remove();
             this.tableRows.forEach((row) => {
                 row.checkboxCells[0]?.remove();
                 if (this.selected.length) {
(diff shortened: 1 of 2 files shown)


Westbrook (commented):
I'm not 100% sure how to prove this is a bad idea, but it _feels_ like we'd be preventing functionality in the only pass that the functionality will have a change to apply, which would later leave the element not functioning as expected. Any thoughts on how we could test that the required functionality here actually happens with the addition of the `?.` sigils?

jianliao:
Hi @Westbrook,

Thanks for sharing your concerns. I took a closer look into our existing table component test cases. Interestingly, some([here](https://github.com/adobe/spectrum-web-components/blob/ffe39ee729a9024dbe948319562b44d3316ecb5c/packages/table/test/table-selects.test.ts#L59C2-L59C2) and [here](https://github.com/adobe/spectrum-web-components/blob/ffe39ee729a9024dbe948319562b44d3316ecb5c/packages/table/test/table.test.ts#L118)) of them already validate that the functionalities remain intact even when using the optional chaining (question mark operator). To further ensure the robustness of this change, I've added a new test case specifically to mimic the scenario that this PR addresses.

Please let me know if you'd like any additional information or clarifications on this!

packages/table/src/Table.ts
```diff
@@ -238,7 +238,7 @@ export class Table extends SizedMixin(SpectrumElement, {
                 this.tableHeadCheckboxCell = document.createElement(
                     'sp-table-checkbox-cell'
                 ) as TableCheckboxCell;
-                this.tableHead.insertAdjacentElement(
+                this.tableHead?.insertAdjacentElement(
```
Westbrook, line 241: If this code runs, and it doesn't do the insert because it ran too early in whatever context, then it won't run again and the `tableHeadCheckboxCell` element will never be inserted into the `tableHead`. Does you use case (in the S2 docs) actually use a Table Head? If it works for you there, then I'm just being over worried, but the test you added (while great) doesn't leverage the alternative timing that you would see in React confirm this.

jianliao: Hi @Westbrook,

I understand your concern regarding the potential risk of the tableHeadCheckboxCell element not being inserted into the tableHead due to early execution.

To address your question: Yes, in the S2 docs use case that I'm working on, the table does indeed have a header. I tested this change locally, and it worked as expected. Interestingly, while running S2 docs in local dev mode, I noticed that the willUpdate() function always gets invoked twice. The first invocation happens before the table headers are populated. However, with the inclusion of optional chaining, by the second invocation, everything proceeds smoothly and as intended.

The new test case I added, though not perfect, is the closest approximation I could come up with to simulate this scenario. I believe it's a good representation of the behavior I observed locally.

I hope this offers some clarity. Always appreciate your diligent feedback!

Westbrook (approved):
Let's to do it!

Issue #3262, opened 2023-05-31
[Bug]: Using arrow keys to toggle buttons in an Radio Group does not trigger change event


yflores814:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Radio Group

### Expected behavior

Using keyboard arrows to select radio buttons in a radio group should trigger change event.

### Actual behavior

We see radio buttons selected when using arrow keys but on change event attached to radio group is not triggered.

### What browsers are you seeing the problem in?

Chrome, Safari

### How can we reproduce this issue?

1. create radio group with a change event and radio buttons
2. use arrow keys to alternate between radio buttons
3. see change event is not called

najikahalsema:
Have you tried listening for a change event on `<sp-radio-group>`?

majornista:
Fix by changing this:
https://github.com/adobe/spectrum-web-components/blob/c5f67662ac54d0e37debaf7cbd7c2df60e8b294e/packages/radio/src/RadioGroup.ts#L54-L56
to this:
```ts
        elementEnterAction: (el: Radio) => {
            this._setSelected(el.value);
        },
```

PR #3592, opened 2023-08-25
fix(radio-group): onChange event not triggering on keyboard navigation

Rajdeepc:
## Description
Added onChange trigger when radio buttons are navigated via keyboard in a radio-group

## Related issue(s)

- https://github.com/adobe/spectrum-web-components/issues/3262

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/radio/src/RadioGroup.ts
+++ b/packages/radio/src/RadioGroup.ts
@@ -52,7 +52,7 @@ export class RadioGroup extends FocusVisiblePolyfillMixin(FieldGroup) {
             });
         },
         elementEnterAction: (el: Radio) => {
-            this.selected = el.value;
+            this._setSelected(el.value);
         },
         elements: () => this.buttons,
         isFocusableElement: (el: Radio) => !el.disabled,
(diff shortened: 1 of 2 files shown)


najikahalsema (changes requested):
Could we get a test in Radio Group for this fix? 

I don't know if we've explicitly discussed this, but it's best practise to write a test when making bug fixes so that we know it's _actually_ fixed and can make sure other additions to the code don't cause our tests to fail. The best way to test this is to write the test before making changes (it will fail), make the change, and watch the test succeed!

the test I'd recommend for this is `it("emits change events on arrow key events")`. You can use the `sinon/spy` package to check the callCount of the change event to make sure it emits as expected.

Let me know if you have more questions!

najikahalsema (approved):
Great! Thank you ☺️

PR #3599, opened 2023-08-29
docs: ensure table loads

Westbrook:
## Description
- don't rewrite HTML to the page after the first time (`code-example` isn't recycled, so this should be fine in our context)
- lazily init the Table demos by the availability of `code-example` AND `sp-table`

## How has this been tested?
-   [ ] _Test case 1_
    1. Go [here](https://table-docs--spectrum-web-components.netlify.app/components/table/)
    2. See that the Virtualized table demos load
    3. Switch the color theme
    4. See that the Virtualized table demos are still live

## Types of changes

## Checklist

### DIFF
--- a/packages/table/README.md
+++ b/packages/table/README.md
@@ -246,8 +246,11 @@ For large amounts of data, the `<sp-table>` can be virtualised to easily add tab
         return items;
     }
 
-    const initTable = () => {
+    const initTable = async () => {
         const table = document.querySelector('#table-virtualized-demo');
+
+        await table.updateComplete;
+        await table.closest('code-example')?.updateComplete;
         table.items = initItems(50);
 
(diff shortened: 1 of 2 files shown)


najikahalsema (dismissed):
LGTM! I'm curious about the usage of the `demo` here. What's the difference between that and the regular `html-live` tag? The async function makes sense to me, but I wonder if there are potential other places where we'd want to denote when to use `demo` or not. If there are docs regarding this I could read to understand the difference, that would be good, too.

Westbrook:
I don't actually know what `demo` does there, and can't seem to find any code that would cover it, even though it's in a couple of other package's README.md. Trying a run without it 🤞🏼

Westbrook:
Removed `demo` and it worked as expected. Shipping!

Issue #2269, opened 2022-05-05
[Bug]: Number field unusable in Safari with Japanese Romaji IME


moisadoru:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

number-field

### Expected behavior

The number field component should be usable across all supported browsers withe the Japanese IME enabled.

### Actual behavior

When Japanese Romaji IME is activated, the number field does not correctly interpret the input. For instance, pressing `5` and then `0` should set the value `50`. In Safari, the value that is set in the field is `550`. Furthermore, when pressing `Backspace`, and extra `5` is added, so the value becomes `5505`. Pressing `1` will add whole chunks of `1` followed by an arbitrary number of `0` digits.

### Screenshots

Here is a small video of the above described behavior of the input field with Japanese ROmaji IME enabled:

https://user-images.githubusercontent.com/109328/166923081-9d74c4be-2b1c-496c-9038-29a86c74c376.mp4

### What browsers are you seeing the problem in?

Safari

### How can we reproduce this issue?

1. Go to the [`number-field` docs page](https://opensource.adobe.com/spectrum-web-components/components/number-field/#example)
2. Change the IME to `Japanese - Romaji` (to add it to the list, go to System Preferences -> Keyboard -> Input Sources, click on `+`, find Japanese, add `Japanese - Romaji`)
3. Focus the number field (optionally clear it's contents)
4. Type `5`, `0`, `<Backspace>`, `1`, `1`...
5. Observe the value in the number field

Westbrook:
I'm only able to recreate this with `Japanese - Hiragana` input, can you confirm that you're seeing this in `Japanese - Romaji`? If there are any other settings you've leveraged on that input source, please share that as well.

joekukish:
@Westbrook I though we had fixed this already with https://github.com/adobe/spectrum-web-components/pull/2019

moisadoru:
I'm reproducing this with `Japanese - Romaji` in Safari 15.4 (macOS Monterey 12.3.1).

Westbrook:
We _may_ be close to a solution with https://number-field-composition--spectrum-web-components.netlify.app/storybook/?path=/story/number-field--default.

It requires extensive testing. On the manual side, we'd love the support of any team looking to leverage this in confirming that we're in a good place here. On the unit side, we need to get some more written, which is awkward as IME is really hard to test.

Westbrook:
Fixed in nFixed by #3610

PR #3610, opened 2023-08-29
fix(number-field): correct composition entry of multi-byte numbers

hunterloftis:
## Description

Current implementation: Intercepts IME-based keystrokes and replaces them with their single-byte cousin. 

This change: Accepts all keystrokes until the component is blurred (loses focus), then converts all IME-based characters into their single-byte cousins. 

## Related issue(s)

- https://github.com/adobe/spectrum-web-components/issues/2269

## Motivation and context

Fix Number Field issues with certain Input Method Editor (IME) input types.

## How has this been tested?

- All existing tests pass
- Manual testing of both IME and non-IME cases

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/number-field/src/NumberField.ts
+++ b/packages/number-field/src/NumberField.ts
@@ -324,6 +324,7 @@ export class NumberField extends TextfieldBase {
     }
 
     private handleKeydown(event: KeyboardEvent): void {
+        if (this.isComposing) return;
         switch (event.code) {
             case 'ArrowUp':
                 event.preventDefault();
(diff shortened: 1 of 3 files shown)


packages/number-field/src/NumberField.ts
```diff
@@ -402,7 +403,27 @@ export class NumberField extends TextfieldBase {
...
 
-    protected override handleInput(): void {
+    protected handleCompositionStart(): void {
+        this.isComposing = true;
+    }
+
+    protected handleCompositionEnd(): void {
+        this.isComposing = false;
+        requestAnimationFrame(() => {
+            this.inputElement.dispatchEvent(
```
Westbrook, line 413: I wish I remembered why I did this in the `rAF()`. It's the only part that seems clearly brittle/directly working to not step on the OS's toes, which means easy to break.

kahol:
@hunterloftis @Westbrook @najikahalsema Can you please review the latest tests using this fix in https://jira.corp.adobe.com/browse/CCEX-47632 ? It sounds like the behavior is better (for the most part, there's no more double entry) but there is still incorrect behavior depending on browser, OS, and language.

PR #3614, opened 2023-09-01
feat(action-button): allow change events to bubble and pierce shadowdom

nickschaap:
## Description

Action button change events should bubble and be composed to match behavior with other SWC components which fire change events such as `sp-switch` and `sp-checkbox`

## Related issue(s)

- #3615 

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/action-group/src/ActionGroup.ts
+++ b/packages/action-group/src/ActionGroup.ts
@@ -183,6 +183,11 @@ export class ActionGroup extends SizedMixin(SpectrumElement, {
         });
     }
 
+    private handleActionButtonChange(event: Event): void {
+        event.stopPropagation();
+        event.preventDefault();
+    }
+
     private handleClick(event: Event): void {
         const target = event.target as ActionButton;
         if (typeof target.value === 'undefined') {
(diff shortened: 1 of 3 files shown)


packages/action-group/src/ActionGroup.ts
```diff
@@ -363,6 +363,7 @@ export class ActionGroup extends SizedMixin(SpectrumElement, {
 
     private manageChildren(changes?: PropertyValues): void {
         this.buttons.forEach((button) => {
+            button.toggles = false;
```
Westbrook, line 366: This could cause an issue when a consumer leverages an Action Group without `selects` but Action Buttons with `toggles` in order to manage the selection of elements in the group via some custom algorithm.

I think with the change to Action Button, that we want to move away from listened on `click` and to listening on `change` and then `event.stopPropagation()`ing the child event. Much like we see in [Radio Group](https://github.com/adobe/spectrum-web-components/blob/main/packages/radio/src/RadioGroup.ts#L112-L116).

nickschaap: @Westbrook Does that mean we want to dispatch a change event for all clicks on action buttons?

Westbrook: Indeed, this is a bit more complex than Radio Group.

The context is that _if_ the Action Group should dispatch `change` events (selects === 'single' | 'multiple') we shouldn't prevent child Action Buttons from dispatching `change` events (`button.toggles = false;`). In that context we should listen for those `change` events and `event.stopPropagation()` on them. We likely also want to `event.preventDefault()` them, as the Action Button is a controller component of the Action Group in that case and does not get to determine its own state.

Does that make next steps here clear enough?

Westbrook (approved):
@nickschaap, made a couple of changes and think this is OK to land. Can you confirm whether you think this would cover your original use case before we merge?

nickschaap:
@Westbrook LGTM thanks!

PR #3617, opened 2023-09-05
chore: switch to fast-glob from glob

Westbrook:
## Description
Not sure why this started to fail at `get-ready` time, but normalizing to `fast-glob` which we use most other places seems to do the trick!

## How has this been tested?
-   [ ] _Test case 1_
    1. run `yarn`
    2. don't get failures

## Types of changes

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -151,7 +151,6 @@
         "fs-extra": "^11.1.1",
         "geckodriver": "4.2.0",
         "gh-pages": "^4.0.0",
-        "glob": "^10.3.0",
         "gunzip-maybe": "^1.4.2",
         "husky": "^8.0.3",
         "latest-version": "^7.0.0",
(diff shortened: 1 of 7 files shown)


TarunAdobe:
Looks good! I have a question... we're importing `glob` at two places only right now (cem-plugin-react-wrapper.js and icons-ui/bin/build.js) Shouldn't we update it at both places and completely remove the `glob` import to avoid any conflicts later on...?

Westbrook:
@TarunAdobe great call! I was just chasing errors in the build scripts, but in that it was really just `cem-plugin-react-wrapper.js` that remained, removing unneeded dependencies is the right path forward. Especially after we saw how long it tool someone to delete `node_modules` this morning. Updated and testing in CI, now!

TarunAdobe (approved):

Issue #3631, opened 2023-09-08
[Bug]: Default Tooltip offset should be 4


spdev3000:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Tooltip

### Expected behavior

Default offset should be 4 or 5 (mobile) regarding to Spectrum doc: https://spectrum.adobe.com/page/tooltip/#Offset

### Actual behavior

default offset is "0"

### How can we reproduce this issue?

3. Scroll to '....'
4. Check console
5. See error

Westbrook:
Related: https://github.com/adobe/spectrum-css/issues/2156

PR #3632, opened 2023-09-08
fix(tooltip): update default offset to match Spectrum defs

spdev3000:
This PR fixes the default tooltip offset regarding to latest Spectrum defs

## Description

Tooltip offset is aligned to latest Spectrum defs, see: https://spectrum.adobe.com/page/tooltip/#Offset

## Related issue(s)
[#3631](https://github.com/adobe/spectrum-web-components/issues/3631)

## Motivation and context
This change aligns tooltip offset with current application and sync with latest Spectrum definitions

## How has this been tested?

manual tested

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/tooltip/src/Tooltip.ts
+++ b/packages/tooltip/src/Tooltip.ts
@@ -25,6 +25,10 @@ import type {
     OverlayOpenCloseDetail,
     Placement,
 } from '@spectrum-web-components/overlay';
+import {
+    IS_MOBILE,
+    MatchMediaController,
+} from '@spectrum-web-components/reactive-controllers/src/MatchMedia.js';
 
 import tooltipStyles from './tooltip.css.js';
 import { ifDefined } from 'lit/directives/if-defined.js';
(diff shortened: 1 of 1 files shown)


packages/tooltip/src/Tooltip.ts
```diff
@@ -126,7 +131,7 @@ export class Tooltip extends SpectrumElement {
     public selfManaged = false;
 
     @property({ type: Number })
-    public offset = 0;
+    public offset = this.isMobile.matches ? 5 : 4;
```
Westbrook, line 134: The Spectrum specs never mean "mobile" and "desktop" in this way. They mean it relative to the `scale` the UI is delivered in. I've reached out to the team about tokens for this before we do anything custom here. If we need to do so, we likely want this in https://github.com/adobe/spectrum-web-components/blob/main/tools/styles/scale-medium.css and https://github.com/adobe/spectrum-web-components/blob/main/tools/styles/scale-large.css

spdev3000: OK, I wasn't certain if this was referring to the scale or Mobile in general.
And yes it would make sense to check, if this can be solved by a token or spectrum-var.
Will close this for now.

Issue #3000, opened 2023-03-09
[Bug]: sp-number-field doesn't always limit to max value upon return


celeryclub:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-slider

### Expected behavior

If you type in a number larger than the `max` attribute and press "return", it should change the number to the max.

### Actual behavior

If you type in a number larger than the `max` attribute and press "return", it doesn't always change the number. However, the number is always changed when the input is blurred. In addition, the number is never changed when `format-options` is used, and the number is only changed the first time you press "return" even without `format-options`.

### Screenshots

<img width="302" alt="Screenshot 2023-03-08 at 4 33 48 PM" src="https://user-images.githubusercontent.com/750809/223884511-d75b5203-1184-40e1-be29-060470347d77.png">

### What browsers are you seeing the problem in?

Chrome

### How can we reproduce this issue?

It's on an unreleased app so I can't share a link here.

### Sample code that illustrates the problem

```
<sp-slider
    variant="filled"
    hide-stepper
    editable
    format-options='{"style": "unit", "unit": "percent"}'
    min="0"
    max="100"
    step="1"
>
    <sp-field-label size="l">Opacity</sp-field-label>
</sp-slider>
```

TarunAdobe:
https://github.com/adobe/spectrum-web-components/assets/137032355/c6a68d65-308b-4dfe-9ebb-370d560e4249

TarunAdobe:
@celeryclub  Please checkout the attached video... I wanna know your thoughts on this pattern

PR #3638, opened 2023-09-12
bug(number-field): update number-field value on pressing "enter"

TarunAdobe:
## Description

If you type in a number larger than the max attribute and press "return", it should change the number to the max.
i.e. The value inside the number-field should update on pressing "return" however right now it only happens when the input is blurred out.

## Related issue(s)

- #3000

## Motivation and context

If you type in a number larger than the max attribute and press "return", it doesn't always change the number. However, the number is always changed when the input is blurred. In addition, the number is never changed when format-options is used, and the number is only changed the first time you press "return" even without format-options.

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/number-field/src/NumberField.ts
+++ b/packages/number-field/src/NumberField.ts
@@ -160,10 +160,11 @@ export class NumberField extends TextfieldBase {
         }
         const oldValue = this._value;
         this._value = value;
-        if (!this.managedInput) {
+        if (!this.managedInput && this.lastCommitedValue != this.value) {
             this.dispatchEvent(
                 new Event('change', { bubbles: true, composed: true })
             );
+            this.lastCommitedValue = this.value;
         }
         this.requestUpdate('value', oldValue);
     }
@@ -180,6 +181,7 @@ export class NumberField extends TextfieldBase {
 
     public override _value = NaN;
     private _trackingValue = '';
+    private lastCommitedValue = NaN;
 
     /**
      * Retreive the value of the element parsed to a Number.
@@ -285,9 +287,12 @@ export class NumberField extends TextfieldBase {
         this.buttons.releasePointerCapture(event.pointerId);
         cancelAnimationFrame(this.nextChange);
         clearTimeout(this.safty);
-        this.dispatchEvent(
-            new Event('change', { bubbles: true, composed: true })
-        );
+        if (this.lastCommitedValue != this.value) {
+            this.dispatchEvent(
+                new Event('change', { bubbles: true, composed: true })
+            );
+            this.lastCommitedValue = this.value;
+        }
         this.managedInput = false;
     }
 
@@ -357,6 +362,7 @@ export class NumberField extends TextfieldBase {
                 this.dispatchEvent(
                     new Event('change', { bubbles: true, composed: true })
                 );
+                this.lastCommitedValue = this.value;
             }, CHANGE_DEBOUNCE_MS) as unknown as number;
         }
         this.managedInput = false;
@@ -400,7 +406,10 @@ export class NumberField extends TextfieldBase {
         }
         this.value = value;
         this.inputElement.value = this.formattedValue;
-        super.handleChange();
+        if (this.lastCommitedValue != this.value) {
+            this.lastCommitedValue = this.value;
+            super.handleChange();
+        }
     }
 
     protected handleCompositionStart(): void {
--- a/packages/number-field/src/NumberField.ts
+++ b/packages/number-field/src/NumberField.ts
@@ -160,7 +160,7 @@ export class NumberField extends TextfieldBase {
         }
         const oldValue = this._value;
         this._value = value;
-        if (!this.managedInput && this.lastCommitedValue != this.value) {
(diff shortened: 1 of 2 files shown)


blunteshwar:
@TarunAdobe some VRTs are failing. Can you look at them

Rajdeepc (dismissed):

packages/number-field/src/NumberField.ts
```diff
@@ -340,6 +340,13 @@ export class NumberField extends TextfieldBase {
                     new Event('change', { bubbles: true, composed: true })
                 );
                 break;
+            case 'Enter':
+                event.preventDefault();
```
Westbrook, line 344: What is the default that you're preventing here?

We haven't gotten to Form Association for these Custom Elements, but in the future, were this associated to a form, there would be some expectation that `Enter` would submit the form by default.

packages/number-field/src/NumberField.ts
```diff
@@ -340,6 +340,13 @@ export class NumberField extends TextfieldBase {
...
+                this.inputElement.value = this.formattedValue;
+                this.dispatchEvent(
+                    new Event('change', { bubbles: true, composed: true })
+                );
```
Westbrook, line 348: This looks good are there any tests that would be beneficial to add in support of this going forward?

packages/number-field/src/NumberField.ts
TarunAdobe: good point! I have removed the `event.preventDefault()` and I've updated the few tests with the current behaviour expectations

packages/number-field/src/NumberField.ts
```diff
@@ -160,6 +160,12 @@ export class NumberField extends TextfieldBase {
         }
         const oldValue = this._value;
         this._value = value;
+        if (!this.managedInput && this.lastCommitedValue != this.value) {
```
Westbrook, line 163: `!=` allows coercion, where at `!==` does not. I'm not sure we'd ever be in a situation for coercion to occur here, but if it's possible to use instead, let's normalize on `!==` when possible.

packages/number-field/src/NumberField.ts
```diff
@@ -400,7 +405,11 @@ export class NumberField extends TextfieldBase {
             }
         }
         this.value = value;
```
Westbrook, line 407: This line will trigger the setter. Will the following lines _never_ accidentally dispatch a second `change` event?

packages/number-field/test/number-field.test.ts
```diff
@@ -566,7 +566,7 @@ describe('NumberField', () => {
                 ],
             });
             expect(inputSpy.callCount).to.equal(4);
-            expect(changeSpy.callCount).to.equal(1);
+            expect(changeSpy.callCount).to.equal(0); // the actual value hasn't changed so changespy won't be called
```
Westbrook, line 569: Great point! The name of this test is no longer valid.

We may want both this test and the previous test, just make sure it doesn't end on the same value as it started...

Westbrook: Also...

```suggestion
            expect(changeSpy.callCount, 'value does not change from initial value so no "change" event is dispatched').to.equal(0);
```

The `expect` method accepts "comments" as the second argument.

packages/number-field/test/number-field.test.ts
```diff
@@ -751,7 +751,7 @@ describe('NumberField', () => {
             await sendKeys({ press: 'Enter' });
             await elementUpdated(el);
             expect(lastInputValue, 'last input value').to.equal(10);
-            expect(lastChangeValue, 'last change value').to.equal(10);
+            expect(lastChangeValue, 'last change value').to.equal(0); // value wouldn't go beyond max, so it remains unchanged after pressing enter and thus no changeSpy called
```
Westbrook, line 754: A change and not a change likely benefit two versions of this, too.

packages/number-field/src/NumberField.ts
Westbrook: Also, will this dispatch a `change` event if the use use `el.value = 'new value'` (but a valid value) from the outside?

Westbrook: If this doesn't do that, we should add a test to confirm for the future. If it does do that, we should add a test to confirm that the alternate approach that you take to get there, too.

Westbrook (commented):
Really like what you've settled on here! A couple of extra cases to confirm and tests to duplicate before we can merge this, but it's really starting to shape up.

packages/number-field/src/NumberField.ts
TarunAdobe: no because in the setter we are updated the lastCommittedValue which will cause the check on the line 409 to fail making sure there is not second `change` event

packages/number-field/src/NumberField.ts
TarunAdobe: thnx for noticing this! updated it🥂

PR #3673, opened 2023-09-26
fix(number-field): negative numbers handled

blunteshwar:
## Description

Number field was not correctly handling negative numbers when step was a floating point number

## Related issue(s)

-

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/number-field/src/NumberField.ts
+++ b/packages/number-field/src/NumberField.ts
@@ -464,8 +464,7 @@ export class NumberField extends TextfieldBase {
     }
 
     private validateInput(value: number): number {
-        const signMultiplier = value < 0 ? -1 : 1;  // Basically I am checking if the the number is negative or not, if it is I am multiplying it by -1 to make it positive and the the function validateinput is doing its job and at the end before returning I am again multiplying by -1.
- 'signMultiplier' adjusts 'value' for 'validateInput' and reverts it before returning.
+        const signMultiplier = value < 0 ? -1 : 1; // 'signMultiplier' adjusts 'value' for 'validateInput' and reverts it before returning.
         value *= signMultiplier;
         if (typeof this.min !== 'undefined') {
             value = Math.max(this.min, value);
--- a/packages/number-field/src/NumberField.ts
+++ b/packages/number-field/src/NumberField.ts
@@ -464,7 +464,7 @@ export class NumberField extends TextfieldBase {
     }
 
     private validateInput(value: number): number {
-        const signMultiplier = value < 0 ? -1 : 1;
+        const signMultiplier = value < 0 ? -1 : 1;  // Basically I am checking if the the number is negative or not, if it is I am multiplying it by -1 to make it positive and the the function validateinput is doing its job and at the end before returning I am again multiplying by -1.
         value *= signMultiplier;
         if (typeof this.min !== 'undefined') {
             value = Math.max(this.min, value);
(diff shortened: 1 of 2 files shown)


Westbrook (commented):
This looks pretty good to me. If you could add a test to confirm that the original issue is now address this will be good to go!

packages/number-field/src/NumberField.ts
```diff
@@ -489,6 +491,7 @@ export class NumberField extends TextfieldBase {
                 }
             }
         }
+        value *= signMultiplier;
```
Rajdeepc, line 494: Is this a repetative statement? See [here](https://github.com/adobe/spectrum-web-components/pull/3673/files#diff-bbe60f2ecb5ae23472a9f446b13eaf5000919b731d68bee594c13a930d9b9933R468)

Rajdeepc (changes requested):

packages/number-field/src/NumberField.ts
blunteshwar: yeah!
Basically I am checking if the the number is negative or not, if it is I am multiplying it by -1 to make it positive and the the function validateinput is doing its job and at the end before returning I am again multiplying by -1.

Rajdeepc:
@blunteshwar if you are able to add a test we can merge this

Rajdeepc (commented):
@blunteshwar if you are able to add a test we will be able to merge this

packages/number-field/src/NumberField.ts
Westbrook: It's an easy trap not to catch this, let's drop a comment in that explains this for future us.

blunteshwar: okay!

blunteshwar:
> @blunteshwar if you are able to add a test we will be able to merge this

done!

packages/number-field/src/NumberField.ts
Westbrook: Get this comment in and we're ready to merge!

blunteshwar: done!

packages/number-field/src/NumberField.ts
```diff
@@ -464,6 +464,8 @@ export class NumberField extends TextfieldBase {
     }
 
     private validateInput(value: number): number {
+        const signMultiplier = value < 0 ? -1 : 1;  // Basically I am checking if the the number is negative or not, if it is I am multiplying it by -1 to make it positive and the the function validateinput is doing its job and at the end before returning I am again multiplying by -1.
```
Rajdeepc, line 467:
```suggestion
        const signMultiplier = value < 0 ? -1 : 1;  // Basically I am checking if the the number is negative or not, if it is I am multiplying it by -1 to make it positive and the the function validateinput is doing its job and at the end before returning I am again multiplying by -1.
 'signMultiplier' adjusts 'value' for 'validateInput' and reverts it before returning.
```

packages/number-field/src/NumberField.ts
```diff
@@ -464,6 +464,9 @@ export class NumberField extends TextfieldBase {
...
+ 'signMultiplier' adjusts 'value' for 'validateInput' and reverts it before returning.
+        value *= signMultiplier;
```
Westbrook, line 469: The comment updates got committed weirdly here.

Can you also update to `main` as @TarunAdobe's work was also in this file and there may be conflicts.

Rajdeepc:
@Westbrook If we are good to go, let's land this 🛩️

Westbrook (approved):
:shipit:!

PR #3689, opened 2023-10-02
fix(menu): allow `change` events to be direct

### LINKED ISSUE #3687: [Bug]: Cannot write to clipboard in Safari from change event of sp-action-menu
### Code of conduct

### Impacted component(s)

sp-action-menu

### Expected behavior

I would like to be able to call `navigator.clipboard.writeText` from inside the `change` event of `sp-action-menu`. Safari only allows this function to be called as the result of a user interaction. Since this is the result of a user interaction it seems like this should work, and I would expect to be able to call `navigator.clipboard.writeText` successfully without an error.

It's not completely surprising or unexpected that this wouldn't work, but what's interesting to me is that it does work in react-spectrum's ActionMenu `onAction` callback, so I'm curious if it could be supported here.

### Actual behavior

When calling `navigator.clipboard.writeText` from the change event in Safari it results in an error:

> NotAllowedError
> The request is not allowed by the user agent or the platform in the current context, possibly because the user denied permission.

### Screenshots

_No response_

Safari

1. Go to https://codesandbox.io/s/solitary-worker-6f26cf?file=/src/App.js
2. Click on the action menu trigger under the heading "react-spectrum ActionMenu"
3. Observe that the string "rsCopy" is now in your clipboard
4. Click on the action menu trigger under the heading "SWC sp-action-menu"
5. Observe the `NotAllowedError` and "rsCopy" is still in your clipboard when it should be "spCopy"

https://codesandbox.io/s/solitary-worker-6f26cf?file=/src/App.js

_No response_

Westbrook:
## Description
Remove an asynchronous break in the call stack between `click` and `change` so that `navigator.clipboard.copyText()` can be safely called during the `change` event.

Test this.
- permissions can be added to Chrome
- flags can be set in Firefox
- Safari does not seem to have testing support in the version of Playwright we're leveragng

Demonstrate it in a handful of stories.
- [menu](https://direct-change--spectrum-web-components.netlify.app/storybook/?path=/story/menu--single-select)
- [action menu](https://direct-change--spectrum-web-components.netlify.app/storybook/?path=/story/action-menu--label-only)

## Related issue(s)
- fixes #3687 
  - @rickharris to reconfirm in his local Safari

## How has this been tested?
-   [ ] _Test case 1_
    1. Go [here](https://direct-change--spectrum-web-components.netlify.app/storybook/?path=/story/menu--single-select)
    2. Select a Menu Item
    3. See that it's `value` (or text in most cases) has been copied to the clipboard
-   [ ] _Test case 2_
    1. Go [here](https://direct-change--spectrum-web-components.netlify.app/storybook/?path=/story/action-menu--label-only)
    2. Open the Menu
    3. Select a Menu Item
    4. See that it's `value` (or text in most cases) has been copied to the clipboard

## Types of changes

## Checklist

### DIFF
--- a/packages/action-menu/stories/action-menu.stories.ts
+++ b/packages/action-menu/stories/action-menu.stories.ts
@@ -23,6 +23,7 @@ import { makeOverBackground } from '../../button/stories/index.js';
 import '@spectrum-web-components/icons-workflow/icons/sp-icon-settings.js';
 import type { MenuItem } from '@spectrum-web-components/menu/src/MenuItem.js';
 import { Placement } from '@spectrum-web-components/overlay/src/overlay-types.js';
+import { Menu } from '@spectrum-web-components/menu';
 
 export default {
     component: 'sp-action-menu',
@@ -191,7 +192,10 @@ export const labelOnly = ({
         ?disabled=${disabled}
         ?open=${open}
         size=${size}
-        @change="${changeHandler}"
+        @change=${(event: Event & { target: Menu }): void => {
+            navigator.clipboard.writeText(event.target.value);
+            changeHandler(event);
+        }}
         .selects=${selects ? selects : undefined}
         value=${selected ? 'Select Inverse' : ''}
     >
--- a/packages/menu/src/Menu.ts
+++ b/packages/menu/src/Menu.ts
@@ -528,7 +528,6 @@ export class Menu extends SizedMixin(SpectrumElement, { noDefaultSize: true }) {
             this.selectedItems = [targetItem];
         }
 
-        await this.updateComplete;
         const applyDefault = this.dispatchEvent(
             new Event('change', {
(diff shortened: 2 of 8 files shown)


najikahalsema (commented):
Is this functionality something that is expected or standard to a menu like this, or is this something we'd like to optionally toggle? I can imagine as an end user if I had some unexpected text copied in my clipboard from a menu item I messed with, I would be annoyed.

Westbrook:
Yes, randomly copying stuff into people's clipboard would be weird! However, all of the copy workflows are part of the demos herein, not the custom elements that are distributed themselves. 😅

najikahalsema:
Oh, duh. My brain didn't register this was only in the stories.

najikahalsema (approved):

Issue #2963, opened 2023-02-27
[Bug]: <sp-picker quiet> doesn't lay out properly


adixon-adobe:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-picker

### Expected behavior

The test of `sp-picker` is vertically centered like it was previously.

### Actual behavior

It looks like there's a new [negative margin ](https://github.com/adobe/spectrum-web-components/blob/main/packages/picker/src/spectrum-picker.css#L368) on the button that probably isn't intended to be there with the quiet attribute.

The result of this is that the layout of the internal <button> element is expanding beyond the bounds of the picker, and causing the picker text to align with the top. This is noticeably off if you have a horizontal layout, which it looks the docs are lacking for picker.

It looks like we should be able to work around this by adjusting the variable used here.

### Screenshots

Here's a quick example of how things can look off:

<img width="183" alt="Screen Shot 2023-02-27 at 12 35 25 PM" src="https://user-images.githubusercontent.com/48841507/221652762-62ec1726-df45-41fa-b9c5-323271dcf5ea.png">

<img width="337" alt="Screen Shot 2023-02-27 at 12 36 01 PM" src="https://user-images.githubusercontent.com/48841507/221652873-f0c30c71-9b39-4a4b-b139-b1d62d52d6f0.png">

adixon-adobe:
I'm guessing this is also a spectrum css bug.

blunteshwar:
PR: [https://github.com/adobe/spectrum-css/pull/2169](url)

blunteshwar:
PR: https://github.com/adobe/spectrum-css/pull/2169

jenndiaz:
I do not see this is a bug. The negative margin was included in recent design guidance. 
We do have an [additional class](https://github.com/adobe/spectrum-css/blob/main/components/picker/index.css#L476) for when the label and picker are side by side, @adixon-adobe do you still see this issue when using this class? 
![Screenshot 2023-09-27 at 2 59 01 PM](https://github.com/adobe/spectrum-web-components/assets/63808889/866715ac-7cbd-47c7-a746-01cc9e2c7eed)


CC:  @pfulton @najikahalsema @blunteshwar

blunteshwar:
@jenndiaz Thanks!, after using this additional class the issue is resolved!

blunteshwar:
[...]

PR #3704, opened 2023-10-06
fix(picker): picker-quiet-layout

blunteshwar:
## Description

Added a new property called `sideLabel` to `sp-picker` to introduce a new UI for sideLabel picker variant.

## Related issue(s)

- fixes https://github.com/adobe/spectrum-web-components/issues/2963

## How has this been tested?

-   [ ] _Test case 1_
    1. Go [here](https://blunteshwar-picker-bug--spectrum-web-components.netlify.app/storybook/?path=/story/picker--side-label)
-   [ ] _Test case 2_
    1. Go [here](https://blunteshwar-picker-bug--spectrum-web-components.netlify.app/storybook/?path=/story/picker--quiet-side-label)

Screenshot 2024-02-21 at 5 22 50 PM
Screenshot 2024-02-21 at 5 22 37 PM

## Types of changes

-   [] Bug fix (non-breaking change which fixes an issue)

## Checklist

## Best practices


Rajdeepc:
@blunteshwar If you can add some tests for this change that would be great!

packages/picker/src/Picker.ts
```diff
@@ -433,7 +433,7 @@ export class PickerBase extends SizedMixin(Focusable, { noDefaultSize: true }) {
                 aria-haspopup="true"
                 aria-labelledby="icon label applied-label"
                 id="button"
-                class="button"
+                class=${this.quiet ? 'button sideLabel' : 'button'}
```
Rajdeepc, line 436: You can simplify by writing like this:
```
class=${`button ${this.quiet ? 'sideLabel' : ''}`}
```

Rajdeepc (commented):
Few comments and also please add in few tests

packages/picker/src/Picker.ts
Westbrook, line 436: What about something being `quiet` also makes it have a "side label"?

blunteshwar: I think this is only specific to `picker`

blunteshwar: I talked to @jenndiaz in spectrum-css and she told me that the might rename this `side-label` class to a more generic class name. And yes, we will have to to use this class

Westbrook:
It would be great if there wasn't a requirement for both the label and the picker to notate the label location. In that way, when this is "done" then the story decorator should "just work", rather than look like: 
image

I shared an possible path forward on this with @TarunAdobe on Tuesday, can yall sync up on that?

Rajdeepc:
> It would be great if there wasn't a requirement for both the label and the picker to notate the label location. In that way, when this is "done" then the story decorator should "just work", rather than look like: image
> 
> I shared an possible path forward on this with @TarunAdobe on Tuesday, can yall sync up on that?

I think the above code is a little misleading. The side label of picker is adjusted with `display:flex` on the parent which is not something we should be doing all the time. This way the picker will not know that it is rendering a side label.
Instead, we should allow the `side-aligned` attribute on `sp-field-label` associated with a picker and target the picker to let it self know that it should allow the `sideLabel` attribute on it to remove any depedency of users adding this attribute from their side. Your thoughts?

Westbrook:
> ```
>if (this.hasAttribute('side-aligned')) {
>             this.resolvedElement.element?.setAttribute('sideLabel', 'true');
>  }
> ```

Roughly, yes. This versions applied this value to everything, which may be right, but may not happen with the same attribute/property over time. What do you think about altering this type to include a second argument that is the entire Field Label class?

```diff
- applyFocusElementLabel?: (label?: string) => void;
+ applyFocusElementLabel?: (appliedLabel: string, labelElement?: FieldLabel) => void;
```

Then we can leverage all of the data on the FieldLabel as we see fit in the labeled element. Something like:

```js
applyFocusElementLabel = (value: string, labelElement: FieldLabel): void => {
    this.appliedLabel = value;
    if (labelElement.sideAligned === 'start') {
       ...
    }
};
```

Rajdeepc:
> ```diff
> FieldLabel
> ```

Yes! I would agree with this approach which would definitely detach the logic from the Field Label class. Thanks

PR #3706, opened 2023-10-06
fix(overlay): allow overlay to persist on hover

### LINKED ISSUE #3366: [Bug][a11y]: Tooltip content not hoverable
### Code of conduct

### Impacted component(s)

Tooltip

### Expected behavior

When a user hovers over an element that triggers a tooltip, the tooltip should appear. The tooltip should remain visible and accessible when the user moves their mouse pointer onto the tooltip itself. This should allow users to interact directly with the content inside the tooltip, such as clicking on a link within the tooltip. The tooltip should only disappear when the user moves their mouse pointer away from both the tooltip and the triggering element, or if the user explicitly dismisses the tooltip (e.g., using the Esc key).

### Actual behavior

When a user navigates to an element with a tooltip and hovers over it, the tooltip appears as expected. However, when the user moves their mouse pointer onto the tooltip itself, the tooltip disappears. This makes it impossible for users to interact directly with the content inside the tooltip.

### Screenshots

_No response_

_No response_

1. Navigate to the element triggering the tooltip
2. Hover over the element to trigger the tooltip
3. Try moving the mouse pointer onto the tooltip content itself

Result:
The tooltip disappears as soon as the mouse pointer leaves the boundary of the triggering element and enters the tooltip.

_No response_

_No response_

najikahalsema:
## Description

A timer based approach to allowing hover content to persist so that a user can hover over it. 

## Related issue(s)
- fixes #3366 

## Motivation and context

Blocks Coachmark PR and Rich Tooltip PR. This also matches the expected behaviour for tooltips outlined by [WACG](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role).

## How has this been tested?

-   [ ] _Before_
    1. Go [here](https://opensource.adobe.com/spectrum-web-components/storybook/?path=/story/tooltip--overlaid-top)
    2. Hover over any of the trigger elements
    3. Attempt to hover over a tooltip
    4. See that it is not possible! 
    5. You can also try the same test [here](https://opensource.adobe.com/spectrum-web-components/storybook/?path=/story/overlay--click-and-hover-targets)
-   [ ] _After_
    1. Go [here](https://halsema-overlay-hover-handler--spectrum-web-components.netlify.app/storybook/?path=/story/tooltip--overlaid-top)
    2. Hover over any of the trigger elements
    3. Attempt to hover over a tooltip
    4. See that it is now possible! 
    5. Once you leave the triggerElement and go somewhere that isn't the overlay, the tooltip should close shortly after. 
    6. You can try the same test [here](https://halsema-overlay-hover-handler--spectrum-web-components.netlify.app/storybook/?path=/story/overlay--click-and-hover-targets).

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/overlay/src/Overlay.ts
+++ b/packages/overlay/src/Overlay.ts
@@ -808,57 +808,18 @@ export class Overlay extends OverlayFeatures {
     // set a timeout once the pointer enters and the overlay is shown
     // give the user time to enter the overlay
 
-    protected handleOverlayPointerenter = (event: PointerEvent): void => {
-        if (
-            this.triggerElement === event.relatedTarget ||
-            (this.hasNonVirtualTrigger &&
-                (this.triggerElement as HTMLElement).contains(
-                    event.relatedTarget as Node
-                ))
-        ) {
-            return;
-        }
-
+    protected handleOverlayPointerenter = (): void => {
         if (this.hoverTimeout) {
             clearTimeout(this.hoverTimeout);
             delete this.hoverTimeout;
         }
     };
 
-    protected handlePointerleave = (event: PointerEvent): void => {
-        if (
-            this === event.relatedTarget ||
-            this.contains(event.relatedTarget as Node) ||
-            [...this.children].find((child) => {
-                if (child.localName !== 'slot') {
-                    return false;
-                }
-                return (child as HTMLSlotElement)
-                    .assignedElements({ flatten: true })
-                    .find((el) => {
-                        return (
-                            el === event.relatedTarget ||
-                            el.contains(event.relatedTarget as Node)
-                        );
-                    });
-            })
-        ) {
-            return;
-        }
-
+    protected handlePointerleave = (): void => {
         this.doPointerleave();
     };
 
-    protected handleOverlayPointerleave = (event: PointerEvent): void => {
-        if (
-            this.triggerElement === event.relatedTarget ||
-            (this.hasNonVirtualTrigger &&
-                (this.triggerElement as HTMLElement).contains(
-                    event.relatedTarget as Node
-                ))
-        ) {
-            return;
-        }
+    protected handleOverlayPointerleave = (): void => {
         this.doPointerleave();
     };
 
--- a/packages/overlay/src/Overlay.ts
+++ b/packages/overlay/src/Overlay.ts
@@ -53,6 +53,7 @@ import { PlacementController } from './PlacementController.js';
 import styles from './overlay.css.js';
 
 const LONGPRESS_DURATION = 300;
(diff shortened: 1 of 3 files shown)


packages/overlay/test/overlay.test.ts
```diff
@@ -768,7 +769,7 @@ describe('Overlay - type="modal"', () => {
     });
 });
 describe('Overlay - timing', () => {
-    it('manages multiple modals in a row without preventing them from closing', async () => {
+    it.only('manages multiple modals in a row without preventing them from closing', async () => {
```
Westbrook, line 772: Remove `.only`

packages/overlay/test/overlay.test.ts
```diff
@@ -833,8 +835,9 @@ describe('Overlay - timing', () => {
                 },
             ],
         });
-        await nextFrame();
-        await nextFrame();
+        // allow "hover" content to close by waiting for its timer to complete
+        await aTimeout(400);
```
Westbrook, line 839: If the tooltip should be “closed” at this point, could we leverage an `sp-closed` listener instead?

Westbrook, line 839: If the tooltip should be “closed” at this point, could we leverage an `sp-closed` listener instead?

packages/overlay/test/overlay.test.ts
```diff
@@ -875,7 +878,7 @@ describe('Overlay - timing', () => {
             ],
         });
         await closed;
-
+        await aTimeout(200);
```
Westbrook, line 881: What does this wait for?

Westbrook (commented):
Let's double check that the tests you change continues to test what it did while adapting to this new feature. At least the comments seem out of date, but it's not clear that we're serving both the original and new intent appropriately.

packages/overlay/src/Overlay.ts
```diff
@@ -53,6 +53,7 @@ import { PlacementController } from './PlacementController.js';
 import styles from './overlay.css.js';
 
 const LONGPRESS_DURATION = 300;
+const HOVER_DELAY = 300;
```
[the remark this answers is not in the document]
najikahalsema: Great question! And great catch! I actually just picked an arbitrary number. I think I started with 400 and subconsciously changed it to 300 when I put this variable next to the longpress one 😅 I'll circle back on this and put some rationale behind this number and change it if needed!

Westbrook (approved):
Nice work here! That removal of the old half logic might make this a new loss in end user JS, too 🚀 

:shipit:

PR #3708, opened 2023-10-09
add lighthouse benchmark in Github Actions

Rajdeepc:
## Description

New Job is added to test.yml workflow which takes in the dynamic docs url based on the branch name and run lighthouse benchmark on index.html
For now only the index page is benchmarks but we will eventually add all the routes so that performance profiling runs on every component routes.

Lighthouse config

Only Performance is being audited now with dom-size audit settings. We can expand this based on our need.

## Motivation and context

Lighthouse reporting is essential to keep our docs page faster and also to run performance profiling on SWC components.

## How has this been tested?

-   [x] _Test case 1_
https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1707302671447-43007.report.html

## Types of changes

## Checklist

## Best practices


Westbrook (commented):
A version of this that added `.lighthouseci` to the `.gitignore` and uses the config to allow our current usage to "pass" would be a great first step here.

With the tool in repo, we could then create issues/branches for each test that would normally fail and use those to incrementally get back to the performance we'd want to target here.

Once `index.html` was passing, we could then move to other pages in the site, too.

packages/radio/src/radio.css
```diff
@@ -23,7 +23,7 @@ governing permissions and limitations under the License.
 /**
  * Work around for https://github.com/adobe/spectrum-css/issues/2251
  **/
-:host([dir="rtl"]) #button:after {
+:host([dir='rtl']) #button:after {
```
castastrophe, line 26: This seems unrelated to the rest of the PR - was this intentionally included?

.github/workflows/main.yml
```diff
@@ -0,0 +1,17 @@
...
+    lighthouseci:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v3
+            - uses: actions/setup-node@v3
+              with:
+                  node-version: '18'
+                  cache: 'yarn'
+                  registry-url: 'https://registry.npmjs.org'
+            - run: npm install -g @lhci/cli@0.12.x
```
castastrophe, line 13: Since lighthouse is installed as part of the monorepo, you shouldn't need to install it separately here. The yarn install that follows will include the lighthouse cli.

.github/workflows/main.yml
```diff
@@ -0,0 +1,17 @@
...
+            - name: Install dependencies
+              run: yarn --frozen-lockfile
+            - run: yarn docs:build
+            - run: lhci autorun
```
castastrophe, line 17: Instead of running this here, I would use the yarn command `yarn lighthouse` so that you're using the locally installed version.

lighthouserc.cjs
```diff
@@ -0,0 +1,44 @@
...
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+module.exports = {
+    ci: {
+        collect: {
+            url: [
+                'http://localhost:8000/',
```
castastrophe, line 16: Do the docs always use 8000 or does it fallback to an open port if that is in use? I ask because you might need to use an environment variable or something else here.

castastrophe (changes requested):

lighthouserc.cjs
Rajdeepc: Docs will always use 8000 as it is 11ty default port

.github/lighthouse/lighthouserc.json
```diff
@@ -0,0 +1,10 @@
+{
+    "ci": {
+        "collect": {
+            "settings": {
+                "onlyCategories": ["performance"],
```
Westbrook, line 5:
```suggestion
                "onlyCategories": [
                    // Excluded categories do not currently pass
                    // "accessibility",
                    // "best-practices",
                    "performance",
                     // "pwa",
                     // "seo"
                ],
```
A

.github/lighthouse/lighthouserc.json
```diff
@@ -0,0 +1,10 @@
...
+                "onlyAudits": ["dom-size"]
```
Westbrook, line 6: Is there an equivalent list we could include like above?

Westbrook (commented):
Getting pretty close! Just want to document some next steps in line of the code and we should be good to go.

.github/lighthouse/lighthouserc.json
Rajdeepc: I don't think we can add comments in json. :(

Westbrook: Can we use a JS config like this one? https://github.com/GoogleChrome/lighthouse/blob/main/docs/configuration.md#usage

Rajdeepc: Updated to
```
export default {
    extends: 'lighthouse:default',
    settings: {
        onlyCategories: [
            // Excluded categories do not currently pass
            // "accessibility",
            // "best-practices",
            'performance',
            // "pwa",
            // "seo"
        ],
        onlyAudits: ['first-meaningful-paint', 'speed-index', 'interactive'],
    },
};
```

.github/lighthouse/lighthouserc.json
Rajdeepc: Yes as here
https://github.com/GoogleChrome/lighthouse/blob/ed04b10ccfa4e82ee5cd5b437e474a12c1ba52d0/core/config/default-config.js#L178

Rajdeepc: also added in the config

Issue #3398, opened 2023-06-29
[Feat][Button]: Support `icon-only` attribute


Westbrook:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Button

### Description of the requested feature

Spectrum CSS recently added support for [icon only Buttons](https://opensource.adobe.com/spectrum-css/button-accent.html#sizing), we should add them to SWC.

### Mockups or screenshots

<img width="48" alt="image" src="https://github.com/adobe/spectrum-web-components/assets/1156657/6e60fdf2-b7ed-4faa-9639-9f3da3c3d1fb">

### Implementation notes or ideas

The `spectrum-Button--iconOnly` class leveraged in the CSS output should be converted to an `icon-only` attribute. See documentation for this sort of conversion, [here](https://opensource.adobe.com/spectrum-web-components/guides/spectrum-config/).

If present this attribute should _visually_ suppress the `id="label"` element that delivers slotted content so that it can still be made available to screen readers.

We'll need to add size, variant, static, and treatment versions of this pattern to the various Storybook entries for Button, and if there was a sensible functional test to include with this addition, that would be great, too.

Follow up work, once v2 of the Overlay API has been shipped, can expand on this pattern to deliver the provided text content in a Tooltip, as outlined in the [specification for this pattern](https://spectrum.adobe.com/page/button/#Label-and-icon).

najikahalsema:
The icon-only button should display the label for button in a tooltip, according to the Spectrum specification. There are also style differences and functional differences that need to be explored and addressed.

PR #3716, opened 2023-10-11
chore(button): CSS for icon-only attribute

Rajdeepc:
## Description
Added CSS for icon-only attribute for `sp-button`

## Motivation and context

To solve the below issue on a rounded button with icon
Screenshot 2023-10-11 at 7 01 14 PM

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices


Westbrook:
FYI: https://github.com/adobe/spectrum-web-components/issues/3398

najikahalsema (approved):

Rajdeepc:
@najikahalsema @Westbrook If everything looks good here can we merge this?

najikahalsema (approved):

najikahalsema (dismissed):
Disregard my approval. Can you investigate why the build is failing to meet test coverage in chromium?

Rajdeepc:
> Disregard my approval. Can you investigate why the build is failing to meet test coverage in chromium?

The build is failing due to code coverage. Can we take this up in a separate ticket as this is unrelated to the changes in this PR?
Screenshot 2023-11-02 at 12 15 49 AM

Rajdeepc:
> Disregard my approval. Can you investigate why the build is failing to meet test coverage in chromium?

Should be good now

Westbrook (commented):
Some things for this:
- please add stories with the icon only button in isolation so we can manage it via regression testing
- Should a user be able to set `buttonRef.iconOnly = true` or do we want to require `buttonRef.toggleAttribute('icon-only', true)`? Either way works, but we should add content about this to the documentation site based on the decisions we make there for today.
- see https://spectrum.adobe.com/page/button/#Label-and-icon, particularly:
  > The label can be hidden to create an icon-only button. If the label is hidden, an icon is required, and the label will appear in a tooltip.
  
  I'm OK not to block on that, though it would be nice to have this feature, too. We can land this for Coachmark work as is (with stories) and then keep #3398 open for the follow up work around a Tooltip.

Rajdeepc:
> Some things for this:
> 
> * please add stories with the icon only button in isolation so we can manage it via regression testing
> * Should a user be able to set `buttonRef.iconOnly = true` or do we want to require `buttonRef.toggleAttribute('icon-only', true)`? Either way works, but we should add content about this to the documentation site based on the decisions we make there for today.
> * see https://spectrum.adobe.com/page/button/#Label-and-icon, particularly:
>   > The label can be hidden to create an icon-only button. If the label is hidden, an icon is required, and the label will appear in a tooltip.
>   
>   
>   I'm OK not to block on that, though it would be nice to have this feature, too. We can land this for Coachmark work as is (with stories) and then keep [[Feat][Button]: Support `icon-only` attribute #3398](https://github.com/adobe/spectrum-web-components/issues/3398) open for the follow up work around a Tooltip.

@Westbrook Added the required stories and also updated the tests and docs to cater icon-only attribute for Icon Only Button element

packages/button/test/button.test.ts
```diff
@@ -84,7 +84,7 @@ describe('Button', () => {
     it('loads default only icon', async () => {
         const el = await fixture<Button>(
             html`
-                <sp-button label="Button">
+                <sp-button label="Button" icon-only>
```
Westbrook, line 87: 👏🏼 👏🏼 👏🏼

Westbrook (approved):
LGTM. Let's land this plane 🛬

PR #3721, opened 2023-10-13
fix(grid): plug a mememory leak from the render process

### LINKED ISSUE #3707: [Memory-leak]: sp-grid not releasing its cached cards, when it's out of DOM, and cards retained as detached dom elements in heap
### Code of conduct

### Impacted component(s)

sp-grid

### Expected behavior

Cards of sp-grid should not take memory as detached dom elements when sp-grid disconnected from DOM.

### Actual behavior

Cards of sp-grid being retained as detached dom nodes taking memory space, even sp-grid got disconnected from DOM.

### Screenshots

Screenshot 2023-10-09 at 12 09 39 PM
Screenshot 2023-10-10 at 2 11 40 PM

_No response_

Do it in MSEdge with it's Detached Elements panel on

1. Go to https://new.express.adobe.com/
2. Click on 'Your-stuff' from left options
3. Get it load some assets
4. Return back to home
5. Observe Detached nodes

or

1. Run the storybook after cloning this repo.
2. Go Grid
3. Let some cards load
4. Go to any other component
5. Reload detached elements and observe detached nodes.

If you will repeat, you can clearly see detached nodes increasing.

_No response_

_No response_

Westbrook:
## Description
Rendering a `lit-html` template with `render()` without caching it's root part means that you can't inform it when the template is connected to the DOM or not. This connection state is required to ensure clean up can happen to any of the template directives that might be leveraged therein. If the Virtualize template directive does not get disconnected it can leak references to the items it had previously rendered to the DOM.

Cache a reference to the "grid part" in the Grid element for the DOM that would otherwise have been rendered to the render root and managed directly by `LitElement`. Manage the connected state of this cached part.

## Related issue(s)
- fixes #3707

## How has this been tested?
-   [ ] _Test case 1_
    1. Follow the steps in https://github.com/adobe/spectrum-web-components/issues/3707#issuecomment-1754719002

## Types of changes

## Checklist

### DIFF
--- a/tools/grid/test/grid-memory.test.ts
+++ b/tools/grid/test/grid-memory.test.ts
@@ -1,0 +1,72 @@
+/*
+Copyright 2023 Adobe. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License. You may obtain a copy
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import { expect, fixture, nextFrame } from '@open-wc/testing';
+import { html, render } from '@spectrum-web-components/base';
+import { Default } from '../stories/grid.stories';
+
+async function usedHeapMB(): Promise<number> {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    const memorySample = performance.measureUserAgentSpecificMemory();
+    return (await memorySample).bytes / (1024 * 1024);
+}
+
+describe('Grid memory usage', () => {
+    it('releases references on disconnect', async function () {
+        if (!window.gc || !('measureUserAgentSpecificMemory' in performance))
+            this.skip();
+
+        this.timeout(10000);
+
+        const iterations = 50;
+        let active = false;
+
+        const el = await fixture<HTMLElement>(
+            html`
+                <div></div>
+            `
+        );
+
+        async function toggle(
+            forced: boolean | undefined = undefined
+        ): Promise<void> {
+            active = forced != null ? forced : !active;
+            render(active ? Default() : html``, el);
+            await nextFrame();
+            await nextFrame();
+        }
+
+        // "shake things out" to get a good first reading
+        for (let i = 0; i < 5; i++) {
+            await toggle();
+        }
+        await toggle(false);
+        const beforeMB = await usedHeapMB();
+
+        for (let i = 0; i < iterations; i++) {
+            await toggle();
+        }
+        await toggle(false);
+        const afterMB = await usedHeapMB();
+
+        /**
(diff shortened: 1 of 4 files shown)


tools/grid/test/grid-memory.test.ts
```diff
@@ -0,0 +1,72 @@
...
+
+        /**
+         * An actually leak here shapes up to be more than 10MB per test,
+         * we could be more linient later, if needed, but the test currently
+         * shows less heap after the test cycle.
+         */
+        expect(
+            afterMB - beforeMB,
+            `before: ${beforeMB}, after: ${afterMB}`
+        ).to.be.lt(0);
```
nickschaap, line 70: You can also potentially use a WeakRef to the Grid element and wait a certain amount of time/ trigger garbage collection and expect the deref() to return undefined. This might be more deterministic than measuring heap memory.

Westbrook: Opted for the more reliable, but slightly more cumbersome to surface, `[performance.measureUserAgentSpecificMemory()](https://chromestatus.com/feature/5685965186138112)` for now. Showing even more positive results from this change when measuring with this API.

I know the team has been looking at tooling with WeakMaps, I wonder how you're actually holding things in that way as it feels like the need for the key would inherently keep a reference and prevent the memory from being released. You wouldn't happen to have any examples to share?

nickschaap: We're mainly using [WeakRef](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef#instance_methods) different from [WeakMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap). The basic idea is to construct a new WeakRef passing in the DOM node you're interested in tracking, trigger garbage collection, and then test whether or not calling WeakRef.deref() on your reference returns undefined. If so, you can. assume the node was garbage collected. It not [fool-proof](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef#avoid_where_possible) but it has been working quite nicely for us.

Here is the original dom inspection snippets that Kin created: https://git.corp.adobe.com/jblas/debug-utils/tree/main/snippets#object-watcher

najikahalsema (dismissed):
LGTM! Never used MSEdge until today 😅

web-test-runner.config.js
```diff
@@ -135,5 +143,5 @@ export default {
         },
     ],
     group: 'unit',
-    browsers: [chromium, firefox, webkit],
+    browsers: [chromiumWithMemoryTooling, firefox, webkit],
```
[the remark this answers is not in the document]
Westbrook: In that the tests run and pass, I have no other path to judge "safety" here.

This is a great point though, and segmenting these out, for safety/performance/understandability/etc, as part of https://github.com/adobe/spectrum-web-components/issues/3722 would make a lot of sense!

tools/grid/test/grid-memory.test.ts
Westbrook: This is cool, I didn't realize that WeakRef was different than WeakMap! Definitely a useful pattern for the future, particularly as a comparison for test speed for tightening the feedback cycle. 

Digging into `measureUserAgentSpecificMemory()` will also give us some other useful tidbits later on, for an expanded perf testing suite, so getting started now seems useful.

PR #3735, opened 2023-10-20
Playwright upgrade

Westbrook:
## Description
- move to node@18
- use playwright@1.37.1
  - there are a couple of newer versions available, but they don't seem to play well with CircleCI, will need to investigate further, separately
- massage config
  - move system wide test concurrency rules to specific browsers
    - firefix gets 1 concurrency due to https://github.com/microsoft/playwright/issues/27717
      - this takes an annoyingly long amount of time, but better stable than fast here
    - safari gets 4 so that it doesn't overrun in CI
    - chrome gets no rule so that it can negotiate with system resources on its own
  - move to concurrent browser usage, as it seems to no longer cause issues
  - add some retries in various places to stabilize results
  - remove chrome with flags as "stable" chrome in this version supports `[popover]`
- skip some tests in WebKit for this issue: https://bugs.webkit.org/show_bug.cgi?id=263081
  - looks like there's already a PR, so hopefully it's in TP/stable soon, however this will mean that the newer versions of playwright will need to work in CircleCI to adopt, see above
- fix a couple tests that were wrong
- update the send mouse plugin to reset the pointer position at the end of each test that uses it 

## Motivation and context
Get closer to current browser matrix at test time

## How has this been tested?
-   [ ] _Test case 1_
    1. Go [here](https://app.circleci.com/pipelines/github/adobe/spectrum-web-components/18266/workflows/7cb84e62-8e56-4f81-b38f-80dabc0ba0ff)
    2. See that CI passed
-   [ ] _Test case 2_
    1. Take the branch onto your local system
    2. `yarn`
    3. `yarn test`
    4. See that the tests pass (I'm averaging about 3 minutes right now with an M1 Max Macbook 😓)

## Types of changes

## Checklist


najikahalsema (approved):
LGTM! Thanks for resetting the pointer location. I timed both test runs on here and main, and it feels like a 40 second difference. Not too bad.

PR #3747, opened 2023-10-26
added overlay story - context menu using declarative api

TarunAdobe:
## Description

added overlay story - context menu using declarative api

## Related issue(s)

- 

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/overlay/stories/overlay.stories.ts
+++ b/packages/overlay/stories/overlay.stories.ts
@@ -1424,10 +1424,13 @@ export const virtualElementDeclaratively = (
         const overlay = document.querySelector(
             'sp-overlay:not([open])'
         ) as Overlay;
-        overlay.triggerElement.updateBoundingClientRect(
-            event.clientX,
-            event.clientY
-        );
+
+        if (overlay.triggerElement instanceof VirtualTrigger) {
+            overlay.triggerElement.updateBoundingClientRect(
+                event.clientX,
+                event.clientY
+            );
+        }
         overlay.willPreventClose = true;
         overlay.open = true;
     };
--- a/packages/overlay/stories/overlay.stories.ts
+++ b/packages/overlay/stories/overlay.stories.ts
@@ -1413,3 +1413,78 @@ export const virtualElement = (args: Properties): TemplateResult => {
 virtualElement.args = {
     placement: 'right-start' as Placement,
 };
+
+export const virtualElementDeclaratively = (
+    args: Properties
+): TemplateResult => {
+    const handleContextmenu = async (event: PointerEvent): Promise<void> => {
+        event.preventDefault();
+        event.stopPropagation();
+
+        const overlay = document.querySelector(
+            'sp-overlay:not([open])'
+        ) as Overlay;
+        overlay.triggerElement.updateBoundingClientRect(
+            event.clientX,
+            event.clientY
+        );
+        overlay.willPreventClose = true;
+        overlay.open = true;
+    };
+    const overlay = (): TemplateResult => html`
+        <sp-overlay
+            offset="0"
+            type="auto"
+            placement=${args.placement}
+            .triggerElement=${new VirtualTrigger(0, 0)}
+        >
+            <sp-popover
+                style="width:300px;"
+                @change=${(event: PointerEvent) => {
+                    event.target?.dispatchEvent(
+                        new Event('close', { bubbles: true })
+                    );
+                }}
+            >
+                <sp-menu>
+                    <sp-menu-group>
+                        <span slot="header">Menu</span>
+                        <sp-menu-item>Deselect</sp-menu-item>
+                        <sp-menu-item>Select inverse</sp-menu-item>
+                        <sp-menu-item>Feather...</sp-menu-item>
+                        <sp-menu-item>Select and mask...</sp-menu-item>
+                        <sp-menu-divider></sp-menu-divider>
+                        <sp-menu-item>Save selection</sp-menu-item>
+                        <sp-menu-item disabled>Make work path</sp-menu-item>
+                    </sp-menu-group>
+                </sp-menu>
+            </sp-popover>
+        </sp-overlay>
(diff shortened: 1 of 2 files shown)


packages/overlay/stories/overlay.stories.ts
```diff
@@ -1413,3 +1413,72 @@ export const virtualElement = (args: Properties): TemplateResult => {
...
+        event.stopPropagation();
+
+        const overlay = document.querySelector('#overlay') as Overlay;
+
+        const virtualTrigger = new VirtualTrigger(event.clientX, event.clientY);
+
+        overlay.triggerElement = virtualTrigger;
+        overlay.placement = args.placement;
+        overlay.offset = 0;
+        overlay.type = 'auto';
```
Westbrook, line 1431: We can place these three declaratively into the story.

packages/overlay/stories/overlay.stories.ts
```diff
@@ -1413,3 +1413,72 @@ export const virtualElement = (args: Properties): TemplateResult => {
...
+                    @click=${(event: PointerEvent) => {
+                        if (
+                            (event.target as HTMLElement).localName ===
+                            'sp-menu-item'
+                        ) {
+                            event.target?.dispatchEvent(
+                                new Event('close', { bubbles: true })
+                            );
+                        }
+                    }}
```
Westbrook, line 1462: I see this is adopted from the other story. I can't remember why I leveraged this, rather than the `change` event on the `<sp-menu>` which already does the element checking for us. Should be able to use:
```
<sp-menu
    @change=${() => {
        event.target?.dispatchEvent(
            new Event('close', { bubbles: true })
        );
    }}
```
Instead.

Westbrook (commented):
I don't see being able to open the _next_ context menu while the first is opened here... can you look into that?

packages/overlay/stories/overlay.stories.ts
```diff
@@ -1413,3 +1413,70 @@ export const virtualElement = (args: Properties): TemplateResult => {
...
+export const virtualElementDeclaratively = (
+    args: Properties
+): TemplateResult => {
+    const handleContextmenu = async (event: PointerEvent): Promise<void> => {
+        event.preventDefault();
+        event.stopPropagation();
+
+        const overlay = document.querySelector('#overlay') as Overlay;
+
+        const virtualTrigger = new VirtualTrigger(event.clientX, event.clientY);
```
Westbrook, line 1426: I think we want to have a stable `virtualTrigger`. This would allow us to attach it to the `<sp-overlay>` via `.trigger=${virtualTrigger}` and then leverage `updateBoundingClientRect()` on the virtual trigger to update the position, which also calls `Overlay.update()` for you under the hood.

packages/overlay/stories/overlay.stories.ts
```diff
@@ -1413,3 +1413,76 @@ export const virtualElement = (args: Properties): TemplateResult => {
...
+        virtualTrigger.updateBoundingClientRect(event.clientX, event.clientY);
+        overlay.willPreventClose = true;
+        overlay.open = true;
+    };
+    const overlay = (): TemplateResult => html`
+        <sp-overlay
+            offset="0"
+            type="auto"
+            placement=${args.placement}
+            .triggerElement=${virtualTrigger}
```
Westbrook, line 1437: Binding these both to the same virtual trigger does a weird thing in which always both menus are in the same place making it look like the previously opened Menu winks closed without making its transition. I think maybe we make a `virtualTrigger` for each and then address them in `handleContextmenu` as `overlay.triggerElement.updatedBindingClientRect()` instead.

TarunAdobe: I updated it with separate virtual triggers... However I don't see any difference in transition :(

packages/overlay/stories/overlay.stories.ts
```diff
@@ -1413,3 +1413,81 @@ export const virtualElement = (args: Properties): TemplateResult => {
...
+        </div>
+    `;
+};
+
+virtualElementDeclaratively.args = {
+    placement: 'right-start' as Placement,
+};
+
+virtualElementDeclaratively.swc_vrt = {
+    skip: true,
```
Westbrook, line 1492: With this, you should be able to remove the changes in `.circleci/config.yml` and CI will go green.

packages/overlay/stories/overlay.stories.ts
Westbrook: When comparing https://65812823fcc49166eb0173e8--spectrum-web-components.netlify.app/storybook/?path=/story/overlay--virtual-element-declaratively to https://chore-overlay-declarative-context-menu--spectrum-web-components.netlify.app/storybook/?path=/story/overlay--virtual-element-declaratively you're not able to see the current minus one Menu fade out? If not, check them out at 10% speed:

Previous (one virtual trigger):

https://github.com/adobe/spectrum-web-components/assets/1156657/dbd51a4d-4f1f-47f6-82eb-0848876df0d3

Current (two virtual triggers):


https://github.com/adobe/spectrum-web-components/assets/1156657/21b869d3-36c9-4cbc-971f-3ffc89a6a6f5

packages/overlay/stories/overlay.stories.ts
Westbrook: Once we're not updating the golden image cache and CI is green then I think this demo is ready to go to `main`.

packages/overlay/stories/overlay.stories.ts
TarunAdobe: ok damn! It wasn't even noticeable for me. Thanx for attaching the video!

Westbrook (approved):
Thanks for working through this nice addition for our consumers! 🥳

PR #3748, opened 2023-10-26
fix: add delay to overlay-trigger

AndreiBaicu26:
This PR adds the `delay` capability back to `overlay-trigger`

## Motivation and context

Adding back the capability of delayed overlays.

## How has this been tested?

-   [ ] _Test case 1_
     Inside storybook created a button with a delayed tooltip through overlay-trigger.

-   [ ] _Test case 2_
     Called the Overlay.open() method to test backwards compatibility.

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/overlay/src/Overlay.ts
+++ b/packages/overlay/src/Overlay.ts
@@ -86,6 +86,12 @@ export class Overlay extends OverlayFeatures {
 
     abortController!: AbortController;
 
+    @query('.dialog')
+    override dialogEl!: HTMLDialogElement & {
+        showPopover(): void;
+        hidePopover(): void;
+    };
+
     /**
      * An Overlay that is `delayed` will wait until a warm-up period of 1000ms
      * has completed before opening. Once the warmup period has completed, all
@@ -95,13 +101,17 @@ export class Overlay extends OverlayFeatures {
      * provided that option.
      */
     @property({ type: Boolean })
-    override delayed = false;
+    override get delayed(): boolean {
+        if (this.elements.length === 0) {
+            return false;
+        }
 
-    @query('.dialog')
-    override dialogEl!: HTMLDialogElement & {
-        showPopover(): void;
-        hidePopover(): void;
-    };
+        if (this.elements[0].getAttribute('delayed') !== null) {
+            return true;
+        }
+
+        return false;
+    }
 
     /**
      * Whether the overlay is currently functional or not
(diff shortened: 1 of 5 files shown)


AndreiBaicu26:
@Westbrook Something like the new refactoring ? 👀
This would be cleaner yes and I believe a getter should be enough ?

packages/overlay/src/Overlay.ts
```diff
@@ -95,13 +101,23 @@ export class Overlay extends OverlayFeatures {
...
-    @query('.dialog')
-    override dialogEl!: HTMLDialogElement & {
-        showPopover(): void;
-        hidePopover(): void;
-    };
+        if (this.elements[0].getAttribute('delayed') !== null) {
+            return true;
+        }
+
+        return this._delayed;
```
Westbrook, line 113: Is this all enough the same as `return this.elements.at(-1)?.hasAttribute('delayed') ?? this._delayed;`?

AndreiBaicu26: Almost, changed to `this.elements.at(-1)?.hasAttribute('delayed') || this._delayed;`

Westbrook:
It would be nice if we could add a test for this. Maybe there's room to add a little time checking between the hover and open events in https://github.com/adobe/spectrum-web-components/blob/main/packages/overlay/test/overlay-trigger-hover-click.test.ts#L92-L111 or so. Probably can go without blocking on this, but maybe you've got an idea on testing this other wise?

After that, I _think_ we're just missing some changes from `main` and this will be good to go. Can you catch this up to that branch?

AndreiBaicu26:
@Westbrook Rebased with main + added a test for delay on click ✅

packages/overlay/test/overlay-trigger-click.test.ts
```diff
@@ -174,4 +177,37 @@ describe('Overlay Trigger - Click', () => {
...
+            </overlay-trigger>
+        `);
+
+        await elementUpdated(el);
+        el.setAttribute('open', 'click');
+        await waitUntil(
+            () => openedSpy.calledOnce,
+            'hover content projected to overlay',
+            { timeout: 8000 }
+        );
```
Westbrook, line 208:
```suggestion
        const start = performance.now();
        const opened = oneEvent(el, 'sp-opened');
        el.setAttribute('open', 'click');
        await opened;
```

If you structure the test like this you can get away without either of the spies and can be sure to only test the open workflow, rather than any delay that might come from the generation of the fixture.

Westbrook (commented):
Looking good! Tiny refactor in the test for simplicity, and then let's ship it!

Westbrook (approved):
LGTM! Thanks for bringing this to our attention and finding such a clean fix. 🙇🏼

Issue #2406, opened 2022-07-08
[Feat]: Name attribute for input components


Illusive7Man:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-textfield, sp-picker, sp-checkbox, sp-switch

### Description of the requested feature

Without name attribute there is no browser autocomplete.
For example a login form made using `<sp-textfield>` components will never be auto-filled by Chrome.

blunteshwar:
@Illusive7Man sp-textfield, sp-checkbox, sp-switch are doable but i think sp-picker is not an input component how can it have a name attribute for autocomplete

Westbrook:
@blunteshwar we should spend some time looking into how [this feature](https://html.spec.whatwg.org/multipage/custom-elements.html#form-associated-custom-elements:face-state-6) supports doing just that or not.

blunteshwar:
Sure!
We can discuss this in our tuesday session

PR #3752, opened 2023-10-31
feat(textfield): added name attribute to textfield

blunteshwar:
## Description

Added name attribute to sp-textfiled, sp-picker, sp-swicth, sp-checkbox

## Related issue(s)
 [[Feat]: Name attribute for input components #2406](https://github.com/adobe/spectrum-web-components/issues/2406)

-

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices


packages/checkbox/src/CheckboxBase.ts
```diff
@@ -24,6 +25,9 @@ export class CheckboxBase extends Focusable {
     @property({ type: Boolean, reflect: true })
     public readonly = false;
 
+    @property({ type: String, reflect: true })
+    public name: string | undefined;
```
najikahalsema, line 29: Hi, I have a question for you. I'm trying to understand the rationale behind the decision of writing the `name` property type like this. If the goal is that if a user doesn't include the `name` attribute, then we want to explicitly state that `name = undefined`, right?
```
@property({type: String, reflect: true})
public name: string = ''; 
```
^ So something like this would defeat the purpose of the goal of form elements using autocomplete. Am I correct in understanding that??

packages/checkbox/src/CheckboxBase.ts
```diff
@@ -54,6 +58,7 @@ export class CheckboxBase extends Focusable {
     protected override render(): TemplateResult {
         return html`
             <input
+                name=${ifDefined(this.name || undefined)}
```
najikahalsema, line 61: Continuing on here, could we do something like 

```
name=${this.name ? this.name : undefined} 
```

instead? I think wrapping `undefined` in an `ifDefined()` doesn't make sense. 

If you were to do that approach, then you could likely write the `name` property as 
```
public name?: string
```

I think this is more in line with our conventions of writing code.

packages/checkbox/test/checkbox.test.ts
```diff
@@ -238,6 +238,13 @@ describe('Checkbox', () => {
         expect(el.checked).to.be.true;
     });
 
+    it('has name attribute', () => {
+        let el = testFixture.querySelector('#checkbox0') as Checkbox;
+
+        el = testFixture.querySelector('#checkbox1') as Checkbox;
+        expect(el.hasAttribute('name'));
+    });
```
najikahalsema, line 246: Could you also include a test for the expected value of `name` if `name = undefined`?

najikahalsema: Also, what is the value of `el.name` here?

najikahalsema: Also, what is the value of `el.name` here?

packages/textfield/src/Textfield.ts
```diff
@@ -237,6 +240,7 @@ export class TextfieldBase extends ManageHelpText(
                 : nothing}
             <!-- @ts-ignore -->
             <textarea
+                name=${ifDefined(this.name || undefined)}
```
najikahalsema, line 243: If `name` is a required attribute and has a value of an empty string no matter what, would this `undefined` state ever occur?

packages/textfield/test/textfield.test.ts
```diff
@@ -342,6 +342,37 @@ describe('Textfield', () => {
             : null;
         expect(input).to.not.be.null;
     });
+    it('handles `name` attribute', async () => {
```
najikahalsema, line 345: Again, curious to see a test where `name` is not specified by the user.

najikahalsema (changes requested):
I have some questions. Thank you for adding tests!

packages/checkbox/src/CheckboxBase.ts
blunteshwar: Autocomplete will work just fine with both `public name: string | undefined;` and `public name: string = ''; `
To answer your question, I used `public name: string | undefined;` because it makes more sense for name attribute to be either undefined or a valid string. When I write `public name: string = ''; ` instead of `public name: string | undefined;` the name attribute can be seen in the inspect element section although the user has not given it any value.
Screenshot 2023-11-10 at 9 19 07 AM
Screenshot 2023-11-10 at 9 20 12 AM

Also when I use  `public name: string = ''; ` I get the error shown above in image

packages/checkbox/src/CheckboxBase.ts
blunteshwar: I am not so sure but I think `name=${this.name ? this.name : undefined} ` will generate an error as shown in the below image
Screenshot 2023-11-10 at 9 25 01 AM

packages/checkbox/test/checkbox.test.ts
blunteshwar: the value here is undefined because  it's not yet initialised

blunteshwar: Added 2 extra tests here for the same

packages/textfield/src/Textfield.ts
blunteshwar: Here, I am using `ifDefined` because it renoves name attribute from dom(inspect element) when name is an empty string. If I don't use `ifDefined` then name attribute can be seen in inspect element section.
Screenshot 2023-11-10 at 9 52 02 AM

packages/textfield/test/textfield.test.ts
blunteshwar: done!

packages/checkbox/src/CheckboxBase.ts
najikahalsema: Gotcha. I think that's fair. However, would 
```
@property({type: String, reflect: true})
public name?: string; 
```
accomplish the same thing?

packages/checkbox/test/checkbox.test.ts
```diff
@@ -238,6 +238,17 @@ describe('Checkbox', () => {
         expect(el.checked).to.be.true;
     });
 
+    it('has name attribute', () => {
+        let el = testFixture.querySelector('#checkbox0') as Checkbox;
+
+        el = testFixture.querySelector('#checkbox1') as Checkbox;
+        expect(el.hasAttribute('name'));
+        expect(el.name).to.be.undefined;
+        expect(el.name == undefined);
```
najikahalsema, line 247:
```suggestion
```

You can remove this line since it's the same as line 246.

packages/textfield/test/textfield.test.ts
```diff
@@ -342,6 +342,34 @@ describe('Textfield', () => {
...
+        const el = await litFixture<Textfield>(
+            html`
+                <sp-textfield placeholder="Enter your name"></sp-textfield>
+            `
+        );
+        expect(el).to.not.equal(undefined);
+        expect(el.name).to.be.undefined;
+        expect(el.name === undefined);
```
najikahalsema, line 353:
```suggestion

```

najikahalsema (changes requested):
Tests look good. I'm curious still what the property looks like if it's optional, ie 

```
public name?: string
```

Let me know if it behaves the same as what you have now.

packages/checkbox/test/checkbox.test.ts
Westbrook: Remember what Hunter found about this syntax.

packages/checkbox/src/CheckboxBase.ts
blunteshwar: Nope, In this case when nothing is assigned to `name`, instead of being undefined it would be equal to "" (empty string).

packages/checkbox/test/checkbox.test.ts
blunteshwar: > You can remove this line since it's the same as line 246.

Done!

Issue #3753, opened 2023-10-31
[Bug]: sp-number-field value doesn't update on ArrowDown/ArrowUp keydowns or on scroll


mizgaionutalexandru:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-number-field

### Expected behavior

When focusing the input field and tapping the ArrowUp/ ArrowDown buttons, the component should display the updated/current value.

### Actual behavior

When focusing the input field and tapping the ArrowUp/ ArrowDown buttons, the component's display value does not change. It updates only when blur happens - clicking outside the component.

### Screenshots

In the first part of the video you can see what currently happens while I'm scrolling with focus on number field, while the last part shows what happens when tapping the ArrowUp/ArrowDown keys.

https://github.com/adobe/spectrum-web-components/assets/67201262/70b1cf9c-3a02-49be-a10f-6c65867aa97d

### What browsers are you seeing the problem in?

Firefox, Chrome, Safari, Microsoft Edge

### How can we reproduce this issue?

1. Go to [SWC Storybook](https://opensource.adobe.com/spectrum-web-components/storybook/?path=/story/number-field-sizes--s)
2. Click on the number field to focus it
3. Tap the ArrowUp/ArrowDown key 
4. See that the value doesn't update
5. Click outside the component
6. See that value updates

Westbrook:
Fixed by  #3754

PR #3754, opened 2023-10-31
fix(number-field): update display value on scroll and arrow up/down

mizgaionutalexandru:
Fixing the displayed value of sp-number-field on arrow keys and on scroll interactions.

## Related issue(s)

https://github.com/adobe/spectrum-web-components/issues/3753

## Motivation and context

It improves the user experience by showing in real time the changing value.

## How has this been tested?

-   [ ] _Test case 1_
    1. Go to number-field storybook
    2. Click on the number field to focus it
    3. Tap the ArrowUp/ArrowDown key or scroll
    4. See that the value updates in real time

In the first part of the video you can see the behaviour while I'm scrolling with focus on number field, while the second part shows what happens when tapping the ArrowUp/ArrowDown keys.

https://github.com/adobe/spectrum-web-components/assets/67201262/3d4270ce-5c02-487b-b1d4-066b62aa1e96

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/number-field/src/NumberField.ts
+++ b/packages/number-field/src/NumberField.ts
@@ -319,8 +319,11 @@ export class NumberField extends TextfieldBase {
         if (isNaN(this.value)) {
             value = min;
         }
+
         this._value = this.validateInput(value);
-        this.dispatchEvent(
+        this.inputElement.value = value.toString();
+
+        this.inputElement.dispatchEvent(
             new Event('input', { bubbles: true, composed: true })
         );
         this.indeterminate = false;
--- a/packages/number-field/test/number-field.test.ts
+++ b/packages/number-field/test/number-field.test.ts
@@ -73,10 +73,12 @@ describe('NumberField', () => {
     describe('receives input', () => {
         it('without language context', async () => {
             const el = await getElFrom(Default({ value: 1337 }));
+
             el.size = 's';
             expect(el.formattedValue).to.equal('1,337');
             expect(el.valueAsString).to.equal('1337');
             expect(el.value).to.equal(1337);
+            expect(el.focusElement.value).to.equal('1,337');
             el.focus();
             await sendKeys({ type: '7331' });
             await elementUpdated(el);
@@ -85,6 +87,7 @@ describe('NumberField', () => {
             expect(el.formattedValue).to.equal('13,377,331');
             expect(el.valueAsString).to.equal('13377331');
             expect(el.value).to.equal(13377331);
+            expect(el.focusElement.value).to.equal('13,377,331');
         });
         it('with language context', async () => {
             const [languageContext] = createLanguageContext('fr');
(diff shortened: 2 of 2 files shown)


packages/number-field/src/NumberField.ts
```diff
@@ -155,8 +155,8 @@ export class NumberField extends TextfieldBase {
         if (value === this.value) {
             return;
         }
-        this.lastCommitedValue = value;
         const oldValue = this._value;
+        this.lastCommitedValue = oldValue;
```
Westbrook, line 159: This feels contrary to what `lastCommittedValue` really means. However, I'll admit that the way the current consumption of that value works is a bit magical feeling...maybe there's something better to be done with that property.

packages/number-field/src/NumberField.ts
```diff
@@ -319,7 +319,8 @@ export class NumberField extends TextfieldBase {
         if (isNaN(this.value)) {
             value = min;
         }
-        this._value = this.validateInput(value);
+        this.value = value;
```
Westbrook, line 322: I'm not sure we want to trigger a render lifecycle in response to the `input` event. This _may_ be the cause of some of your failing tests.

Westbrook: My bad, indeed. But, I still don't see this as expecting to update the `lastCommittedValue`, which is what happens when leveraging the setter liker this. _Maybe_ it would be worth it to do `this.inputElement.value = this.validateInput(value);` and then dispatching the `input` event on `this.inputElement` to tie into that `input` workflow? 🤔

mizgaionutalexandru: I think that works as well, although I see that the `requestUpdate` part of the value setter is not used when the component is by itself but only when it gets set from the outside (for example in sp-slider), if we go down this imperative path. This happens because `this._value` gets computed in `stepBy` and then `setValue` gets called in which the if guard will always return. I wonder if this is intended/best.
I think for this specific issue this imperative approach might be better but i truly wonder if long term the value setter and the `lastCommitedValue` could be slightly changed to have their meaning a little bit clearer.

Westbrook (commented):
This looks a lot nicer! Do you think we could get a test added to ensure we don't drop this again in the future?

packages/number-field/test/number-field.test.ts
```diff
@@ -73,10 +73,14 @@ describe('NumberField', () => {
     describe('receives input', () => {
         it('without language context', async () => {
             const el = await getElFrom(Default({ value: 1337 }));
+            const inputElement: HTMLInputElement =
```
Westbrook, line 76: I _think_ this is the same as `el.focusElement`. If so, would you mind swapping that?

Westbrook (approved):
LGTM! I'll catch this up to `main` and if CI still likes what you've done here, we'll merge it in. Thanks for contributing this fix 🙇🏼

PR #3775, opened 2023-11-06
perf(checkbox): refactor architecture for more rendering perf and DOM element count

Westbrook:
## Description
Refactor the render method and inheritance order of the Checkbox element to reduce render and initialization costs.

## How has this been tested?
-   [ ] _Test case 1_
    1. The tests pass

## Types of changes
-   [x] Performance refactor

## Checklist

### DIFF
--- a/packages/checkbox/src/Checkbox.ts
+++ b/packages/checkbox/src/Checkbox.ts
@@ -16,10 +16,12 @@ import {
     html,
     PropertyValues,
     SizedMixin,
+    SpectrumElement,
     TemplateResult,
 } from '@spectrum-web-components/base';
 import { property } from '@spectrum-web-components/base/src/decorators.js';
-import { CheckboxBase } from './CheckboxBase.js';
+import { CheckboxMixin } from './CheckboxMixin.js';
+import checkboxStyles from './checkbox.css.js';
 import '@spectrum-web-components/icons-ui/icons/sp-icon-checkmark75.js';
 import '@spectrum-web-components/icons-ui/icons/sp-icon-checkmark100.js';
 import '@spectrum-web-components/icons-ui/icons/sp-icon-checkmark200.js';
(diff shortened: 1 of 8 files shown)


Westbrook:
Surfacing initial perf scores:

*Chrome*
image

*Firefox*
image

packages/checkbox/test/benchmark/test-basic.ts
```diff
@@ -15,5 +15,7 @@ import { html } from 'lit';
 import { measureFixtureCreation } from '../../../../test/benchmark/helpers.js';
 
 measureFixtureCreation(html`
-    <sp-checkbox id="checkbox0" tabindex="5">Component</sp-checkbox>
+    <sp-checkbox id="checkbox1">Component</sp-checkbox>
+    <sp-checkbox id="checkbox2" checked>Component</sp-checkbox>
+    <sp-checkbox id="checkbox3" indeterminate>Component</sp-checkbox>
```
lazd, line 20: Should this change be merged first in order to have a baseline that matches? If the benchmark has been changed in this branch only, won't the results be off? Perhaps it's even faster than the current reported results!

lazd (commented):
Definitely looks like a positive change! Where are the gains in DOM element count? I see the icon templates aren't created until they're needed, but it seems like overall DOM structure remains the same.

Also, have you considered removing `Focusable`? The `<input>` itself should be nicely focusable... I know that will kind of break the contract if people call `instance.focus()` or try to set `tabindex`, but perhaps those can be proxied to the input as well...

packages/checkbox/test/benchmark/test-basic.ts
Westbrook: Baseline is run against the previously release version, every time.

packages/checkbox/src/Checkbox.ts
```diff
@@ -28,72 +30,98 @@ import '@spectrum-web-components/icons-ui/icons/sp-icon-dash75.js';
...
+    },
 };
 
 /**
  * @element sp-checkbox
  * @slot - content to display as the label for the Checkbox
  * @fires change - Announces a change in the `checked` property of a Checkbox
  */
-export class Checkbox extends SizedMixin(CheckboxBase, {
+export class Checkbox extends SizedMixin(CheckboxMixin(SpectrumElement), {
```
Westbrook, line 111: @lazd this removes `focusable`, but only for Checkbox.

packages/checkbox/src/Checkbox.ts
```diff
@@ -116,24 +164,41 @@ export class Checkbox extends SizedMixin(CheckboxBase, {
         return html`
             ${super.render()}
             <span id="box">
-                ${checkmarkIcon[this.size as DefaultElementSize]}
-                ${dashIcon[this.size as DefaultElementSize]}
+                ${this.checked
```
Westbrook, line 167: This is the DOM count win.

Rather than _always_ rendering both the checkmark AND dash icon, they'll now be left off the DOM _unless_ the element is `checked` or `indeterminate`.

A much more contained change than what you'd been looking at, but something that stays within the large "Spectrum" family until we're able to get the CSS techniques updated.

packages/checkbox/test/benchmark/test-basic.ts
Westbrook: We collapse them by default, but you can find the most recent run in [this comment](https://github.com/adobe/spectrum-web-components/pull/3775#issuecomment-1795008308).

packages/checkbox/src/Checkbox.ts
lazd: Huh, but `CheckboxMixin` still extends `Focusable`, right? So it's not removed?

packages/checkbox/test/benchmark/test-basic.ts
lazd: I see, so it's running this `test-basic.ts` against release `<sp-checkbox>`, then running it against the branch `<sp-checkbox>`, so there's no need update benchmarks in main for a fair comparison. Got it!

packages/checkbox/src/Checkbox.ts
lazd: Got it, that makes sense!

najikahalsema (approved):
Beautiful. Nice work.

Issue #3773, opened 2023-11-06
Update lit to v3


damooo:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

All

### Description of the requested feature

Lit is updated to v3. There are not many breaking changes. It will be great if swc update the dependency too. Otherwise bundler seems including both versions, and types are incompatible when used along with new lit versions, and other ecosystem components.

Westbrook:
We are tracking this in #3209, but it will likely still be a little while yet before we are able to release fully `lit@3.0` versions of the library. Depending on install order and package manager configurations, there _should_ be no issue in using our components in a `lit@3.0` application. You _will_ need to have both versions in your build, but in terms of client side libraries is quite a small price to pay. I can say that you _won't_ be able to force the next version (via `resolutions` or `overrides`) without issue today, but maybe there's an interim step that we could do to unlock that in the time being.

damooo:
> I can say that you won't be able to force the next version (via resolutions or overrides) without issue today, but maybe there's an interim step that we could do to unlock that in the time being.

As update being already working upon, it is not an issue to include both versions now, and also no particular other reason to override. swc seems re-exporting lit functionalities.  I would be importing them, as we are anyway using `SpectrumElement` as baseclass.

Westbrook:
It shouldn't be an issue to include both versions.

In face, if you're always importing through SWC then you'll always get v2, which means you wouldn't really be using two versions!

PR #3209, opened 2023-05-17
chore: use lit@3.0

Westbrook:
## Description
Lit@3.0 is coming. It's focus is on removing support for out-of-date browsers, and cleaning up deprecations internal to their APIs, with the intent of including the smallest possible breaking changes.

This clarifies any breaking/required changes for the library to get updated to it.

## How has this been tested?
-   [ ] _Test case 1_
    1. All tests pass
-   [ ] _Test case 2_
    1. Storybook/the docs site are usable

## Types of changes
-   [x] Chore (minor updates related to the tooling or maintenance of the repository, does not impact compiled assets)

## Checklist
-   [x] I have signed the [Adobe Open Source CLA](http://opensource.adobe.com/cla.html).
-   [x] My code follows the code style of this project.
-   [ ] If my change required a change to the documentation, I have updated the documentation in this pull request.
-   [x] I have read the **[CONTRIBUTING](<(https://github.com/adobe/spectrum-web-components/blob/main/CONTRIBUTING.md)>)** document.
-   [ ] I have added tests to cover my changes.
-   [x] All new and existing tests passed.

PR #3776, opened 2023-11-06
refactor: prepare for lit@3.0 usage

Westbrook:
## Description
Fast forward changes needed to be compliant with `lit@3.0` so that consumers could choose to force that dependency as needed. This also opens the door for an interstitial step wherein we depend on `^2 || ^3` if needed by consuming projects for a period of time.

## Related issue(s)
- refs #3773
- refs #3209

## How has this been tested?
-   [ ] _Test case 1_
    1. Tests pass

## Types of changes
-   [x] Refactor

## Checklist

### DIFF
--- a/packages/overlay/test/index.ts
+++ b/packages/overlay/test/index.ts
@@ -628,7 +628,7 @@ export const runOverlayTriggerTests = (type: string): void => {
                 expect(await isOnTopLayer(this.hoverContent)).to.be.false;
 
                 const rect = this.outerTrigger.getBoundingClientRect();
-                const close = oneEvent(this.outerTrigger, 'sp-closed');
+                const open = oneEvent(this.outerTrigger, 'sp-opened');
                 await sendMouse({
                     steps: [
                         {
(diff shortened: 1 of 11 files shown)


packages/overlay/test/index.ts
najikahalsema, line 1: Is this just an opportunistic testing refactor, or would it be affected by a 3.0 upgrade? (Beyond the render time being faster, which is why i'm assuming we wouldn't need so many `nextFrame()`s here?)

packages/sidenav/test/sidenav.test.ts
```diff
@@ -393,6 +394,11 @@ describe('Sidenav', () => {
     it('focuses the child anchor not the root when [tabindex=-1]', async () => {
         const el = await fixture<SideNav>(manageTabIndex());
 
+        await nextFrame();
+        await nextFrame();
+        await nextFrame();
+        await nextFrame();
```
najikahalsema, line 400: Why all the nextFrames here?

najikahalsema (approved):
These look like good refactors. I have some questions but none that would prevent merging this. Good work.

packages/overlay/test/index.ts
Westbrook: On the lest we assume that `close` will happen, but without confirming `open` first it was having trouble passing.

packages/sidenav/test/sidenav.test.ts
Westbrook: This is working to ensure that all child elements have `el.updateComplete`'ed. We would likely benefit from an advanced version of `fixture` or other helper that ensured whole tree of elements has updated before running tests.

Issue #3531, opened 2023-08-09
[Feat]: expand Action Group management to include Action Menu children in Action Bar


l3vi-7:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Action Bar

### Description of the requested feature

Action Menu is not currently an accepted child of Action Bar. Expanding the Action Group management that is internally leveraged by Action Bar to include Action Menu would enable clients to use Action Menu with slot "buttons" and may also help resolve the Menu overlay overlap issue. (Please refer the attached image)
![image (6)](https://github.com/adobe/spectrum-web-components/assets/42098011/4df356d8-116a-4f38-b715-66fcd0945a34)

Westbrook:
In lieu of #1719 I'm a bit remiss to say it, but likely the right path here is to update `protected _buttonSelector = 'sp-action-button';` to `protected _buttonSelector = 'sp-action-button, sp-action-menu';` and then add some new tests leveraging `sp-action-menu`s or some `sp-action-menu`s to the existing tests to ensure appropriate coverage.

_Some_ of the functionality here won't be available until #3530 is complete, so it's probably a good idea to start there.

@Rajdeepc, could you look at who on the team might be able to pick this up?

najikahalsema:
> Good morning team ... I have a few questions about the action group [pr](https://github.com/adobe/spectrum-web-components/pull/3780). With the latest comment from Westbrook (ain't he supposed to be off the grid for a few weeks ) it looks like he wants to extend Action Group to support Action Menu as children too. Currently action group is written in a way to support action buttons only and if we want to extend it for action m**enu I'd need more clarification on certain implementation details** and apart from this, a major doubt that I have is with the consumption of Action Menu inside of the Action Bar - If we have a Action Menu inside of an Action Group in the Action Bar that we cannot tab into the Action Menu (cz Action Menu and Action button both are a part of a group).  and leads to the following error 
![image](https://github.com/adobe/spectrum-web-components/assets/44980010/cbbbc6a2-2a41-41c4-88cc-1caaba943ca2)


Hey Tarun. I would agree that in order to make this a complete fix, we need to think about including the action menu as part of the action group. What clarifications are you looking for on the implementation details?

The other question I have is what you mean by "If we have a Action Menu inside of an Action Group in the Action Bar that we cannot tab into the Action Menu (cz Action Menu and Action button both are a part of a group)"? Where is that error coming from? Is it an accessibility error or something else?

[...]

PR #3780, opened 2023-11-07
feat(action-bar): support for action-menus

TarunAdobe:
## Description

Action Menu is not currently an accepted child of Action Bar. Expanding the Action Group management that is internally leveraged by Action Bar to include Action Menu would enable clients to use Action Menu with slot "buttons" and may also help resolve the Menu overlay overlap issue. 

## Related issue(s)

- #3531 

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

image

## Types of changes

## Checklist

## Best practices


najikahalsema (changes requested):
Any clue why [these VRTs](https://5bee72719967d036ba9fee3f8e41ba6a--spectrum-web-components.netlify.app/review/#OverlayElementStories/actionGroupWithFilters.png) are changing? That seems unexpected to me. I'm fine with the background changing since I believe that is supposed to match the other action buttons in the action group, but I'm curious about the corners being unrounded. Here's a [link](https://feat-action-menu-in-action-bar--spectrum-web-components.netlify.app/storybook/?path=/story/overlay-element--action-group-with-filters) to the story in your branch and a [link](https://opensource.adobe.com/spectrum-web-components/storybook/?path=/story/overlay-element--action-group-with-filters) to what's live now.

Also, in order to make sure the addition of the action menu to the ActionBar stories is captured by the VRTs, can you add one extra story that's called "With Action Menu" or something, and have that story start with the ActionMenu in an `open` state?  

Please investigate why the tests are failing. Looks like there's some accessibility issues that need addressing. I think it'd be good to add some tests. Generally, we want to add tests for every new feature we add to a component. A place to start... I did notice that the tab order is broken in your PR in the [stories I linked previously,](https://feat-action-menu-in-action-bar--spectrum-web-components.netlify.app/storybook/?path=/story/overlay-element--action-group-with-filters) because I can't tab into the action group before the action menu. It also doesn't work as expected in the [ActionBar's default story](https://feat-action-menu-in-action-bar--spectrum-web-components.netlify.app/storybook/?path=/story/action-bar--default). Some stories regarding this tab index behaviour are probably a good thing to cover in the test file.

Westbrook (commented):
NOTE: for this to be accurately supported in Action Bar, it needs to be accurately supported in Action Groups. Not only does it need to accept the `size`, `static`, etc. attribute from the Action Group, but it should play as expected into the keyboard interaction model of the group as well.

Westbrook:
In https://feat-action-menu-in-action-bar--spectrum-web-components.netlify.app/storybook/?path=/story/action-bar--default when I navigate to the Action Menu with the keyboard, I am unable to access any of the Menu Items once it is opened.

nikkimk (changes requested):
Action Menu is not focusable via keyboard within action menu:

1. Go to https://feat-action-menu-in-action-bar--spectrum-web-components.netlify.app/storybook/?path=/story/action-bar--fixed
2. Navigate to the action bar via keyboard, so that the **Close** button has focus.
3. Press <kbd>Tab</kbd>, so that the **Edit** button has focus.
3. Press <kbd>Tab</kbd> again, and the **More** action menu should have focus, but instead focus is moved outside of the action bar to the System Picker after it.

TarunAdobe:
@nikkimk This is an expected behaviour as we've placed the edit button and the action-menu in the action-group so you should be able to use arrow keys to navigate between the two items inside of the action-group but using tab will jump to the next focusable element which in this case is the picker.

packages/menu/src/Menu.ts
```diff
@@ -587,22 +587,21 @@ export class Menu extends SizedMixin(SpectrumElement, { noDefaultSize: true }) {
 
     protected navigateBetweenRelatedMenus(event: KeyboardEvent): void {
         const { key } = event;
+        event.stopPropagation();
```
blunteshwar, line 590: Why did we removed `event.stopPropagation();` from line 597 and 605 and added it here?

TarunAdobe: because now we want to make sure that the keyboardEvents don't surface upto the rovingtabIndexController once the menu is open otherwise it leads to an in-accessible state of action-group.

Rajdeepc (approved):

TarunAdobe:
@Rajdeepc  I have addressed all the comments on the PR. Please take a look.

Rajdeepc (approved):

blunteshwar (approved):
LGTM❗

nikkimk:
LGTM

PR #3790, opened 2023-11-09
feat(action-menu): support for emphasized variant in action menu

TrojanRocks:
Emphasized attribute support added to action Menu

## Description
Action Menu internally uses sp-action-button, this component supports emphasized attribute but is not being passed the same, so added the support of emphasized attribute

## Motivation and context

Some Design changes required support of emphasized variant in actionButton

Screenshot 2023-11-09 at 9 28 45 AM

## Checklist

## Best practices


packages/action-menu/src/ActionMenu.ts
```diff
@@ -51,6 +51,9 @@ export class ActionMenu extends ObserveSlotPresence(
     @property({ type: String, reflect: true })
     public static: 'white' | 'black' | undefined = undefined;
 
+    @property({ type: Boolean, reflect: true })
```
Westbrook, line 54: This does not need to `reflect` as it has no effect of the visual delivery of the `<sp-action-menu>` itself.

Westbrook (commented):
This change makes sense, but be aware that this attribute won't make this deliver with the UI in you screenshot. That leverages an `<sp-button>` instead.

If you'd OK with that difference, then this is looking petty good and we'd just need some stories added to represent this change so that the delivery of element with this attribute are included in our visual regression testing processes.

Westbrook:
> If you'd OK with that difference, then this is looking petty good and we'd just need some stories added to represent this change so that the delivery of element with this attribute are included in our visual regression testing processes.

Westbrook:
@TrojanRocks would you be available to catch this up to `main` and add a story to support this update at VRT time?

Westbrook:
Also, when the story and change works as you'd like it to, be sure to checkout https://github.com/adobe/spectrum-web-components#keeping-ci-assets-updated to manage the VRT cache based on this change.

PR #3793, opened 2023-11-10
docs: fix button.ts typo

ichengde:


Westbrook:
This will cause the slot to be documented as having the name “text” which is incorrect.

ichengde:
I thought it was a mistake, but now I get it. Thank you

Issue #3383, opened 2023-06-28
[Bug][a11y]: Update resizable examples in documentation using the label prop available in the api


dineshy87:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Split-view

### Expected behavior

Not a big deal, but I think it's better to have the resizable examples in [documentation](https://opensource.adobe.com/spectrum-web-components/components/split-view/) use label prop. It also helps the devs using it understand that the separator is generally labelled after the primary pane. 

[ "The splitter also has an accessible name that matches the name of the primary pane."](https://www.w3.org/WAI/ARIA/apg/patterns/windowsplitter/#aboutthispattern)

### Actual behavior

None of the examples are labelled.

castastrophe:
@Rajdeepc @dineshy87 Would this be a case where [aria-labelledby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby) might be more appropriate?

> separator is generally labelled after the primary pane.

Since the separator is labelled for the primary pane, we add an ID to it and point our splitter to that id:

```html
<sp-split-view resizable aria-labelledby="left-panel">
    <div>
        <h1 id="left-panel">Left panel</h1>
        <p>
            Lorem Ipsum is simply dummy text of the printing and typesetting
            industry.
        </p>
    </div>
    <div>
        <h2>Right panel</h2>
        <p>
            It is a long established fact that a reader will be distracted by
            the readable content of a page when looking at its layout.
        </p>
    </div>
</sp-split-view>
```

castastrophe:
I'll add we could also get a little smart with this if the split view has any awareness of which panel is _more_ visible. If one exceeds 60% of the available space for example, we could swap the labelledby to point to the more dominant panel. Just an idea. I'm not sure how helpful that would be necessarily to a screen-reader experience without testing it in action.

Rajdeepc:
Definitely this sounds intuitive but we need to test it out in action to get a broader perspective of how it will read out in screen reader.

PR #3800, opened 2023-11-14
docs(split-view): added label for resizable examples for documentation

Rajdeepc:
## Description

Added label property for resizable elements of split-view

## Related issue(s)

- https://github.com/adobe/spectrum-web-components/issues/3383

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/split-view/README.md
+++ b/packages/split-view/README.md
@@ -38,7 +38,13 @@ import { SplitView } from '@spectrum-web-components/split-view';
 ### Horizontal Resizable
 
 ```html
-<sp-split-view resizable primary-min="50" secondary-min="50" primary-size="100">
+<sp-split-view
+    resizable
+    primary-min="50"
+    secondary-min="50"
+    primary-size="100"
+    label
+>
     <div>
         <h1>Left panel</h1>
         <p>
@@ -59,7 +65,7 @@ import { SplitView } from '@spectrum-web-components/split-view';
 ### Horizontal Resizable & Collapsible
 
 ```html
-<sp-split-view resizable>
+<sp-split-view resizable label>
     <div>
         <h1>Left panel</h1>
         <p>
@@ -95,6 +101,7 @@ import { SplitView } from '@spectrum-web-components/split-view';
     primary-min="50"
     primary-max="150"
     secondary-min="50"
+    label
 >
     <div>
         <h1>Top panel</h1>
@@ -116,7 +123,7 @@ import { SplitView } from '@spectrum-web-components/split-view';
 ### Vertical Resizable & Collapsible
 
 ```html
-<sp-split-view vertical resizable style="height: 300px;">
+<sp-split-view vertical resizable style="height: 300px;" label>
     <div>
         <h1>Top panel</h1>
         <p>


castastrophe:
@Rajdeepc I added an idea to the original issue if you wanted to weigh in there!

najikahalsema (changes requested):
Could you actually give a value to the `label`s here?

najikahalsema (approved):
Oh, now that I see how the label works in the code, I think this is fine for  now because we're addressing the issue. In the future, I wonder if it would be worth revisiting this attribute because, to me, `label` implies you can customise it. Setting a label like this isn't good for internationalisation, either, so I'll make a new issue for this.

najikahalsema:
I think that the way the property is set though would allow for a custom label, though, right? Because the logic is 

```
        const label =
            this.label || this.resizable ? 'Resize the panels' : undefined;
```

And the consumer can set `this.label` ... so are the only values 'Resize the panels' or `undefined`? What happens if you set `this.label` to some string value?

najikahalsema:
Sorry, I accidentally closed the PR when trying to make this comment.

I think that the way the property is set though would allow for a custom label, though, right? Because the logic is 

```
        const label =
            this.label || this.resizable ? 'Resize the panels' : undefined;
```

And the consumer can set `this.label` ... so are the only values 'Resize the panels' or `undefined`? What happens if you set `this.label` to some string value?

Rajdeepc:
> Sorry, I accidentally closed the PR when trying to make this comment.
> 
> I think that the way the property is set though would allow for a custom label, though, right? Because the logic is
> 
> ```
>         const label =
>             this.label || this.resizable ? 'Resize the panels' : undefined;
> ```
> 
> And the consumer can set `this.label` ... so are the only values 'Resize the panels' or `undefined`? What happens if you set `this.label` to some string value?

If we set this.label to some other string values. `aria-label` value still remains as 'Resize the panels`
```
<sp-split-view label="Some other string">
#shadow-root
   <div id="splitter" aria-label="Resize the panels"></div>
</sp-split-view>
```

najikahalsema:
Okay. I gotcha. I guess we can merge like how it is now and make a follow up issue.

najikahalsema (dismissed):

Westbrook (changes requested):
`label` without a value is a non-sensical addition here. We should include a value in this change.

We should address the issue that was created herein, and in particular the reality that I wrote the logic to apply the `aria-label` value incorrectly. 🙈 

```
const label =
            this.label || this.resizable ? 'Resize the panels' : undefined;
```

Should likely be
```
const label =
            this.label || (this.resizable ? 'Resize the panels' : undefined);
```

Or possibly:

```
const label =
            this.resizable ? (this.label || 'Resize the panels') : undefined;
```

So that the `label` attribute actually does something here...

najikahalsema (approved):
Thanks, @Westbrook , for being able to articulate what I could not. ^^; Thanks for fixing this, Rajdeep.

Westbrook (approved):

Issue #3787, opened 2023-11-08
[Tooltip] Expand `self-managed` documentation


Westbrook:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Description of issue

Expand https://opensource.adobe.com/spectrum-web-components/components/tooltip/#self-managed-overlays to include information about a focusable element, as per the Dev Mode warning and the discussion in https://github.com/adobe/spectrum-web-components/discussions/3781#discussioncomment-7504620

PR #3815, opened 2023-11-20
chore(tooltip): extend tooltip self-managed documentation

TarunAdobe:
Expand https://opensource.adobe.com/spectrum-web-components/components/tooltip/#self-managed-overlays to include information about a focusable element, as per the Dev Mode warning and the discussion in https://github.com/adobe/spectrum-web-components/discussions/3781#discussioncomment-7504620

## Description

Extended `self-managed` tooltip documentation to include an example of how to use tooltip for non-interactive elements by wrapping them with an interactive element and added notes about why the current pattern is supported the way it is.

## Related issue(s)

- #3787

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/tooltip/README.md
+++ b/packages/tooltip/README.md
@@ -46,8 +46,7 @@ Tooltips can be top, bottom, left, or right.
 By default, Tooltip provides styling without behavior.
 You must combine it with an [Overlay Trigger](https://opensource.adobe.com/spectrum-web-components/components/overlay-trigger/#%22hover%22-content-only) in order to manage its overlay behavior.
 
-You can instead apply the `self-managed` attribute for this common case,
-which automaticaly binds to the parent element's hover interaction:
+You can use the tooltip as the descendant of an interactive element, such as [Action Button](https://opensource.adobe.com/spectrum-web-components/components/action-button/), by applying the `self-managed` attribute which automatically binds the Tooltip to its first interactive ancestor element's focus/hover interactions:
 
 ```html
 <sp-action-button>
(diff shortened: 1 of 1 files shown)


packages/tooltip/README.md
```diff
@@ -46,8 +46,7 @@ Tooltips can be top, bottom, left, or right.
 By default, Tooltip provides styling without behavior.
 You must combine it with an [Overlay Trigger](https://opensource.adobe.com/spectrum-web-components/components/overlay-trigger/#%22hover%22-content-only) in order to manage its overlay behavior.
 
-You can instead apply the `self-managed` attribute for this common case,
-which automaticaly binds to the parent element's hover interaction:
+You can use the tooltip with an interactive element, such as [Action Button](https://opensource.adobe.com/spectrum-web-components/components/action-button/), and apply the `self-managed` attribute which automaticaly binds to the parent element's hover interaction:
```
Westbrook, line 49:
```suggestion
You can use the tooltip as the descendant of an interactive element, such as [Action Button](https://opensource.adobe.com/spectrum-web-components/components/action-button/), by applying the `self-managed` attribute which automatically binds the Tooltip to its first interactive ancestor element's focus/hover interactions:
```

Westbrook (commented):
We should probably define “interactive element” inline with what the code is looking for here, there’s a long CSS selector to base this off of for consumers.

Westbrook (approved):
Super useful addition. Thanks for editing through this! :shipit:

PR #3840, opened 2023-11-30
fix(tabs): prevent vertical scrolling in overflow tabs

### LINKED ISSUE #3813: [Bug]: Overflow tabs on mobile are draggable
### Code of conduct

### Impacted component(s)

sp-tabs-overflow

### Expected behavior

The user should be able to only swipe left or right to select a tab. 

### Actual behavior

The user can drag the whole component inside its container.

### Screenshots

https://github.com/adobe/spectrum-web-components/assets/56998711/20ad8eb7-d220-419a-8621-fad3b389e249

_No response_

Repro on iOS device iPhone 14. iOS version 17.0.2
Go to https://opensource.adobe.com/spectrum-web-components/components/tabs-overflow/ 
Tap on one tab.
Hold and drag on tabs .

_No response_

_No response_

Westbrook:
## Description
Use a more complete value for `overflow`, `auto hidden`, to prevent overflow scrolling in touch environments.

I'm not sure how one might do a regression test on this sort of thing, but maybe we don't need to? 

## Related issue(s)
- fixes #3813

## How has this been tested?
-   [ ] _Test case 1_
    1. Go [here](https://opensource.adobe.com/spectrum-web-components/components/tabs-overflow/#example)
    2. In an actual iOS device, maybe an emulator would be enough, but I didn't try
    3. See that the Tabs will scroll up/down as well as left/right, even if they always go back to root
-   [ ] _Test case 2_
    1. Go [here](https://overflow-tabs-scrolling--spectrum-web-components.netlify.app/components/tabs-overflow/#example)
    2. In the same device
    3. See that the Tabs will NO LONGER scroll up/down

## Types of changes

## Checklist

### DIFF
--- a/packages/tabs/src/tabs.css
+++ b/packages/tabs/src/tabs.css
@@ -40,6 +40,10 @@ governing permissions and limitations under the License.
     justify-content: var(--swc-tabs-list-justify-content);
 }
 
+:host([disabled]) #list {
+    pointer-events: none;
+}
+
 :host([disabled]) #list #selection-indicator {
     background-color: var(
         --mod-tabs-color-disabled,
@@ -49,10 +53,6 @@ governing permissions and limitations under the License.
 
 :host([disabled]) ::slotted(sp-tab) {
     color: var(--mod-tabs-color-disabled, var(--spectrum-tabs-color-disabled));
-}
-
-:host([disabled]) #list {
-    pointer-events: none;
 }
 
 /*
@@ -70,8 +70,9 @@ governing permissions and limitations under the License.
 }
 
 :host([dir][direction='horizontal']) #list.scroll {
-    overflow-x: auto;
     scrollbar-width: none;
+    overflow-x: auto;
+    overflow-y: hidden;
 }
 
 :host([dir][direction='horizontal']) #list.scroll::-webkit-scrollbar {


najikahalsema (commented):
Your link for the second test is broken 😉 But this doesn't seem to fix the problem for me. I'm on iOS 16.6 on iPhone 14, Chrome ver 119.0.6045 :( I will update to iOS. 17.1.2 and let you know if there's a difference.

https://github.com/adobe/spectrum-web-components/assets/44980010/5a82269d-f2d5-493c-9c44-2c8b96108863

Westbrook:
Both of those links are bad...at least in relation to the work addressed here. I've updated both.

Here's I've focused specifically on when the `sp-tabs-overflow` element is leveraged. Do you think prevents the off axis scrolling on ALL implementations would be acceptable? I'm not seeing a reason why not, but also not sure what a team would do otherwise if they expected scrolling in this case.

najikahalsema:
Okay, great, it works! I think it's worth doing while we're here, and if someone complains about it, we can... revert it? I don't see why we wouldn't want it prevented in all cases.

najikahalsema (approved):
That's fine with me. Approved!

PR #3845, opened 2023-12-02
fix(overlay): ensure events are only bound once

### LINKED ISSUE #3795: [Bug]: Overlay - triggerElement might be null in doPointerleave
### Code of conduct

### Impacted component(s)

Overlay

### Expected behavior

No error should be thrown when pointer leaves trigger element

### Actual behavior

Sometimes it happens, that the triggerElement is null, when function `doPointerleave` is called, leading to an exception when trying to exec 
```
if (this.focusedin && triggerElement.matches(":focus-visible"))
      return;
```
It happens with triggers for both tooltip and overlay in some (more complex) use cases.
It might be fixable by using conditional operator in `triggerElement?.matches(":focus-visible")` 

Exception thrown:
```
TypeError: Cannot read properties of null (reading 'matches')
    at _Overlay2.doPointerleave (Overlay.ts:851:46)
    at Button.handlePointerleave (Overlay.ts:841:14)
    ...
 ```

### Screenshots

_No response_

Chrome

3. Scroll to '....'
4. Check console
5. See error

_No response_

_No response_

Westbrook:
## Description
- clean up event bindings before binding.
- merge multiple possible calls to `manageTriggerElement()` into a single call

## Related issue(s)
- fixes #3795

## How has this been tested?
-   [ ] _Test case 1_
    1. Go [here](https://do-pointer-leave--spectrum-web-components.netlify.app/storybook/?path=/story/overlay-element--transient-hover)
    2. Trigger the tooltip on the button
    3. Click the button to trigger the popover
    4. Move the mouse out of the button
    5. See that no error in thrown in the console

## Types of changes

## Checklist

### DIFF
--- a/packages/overlay/src/Overlay.ts
+++ b/packages/overlay/src/Overlay.ts
@@ -504,6 +504,8 @@ export class Overlay extends OverlayFeatures {
 
     protected bindEvents(): void {
         if (!this.hasNonVirtualTrigger) return;
+        // Clean up listeners if they've already been bound
+        this.abortController?.abort();
         this.abortController = new AbortController();
         const nextTriggerElement = this.triggerElement as HTMLElement;
         switch (this.triggerInteraction) {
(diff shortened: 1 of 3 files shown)


spdev3000 (approved):
lgtm

PR #3846, opened 2023-12-04
fix(number-field): validate min/max in more contexts

### LINKED ISSUE #3827: [Bug]: Number Field min-max limits not correctly working
### Code of conduct

### Impacted component(s)

Number Field

### Expected behavior

When using Number fields with limit (min/max) it should act like:
- using keyboard arrow keys to modify values should stop at limit
- using mouse wheel to modify values it should stop at limit
- using stepper buttons to modify values it should stop at limit

### Actual behavior

Following wrong behavior can be experienced:
1. Defining a max value:
- if using a max value and click into Number Field a user can successfully step up (using `cursor up` key or mouse wheel) to `max + 1` value as display value, it will be corrected, once it looses the focus
- if using a max value and click into Number Field a user can successfully step up (using SHIFT modifier and `cursor up` key or mouse wheel) to `max + 10` value, it will be corrected, once it looses the focus

2. Defining a min value:
- if using a min value and click into Number Field a user can successfully step down (using `cursor down` key or mouse wheel) to `min - 1` value, it will be corrected, once it looses the focus
- if using a min value and click into Number Field a user can successfully step down (using SHIFT modifier and `cursor down` key or mouse wheel) to `min + 10` value, it will be corrected, once it looses the focus
- if using a min value of `0` - there seems no real min limit set and user can step down to `-1 * max` 
- if using a min value of `0` - user can use the stepper to step to `0` - button seems disabled, but clicking it again re-enables the button and steps further down to `-1 * max` , this won't be corrected

### Screenshots

Screenshot 2023-11-24 at 17 45 16

_No response_

1. Go to https://studio.webcomponents.dev/edit/JAP5C1EoJO0WvZGrZx4I/src/index.ts?p=stories
2. Try out several use cases with stepper, cursor keys, mouse wheel and SHIFT modifier.

_No response_

_No response_

Westbrook:
## Description
Validate limits more regularly to prevent slippage and ensure the value stays within them.

## Related issue(s)
- fixes #3827

## How has this been tested?
-   [ ] _Test case 1_
    1. Go [here](https://number-field-limits--spectrum-web-components.netlify.app/storybook/?path=/story/number-field--min-max)
    2. Use the `ArrowDown` key until the Number Field hits `0`, the min value
    3. See that the value does not go below that value
-   [ ] _Test case 2_
    1. Go [here](https://number-field-limits--spectrum-web-components.netlify.app/storybook/?path=/story/number-field--min-max)
    2. Set the value to 254
    3. use the `ArrowUp` key until the Number Field hits `255`, the max value
    4. See that the value does not go above that

See other use cases in #3827

## Types of changes

## Checklist

### DIFF
--- a/packages/number-field/src/NumberField.ts
+++ b/packages/number-field/src/NumberField.ts
@@ -319,7 +319,9 @@ export class NumberField extends TextfieldBase {
         if (isNaN(this.value)) {
             value = min;
         }
+        value = this.valueWithLimits(value);
 
+        this.requestUpdate();
         this._value = this.validateInput(value);
         this.inputElement.value = value.toString();
 
@@ -474,15 +476,22 @@ export class NumberField extends TextfieldBase {
         this.inputElement.setSelectionRange(nextSelectStart, nextSelectStart);
     }
 
-    private validateInput(value: number): number {
-        const signMultiplier = value < 0 ? -1 : 1; // 'signMultiplier' adjusts 'value' for 'validateInput' and reverts it before returning.
-        value *= signMultiplier;
+    private valueWithLimits(nextValue: number): number {
+        let value = nextValue;
         if (typeof this.min !== 'undefined') {
             value = Math.max(this.min, value);
         }
         if (typeof this.max !== 'undefined') {
             value = Math.min(this.max, value);
         }
+        return value;
+    }
+
+    private validateInput(value: number): number {
+        value = this.valueWithLimits(value);
+        const signMultiplier = value < 0 ? -1 : 1; // 'signMultiplier' adjusts 'value' for 'validateInput' and reverts it before returning.
+        value *= signMultiplier;
+
         // Step shouldn't validate when 0...
         if (this.step) {
             const min = typeof this.min !== 'undefined' ? this.min : 0;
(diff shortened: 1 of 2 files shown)


spdev3000:
When I enter a negative value into a number field with min="0" and focus leaves the input (blur) or "Enter" is pressed (while focus is still in the number field) - the negative number is still displayed (and not set to "o" = min-value)
Screenshot 2023-12-04 at 17 44 10

spdev3000:
Also entering a really large **_negative_** number (such as -200000) for a min="0" and max="255" field and pressing "Enter" sets the value to -255 (-1 * max).
Screenshot 2023-12-04 at 17 47 51

packages/number-field/src/NumberField.ts
```diff
@@ -474,15 +476,22 @@ export class NumberField extends TextfieldBase {
...
             value = Math.max(this.min, value);
         }
         if (typeof this.max !== 'undefined') {
             value = Math.min(this.max, value);
         }
+        return value;
+    }
+
+    private validateInput(value: number): number {
+        value = this.valueWithLimits(value);
```
Westbrook, line 491: Thanks for the deep QE here @spdev3000. It seems this corrects the `-20000` and `-255` input issues you found.

@blunteshwar can you take a look as it seems this stems from your work in https://github.com/adobe/spectrum-web-components/pull/3673/files

blunteshwar: Looks good to me!
It seems that due to Initially multiplying by the `signMultiplier`  `value = Math.max(this.min, value);` was not able to limit the value.

blunteshwar (approved):

Issue #2559, opened 2022-09-01
[Feat]: Add Data Picker Pattern


Westbrook:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Date Picker, Calendar

### Description of the requested feature

Please add the Date Picker pattern

cc: @najikahalsema @Jabark 

### Mockups or screenshots

https://opensource.adobe.com/spectrum-css/datepicker.html

### Implementation notes or ideas

It's possible that date range could be scoped out for an initial release, as it's much more complicated, but unsure what side effects that would have on the API.

Ashish-Dobhal:
Hi team,
is there an update on the latest progress of this task. We are planning to use this component and wanted to know if its in the pipelines before using developing/using something else

najikahalsema:
@Jabark is the contributing developer for this component, so I would check with him. Our team is not actively developing this component.

davidjgonzalez:
@Jabark Any updates on this? I need a date picker in my project.

I was hoping in the interim i could use sp-textfield and set type="date" and let the browser do the heavy lifting but that didn't work -- not sure if you have any tips on how I could make that work.

Westbrook:
@paulodetarsofm has picked up this work and the progress is available in https://github.com/adobe/spectrum-web-components/pull/3464

davidjgonzalez:
Thanks - left a comment on that! 🙏

castastrophe:
We have this work tracked now in Jira so I'm going to close. Please do reach out if you still need a status update on this feature.

PR #3856, opened 2023-12-06
feat(date-time-picker): add date/time input + picker 

mirekszot:
## Description

Adding the third component that will be part of the date/time selector: the date/time field itself. The time field, which had been created in this [PR](https://github.com/adobe/spectrum-web-components/pull/3464), will no longer be used, inside the popover we will only have the calendar, at least for now.

image

## Related issue(s)

## Motivation and context

Allow the user to set the date and time to publish a post on social media.

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices


packages/date-time-picker/package.json
```diff
@@ -1,18 +1,18 @@
 {
-    "name": "@spectrum-web-components/time-field",
+    "name": "@spectrum-web-components/date-time-picker",
```
Westbrook, line 2: Confirming this is meant to rename the other working package, not create a new package in parallel.

packages/date-time-picker/package.json
```diff
@@ -57,7 +57,16 @@
         "lit-html"
     ],
     "dependencies": {
-        "@spectrum-web-components/input-segments": "^0.0.1"
+        "@spectrum-web-components/calendar": "^0.0.1",
+        "@spectrum-web-components/field-label": "^0.40.0",
```
Westbrook, line 61: Versions, here, too. The new packages (`@spectrum-web-components/calendar`, etc.) will manage themselves, so no need to worry about them.

packages/date-time-picker/src/DateTimePicker.ts
```diff
@@ -0,0 +1,158 @@
...
+    state,
+} from '@spectrum-web-components/base/src/decorators.js';
+import { InputSegments } from '@spectrum-web-components/input-segments';
+
+import styles from './date-time-picker.css.js';
+
+import '@spectrum-web-components/calendar/sp-calendar.js';
+import '@spectrum-web-components/field-label/sp-field-label.js';
+import '@spectrum-web-components/icons-workflow/icons/sp-icon-calendar.js';
+import '@spectrum-web-components/overlay/sp-overlay.js';
```
Westbrook, line 28: Please add a note to the to do lists that are tracked around adding lazy loading for some of these dependancies.

packages/date-time-picker/src/DateTimePicker.ts
```diff
@@ -0,0 +1,158 @@
...
+                                @change=${this.handleDate}
+                            ></sp-calendar>
+                        </div>
+                    </sp-popover>
+                </sp-overlay>
+            </div>
+        `;
+    }
+
+    public openPicker(): void {
```
Westbrook, line 107: Can we use `showPicker()` to normalize to the platform API: https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/showPicker

This would imply `hidePicker()` below.

packages/date-time-picker/src/DateTimePicker.ts
```diff
@@ -0,0 +1,158 @@
...
+        this.pickerDate = undefined;
+        this.open = false;
+    }
+
+    /**
+     * Updates the value of the internal property used to define the date used by the calendar with the value received
+     *
+     * @param event - Event with the value emitted by the calendar
+     */
+    public handleDate(event: CustomEvent<Date | undefined>): void {
```
Westbrook, line 125: Can this be `private`?

packages/date-time-picker/src/DateTimePicker.ts
```diff
@@ -0,0 +1,158 @@
...
+        super.yearSegment.value = year;
+        super.formatValue(super.yearSegment);
+
+        super.monthSegment.value = month;
+        super.formatValue(super.monthSegment);
+
+        super.daySegment.value = day;
+        super.formatValue(super.daySegment);
+
+        super.requestUpdate();
```
Westbrook, line 156: Why `super` instead of `this`?

packages/date-time-picker/src/DateTimePicker.ts
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    public openPicker(): void {
+        this.pickerDate = super.getDateFromSegments();
+        this.open = true;
+    }
+
+    public closePicker(): void {
+        super.setNewDateTime();
+        super.emitNewDateTime();
```
Westbrook, line 114: Why `super` instead of `this`?

packages/date-time-picker/src/date-time-picker.css
```diff
@@ -0,0 +1,58 @@
...
+    --datetime-picker-spacing: calc(
+        var(
+                --mod-textfield-spacing-inline,
+                var(--spectrum-textfield-spacing-inline)
+            ) -
+            var(
+                --mod-textfield-border-width,
+                var(--spectrum-textfield-border-width)
+            )
+    );
```
Westbrook, line 22: Looks like this is only used once. Can we apply it inline, rather than creating a new custom property for it?

packages/date-time-picker/src/DateTimePicker.ts
```diff
@@ -0,0 +1,158 @@
...
+                    @click=${this.openPicker}
+                >
+                    <slot name="calendar-icon" slot="icon">
+                        <sp-icon-calendar></sp-icon-calendar>
+                    </slot>
+                </sp-picker-button>
+
+                <!-- TODO: Enable "receives-focus" when calendar is navigable via keyboard -->
+                <sp-overlay
+                    placement="bottom-start"
```
Westbrook, line 88: As per https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-datepicker/ we likely want this to trap the tab order... _sometimes_. Can you add reviewing this to the to do list?

Generally, this is added via `type="modal"` but this _always_ traps the tab order, and we only want to do so _if_ the user enters the picker menu, not when the picker menu opens, which would be the current functionality.

packages/date-time-picker/stories/date-time-picker.stories.ts
```diff
@@ -215,7 +125,10 @@ export const timeGranularity = (args: StoryArgs = {}): TemplateResult => {
         timeGranularity: args.timeGranularity,
     };
 
-    return renderTimeField(`Time Granularity: ${args.timeGranularity}`, args);
+    return renderDateTimePicker(
+        `Time Granularity: ${args.timeGranularity}`,
```
Westbrook, line 129: This story doesn't seem to change based off of the select `timeGraularity` value.

packages/date-time-picker/stories/date-time-picker.stories.ts
```diff
@@ -9,182 +9,92 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
...
+
+    requestAnimationFrame(() => {
+        const container = document.querySelector(
+            `.story-container-${randomId}`
+        );
+
+        if (container) {
+            render(story, container as HTMLElement);
+        }
+    });
```
Westbrook, line 85: Why is this approach being leveraged as opposed to rendering the story directly in the returned `html`? If it is required by something specific to these elements, please add a comment for the reasoning.

packages/date-time-picker/test/benchmark/basic-test.ts
```diff
@@ -9,10 +9,10 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
...
 governing permissions and limitations under the License.
 */
-import '@spectrum-web-components/time-field/sp-time-field.js';
+import '@spectrum-web-components/date-time-picker/sp-date-time-picker.js';
 import { html } from '@spectrum-web-components/base';
 import { measureFixtureCreation } from '../../../../test/benchmark/helpers.js';
 
 measureFixtureCreation(html`
-    <sp-time-field></sp-time-field>
+    <sp-date-time-picker></sp-date-time-picker>
```
Westbrook, line 17: Please configure this picker with some defaults to better exhibit the performance profile of the element.

packages/date-time-picker/test/date-time-picker.test.ts
```diff
@@ -9,25 +9,25 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
...
+            await fixture<DateTimePicker>(
                 html`
-                    <sp-time-field></sp-time-field>
+                    <sp-date-time-picker></sp-date-time-picker>
                 `
             )
     );
-    it('loads default time-field accessibly', async () => {
-        const el = await fixture<TimeField>(
+    it('loads default sp-date-time-picker accessibly', async () => {
```
Westbrook, line 27: If you were able to stub out some tests you might expect in the future, the would be much appreciated.

```js
it('does something that Mirek built', aync () => {});
it('does something else that Mirek built', aync () => {});
```

This way we don't have to think _what_ to test in future passes, just _how_ to test them.

packages/picker-button/src/PickerButton.ts
```diff
@@ -57,6 +60,7 @@ export class PickerButton extends SizedMixin(
             root: true,
             uiicononly: !this.hasText,
             textuiicon: this.hasText,
+            'spectrum-PickerButton--quiet': this.quiet,
```
Westbrook, line 63: If you update https://github.com/adobe/spectrum-web-components/blob/main/packages/picker-button/src/spectrum-config.js#L212-L215 to be `spectrum-PickerButton--quiet` instead of `spectrum-PickerButton--low` we won't need to effect the render method with this change.

packages/picker-button/src/PickerButton.ts
```diff
@@ -45,6 +45,9 @@ export class PickerButton extends SizedMixin(
     @property({ type: Boolean, reflect: true })
     invalid = false;
 
+    @property({ type: Boolean, reflect: true })
```
Westbrook, line 48: Thanks for adding this!

packages/picker-button/stories/picker-button-sizes.stories.ts
```diff
@@ -12,7 +12,8 @@ governing permissions and limitations under the License.
 import { TemplateResult } from '@spectrum-web-components/base';
 
 import { argTypes, StoryArgs, Template } from './index.js';
-import '../sp-picker-button.js';
+
+import '@spectrum-web-components/picker-button/sp-picker-button.js';
```
Westbrook, line 16: 🔥

tools/input-segments/src/InputSegments.ts
```diff
@@ -257,129 +261,218 @@ export class InputSegments extends TextfieldBase {
         };
 
         /**
-         * TODO: Include ARIA attributes for editable segments
-         * TODO: Use `@input`/`@beforeinput` events to handle data input/content cleanup
-         * TODO: Move `handleKeydown()` call to a cache so that it doesn't cycle on the binding in each render pass
```
Westbrook, line 262: This doesn't seem to be addressed here. The issue that is looks to resolve seems to exist in the `beforeinput` and `input` event listeners as well, now.

tools/input-segments/src/InputSegments.ts
```diff
@@ -257,129 +261,218 @@ export class InputSegments extends TextfieldBase {
...
+                @beforeinput=${(event: InputEvent) => {
+                    this.handleBeforeInput(segment, event);
+                }}
+                @input=${(event: InputEvent) => {
+                    this.handleInputEvent(segment, event);
+                }}
+                .innerText=${when(
+                    usePlaceholder,
+                    () => segment.placeholder ?? '',
+                    () => segment.formatted ?? ''
```
Westbrook, line 289: It would be great if we could use stable methods for this as well.

tools/input-segments/src/InputSegments.ts
```diff
@@ -257,129 +261,218 @@ export class InputSegments extends TextfieldBase {
...
+        const data: string | null = event.data;
 
-        if (details === undefined) {
+        if (details === undefined || data === null) {
             return;
         }
 
-        const typedValue = this.numberParser.parse(event.key);
+        if (
+            data === ' ' || // Space is considered “valid number” by the following validations... 🤦🏻‍♂️
```
Westbrook, line 387: 🤣

tools/input-segments/src/InputSegments.ts
```diff
@@ -257,129 +261,218 @@ export class InputSegments extends TextfieldBase {
...
+
+        this.valueChanged(segment, event);
+    }
+
+    /**
+     * Increments the segment value respecting the minimum and maximum limits
+     *
+     * @param segment - Segment on which the event was triggered (the segment being changed)
+     * @param event - Triggered event details
+     */
```
Westbrook, line 411: Many thanks for all of the JSDoc comments!

tools/input-segments/src/InputSegments.ts
```diff
@@ -415,163 +508,416 @@ export class InputSegments extends TextfieldBase {
...
 
-        if (!this.isValidTime(this.selectedDateTime)) {
-            this.selectedDateTime = undefined;
-            return;
+        const segment: Segment = {
+            type,
+            formatted,
+            ...(value !== undefined && { value }),
+            ...(minValue !== undefined && { minValue }),
+            ...(maxValue !== undefined && { maxValue }),
```
Westbrook, line 632: Is it possible to get the same here via:
```suggestion
            ...this.getValueAndLimits(type)
```
As `undefined` destructures nothing into the object?

tools/input-segments/src/InputSegments.ts
```diff
@@ -415,163 +508,416 @@ export class InputSegments extends TextfieldBase {
...
-            year = this.currentDateTime.year;
-            month = this.currentDateTime.month;
-            day = this.currentDateTime.day;
+        /**
+         * If the day being formatted is February 29th but the year segment has not yet been filled, we need to use a
+         * leap year to allow the 29th to remain, otherwise, if we use the current year and it is not a leap year, the
+         * day that would be displayed would be March 1st, as February 29th would not exist and JavaScript “moves” the
+         * day to the next day. As this year is only used to format the day and month, we use the year 2000 as the "base
+         * year" for formatting
+         */
```
Westbrook, line 814: 🤯 How did you catch this!? Great work. Hopefully this is in the tests either in actuality or as a stub.

tools/input-segments/src/input-segments.css
```diff
@@ -50,14 +50,8 @@ governing permissions and limitations under the License.
     text-align: start;
 }
 
-.placeholder {
-    display: block;
-    width: 100%;
-    height: 0;
-    visibility: hidden;
+.editable-segment.is-placeholder {
     font-style: italic;
```
Westbrook, line 54: This needs to be forced as non-italic is specific locales:
```
.is-placeholder:lang(ja),
is-placeholder:lang(ko),
.is-placeholder:lang(zh) {
    font-style: normal;
}
```

tools/input-segments/src/input-segments.css
```diff
@@ -50,14 +50,8 @@ governing permissions and limitations under the License.
     text-align: start;
 }
 
-.placeholder {
-    display: block;
-    width: 100%;
-    height: 0;
-    visibility: hidden;
+.editable-segment.is-placeholder {
```
Westbrook, line 53: Does this need the double specificity, or is `.is-placeholder` enough here?

Westbrook (commented):
Really great work so far here @mirekszot! Lots of notes and small changes, let me know if any need further explanation.

mirekszot:
> Really great work so far here @mirekszot! Lots of notes and small changes, let me know if any need further explanation.

Thanks @Westbrook for reviewing. All kudos should go to @paulodetarsofm, he's the author of this PR 👏 👏.
I will try to implement your comments ASAP!

packages/date-time-picker/stories/date-time-picker.stories.ts
paulodetarsofm: @Westbrook, this solution was suggested by you, I don't remember completely but I think this is necessary because we have more than one example using the same component in the same story, and only the first one was being rendered

packages/date-time-picker/README.md
```diff
@@ -0,0 +1,34 @@
+## Description
+
```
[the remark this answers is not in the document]
paulodetarsofm: As we did with the previous components, the description will be updated later. Documentation update is something we add to the TODO list of all new components

packages/date-time-picker/src/DateTimePicker.ts
paulodetarsofm: What do you mean exactly, @Westbrook?

packages/date-time-picker/src/DateTimePicker.ts
paulodetarsofm: Added

packages/date-time-picker/stories/date-time-picker.stories.ts
Westbrook: Well, past Westbrook seems like a smart guy...could we comment that as future Westbrook is looking to continue the trend of less intelligent Westbrooks. 🙈

tools/input-segments/src/InputSegments.ts
paulodetarsofm: In fact, this `TODO` was not resolved, it only had the text adjusted, as previously we only had `keydown`, but now we also have two new events, `input` and `beforeinput`.

Even so, we will need help with this improvement, because although we have talked about this in the past, I don't remember how we could actually improve this process, using what you called "call to a cache".

tools/input-segments/src/InputSegments.ts
```diff
@@ -257,129 +271,214 @@ export class InputSegments extends TextfieldBase {
...
-                    () => segment.formatted
-                )}
-            </div>
+                @beforeinput=${(event: InputEvent) => {
+                    this.handleBeforeInput(segment, event);
+                }}
+                @input=${(event: InputEvent) => {
+                    this.handleInputEvent(segment, event);
+                }}
+                .innerText=${this.renderSegmentText(segment)}
```
Westbrook, line 296: We'll likely benefit from a comment on this one.

tools/input-segments/src/InputSegments.ts
Westbrook: Is there a test or stub for this?

packages/calendar/src/Calendar.ts
```diff
@@ -89,8 +89,8 @@ export class Calendar extends SpectrumElement {
     @property({ type: Boolean, reflect: true })
     padded = false;
 
-    @state()
-    private currentDate!: CalendarDate;
+    @property({ attribute: false })
+    currentDate!: CalendarDate;
```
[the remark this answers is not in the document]
Westbrook: I see this is used in tests. Are there other use cases? If not, we should likely move it back to `private` and work around that reality in tests, rather than surfacing a property that we don't need to the public API.

Westbrook (commented):
Small questions and then we're good to merge this down to the project branch!

tools/input-segments/src/InputSegments.ts
paulodetarsofm: Added

packages/date-time-picker/test/date-time-picker.test.ts
Westbrook: I'm OK with skipping test for the near term, if you are. It seems like it would be a good idea to list out some of the things that the element does that _could_ be tested now, while the changes are less cold in your mind. This was a comment asking that we list some things that it does:
```
it('makes me a sandwich', async function() {});
// etc...
```
So that we know what needs tests when it comes time for them.

Rajdeepc:
@paulodetarsofm Can you get this targeted to the main branch and then we can do a final review and get it landed? Thanks

packages/calendar/test/calendar.test.ts
```diff
@@ -10,29 +10,159 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 import { elementUpdated, expect, fixture, html } from '@open-wc/testing';
+import sinon, { spy } from 'sinon';
 
-import '../sp-calendar.js';
-import { Calendar } from '..';
 import { testForLitDevWarnings } from '../../../test/testing-helpers.js';
+import { Calendar } from '../src/Calendar.js';
```
Rajdeepc, line 16: This 
```suggestion
import { Calendar } from '../src/Calendar.js';
```
```suggestion
import { Calendar } from '@spectrum-web-components/calender';
```

Rajdeepc (changes requested):

packages/calendar/package.json
```diff
@@ -69,7 +69,8 @@
         "@spectrum-web-components/reactive-controllers": "^0.41.2"
     },
     "devDependencies": {
-        "@spectrum-css/calendar": "^4.2.4"
+        "@spectrum-css/calendar": "^4.2.4",
+        "@spectrum-web-components/story-decorator": "^0.41.2"
```
Rajdeepc, line 73: I think this should be in dependencies and not in devDependencies. Do yo not need `defaultLocale` in prod?

packages/calendar/src/Calendar.ts
```diff
@@ -311,12 +314,13 @@ export class Calendar extends SpectrumElement {
                 class="spectrum-Calendar-tableCell"
```
Rajdeepc, line 314:
```suggestion
                class="tableCell"
```
`spectrum-config`
```
converter.classToClass(
                    'spectrum-Calendar-tableCell',
                    'tableCell'
                ),
```

packages/calendar/src/Calendar.ts
```diff
@@ -176,6 +177,7 @@ export class Calendar extends SpectrumElement {
                     aria-label="Previous"
                     title="Previous"
                     class="spectrum-Calendar-prevMonth"
```
Rajdeepc, line 179: same as below

packages/calendar/src/Calendar.ts
```diff
@@ -196,6 +198,7 @@ export class Calendar extends SpectrumElement {
                     aria-label="Next"
                     title="Next"
                     class="spectrum-Calendar-nextMonth"
```
Rajdeepc, line 200: same as below

packages/calendar/stories/calendar.stories.ts
```diff
@@ -9,177 +9,71 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
...
-interface StoryArgs {
-    lang?: string;
-
-    padded?: boolean;
-    disabled?: boolean;
+type ComponentArgs = {
     selectedDate?: Date;
     min?: Date;
     max?: Date;
+    padded?: boolean;
```
Rajdeepc, line 33: this is not understandable, can you change it to something like `calenderPadding` instead

packages/calendar/stories/calendar.stories.ts
```diff
@@ -9,177 +9,71 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
...
+type StoryArgs = ComponentArgs & {
+    onChange?: (dateTime: Date) => void;
+};
 
+interface SpreadStoryArgs {
     [prop: string]: unknown;
 }
 
 const renderCalendar = (
     title: string,
```
Rajdeepc, line 46: why title is not a part of args?

packages/calendar/test/calendar.test.ts
```diff
@@ -10,12 +10,24 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 import { elementUpdated, expect, fixture, html } from '@open-wc/testing';
+import sinon, { spy } from 'sinon';
```
Rajdeepc, line 13: SWC doesn't use sinon. Can you use testing helpers from [here](https://github.com/adobe/spectrum-web-components/blob/mirekszot/date-time-picker-ui-fixes/test/testing-helpers.ts#L0-L1)

packages/date-time-picker/README.md
```diff
@@ -0,0 +1,36 @@
...
+import { DateTimePicker } from '@spectrum-web-components/date-time-picker';
+```
+
+## Example
+
+```html
+<sp-date-time-picker></sp-date-time-picker>
+```
+
+## To-do list
```
Rajdeepc, line 30: Can we remove this?

packages/date-time-picker/src/DateTimePicker.ts
```diff
@@ -0,0 +1,159 @@
...
+
+                <sp-picker-button
+                    class="picker-button"
+                    ?open=${this.open}
+                    ?quiet=${this.quiet || this.readonly}
+                    ?invalid=${this.invalid}
+                    ?disabled=${isDisabled}
+                    @click=${this.showPicker}
+                >
+                    <slot name="calendar-icon" slot="icon">
```
Rajdeepc, line 82: This is a bit _erratic_ here. Why would you create both a slot definition and use slot="icon"?

packages/date-time-picker/src/DateTimePicker.ts
```diff
@@ -0,0 +1,159 @@
...
+
+                <!-- TODO: Enable "receives-focus" when calendar is navigable via keyboard -->
+                <sp-overlay
+                    placement="bottom-start"
+                    receives-focus="false"
+                    .triggerElement=${this as HTMLElement}
+                    ?open=${this.open}
+                    @sp-closed=${this.hidePicker}
+                >
+                    <sp-popover class="popover">
```
Rajdeepc, line 95: FYI: This creates a stringent boundary when we would want to render this in mobile!
In mobile the calender should render in a tray, not sure if it has been design approved or not!

packages/date-time-picker/src/DateTimePicker.ts
```diff
@@ -0,0 +1,159 @@
...
+            </div>
+        `;
+    }
+
+    public showPicker(): void {
+        this.pickerDate = this.getDateFromSegments();
+        this.open = true;
+    }
+
+    public hidePicker(): void {
```
Rajdeepc, line 113: We would love to have a documentation block!

packages/date-time-picker/src/DateTimePicker.ts
```diff
@@ -0,0 +1,159 @@
...
+     * Updates the value of the internal property used to define the date used by the calendar with the value received
+     *
+     * @param event - Event with the value emitted by the calendar
+     */
+    private handleDate(event: CustomEvent<Date | undefined>): void {
+        event.stopPropagation();
+
+        const dateTime = event.detail;
+
+        if (dateTime) {
```
Rajdeepc, line 131: How are you handling the edge case here if dateTime is undefined?

packages/date-time-picker/src/DateTimePicker.ts
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        this.yearSegment.value = year;
+        this.formatValue(this.yearSegment);
+
+        this.monthSegment.value = month;
+        this.formatValue(this.monthSegment);
+
+        this.daySegment.value = day;
+        this.formatValue(this.daySegment);
```
Rajdeepc, line 155: can you not do these formatValue early and do it at the last execution statements? Much cleaner approach

Rajdeepc (commented):
Good Progress! Few review comments then we can land this!

packages/calendar/stories/calendar.stories.ts
mizgaionutalexandru: I think what was meant here was to distinguish between the component attributes/properties and the <h1> used in each story. I will address this in a future PR.

packages/date-time-picker/README.md
mizgaionutalexandru: I don't think it's the time to remove them. I will remove them incrementally once they are completed.

packages/date-time-picker/src/DateTimePicker.ts
mizgaionutalexandru: This was done to use a default icon for the picker button as well as giving the consumers the posibility of choosing another one, as per the component's RFC.

packages/date-time-picker/src/DateTimePicker.ts
mizgaionutalexandru: AFAIK there's no design for mobile. If the spectrum design team has something specific in mind we can do that. I agree that aligning with the React Spectrum approach (using the tray) should be desired but for the moment this approach can work. I can change this in a future PR before release and I think the MatchMediaController can be used here.

Rajdeepc: Right! This is just na FYI

packages/calendar/src/Calendar.ts
Rajdeepc: Would like you to change this to the above naming pattern used in the library

Rajdeepc:
Good to go into a feature branch! Please make the above suggested changes in your feature branch and we can take this forward. 
Thanks for doing this!

Issue #3855, opened 2023-12-06
[Bug]: Picker flickering and tray is not showing up on first load - [MOBILE ONLY]


Rajdeepc:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Picker

### Expected behavior

Picker should open as tray in mobile without flickering on every click.

### Actual behavior

Picker is flickering on first load in mobile and clicking on ESC and then reopening it opens as tray as expected. Only fails in first load in mobile

### Screenshots

sp-tray should have open attribute. This happens only on first click of the picker. If you press ESC and click then it has the open attribute

![Image](https://github.com/adobe/spectrum-web-components/assets/8790510/e27fa0c5-a8a9-4431-b082-dea4b504a3cf)

Video:

[screen-capture.webm](https://github.com/adobe/spectrum-web-components/assets/8790510/984bc3a8-a690-421f-bf70-65687d2d35ce)

### What browsers are you seeing the problem in?

All Browsers in Mobile

### How can we reproduce this issue?

1. Go to https://opensource.adobe.com/spectrum-web-components/storybook/index.html?path=/story/picker--default
2. Emulate in MOBILE VIEW
3. Click on the Picker
4. Check the picker flashes and doesn't open the tray
5. Click ESC or reload the page.
6. Open the picker again. It flashed but opens as tray

### Logs taken while reproducing problem

No logs in console

PR #3858, opened 2023-12-07
chore(overlay): shadow elements receives open when open is true in sp…

Rajdeepc:
## Description

We have added the logic to the slotted shadow element of `<sp-overlay>` will receive an `open` attribute whenever the overlay is opened. This behavior is particularly useful when you want to add an `open` attribute to the slotted shadow element dynamically.

## Related issue(s)

- #3855 

## Motivation and context

Picker is flickering on first load in mobile and clicking on ESC and then reopening it opens as tray as expected. Only fails in first load in mobile

## How has this been tested?

Fix: 
Uploading issue 7 fix.mov…

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/overlay/README.md
+++ b/packages/overlay/README.md
@@ -169,6 +169,10 @@ When a `triggerElement` is present, either through an ID reference established v
     -   **if you have a `placement` but not a `triggerElement`** the `<sp-overlay>` will not be positioned due to their being nothing for the `plaement` to reference when so doing
 -   **if no `placement` is available** the content will not be placed with the expectation that the content itself or the consuming application will handle placement of the overlaid content. This is commonly what will happen for `type="modal"` and `type="page"` overlays as they are likely meant to cover the entire screen, whether visibly (via an `<sp-underlay>`, an element that includes one, or similar) or figuratively (as when modal content is not delivered with a backdrop or scrim).
 
+Additionally, a new behavior has been introduced to `<sp-overlay>`:
+
+-   **if a `open` is true**, the slotted shadow element of `<sp-overlay>` will receive an `open` attribute whenever the overlay is opened. This behavior is particularly useful when you want to add an `open` attribute to the slotted shadow element dynamically.
+
 ### Type
 
 The `type` of an Overlay outlines a number of things about the interaction model within which is works.
--- a/packages/overlay/src/OverlayDialog.ts
+++ b/packages/overlay/src/OverlayDialog.ts
@@ -55,9 +55,7 @@ export function OverlayDialog<T extends Constructor<AbstractOverlay>>(
             const start =
                 (el: OpenableElement, index: number) =>
                 async (): Promise<void> => {
-                    if (typeof el.open !== 'undefined') {
-                        el.open = targetOpenState;
-                    }
+                    el.open = targetOpenState;
                     if (!targetOpenState) {
                         const close = (): void => {
                             el.removeEventListener('close', close);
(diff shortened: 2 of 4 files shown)


packages/overlay/README.md
```diff
@@ -169,6 +169,10 @@ When a `triggerElement` is present, either through an ID reference established v
     -   **if you have a `placement` but not a `triggerElement`** the `<sp-overlay>` will not be positioned due to their being nothing for the `plaement` to reference when so doing
 -   **if no `placement` is available** the content will not be placed with the expectation that the content itself or the consuming application will handle placement of the overlaid content. This is commonly what will happen for `type="modal"` and `type="page"` overlays as they are likely meant to cover the entire screen, whether visibly (via an `<sp-underlay>`, an element that includes one, or similar) or figuratively (as when modal content is not delivered with a backdrop or scrim).
 
+Additionally, a new behavior has been introduced to `<sp-overlay>`:
+
+-   **if a `open` is true**, the slotted shadow element of `<sp-overlay>` will receive an `open` attribute whenever the overlay is opened. This behavior is particularly useful when you want to add an `open` attribute to the slotted shadow element dynamically.
```
Westbrook, line 174: An element can't _technically_ be both "slotted" and a "shadow" element.

packages/overlay/README.md
```diff
@@ -169,6 +169,10 @@ When a `triggerElement` is present, either through an ID reference established v
     -   **if you have a `placement` but not a `triggerElement`** the `<sp-overlay>` will not be positioned due to their being nothing for the `plaement` to reference when so doing
 -   **if no `placement` is available** the content will not be placed with the expectation that the content itself or the consuming application will handle placement of the overlaid content. This is commonly what will happen for `type="modal"` and `type="page"` overlays as they are likely meant to cover the entire screen, whether visibly (via an `<sp-underlay>`, an element that includes one, or similar) or figuratively (as when modal content is not delivered with a backdrop or scrim).
 
+Additionally, a new behavior has been introduced to `<sp-overlay>`:
```
Westbrook, line 172: We should probably add content at the end of the `Example` section, instead. Something like:

```
When a `<sp-overlay>` element is opened, it will pass that state to its direct children elements as the property `open`, which it will set to `true`. Elements should react to this by initiating any transition between closed and open that they see fit. Similarly, `open` will be set to `false` when the `<sp-overlay>` element is closed.
```

Westbrook (commented):
We should specifically update [this](https://github.com/adobe/spectrum-web-components/blob/main/projects/documentation/content/migrations/2023-8-18.md?plain=1#L21) to say the `property` will be addressed and that non-SWC patterns should ensure that any reflection of the property the attribute is handled internal to the element in question.

projects/documentation/content/migrations/2023-8-18.md
```diff
@@ -18,7 +18,7 @@ If you had previously done work in your application to interact directly with ov
 
 ### Remove the `open` attribute from content that is meant to be overlaid.
 
-The `open` attribute will be addressed directly by the Overlay API itself to manage the visual transition from the closed to open state of the overlaid content, and management of this attribute by the parent application could prevent those transitions from occurring as expected.
+The `open` attribute will be addressed and the non-SWC patterns should ensure that any reflection of the property the attribute is handled internal to the element in question.
```
Westbrook, line 21:
```suggestion
The `open` property will be addressed directly by the Overlay API itself. Elements wanted to to manage the visual transition from the closed to open state of the overlaid content should be handled reactively in response to that change. Management of this attribute by the parent application could prevent those transitions from occurring as expected.
```

Westbrook (approved):
LGTM!

PR #3859, opened 2023-12-08
bug(checkbox): added missing readonly prop value

Rajdeepc:
## Description

Added missing readonly prop value

## Related issue(s)

- 

## Motivation and context

Showing an error on readonly prop in `CheckBox` swc-react component.

## How has this been tested?

`yarn link` at packages/checkbox in SWC

and `yarn link "@spectrum-web-components/checkbox"`

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

Screenshot 2023-12-08 at 2 22 05 PM

## Types of changes

## Checklist

## Best practices


Westbrook (commented):
Is there a way to trigger the error that raised this? Being typings and not functionality, it seems like maybe not, but it would be good if we could add something that prevented issues like this in future refactors.

Westbrook:
Related, looks like Spectrum CSS moves the `input` to `disabled` in this state, maybe we should do the same? 
image

packages/checkbox/src/CheckboxMixin.ts
```diff
@@ -78,6 +83,19 @@ export function CheckboxMixin<T extends Constructor<ReactiveElement>>(
...
+        protected override updated(changes: PropertyValues): void {
+            super.updated(changes);
+            if (
+                changes.has('readonly') &&
+                (typeof changes.get('readonly') !== 'undefined' ||
+                    this.readonly)
+            ) {
+                if (this.readonly) {
+                    this.inputElement.disabled = this.readonly;
+                }
```
Westbrook, line 96: This doesn't look like it supports removing `readonly`?

Rajdeepc: Yup. great catch! update to allow support for both setting and removing readonly

packages/checkbox/src/CheckboxMixin.ts
```diff
@@ -78,6 +83,17 @@ export function CheckboxMixin<T extends Constructor<ReactiveElement>>(
...
         }
+
+        protected override updated(changes: PropertyValues): void {
+            super.updated(changes);
+            if (
+                changes.has('readonly') &&
+                (typeof changes.get('readonly') !== 'undefined' ||
+                    this.readonly)
+            ) {
+                this.inputElement.disabled = this.readonly;
```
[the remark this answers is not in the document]
Rajdeepc: ok added the disabled property in the render:
```
            return html`
                <input
                    id="input"
                    name=${ifDefined(this.name || undefined)}
                    type="checkbox"
                    .checked=${this.checked}
                    ?disabled=${this.readonly} 
                    @change=${this.handleChange}
                />
            `;
        }
```

Westbrook (commented):
Is there any way that we can test the specific issue that arose here in your consumption?

Rajdeepc:
> Is there any way that we can test the specific issue that arose here in your consumption?

Yes should be added now.

Issue #586, opened 2020-04-01
Add fill-start/fill-offset functionality to Slider


Westbrook:
No fill start/fill-offset example: https://opensource.adobe.com/spectrum-css/slider.html#filled-offset

![image](https://user-images.githubusercontent.com/1156657/78150671-9dd71b00-7405-11ea-9d25-7bd3e27e1489.png)

PR #3876, opened 2023-12-15
fix(slider): added filled offset variant

Rajdeepc:
## Description
Add fill-start/fill-offset functionality to Slider

## Related issue(s)

- https://github.com/adobe/spectrum-web-components/issues/586

## How has this been tested?

-   [x] _Test case 1_
    1. [Go here](https://github.com/adobe/spectrum-web-components/pull/3876/files#diff-d739a2f7294524812acb0533f22299b6248f763b10c7f4fd487923165cf5af15R847)
    2. Run this test

https://github.com/adobe/spectrum-web-components/assets/8790510/cd77eda0-16f2-4877-9c88-b588a7accc7c

With only fill-start, fill starting from center
Screenshot 2023-12-18 at 2 40 17 PM

With a fill-start value, fill starting from fill-start value
Screenshot 2024-01-08 at 10 52 54 AM

## Types of changes

## Checklist

## Best practices


packages/slider/src/Slider.ts
```diff
@@ -38,7 +38,7 @@ import { SliderHandle } from './SliderHandle.js';
 import { streamingListener } from '@spectrum-web-components/base/src/streaming-listener.js';
 import type { NumberFormatter } from '@internationalized/number';
 
-export const variants = ['filled', 'ramp', 'range', 'tick'];
+export const variants = ['filled', 'offset', 'ramp', 'range', 'tick'];
```
Westbrook, line 41: Does this require a new variant, or can it be achieved via the additions of a `fill-start` attribute/property that would also allow for the offset to start at somewhere other than the middle value of the range?

packages/slider/src/Slider.ts
```diff
@@ -160,6 +164,9 @@ export class Slider extends SizedMixin(ObserveSlotText(SliderHandle, ''), {
     @property({ type: Boolean, reflect: true })
     public override disabled = false;
 
+    @property({ type: Boolean, reflect: true, attribute: 'fill-start' })
+    public fillStart = false;
```
Westbrook, line 168: I don't think this is supposed to be a boolean. I think it is supposed to represent where the fill _starts_ from. By default it would equal the same as `min` but then set it would allow for the fill to start from whereever.

Westbrook: Also, maybe be unneeded work, but should we also surface a `relativeValue` property that give the value relative to the start? Calculable from the outside, so maybe not.

Rajdeepc: should min the start value or the initial value represented as 'value' ?
fillStart boolean creates the filled highlight on the track.

Westbrook: I may be misunderstand, so let's check with the design team, but I understand the workflow to be:
```
<sp-slider></sp-slider> // No fill

<sp-slider
    variant="filled"
></sp-slider> // Fill from `min`, or 0 by default

<sp-slider
    variant="filled"
    fill-start="5"
    min="0"
    max="10"
></sp-slider> // Fill from 5, or half way
```

Rajdeepc: Sure I was wondering if fill-start="5" and value="6" then how will the slider behave? which takes precendence?

Westbrook:
```
0                    20
|----==0-------------|
```
I think it would look like this where the `fill-start`ed at 5 and the fill went to 6 where the handle stood. More clear might be `variant="filled" fill-start value="15" min="0" max="20"`, `variant="filled" fill-start="5" value="15" min="0" max="20"`, and `variant="filled" fill-start="5" value="0" min="0" max="20"`  as comparison:
```
0                    20
|----------====0-----|
```
```
0                    20
|----==========0-----|
```
```
0                    20
0===-----------------|
```
When `fill-start` is present, the fill starts at half the value like you've previously outlined. When `fill-start` is supplied a value, the fill starts from that value. The fill _goes_ from that number to the `value` provided. The delta from the `fill-start` to the `value` would be the `relativeValue` if we chose to surface it.

Rajdeepc: 1. For this variant `variant="filled" fill-start value="15" min="0" max="20"` , handle will be at 10 i.e the center. 
When user slides from 10 - 20 and 10 - 0 should the fill be visible for both or it should only be visible from 10-15?

2. For this variant `variant="filled" fill-start="5" value="15" min="0" max="20"` 
the handle should start from the fill-start value of "5" or from `value="15"` itself? And also do we show fill from 5-15 only or from 5-0 also?

Westbrook: `variant="filled" fill-start value="15" min="0" max="20"`
In this case the "fill" will start at half (so 10) but the value will be 15, so the "fill" will take up 1/4 of the track starting from the middle.
```
0                    20
|----------====0-----|
```
---
`variant="filled" fill-start value="10" min="0" max="20"`
Fill starting at half (10) and value at 10.
```
0                    20
|----------0---------|
```
User moves handle to 20
```
0                    20
|----------==========0
```
User moves handle to 0
```
0                    20
0===========---------|
```
---
`variant="filled" fill-start="5" value="15" min="0" max="20"`
Fill takes up half of the track starting 1/4 of the way along the track.
```
0                    20
|-----=========0-----|
```
User moves to 0, fill size is 1/4 of the track starting at the beginning of the track.
```
0                    20
0=====---------------|
```

Rajdeepc: Attached video recording for preview and review. Let me know how this solution looks now.

packages/slider/src/Slider.ts
```diff
@@ -455,4 +523,18 @@ export class Slider extends SizedMixin(ObserveSlotText(SliderHandle, ''), {
...
+        if (changes.has('value') && changes.has('fillStart')) {
+            this._cachedValue = Number(this.value);
+            if (typeof this.fillStart === 'number') {
+                this.fillStartPoint = this.fillStart;
+            } else {
+                this.fillStartPoint =
+                    (Number(this.max) - Number(this.min)) / 2 +
+                    Number(this.min);
+            }
+        }
```
Westbrook, line 538: Could this just be a getter instead? That would allow us to not place `fillStartPoint` as a reactive property, as well.

packages/slider/src/Slider.ts
```diff
@@ -348,6 +357,60 @@ export class Slider extends SizedMixin(ObserveSlotText(SliderHandle, ''), {
...
+
+    private renderFillOffset(): TemplateResult {
+        if (!this.fillStart || !this.fillStartPoint || !this._cachedValue) {
+            return html``;
+        }
+        return html`
+            <div
+                class=${`fill ${
+                    !!(this.value > this.fillStartPoint) ? 'offset' : ''
+                }`}
```
Westbrook, line 397: using `classMap({fill: true, offset: this.value > this.fillStartPoint})` might make this easier to read.

packages/slider/src/Slider.ts
```diff
@@ -348,6 +357,60 @@ export class Slider extends SizedMixin(ObserveSlotText(SliderHandle, ''), {
...
+                        this.value > this.fillStartPoint
+                            ? this.getOffsetPosition(this.fillStartPoint)
+                            : this.getOffsetPosition(this.value)
+                    }%`,
+                    width: `${this.getOffsetWidth(
+                        this.fillStartPoint,
+                        this.value,
+                        this._cachedValue
+                    )}%`,
+                })}
```
Westbrook, line 409: With a complex map like this, we may want to create it as a variable outside of the template, so that it is easier to read.

packages/slider/src/Slider.ts
```diff
@@ -438,8 +502,12 @@ export class Slider extends SizedMixin(ObserveSlotText(SliderHandle, ''), {
...
         const styles: StyleInfo = {
             width: `${size * 100}%`,
-            '--spectrum-slider-track-background-size': `${(1 / size) * 100}%`,
-            '--spectrum-slider-track-segment-position': `${start * 100}%`,
+            ...(this.handleController.size > 1 && {
+                '--spectrum-slider-track-background-size': `${
+                    (1 / size) * 100
+                }%`,
+                '--spectrum-slider-track-segment-position': `${start * 100}%`,
+            }),
```
Westbrook, line 510: What does this change do?

packages/slider/stories/slider.stories.ts
```diff
@@ -132,6 +132,69 @@ export const Default = (args: StoryArgs = {}): TemplateResult => {
...
+        </div>
+    `;
+};
+
+export const withOnlyFillStart = (args: StoryArgs = {}): TemplateResult => {
+    return html`
+        <div style="width: 500px; margin-inline: 20px;">
+            <sp-slider
+                max="1"
+                fill-start
```
Westbrook, line 160: I _think_ we still want `variant="filled"` here. I know this is less things, but it's also less clear. I could be convinced away, but it feels more expected for each of the "filled" sliders to have the same `variant` and for the `fill-start` attribute to change the way that the fill is delivered.

packages/slider/test/slider.test.ts
```diff
@@ -844,6 +844,53 @@ describe('Slider', () => {
...
+                        200,
+                        handleBoundingRect.y + handleBoundingRect.height + 100,
+                    ],
+                },
+            ],
+        });
+        await nextFrame();
+
+        expect(el.value).to.equal(24);
+        expect(el.shadowRoot.querySelector('.fill') as HTMLDivElement).to.exist;
```
Westbrook, line 892: Probably a good idea to assert that `.fill` doesn't exist to get started with?

Westbrook (commented):
One visual outcome in the stories that seems wrong: 
image

Beyond that, I think we're in alignment on how things should end up, just a couple of nips and tucks and this should be on its way!

packages/slider/stories/slider.stories.ts
Rajdeepc: So we cannot use variant="fiiled" when using this property `fill-start` becuase of this css
```
:host([variant='filled']) .track:first-child:before {
    background: var(
        --highcontrast-slider-track-fill-color,
        var(
            --mod-slider-track-fill-color,
            var(--spectrum-slider-track-fill-color)
        )
    );
}
```

If fill-start has value then the track will be filled even if it is not starting from 0

packages/slider/src/Slider.ts
Rajdeepc: This is redundant now. This was done to remove discrepancy in css on tick label sliders

packages/slider/README.md
```diff
@@ -83,14 +83,91 @@ import { Slider } from '@spectrum-web-components/slider';
...
+                initSlider();
+            });
+        });
+</script>
+```
+
+<script type="module">
+    const initSlider = async () => {
+        const slider = document.querySelector('#fill-start-slider');
+        slider.fillStart = 0.3
```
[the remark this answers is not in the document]
Rajdeepc: Yes! I have adjusted the logic to add fill-start as an empty attribute as well as with a value

packages/slider/test/slider.test.ts
Rajdeepc: class fill will be there when fill-start or fill-start with value is there with width as 0% when only fill-start

Westbrook: Then a presence check is likely not enough to actually guarantee that the role that element is playing is complete to the situation in question.

packages/slider/src/Slider.ts
Rajdeepc: This has been simplified to make it more robust and DRY

packages/slider/stories/slider.stories.ts
Rajdeepc: Thanks for pointing this use case. This should be handled now

packages/slider/src/Slider.ts
Rajdeepc: Now using the `fillStartPoint` getter
```
private get fillStartPoint(): number {
        if (this.fillStart) {
            return Number(this.fillStart);
        } else {
            return (Number(this.max) - Number(this.min)) / 2 + Number(this.min);
        }
    }
```

Rajdeepc:
> One visual outcome in the stories that seems wrong: image
> 
> Beyond that, I think we're in alignment on how things should end up, just a couple of nips and tucks and this should be on its way!

Updated the logic to handle the missed use case. Storybook and Docs are updated.

Rajdeepc:
@Westbrook 
https://github.com/adobe/spectrum-web-components/assets/8790510/17e06f48-065e-49f8-a283-372328409f69

packages/slider/src/Slider.ts
```diff
@@ -348,6 +355,70 @@ export class Slider extends SizedMixin(ObserveSlotText(SliderHandle, ''), {
...
+        fillStartValue: number,
+        currentValue: number,
+        cachedValue: number
+    ): number {
+        let distance: number;
+        if (fillStartValue === cachedValue) {
+            distance = Math.abs(currentValue - fillStartValue);
+        } else {
+            distance = Math.abs(currentValue - fillStartValue);
+        }
```
Westbrook, line 377: This seems like the same on both sides of the logic.

packages/slider/stories/slider.stories.ts
```diff
@@ -132,6 +132,81 @@ export const Default = (args: StoryArgs = {}): TemplateResult => {
...
 };
 
+export const Filled = (args: StoryArgs = {}): TemplateResult => {
+    return html`
+        <div style="width: 500px; margin-inline: 20px;">
+            <sp-slider
+                max="1"
+                variant="filled"
+                min="0"
+                value=".5"
```
Westbrook, line 142: Nit, can we use a value other than the middle of the range so that this story shows the fill by default in VRT so that we can test that the output stays the same over time.

Westbrook (commented):
Sorry, found two quick nits.

packages/slider/stories/slider.stories.ts
Rajdeepc: Screenshot 2024-01-17 at 11 34 00 AM

Westbrook: I'm sorry, I put this comment on the wrong story. 😞 I meant for https://bug-slider-filled--spectrum-web-components.netlify.app/storybook/?path=/story/slider--fill-start for all the same reasons. I swear this is the last thing!

Rajdeepc: Screenshot 2024-01-17 at 6 45 46 PM

Westbrook (commented):
This LGTM! Checking with @beeduul to confirm this covers a use case that recently came up for him and then we can merge this in. Thanks for getting this together!

beeduul:
Thanks for checking @Westbrook.  My use case only requires the vanilla `fill-start` attribute to start in the middle of the slider, so lgtm!

Westbrook (approved):
:shipit: 🥳

PR #3880, opened 2023-12-18
chore: update playwright and @web dependencies

Westbrook:
## Description
- update playwright to 1.40.1
  - requires new golden image cache
  - requires Overlay test workflow changes
  - requires Picker a11y delivery changes
  - normalizes Firefox delivery of Picker a11y model
  - requires Split Button a11y delivery updates
- update `@web` packages
  - requires stopSession package patch
  - requires commonJS plugin when building storybook

## Types of changes

## Checklist


packages/overlay/test/index.ts
```diff
@@ -429,6 +420,20 @@ export const runOverlayTriggerTests = (type: string): void => {
                     'inner click content available at point'
                 ).to.be.true;
 
+                // Why does this make the test pass in Chromium? 🤷🏻‍♂️
```
najikahalsema, line 423: How did you figure this out?

najikahalsema (approved):
LGTM! curious about the chromium tests passing but not a blocker. more curious how you discovered that.

packages/overlay/test/index.ts
Westbrook: In the next test, it was previously using `preseSpace()` (though I also don't know why). When the test started to fail, I switched that to an actual keyboard interaction and it started to pass, so I adopted the approach here, too.

najikahalsema: Makes sense to me.... interesting. Thanks for explaining.

PR #3894, opened 2023-12-19
fix(combobox): add combobox pattern

### LINKED ISSUE #478: Add Spectrum Combobox Pattern

Westbrook:
## Description
Add a Combobox pattern to the library:
```
<sp-combobox label="Declarative combobox">
    <sp-menu-item value="0">Declare this</sp-menu-item>
    <sp-menu-item value="1">Declare that</sp-menu-item>
    <sp-menu-item value="2">Declare the other</sp-menu-item>
</sp-combobox>

<sp-combobox
  label="Imparative combobox"
  options='[
    {"value": "0", "itemText": "Imperative this"},
    {"value": "1", "itemText": "Imperative that"},
    {"value": "2", "itemText": "Imperative the other"}
  ]'
>
</sp-combobox>
```
Visit [the docs](https://combobox--spectrum-web-components.netlify.app/component/combobox) for more information.

## Related issue(s)
- fixes #478

## How has this been tested?
-   [ ] _Test case 1_
    1. Go [here](https://combobox--spectrum-web-components.netlify.app/storybook/index.html?path=/story/combobox--default)
    2. Test an `autocomplete="none"` Combobox
-   [ ] _Test case 2_
    1. Go [here](https://combobox--spectrum-web-components.netlify.app/storybook/index.html?path=/story/combobox--list-autocomplete)
    2. Test an `autocomplete="list"` Combobox

image

## Types of changes

## Checklist


packages/combobox/src/Combobox.ts
```diff
@@ -0,0 +1,577 @@
...
+import { Textfield } from '@spectrum-web-components/textfield';
+
+import styles from './combobox.css.js';
+import chevronStyles from '@spectrum-web-components/icon/src/spectrum-icon-chevron.css.js';
+import { Menu, MenuItem } from '@spectrum-web-components/menu';
+
+export type ComboboxOption = {
+    id: string;
+    value: string;
+};
```
najikahalsema, line 46: Question that came up for me as I work on the benchmark test: do we want to expand this API to _also_ to accept an array of strings that could get converted to `ComboboxOptions`? The `id` could match the `value`, or we could do some fancy RegEx or whatever to convert the `value` to something more multipurpose. 

The reason I ask is because I have an `String` array of a list of countries, and it's annoying for me to have to convert them into an object so I can fit it into this API. It'd be nice if I could just have the combobox do the work of converting it for me. I'm also thinking about that in case we end up using the select-only combobox as a basis for the lazy picker or something... What do you think, @Westbrook?

najikahalsema: Currently, we can do this with a map and have all the values in the lightDOM, but I wouldn't want that to be the only option.

```
        <sp-combobox>
            ${countryList.map(
                (value, index) => html`
                    <sp-menu-item id=${index} value=${value}>
                        ${option.value}
                    </sp-menu-item>
                `
            )}
        </sp-combobox>
```

Westbrook: This is a great question for potential consumers of the element. I'm not opposed to it so much as I would be surprised to see someone with a functional data set that didn't already have not a unique identifier AND a user readable value.

In that case that there were, conditioning data to support this is also not too crazy:

```html
<sp-combobox
    .options=${countryList.map((value, index) => ({ value, index }))}
></sp-combobox>
```

This repeats the data processing more than would be appropriate, so you would probably cache a conditioned array in the parent context. That's not so different than us doing it as well, so maybe we would just do it by default.

The question will always go back to whether it actually makes sense to for consumers to actually reference values by their visible value vs always having a unique ID for them.

hunterloftis (approved):

PR #3907, opened 2024-01-02
test(combobox): update data management tests for current API

Westbrook:
- decrease failing test count
- remove obsolete test file
- refactor tests for updated element API
- update element lifecycle for latest API


packages/combobox/test/combobox.data.test.ts
```diff
@@ -275,20 +322,28 @@ describe('Combobox Data', () => {
...
+        const processedOptions = el.availableOptions.map(({ id, value }) => ({
+            id,
+            value,
+        }));
+
+        expect(processedOptions).to.deep.equal(options);
     });
-    it('accepts adding through slots', async () => {
+    it('accepts adding through slots', async function () {
+        this.retries(0);
```
najikahalsema, line 337: Do we want to remove this or keep it for now and remove it when we merge to main? If so, we should probably make a note.

najikahalsema (approved):
From 26 failures to 19!! Hooray!! 👏

packages/combobox/test/combobox.data.test.ts
Westbrook: Good catch; removed!

PR #3909, opened 2024-01-02
docs(icons-ui,icons-workflow): correct typos

mattnovelli:
## Description

Fixed spelling of "like"

## Motivation and context

Nitpick.

### DIFF
--- a/packages/icons-ui/README.md
+++ b/packages/icons-ui/README.md
@@ -103,7 +103,7 @@ Every icons can be customized via the following options:
 
 ### Extended use cases
 
-The default exports of this package are pre-wrapped via `setCustomTemplateLiteralTag` in the `html` template tag from `lit-html`, and works like the following::
+The default exports of this package are pre-wrapped via `setCustomTemplateLiteralTag` in the `html` template tag from `lit-html`, and work like the following::
 
 ```js
 import { Arrow75Icon } from '@spectrum-web-components/icons-ui';
--- a/packages/icons-ui/README.md
+++ b/packages/icons-ui/README.md
@@ -103,7 +103,7 @@ Every icons can be customized via the following options:
 
 ### Extended use cases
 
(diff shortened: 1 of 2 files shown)


packages/icons-ui/README.md
```diff
@@ -103,7 +103,7 @@ Every icons can be customized via the following options:
 
 ### Extended use cases
 
-The default exports of this package are pre-wrapped via `setCustomTemplateLiteralTag` in the `html` template tag from `lit-html`, and work liek the following::
+The default exports of this package are pre-wrapped via `setCustomTemplateLiteralTag` in the `html` template tag from `lit-html`, and works like the following::
```
Westbrook, line 106: I'm pretty sure the via subject verb agreement the correct plurality is `work` being we're talking about `exports`.

Westbrook:
```suggestion
The default exports of this package are pre-wrapped via `setCustomTemplateLiteralTag` in the `html` template tag from `lit-html`, and work like the following::
```

packages/icons-workflow/README.md
```diff
@@ -103,7 +103,7 @@ Every icons can be customized via the following options:
 
 ### Extended use cases
 
-The default exports of this package are pre-wrapped via `setCustomTemplateLiteralTag` in the `html` template tag from `lit-html`, and work liek the following::
+The default exports of this package are pre-wrapped via `setCustomTemplateLiteralTag` in the `html` template tag from `lit-html`, and works like the following::
```
Westbrook, line 106:
```suggestion
The default exports of this package are pre-wrapped via `setCustomTemplateLiteralTag` in the `html` template tag from `lit-html`, and work like the following::
```

Westbrook (approved):
Thanks for this @mattnovelli 🙇🏼

PR #3920, opened 2024-01-05
[DO NOT MERGE] Seeing if we can get perf test output with the current changes.

Westbrook:
Moves from `.css.js` files to `.min.css` files with `assert { type: 'css' }` in a way that will only work in Chromium, but would point to the important of pushing on these APIs with other browser implementors.


caseyisonit:
@Rajdeepc @rubencarvalho can we close this PR? This doesnt seem like something we will be adopting

rubencarvalho:
Closing because it is currently not relevant to our goals - we might revisit this under the scope of unification efforts.

PR #3922, opened 2024-01-06
chore(combobox): cleanup unused code

Westbrook:
## Description
Remove commented out code and unused `spectrum-textfield.css`.

Rename some `on*` handlers to `handle*` to prevent overlaps with other UI frameworks.

Works through the new VRTs from `main` which causes failures in that part of the test cycle.


najikahalsema (changes requested):
Actual code removal looks good to me. We could probably also remove that `test/index.html` file. Can we get CI updated so that it passes the VRTs for the combobox sizes?

Westbrook:
@najikahalsema removed!

najikahalsema (approved):
Thank you!

PR #3953, opened 2024-01-18
test(combobox): add accessibility tests

najikahalsema:
## Description

I have made the executive decision that a combobox, even with a tooltip, still needs a label. 

To-do: 
- [x] Label the combobox
- [x] get a11y snapshot tests passing
- [x] A test that proves `aria-activedescendent` matches the id of the selected menu item when press ArrowDown from the textfield. 
- [x] Add demo with help text.
Will need to abstract the `name` attribute magic happening in the tests into a helper method at a later date. Someday... but not today. 

## Related issue(s)

-

## Motivation and context

accessibility is cool :) 

## How has this been tested?

Unit tests!

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices


najikahalsema:
why is my commit history always goofy :/

packages/combobox/src/Combobox.ts
```diff
@@ -306,17 +339,13 @@ export class Combobox extends Textfield {
...
-                aria-label=${ifDefined(this.label || this.placeholder)}
-                aria-invalid=${ifDefined(this.invalid || undefined)}
                 maxlength=${ifDefined(
                     this.maxlength > -1 ? this.maxlength : undefined
                 )}
                 minlength=${ifDefined(
                     this.minlength > -1 ? this.minlength : undefined
                 )}
                 pattern=${ifDefined(this.pattern)}
-                placeholder=${ifDefined(this.placeholder)}
```
Westbrook, line 319: Did we intend to remove this here?

packages/combobox/stories/combobox.stories.ts
```diff
@@ -15,57 +15,61 @@ import { html, TemplateResult } from '@spectrum-web-components/base';
...
+    { id: 'o5', value: 'Abcdeaaaaaaaa' },
+    { id: 'o6', value: 'Abcdefaaaaaaa' },
+    { id: 'o7', value: 'Abcdefgaaaaaa' },
+    { id: 'o8', value: 'Abcdefghaaaaa' },
+    { id: 'o9', value: 'Abcdefghiaaaa' },
+    { id: 'o10', value: 'Abcdefghijaaa' },
+    { id: 'o11', value: 'Abcdefghijkaa' },
+    { id: 'o12', value: 'Abcdefghijkla' },
+    { id: 'o13', value: 'Abcdefghijklm' },
+];
```
Westbrook, line 52: Note to selves, we should probably come up with a closer to real life set of demo data...

packages/combobox/test/combobox-a11y.test.ts
```diff
@@ -0,0 +1,309 @@
...
+import { a11ySnapshot, findAccessibilityNode } from '@web/test-runner-commands';
+import { TestableCombobox } from './index.js';
+import { withFieldLabel, withHelpText } from '../stories/combobox.stories.js';
+import { MenuItem } from '@spectrum-web-components/menu';
+
+const isWebKit =
+    /AppleWebKit/.test(window.navigator.userAgent) &&
+    !/Chrome/.test(window.navigator.userAgent);
+
+const comboboxFixture = async (): Promise<TestableCombobox> => {
```
Westbrook, line 28: Is one of the storybook stories sufficient to test against so that we don't need another instance of this?

packages/combobox/test/combobox-a11y.test.ts
```diff
@@ -0,0 +1,309 @@
...
+    );
+
+    return el;
+};
+
+describe('Combobox accessibility', () => {
+    afterEach(() => {
+        const overlays = document.querySelectorAll('active-overlay');
+        overlays.forEach((overlay) => overlay.remove());
+    });
```
Westbrook, line 51: From old Overlay API, please remove.

packages/combobox/test/combobox-a11y.test.ts
```diff
@@ -0,0 +1,309 @@
...
+        const test = await fixture<HTMLDivElement>(html`
+            <div>${withFieldLabel()}</div>
+        `);
+        const el = test.querySelector('sp-combobox') as unknown as Combobox;
+
+        await elementUpdated(el);
+
+        await expect(el).to.be.accessible();
+    });
+    it('renders accessibly with <sp-field-label>', async () => {
```
Westbrook, line 71:
```suggestion
    it('renders accessibly with <sp-help-text>', async () => {
```
Maybe?

packages/combobox/test/combobox-a11y.test.ts
```diff
@@ -0,0 +1,309 @@
...
+        await expect(el).to.be.accessible();
+    });
+    it('renders with an active descendent', async () => {
+        const el = await comboboxFixture();
+
+        const opened = oneEvent(el, 'sp-opened');
+        el.open = true;
+        await opened;
+
+        el.activeDescendant = el.availableOptions[0];
```
Westbrook, line 202: Does this practically effect the UI? Or do we need to activate a descendant by interacting with the Combo Box first?

packages/combobox/test/combobox-a11y.test.ts
```diff
@@ -0,0 +1,309 @@
...
+        await elementUpdated(el);
+
+        await expect(el).to.be.accessible();
+    });
+    it('manages aria-activedescendant', async () => {
+        // a11ySnapshot does not track the aria-activedescendant, hence querySelecting
```
Westbrook, line 208: 🙇🏼

packages/combobox/test/combobox-a11y.test.ts
```diff
@@ -0,0 +1,309 @@
...
+        const el = await comboboxFixture();
+        await elementUpdated(el);
+
+        expect(el.activeDescendant).to.be.undefined;
+
+        el.focus();
+        await elementUpdated(el);
+
+        el.focusElement.dispatchEvent(arrowDownEvent());
```
Westbrook, line 217: Please use `sendKeys()` whenever possible.

Westbrook (commented):
This is reviewing the code. I'll need to manual confirm the accessibility, probably in the morning.

packages/combobox/src/Combobox.ts
najikahalsema: Oh, yeah, I did. There's no placeholder text for the combobox. We didn't even have `this.placeholder` defined, iirc

packages/combobox/test/combobox-a11y.test.ts
najikahalsema: I liked that the test fixture only had 4 options, but I guess we could configure the stories to accept the `options` as an argument in the stories and just feed that into the test. Idk.

packages/combobox/test/combobox-a11y.test.ts
najikahalsema: 😅

packages/combobox/test/combobox-a11y.test.ts
najikahalsema: It does not! I will change this test to be more practical.

najikahalsema: Actually, what's the goal here? 

The test tagline is "renders with an active descendent", but then the actual test criteria is `await expect(el).to.be.accessible();`. 

If we're going to affect the UI, would the next test, "manages `aria-activedescendant`", be sufficient if we added the `expect(el).to.be.accessible();` to that test? Or is the goal actually to render something with an active descendant by default? I can't imagine that scenario actually being a use case, though...

Westbrook: That seems reasonable to me. It is pretty similar to the following test, yes.

packages/combobox/test/combobox-a11y.test.ts
Westbrook: Maybe we make the story template take options as an argument?

This also links up with https://github.com/adobe/spectrum-web-components/pull/3953#discussion_r1466918896 and can be addressed later, though we should keep notes on any tasks of which we allow a deferral.

packages/combobox/src/Combobox.ts
Westbrook: I thought [this](https://github.com/adobe/spectrum-web-components/blob/main/packages/textfield/src/Textfield.ts#L77) would apply here, but maybe not. As much a design and accessibility speak against placeholder content, many consumers want control over it, so we likely should include it when possible.

najikahalsema: The point I'm trying to make is that a combobox doesn't have placeholder content to begin with. It's not part of the design spec or the WAI ARIA spec...

Westbrook: Ah yes, `placeholder` is a part of the underlying `<input type="text">` https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/placeholder, for good or ill people like to use it...

packages/combobox/test/benchmark/basic-test.ts
```diff
@@ -14,11 +14,11 @@ import '@spectrum-web-components/combobox/sp-combobox.js';
 import '@spectrum-web-components/menu/sp-menu-item.js';
 import { html } from '@spectrum-web-components/base';
 import { measureFixtureCreation } from '../../../../test/benchmark/helpers.js';
-import { comboboxOptions } from '../index.js';
+import { benchmarkOptions } from '../index.js';
```
najikahalsema, line 17: changed this var name because it felt too generic.

PR #3956, opened 2024-01-19
fix(overlay): clean position data on close

### LINKED ISSUE #3955: [Bug]: re-opening an action menu on the right side of the page can open outside window bounds
### Code of conduct

### Impacted component(s)

sp-action-menu

### Expected behavior

If there's space on the page, an action menu should always open within the page bounds.

### Actual behavior

If an action menu is on the right side of the page, you can get the menu to open off the right side of the page after opening it a few times. I was able to repro this in the story book by hacking the menu css to push the action menu to the right.

### Screenshots

action-menu-offscreen

_No response_

1. Go to https://opensource.adobe.com/spectrum-web-components/storybook/index.html?path=/story/action-menu--icon-only
2. Hack the css for the action menu to push it to the right of the page (e.g. by adding `float: right;`)
3. Open and close the menu a few times
4. Eventually you should be able to see the issue. 

_No response_

_No response_

Westbrook:
## Description
Clear styling applied when positioning after an Overlay is closed again so that previous measurements do not negatively effect subsequent positioning.

~If you can think of a non-manual test for this, I'm all ears.~

## Related issue(s)
- fixes #3955

## How has this been tested?
-   [ ] _Test case 1_
    1. Go [here](https://repeat-placement--spectrum-web-components.netlify.app/storybook/?path=/story/action-menu--icon-only&args=align:end)
    2. Open/close the Action Menu repreatedly
    3. See that at no point does the Menu flow off the edge of the page

## Types of changes

## Checklist

### DIFF
--- a/packages/overlay/src/PlacementController.ts
+++ b/packages/overlay/src/PlacementController.ts
@@ -267,13 +267,20 @@ export class PlacementController implements ReactiveController {
         }
     }
 
-    public resetOverlayPosition = (): void => {
-        if (!this.target || !this.options) return;
-
+    public clearOverlayPosition(): void {
+        if (!this.target) {
+            return;
+        }
         this.target.style.removeProperty('max-height');
-        this.target.style.removeProperty('height');
+        this.target.style.removeProperty('max-width');
         this.initialHeight = undefined;
         this.isConstrained = false;
+    }
+
+    public resetOverlayPosition = (): void => {
+        if (!this.target || !this.options) return;
+        this.clearOverlayPosition();
+
         // force paint
         this.host.offsetHeight;
         this.computePlacement();
--- a/packages/action-menu/stories/action-menu.stories.ts
+++ b/packages/action-menu/stories/action-menu.stories.ts
@@ -10,6 +10,7 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 import { html, TemplateResult } from '@spectrum-web-components/base';
+import { ifDefined } from '@spectrum-web-components/base/src/directives.js';
 
 import '@spectrum-web-components/action-menu/sp-action-menu.js';
 import '@spectrum-web-components/menu/sp-menu.js';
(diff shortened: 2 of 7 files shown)


packages/overlay/src/PlacementController.ts
```diff
@@ -267,13 +267,17 @@ export class PlacementController implements ReactiveController {
         }
     }
 
-    public resetOverlayPosition = (): void => {
-        if (!this.target || !this.options) return;
-
+    public clearOverlayPosition(): void {
         this.target.style.removeProperty('max-height');
-        this.target.style.removeProperty('height');
+        this.target.style.removeProperty('max-width');
```
adixon-adobe, line 272: Little nit, but this pattern looks a little fragile (especially since these are set with intercap naming). I'm not sure if this is a frequent pattern in the library, but if it is, it might be worth adding a utility that makes the reset of style overrides more automatic. So e.g.


```
// To set overrides
this.targetOverrides = new StyleOverride(target, { maxWidth: ..., maxHeight: .... })`

// To clear
this.targetOverrides.reset()
```

adixon-adobe (approved):
I have a small nit/suggestion that I could see turning into a followup ticket. Again, only if this is a common pattern, since it looks like it'd be easy to introduce regressions when refactoring style overrides like this.

adixon-adobe:
This makes sense to me, and appears to address the bug. Thanks for the quick followup!

packages/overlay/src/PlacementController.ts
Westbrook: This is a nice looking pattern. If we were doing this elsewhere I'd definitely pick up a version of this to simplify things!

hunterloftis:
> If you can think of a non-manual test for this, I'm all ears.

With the original repro, when using the mouse, I get nondeterministic results (eventually a click does trigger the issue):

https://opensource.adobe.com/spectrum-web-components/storybook/index.html?path=/story/action-menu--icon-only

But when using the spacebar, so far, 100% of the time I can reproduce the issue (first activation -> opens correctly, second activation -> opens offscreen). If keyboard input is also deterministic in the test suite, then maybe that's a route to write a solid test?

TarunAdobe:
I am not able to open the action-menu on pressing spacebar while the it is focused. However pressing enter does open it as expected.

Westbrook:
@hunterloftis this approach worked! The new test fails when run against the proceeding branch/commit and passes when run against the changes herein.

TarunAdobe (approved):

PR #3967, opened 2024-01-25
feat: basic POC for Component Fragments

lazd:
## Description

This PR adds new API `Theme.registerComponentFragment(tagName, CSSResult)` that enables injecting CSS into a given component's ShadowRoot. This can enable theming, patching, and more beyond what the current `--mod-*` tokens API provides, and should be used alongside of it.

The API is very similar to `Theme.registerThemeFragment()`, but except for targeting the `sp-theme` element, it targets whatever SWC component you pass. Any component that uses the `SpectrumMixin` will automatically pick up registered component fragments, as long as they're registered before the component's `connectedCallback` is called for the first time. This is different than `registerThemeFragment`, which supports adding theme fragments at any time. We chose not to support adding component fragments after `connectedCallback` as we'd have to maintain a reference to each component instance in order to have them adopt new styles after render.

This approach is documented on this [internal wiki page](https://wiki.corp.adobe.com/display/horizon/Hz+Design+System) for Adobe employees that are interested in why we're proposing this.

## Related issue(s)

-

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices


packages/action-button/stories/index.ts
```diff
@@ -16,7 +16,8 @@ import '@spectrum-web-components/action-group/sp-action-group.js';
 import '@spectrum-web-components/icon/sp-icon.js';
 import '@spectrum-web-components/icons-workflow/icons/sp-icon-edit.js';
 
-import '@spectrum-web-components/action-button/sp-action-button.js';
+// Use dynamic imports otherwise the registered theme fragments doesn't run in time
+import('@spectrum-web-components/action-button/sp-action-button.js');
```
lazd, line 20: Since styles get finalized at class declaration time, we have to delay importing ActionButton at all until we've registered our theme fragments, otherwise they have no effect.

The idea is that users would first import `Theme` and call `Theme. registerComponentFragment(componentName, css)` for all of their component fragments before importing the actual components.

I'm a bit iffy on how that would actually work, more exploration is required.

tools/base/src/Base.ts
```diff
@@ -10,6 +10,7 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
+import { Theme } from '@spectrum-web-components/theme';
```
lazd, line 13: This breaks everything. I'm not sure what I'm doing wrong, but I get this error when running `yarn install`

```
tools/base/src/Base.ts:13:23 - error TS2307: Cannot find module '@spectrum-web-components/theme' or its corresponding type declarations.

13 import { Theme } from '@spectrum-web-components/theme';
                         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

tools/theme/package.json
```diff
@@ -133,8 +133,8 @@
         "lit-html"
     ],
     "dependencies": {
-        "@spectrum-web-components/base": "^0.40.3",
-        "@spectrum-web-components/styles": "^0.40.3"
+        "@spectrum-web-components/styles": "^0.40.3",
+        "lit": "^2.5.0"
```
lazd, line 137: I tried to remove circular dependencies by pulling things from `lit` directly.

tools/base/src/Base.ts
lazd: (it was `tsconfig.json`'s `references` array)

tools/base/src/Base.ts
```diff
@@ -126,6 +133,28 @@ export function SpectrumMixin<T extends Constructor<ReactiveElement>>(
             }
         }
 
+        protected adoptComponentFragments(): void {
+            const componentFragments = Theme.getComponentFragments(
+                this.constructor.name
```
moisadoru, line 138: This will not work with bundlers that mangle names.

tools/theme/src/Theme.ts
```diff
@@ -495,6 +494,23 @@ export class Theme extends HTMLElement implements ThemeKindProvider {
         }
     }
 
+    static registerComponentFragment(
+        name: ComponentName,
+        styles: CSSResultGroup
+    ): void {
```
moisadoru, line 500: To solve the potential issue with bundlers that mangle names, and to avoid accumulating styles in the `fragmentsByComponent` map, you could use a `WeakMap<HTMLElement, CSSResultGroup[]>` instead:

```typescript
    static fragmentsByComponent = new WeakMap<HTMLElement, CSSResultGroup[]>();

    // ...
    static private _getComponentFragments(component: HTMLElement): CSSResultGroup[] {
        let fragments = Theme.fragmentsByComponent.get(component);
        if (!Array.isArray(fragments)) {
          Theme.fragmentsByComponent.set(component, (fragments = []));
        }
        return fragments;
    }

    static registerComponentFragment(
        component: HTMLElement,
        styles: CSSResultGroup
    ): void {
        const fragments = Theme._getComponentFragments(component);
        fragments.push(styles);
```

tools/base/src/Base.ts
```diff
@@ -126,6 +133,28 @@ export function SpectrumMixin<T extends Constructor<ReactiveElement>>(
...
+                        );
+                    }
+                } else {
+                    componentFragments.forEach((s) => {
+                        const style = document.createElement('style');
+                        style.textContent = (s as CSSResult).cssText;
+                        this.shadowRoot.appendChild(style);
+                    });
+                }
+            }
```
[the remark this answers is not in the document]
lazd: Lit's `adoptStyles()` [overwrites the existing styles](https://github.com/lit/lit/blob/12109c25997ef03180d7eefe05c64e0fb20dd2b0/packages/reactive-element/src/css-tag.ts#L175-L177), we want to add to them.

tools/theme/src/Theme.ts
lazd: Hmmm, for this to work, we'd need an actual HTMLElement instance to pass to `registerComponentFragment`, so that wouldn't work. Component Fragments need to apply to all instances of a given component, hence using the constructor's name as a key.

However, we could switch this to `this.tagName` instead...

tools/base/src/Base.ts
lazd: Good catch, what if we use `this.tagName`?

lazd: I went ahead and switched it over to `tagName`, let me know your thoughts.

Westbrook:
We're currently discussing whether `Theme.registerThemeFragment()` (already available in production) as seen in https://studio.webcomponents.dev/edit/EYQn9T6wOnieZbv4xnPj/src/index.ts?branch=custom-system%40wfVpXdsTYhf9mY2slxW4t2Y2SjC3&p=stories is an equivalent or better path for supporting this situation.

Westbrook:
This isn't currently targeted for a merge, so closing for now. We can come back to this if it's fully needed in the future of the project.

PR #3972, opened 2024-01-29
docs(combobox): create initial documentation

hunterloftis:
## Description

First draft of combobox docs.

### DIFF
--- a/packages/combobox/README.md
+++ b/packages/combobox/README.md
@@ -26,12 +26,11 @@ import { Combobox } from '@spectrum-web-components/combobox';
 ## Sizes
 
 <sp-tabs selected="m" auto label="Size Attribute Options">
-
 <sp-tab value="s">Small</sp-tab>
 <sp-tab-panel value="s">
 
 ```html
-<sp-combobox label="Color">
+<sp-combobox size="s" label="Color">
     <sp-menu-item value="red">Red</sp-menu-item>
     <sp-menu-item value="green">Green</sp-menu-item>
     <sp-menu-item value="blue">Blue</sp-menu-item>
@@ -39,12 +38,11 @@ import { Combobox } from '@spectrum-web-components/combobox';
 ```
 
 </sp-tab-panel>
-
 <sp-tab value="m">Medium</sp-tab>
 <sp-tab-panel value="m">
 
 ```html
-<sp-combobox label="Color">
+<sp-combobox size="m" label="Color">
     <sp-menu-item value="red">Red</sp-menu-item>
     <sp-menu-item value="green">Green</sp-menu-item>
     <sp-menu-item value="blue">Blue</sp-menu-item>
@@ -52,12 +50,11 @@ import { Combobox } from '@spectrum-web-components/combobox';
 ```
 
 </sp-tab-panel>
-
 <sp-tab value="l">Large</sp-tab>
 <sp-tab-panel value="l">
 
 ```html
-<sp-combobox label="Color">
+<sp-combobox size="l" label="Color">
     <sp-menu-item value="red">Red</sp-menu-item>
     <sp-menu-item value="green">Green</sp-menu-item>
     <sp-menu-item value="blue">Blue</sp-menu-item>
@@ -65,12 +62,11 @@ import { Combobox } from '@spectrum-web-components/combobox';
 ```
 
 </sp-tab-panel>
-
 <sp-tab value="xl">Extra Large</sp-tab>
 <sp-tab-panel value="xl">
 
 ```html
-<sp-combobox label="Color">
+<sp-combobox size="xl" label="Color">
     <sp-menu-item value="red">Red</sp-menu-item>
     <sp-menu-item value="green">Green</sp-menu-item>
     <sp-menu-item value="blue">Blue</sp-menu-item>
@@ -78,12 +74,13 @@ import { Combobox } from '@spectrum-web-components/combobox';
 ```
 
 </sp-tab-panel>
-
 </sp-tabs>
 
 ## Labeling
 
-A combobox must be labeled. Either provide an accessible label via the `label` attribute, or render a visible label via `<sp-field-label>`:
+A combobox must be labeled.
+Typically, you should render a visible label via `<sp-field-label>`.
+For exceptional cases, provide an accessible label via the `label` attribute.
 
 ```html
 <sp-field-label for="color">Color</sp-field-label>
@@ -102,15 +99,15 @@ Combobox options are presented as a popup menu. Menu items can be provided via m
 
 Instead of providing `<sp-menu-item>` children, you can assign an array of `ComboboxOptions` to the `options` property, and `<sp-combobox>` will create matching menu items:
 
-```html
-<sp-combobox id="colors"></sp-combobox>
+```html-no-demo
+<sp-combobox id="color" label="Color"></sp-combobox>
 
 <script>
-    document.getElementById('colors').options = {[
-        { id: "red", value: "Red" },
-        { id: "green", value: "Green" },
-        { id: "blue", value: "Blue" }
-    ]};
+    document.getElementById('color').options = [
+        { value: "red", itemText: "Red" },
+        { value: "green", itemText: "Green" },
+        { value: "blue", itemText: "Blue" }
+    ];
 </script>
 ```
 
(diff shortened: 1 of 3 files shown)


packages/combobox/README.md
```diff
@@ -21,8 +23,73 @@ When looking to leverage the `Combobox` base class as a type and/or for extensio
...
+```html
+<sp-combobox id="colors"></sp-combobox>
+
+<script>
+    document.getElementById('colors').options = {[
+        { id: "red", value: "Red" },
+        { id: "green", value: "Green" },
+        { id: "blue", value: "Blue" }
+    ]};
+</script>
```
Westbrook, line 55: I don't think this will work. @najikahalsema was making some similar updates to UEC that might be applicable here.

This however, points out that I need to add this to the `bundle` package so that it is even available in the docs pages like this... I'll get a separate PR up for that.

najikahalsema: Yeah, you'd probably want to do something like 

```
```html-live
<sp-combobox id="colors"></sp-combobox>

<script type="module">
    const initCombobox = () => {
        const combobox = document.querySelector('#colors');
        const options = [
            { id: "red", value: "Red"}, ......
         ];
       combobox.options = options;
    };
    customElements.whenDefined('sp-combobox').then(() => {
        initCombobox();
    });
</script>
```
And then immediately below it (but outside of the html-live tag):
```
<script type="module">
    const initCombobox = async () => {
        const combobox = document.querySelector('#colors');
        await combobox.updateComplete;
        await combobox.closest('code-example')?.updateComplete;

        const options = [
            { id: "red", value: "Red"}, ......
         ];
       combobox.options = options;
    };
    customElements.whenDefined('code-example').then(() => {
        customElements.whenDefined('sp-combobox').then(() => {
            initCombobox();
        });
    });
</script>
```

najikahalsema (changes requested):
I agree with westbrook's notes. Also, some more examples might be good :) Hopefully you're able to get them working with that code snippet.

packages/combobox/README.md
```diff
@@ -21,8 +23,73 @@ When looking to leverage the `Combobox` base class as a type and/or for extensio
...
+The combobox supports both mouse and keyboard navigation. Mobile behavior is currently unspecified.
+
+When an `<sp-combobox>` is focused, pressing the down arrow moves focus to the first menu item in the popup menu.
+The up and down arrows then move between available menu items.
+
+The escape key dismisses the popup menu if open. Otherwise, it clears the combobox's textfield.
+
+The enter key sets the `value` of the focused `<sp-combobox>`.
+If the popup menu is open, the value is set to the `value` of the selected menu item,
+returning focus back to the combobox's textfield.
```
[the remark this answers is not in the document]
Rajdeepc: should we also add how to test in different accessibility modes?

packages/combobox/README.md
```diff
@@ -21,8 +23,73 @@ When looking to leverage the `Combobox` base class as a type and/or for extensio
...
 ```
+
+```js
+this.colorOptions = [...this.colorOptions, { id: 'purple', value: 'Purple' }];
+```
+
+## Autocomplete
+
+<!--- Should there be an option to make matching case-sensitive? -->
+<!--- Should there be an option to allow matching anywhere in a string rather than the beginning? -->
```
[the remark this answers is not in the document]
hunterloftis: Added!

packages/combobox/README.md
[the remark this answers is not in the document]
hunterloftis: I think so, since it details some of the edge cases that could be unclear (eg, does "enter" always lock in the typed value or does the menuitem value supercede when the menu is open? Does escape do anything when the menu is closed?)

If these facets are covered in unit tests, though, we could remove their human documentation and then a user can always find out by experimenting in storybook.

packages/combobox/README.md
hunterloftis: Is this a place where it deviates from the RFC? [Using an object](https://github.com/adobe/spectrum-web-components/discussions/3897) looks to be assigning an array to the `options` property, just via Lit syntax.

hunterloftis: Or do you just mean this won't work in the docs demo environment?

najikahalsema: It won't work in the demo environment... maybe. Unless you find that it does! In which case, never mind!

packages/combobox/README.md
```diff
@@ -23,25 +23,84 @@ When looking to leverage the `Combobox` base class as a type and/or for extensio
...
 
-### Options Array
+</sp-tab-panel>
 
-Instead of providing `<sp-menu-item>` children, you can assign an array of `ComboboxOptions` to the `options` property, and `<sp-combobox>` will create matching menu items:
+<sp-tab value="m">Medium</sp-tab>
+<sp-tab-panel value="m">
+
+```html
+<sp-combobox label="Color">
```
Westbrook, line 47: This also accepts that `size="m"` value across the examples.

packages/combobox/README.md
hunterloftis: Thanks for the example!

packages/combobox/src/Combobox.ts
```diff
@@ -255,6 +255,7 @@ export class Combobox extends Textfield {
 
     public toggleOpen(): void {
         this.open = !this.open;
+        this.inputElement.focus();
```
Westbrook, line 258: Does this cause any issues if a user clicks on a focusable element is IS NOT within the Combobox element?

packages/combobox/README.md
```diff
@@ -21,8 +23,136 @@ When looking to leverage the `Combobox` base class as a type and/or for extensio
...
+    <sp-menu-item value="blue">Blue</sp-menu-item>
+</sp-combobox>
+```
+
+</sp-tab-panel>
+</sp-tabs>
+
+## Labeling
+
+A combobox must be labeled.
```
Westbrook, line 81: Did we want this on an line of its own? If so, we need an extra space for the lines not to be merged when covered to HTML.

packages/combobox/README.md
```diff
@@ -21,8 +23,136 @@ When looking to leverage the `Combobox` base class as a type and/or for extensio
...
+        { value: "red", itemText: "Red" },
+        { value: "green", itemText: "Green" },
+        { value: "blue", itemText: "Blue" }
+    ];
+</script>
+```
+
+### Dynamic options
+
+When you update the `options` property or add/remove `<sp-menu-item>` children, the `<sp-combobox>` will detect that change and update its popup menu contents. For example, using [Lit](https://lit.dev/):
```
Westbrook, line 116: Is there a concise way to inform users that updating the `options` property must include providing a _new_ array? You do this in the example, maybe that's enough?

packages/combobox/README.md
```diff
@@ -21,8 +23,136 @@ When looking to leverage the `Combobox` base class as a type and/or for extensio
...
+`autocomplete="list"`
+
+The popup menu items are filtered to only those completing the currently-input value.
+
+## Focus and Accessibility
+
+The combobox supports both mouse and keyboard navigation. Mobile behavior is currently unspecified.
+
+When an `<sp-combobox>` is focused, pressing the down arrow moves focus to the first menu item in the popup menu.
+The up and down arrows then move between available menu items.
```
Westbrook, line 152: There's a couple of new lines in this section that will be merged into single `<p>` tags here, are we OK with that?

Westbrook (commented):
All minor things...

packages/combobox/README.md
hunterloftis: No, I just manually wrapped it in the MD file 😁

packages/combobox/README.md
hunterloftis: Yeah I'm just putting newlines for easier reading of the source MD. Happy to merge onto a single line if you prefer.

packages/combobox/README.md
```diff
@@ -21,8 +23,140 @@ When looking to leverage the `Combobox` base class as a type and/or for extensio
...
+        { value: "red", itemText: "Red" },
+        { value: "green", itemText: "Green" },
+        { value: "blue", itemText: "Blue" }
+    ];
+</script>
+```
+
+### Dynamic options
+
+When you replace the `options` Array with a new object or add/remove `<sp-menu-item>` children, the `<sp-combobox>` will detect that change and update its popup menu contents.
```
Westbrook, line 117:
```suggestion
When you replace the `options` Array with a new Array or add/remove `<sp-menu-item>` children, the `<sp-combobox>` will detect that change and update its popup menu contents.
```

Is it confusing to replace an Array with an object?

hunterloftis: Just removed the "with a new object" part, left "replace the Array"

Westbrook (approved):
LGTM!

PR #3988, opened 2024-01-31
docs(combobox): add story demonstrating controlled-component usage

hunterloftis:
This verifies that the value of `<sp-combobox>` can be controlled by its container, and demonstrates how.

### DIFF
--- a/packages/combobox/stories/combobox.stories.ts
+++ b/packages/combobox/stories/combobox.stories.ts
@@ -10,7 +10,11 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import { html, TemplateResult } from '@spectrum-web-components/base';
+import {
+    html,
+    LitElement,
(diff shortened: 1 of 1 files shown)


packages/combobox/stories/combobox.stories.ts
```diff
@@ -145,3 +153,47 @@ export const withHelpText = (): TemplateResult => {
...
+        };
+    }
+}
+defineElement('controlled-combo', ControlledCombo);
+
+export const controlled = (): TemplateResult => {
+    return html`
+        <controlled-combo></controlled-combo>
+    `;
+};
```
Westbrook, line 199:
```suggestion
};
controlled.swc_vrt = {
    skip: true,
};
```
There's not additional VRT data here, so skipping should be fine and will get CI green again.

Westbrook (approved):
LGTM! If we had a "new component check list" ensuring we have a "controlled component demo" should be on it 😅

Issue #2065, opened 2022-01-21
[Feat]: Default-values for slider / sider-handle


spdev3000:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Slider, SliderHandle

### Description of the requested feature

We'd like to have a feature implemented in Slider or SliderHandle to provide a `default-value`. If the user would then double click on the slider or the handle the value would reset to the provided default-value.

### Implementation notes or ideas

I tested also detecting a double click on the slider track. This might be possible to let the user even reset a slider when not double clicking on the slider handle.
We need to consider resetting values also for keyboard-users. Maybe we can use Escape key when the handle has focus to reset to default value (if given).

jnurthen:
Please also consider touch users, screen reader users and voice recognition users when thinking about how this should work.

hunterloftis:
@jnurthen do you have concerns about accessibility with the core [spec](https://spectrum.adobe.com/page/slider/#Double-click-to-reset)?

If so, is there an alternative input to a double-click that you would recommend replacing the Spectrum behavior with?

jnurthen:
I have concerns - it doesn't appear to take account of any input mechanisms other than pointers

jnurthen:
I filed https://jira.corp.adobe.com/browse/SDS-12156

hunterloftis:
Thanks @jnurthen!

najikahalsema:
related to #586

Westbrook:
Erroneously closed.

Rajdeepc:
Needs some testing on touch users, screen reader users and voice recognition

PR #3991, opened 2024-02-01
fix(slider): double click on slider handle to reset slider position

Rajdeepc:
## Description

Double click on slider handle to reset slider position

## Related issue(s)

- https://github.com/adobe/spectrum-web-components/issues/2065

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

https://github.com/adobe/spectrum-web-components/assets/8790510/f1a8432c-275e-4e12-9d4d-2a13d9c87059

## Types of changes

## Checklist

## Best practices


Rajdeepc:
@Westbrook Slider Handle UI is updated according to S1 designs
Screenshot 2024-02-14 at 3 48 28 PM
Screenshot 2024-02-14 at 3 48 34 PM

packages/slider/src/HandleController.ts
```diff
@@ -88,6 +88,21 @@ export class HandleController {
         }
         return result;
     }
+    private clickTimer: number | NodeJS.Timeout | null = null;
+
+    public get isDoubleClick(): boolean {
```
Westbrook, line 93: https://developer.mozilla.org/en-US/docs/Web/API/Element/dblclick_event

Westbrook (commented):
There is currently no way of knowing that this interaction exists. Further, it's not clear how to trigger it when interacting with a keyboard or with a screen reader.

packages/slider/src/HandleController.ts
```diff
@@ -88,6 +88,21 @@ export class HandleController {
...
+        if (this.clickTimer !== null) {
+            clearTimeout(this.clickTimer);
+            this.clickTimer = null;
+            return true;
+        }
+
+        this.clickTimer = setTimeout(() => {
+            this.clickTimer = null;
+        }, 300);
```
spdev3000, line 102: Just a note: PsWeb is using 500ms (as defined by Microsoft) to detect a double click:
see https://docs.microsoft.com/en-us/windows/win32/controls/ttm-setdelaytime?redirectedfrom=MSDN

spdev3000:
Did you try listen to `dblclick` event to avoid timers here?

Rajdeepc:
> Did you try listen to `dblclick` event to avoid timers here?

Yes thats the way to go forward to include on all the input modalities. The code should now listen to a dblClick event instead of handling it with a timer function. Thanks for pointing it out.

TarunAdobe:
@spdev3000 I've updated the slider to listen to a dblclick and reset the value accordingly. On a side note I am resetting the value of the slider on pressing enter key (once)... Let me know if this makes sense or is there some other way we want to deliver this.
cc: @Westbrook

spdev3000:
lots of good stuff here.
Can we add a storybook use case for providing a defaultValue via slider attribute?
Also can we add a test, that compares the (provided from outside) defaultValue with the value after double click?

Regarding 
```
On a side note I am resetting the value of the slider on pressing enter key (once)
```
Do you mean resetting when the slider handle has been focused and the user then press Enter key? Sounds ok to me for the use case, when editable=false (aka NumberField isn't visible).

packages/slider/src/HandleController.ts
```diff
@@ -432,7 +445,15 @@ export class HandleController {
         this.requestUpdate();
     };
 
-    private onInputKeydown = (event: Event): void => {
+    private onInputKeydown = (event: KeyboardEvent): void => {
+        if (event.key == 'Enter') {
```
Westbrook, line 449: The `Enter` key, when pressed while focusing a form field, should trigger the `change` event on that form field, not reset it. The accessibility team spoke of possibly accepting `Escape` as a resetting key, but we should confirm with this this behavior and any associated messaging that should be surfaced by the element to inform screen reader users that this functionality is available. Further, we should then confirm how the use of `Escape` in that context may or may not interfere with pressing `Escape` while in a modal, at which point the `Escape` key would be expected to close the modal...

TarunAdobe: Thank you for this feedback... Super helpful. I will switch the `Enter` key with `Escape`.
1. For the accessibility part I'm thinking if we can have an `aria-describedby` on the input that points to the description of this specific functionality?
2. And I verified the current code in case of an modal and currently on pressing `escape` the modal does close however the value also resets to default. Is this expected or do we want to not reset the value and just close the modal if the slider is in a modal?

spdev3000:
If it's not clear whether Escape or Enter or any other key might be used I vote for skipping it now and update later.
If we decide to use Escape key - is it possible to stop bubbling the event when it has been detected and default value has been set? Otherwise we may do so from the outside of slider component.

TarunAdobe:
@spdev3000  I agree to you we can think of skipping it if we don't get the right direction from the accessibility team (I've dropped a message to start some discussion on this). 
And yes I think it is possible and we can stop the event from bubbling up but I'm not yet convinced if this is the right path. 
On a side note I was thinking about this - 
```Can we add a storybook use case for providing a defaultValue via slider attribute?```
Currently we set the default value of the slider based on its initial value, i.e. we don't provide consumers a way to set the default value through an attribute in this implementation. I can be convinced to have that through attribute if you help my understand how that might bean  important addition.

**Comment by spdev3000** [comment]
looks good to me so far - one question:
If the slider is part of an overlaid popover or modal dialog and we want to leverage a default-value:
If the user clicks and drags the slider handle and then presses `Escape` key to close/dismiss the outer popover overlay or the modal dialog - the slider would first reset its value and then the overlay would close, correct?
If so: any idea how to let the user only close the overlay (containing the slider) and not reset the default value?
Should we add another property to control/disable that (Escape=reset) behavior for such use cases?

**Comment by TarunAdobe** [comment]
Valid question @spdev3000. You are right if the user drags the slider that is contained inside of an overlaid content and then presses `escape` the slider value will reset instead of closing the modal... This interaction pattern was suggested by someone from the accessibility team. 
And for your second question: I don't think it's possible to avoid reseting the value of the slider in case the slider has a `default-value` attribute and the user presses `escape` key. As for having a separate attribute/property to control this behaviour, It's possible to have it but we need a few more eyes on this... Maybe @Westbrook and @Rajdeepc  can help us decide the correct path forward.

File: packages/slider/src/HandleController.ts [hunk 4]
```diff
@@ -341,6 +341,19 @@ export class HandleController {
 
     private _activePointerEventData!: DataFromPointerEvent | undefined;
 
+    /**
+     * @description check for defaultvalue(value) property in sp-slider and reset on double click on sliderHandle
+     * @param event
+     */
+    public handleDoubleClick(event: PointerEvent): void {
+        const { input } = this.extractDataFromEvent(event);
+        if (input.model?.handle.defaultValue) {
```
Westbrook, line 350: If the default value in `0` then this test will not work appropriately.

packages/slider/stories/slider.stories.ts
```diff
@@ -152,6 +152,26 @@ export const Filled = (args: StoryArgs = {}): TemplateResult => {
...
 
+export const HasADefaultValue = (args: StoryArgs = {}): TemplateResult => {
+    return html`
+        <div style="width: 500px; margin-inline: 20px;">
+            <sp-slider
+                max="1"
+                min="0"
+                value=".5"
+                step="0.01"
+                default-value="0.2"
```
Westbrook, line 163: Might be useful to put `default-value` into the `args`, but not a blocker.

packages/slider/test/slider.test.ts
```diff
@@ -1604,4 +1603,75 @@ describe('Slider', () => {
         await elementUpdated(el);
         expect(el.values).to.deep.equal({ a: 10, b: 20, c: 29 });
     });
+    it('resets to default value on double click after moving pointer', async () => {
```
Westbrook, line 1606: We should add a test where the Slider is in an Overlay to confirm it does not close the Overlay.

We should also add `default-value` to one or some of the Stories in Overlay to confirm this functionality there, as well.

packages/slider/test/slider.test.ts
```diff
@@ -1604,4 +1603,75 @@ describe('Slider', () => {
...
+                    default-value="50"
+                ></sp-slider>
+            `
+        );
+        await elementUpdated(el);
+        expect(el.value, 'initial').to.equal(50);
+        expect(pointerId).to.equal(-1);
+        const handle = el.shadowRoot.querySelector('.handle') as HTMLDivElement;
+        el.track.setPointerCapture = (id: number) => (pointerId = id);
+        el.track.releasePointerCapture = (id: number) => (pointerId = id);
```
Westbrook, line 1622: If we use `sendMouse()` all the way through the test here, instead of synthetically dispatching events. Then we should be able to exclude this part of the test all together. Here we are mimicking the default pointerCapture functionality, which the actual pointer interactions in `sendMouse()` do not need.

packages/slider/src/Slider.ts
```diff
@@ -548,4 +554,8 @@ export class Slider extends SizedMixin(ObserveSlotText(SliderHandle, ''), {
             }
         }
     }
+
+    protected override firstUpdated(changes: PropertyValues<this>): void {
+        super.firstUpdated(changes);
+    }
```
Westbrook, line 560: If we're not actually doing work here than empty additions to the call stack are just costing us tiny amounts of performance.

packages/slider/src/Slider.ts
```diff
@@ -442,6 +443,7 @@ export class Slider extends SizedMixin(ObserveSlotText(SliderHandle, ''), {
                         ['pointerup', 'pointercancel', 'pointerleave'],
                         this.handlePointerup,
                     ],
+                    streamOutside: ['dblclick', this.handleDoubleClick],
```
Westbrook, line 446: Half surprised this works, but it does make a nice even API for this.

packages/slider/src/Slider.ts
```diff
@@ -409,10 +408,12 @@ export class Slider extends SizedMixin(ObserveSlotText(SliderHandle, ''), {
             ></div>
         `;
     }
-
     private renderHandle(): TemplateResult {
+        if (this.variant === 'tick') {
```
Westbrook, line 412: 👍🏼

packages/slider/src/HandleController.ts
```diff
@@ -522,12 +548,17 @@ export class HandleController {
...
                     aria-labelledby=${ariaLabelledBy}
                     aria-valuetext=${this.formattedValueForHandle(model)}
+                    aria-describedby="sliderDesc"
                     @change=${this.onInputChange}
                     @focus=${this.onInputFocus}
                     @blur=${this.onInputBlur}
                     @keydown=${this.onInputKeydown}
                     .model=${model}
                 />
+                <p id="sliderDesc" class="visually-hidden">
```
Westbrook, line 558: This should be OK as just `display: none` as it is `describedby` content. Otherwise you get multiple stops of the screen reader cursor. This should be confirmed manually after the change, preferably in [assistivlabs](https://assistivlabs.com/) across more than one screen reader, as also in a unit test using accessibility snapshotting.

Probably OK to make it a `span` over a `p` as it's not visible, too.

packages/slider/src/HandleController.ts
```diff
@@ -432,7 +445,20 @@ export class HandleController {
         this.requestUpdate();
     };
 
-    private onInputKeydown = (event: Event): void => {
+    private onInputKeydown = (event: KeyboardEvent): void => {
+        if (event.key == 'Escape') {
+            const input = event.target as InputWithModel;
+            if (
+                input.model.handle?.defaultValue &&
```
Westbrook, line 452: This will also fail on `0`.

packages/slider/src/HandleController.ts
```diff
@@ -522,12 +548,17 @@ export class HandleController {
...
                     aria-valuetext=${this.formattedValueForHandle(model)}
+                    aria-describedby="sliderDesc"
                     @change=${this.onInputChange}
                     @focus=${this.onInputFocus}
                     @blur=${this.onInputBlur}
                     @keydown=${this.onInputKeydown}
                     .model=${model}
                 />
+                <p id="sliderDesc" class="visually-hidden">
+                    Press escape or double click to reset the slider to its
```
Westbrook, line 559: Please add an entry for this content to https://github.com/adobe/spectrum-web-components/issues/1975

spdev3000:
As another requirement: the double click to `default-value` use case should also work for editable sliders, as most of our use cases have editable versions and need that feature working there as well.

packages/slider/test/slider.test.ts
Westbrook: Ping.

packages/slider/test/slider.test.ts
```diff
@@ -1604,4 +1607,128 @@ describe('Slider', () => {
...
+                    </sp-overlay>
+                </div>
+            `
+        );
+
+        await elementUpdated(el);
+
+        // open the overlay
+        const trigger = el.querySelector('#trigger') as HTMLButtonElement;
+        trigger.click();
```
Westbrook, line 1701: Overlay open action should follow the pattern:

```
const opened = oneEvent(el, 'sp-opened');
trigger.click(); // or similiar
await opened;
```
Otherwise there's no guarantee the content actually opened and was promoted to the top layer.

packages/slider/test/slider.test.ts
```diff
@@ -1604,4 +1607,128 @@ describe('Slider', () => {
...
+        expect(slider.value).to.equal(50);
+
+        // and the overlay should be in open state
+        const overlay = el.querySelector('sp-overlay') as Overlay;
+        expect(overlay.open).to.be.true;
+
+        // send escape key again
+        await sendKeys({
+            press: 'Escape',
+        });
```
Westbrook, line 1727: Needs to reverse of the above, e.g. `const closed = oneEvent(...);`

packages/slider/src/HandleController.ts
```diff
@@ -341,6 +341,22 @@ export class HandleController {
...
+     * @param event
+     */
+    public handleDoubleClick(event: PointerEvent): void {
+        const input = (event.target as Element).querySelector(
+            '.input'
+        ) as InputWithModel;
+
+        if (input.model?.handle.defaultValue !== undefined) {
+            input.model.handle.value = input.model.handle.defaultValue;
+            this.dispatchChangeEvent(input, input.model.handle);
```
Westbrook, line 355: Confirm whether this dispatches an `input` event with the same data. See the result of clicking a new value in an `<input type="range">` as an example: https://codepen.io/Westbrook/pen/ExJjVRZ?editors=1111

Westbrook (commented):
Thanks for updating the demos and adding the test, this is looking much closer to being ready to ship. A couple of small nuances and then we're ready to go!

packages/slider/test/slider.test.ts
TarunAdobe: got it!

packages/slider/src/HandleController.ts
TarunAdobe: added  `dispatchInputEvent` on both double-click and escape key interactions

Westbrook: What's a good test to confirm that this happens before we close this up?

TarunAdobe: would it make sense if I extend the `dispatches input of the animation frame` test (line 388) or should I make a separate one?

Westbrook (approved):
LGTM!

Catching this up to `main` and if CI stays green, I'll merge this in a bit.

PR #4006, opened 2024-02-05
chore: take latest CSS dependencies

### LINKED ISSUE #2598: [Bug]: We need a way to disable hover state for all spectrum buttons on mobile devices
### Code of conduct

### Impacted component(s)

ProjectX

### Expected behavior

There's a lingering hover effect on the mobile browser with button-related components such as `sp-menu-item`, `sp-action-button`, and `sp-button`; this is causing confusion for users.

### Actual behavior

SWC should have a way to detect on mobile devices and disabled the hover effects

### Screenshots

PixelSnap 2022-09-19 at 09 28 57

https://user-images.githubusercontent.com/4380525/191066654-0b550d62-9ba3-4688-a2dd-c7c7f72e2d03.MP4

Safari

1. Go to ProjectX 
2. Click on Plus menu
3. Scroll to through the list of menu items
4. Observe there are some items with linger hover states

_No response_

_No response_

Westbrook:
## Description
Take CSS update the sequesters `:hover` styles within the `@media (hover: hover) {}` media query, which prevents it in the _majority_ of mobile/touch-only contexts.

## Related issue(s)
- fixes #2598

## How has this been tested?
-   [ ] _Test case 1_
    1. Go [here](https://spectrum-css--spectrum-web-components.netlify.app/storybook/index.html?path=/story/menu--default)
    2. In a mobile or touch-only context (can be triggered with Chrome Dev Tools)
    3. Click on of the Menu Items
    4. See that the "hover" state is not triggered
-  [ ] _Test case 2_
    1. Go [here](https://73874bdaecee6ea6d7a506b37d5d256d--spectrum-web-components.netlify.app/review/)
    2. See the accepted VRT updates

## Types of changes

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ executors:
 parameters:
     current_golden_images_hash:
         type: string
-        default: 072516e18cdabf5c159ea72b6cfcc8c7f95843ed
+        default: 15e9cb73f536a5f74d7978a8ae3fdef174935f6f
     wireit_cache_name:
         type: string
         default: wireit
--- a/packages/accordion/package.json
+++ b/packages/accordion/package.json
@@ -73,7 +73,7 @@
         "@spectrum-web-components/shared": "^0.40.4"
     },
     "devDependencies": {
-        "@spectrum-css/accordion": "^4.1.1"
+        "@spectrum-css/accordion": "^4.2.0"
     },
     "types": "./src/index.d.ts",
     "customElements": "custom-elements.json",
--- a/packages/accordion/src/spectrum-accordion-item.css
+++ b/packages/accordion/src/spectrum-accordion-item.css
@@ -213,22 +213,6 @@ governing permissions and limitations under the License.
         var(--spectrum-accordion-item-header-color-default)
     );
 }
-#header:hover {
-    background-color: var(
-        --mod-accordion-background-color-hover,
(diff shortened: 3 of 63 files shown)


hunterloftis:
This fixes the referenced issue, in that lingering hover states no longer appear in the mobile context. But I would expect *some* feedback about the touch events (like the hover state, but applied only during the touch). Instead there's no feedback, so it feels like my input is being ignored:


https://github.com/adobe/spectrum-web-components/assets/364501/20b0deac-c415-49c1-ac6c-d340e32817ab

Westbrook:
> it feels like my input is being ignored:

Seems like good feedback for the design specification, but would also the actual "click" event doing something in the application be an acceptable proxy to that?

hunterloftis:
I think if the item isn't disabled, then it should inherently provide feedback (like a button - even if it's not hooked up to application state changes, you can tell that your input is being received by its built-in feedback).

I'll open an issue on the design side & approve this, since it is a fix for the bug & moves closer to the "right" state.

hunterloftis (dismissed):

Westbrook:
_Actually_, I found a piece of CSS that we weren't processing. Pushing now, and I think it covers what you're looking for here.

Westbrook:
This _only_ addresses the issue in Menu Item, so I've opened #4007 to follow up in other patterns.

PR #4020, opened 2024-02-08
fix(underlay): dispatch a "close" event rather than relying naively on "click"

Westbrook:
## Description
Relying on `click` directly means that patterns that open an Overlay on `pointerdown`, like `<sp-picker>`, could inadvertently close the Overlay as part of that initial interaction.

## How has this been tested?
-   [ ] _Test case 1_
    1. Go [here](https://picker-mobile--spectrum-web-components.netlify.app/)
    2. Use a mobile device, emulator, or Dev Tools to mock the mobile context
    3. Open one of the theme Pickers
    4. See that it stays open.

## Types of changes

## Checklist

### DIFF
--- a/packages/underlay/src/Underlay.ts
+++ b/packages/underlay/src/Underlay.ts
@@ -22,16 +22,40 @@ import styles from './underlay.css.js';
 
 /**
  * @element sp-underlay
+ *
+ * @fires close - When the underlay is "clicked" and the consuming pattern should chose whether to close based on that interaction
  */
 export class Underlay extends SpectrumElement {
     public static override get styles(): CSSResultArray {
         return [styles];
     }
 
+    private canClick = false;
+
     @property({ type: Boolean, reflect: true })
     public open = false;
 
+    public override click(): void {
+        this.dispatchEvent(new Event('close'));
+    }
+
+    protected handlePointerdown(): void {
+        this.canClick = true;
+    }
+
+    protected handlePointerup(): void {
+        if (this.canClick) {
+            this.click();
+        }
+        this.canClick = false;
+    }
+
     protected override render(): TemplateResult {
         return html``;
+    }
+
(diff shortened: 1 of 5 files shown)


hunterloftis:
I'm just using emulated mobile behaviors in Chrome, which isn't always 100% accurate, but the picker doesn't stay open in this case:


https://github.com/adobe/spectrum-web-components/assets/364501/3a5a5344-b07b-4032-becb-783377335b98

Westbrook:
@hunterloftis that is NOT the experience that I was addressing, but I'll dig into that, too! 😳

In a mobile context, so smaller window but still touch, a Tray will be opened with an Underlay, instead of the Popover. That is hopefully addressed by this change. Will report back when I know more about touch more broadly.

packages/underlay/src/Underlay.ts
```diff
@@ -22,16 +22,40 @@ import styles from './underlay.css.js';
...
+        this.dispatchEvent(new Event('close'));
+    }
+
+    protected handlePointerdown(): void {
+        this.canClick = true;
+    }
+
+    protected handlePointerup(): void {
+        if (this.canClick) {
+            this.dispatchEvent(new Event('close'));
```
hunterloftis, line 48: Another option here could be `this.click()` - today the lines are exactly the same, but sending both paths through the same method would guarantee that the behavior of a user-driven "click" and a code-driven "click" stay the same.

hunterloftis (approved):
Ah I see, yep repro'd both the existing & the fixed behavior.

packages/underlay/src/Underlay.ts
Westbrook: Yes!

najikahalsema (approved):
Reproduced successfully on my iPhone 👍 Looks good.

PR #4027, opened 2024-02-12
ci: expand lighthouse usage and reporting

Westbrook:
## Description
- build the test URL, rather than extract it from other functionality
- add a comparison to `main`
- [x] report the results into the PR
- [x] possibly compare to `latest`
- go from there

## How has this been tested?
-   [ ] _Test case 1_
    1. At some point the lighthouse results pre/post a PR should display in a comment below...

## Types of changes

## Checklist

### DIFF
--- a/.github/lighthouse/lighthouse-config.js
+++ b/.github/lighthouse/lighthouse-config.js
@@ -2,12 +2,11 @@ export default {
     extends: 'lighthouse:default',
     settings: {
         onlyCategories: [
-            // Excluded categories do not currently pass
-            // "accessibility",
-            // "best-practices",
+            'accessibility',
+            'best-practices',
             'performance',
-            // "pwa",
-            // "seo"
+            'pwa',
+            'seo',
         ],
         // Excluded audits do not currently pass
         // metrics/first-contentful-paint', 'metrics/largest-contentful-paint','metrics/first-meaningful-paint', 'metrics/speed-index', 'screenshot-thumbnails',
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -41,7 +41,7 @@ jobs:
                         repo: context.repo.repo,
                         issue_number: context.issue.number,
                       }).then(({data}) => {
-                        const priorComment = data.find(comment => comment.body.startsWith('# Branch Preview'));
+                        const priorComment = data.find(comment => comment.body.startsWith('## Branch preview') || comment.body.startsWith('# Branch Preview'));
                         if (priorComment) {
                           github.rest.issues.updateComment({
                             owner: context.repo.owner,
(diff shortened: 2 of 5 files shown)


.github/lighthouse/lighthouse-config.js
```diff
@@ -2,12 +2,11 @@ export default {
     extends: 'lighthouse:default',
     settings: {
         onlyCategories: [
-            // Excluded categories do not currently pass
-            // "accessibility",
-            // "best-practices",
+            'accessibility',
```
Westbrook, line 5: Turning on the categories here does not seem to turn on the audits. This way we can report their scores to the PR.

.github/workflows/test.yml
```diff
@@ -41,7 +41,7 @@ jobs:
                         repo: context.repo.repo,
                         issue_number: context.issue.number,
                       }).then(({data}) => {
-                        const priorComment = data.find(comment => comment.body.startsWith('# Branch Preview'));
+                        const priorComment = data.find(comment => comment.body.startsWith('## Branch preview') || comment.body.startsWith('# Branch Preview'));
```
Westbrook, line 44: This messages out the older version from existing PRs by replacing them or the new versions with the updated comment.

.github/workflows/test.yml
```diff
@@ -78,27 +78,52 @@ jobs:
...
-                  script: |
-                      const { buildPreviewURLComment } = await import('${{ github.workspace }}/tasks/build-preview-urls-comment.js');
-                      const body = buildPreviewURLComment(process.env.GITHUB_HEAD_REF);
-                      const urlRegex = /(https?:\/\/[^\s]+)/g;
-
-                      const docsURL = body.match(urlRegex)?.[0];
-                      core.setOutput('docsURL', docsURL);
+            - name: Extract branch name
+              shell: bash
+              run: echo "branch=${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}" >> $GITHUB_OUTPUT
```
Westbrook, line 83: This still feels like too much for for GitHub to be able to tell us the branch, but it's less than reaching into an unrelated script.

.github/workflows/test.yml
```diff
@@ -78,27 +78,52 @@ jobs:
...
-                  uploadArtifacts: true
+                  runs: 3
                   temporaryPublicStorage: true
+                  uploadArtifacts: true
+                  urls: |
+                      https://opensource.adobe.com/spectrum-web-components/
+                      https://main--spectrum-web-components.netlify.app/
+                      https://${{ steps.extract_branch.outputs.branch }}--spectrum-web-components.netlify.app/
+
+            - name: Post Results
```
Westbrook, line 99: This is the third time we've using this style of job in the GH CI, totally worth abstracting, but 100% not something I'm going to spend time on at this juncture.

.github/workflows/test.yml
```diff
@@ -78,27 +78,52 @@ jobs:
...
+              uses: actions/github-script@v6
+              with:
+                  script: |
+                      const { buildLighthouseComment } = await import('${{ github.workspace }}/tasks/build-lighthouse-comment.js');
+                      const body = buildLighthouseComment(${{ steps.lighthouse.outputs.links }}, ${{ steps.lighthouse.outputs.manifest }});
+                      github.rest.issues.listComments({
```
hunterloftis, line 105: I'm not sure what context this runs in... would `const data = await github.rest.issues.listComments(...)` work?

hunterloftis (approved):

.github/workflows/test.yml
Westbrook: It's no longer recent, but with the post initial creation move to `await import()` above, we should have access to that here and be able to flatten out the asynchrony here. I'll put this and the abstraction process into an issue to track and clarify steps. Great call!

Issue #3987, opened 2024-01-31
[Feat]: Tabs: Automatically bring the selected tab into view


Rocss:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-tabs

### Description of the requested feature

Given that I have the multiple tabs, which are wrapped in `sp-tabs-overflow`, and I already have a selected tab when the component is rendered, if the selected tab is not into view, it is confusing for users to understand what tab is selected.
You have to use the right arrow to scroll and discover the selected tab.

Introducing an optional attribute which scrolls to the selected tab would improve the user experience in use cases with small viewports, or large number of tabs.

### Implementation notes or ideas

`sp-tabs` component already knows what is the selected tab. The implementation would modify the `willUpdate` method (https://github.com/adobe/spectrum-web-components/blob/main/packages/tabs/src/Tabs.ts#L305) to verify if the selection is into view. If the child is not into view, scroll it into view (either at the center, or the start/end of the `sp-tabs-overflow` view space).

Westbrook:
Questions:
- Should `<sp-tabs-overflow>` _always_ "autoscroll"?
- Is it possible to leverage https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView so that we don't need to use `offsetLeft` and `clientWidth` which require a paint to leverage?
- If we use `scrollIntoView()` is it possible to move the majority (or all) of this work into `<sp-tabs-overflow>`?
- In theory, we could move the work from `_waitForChildrenToUpdate` into `getUpdateComplete` and support both this use case with a simple `await tabs.updateComplete` and remove the need to override `manageAutoFocus` to begin with.

Rocss:
> Should `<sp-tabs-overflow>` _always_ "autoscroll"?

I don't have a strong opinion on this. As a user I do prefer for it to scroll, as I can view the tab that's selected without me needing to do anything.
I do think it should be at least optional and if it were to be optional, the name can also be improved into something that would tell you it is only _initially scrolling_. What do you think?

> Is it possible to leverage https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView so that we don't need to use `offsetLeft` and `clientWidth` which require a paint to leverage?

[...]

PR #4032, opened 2024-02-13
feat(tabs): adds optional property to bring selected tab into view when first rendered

Rocss:
## Description

 If the default selected tab is out of view, the tabs overflow component does not ensure that tab is scrolled into view.

## Related issue(s)

See https://github.com/adobe/spectrum-web-components/issues/3987

## Motivation and context

As a user, I can easily see what tab is selected without having to scroll to discover it.

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

https://github.com/adobe/spectrum-web-components/assets/13311865/3de0a2be-07af-4e55-8da9-5f3355aec6b3

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/tabs/src/Tabs.ts
+++ b/packages/tabs/src/Tabs.ts
@@ -229,7 +229,9 @@ export class Tabs extends SizedMixin(Focusable, { noDefaultSize: true }) {
         return {};
     }
 
-    private _waitForChildrenToUpdate(): Promise<boolean[]> {
+    override async getUpdateComplete(): Promise<boolean> {
+        const complete = await super.getUpdateComplete();
+
         const tabs = [...this.children] as Tab[];
         const tabUpdateCompletes = tabs.map((tab) => {
             if (typeof tab.updateComplete !== 'undefined') {
@@ -237,30 +239,9 @@ export class Tabs extends SizedMixin(Focusable, { noDefaultSize: true }) {
             }
             return Promise.resolve(true);
         });
-        return Promise.all(tabUpdateCompletes);
-    }
 
-    protected override manageAutoFocus(): void {
-        this._waitForChildrenToUpdate().then(() => super.manageAutoFocus());
-    }
-
-    public scrollToSelected(): void {
-        this._waitForChildrenToUpdate().then(() => {
-            const selectedTab = this.querySelector(
-                `[role="tab"][value="${this.selected}"]`
-            ) as Tab;
-
-            // only if the selected tab is not visible
-            if (selectedTab && selectedTab.offsetLeft > this.clientWidth) {
-                // scroll it to the center of the container
-                this.scrollTabs(
-                    selectedTab.offsetLeft -
-                        this.clientWidth / 2 +
-                        selectedTab.clientWidth / 2,
-                    'instant'
-                );
-            }
-        });
+        await Promise.all(tabUpdateCompletes);
+        return complete;
     }
 
     protected managePanels({
--- a/packages/tabs/src/Tabs.ts
+++ b/packages/tabs/src/Tabs.ts
@@ -229,7 +229,7 @@ export class Tabs extends SizedMixin(Focusable, { noDefaultSize: true }) {
         return {};
     }
 
-    protected override manageAutoFocus(): void {
+    private _waitForChildrenToUpdate(): Promise<boolean[]> {
         const tabs = [...this.children] as Tab[];
         const tabUpdateCompletes = tabs.map((tab) => {
             if (typeof tab.updateComplete !== 'undefined') {
@@ -237,7 +237,30 @@ export class Tabs extends SizedMixin(Focusable, { noDefaultSize: true }) {
             }
             return Promise.resolve(true);
         });
-        Promise.all(tabUpdateCompletes).then(() => super.manageAutoFocus());
+        return Promise.all(tabUpdateCompletes);
+    }
+
+    protected override manageAutoFocus(): void {
+        this._waitForChildrenToUpdate().then(() => super.manageAutoFocus());
+    }
+
+    public scrollToSelected(): void {
+        this._waitForChildrenToUpdate().then(() => {
+            const selectedTab = this.querySelector(
+                `[role="tab"][value="${this.selected}"]`
+            ) as Tab;
+
+            // only if the selected tab is not visible
+            if (selectedTab && selectedTab.offsetLeft > this.clientWidth) {
+                // scroll it to the center of the container
+                this.scrollTabs(
+                    selectedTab.offsetLeft -
+                        this.clientWidth / 2 +
+                        selectedTab.clientWidth / 2,
+                    'instant'
+                );
+            }
+        });
     }
 
     protected managePanels({
--- a/packages/tabs/src/TabsOverflow.ts
+++ b/packages/tabs/src/TabsOverflow.ts
@@ -27,6 +27,7 @@ import {
 import { classMap } from '@spectrum-web-components/base/src/directives.js';
 import { ResizeController } from '@lit-labs/observers/resize-controller.js';
 import { Tabs } from './Tabs.js';
+import { Tab } from './Tab.js';
 import '@spectrum-web-components/action-button/sp-action-button.js';
 import '@spectrum-web-components/icons-ui/icons/sp-icon-chevron100.js';
 import chevronIconStyles from '@spectrum-web-components/icon/src/spectrum-icon-chevron.css.js';
@@ -78,9 +79,7 @@ export class TabsOverflow extends SizedMixin(SpectrumElement) {
         });
     }
 
-    protected override async firstUpdated(
-        changes: PropertyValues
-    ): Promise<void> {
+    protected override firstUpdated(changes: PropertyValues): void {
         super.firstUpdated(changes);
         // enable scroll event
         const [tabs] = this.scrollContent;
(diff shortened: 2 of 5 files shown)


packages/tabs/src/TabsOverflow.ts
```diff
@@ -27,6 +27,7 @@ import {
 import { classMap } from '@spectrum-web-components/base/src/directives.js';
 import { ResizeController } from '@lit-labs/observers/resize-controller.js';
 import { Tabs } from './Tabs.js';
+import { Tab } from './Tab.js';
```
Westbrook, line 30:
```suggestion
import type { Tabs } from './Tabs.js';
import type { Tab } from './Tab.js';
```
I _think_ you can do this for `Tabs`, too, but if not, I retract my statement. Having these at `type` imports isolate the code inclusion to the TS context and _in some cases_ can reduce the final about of JS passed to the browser.

packages/tabs/src/TabsOverflow.ts
```diff
@@ -83,6 +90,15 @@ export class TabsOverflow extends SizedMixin(SpectrumElement) {
             tabs.enableTabsScroll = true;
         }
         this.resizeController.observe(this.overflowContainer);
+
+        if (this.autoscroll && tabs?.selected) {
```
Westbrook, line 94: This is a clear implementation of the described use case. Related question, should there be API/handling so that were the `selected` value of the `<sp-tabs>` element updated from the outside to include content what was not currently visible that this happened as well? Even if it makes sense, I can see separating the work for the two cases, but want to make sure we think through the whole workflow in case there would be any differences in the API application in that case.

packages/tabs/stories/tabs-overflow.stories.ts
```diff
@@ -23,3 +23,11 @@ export const compact = (args: OverflowProperties): TemplateResult => {
 compact.args = {
     compact: true,
 };
+
+export const autoscroll = (args: OverflowProperties): TemplateResult => {
+    return renderTabsOverflowExample(args);
+};
+autoscroll.args = {
```
Westbrook, line 30: 🔥

packages/tabs/test/tabs-overflow.test.ts
```diff
@@ -73,7 +93,7 @@ describe('TabsOverflow', () => {
         const el = await fixture<TabsOverflow>(
             html`
                 <sp-tabs-overflow>
-                    <sp-tabs size="m" selected="1" enableTabsScroll=${true}>
+                    <sp-tabs size="m" selected="1" ?enableTabsScroll=${true}>
```
Westbrook, line 96: This is always setting `true` so we could just do:
```suggestion
                    <sp-tabs size="m" selected="1" enableTabsScroll>
```
However, is it technically correct for this property to be set from an attribute, or should it _always_ be managed by `<sp-tabs-overflow>`?

packages/tabs/test/tabs-overflow.test.ts
```diff
@@ -157,4 +189,21 @@ describe('TabsOverflow', () => {
         const slotContent = slot?.assignedElements() || '';
         expect(slotContent[0].toString()).to.not.contains('Tabs');
     });
+
+    it('should automatically bring the selected tab into view', async () => {
+        const spy = stub(HTMLElement.prototype, 'scrollIntoView');
```
Westbrook, line 194: Do we really want to stub a method, which is an implementation detail and can change, or would it be better to use `el.getBoundingClientRect()` to see that the child exists within the bounds of the parent which would be the experience we're specifically looking to deliver to the end user?

Westbrook (commented):
Coming along nicely! A couple of nits and questions and we'll be ready to land this.

packages/tabs/src/TabsOverflow.ts
Rocss: Now that you mentioned it I think would be more extensible to have a public method on `Tabs`: 
```
    public scrollToSelection(): void {
        if (this.enableTabsScroll && this.selected) {
                const selectedTab = this.querySelector(
                    `[role="tab"][value="${this.selected}"]`
                ) as Tab;
                selectedTab?.scrollIntoView({ inline: 'center', behavior: 'smooth' });
            });
        }
    }
```
similar to the `scrollTo` one which consumers can call if needed.  It can even receive scrollOptions parameters.

The `autoscroll` (or `autoscroll-to-selection`?) property on `TabsOverflow` would just enable this behaviour for you  (would pass a boolean similar to `enableTabsScroll`) and call `scrollToSelection` on `updated`.

```
    protected override updated(changedProperties: PropertyValueMap<this>): void {
        if (this.shouldAutoScroll && changedProperties.has('selected')) {
           // do check if it is out of view first and then:
            this.scrollToSelection();
        }
    }
```


For this automatic behaviour you would need to calculate the boundings though, so that you don't do unnecessary scrolls on tab click. But I think this enables out of the box scrolling, and adds an API which can be used for custom use cases.

Attaching a video for demonstration, I don't have a strong opinion on the API, so open to any suggestions!


https://github.com/adobe/spectrum-web-components/assets/13311865/f3e9df29-f88e-4fa5-a376-8de36dc20966

packages/tabs/test/tabs-overflow.test.ts
Rocss: Right, I think it is has no effect when set from an attribute and should be managed by `<sp-tabs-overflow>`. They way I see it, it is just a helper property to let the `<sp-tabs>` know it is wrapper in a `<sp-tabs-overflow>`.

packages/tabs/test/tabs-overflow.test.ts
Rocss: Nope, will try to calculate the `boundaries` instead!

packages/tabs/src/TabsOverflow.ts
Westbrook: Does the use of `scrollIntoView` require the use of `_isSelectionOutOfViewport`? I like the direction this is going in, but I'm a but remiss to add DOM measurement to the workflow here...

Rocss: @Westbrook I can totally get rid of the DOM measurement checks, but then the Tabs will always try to scroll even if not necessary

Westbrook: Hrm. What if we use the following CSS on the `sp-tabs`:
```
:host {
    scroll-margin-inline: var(--mod-tabs-item-horizontal-spacing, var(--spectrum-tabs-item-horizontal-spacing) );
}
```
And `selectedTab?.scrollIntoView();` with no arguments? Do we get something similar to what we want when a Tab is offscreen and no (I think no, but I'd accept a very small amount, too) scrolling when the element is in view?

packages/tabs/test/tabs-overflow.test.ts
Westbrook: The `<sp-tabs-overlay>` sets this value, meaning a consumer of the components shouldn't need to set this value, meaning the story probably shouldn't set a bad example by setting this value...

packages/tabs/src/Tabs.ts
```diff
@@ -229,15 +233,40 @@ export class Tabs extends SizedMixin(Focusable, { noDefaultSize: true }) {
...
+    public scrollToSelection(): void {
+        if (this.enableTabsScroll && this.selected) {
+            this.updateComplete.then(() => {
+                const selectedTab = this.querySelector(
+                    `[role="tab"][value="${this.selected}"]`
+                ) as Tab;
+                selectedTab?.scrollIntoView();
+            });
+        }
+    }
```
Westbrook, line 260:
```suggestion
    public async scrollToSelection(): Promise<void> {
        if (!this.enableTabsScroll || !this.selected) {
            return;
        }
        await this.updateComplete;
        const selectedTab = this.tabs.find((tab) => tab.value === this.selected);
        selectedTab?.scrollIntoView();
    }
```
Structure nit.

packages/tabs/src/Tabs.ts
```diff
@@ -124,6 +125,9 @@ export class Tabs extends SizedMixin(Focusable, { noDefaultSize: true }) {
     @property({ attribute: false })
     public shouldAnimate = false;
 
+    @property({ attribute: false })
+    public shouldAutoscroll = false;
```
Westbrook, line 129: I feel like I asked this, but can't find it in previous comments, with this structure, does this need to be opt-in? Feels like we should always scroll to the item. Most time it wouldn't do anything, but the couple of times it did, it would be a win for everyone, right?

packages/tabs/src/Tabs.ts
```diff
@@ -229,15 +233,40 @@ export class Tabs extends SizedMixin(Focusable, { noDefaultSize: true }) {
...
+            });
+        }
+    }
+
+    protected override updated(
+        changedProperties: PropertyValueMap<this>
+    ): void {
+        super.updated(changedProperties);
+
+        if (changedProperties.has('selected') && this.shouldAutoscroll) {
```
Westbrook, line 267: If we don't remove `this.shouldAutoscroll`, would it be better as a gate internal to `scrollToSelection()`?

packages/tabs/test/tabs-overflow.test.ts
```diff
@@ -16,23 +16,38 @@ import '@spectrum-web-components/tabs/sp-tab.js';
...
+    selected = 1,
+    autoscroll = false,
+}: OverflowProperties): Promise<HTMLDivElement> => {
     const tabsContainer = await fixture<HTMLDivElement>(
         html`
             <div class="container" style="width: 200px; height: 150px;">
-                <sp-tabs-overflow>
-                    <sp-tabs size=${size} selected="1" enableTabsScroll=${true}>
+                <sp-tabs-overflow ?autoscroll=${autoscroll}>
+                    <sp-tabs size=${size} selected=${selected}>
```
Westbrook, line 50: :fire:

Westbrook (commented):
This likely also needs a bump in the VRTs after other changes, but with this last nit or two it's looking ready to go!

Westbrook (approved):
I kicked the flaky test past and it looks like we're good to go here.

Thanks for driving this home! :shipit: LGTM.

PR #4042, opened 2024-02-14
chore(slider): updated handle ui to s1 design

Rajdeepc:
## Description
Update `sp-slider` handle to S1 design

Screenshot 2024-02-14 at 3 48 28 PM
Screenshot 2024-02-14 at 3 48 34 PM

## Related issue(s)

- With the design team pushing back on the recent Slider Handle changes, and it being a blocker for PsWeb catching up to more recent released of SWC.
CSS: Current implementation is correct
SWC: May eventually be this way in S2 but we need to update for S1

## How has this been tested?

-   [x] _Test case 1_
    1. Go [here](https://slider-handleui--spectrum-web-components.netlify.app/storybook/?path=/story/slider--default)

## Types of changes

## Checklist

## Best practices


Westbrook:
Sometimes when there are a lot of changes it take a couple of times to shake out the VRT update. Please take another look at that so we can get this passing across the board.

Rajdeepc:
> Sometimes when there are a lot of changes it take a couple of times to shake out the VRT update. Please take another look at that so we can get this passing across the board.

@Westbrook Good to land now.

packages/slider/src/Slider.ts
```diff
@@ -410,6 +410,15 @@ export class Slider extends SizedMixin(ObserveSlotText(SliderHandle, ''), {
         `;
     }
 
+    private renderHandle(): TemplateResult {
+        if (this.variant === 'tick') {
+            return html``;
+        }
+        return html`
+            ${this.handleController.render()}
+        `;
```
spdev3000, line 419: NIT: Can be shortened to: 
```
return this.variant === 'tick' ?  html`` : this.handleController.render()
```

spdev3000 (approved):
lgtm

packages/slider/src/Slider.ts
Westbrook: I'd prefer this as well.

Westbrook (commented):
Please update one of the stories to have a minimum or maximum value by default to make it easier to track this specific style of change in the future.

spdev3000 (approved):
lgtm, thx!

packages/slider/src/Slider.ts
```diff
@@ -410,6 +410,12 @@ export class Slider extends SizedMixin(ObserveSlotText(SliderHandle, ''), {
         `;
     }
 
+    private renderHandle(): TemplateResult {
+        return html`
+            ${this.variant === 'tick' ? html`` : this.handleController.render()}
+        `;
```
Westbrook, line 416: At some point we should do focused testing on whether the use of stacked templates here causes us any performance costs, but the small scale use here doesn't seem to be tweaking the performance of the component itself, so :shipit:

Westbrook (approved):

PR #4050, opened 2024-02-16
fix(overlay): leverage "transition-behavior" to persist top-layer content while closing

Westbrook:
## Description
Currently when the browser closers an Overlay for us, it will fall off of the `#top-layer` immediately and allow the Overlay content to paint behind standard page content while its close animation is playing.

`transition-behavior`, in association with the `overlay` property, allows the content to stay on the `#top-layer` through the course of the animation.

*TO DO*
- [x] decide whether the [open Picker demo](https://transition-behavior--spectrum-web-components.netlify.app/storybook/?path=/story/picker--open) should continue to be forward facing and test this specific feature, while being broken in Firefox/WebKit, or whether it should be normalized for "working as expected" everywhere

## How has this been tested?
-   [ ] _Test case 1_
    1. Go [here](https://transition-behavior--spectrum-web-components.netlify.app/storybook/?path=/story/picker--open) (Currently Chromium only, [more info](https://developer.mozilla.org/en-US/docs/Web/CSS/overlay))
    2. Open the Picker (it should be by default)
    3. Click away from the Menu, so the Picker closes
    4. See that the Menu does not appear "behind" the second Picker while its close animation plays

## Types of changes

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ executors:
 parameters:
     current_golden_images_hash:
         type: string
-        default: 686fa6e78586a47ae3fd13d4171392ca5aa5d42d
+        default: 824fb67e89d53a25416b315fc8dda02aa83e5482
     wireit_cache_name:
         type: string
         default: wireit
--- a/packages/overlay/src/AbstractOverlay.ts
+++ b/packages/overlay/src/AbstractOverlay.ts
@@ -237,6 +237,9 @@ export class AbstractOverlay extends SpectrumElement {
     protected placementController!: PlacementController;
     receivesFocus!: 'true' | 'false' | 'auto';
     protected requestSlottable(): void {}
+    protected returnFocus(): void {
+        return;
+    }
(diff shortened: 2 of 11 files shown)


hunterloftis:
In FF (123.0), the lower (closed) picker renders over the upper (open) picker:

image

Westbrook:
@hunterloftis this is expected as it falls outside of the support matrix of the polyfill we deliver the Overlay with when `popover` is not supported. See:

image

Do you think we should roll back that change in the demos? Firefox is tracking v125 support for `popover` and is currently at v123.

hunterloftis:
@Westbrook ah I see, it's failing due to the backdrop filter.

Personally I was thrown by the demo (it looked like a bug, rather than a demonstration of an edge case). I think it's fine to leave it as long as we make it clear that it's intending to demonstrate a cross-browser limitation. Naming it something other than "open," or putting text around the demo itself describing what it's showing.

hunterloftis (approved):

PR #3669, opened 2023-09-21
fix(menu): allow Menu elements to be controlled

Westbrook:
## Description
Allow the `selected` property of an `<sp-menu>` to drive associated state so that the element can be leveraged as a "controlled component".

## Related issue(s)
- fixes #3667

## How has this been tested?
-   [ ] _Test case 1_
    1. Go [here](https://controlled-menu--spectrum-web-components.netlify.app/storybook/?path=/story/menu--controlled)
    2. Click any item
    3. See that the `selected` value is forced to `[ 'Select and Mask...' ]`.
    4. Click any _other_ item
    5. See that the `selected` value is forced to `[ 'Select and Mask...' ]`.

## Types of changes
-   [x] New feature (non-breaking change which adds functionality)

## Checklist
-   [x] I have signed the [Adobe Open Source CLA](http://opensource.adobe.com/cla.html).
-   [x] My code follows the code style of this project.
-   [ ] If my change required a change to the documentation, I have updated the documentation in this pull request.
-   [x] I have read the **[CONTRIBUTING](<(https://github.com/adobe/spectrum-web-components/blob/main/CONTRIBUTING.md)>)** document.
-   [x] I have added tests to cover my changes.
-   [x] All new and existing tests passed.

PR #4051, opened 2024-02-16
Fix controlled menuitems

hunterloftis:
## Description

"A" (probably not "the") fix for the currently-broken controlled action-menu items:

- https://opensource.adobe.com/spectrum-web-components/storybook/?path=/story/action-menu--controlled

(opening and closing, without clicking on any items, will clear all items' selected states)
(it probably should not trigger the "action" popup immediately; that may be a separate issue)

The root cause is that `PickerBase` always assigns its child `<sp-menu>`'s `.selected` array. Since `ActionMenu` extends `PickerBase`, it _also_ always assigns the selected array. So an action menu like this:

```html
<sp-action-menu>
  <sp-menu-item selected>Red</sp-menu-item>
  <sp-menu-item selected>Green</sp-menu-item>
  <sp-menu-item selected>Blue</sp-menu-item>
</sp-action-menu>
```

...will render correctly, once (all items visibly selected), but upon opening & closing the menu, the action-menu will reset the `.selected` array to `[]` which will clear the selected state from every item. This behavior started here:

https://github.com/adobe/spectrum-web-components/pull/3669

This demonstration fix sidesteps the issue by not setting `selected` on `unmanaged` action menus. That change can't be applied directly to `PickerBase` because then there are side effects for other components that inherit from `PickerBase`, and even for action menus that are intended to be used by assigning values to `.selected`.

Sharing this imperfect/hacky fix in order to start a conversation about what a long-term fix could look like.

### DIFF
--- a/packages/action-menu/src/ActionMenu.ts
+++ b/packages/action-menu/src/ActionMenu.ts
@@ -45,6 +45,9 @@ export class ActionMenu extends ObserveSlotPresence(
         return [actionMenuStyles];
     }
 
+    @property({ type: Boolean })
+    public unmanaged = false;
+
     @property({ type: String })
     public override selects: undefined | 'single' = undefined;
 
@@ -116,8 +119,37 @@ export class ActionMenu extends ObserveSlotPresence(
             >
                 ${this.buttonContent}
             </sp-action-button>
-            ${this.renderMenu} ${this.renderDescriptionSlot}
+            ${this.unmanaged ? this.renderUnmanagedMenu : this.renderMenu}
+            ${this.renderDescriptionSlot}
         `;
+    }
+
+    protected get renderUnmanagedMenu(): TemplateResult {
+        const menu = html`
+            <sp-menu
+                aria-labelledby="applied-label"
+                @change=${this.handleChange}
+                id="menu"
+                @keydown=${{
+                    handleEvent: this.handleEnterKeydown,
+                    capture: true,
+                }}
+                role=${this.listRole}
+                size=${this.size}
+                @sp-menu-item-added-or-updated=${this.shouldManageSelection}
+            >
+                <slot @slotchange=${this.shouldScheduleManageSelection}></slot>
+            </sp-menu>
+        `;
+        this.hasRenderedOverlay =
+            this.hasRenderedOverlay ||
+            this.focused ||
+            this.open ||
+            !!this.deprecatedMenu;
+        if (this.hasRenderedOverlay) {
+            return this.renderOverlay(menu);
+        }
+        return menu;
     }
 
     protected override update(changedProperties: PropertyValues<this>): void {
--- a/packages/action-menu/stories/action-menu.stories.ts
+++ b/packages/action-menu/stories/action-menu.stories.ts
@@ -320,6 +320,7 @@ export const controlled = ({ align = 'start' } = {}): TemplateResult => {
     return html`
         <sp-action-menu
             label="View"
+            unmanaged
             @change=${onChange}
             style=${ifDefined(
                 align === 'end' ? 'float: inline-end;' : undefined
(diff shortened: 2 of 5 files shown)


Westbrook:
Could similar be achieved by moving the application of `selected` from the declarative renderer to something like:
```
    protected override updated(): void {
        if (this.selects) {
            (this.shadowRoot.querySelector('sp-menu') as Menu).selected= this.value ? [this.value] : [];
        }
    }
```
Types are not so happy about it at current, but you can seemingly do this declarative with:
```
    .selected=${this.selects ? (this.value ? [this.value] : []) : undefined}
```

hunterloftis:
I'm not sure what the contract is exactly between `selects` and `selected` - if `selects` is undefined, is the component allowed to ignore `selected`?

If so then 👍 that's the way to go.

Westbrook:
> When the `selects` attribute is not present a `value` will not be maintained and the Menu Item children of this Menu will not have their `selected` state managed.

No idea whether there are tests that behave differently or not on this.

najikahalsema:
> No idea whether there are tests that behave differently or not on this.

It looks like there are tests in menu for when `selects` is actually set, but no tests for situations where it's undefined?

hunterloftis:
> No idea whether there are tests that behave differently or not on this.

It causes this test to fail: https://github.com/adobe/spectrum-web-components/blob/4a31c5b407e3cb4f3d684363d8d35008488ec617/packages/action-menu/test/index.ts#L499

So far I've only come up with fixes that have a breaking API change: either requiring `selects` to be some value in order for `selected` to apply, or requiring `selects` to be some special sentinel value (or adding another sentinel like `unmanaged`) in order for `selected` to _not_ apply.

Westbrook:
What happens if you change https://github.com/adobe/spectrum-web-components/blob/main/packages/menu/src/Menu.ts#L95 to be if (selected === this.selected || !this.selects) {`?

There's also likely a reasonable path to retiring a test or two if we need to in order to have an understandable API for this....

hunterloftis:
@Westbrook that also fails the "allows top-level selection state to change" since right now, setting `menu.selected` works regardless of the value of `selects` & that's how it's tested. I think it would be reasonable to change that, but I'm not sure what to, since the values available to `selects` seems to mean "manage selected automatically/internally" whereas setting `selected` means "I'll set this explicitly/externally."

Maybe a new value like `selects = "selected"`? That's a breaking API change but it seems like a decent path forward.

The alternative is the inverse, like `selects = "none"`, which would be a sentinel to block any internal management of the `selected` value. I slightly prefer that option because it seems a little clearer & easier to explain.

spdev3000:
Just to add my 2 cents here, as I also try to extend current ActionMenu to a CustomActionMenu for different reasons, and the selection behavior was one problem to solve here, as we want to control the selection states of the menu-items from the outside.
So `selects` of the ActionMenu is `undefined` and setting `sp-menu-item` to `selected` and then open/close ActionMenu removes the selection state. 
I think a solution might be to change the line in `PickerBase.renderMenu` 
from 
```
 .selected=${this.value ? [this.value] : []} 
 ``` 
  to 
  ```
  .selected=${this.selects && this.value ? [this.value] : []}
 ```
  
see https://github.com/adobe/spectrum-web-components/blob/7311822ef254de843e3aacce9063fd3d30cd666e/packages/picker/src/Picker.ts#L662
without the need to add any extra logic for "unmanaged" menus.
Of course tests for that use case needs to be added.

**Comment by spdev3000** [comment]
Just to add my 2 cents here, as I also try to extend current ActionMenu to a CustomActionMenu for different reasons, and the selection behavior was one problem to solve here, as we want to control the selection states of the menu-items from the outside.
So selects of the ActionMenu is `undefined` and setting sp-menu-item to `selected` and then open/close ActionMenu removes the selection state.
Can we maybe try to only set the `.selected` value of `PickerBase.renderMenu` if the ActionMenu has a `selects` value !== `undefined`?
Maybe you can skip the `unmanaged` property and just compare for `this.selects` resulting in:
```
${this.selects ? this.renderMenu : this.renderUnmanagedMenu}
```

Westbrook:
Would be better to confirm that selection is appropriately managed when using `event.preventDefault()` on the `change` event rather than expanding the API here. Closing in preference of a solution in that shape.

PR #4105, opened 2024-03-01
chore(split-button): initiate deprecation notice

Rajdeepc:
## Description

This is the first pass of putting a deprecation notice on split-button.
We will be adding deprecation notice on docs site as well as pn storybook for this element.

## Related issue(s)

-

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

image

## Types of changes

## Checklist

## Best practices


packages/split-button/src/SplitButton.ts
```diff
@@ -115,6 +115,14 @@ export class SplitButton extends SizedMixin(PickerBase) {
                 this.selects = 'single';
             }
         }
+        if (window.__swc.DEBUG) {
+            window.__swc.warn(
+                this,
+                `<${this.localName}> is deprecated. Use a [button group](https://opensource.adobe.com/spectrum-web-components/components/button-group/) to show any additional actions related to the most critical action.`,
+                'https://opensource.adobe.com/spectrum-web-components/components/button-group/',
```
Westbrook, line 122: This should likely point to a “Deprecated” section that we add to the Split Button readme to ensure the same message is received by devs who do not leverage DEV Mode.

Rajdeepc: Yes that's the final motive. I have added the change to be added here in the description

Rajdeepc: Screenshot 2024-03-04 at 4 16 29 PM

projects/documentation/src/components/styles.css
```diff
@@ -950,6 +951,10 @@ sp-link:not(:defined) {
     border-radius: 6px;
 }
 
+.deprecation-container {
+    margin: 3em 0;
```
Westbrook, line 955: The demo content and page header both use `2em` for their spacing, not sure if that's something to take in mind when making the decision on spacing here.

packages/split-button/README.md
```diff
@@ -2,6 +2,10 @@
 
 An `sp-split-button` surfaces an immediately envokable action via it's main button, as well as a list of alternative actions in its toggleable menu overlay. By default, any actions envoked from the overlay will replace the main action button. When leveraging `[type="more"]` the action will be envoked, but the main button will remain the action initally persribed by the implementor.
 
+## Deprecated
```
Westbrook, line 5: Sorry, I didn't mean that we needed new content (or in this case, duplicate), just that the Inline Alert that you've added should include an anchor link that we could address with `#deprecation` to ensure that the link shared in Dev Mode got people to the appropriate part of the page.

projects/css-custom-vars-viewer/package.json
```diff
@@ -1,7 +1,6 @@
 {
     "name": "@spectrum-web-components/custom-vars-viewer",
     "version": "0.41.1",
-    "private": "true",
```
Westbrook, line 4: This is supposed to be private.

projects/documentation/scripts/copy-component-docs.js
```diff
@@ -57,6 +57,28 @@ const findDeclaration = (customElements, test) => {
...
+                `../../tools/${packageName}/package.json`,
+                {
+                    assert: { type: 'json' },
+                }
+            ).then((packageDefault) => packageDefault.default);
+        } catch (error) {}
+    }
+    return packageJSON.deprecationNotice;
+};
+
```
Westbrook, line 81: I think we know whether something is `packages` or `tools` in a way that doesn't require this `try/catch` gate.

packages/split-button/src/SplitButton.ts
```diff
@@ -115,6 +115,14 @@ export class SplitButton extends SizedMixin(PickerBase) {
                 this.selects = 'single';
             }
         }
+        if (window.__swc.DEBUG) {
+            window.__swc.warn(
```
jnjosh, line 119: I wonder if we should introduce a deprecated notice function here, not because it would do something different but we could do different things in the future and have all deprecations do the same thing.

Westbrook: FYI, this displays as follows for a development time user: 

image

Westbrook (approved):
Ship it!

Issue #3549, opened 2023-08-15
[Feat]: Directive based approach to overlay-trigger


benjamind:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

overlay-trigger, tooltip

### Description of the requested feature

We should explore creating a directive for creating behavior similar to `overlay-trigger` but using a directive to support lazy rendering of the overlay content. This should have significant improvement to element count and initial render time as it will eliminate a large number of unnecessary DOM nodes, only rendering them on demand.

Additionally we could extend the `trigger` directive to have a special case for `tooltip` for easy of use.

### Mockups or screenshots

Example of a trigger directive in use:
```html
            <sp-button
                ${trigger(html`
                    <sp-popover>
                        <p>This is the content.</p>
                    </sp-popover>
                `)}
            >
                Trigger
            </sp-button>
```

Similarly tooltip could be used like this:
```html
            <sp-button ${tooltip('A tooltip appears!')}>
                Hover me
            </sp-button>
```

PR #4112, opened 2024-03-01
chore(overlay): add "slottable-request" events to the very beginning/end of the Overlay lifecycle

Westbrook:
## Description
Surface as yet unused `slottable-request` event infrastructure to unlock in situ testing of some theoretical performance optimizations. Use `chore` in the commit so that this change doesn't appear on the change log.

## Related issue(s)
- refs #3549 

## How has this been tested?
-   [ ] _Test case 1_
    1. Namely checkout the tests that run, but know this is an experimental feature that could be removed or changed at any time.

## Types of changes

## Checklist

### DIFF
--- a/packages/combobox/src/Combobox.ts
+++ b/packages/combobox/src/Combobox.ts
@@ -507,8 +507,11 @@ export class Combobox extends Textfield {
             this.open = false;
         }
         if (changed.has('activeDescendant')) {
-            if (changed.get('activeDescendant')) {
-                (changed.get('activeDescendant') as MenuItem).focused = false;
+            const previouslyActiveDescendant = changed.get(
+                'activeDescendant'
+            ) as unknown as MenuItem;
+            if (previouslyActiveDescendant) {
+                previouslyActiveDescendant.focused = false;
             }
             if (
                 this.activeDescendant &&
--- a/packages/overlay/src/slottable-request-event.ts
+++ b/packages/overlay/src/slottable-request-event.ts
@@ -1,0 +1,41 @@
+/*
+Copyright 2024 Adobe. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License. You may obtain a copy
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+export class SlottableRequestEvent extends Event {
+    readonly data: unknown;
+    readonly name: string;
+    readonly slotName: string;
+    constructor(name: string, data: unknown, key?: string) {
+        super('slottable-request', {
+            bubbles: false,
+            cancelable: true,
+            composed: false,
+        });
+        this.name = name;
+        this.data = data;
+        this.slotName = key !== undefined ? `${name}.${key}` : name;
+        if (window.__swc.DEBUG) {
+            window.__swc.warn(
+                undefined,
+                `⚠️  WARNING ⚠️ : \`slottable-request\` events are experimental and there is no guarantees behind usage of them in an application!! Their shape and presence within the library could be changed at anytime.
+                
+Learn more about the protocol these events are based on below:`,
+                'https://github.com/webcomponents-cg/community-protocols/pull/45',
+                {
+                    level: 'high',
+                    type: 'api',
+                }
+            );
+        }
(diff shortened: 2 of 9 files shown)


najikahalsema (commented):
Doesn't have to be in this PR, of course, but it would be cool to see examples of how this could be used. That's getting super into the weeds, though, I think haha. :^) Doesn't seem like it'd cause issues beyond simply firing an event, and i like how segmented it is and the inclusion of the dev mode warning.

Want to hear from others about their thoughts, too!

packages/overlay/test/overlay-element.test.ts
```diff
@@ -74,6 +79,140 @@ describe('sp-overlay', () => {
         });
     }
 
+    describe.only('`slottable-request` event', () => {
```
tjgupta, line 82: I assume `.only()` wasn't meant to be checked in?

tjgupta:
> Doesn't have to be in this PR, of course, but it would be cool to see examples of how this could be used. That's getting super into the weeds, though, I think haha. :^) Doesn't seem like it'd cause issues beyond simply firing an event, and i like how segmented it is and the inclusion of the dev mode warning.
> 
> Want to hear from others about their thoughts, too!

I agree with this, seeing an example usage would help me, but I can find the example elsewhere too if that's easier since this is early stage.

packages/overlay/test/overlay-element.test.ts
Westbrook: Thanks for catching this. I had thought we'd put linting for this in previously, but it looks like I'll need to add that in a separate PR.

Fixed!

packages/combobox/src/Combobox.ts
```diff
@@ -508,7 +508,9 @@ export class Combobox extends Textfield {
         }
         if (changed.has('activeDescendant')) {
             if (changed.get('activeDescendant')) {
-                (changed.get('activeDescendant') as MenuItem).focused = false;
+                (
+                    changed.get('activeDescendant') as unknown as MenuItem
+                ).focused = false;
```
hunterloftis, line 513: I'm sure looking up 'activeDescendant' twice isn't expensive, but doing it only once would preclude typos with the key:

```
const descendant = changed.get('activeDescendant') as unknown as MenuItem | undefined;
if (descendant) {
  descendant.focused = false;
}
```

packages/overlay/src/slottable-request-event.ts
```diff
@@ -0,0 +1,41 @@
...
+                {
+                    level: 'high',
+                    type: 'api',
+                }
+            );
+        }
+    }
+}
+
+export const remove = Symbol('remove-slottable-request');
```
hunterloftis, line 41: `remove` is short & generic so it's likely to be shadowed by consuming code.

I think you should keep it as-is since it maps to the proposal, but I'm curious if you think it's worth suggesting a proposal spec change to something like `SlottableRequestEvent.remove` as a static symbol.

hunterloftis (approved):

packages/overlay/test/overlay-element.test.ts
```diff
@@ -74,6 +79,140 @@ describe('sp-overlay', () => {
...
+            await nextFrame();
+
+            const closed = oneEvent(el, 'sp-closed');
+            el.open = false;
+            await closed;
+
+            expect(
+                slottableRequestTime,
+                `slottable-request: ${slottableRequestTime}, sp-closed: ${closedTime}`
+            ).to.be.gte(closedTime);
```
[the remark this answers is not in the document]
Westbrook: This was failing without the `gte`, which is why I switched to this for all of the comparisons. I'll take another look to confirm.

packages/overlay/src/slottable-request-event.ts
Westbrook: I'm not sure the name of the variable needs to be fixed, at least for now. Changing here.

Would you be able to comment into the protocol PR on this topic?

packages/overlay/test/overlay-element.test.ts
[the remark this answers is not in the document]
hunterloftis: It seems like `gte` means "didn't happen in the past?" But maybe that's not quite right, I'm not familiar with the timings.

If we expect there to be any human-detectable amount of time between signaling a close and a close completing, I would expect `performance.now` to pick it up as > 0.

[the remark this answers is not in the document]
Westbrook: Agreed. I've altered the workflow for this to ensure that the `slottable-request` event goes out after the `sp-closed` event in a more strict manner.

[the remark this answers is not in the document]
Westbrook: For reference, only WebKit is _SO_ fast the it can dispatch both of those events in the same millisecond.

PR #4122, opened 2024-03-04
chore(overlay, tooltip): v0 directive for Overlay v2 testing

Westbrook:
## Description
Directives, demos, and tests for lazily rendering Overlay content.

This prepares for an initial pre-release for use in major consuming clients for testing purposes.

*EVERYTHING HERE IS EXPERIMENTAL!*

We're looking at pre-releasing this to `latest` and then revisit after some testing for documentation/announcement.

_Known issues:_
- directives do not orchestrate with each other, so using more than one on the same trigger is not currently suggested
- lazily populating an `<sp-overlay>` should be _roughly_ as performant as the directive, but not currently sure why there's a performance delta between the two (this lead to some gaming of the class structure that could technically be reverted)
- I'm not super happy with this argument shape or the defaults it represents:
  ```
        triggerInteraction: 'hover',
        overlayOptions: {
            placement: 'top-start',
            type: 'auto',
            offset: 0,
        },
  ```
  further pre-release experimentation will likely smooth that out.
- there are additional DEV MODE warnings included for the experimental nature of this code
- there's not a lot of tests of the directive

## Related issue(s)
- refs #3549

## How has this been tested?
-   [ ] _Test case 1_
    1. Go [here](https://overlay-directive-v2--spectrum-web-components.netlify.app/storybook/?path=/story/overlay-directive--default)
    2. See the tooltip and the popover related to the button

## Types of changes

## Checklist

### DIFF
--- a/packages/overlay/src/overlay-trigger-directive.ts
+++ b/packages/overlay/src/overlay-trigger-directive.ts
@@ -1,0 +1,200 @@
+/*
+Copyright 2023 Adobe. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License. You may obtain a copy
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+import { ElementPart, nothing, render, TemplateResult } from 'lit';
+import { AsyncDirective, directive } from 'lit/async-directive.js';
+import { Overlay } from './overlay.js';
+import { OverlayContentTypes } from './OverlayTrigger.js';
+import {
+    OverlayOpenCloseDetail,
+    OverlayTriggerInteractions,
+    Placement,
+} from './overlay-types.js';
+
+export type OverlayTriggerOptions = {
+    triggerOn: OverlayContentTypes;
+    placement: Placement;
+    type: OverlayTriggerInteractions;
+    offset: number;
+};
+
+export class OverlayTriggerDirective extends AsyncDirective {
+    private template?: TemplateResult;
+    private target?: HTMLElement;
+
+    protected defaultOptions: OverlayTriggerOptions = {
+        triggerOn: 'hover',
+        placement: 'top-start',
+        type: 'inline',
+        offset: 0,
+    };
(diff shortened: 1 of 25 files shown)


packages/overlay/src/overlay-trigger-directive.ts
```diff
@@ -0,0 +1,142 @@
...
+
+        render(willRemoveSlottable ? undefined : this.template(), this.overlay);
+        if (willRemoveSlottable) {
+            this.overlay.remove();
+        } else {
+            Overlay.applyOptions(this.overlay, {
+                ...this.options,
+                trigger: this.target,
+            });
+            this.target.insertAdjacentElement('afterend', this.overlay);
```
hunterloftis, line 117: Should we make the insertion point configurable or is after the target always ok?

packages/overlay/stories/overlay-directive.stories.ts
```diff
@@ -0,0 +1,223 @@
...
+        <sp-button
+            variant="primary"
+            ${tooltip(
+                () =>
+                    html`
+                        Click to open a popover.
+                    `
+            )}
+            ${trigger(
+                () => html`
```
hunterloftis, line 154: I know it's just a story, but I wonder what you think of naming the various parts as render functions just for readability:

```
const renderPopover = () => html`
  ...
`;

<sp-button ${trigger(renderPopover)}>
```

packages/tooltip/src/tooltip-directive.ts
```diff
@@ -0,0 +1,31 @@
...
+    options?: Partial<OverlayTriggerOptions & { variant: string }>
+): ReturnType<typeof trigger> {
+    return trigger(() => {
+        import('@spectrum-web-components/tooltip/sp-tooltip.js');
+        return html`
+            <sp-tooltip variant=${ifDefined(options?.variant)}>
+                ${tooltipContent()}
+            </sp-tooltip>
+        `;
+    }, options);
```
hunterloftis, line 30: 🔥 🔥 🔥

hunterloftis (dismissed):
Very interested to see what impact this has.

I also suspect that adopting these directives will expose places where apps are inadvertently relying on the DOM to maintain state.

packages/overlay/src/overlay-trigger-directive.ts
Westbrook: Added.

beeduul:
> I also suspect that adopting these directives will expose places where apps are inadvertently relying on the DOM to maintain state.

that could be really interesting if so.  and a statement that doing things the "right way" benefits performance AND robustness/correctness.

hunterloftis (approved):

Issue #3393, opened 2023-06-29
[Bug][a11y]: Virtualized Table should include aria-rowcount attribute


majornista:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Table

### Expected behavior

When a Table is virtualized it should include an [`aria-rowcount`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-rowcount) attribute to communicate the number of total rows in the table, otherwise screen readers will announce only the number of rendered rows.

### Actual behavior

Virtualized table lacks `aria-rowcount` attribute, so screen readers will announce only the number of rendered rows rather than the total rows in the table.

### How can we reproduce this issue?

3. Scroll to '....'
4. Check console
5. See error

PR #4126, opened 2024-03-05
fix(table): Add aria-rowcount to virtualized tables

blaabaer:
## Description

The `sp-table` component can be virtualized, which only renders table rows that are currently visible in the viewport, which helps with performance on large tables. However, this can trip up screenreaders, which only detect and announce the rendered rows, and not the true size of the table. `aria-rowcount` is a screenreader attribute which announces the true unvirtualized size of the table.

This PR implements the automatic application of `aria-rowcount` for virtualized tables. 

## Related issue(s)

- #3393 

## Motivation and context

Accuracy and efficacy of screenreader accessibility.

## How has this been tested?

-   [ ] Run storybook - Table > Virtualized > [Any of these]
    1. Inspect the `sp-table` element
    2. Observe that the `aria-rowcount` value is correct.

image

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/table/src/Table.ts
+++ b/packages/table/src/Table.ts
@@ -507,6 +507,9 @@ export class Table extends SizedMixin(SpectrumElement, {
                 }
             );
         }
+        if (this.items.length) {
+            this.setAttribute('aria-rowcount', `${this.items.length}`);
+        }
         const config: VirtualizeDirectiveConfig<Record<string, unknown>> = {
             items: this.items,
             renderItem: this.renderItem,
--- a/packages/table/src/Table.ts
+++ b/packages/table/src/Table.ts
@@ -508,6 +508,8 @@ export class Table extends SizedMixin(SpectrumElement, {
             );
         }
         if (this.items.length) {
+            // Ensures screenreaders can announce the true size of the table
+            // despite virtualization only rendering a subset.
             this.setAttribute('aria-rowcount', `${this.items.length}`);
         }
         const config: VirtualizeDirectiveConfig<Record<string, unknown>> = {
--- a/packages/table/stories/table-virtualized.stories.ts
+++ b/packages/table/stories/table-virtualized.stories.ts
@@ -97,7 +97,6 @@ class VirtualTable extends SpectrumElement {
     protected override render(): TemplateResult {
         return html`
             <sp-table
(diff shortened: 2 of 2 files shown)


packages/table/src/Table.ts
```diff
@@ -507,6 +507,11 @@ export class Table extends SizedMixin(SpectrumElement, {
                 }
             );
         }
+        if (this.items.length) {
+            // Ensures screenreaders can announce the true size of the table
+            // despite virtualization only rendering a subset.
+            this.setAttribute('aria-rowcount', `${this.items.length}`);
+        }
```
Westbrook, line 514: This method is gated by this test, so we might not need it here.

Also, do we want to remove this attribute ever? E.G. we wouldn't come into this method is `this.items` was set to `[]` later on, but should `aria-rowcount="0"` in that case, or be removed?

packages/table/stories/table-virtualized.stories.ts
```diff
@@ -97,7 +97,6 @@ class VirtualTable extends SpectrumElement {
     protected override render(): TemplateResult {
         return html`
             <sp-table
-                aria-rowcount="50"
```
Westbrook, line 100: Nice catch to remove this here! Were you able to see the tests fail with this removed before you added new code to `Table.ts`?

blaabaer: Yes, I observed that tests failed when it was removed, and they were repaired when my code added it back programmatically.

packages/table/src/Table.ts
blaabaer: You're right, this conditional isn't necessary here!

packages/table/src/Table.ts
```diff
@@ -507,6 +507,11 @@ export class Table extends SizedMixin(SpectrumElement, {
                 }
             );
         }
+
+        // Ensures screenreaders can announce the true size of the table
+        // despite virtualization only rendering a subset of rows.
+        this.setAttribute('aria-rowcount', `${this.items.length}`);
```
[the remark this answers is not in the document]
blaabaer: Yes @audknu and I have updated the description with screenshots.

sarahszhou (approved):
LGTM!

PR #4144, opened 2024-03-06
fix(tab-overflow): improve tab navigation experience and support custom aria labels

### LINKED ISSUE #3270: [Bug][a11y]: Tabs overflow component has action buttons that don't necessarily need keyboard focus
### Code of conduct

### Impacted component(s)

sp-tabs-overflow

### Expected behavior

Please take overflow related previous and next buttons out of taborder by using tabindex="-1" on them. They're also not labelled, so please add aria-label to both the buttons.

### Actual behavior

Having tabstops before the tabs component to reveal any hidden tabs seems unnecessary. Keyboard users are able to use arrow keys to navigate through all the tabs and the visual affordance allows them to know that there are more tabs. For screen reader users, the number of tabs is already conveyed by the tablist. So, overall addition of two tabstops (for back and forward buttons) is not necessary. Even on mobile, using a screenreader, a user can go through all the tabs without needing to use these buttons.

### Screenshots

_No response_

_No response_

_No response_

_No response_

_No response_

blaabaer:
image

`<sp-tab-overflow>` is an optional add-on to the Tabs component that adds a [<] and [>] button to help scroll, if there are so many tabs that they overflow off the page. These buttons are currently Keyboard-Navigable, and capture tab focus before the TabGroup itself. 

These buttons should be removed from keyboard navigation for the following reasons:
- Keyboard navigators can already use their arrow keys to navigate through the entire TabGroup
- Every time keyboard navigation enters the TabGroup component, it will always initially focus the very first Tab, regardless of whether it is currently rendered on or off the page. 

Therefore, it is superfluous to have these buttons part of the tab navigation journey.

In addition, these buttons were not labeled. This PR gives them default `aria-label`s and allows overrides. These default labels are "Scroll to previous tabs" and "Scroll to next tabs" - specifically avoided left/right language in case of Right-To-Left locales.

## Related issue(s)

- Fixes #3270 

## Motivation and context

- Improves Keyboard Accessibility by removing UI that is not relevant to Keyboard Navigation experience.
- Improves Screenreader Accessibility by labeling interactive components.

## How has this been tested?

-   [x] On Storybook and on the Docs website:
    1. Navigate to "Tabs Overflow"
    2. Observe that the Overflow buttons are no longer keyboard navigable, and arrow-key + spacebar navigation works correctly to move between tabs.
    3. Inspect the Overflow buttons and observe that they have default labels: "Scroll to previous tabs" and "Scroll to next tabs".

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/tabs/src/TabsOverflow.ts
+++ b/packages/tabs/src/TabsOverflow.ts
@@ -46,16 +46,16 @@ export class TabsOverflow extends SizedMixin(SpectrumElement) {
     }
 
     @property({ type: Boolean, reflect: true })
-    compact = false;
+    public compact = false;
+
+    @property({ type: String, attribute: 'label-previous' })
+    public labelPrevious = 'Scroll to previous tabs';
+
+    @property({ type: String, attribute: 'label-next' })
+    public labelNext = 'Scroll to next tabs';
 
     @property({ reflect: true })
     public override dir!: 'ltr' | 'rtl';
-
-    @property({ reflect: true, type: String, attribute: 'label-previous' })
-    public labelPrevious: string | undefined;
-
-    @property({ reflect: true, type: String, attribute: 'label-next' })
-    public labelNext: string | undefined;
 
     @state()
     private overflowState: TabsOverflowState = {
@@ -138,9 +138,9 @@ export class TabsOverflow extends SizedMixin(SpectrumElement) {
 
     protected override render(): TemplateResult {
         const { canScrollRight, canScrollLeft } = this.overflowState;
-        const ariaLabelPrevious =
-            this.labelPrevious || 'Scroll to previous tabs';
-        const ariaLabelNext = this.labelNext || 'Scroll to next tabs';
+        const ariaLabelPrevious = this.labelPrevious;
+        // this.labelPrevious || 'Scroll to previous tabs';
+        const ariaLabelNext = this.labelNext;
         return html`
             <div
                 class=${classMap({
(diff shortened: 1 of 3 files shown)


packages/tabs/src/TabsOverflow.ts
```diff
@@ -51,6 +51,12 @@ export class TabsOverflow extends SizedMixin(SpectrumElement) {
     @property({ reflect: true })
     public override dir!: 'ltr' | 'rtl';
 
+    @property({ reflect: true, type: String, attribute: 'label-previous' })
+    public labelPrevious: string | undefined;
```
audknu, line 55: if this can be defined can this be:
```suggestion
    public labelPrevious?: string;
```

audknu: If the next and previous buttons should not be keyboard accessible, is it necessary to introduce these labels? Instead, should we ensure that by keyboard, the next and previous buttons are not interacted with?

packages/tabs/src/TabsOverflow.ts
```diff
@@ -132,6 +138,9 @@ export class TabsOverflow extends SizedMixin(SpectrumElement) {
 
     protected override render(): TemplateResult {
         const { canScrollRight, canScrollLeft } = this.overflowState;
+        const ariaLabelPrevious = this.labelPrevious;
+        // this.labelPrevious || 'Scroll to previous tabs';
```
sarahszhou, line 142: Did you mean to keep this comment here?

packages/tabs/src/TabsOverflow.ts
blaabaer: I think that screenreader labeling can still be useful even without keyboard accessibility. (thinking of mobile users, who use tap controls and screenreaders read out what is tapped)

audknu (approved):

sarahszhou (approved):
Looks good to me!

Issue #3396, opened 2023-06-29
[Bug][a11y]: Table checkboxes to select row or Select All have no label


majornista:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Table

### Expected behavior

Checkboxes used to select the row within the table should be labelled by "Select " + the rowheader. Select All checkbox should be labelled "Select All."

### Actual behavior

Checkboxes within the table have no accessibility name and are indistinguishable from each other.

### How can we reproduce this issue?

3. Scroll to '....'
4. Check console
5. See error

najikahalsema:
related to https://github.com/adobe/spectrum-web-components/issues/3394

najikahalsema:
fixed via #4145

Issue #3394, opened 2023-06-29
[Bug][a11y]: Table examples show no way of declaring a cell as the row header


majornista:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Table

### Expected behavior

It should be possible to declare a column as the rowheader for a virtualized table, and the rowheader should be used to help label the checkbox used to select the row.

### Actual behavior

While it is possible to override the default role for `sp-table-cell` of `gridcell` with [`rowheader`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/rowheader_role), documentation and storybook examples show no way of declaring a cell as the row header for the row.
Being able to declare the rowheader is important for setting the label for the checkbox that selects a row.

Additional concerns: the `role` for `sp-table-cell` should be type safe, `'gridcell' | 'rowheader' | 'cell'`. 

### How can we reproduce this issue?

3. Scroll to '....'
4. Check console
5. See error

najikahalsema:
Tasks:
- [ ] Update the type for role to be typesafe
- [ ] Update documentation to include usage of the types 
- [ ] Ensure `rowheader` labels the checkbox

najikahalsema:
Fixed via #4145

PR #4145, opened 2024-03-07
fix(table): adding missing aria-labels

ingorichter:
## Description

The PR implements the following changes

- add `rowheader` role for `<sp-table-cell>`
- tables with `single` or `multiple` select will now have an `aria-label` attribute for the checkboxes in the table. There is an english default string for checkboxes that select a row is **Select** and for the checkbox selecting all rows the default is **Select All**
- The string that will be used for the checkboxes can be provided for the `<sp-table>` element and `<sp-table-cell>` supports an attribute to allow an explicit override for that row. This is intended for clients to support a localized version for the checkbox
- Clients can provide a localized version for **Select All** checkbox string for `<sp-table>` component

## Related issue(s)

This fix addresses

- #3394
- #3396 

## Motivation and context

Screen Readers were unable to distinguish the different checkboxes for each table row

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/checkbox/src/CheckboxMixin.ts
+++ b/packages/checkbox/src/CheckboxMixin.ts
@@ -45,13 +45,6 @@ export function CheckboxMixin<T extends Constructor<ReactiveElement>>(
         @property({ type: Boolean, reflect: true })
         public readonly = false;
 
-        /**
-         * An accessible label that describes the component.
-         * It will be applied to aria-label, but not visually rendered.
-         */
-        @property({ type: String })
-        public label?: string;
-
         @query('#input')
         inputElement!: HTMLInputElement;
 
@@ -78,9 +71,7 @@ export function CheckboxMixin<T extends Constructor<ReactiveElement>>(
         protected render(): TemplateResult {
             return html`
                 <input
-                    aria-label="${ifDefined(
-                        this.label || this.ariaLabel || undefined
-                    )}"
+                    aria-label="${ifDefined(this.ariaLabel || undefined)}"
                     id="input"
                     name=${ifDefined(this.name || undefined)}
                     type="checkbox"
--- a/packages/checkbox/src/CheckboxMixin.ts
+++ b/packages/checkbox/src/CheckboxMixin.ts
@@ -45,6 +45,13 @@ export function CheckboxMixin<T extends Constructor<ReactiveElement>>(
         @property({ type: Boolean, reflect: true })
         public readonly = false;
 
+        /**
+         * An accessible label that describes the component.
+         * It will be applied to aria-label, but not visually rendered.
+         */
+        @property({ type: String })
+        public label?: string;
+
         @query('#input')
         inputElement!: HTMLInputElement;
 
@@ -71,6 +78,9 @@ export function CheckboxMixin<T extends Constructor<ReactiveElement>>(
         protected render(): TemplateResult {
             return html`
                 <input
+                    aria-label="${ifDefined(
+                        this.label || this.ariaLabel || undefined
+                    )}"
                     id="input"
                     name=${ifDefined(this.name || undefined)}
                     type="checkbox"
--- a/packages/checkbox/stories/checkbox.stories.ts
+++ b/packages/checkbox/stories/checkbox.stories.ts
@@ -30,6 +30,12 @@ export const readonly = (): TemplateResult => {
     `;
 };
 
+export const WithLabel = (): TemplateResult => {
+    return html`
+        <sp-checkbox label="Checkbox Label">Checkbox</sp-checkbox>
+    `;
+};
+
 export const Indeterminate = (): TemplateResult => {
     return html`
         <sp-checkbox indeterminate>
--- a/packages/checkbox/stories/checkbox.stories.ts
+++ b/packages/checkbox/stories/checkbox.stories.ts
@@ -32,7 +32,7 @@ export const readonly = (): TemplateResult => {
 
 export const WithLabel = (): TemplateResult => {
     return html`
-        <sp-checkbox label="Checkbox Label">Checkbox</sp-checkbox>
+        <sp-checkbox aria-label="Explicit Checkbox Label">Checkbox</sp-checkbox>
     `;
 };
 
--- a/packages/table/README.md
+++ b/packages/table/README.md
@@ -873,7 +873,7 @@ For each table column you want to sort, use the `sortable` attribute in its resp
             cell2.textContent = `Row Item Beta ${item.date}`;
             cell3.textContent = `Index: ${index}`;
             return [cell1, cell2, cell3];
-        }
+        };
         table.addEventListener('sorted', (event) => {
             const { sortDirection, sortKey } = event.detail;
             items = items.sort((a, b) => {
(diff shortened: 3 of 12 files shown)


packages/table/src/Table.ts
```diff
@@ -157,6 +158,19 @@ export class Table extends SizedMixin(SpectrumElement, {
     @property({ type: Boolean, reflect: true })
     public quiet = false;
 
+    /**
+     * Specify the string that is used to provide a label for a row checkbox
+     */
+    @property({ type: String })
+    public selectRowString = 'Select';
```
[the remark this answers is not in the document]
audknu: To align with the existing pattern for including sub-labels for a SWC component, https://github.com/adobe/spectrum-web-components/blob/f89ea42fdb7b82ff05283f6cd871139f65117ae6/packages/color-area/src/ColorArea.ts#L64-L68, can this be renamed to `label-select-row` and `label-select-all-rows`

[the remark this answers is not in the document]
ingorichter: I couldn't find anything consistent. I can rename them to `selectRowLabel` and `selectAllRowsLabel` to remove the string part in the property name

packages/checkbox/src/CheckboxMixin.ts
```diff
@@ -71,6 +78,9 @@ export function CheckboxMixin<T extends Constructor<ReactiveElement>>(
         protected render(): TemplateResult {
             return html`
                 <input
+                    aria-label="${ifDefined(
+                        this.label || this.ariaLabel || undefined
```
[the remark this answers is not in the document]
ingorichter: Yes, the component supports ariaLabel implicitly. I can't take a screenshot of the popover with that information from my ide, but it redirects to a document explaining the internals https://wicg.github.io/aom/explainer.html#default-semantics-for-custom-elements-via-the-elementinternals-object

packages/checkbox/stories/checkbox.stories.ts
```diff
@@ -30,6 +30,12 @@ export const readonly = (): TemplateResult => {
     `;
 };
 
+export const WithLabel = (): TemplateResult => {
+    return html`
+        <sp-checkbox label="Checkbox Label">Checkbox</sp-checkbox>
+    `;
+};
+
 export const Indeterminate = (): TemplateResult => {
```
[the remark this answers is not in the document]
ingorichter: Without an explicit label, the Checkbox will be accessible and uses the text between the tags. With the label you can explicitly set the `aria-label` of the underlying `input` tag to support checkboxes without a provided label text between tags. For the table use case this is important, since there we only have empty sp-checkbox tag

packages/checkbox/src/CheckboxMixin.ts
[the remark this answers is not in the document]
sarahszhou: Will `this.ariaLabel` ever be defined? I don't see it as a property on `CheckboxMixin`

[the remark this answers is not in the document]
ingorichter: if you do <sp-checkbox aria-label="Test"></sp-checkbox> then it will be used. Perhaps it's better to rely only on `aria-label` instead of `label`. That won't require any changes on the checkbox at all. What do you think?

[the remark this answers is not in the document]
sarahszhou: Yeah, that sounds good! Might be cleaner that way.

packages/table/README.md
```diff
@@ -893,3 +893,116 @@ For each table column you want to sort, use the `sortable` attribute in its resp
...
+
+## Accessibility
+
+Tables can be created to allow the selection of single or multiple rows. In both cases, the checkbox in front of a row will have an `aria-label` that improves accessibility.
+
+The `<sp-table>` component is built to be accessible by default. The `aria-label` attribute is automatically assigned to the checkbox, providing accessibility support. The English default value for a row checkbox is 'Select'. The checkbox in the table header will have the value 'Select All'.
+
+Both strings can be customized by setting the `selectRowLabel` and `selectAllRowsLabel` attributes of the `<sp-table>` component. Providing selectRowLabel for the `<sp-table-row>` component will override the default value for that row. Providing selectAllRowsLabel for the `<sp-table>` component will override the default value for the row checkbox.
+
+### Combination of `selects`, `selectRowLabel` and `selectAllRowsLabel`
```
sarahszhou, line 905: Can `selects` be capitalized to be consistent with the table below? => `Selects`

packages/table/README.md
```diff
@@ -893,3 +893,116 @@ For each table column you want to sort, use the `sortable` attribute in its resp
...
+
+| Component    | selectRowLabel | selectAllRowsLabel | Selects  | Result                                                                                                                                             |
```
sarahszhou, line 907: Can `Selects` be in lower case to match the attribute?

packages/table/src/Table.ts
```diff
@@ -348,6 +363,22 @@ export class Table extends SizedMixin(SpectrumElement, {
...
+                // be used as the label for the checkbox. If not present, the default
+                // selectRowLabel will be used.
+                const rowHeader = row.querySelector('[role="rowheader"]');
+                if (rowHeader) {
+                    const cell = rowHeader as TableCell;
+                    checkbox.ariaLabel = cell.selectRowLabel
+                        ? cell.selectRowLabel
+                        : `${
+                              this.selectRowLabel
+                          } ${rowHeader.textContent?.trim()}`;
```
sarahszhou, line 378: Did we add a unit test for when `textContent` is undefined?

packages/table/src/TableCheckboxCell.ts
```diff
@@ -71,6 +72,7 @@ export class TableCheckboxCell extends SpectrumElement {
                 ?disabled=${this.disabled}
                 ?emphasized=${this.emphasized}
                 aria-hidden=${ifDefined(this.selectsSingle)}
+                aria-label="${this.ariaLabel || nothing}"
```
sarahszhou, line 75: Maybe we can just align on using
`ifDefined(this.ariaLabel || undefined)` OR `this.ariaLabel || nothing` 

I guess it feels inconsistent to me that here we are using `nothing` and above we are using the other variation

packages/table/README.md
ingorichter: sure, let me fix that

sarahszhou (approved):
Some small change requests here and there but overall LGTM!

packages/table/src/Table.ts
ingorichter: not yet. I will add a unit test. Thanks

sarahszhou:
I've re-run the build on the tests that are failing @ingorichter - still see some failing. Re-running

I think for the purposes of the hackathon, it might be OK to merge into acpc/hackathon-swc , even with tests failing, in case we want to demo any of this behavior. 

But, it might be good to investigate why some of your changes have affected the existing tests since the ultimate goal here is to get these changes into `main`.

packages/table/src/TableCheckboxCell.ts
ingorichter: I agree with you. The linter was complaining about the `IfDefined(this.ariaLabel)` and I found something in the lit document that was mentioning `this.ariaLabel || nothing` instead. I would rather keep the other variant, but I was unable to silence the linter

ingorichter: https://lit.dev/docs/templates/conditionals/#conditionally-rendering-nothing

ingorichter:
@sarahszhou thanks for re-running the failing tests. I'll keep an eye on it, since they don't fail locally

Issue #4107, opened 2024-03-01
[Bug]: ActionMenu closing prematurely on iPad pro device


godanny86:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

ActionMenu

### Expected behavior

The `sp-action-menu` should work on iPad pro and other larger touch devices. Either a mobile tray implementation should be used or the standard overlay should work. 

### Actual behavior

What we are observing with the latest version of SWC is that on a large touch device such as an iPad pro the action menu looks to be using the `sp-overlay` + `sp-popover`. However on touch end it seems to close the Overlay and the  result is the menu just flickers. On smaller devices the mobile tray implementation is triggered as expected and this doesn't happen.

Can be reproduced using the documentation and/or sandbox: https://opensource.adobe.com/spectrum-web-components/components/action-menu/, https://studio.webcomponents.dev/edit/OfeOwrikpHyrhWv6YvUI/package.json?p=stories

### Screenshots

Video below is a screen recording using Safari on an iPad Pro simulator:

https://github.com/adobe/spectrum-web-components/assets/8974514/a49f08a9-2675-4ac0-876b-53bee752217f

Sandbox:  https://studio.webcomponents.dev/edit/OfeOwrikpHyrhWv6YvUI/package.json?p=stories

### What browsers are you seeing the problem in?

Chrome, Safari

### How can we reproduce this issue?

1. Navigate to a site using the ActionMenu on a large touch device such as an iPad Pro
2. Try to open the menu by tapping it
3. Observe that the menu flickers and then closes on the end of the touch 

**note** we started seeing this issue after upgrading to SWC 0.45+

Westbrook:
@TarunAdobe would you have any availability to start investigating this, after we get that Slider PR landed?

PR #4150, opened 2024-03-07
fix(picker): use tray on all touch devices

TarunAdobe:
## Description

What we are observing with the latest version of SWC is that on a large touch device such as an iPad pro the action menu looks to be using the sp-overlay + sp-popover. However on touch end it seems to close the Overlay and the result is the menu just flickers. On smaller devices the mobile tray implementation is triggered as expected and this doesn't happen.

Can be reproduced using the documentation and/or sandbox: https://opensource.adobe.com/spectrum-web-components/components/action-menu/, https://studio.webcomponents.dev/edit/OfeOwrikpHyrhWv6YvUI/package.json?p=stories

## Related issue(s)

- #4107

## Motivation and context

The sp-action-menu should work on iPad pro and other larger touch devices. Either a mobile tray implementation should be used or the standard overlay should work.

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/picker/src/Picker.ts
+++ b/packages/picker/src/Picker.ts
@@ -45,10 +45,7 @@ import type {
     MenuItemChildren,
 } from '@spectrum-web-components/menu';
 import { Placement } from '@spectrum-web-components/overlay';
-import {
-    IS_MOBILE,
-    MatchMediaController,
-} from '@spectrum-web-components/reactive-controllers/src/MatchMedia.js';
+import { MatchMediaController } from '@spectrum-web-components/reactive-controllers/src/MatchMedia.js';
 import type { Overlay } from '@spectrum-web-components/overlay/src/Overlay.js';
 import type { FieldLabel } from '@spectrum-web-components/field-label';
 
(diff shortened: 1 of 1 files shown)


packages/picker/src/Picker.ts
```diff
@@ -61,7 +58,10 @@ const chevronClass = {
 
 export const DESCRIPTION_ID = 'option-picker';
 export class PickerBase extends SizedMixin(Focusable, { noDefaultSize: true }) {
-    protected isMobile = new MatchMediaController(this, IS_MOBILE);
+    protected isMobile = new MatchMediaController(
+        this,
+        '(hover: none) and (pointer: coarse)'
```
Rajdeepc, line 63: Can we put this in a constant to make it DRY

Rajdeepc (changes requested):

Westbrook (commented):
This is a good workaround, because it works. But, it is just a work around. As per https://spectrum.adobe.com/page/picker/#Menu-container only "mobile" devices should even be considered for Tray delivery of these menu items. With that in mind, as a short term fix, we can definitely be OK with this, but would you like to put some additional research into this before we commit to shipping something we know we'll need to revert in the future?

PR #4156, opened 2024-03-07
fix(table): Add aria-rowcount to virtualized tables

### LINKED ISSUE #3393: [Bug][a11y]: Virtualized Table should include aria-rowcount attribute

blaabaer:
## Description

The `sp-table` component can be virtualized, which only renders table rows that are currently visible in the viewport, which helps with performance on large tables. However, this can trip up screenreaders, which only detect and announce the rendered rows, and not the true size of the table. `aria-rowcount` is a screenreader attribute which announces the true unvirtualized size of the table.

This PR implements the automatic application of `aria-rowcount` for virtualized tables. 

## Related issue(s)

- fixes #3393 

## Motivation and context

Accuracy and efficacy of screenreader accessibility.

## How has this been tested?

-   [x] Run storybook - Table > Virtualized > [Any of these]
    1. Inspect the `sp-table` element
    2. Observe that the `aria-rowcount` value is correct.
-   [x] Visit docs site - Table > Virtualized
    1. Inspect the `sp-table` element
    2. Observe that the `aria-rowcount` value is correct.

image

image

Before my changes
image

After my changes
image

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/table/src/Table.ts
+++ b/packages/table/src/Table.ts
@@ -507,6 +507,9 @@ export class Table extends SizedMixin(SpectrumElement, {
                 }
             );
         }
+        if (this.items.length) {
+            this.setAttribute('aria-rowcount', `${this.items.length}`);
+        }
         const config: VirtualizeDirectiveConfig<Record<string, unknown>> = {
             items: this.items,
             renderItem: this.renderItem,
--- a/packages/table/src/Table.ts
+++ b/packages/table/src/Table.ts
@@ -508,6 +508,8 @@ export class Table extends SizedMixin(SpectrumElement, {
             );
         }
         if (this.items.length) {
+            // Ensures screenreaders can announce the true size of the table
+            // despite virtualization only rendering a subset.
             this.setAttribute('aria-rowcount', `${this.items.length}`);
         }
         const config: VirtualizeDirectiveConfig<Record<string, unknown>> = {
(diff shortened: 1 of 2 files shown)


packages/table/src/Table.ts
```diff
@@ -507,6 +507,11 @@ export class Table extends SizedMixin(SpectrumElement, {
                 }
             );
         }
+
+        // Ensures screenreaders can announce the true size of the table
+        // despite virtualization only rendering a subset of rows.
+        this.setAttribute('aria-rowcount', `${this.items.length}`);
```
Westbrook, line 513: Because this code path is only run when `items` has a `length` were this table to be used with filtering or searching and the length could conceivable become `0`, there is a change that this data would be out of date. Should we remove this attribute in the implied `else` at line 474?

Westbrook (approved):
LGTM! Thanks @blaabaer 🙇🏼

Issue #4151, opened 2024-03-07
[Bug]: Unable to set aria-label for sp-progress-bar and sp-progress-circle


lehelen19:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-progress-bar, sp-progress-circle

### Expected behavior

When setting the `aria-label` for a component, it should show up in the DOM and be able to be read by screen readers.

### Actual behavior

We are unable to set an `aria-label` for these components unless a `label` is explicitly set.

### What browsers are you seeing the problem in?

Firefox, Chrome, Safari, Microsoft Edge

### How can we reproduce this issue?

1. Go to the code playground for sp-progress-circle or sp-progress-bar: https://studio.webcomponents.dev/edit/LfliuY0UocICDCBr21uy?p=stories https://studio.webcomponents.dev/edit/WjcwSATAB2ZGd0FopIyP?p=stories
2. Fork and add an `aria-label` attribute to an instance of the component.
3. Inspect the element and observe the lack of an `aria-label`.

### Sample code that illustrates the problem

Code snippet:
`<sp-progress-circle indeterminate aria-label="progress-circle"></sp-progress-circle>`
DOM result:
`<sp-progress-circle indeterminate="" dir="ltr" size="m" role="progressbar"></sp-progress-circle>`

Code snippet:
`<sp-progress-circle indeterminate label="progress-circle"></sp-progress-circle>` 
DOM result:
`<sp-progress-circle indeterminate="" label="progress-circle" dir="ltr" size="m" role="progressbar" aria-label="progress-circle"></sp-progress-circle>`

Westbrook:
Possible approach from out of band conversation:
> I think we can add logic to the `else` that only removes `aria-label` if it's value and the value of `changes.get('label')` are the same. But, happy to have better suggestions on this as you dig deeper.

PR #4157, opened 2024-03-07
fix(progress-bar, progress-circle): remove aria-label only if set by label and label is empty

lehelen19:
## Description

Following @Westbrook's suggestion, remove `aria-label` when it matches the previously set (and now empty) `label`. This allows an `aria-label` to remain when it is set independently of the label, but correctly removes an `aria-label` when its corresponding `label` is removed.

## Related issue(s)

- fixes https://github.com/adobe/spectrum-web-components/issues/4151

## Motivation and context

We are unable to set an `aria-label` for `sp-progress-bar` or `sp-progress-circle` because it gets removed when `label` is not set. This lets us set an `aria-label` appropriately to meet accessibility standards.

## How has this been tested?

I tested adding and removing `label` and `aria-label` attributes in the storybooks for the affected components:

-   [ ] Component should support `aria-label`
    1. Add `aria-label` to component without `label`
    2. Observe `aria-label` in DOM as expected
-   [ ] Component should support `label` that sets `aria-label` and removes `aria-label` when `label` is removed
    1. Add `label` to component without `aria-label`
    2. Observe `label` and `aria-label` in DOM as expected
    3. Remove `label` from the DOM and observe `aria-label` is also removed as expected

I also added a few unit tests.

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/progress-bar/src/ProgressBar.ts
+++ b/packages/progress-bar/src/ProgressBar.ts
@@ -139,7 +139,9 @@ export class ProgressBar extends SizedMixin(
         if (changes.has('label')) {
             if (this.label.length) {
                 this.setAttribute('aria-label', this.label);
-            } else {
+            } else if (
+                changes.get('label') === this.getAttribute('aria-label')
+            ) {
                 this.removeAttribute('aria-label');
             }
         }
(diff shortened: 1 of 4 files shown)


Westbrook (approved):
Thanks @lehelen19 this looks great!

Issue #3406, opened 2023-06-30
[Bug][a11y]: Toast should have API for overriding the icon's alt text


majornista:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Toast

### Expected behavior

It should be possible to override the string used as the `label` for the `sp-icon` element in the Shadow DOM, both for i18n purposes and for clarity, because sometimes an "Warning" is not an "Error."

This relates to https://github.com/adobe/spectrum-web-components/issues/1975

### Actual behavior

The alt text for the icon displayed in the "negative", "positive", and "info" Toast variants is hardcoded and cannot be overridden as "Error", "Success", and "Information" respectively.

### How can we reproduce this issue?

3. Scroll to '....'
4. Check console
5. See error

PR #4159, opened 2024-03-08
fix(toast): Toast should have API for overriding the icon's alt text

susmithayenugula:
Added `label` prop to Toast that allows customization of the `sp-icon`'s label.

## Description

- Previously, the label for the `sp-icon` used in a `sp-toast` was hardcoded to `Success`, `Error` and `Information` for `positive`, `negative` and `info` toasts.
- The addition of the label prop allows one to customize the string. This is useful when we want to use the string `Warning` for a `negative` toast.
- Update storybook to add `label` prop and set value of this prop.

## Related issue(s)
fixes https://github.com/adobe/spectrum-web-components/issues/3406

## Motivation and context

The addition of the label prop allows one to customize the string. This is useful when we want to use the string `Warning` for a `negative` toast.

## How has this been tested?
1. Navigate to storybook locally
2. Go to the Toast component in the sidebar.
5. In DevTools, select the icon with the inspect tool. `sp-icon-x` is in `sp-toast`'s shadow-root.
6. Notice in the `Controls` tab, locate  the new attribute `label`.
4. Type in any string and re-render the toast using `Remount component` in the top bar.
7. Verify that the string typed in step 4 is the present under the `label` prop of `sp-icon-x`. 

Screenshot 2024-03-07 at 10 19 37 PM

## Types of changes

## Checklist

## Best practices


packages/icon/src/IconBase.ts
```diff
@@ -36,6 +36,7 @@ export class IconBase extends SpectrumElement {
         if (changes.has('label')) {
             if (this.label) {
                 this.removeAttribute('aria-hidden');
+                this.setAttribute('aria-label', this.label);
```
susmithayenugula, line 39: Per the [readme](https://github.com/adobe/spectrum-web-components/tree/main/packages/icon#accessibility) for icon:

aria-hidden is set to true by default for Icons. The label attribute suppresses this and adds the label text as the aria-label of the icon.

This was not implement so fixed it in this PR

packages/toast/test/toast.test.ts
```diff
@@ -56,14 +56,13 @@ describe('Toast', () => {
...
-                    <sp-toast variant=${variant} open>
+                    <sp-toast variant=${variant} open label="testLabel">
                         This toast is of the \`${variant}\` variant.
                     </sp-toast>
                 `
             );
-
             await elementUpdated(el);
-
+            await expect(el.getAttribute('label')).to.equal('testLabel');
```
susmithayenugula, line 65: I couldn't check if the `sp-icon`'s label in this test since `sp-icon` isn't exposed in `sp-toast`

packages/toast/src/Toast.ts
```diff
@@ -116,28 +116,38 @@ export class Toast extends FocusVisiblePolyfillMixin(SpectrumElement) {
...
+    }
+
+    public get label(): string {
+        return this._label;
+    }
+
     private _variant: ToastVariants = '';
 
-    private renderIcon(variant: string): TemplateResult {
+    private _label: string = '';
```
susmithayenugula, line 132: not sure but seems like changing the default string to be an empty string as opposed to the existing `Success`, `Error`, and `Information` seems like it could break things?

audknu: Agreed, if the consumer does not define a label, should we fall back to the existing strings?
```suggestion
    <sp-icon-info label=${label === "" ? "Information" : label} class="type"></sp-icon-info>
```

sarahszhou: Yeah falling back to the default makes sense to me!

packages/toast/test/toast.test.ts
sarahszhou: Could you try getting it through the `shadowRoot`? So: 
```
el.shadowRoot.querySelector(
   sp-icon-checkmark-circle
) as HTMLElement;
```

I see that used in https://github.com/adobe/spectrum-web-components/blob/main/packages/tray/test/tray.test.ts#L105

packages/icon/src/IconBase.ts
ingorichter: I wonder why we don't use `aria-label` from the component tag instead of introducing a new `label` property? I removed the custom `label` property and use the `aria-label` that can be specified at the component level

Westbrook: At the API level we _generally_ use `label` because sometimes that element must be labeled at the host (which would be the same as using `aria-label`, but other times that label needs to be applied to something _within_ the host, and forwarding `aria-label` is a tricky accessibility/timing cup game of dispair. Having a single entry point across as many elements as possible intends that it is less likely that a consumer makes the mistake of using the wrong one when switching back and forth between them over the course of developing a larger application.

packages/toast/test/toast.test.ts
susmithayenugula: @sarahszhou Thank you, worked perfectly!

Rajdeepc:
Hi @susmithayenugula This PR is pointing to `acpc/hackathon-swc`. Please catch this up to `main` and we can take this forward.

susmithayenugula:
@Rajdeepc I'll take care of this today, thank you!

susmithayenugula:
Hi @Rajdeepc , I've updated the PR as requested. Please take a look!

packages/icon/src/IconBase.ts
```diff
@@ -26,7 +26,7 @@ export class IconBase extends SpectrumElement {
         return [iconStyles];
     }
 
-    @property()
+    @property({ type: String, reflect: true })
```
Westbrook, line 29: What is the benefit of reflecting here?

packages/icon/src/IconBase.ts
Westbrook, line 39: Great add. However, this whole `updated` method probably should be moved to `Icon.ts` instead. Some `IconBase.ts` extensions will pass the label data into an icon literal, which also manages `aria-hidden`, however these management would be beneficial to the `<sp-icon>` element as it accepts a slotted icon rather than owning icon DOM itself.

packages/toast/src/Toast.ts
```diff
@@ -116,28 +116,44 @@ export class Toast extends FocusVisiblePolyfillMixin(SpectrumElement) {
...
+    @property({ type: String })
+    public set label(label: string) {
+        this._label = label;
+        this.setAttribute('label', label);
+        this.requestUpdate('label', label);
+    }
+
+    public get label(): string {
+        return this._label;
+    }
```
Westbrook, line 128: Can we not use a simple decorator here:

```
@property()
label = '';
```
And then avoid all the rest of the manual administration?

As far as naming, this isn't so much the `label` of the toast as much as the label of the icon. Should we have a more specific attribute/property name that represents that? Which ever way we go, it'll be good to have a JSDoc description as to how the value is leveraged within the element.

packages/toast/src/Toast.ts
```diff
@@ -116,28 +116,44 @@ export class Toast extends FocusVisiblePolyfillMixin(SpectrumElement) {
...
-    private renderIcon(variant: string): TemplateResult {
+    private _label: string = '';
+
+    private renderIcon(variant: string, label: string): TemplateResult {
         switch (variant) {
             case 'info':
                 return html`
                     <sp-icon-info
-                        label="Information"
+                        label=${label === '' ? 'Information' : label}
```
Westbrook, line 139: `''` is falsy, so this can be simplified to `label || 'Information'`.

packages/toast/test/toast.test.ts
```diff
@@ -56,15 +56,82 @@ describe('Toast', () => {
...
+                    case 'negative':
+                    case 'error': // deprecated
+                    case 'warning': // deprecated
+                        return 'sp-icon-alert';
+                    case 'positive':
+                        return 'sp-icon-checkmark-circle';
+                    default:
+                        return '';
+                }
+            };
```
Westbrook, line 91: Can this be alleviated by the fact that the icon always has `class="type"`?

packages/toast/test/toast.test.ts
```diff
@@ -56,15 +56,82 @@ describe('Toast', () => {
...
+                    break;
+                case 'positive':
+                    iconType = 'sp-icon-checkmark-circle';
+                    defaultLabel = 'Success';
+                    break;
+                default:
+                    iconType = '';
+                    defaultLabel = '';
+                    break;
+            }
```
Westbrook, line 130: Some of this can be simplified via `class="type"` being on the icon element. The rest of the data feels like it should be externalized some how, but I don't have a good suggestion on cleaning it up just now. Maybe an map of some sort outside of the test case?

TarunAdobe:
Hey @susmithayenugula. Let us know if you still have the time and availability to get this over the line. We would really love to have this contribution in our library.

packages/icon/src/IconBase.ts
caseyisonit: removing as its not included in the issue and should be filed separately with more context

caseyisonit:
This will be resolved in the PR #4944. Closing this PR.

PR #4162, opened 2024-03-08
docs(badge): updated non semantics docs for mod override

Rajdeepc:
## Description

Updated the `--mod-badge-background-color-default` override for `seafoam,indigo,purple,fuchsia,magenta,yellow` non semantic variants property which should be
```
--mod-badge-background-color-seafoam
--mod-badge-background-color-indigo
--mod-badge-background-color-purple
--mod-badge-background-color-fuchsia
--mod-badge-background-color-magenta
--mod-badge-background-color-yellow
```

## Related issue(s)

-

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ executors:
 parameters:
     current_golden_images_hash:
         type: string
-        default: a3a63503e5584396e8b34ec0b02e8a1a26ebc1ae
+        default: d95e1f5e29f6a847ce8190ad6ad6028456707b71
     wireit_cache_name:
         type: string
         default: wireit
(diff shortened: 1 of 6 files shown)


Westbrook:
Interesting. I wish I could keep the historical trajectory of every component in my mind, but alas here is one that I can't. I can say how we got to where we are now, but it seems that being these `variants` are "official" that we don't need those customizations at all to achieve the colors desired for those Badges. Could we remove them instead? Also, it seem that the same content in the Storybook could use the same.

Related, it seems that there are about 9 variants that we aren't supporting here: https://github.com/adobe/spectrum-web-components/blob/main/packages/badge/src/spectrum-badge.css#L269-L274 Would you be able to include those in the CSS processing and the various pieces of documentation?

Westbrook (approved):
LGTM. Catching up `main` and then let’s merge this in!

PR #4180, opened 2024-03-14
chore: fixes #4174 by removing loose ^ selector in internal dependencies

### LINKED ISSUE #4174: [Bug]: Transitive internal dependencies versions are too loose
### Code of conduct

### Impacted component(s)

all

### Expected behavior

We should be able to install arbitrary sets of SWC components using hard pinned versions, e.g. `0.41.1` such that all internal dependencies thereby brought in by those components will internally resolve to not cause conflictings SWC component versions.

### Actual behavior

When installing the following set of dependencies:
```
"@spectrum-web-components/menu": "0.41.1",
"@spectrum-web-components/action-menu": "0.41.1"
```
Its possible for package resolution to result in a duplicate install of `@spectrum-web-components/menu` at `0.41.2`.

This is caused by the internal dependency in `action-menu` on `picker` with the semver `^0.41.1` being specified. This allows `0.41.2` to be installed for `picker` which in turn brings in `menu` at `^0.41.2`. This second transitive dependencies on `menu` is then more restrictive than the `menu` installed by the app and thus we end up with two versions in the repository.

Suggest making internal dependencies use the `0.41.1` form of semver (with no `^`) to ensure that when you install a component at a specific version you only get packages at that version installed. This will avoid unexpected installation of newer versions which can cause such conflicts.

### Screenshots

_No response_

_No response_

1. Add `@spectrum-web-components/menu` and `@spectrum-web-components/action-menu` to a project with a fixed version, e.g. `0.41.1`
2. Install using `pnpm`
3. Inspect`pnpm-lock.yaml` and confirm that two versions of `menu` are installed.

_No response_

_No response_

benjamind:
## Description

Updates all internal version dependencies to use strict versions rather than the looser `^` selector, this is to avoid unexpected installation of newer versions of transitive dependencies.

## Related issue(s)
#4174 

## Motivation and context

When installing a specific version of spectrum-web-components packages we would expect all internal dependencies to be the same requested version. This helps us also avoid unexpected duplication of mismatched versions.

## How has this been tested?

The effect of this change was tested in a separate repository by using `pnpmfile.js` to rewrite the `dependencies` fields of SWC components as they installed. This fixed all transitive dependency mismatched errors.

## Types of changes

## Checklist

## Best practices


caseyisonit:
closing, resolved in #5105

PR #4181, opened 2024-03-14
Selection management fixes

Westbrook:
## Description
- force Button content to be `pointer-events: none` so that the host can be the target for all click interactions, particularly ones leading to selection
- ensure the Action Group can manage selection state through multiple slots

## How has this been tested?
-   [ ] _Test case 1_
    1. See that added Action Group tests handle "icons" that are not managed by SWC and therefor do not prevent their own `pointer-events`
-   [ ] _Test case 2_
    1. See that added Action Group tests handle selection through multiple slots; on of their own and own of their parent's

## Types of changes

## Checklist

### DIFF
--- a/packages/action-group/test/action-group.test.ts
+++ b/packages/action-group/test/action-group.test.ts
@@ -699,7 +699,7 @@ describe('ActionGroup', () => {
         expect(secondButton.selected, 'second button selected').to.be.true;
     });
 
-    it('does not allow interaction with child content to interupt the selection mechanism', async () => {
+    it('does not allow interaction with child content to interrupt the selection mechanism', async () => {
         const el = await singleSelectedActionGroup([]);
         await elementUpdated(el);
         expect(el.selected.length).to.equal(0);
@@ -723,8 +723,8 @@ describe('ActionGroup', () => {
 
         expect(el.selected.length).to.equal(1);
         expect(el.selected).to.deep.equal(['first']);
-        expect(firstButton.selected, 'first button not selected').to.be.true;
-        expect(secondButton.selected, 'second button selected').to.be.false;
+        expect(firstButton.selected, 'first button selected').to.be.true;
+        expect(secondButton.selected, 'second button not selected').to.be.false;
 
         const rect = icon.getBoundingClientRect();
         await sendMouse({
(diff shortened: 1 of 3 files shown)


packages/action-group/test/action-group.test.ts
```diff
@@ -659,6 +699,54 @@ describe('ActionGroup', () => {
         expect(secondButton.selected, 'second button selected').to.be.true;
     });
 
+    it('does not allow interaction with child content to interupt the selection mechanism', async () => {
```
hunterloftis, line 702:
```suggestion
    it('does not allow interaction with child content to interrupt the selection mechanism', async () => {
```

packages/action-group/test/action-group.test.ts
```diff
@@ -659,6 +699,54 @@ describe('ActionGroup', () => {
...
+        expect(firstButton.selected, 'first button not selected').to.be.false;
+        expect(secondButton.selected, 'second button selected').to.be.true;
+
+        firstButton.click();
+        await elementUpdated(el);
+
+        expect(el.selected.length).to.equal(1);
+        expect(el.selected).to.deep.equal(['first']);
+        expect(firstButton.selected, 'first button not selected').to.be.true;
+        expect(secondButton.selected, 'second button selected').to.be.false;
```
hunterloftis, line 727:
```suggestion
        expect(firstButton.selected, 'first button selected').to.be.true;
        expect(secondButton.selected, 'second button not selected').to.be.false;
```

packages/action-group/test/action-group.test.ts
```diff
@@ -659,6 +699,54 @@ describe('ActionGroup', () => {
...
+                {
+                    type: 'click',
+                    position: [
+                        rect.left + rect.width / 2,
+                        rect.top + rect.height / 2,
+                    ],
+                },
+            ],
+        });
+        icon.click();
```
hunterloftis, line 741: Just FMI, what was happening before on icon click?

hunterloftis (approved):

packages/action-group/test/action-group.test.ts
Westbrook: It would ignore the click. See this early exit in the handler:
```
const target = event.target as ActionButton;
if (typeof target.value === 'undefined') {
    return;
}
```
When the click event happens on the Icon then it won't have a value and it will return. There's likely other issues to handle therein, but I was focusing on the immediate error that we had discovered.

PR #4199, opened 2024-03-21
[DO NOT MERGE]: S2 delegates

TarunAdobe:
## Description

Introducing a new delegate based pattern to manage specific controllers based on the current spectrum version.

## Related issue(s)

-

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices


Westbrook:
If you catch this up to main, do the no change diffs go away?

packages/button/src/ButtonBase.ts
```diff
@@ -32,6 +32,10 @@ import buttonStyles from './button-base.css.js';
 export class ButtonBase extends ObserveSlotText(LikeAnchor(Focusable), '', [
     'sp-overlay,sp-tooltip',
 ]) {
+    spectrumConfig = {
+        downstate: true,
```
Westbrook, line 36: We likely want the value here not to be a boolean but an array of values that could include the current "system", e.g.:
```suggestion
        downstate: ['spectrum2'],
```

Westbrook: This way, some functionality can be delivered a la:
```
    downstate: ['spectrum1', 'spectrum2']
```

We should work to include these values an enums, to simplify sharing and consumption.

TarunAdobe: Great idea! I'll update it accordingly

TarunAdobe:
> If you catch this up to main, do the no change diffs go away?

No I don't think that'd be the case as I've manually created fake stylesheets for S2 theme so this PR isn't mergeable

Rajdeepc:
> If you catch this up to main, do the no change diffs go away?

I think this must be due to `git add .` . There are many stale files went in! 
cc @TarunAdobe

Rajdeepc:
@TarunAdobe Can we close this?

TarunAdobe:
Closing this as we have decided to make separate PRs for individual feature

Issue #4026, opened 2024-02-12
[Bug]: Add the posibility to inject new icons in toasts


AndreiBaicu26:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-toast

### Expected behavior

The express app uses different icons than spectrum classic. Additionally, there are requests for a "premium" toast that involves a "premium" icon. 
The toast component should provide a way to add custom icons. 
Example : https://www.figma.com/file/k5FYSaFhxiLUuPjvWfPSpi/Express-DS--%3E-Atoms?node-id=13%3A14026&mode=dev

### Actual behavior

Currently, the icon is rendered only by the toast element.

Westbrook:
Express Icon customization _generally_ should occur with less developer interference. See https://github.com/adobe/spectrum-web-components/issues/2845. The team has yet to commit resourcing to completing the work, but the way that it addresses all icons and not just icons in one pattern may change the priority of work you are addressing here?

If the work above is completed, the idea of the "premium" icon becomes more clear. The idea of a single "variant" wherein the icon is fully customizable simplified the element API here for a diversity of consumers. Wether that is a _new_ variant or the _default_ variant is a good conversation to have with design.

AndreiBaicu26:
Closing this issue as https://github.com/adobe/spectrum-web-components/pull/4200 will add the capability to automatically switch to express icons when neede 💯

PR #4200, opened 2024-03-21
feat(icons-workflow): add support for Express workflow icons

Rocss:
## Description

Initial PR: https://github.com/adobe/spectrum-web-components/pull/3169
This PR is a follow up of the one above.  This new feature allows consumers to use the [Express icons](https://a4u-gateway.corp.adobe.com/bundles/62a0dcda6a689c08a6c091e4/overview) along with the Express theme variant by leveraging [conditional exports](https://nodejs.org/api/packages.html#conditional-exports).

To preview this feature you can go ahead and change the export conditions in storybook's `main.js` by adding **express** string as such:
```
exportConditions: ['browser', 'development', 'express'],
```

In a nutshell it:
- Adds `@adobe/spectrum-css-ccx-workflow-icons` which comes with Express icons.
- Modifies the build process of `icons-workflow` package to also process the Express icons.
- Adds entries in `icons-workflow/package.json` to specify where to find icons based on theme used (`express` vs `default`)

**Very important**: 
There may be some icons (5 icons at the moment) which are present in Express icons package but not in the Spectrum one. In this PR we are making available **only the icons that do have a Spectrum equivalent**. Rest will be ignored.

**Pretty important**: 
There are around ~800 Spectrum icons and around ~200 Express icons.
Therefore, since not all the Spectrum icons have an equivalent in Express, if you are using `express` export condition and you're looking for an icon that is not in the [Express icons package](https://a4u-gateway.corp.adobe.com/bundles/62a0dcda6a689c08a6c091e4/overview), you'll get the fallback, which is the Spectrum one.

## Related issue(s)

- https://github.com/adobe/spectrum-web-components/issues/2845

## Motivation and context

We should add built-in support for [Spectrum Express Workflow Icons](https://github.com/adobe/spectrum-css-workflow-icons/tree/main/dist/spectrum-css-ccx-workflow-icons). In this way we can continue to close the gap on full support for the Express variant of the Spectrum design language.

## How has this been tested?

-   [x] Toggling between "express" and "spectrum" icons
    1. Change the `exportConditions` of the storybook by appending "express" in the array: https://github.com/adobe/spectrum-web-components/blob/rocss/express-workflow-icons/storybook/main.js#L36
    2. Open storybook
    3. Compare it with https://opensource.adobe.com/spectrum-web-components/storybook/index.html?path=/story/icons-workflow--elements
    4. Observe there are icons such as `sp-icon-delete` which look different.  Indeed there are a lot which did not change... because there are ~800 spectrum and only ~200 express icons

## Types of changes

## Checklist

## Best practices


Rocss:
Is it possible the build process runs different on the CI and I am overlooking something, @Westbrook? I don't want to say "it works on my machine" but this build passes on my local. From the logs looks like the `icons.ts` file is not being populated in via the `postbuild` script.

Westbrook:
@Rocss I'd start by running `git clean -dfX` locally and seeing if the build works in that case. I'd venture to say that something is in the cache that shouldn't be or out of the cache that should be, but it would be hard to say what without digging deeper on this. It feels like it would be something in https://github.com/adobe/spectrum-web-components/pull/4200/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R347-R353 but that could be a red herring. Let me know if that doesn't get you moving in the right direction and we can take a look together tomorrow or so.

Westbrook:
With this change in Storybook, I am able to see the change from the previous delivery:

image

To the new delivery:

image

Which implies that this approach (is appropriate) would resolve #4026. @AndreiBaicu26 can you confirm this as well?

Westbrook:
I'd love any thoughts you might have on applying this new condition to https://github.com/adobe/spectrum-web-components/blob/main/.circleci/config.yml#L66 so that when testing an Express theme the intended icons were available in that context.

Westbrook:
In preparation for next steps:

@Rocss do you think your project is in a place to take a pre-release of this change to test therein? It "working" and it "working enough to support major products" are two different things, so if there are other testing paths that you would be interested in, any insight in this area would be greatly appreciated, too!

Rocss:
@Westbrook Fair point, I'll discuss within the team the possibility to test this as a pre-release in Express.

Rocss:
Update:
I'm waiting for a decision from design on whether Express will be switching to Spectrum 2 default icons, as we're considering the work in this PR to no longer be relevant in the context of unification.

I'm also thinking of how the complexity of the build would change if Spectrum 2 will have different sets of icons for different themes as well. We'll have to basically support 4 icon repositories (2 for Spectrum Express and 2 for Spectrum classic), as SWC will support both S1 and S2 for a while.

PR #4202, opened 2024-03-21
build: exclude erroneously interpreted event dispatch from CEM management

Westbrook:
## Description
The `SlottableRequestEvent` gets interpreted by CEM incorrectly and fails the React build when attempting to make a release. This ignores the automatic processing of the event and places it manually in the JSDocs.

## How has this been tested?
-   [ ] _Test case 1_
    1. Run `build:react`
    2. See that the process completes.

## Types of changes

## Checklist

### DIFF
--- a/packages/overlay/src/Overlay.ts
+++ b/packages/overlay/src/Overlay.ts
@@ -77,6 +77,7 @@ export const strategies = {
  *
  * @fires sp-opened - announces that an overlay has completed any entry animations
  * @fires sp-closed - announce that an overlay has compelted any exit animations
+ * @fires slottable-request - requests to add or remove slottable content
  */
 export class Overlay extends OverlayFeatures {
     static override styles = [styles];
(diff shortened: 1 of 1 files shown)


packages/overlay/src/Overlay.ts
```diff
@@ -536,6 +537,9 @@ export class Overlay extends OverlayFeatures {
         if (!this.open) {
             document.body.offsetHeight;
         }
+        /**
```
Rajdeepc, line 540: Why can't we move this custom Event to surface to the react events?

Westbrook: There’s not currently tooling to surface enough data to do so. We need to spend some time seeing if we can support brining that to life, but I wanted to make sure the library was in a publishable form while we investigate that. For the last two releases I had corrected it manually, but I can never be sure that it really worked as we don’t have any (post publication) testing for the wrappers.

Rajdeepc (approved):

PR #4210, opened 2024-03-22
fix(help-text): apply aria-live to ensure full help text is read to user

### LINKED ISSUE #3377: [Bug]: sp-help-text error message suppressed when continue typing
### Code of conduct

### Impacted component(s)

sp-help-text

### Expected behavior

When sp-help-text is used in the negative-help-text slot it is expected that the error message will be read to the user. This works great if the user simply presses a single keypress but when the user types more info into the field the error message is supressed.

### Actual behavior

The error message doesn't get read as it its reading is suppressed by the next keypress.

### Screenshots

_No response_

Firefox, Chrome, Microsoft Edge

1. Start JAWS or NVDA
2. Go to https://studio.webcomponents.dev/preview/Tea8IMwcfLarfCQwSnxa/src?branch=Textfield-invalid-text@7fBWg9Y3wmUwOKofZzrLDd2SRAr2&p=stories
3. Click on the first name field
4. Type 3
5. Note the error message is read
6. Delete the number and type some letters into the field
7. Type 33 into the field
8. Note the error message is not read.

_No response_

_No response_

Westbrook:
## Description
Use `aria-live="assertive"` to ensure the Help Text content (when changing) has higher screen reader priority than Textfield input.

## Related issue(s)
- fixes #3377

## How has this been tested?
-   [ ] _Test case 1_
    1. Go [here](https://assertive-help-text--spectrum-web-components.netlify.app/components/textfield/#help-text)
    2. Turn on your screen reader
    3. Focus the `Stay "Positive"` text field
    4. Type or delete anything and then keep typing
    5. Ensure that the screen reader reads ALL of the updated help text when moving on and off of the value `Positive`

## Types of changes

## Checklist


packages/help-text/src/HelpTextManager.ts
```diff
@@ -43,7 +43,10 @@ export class HelpTextManager {
         // `pass-through-help-text-${this.instanceCount}` makes the slot effectively unreachable from
         // the outside allowing the `help-text` slot to be preferred while `negative === false`.
         return html`
-            <div id=${ifDefined(this.isInternal ? this.id : undefined)}>
+            <div
+                id=${ifDefined(this.isInternal ? this.id : undefined)}
```
Rajdeepc, line 47: Does a truthy `this.isInternal` always specify that an `id` will be present?

Westbrook: Yes it does. And, while it's _technically_ possible that a custom element consumer would _unset_ the value of ID, but that would be highly irregular. Without negative intent, the `id` as set in the constructor (note that this is a helper class and not a custom element class) would always be available for use here.

As this isn't new functionality, if there's a concern we want to address here, it would be better to track it in an issue for a later update.

hunterloftis:
Should I expect "Tell us how you are feeling today" and/or "Please be positive?"

I get neither - instead, VoiceOver gives me:
- Test
- More content available, You are currently on a text field.
- (Tab off, then on)
- Test, text contents selected, invalid data, edit.

nikkimk (approved):
It worked, but we should have some examples of how to use it accessibly in the docs.

Rajdeepc:
@nikkimk Would you like to take this up?

Rajdeepc (commented):
This looks very clean now and helpful! Let me know when it will be ready for review. Thanks

Rajdeepc (approved):
Good work!

Issue #3973, opened 2024-01-29
[Bug]: Slider with `variant="filled"` AND `fill-start` fills in unexpected ways.


Westbrook:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Slider

### Expected behavior

When using the Slider like so:

```html
<sp-slider
  variant="filled"
  fill-start
></sp-slider>
```

I'd expect the fill to start at 50% of the track and fill from there to the current handle location.

### Actual behavior

The fill starts at 0 and goes to the handle, if the handle is less that 50% it will continue through the handle to the 50% point.

### Screenshots

<img width="754" alt="image" src="https://github.com/adobe/spectrum-web-components/assets/1156657/4799617d-0703-4be1-910b-d729e12ebedc">

PR #4217, opened 2024-03-26
fix(slider): no change in background for variant filled and fill start

Rajdeepc:
## Description
This PR adds a condition `this.fillStart === undefined` to the set `variant` method in order to address a specific issue. The condition ensures that the variant attribute is only set if the fillStart property is undefined, preventing unintended behavior when both variant and `fillStart` attributes are present simultaneously.

## Changes
Added a condition `this.fillStart === undefined `to the set variant method.
Updated the logic to set or remove the variant attribute based on the new condition.

## Related issue(s)

- #3973 

## How has this been tested?

```
<sp-slider
    max="20"
    variant="filled"
    fill-start
    min="0"
    value="10"
    step="1"
></sp-slider>
```

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/slider/src/Slider.ts
+++ b/packages/slider/src/Slider.ts
@@ -101,7 +101,7 @@ export class Slider extends SizedMixin(ObserveSlotText(SliderHandle, ''), {
         if (variant === this.variant) {
             return;
         }
-        if (variants.includes(variant)) {
+        if (variants.includes(variant) && this.fillStart === undefined) {
             this.setAttribute('variant', variant);
             this._variant = variant;
         } else {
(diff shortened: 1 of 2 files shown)


packages/slider/src/Slider.ts
```diff
@@ -101,7 +101,7 @@ export class Slider extends SizedMixin(ObserveSlotText(SliderHandle, ''), {
         if (variant === this.variant) {
             return;
         }
-        if (variants.includes(variant)) {
+        if (variants.includes(variant) && this.fillStart === undefined) {
```
Westbrook, line 104: We'll need text in the readme as to the prioritization here.

Westbrook (commented):
Would it be useful to have a VRT that intersected with this change to ensure we do not regress on this?

Westbrook (approved):
LGTM! Thanks for clearing this intersection up for consumers. 🙇🏼

PR #4226, opened 2024-03-27
chore(menu): support delivering submenus via directives

Westbrook:
## Description
Expand use of `slottable-request` based laziness to support submenus.

## How has this been tested?
-   [ ] _Test case 1_
    1. Go [here](https://submenu-directive--spectrum-web-components.netlify.app/storybook/?path=/story/action-menu--directive)
    2. See that the submenu can be activated with the mouse or the keyboard
    3. When using the keyboard confirm that the "visible focus" returns to the submenu host, not the first menu item in the parent menu

## Types of changes

## Checklist

### DIFF
--- a/packages/menu/src/MenuItem.ts
+++ b/packages/menu/src/MenuItem.ts
@@ -39,6 +39,7 @@ import checkmarkStyles from '@spectrum-web-components/icon/src/spectrum-icon-che
 import type { Menu } from './Menu.js';
 import { MutationController } from '@lit-labs/observers/mutation-controller.js';
 import type { Overlay } from '@spectrum-web-components/overlay';
+import { SlottableRequestEvent } from '@spectrum-web-components/overlay/src/slottable-request-event.js';
 
 /**
  * Duration during which a pointing device can leave an `<sp-menu-item>` element
@@ -164,6 +165,8 @@ export class MenuItem extends LikeAnchor(
     @query('sp-overlay')
     public overlayElement!: Overlay;
 
+    private submenuElement?: HTMLElement;
+
     public override get focusElement(): HTMLElement {
         return this;
     }
@@ -210,7 +213,14 @@ export class MenuItem extends LikeAnchor(
                 childList: true,
                 subtree: true,
             },
-            callback: () => {
+            callback: (mutations) => {
+                const isSubmenu = mutations.every(
+                    (mutation) =>
+                        (mutation.target as HTMLElement).slot === 'submenu'
+                );
+                if (isSubmenu) {
+                    return;
+                }
                 this.breakItemChildrenCache();
             },
         });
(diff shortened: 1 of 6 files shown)


najikahalsema:
The submenu flickers a little when I open it with my mouse. 🤔 Does that happen for you? I'm assuming "no"...

https://github.com/adobe/spectrum-web-components/assets/44980010/40b6a261-6b67-4736-b21a-0f1b6e61b34c

packages/overlay/src/Overlay.ts
```diff
@@ -534,6 +540,9 @@ export class Overlay extends OverlayFeatures {
     }
 
     protected override requestSlottable(): void {
+        if (this.lastRequestSlottableState === this.open) {
+            return;
+        }
```
Westbrook, line 545: Great catch on the stutter, @najikahalsema, we should use broken browsers more often? This will prevent duplicate `slottable-request` events from being dispatched with the same `data`, which appears to be what was causing what you captured, even if it was really difficult to run into normally.

packages/menu/src/MenuItem.ts
```diff
@@ -39,6 +39,7 @@ import checkmarkStyles from '@spectrum-web-components/icon/src/spectrum-icon-che
 import type { Menu } from './Menu.js';
 import { MutationController } from '@lit-labs/observers/mutation-controller.js';
 import type { Overlay } from '@spectrum-web-components/overlay';
+import { SlottableRequestEvent } from '@spectrum-web-components/overlay/src/slottable-request-event.js';
```
Rajdeepc, line 42: I am not sure whether we have expanded the readme to include this usage but we can definately add some context on this for the consumers in a good document on how and when to use it?

packages/menu/src/MenuItem.ts
```diff
@@ -343,10 +360,10 @@ export class MenuItem extends LikeAnchor(
     }
 
     protected manageSubmenu(event: Event & { target: HTMLSlotElement }): void {
-        const assignedElements = event.target.assignedElements({
+        this.submenuElement = event.target.assignedElements({
             flatten: true,
-        });
-        this.hasSubmenu = !!assignedElements.length;
+        })[0] as HTMLElement;
```
Rajdeepc, line 365: Can we keep a null pointer check here?

packages/menu/src/MenuItem.ts
Westbrook: This unlocks the last bit of scope for the initial test interaction with the PS team. Once that test cycle is complete, the last piece of confirmation before documentation/"official release" will be confirming how the directives work in partnership with requests for content in slots other than the default slot AND THEN I will start writing documentation with all available speed. One part of that documentation will also be internal outlining path for future use of this form of "enhancement testing process".

packages/menu/src/MenuItem.ts
Westbrook: The [return value](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSlotElement/assignedElements#return_value) for `assignedElements()` is always an array, even if it's empty, which means that accessing the `0` index will either get an `Element` or `undefined`, which are the two values that we expect here. Is there an alternative form of protection that you'd like to see a null pointer check guard us from?

hunterloftis (approved):

PR #4229, opened 2024-04-01
fix(base,reactive-controllers): add delegates API and downstate delegate

TarunAdobe:
## Description

This PR adds the delegate strategy in SWC that allows us to add `downstate` functionality over our existing components. This is needed so that we can support features like `downstate` in Spectrum 2 while keeping everything as it is in Spectrum 1

Downstate Activation is added to this [PR](https://github.com/adobe/spectrum-web-components/pull/4368)

## Motivation and context

The downstate feature in spectrum-two enhances user interaction by providing visual feedback when components are clicked or activated. It improves accessibility and user experience by clearly indicating the active state of elements. Developers can enable downstate for components, and it seamlessly integrates with existing functionality in the framework. This addition enriches the user interface, making interactions more intuitive and engaging.

[Technical Document](https://paper.dropbox.com/doc/Tech-Adding-downstate-in-SWC-with-delegates--CM7vMhN7q6jYexR156mMlhl2Ag-V2v21xDURLM1rFpoacWTI)

## Types of changes

## Checklist

## Best practices


packages/button/src/ButtonBase.ts
```diff
@@ -32,6 +32,10 @@ import buttonStyles from './button-base.css.js';
 export class ButtonBase extends ObserveSlotText(LikeAnchor(Focusable), '', [
     'sp-overlay,sp-tooltip',
 ]) {
+    spectrumConfig = {
```
Rajdeepc, line 35: We need to add this config to our documentation too. Would you be interested to layout an idea on how we can better communicate this additional config for the users.

tools/base/src/Base.ts
```diff
@@ -185,8 +186,14 @@ export function SpectrumMixin<T extends Constructor<ReactiveElement>>(
     return SpectrumMixinElement;
 }
 
+export type SpectrumConfig = {
+    downstate?: boolean;
```
Rajdeepc, line 190: This is no longer a `boolean`. Rather an `array`

tools/reactive-controllers/src/SpectrumDelegates.ts
```diff
@@ -0,0 +1,42 @@
...
+import { SpectrumElement } from '@spectrum-web-components/base';
+import { SpectrumTwoDownStateController } from './delegate-controllers/SpectrumTwoDownStateController.js';
+
+export class SpectrumDelegates implements ReactiveController {
+    private host: SpectrumElement;
+
+    private spectrumTwoDownStateController: SpectrumTwoDownStateController | null =
+        null;
+
+    protected theme: string | null = null;
```
Rajdeepc, line 23: is `theme` not throwing a compile time error like `theme has already been initialised`?

tools/reactive-controllers/src/SpectrumDelegates.ts
```diff
@@ -0,0 +1,42 @@
...
+            this.spectrumTwoDownStateController?.unmanage();
+        }
+    }
+
+    constructor(host: SpectrumElement) {
+        this.host = host;
+        this.host.addController(this);
+    }
+
+    public hostConnected(): void {}
```
Rajdeepc, line 39: if `hostConnected & hostDisconnected` are not required we can remove this.

tools/reactive-controllers/src/delegate-controllers/SpectrumTwoDownStateController.ts
```diff
@@ -0,0 +1,72 @@
...
+    };
+
+    private updateDownStateStyles(isPressed: boolean): void {
+        if (isPressed) {
+            const height = this.host.offsetHeight;
+            const width = this.host.offsetWidth;
+            const transform = `perspective(max(${height}px, ${
+                width / 3
+            }px)) translate3d(0, 0, -2px)`;
+            this.host.style.setProperty('transform', transform);
```
Rajdeepc, line 67: Please check how the css team is doing this
https://github.com/adobe/spectrum-css/blob/a4490cb197a1c29a8638e04fb5b100d8d156f95f/.storybook/decorators/withDownStateDimensionCapture.js#L5

Rajdeepc (changes requested):

packages/button/src/ButtonBase.ts
Westbrook: Also this is not currently static like it in the `Base.ts`, should it be?

tools/reactive-controllers/src/delegate-controllers/SpectrumTwoDownStateController.ts
```diff
@@ -0,0 +1,72 @@
+/*
```
Westbrook, line 1: Please don't call this `SpectrumTwo`. It only happens to be a `DownStateController` that applies to "Spectrum Two".

packages/button/src/ButtonBase.ts
Rajdeepc: Yes this should be a static property also
```js
static override spectrumConfig = {
        downstate: ['spectrum-two'],
    };
```

tools/base/src/Base.ts
Rajdeepc: Changed to
```js
export type SpectrumConfig = {
    downstate?: string[];
};
```

tools/reactive-controllers/src/SpectrumDelegates.ts
Rajdeepc: Removed as `theme` variable not required here

tools/reactive-controllers/src/delegate-controllers/SpectrumTwoDownStateController.ts
Rajdeepc: Renamed to `DownStateController`

tools/reactive-controllers/src/delegate-controllers/SpectrumTwoDownStateController.ts
```diff
@@ -0,0 +1,72 @@
...
+export class SpectrumTwoDownStateController implements ReactiveController {
+    private host: SpectrumElement;
+
+    constructor(host: SpectrumElement) {
+        this.host = host;
+        this.host.addController(this);
+    }
+
+    public hostConnected(): void {
+        if (this.host.spectrumConfig?.downstate?.includes('spectrum-two')) {
```
Rajdeepc, line 25: you will not be able to access spectrumConfig from `this.host` since you can creating an instance of SpectrumElement and static property will not be accessible.
I have updated `spectrumConfig` to be public in `SpectrumElement`
```js
 public spectrumConfig: SpectrumConfig = {};
 ```

**Comment by Westbrook** [comment]
Build is unhappy here. Sadly, typescript requires that local dependencies also be listed as references `"references": [{ "path": "../base" }]` in `tsconfig.json` in order to support the build order of the project.

**Comment by Rajdeepc** [comment]
> Build is unhappy here. Sadly, typescript requires that local dependencies also be listed as references `"references": [{ "path": "../base" }]` in `tsconfig.json` in order to support the build order of the project.

Don't we also need to add `@spectrum-webcomponents/base` should also be added in the `dependencies` inside `reactive-controllers?

File: tools/reactive-controllers/src/SpectrumDelegates.ts [hunk 8]
```diff
@@ -0,0 +1,38 @@
...
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import type { ReactiveController } from 'lit';
+import { SpectrumElement } from '@spectrum-web-components/base';
```
Westbrook, line 14: Please import as
```suggestion
import type { SpectrumElement } from '@spectrum-web-components/base';
```

tools/base/src/Base.ts
```diff
@@ -12,6 +12,7 @@ governing permissions and limitations under the License.
 
 import { LitElement, ReactiveElement } from 'lit';
 import { version } from '@spectrum-web-components/base/src/version.js';
+import { SpectrumDelegates } from '@spectrum-web-components/reactive-controllers/src/SpectrumDelegates.js';
```
Westbrook, line 15: This causes a cyclical dependency. You need to make some decisions about where the types should live that allows a unidirectional dependency chain in order to allow this to build.

tools/reactive-controllers/src/SpectrumDelegates.ts
```diff
@@ -0,0 +1,38 @@
...
+    private host: SpectrumElement;
+
+    private downStateController: DownStateController | null = null;
+
+    set theme(value: string | null) {
+        if (value === 'spectrum-two') {
+            this.downStateController = new DownStateController(this.host);
+        } else {
+            this.downStateController?.unmanage();
+        }
```
Westbrook, line 27: This isn't the logic by which we chose to add this, right? It should be something like:
```
const { spectrumConfig } = this.host.constructor;
Object.entries(spectrumConfig).forEach(([controller, themes]) => {
    if (themes.includes(value)) {
        this[controller] = new Delegates[Controller](this.host);
    }
});
```
This way both multiple controllers can be added and individual controllers can be added on multiple themes?

Separately, should `theme` now be `system`?

tools/reactive-controllers/src/delegate-controllers/DownStateController.ts
```diff
@@ -0,0 +1,72 @@
+/*
```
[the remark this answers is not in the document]
Rajdeepc: `SpectrumDelegates` has been renamed to `Delegates` and `DownStateController.ts` has been renamed to `DownState.ts`

Westbrook:
You're missing test coverage here.

tools/reactive-controllers/src/downstate.ts
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    addEventListeners(): void {
+        this.host?.addEventListener('pointerdown', this.handlePointerDown);
+        this.host?.addEventListener('pointerup', this.handlePointerUp);
+    }
+
+    removeEventListeners(): void {
+        this.host?.removeEventListener('pointerdown', this.handlePointerDown);
+        this.host?.removeEventListener('pointerup', this.handlePointerUp);
```
Westbrook, line 46: Please use abort signals to remove event listeners.

Rajdeepc:
```js
addEventListeners(): void {
        this.host?.addEventListener('pointerdown', this.handlePointerDown, { signal: this.controller.signal });
        this.host?.addEventListener('pointerup', this.handlePointerUp, { signal: this.controller.signal });
    }
```

tools/reactive-controllers/src/downstate.ts
```diff
@@ -0,0 +1,69 @@
...
+            const height = this.host.offsetHeight;
+            const width = this.host.offsetWidth;
+            const transform = `perspective(max(${height}px, ${
+                width / 3
+            }px)) translate3d(0, 0, -2px)`;
+            this.host.style.setProperty('transform', transform);
+        } else {
+            this.host.style.removeProperty('transform');
+        }
+    }
```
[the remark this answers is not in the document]
Rajdeepc: This has been changed to capture the below properties on `style` with the new css attributes to handle `down state` in S2
```js
 if (isPressed) {
      const height = this.host.offsetHeight;
      const width = this.host.offsetWidth;
      this.host.style.setProperty('--spectrum-downstate-width', `${width}px`);
      this.host.style.setProperty('--spectrum-downstate-height', `${height}px`);
      
  } else {
      this.host.style.removeProperty('--spectrum-downstate-width');
      this.host.style.removeProperty('--spectrum-downstate-height');
  }
```

Rajdeepc:
@adobe/spectrum-css `spectrum-button.css` is overridden with logic to hide icon in `pending` state of `button`. An issue is raised on the S2 button implementation to fix this.

TarunAdobe:
@Rajdeepc Since we're not adopting any S2 css in this PR we're probably good to go here. However on the #4241 where we are adopting those styles, we need to workout the custom css to support all the broken S1 VRTs after migration to S2 styles including but not limited to pending button. Thanks for pointing it out!

Rajdeepc:
> @Rajdeepc Since we're not adopting any S2 css in this PR we're probably good to go here. However on the #4241 where we are adopting those styles, we need to workout the custom css to support all the broken S1 VRTs after migration to S2 styles including but not limited to pending button. Thanks for pointing it out!

Yes that's the plan! Thanks

tools/theme/src/Theme.ts
```diff
@@ -533,6 +534,10 @@ export class Theme extends HTMLElement implements ThemeKindProvider {
                 this.shadowRoot.appendChild(style);
             });
         }
+        this.trackedChildren.forEach((el) => {
```
Westbrook, line 537: Would `trackedChildren` practically be anything other than SpectrumElements here? Should we change the type of that? Or, should we add extra gating in the callback to make sure we're working with a SpectrumElement, if there are other elements that could be gathered here?

yarn.lock
```diff
@@ -5101,10 +5101,10 @@
   resolved "https://registry.yarnpkg.com/@spectrum-css/card/-/card-7.0.2.tgz#4db4772fa6e67ce46a823faca390d57f3afe4512"
   integrity sha512-ftp5BEts14oCFxy9iA0RmndZs0lbbqQCO8Yovd60E73gJmlzlk9SuSiRTGJCGzfPZjm7kZN5q2qfXG9Iwr5Qdw==
 
-"@spectrum-css/checkbox@^8.1.5":
-  version "8.1.5"
-  resolved "https://registry.yarnpkg.com/@spectrum-css/checkbox/-/checkbox-8.1.5.tgz#4dd8ff605f37e8d24b4991528939c2c84ecaba5a"
-  integrity sha512-91xOm+Hj4IrfbPlyNchLf5YW4ohBJlEQq1Urw+mqHSrzjqH9MG+IXLRT9OAY9zhuKTHfEOTqXatgYtfFqR6ILQ==
+"@spectrum-css/checkbox@^14.0.0-next.2":
```
Westbrook, line 5104: Here's where those new Checkbox styles came from. We likely don't want this in this PR?

packages/checkbox/package.json
```diff
@@ -71,7 +71,7 @@
         "@spectrum-web-components/shared": "^0.42.2"
     },
     "devDependencies": {
-        "@spectrum-css/checkbox": "^8.1.5"
+        "@spectrum-css/checkbox": "^14.0.0-next.2"
```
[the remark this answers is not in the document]
Rajdeepc: No not yet! but I will get in touch with Patrick for a timeline

packages/checkbox/src/spectrum-checkbox.css
```diff
@@ -34,7 +34,7 @@ governing permissions and limitations under the License.
     --spectrum-checkbox-control-color-disabled: var(
         --spectrum-disabled-content-color
     );
-    --spectrum-checkbox-checkmark-color: var(--spectrum-gray-75);
+    --spectrum-checkbox-checkmark-color: var(--spectrum-gray-50);
```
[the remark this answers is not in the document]
Rajdeepc: This new checkbox css version is imported to test downstate for checkbox

packages/checkbox/src/spectrum-config.js
```diff
@@ -50,6 +53,20 @@ const config = {
...
+                },
+                {
+                    find: {
+                        type: 'pseudo-class',
+                        kind: 'not',
+                        selectors: [[builder.class('is-readOnly')]],
+                    },
+                    replace: {
+                        type: 'pseudo-class',
+                        kind: 'not',
```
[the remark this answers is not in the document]
Rajdeepc: Yes this change is deliberate to add the newer css properties for checkbox to add a transform property to achieve the downstate as here
https://github.com/adobe/spectrum-css/pull/2520/files#diff-e8f4c2b6fa7f08b9bbd6104c59c7985ce8937381b5f384dd6b71a67aae6c8061R153

projects/story-decorator/src/StoryDecorator.ts
```diff
@@ -91,6 +91,7 @@ const reduceMotionProperties = css`
     --spectrum-animation-duration-2000: 0ms;
     --spectrum-animation-duration-4000: 0ms;
     --spectrum-animation-duration-6000: 0ms;
+    --spectrum-button-animation-duration: 0ms;
```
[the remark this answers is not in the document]
Rajdeepc: not required anymore as I was testing the downstate with button previously!

projects/story-decorator/src/StoryDecorator.ts
```diff
@@ -277,6 +278,12 @@ export class StoryDecorator extends SpectrumElement {
             ] as SpectrumElement[];
             descendents.push(...gathered);
         });
+        // run over the slotted descendants and add theme property
+        descendents.forEach((element) => {
+            if (element instanceof SpectrumElement) {
+                element.spectrumDelegates.theme = this.theme;
+            }
+        });
```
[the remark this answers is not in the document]
Rajdeepc: in Theme.ts I don't feel this is needed as this is very local to storybook.

tools/reactive-controllers/downstate.md
```diff
@@ -0,0 +1,45 @@
+## Description
+
+The `downstate` controller is designed to manage the down state styles of a custom element in a web application. This controller is particularly useful for creating interactive elements that visually respond to user input, such as buttons or interactive components.
```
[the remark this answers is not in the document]
Rajdeepc: I like the updated one! Thanks

tools/reactive-controllers/src/delegates.ts
```diff
@@ -0,0 +1,55 @@
...
+    downstate?: string[];
+};
+
+export class Delegates implements ReactiveController {
+    private host: LitElement;
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    private controllers: Record<string, any> = {};
+
+    set theme(value: string) {
```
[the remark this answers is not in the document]
Rajdeepc: Yes!

tools/reactive-controllers/src/downstate.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    addEventListeners(): void {
+        this.host?.addEventListener('pointerdown', this.handlePointerDown, {
+            signal: this.controller.signal,
+        });
+        this.host?.addEventListener('pointerup', this.handlePointerUp, {
+            signal: this.controller.signal,
+        });
+    }
+
```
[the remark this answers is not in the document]
Rajdeepc: I have made sure to call the `AbortController` during disconnet.
```js
public hostDisconnected(): void {
        this.removeEventListeners();
        this.controller.abort(); // Abort the controller when disconnecting
    }
```

yarn.lock
Rajdeepc: This is needed to create the test case and to surface up the VRTs of downstate.

packages/checkbox/src/spectrum-checkbox.css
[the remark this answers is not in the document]
Westbrook: The concept behind "Spectrum 2 Foundations" implies to me that Downstate should work even if we don't update the CSS. Please confirm whether that's true of our code or not.

packages/checkbox/src/spectrum-config.js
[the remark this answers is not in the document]
Westbrook: Ah, I see the output now:
```
:host(:not([readonly]):is(:active, [active])) #input:not(:disabled) + #box {
    transform: perspective(
            var(--spectrum-component-size-minimum-perspective-down)
        )
        translateZ(var(--spectrum-component-size-difference-down));
}
```
Gotcha.

projects/story-decorator/src/StoryDecorator.ts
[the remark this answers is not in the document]
Westbrook: Doesn't Theme already do this? https://github.com/adobe/spectrum-web-components/pull/4229/files#diff-d7278a185ea2e082c98b7e0c1192965a21208c1e16133411f5860048b9d030f2R537

tools/reactive-controllers/src/delegates.ts
```diff
@@ -0,0 +1,55 @@
...
+                    this.controllers[controller] = new ControllerClass(
+                        this.host
+                    );
+                }
+            }
+        );
+    }
+
+    constructor(host: LitElement) {
+        this.host = host;
```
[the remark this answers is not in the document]
Rajdeepc: Too add the element in the Delegates controller we need to make sure that `this` is a part of the ReactiveController itself. So if we remove implements `ReactiveController` we will be missing the type.

tools/theme/src/Theme.ts
Rajdeepc: Yes we should check for an instance of the SpectrumElement to add the system property
```js
this.trackedChildren.forEach((el) => {
    if (el instanceof SpectrumElement) {
        el.spectrumDelegates.system = this.system;
    }
});
```

tools/reactive-controllers/src/delegates.ts
```diff
@@ -0,0 +1,55 @@
...
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const DelegatesList: Record<string, any> = {
+    downstate: DownState,
+};
+
+export type SpectrumConfig = {
+    downstate?: string[];
+};
+
+export class Delegates implements ReactiveController {
```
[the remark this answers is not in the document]
Rajdeepc: Yes! This should not reflect on the ReactiveController since we don't need its methods here.

tools/reactive-controllers/src/downstate.ts
```diff
@@ -0,0 +1,84 @@
...
+    constructor(host: LitElement) {
+        this.host = host;
+        this.controller = new AbortController();
+        this.host.addController(this);
+    }
+
+    public getElement(): LitElement {
+        return this.host;
+    }
```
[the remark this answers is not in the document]
Rajdeepc: No we don't. The test has been updated

tools/styles/tokens/express/global-vars.css
```diff
@@ -1008,6 +1008,8 @@
     --system-spectrum-checkbox-control-color-focus: var(--spectrum-gray-900);
 }
 
+/*# sourceMappingURL=components/checkbox/themes/express.css.map */
```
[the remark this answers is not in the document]
Rajdeepc: Removed let's see from where this came from

projects/story-decorator/src/StoryDecorator.ts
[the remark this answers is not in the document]
Rajdeepc: Yes a little update here will add the logic.
```js
this.trackedChildren.forEach((el) => { if (el instanceof SpectrumElement) { el.spectrumDelegates.system = this.system; } });
```

tools/base/src/Base.ts
```diff
@@ -185,8 +187,14 @@ export function SpectrumMixin<T extends Constructor<ReactiveElement>>(
...
 }
 
+export type SpectrumConfig = {
+    downstate?: string[];
+};
+
 export class SpectrumElement extends SpectrumMixin(LitElement) {
     static VERSION = version;
+    public spectrumConfig: SpectrumConfig = {};
+    public spectrumDelegates = new Delegates(this);
```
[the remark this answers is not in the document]
Rajdeepc: Curious to know how useful this will be if we keep this in the `SpectrumMixin`?

packages/checkbox/src/spectrum-checkbox.css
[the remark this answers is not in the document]
Rajdeepc: As per Josh, the delivery of features should fall in like this:
1.0 - foundations only
1.1 - (maybe) downstate
1.2 - something else
1.3 - new font?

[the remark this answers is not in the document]
jnjosh: That future roadmap/feature list is highly made up too… :)  If we feel like the foundation part of Downstate should be part of foundations as Westbrook said, that can go in. We just might not apply it to all components expecting that behavior until future iterations.

Rajdeepc:
This PR only holds the downstate logic and the downstate activation of checkbox has been abstracted to #4368

tools/reactive-controllers/test/downstate.test.ts
```diff
@@ -0,0 +1,45 @@
...
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+import { expect, fixture, nextFrame } from '@open-wc/testing';
+import { html, LitElement } from 'lit';
+import { DownState } from '@spectrum-web-components/reactive-controllers/src/downstate.js';
+
+class TestDownstateEl extends LitElement {}
```
Westbrook, line 16: Can we use `spectrumConfig` here to ensure both the Controller and the delegates approach?

tools/reactive-controllers/src/downstate.ts
```diff
@@ -0,0 +1,81 @@
...
+            signal: this.abortController.signal,
+        });
+        this.host?.addEventListener('pointerup', this.handlePointerUp, {
+            signal: this.abortController.signal,
+        });
+    }
+
+    removeEventListeners(): void {
+        this.host?.removeEventListener('pointerdown', this.handlePointerDown);
+        this.host?.removeEventListener('pointerup', this.handlePointerUp);
```
Westbrook, line 53: We should remove the listeners with the abort signal.

tools/reactive-controllers/src/downstate.ts
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    public handlePointerDown = (): void => {
+        this.updateDownStateStyles(true);
+    };
+
+    public handlePointerUp = (): void => {
+        this.updateDownStateStyles(false);
+    };
```
Westbrook, line 62: This applies these changes synchronously, which is probably correct. However, the majority of our render reactions are async, due to the reactivity of LitElement. Do you se any issue with this?

tools/reactive-controllers/test/downstate.test.ts
Westbrook: May also want to wrap it all in an `<sp-theme>` to get ALL of the contracts covered.

Westbrook (commented):
Not so sure why there's churn in the lock. I've seen it in other branches lately, so it might be related to renovate bot, or this might not be _fully_ up to date with `main`?

tools/reactive-controllers/src/downstate.ts
Rajdeepc: Already doing here
```js
 public hostDisconnected(): void {
        this.removeEventListeners();
        this.abortController.abort();
    }
```

tools/reactive-controllers/src/downstate.ts
Rajdeepc: I feel updating the down state styles is a UI-specific operation and not directly related to LitElement's reactivity, handling it synchronously shouldn't cause any problems, but would love to hear your thoughts why you foresee some issues here?

tools/reactive-controllers/src/downstate.ts
```diff
@@ -0,0 +1,81 @@
...
+            const height = this.host.offsetHeight;
+            const width = this.host.offsetWidth;
+            this.host.style.setProperty(
+                '--spectrum-downstate-width',
+                `${width}px`
+            );
+            this.host.style.setProperty(
+                '--spectrum-downstate-height',
+                `${height}px`
+            );
```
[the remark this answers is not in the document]
Rajdeepc: Yes you are right. This is much more cleaner. But without setProperty it will not attach to the DOM.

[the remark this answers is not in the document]
Westbrook: Ooh, custom properties don't work this way 😱 ... TIL!

What a shame. Thanks for the clarification.

tools/reactive-controllers/src/downstate.ts
Rajdeepc: This has been updated!

tools/reactive-controllers/src/downstate.ts
Rajdeepc:
```js
removeEventListeners(): void {
        this.abortController.abort();
        this.host?.removeEventListener('pointerdown', this.handlePointerDown);
        this.host?.removeEventListener('pointerup', this.handlePointerUp);
    }
```

tools/reactive-controllers/test/downstate.test.ts
Rajdeepc:
```js
class TestDownstateEl extends SpectrumElement {
    override spectrumConfig = {
        downstate: ['spectrum-two'],
    };
        return html`
            <sp-theme system="spectrum-two" color="light" scale="medium">
                <slot></slot>
            </sp-theme>
        `;
    }
}
```

tools/reactive-controllers/test/downstate.test.ts
```diff
@@ -0,0 +1,45 @@
...
+                <test-downstate-el></test-downstate-el>
+            `
+        );
+
+        const controller = new DownState(
+            el as LitElement & { shadowRoot: ShadowRoot }
+        );
+        await nextFrame();
+        await nextFrame();
+        controller.handlePointerDown();
```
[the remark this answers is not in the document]
Rajdeepc: Sure! the directly call is removed with a sendMouse trigger to mimic the call!

tools/reactive-controllers/src/downstate.ts
Westbrook: The use of the abort controller here should prevent the need for `removeEventListener` code.

tools/reactive-controllers/src/delegates.ts
```diff
@@ -0,0 +1,50 @@
...
+    downstate?: string[];
+};
+
+export class Delegates {
+    private host: LitElement;
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    private controllers: Record<string, any> = {};
+
+    set system(value: string) {
```
rubencarvalho, line 31: Could this not be more strictly typed, for example using `SystemVariant`?

tools/base/src/Base.ts
```diff
@@ -185,8 +187,14 @@ export function SpectrumMixin<T extends Constructor<ReactiveElement>>(
     return SpectrumMixinElement;
 }
 
+export type SpectrumConfig = {
+    downstate?: string[];
```
rubencarvalho, line 191: I am not sure I follow why this needs to be configurable. If all the downstate controller does is update the value of some CSS variables, why does it need to be system-specific?
I would expect these CSS variables to not produce effect on non-supported systems.

TarunAdobe: So this is supposed to be configurable because we wanted to cover up for any further delegates that we might have to add specific to the system... Downstate is just one example that we already know.

rubencarvalho: Right, but by doing this we are introducing a JS-based feature toggling mechanism that, in this particular instance, could be avoided. Given that the CSS variables should inherently not affect unsupported systems, it seems unnecessary to add this extra layer of configuration.

I'm still not 100% convinced that adding extra code now in anticipation of unknown future features is the best approach, especially because it moves us away from CSS being the primary source of truth, which represents a significant shift in our mental model.

caseyisonit:
@TarunAdobe do you have a status on this PR and is there a Jira ticket associated with it? ive added a stale label

TarunAdobe:
> @TarunAdobe do you have a status on this PR and is there a Jira ticket associated with it? ive added a stale label

This was also for demo purposes. Which it served well. We will revisit the approach when we'll work on the downsate in s2 (prolly in a loooong future). So closing this rn

PR #4231, opened 2024-04-01
fix(shared): ensure the "updateComplete" in Focusable is stable

Westbrook:
## Description
- Leveraging non-static promises can cause faulty timing in `updateComplete` callbacks.
- Only delaying `updateComplete` when `autofocus` is true accelerated the render of the majority of focusable elements
- add the `end` or `-e` flag to the benchmark suite that accepts `updateComplete` (the default) or `paint` which waits till the browser paints before ending the performance measurement

## Motivation and context
The `paint` timing of `focusable` element can be pushed back for the active `requestAnimationFrame`, we shouldn't charge every element that fee at render time.

## How has this been tested?
-   [ ] _Test case 1_
    1. Go [here](https://perf-refactor--spectrum-web-components.netlify.app/storybook/?path=/story/button-accent-fill--default)
    2. See that the element renders as expected
-   [ ] _Test case 2_
    1. Review the timing of the tests

## Types of changes

## Checklist

### DIFF
--- a/packages/button/test/button.test.ts
+++ b/packages/button/test/button.test.ts
@@ -97,6 +97,41 @@ describe('Button', () => {
         expect(el).to.not.be.undefined;
         await expect(el).to.be.accessible();
     });
+    it('has a stable/predictable `updateComplete`', async () => {
+        const test = await fixture<HTMLDivElement>(
+            html`
+                <div></div>
+            `
+        );
+
+        let keydownTime = -1;
+        let updateComplete1 = -1;
+        let updateComplete2 = -1;
+
+        const el = document.createElement('sp-button');
+        el.autofocus = true;
+        el.addEventListener('keydown', () => {
+            keydownTime = performance.now();
+        });
+        el.updateComplete.then(() => {
+            updateComplete1 = performance.now();
+        });
+        el.updateComplete.then(() => {
+            updateComplete2 = performance.now();
+        });
+        test.append(el);
+        // don't use elementUpdated(), as it is under test...
+        await nextFrame();
+        await nextFrame();
+        await nextFrame();
+        await nextFrame();
+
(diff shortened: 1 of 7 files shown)


packages/button/test/button.test.ts
```diff
@@ -97,6 +97,41 @@ describe('Button', () => {
...
+
+        const el = document.createElement('sp-button');
+        el.autofocus = true;
+        el.addEventListener('keydown', () => {
+            keydownTime = performance.now();
+        });
+        el.updateComplete.then(() => {
+            updateComplete1 = performance.now();
+        });
+        el.updateComplete.then(() => {
```
Rajdeepc, line 119: These two `updateComplete` promises can potentially be triggered at the same time

Rajdeepc (changes requested):
Can we also make sure we keep this PR clean from any css changes as below
Screenshot 2024-04-02 at 4 18 05 PM

packages/button/test/button.test.ts
Westbrook: That's actually the goal. The previous code positioned the first to happen _after_ the second 😞

test/benchmark/helpers.ts
```diff
@@ -20,9 +20,26 @@ declare global {
     interface Window {
         tachometerResult: undefined | number;
         tachometerStart: undefined | 'page' | 'element';
+        tachometerEnd: undefined | 'updateComplete' | 'paint';
     }
 }
 
+/**
+ * Runs `callback` shortly after the next browser Frame is produced.
+ */
```
hunterloftis, line 29: If there's a URL source for this wizardry it would be good to include so we can monitor it over time.

hunterloftis (approved):

PR #4250, opened 2024-04-08
fix: Mitigate Safari memory leak for input element

loredanaspataru:
## Description

This is an attempt to reduce the memory leak impact caused by the [Safari bug](https://bugs.webkit.org/show_bug.cgi?id=262219) with HTML input elements that aren't properly garbage collected.  

- input event listeners are added using `addEventListener` API instead of the lit declarative way
- event listeners and form element are removed on `disconnectedCallback`

The PR should be treated as POC to showcase the changes we need in SWC and to confirm the fix works as expected on our side.

## Related issue(s)

- [https://github.com/adobe/spectrum-web-components/discussions/4197](https://github.com/adobe/spectrum-web-components/discussions/4197)

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

Here's a screenshot that compares DOM trees that use `sp-search`; left one is current SWC behavior, right one is patched using the changes in this PR.  After detaching child nodes from their parent element, the ones in red are garbage collected, while the remaining ones are retained.

compare-patches

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/textfield/src/Textfield.ts
+++ b/packages/textfield/src/Textfield.ts
@@ -204,6 +204,17 @@ export class TextfieldBase extends ManageHelpText(
         return this.inputElement;
     }
 
+    private _eventHandlers = {
+        input: this.handleInput.bind(this),
+        change: this.handleChange.bind(this),
+        focus: this.onFocus.bind(this),
+        blur: this.onBlur.bind(this),
+    };
+
+    protected _firstUpdateAfterConnected = false;
+
+    protected abortController?: AbortController;
+
     /**
      * Sets the start and end positions of the current selection.
      *
@@ -316,10 +327,6 @@ export class TextfieldBase extends ManageHelpText(
                 pattern=${ifDefined(this.pattern)}
                 placeholder=${this.placeholder}
                 .value=${this.displayValue}
-                @change=${this.handleChange}
-                @input=${this.handleInput}
-                @focus=${this.onFocus}
-                @blur=${this.onBlur}
                 ?disabled=${this.disabled}
                 ?required=${this.required}
                 ?readonly=${this.readonly}
@@ -351,10 +358,6 @@ export class TextfieldBase extends ManageHelpText(
                 pattern=${ifDefined(this.pattern)}
                 placeholder=${this.placeholder}
                 .value=${live(this.displayValue)}
-                @change=${this.handleChange}
-                @input=${this.handleInput}
-                @focus=${this.onFocus}
-                @blur=${this.onBlur}
                 ?disabled=${this.disabled}
                 ?required=${this.required}
                 ?readonly=${this.readonly}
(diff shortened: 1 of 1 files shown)


packages/search/src/Search.ts
```diff
@@ -133,6 +177,58 @@ export class Search extends Textfield {
...
+
+        if (changedProperties.has('value')) {
+            this._manageClearButtonListeners();
+        }
+    }
+
+    protected override firstUpdateAfterConnected(): void {
+        super.firstUpdateAfterConnected();
+
+        this.formAbortController = new AbortController();
```
Rajdeepc, line 213: Instead of creating a new AbortController instance every time the method is called, you can initialize it once in the constructor or connectedCallback and reuse it

packages/search/src/Search.ts
```diff
@@ -98,15 +130,28 @@ export class Search extends Textfield {
         );
     }
 
+    private _manageClearButtonListeners(): void {
+        // add clear button listener when button is added to the DOM
+        if (this.clearButton && !this.clearButtonAbortController) {
+            this.clearButtonAbortController = new AbortController();
```
Rajdeepc, line 136: Same here you are recreating this controller everytime this method reinstantiates.

Rajdeepc (changes requested):
Good Start!! Also can you check if you can better abstract the AbortController to a new controller file so that the rest of the library can also leverage it?

packages/search/src/Search.ts
```diff
@@ -98,15 +130,28 @@ export class Search extends Textfield {
         );
     }
 
+    private _manageClearButtonListeners(): void {
```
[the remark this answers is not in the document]
loredanaspataru: Will update, thanks for the suggestion

packages/search/src/Search.ts
```diff
@@ -133,6 +177,58 @@ export class Search extends Textfield {
...
+        this._firstUpdateAfterConnected = true;
+        this.requestUpdate();
+    }
+
+    public override disconnectedCallback(): void {
+        // Cleanup all form & button event listeners and remove form element from DOM
+        this.formAbortController?.abort();
+        this.clearButtonAbortController?.abort();
+        this.clearButtonAbortController = undefined;
+        this.form.remove();
```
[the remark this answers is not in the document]
loredanaspataru: Calling `formAbortController.abort()` removes all the event listeners that were created using that controller's `signal`. You can read more about that here [https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#signal](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#signal)

packages/search/src/Search.ts
loredanaspataru: I can't do that because once the `controller.abort()` method has been called its `signal` can't be reused to add the event listeners again. So the current flow is as follows: 
- on first update after `connected` I create an instance of `AbortController`
- I use the controller's `signal` to add the form event listeners
- the form event listeners are removed on `disconnectedCallback` by calling `controller.abort()`

packages/search/src/Search.ts
loredanaspataru: As mentioned previously I can't reuse the same controller. If it seems like a significant issue I can remove the `AbortController` and instead use `add/removeEventListener` API along with a flag that checks the listeners have been added to achieve the same result

PR #4251, opened 2024-04-09
fix(link): added feature to stop click propagation for disabled link

Rajdeepc:
## Description
`disabled` attribute should not allow click propagation in `sp-link`
Added a logic to stopPropagation and prevent any clicks to bubble up from the element.

## Related issue(s)

- #4252 

## How has this been tested?

-   [x] _Test case 1_
    1. Go [here](https://link-disabled-click--spectrum-web-components.netlify.app/storybook/?path=/story/link--disabled)
    2. Try to TAB or click on the link.

## Types of changes

## Checklist

## Best practices


packages/link/src/Link.ts
```diff
@@ -47,6 +47,22 @@ export class Link extends LikeAnchor(Focusable) {
...
+        this.addEventListener('click', this.handleClickCapture, {
+            capture: true,
+        });
+    }
+
+    private handleClickCapture(event: Event): void | boolean {
+        if (this.disabled) {
+            event.preventDefault();
+            event.stopImmediatePropagation();
+            event.stopPropagation();
```
Westbrook, line 61: https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events#none

Rajdeepc: Yes this was one of the simplest ways. I feel there are many places in the codebase where we are adding the above code to handle event propagation.

Rajdeepc:
> I _think_ this "corrects" this functionally but I'm worried about it actually working while looking the same as a NOT disabled link. Can we get some design direction?

There is a comment on this by @KayiuCarlson on [this](https://adobedesign.slack.com/archives/CESK60MQD/p1713364564641149?thread_ts=1712614724.465019&cid=CESK60MQD) discussion thread as:

> Disabled state for [Link](https://spectrum.corp.adobe.com/page/link/) is not currently available and being considered in Spectrum. In order for Spectrum design to consider this request (and have it be properly captured to the design tokens), we'd need to understand the use case as well as making sure that there's accessibility consideration. Therefore, it's not just a simple color change. If you're working with a designer, please have them reach out to us at Spectrum design. If you're not working with a designer, we'd still want to see the design of which a disabled Link would like to be used and its context. Without that, this would be a custom change on the product side only.

Westbrook (approved):
Lets take this in. Rebasing to `main` and will merge on green.

Issue #4201, opened 2024-03-21
[Bug]: Editable Slider not rounding values 


Rajdeepc:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Slider

### Expected behavior

Editable Slider onChange `event.target.value` should print 0.56 after adding`0.55999999999` on the NumberField 

### Actual behavior

Slider is printing 0.55999999999 instead of 0.56

### Screenshots

<img width="1272" alt="Screenshot 2024-03-21 at 6 24 26 PM" src="https://github.com/adobe/spectrum-web-components/assets/8790510/e7e6bb0a-bcfb-496c-92f2-cf92f27d885a">

### What browsers are you seeing the problem in?

Firefox, Chrome, Safari, Microsoft Edge

### How can we reproduce this issue?

Following are steps :-

Change min="0.25" to min="0.10"
In textbox , type 0.55999999999
In change i can see 0.55999999999 is getting logged instead of 0.56
Note: this bug is reproducible for few number like 0.55999999999  and NOT for [0.5999999999](tel:05999999999). And it is reproducible only for min value less than 0.11. 

JIRA bug: https://jira.corp.adobe.com/browse/CCEX-102612

Westbrook:
Would this be considered a duplicate of #3817?

PR #4263, opened 2024-04-12
fix(number-field, slider): floating point roundoff precision bug

blunteshwar:
## Description

While calculating the `moduloStep` in `validateInput` we were not rounding the value due to which when the value of `moduloStep` was expected to be 0, it was actually 1.7e-17 and causing the error

## Related issue(s)

- #4201 

## Motivation and context

This change was required because due to wrong value calculation of `moduloStep` the value of number-field was also incorrect. Now that we are correctly calculating `moduloStep` the value of number-field is also correct.

## How has this been tested?

-   [x] _Test case 1_
    1. Go [here](https://blunteshwar-slider-not-rounding-values--spectrum-web-components.netlify.app/storybook/?path=/story/slider--editable)
    2. Add 0.56 in the input field and check the console.

Before Fix

https://github.com/adobe/spectrum-web-components/assets/32033946/d62f08b3-9249-472c-85d3-dd9d4c926026

After Fix

https://github.com/adobe/spectrum-web-components/assets/32033946/7c3cf632-2c5f-4590-a5a7-95c5901192ae

## Types of changes

## Checklist

## Best practices


packages/number-field/test/inputs.test.ts
```diff
@@ -314,6 +314,27 @@ describe('NumberField - inputs', () => {
             expect(el.value).to.equal(255);
         });
     });
+    describe('user suplied fractional numbers', () => {
```
Rajdeepc, line 317:
```suggestion
    describe('with floating point numbers', () => {
```

packages/number-field/test/inputs.test.ts
```diff
@@ -314,6 +314,27 @@ describe('NumberField - inputs', () => {
...
+        it('do not crash the Number Field', async () => {
+            const el = await getElFrom(minMax(minMax.args));
+            el.setAttribute('min', '0.1');
+            el.setAttribute('step', '0.01');
+            el.setAttribute('value', '0.5');
+            await elementUpdated(el);
+
+            el.focus();
+            await sendKeys({
+                type: '6',
```
Rajdeepc, line 327: why type: '6'?

packages/slider/stories/slider.stories.ts
```diff
@@ -450,6 +450,32 @@ editableWithDefaultValue.swc_vrt = {
...
+                ...=${spreadProps(args)}
+            >
+                Angle
+            </sp-slider>
+        </div>
+    `;
+};
+
+editableWithFractionValue.swc_vrt = {
+    skip: true,
```
Rajdeepc, line 476: why are you skipping vrt here?

packages/number-field/test/inputs.test.ts
blunteshwar: This is because the initial value in number-field is '0.5' and appending it with'6'

packages/slider/stories/slider.stories.ts
blunteshwar: Because as such  there are no visual changes in this test

Rajdeepc (dismissed):

Westbrook:
image

I still see the bug outlined in the issue.

We likely need to leverage `formatOptions` and possibly https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#maximumfractiondigits in some way to step around the expected floating point number issues of working JS.

packages/number-field/src/NumberField.ts
```diff
@@ -495,7 +495,17 @@ export class NumberField extends TextfieldBase {
...
-            const moduloStep = (value - min) % this.step;
+            const digitsAfterDecimals =
+                this.step != Math.floor(this.step)
+                    ? this.step.toString().split('.')[1].length
+                    : 0;
+            const formatter = new Intl.NumberFormat('en-US', {
+                maximumFractionDigits: digitsAfterDecimals,
+            });
+            const moduloStep = parseFloat(
+                formatter.format((value - min) % this.step)
```
Rajdeepc, line 506: How will you catch if the value is less than min?

packages/number-field/src/NumberField.ts
```diff
@@ -495,7 +495,17 @@ export class NumberField extends TextfieldBase {
         // Step shouldn't validate when 0...
         if (this.step) {
             const min = typeof this.min !== 'undefined' ? this.min : 0;
-            const moduloStep = (value - min) % this.step;
+            const digitsAfterDecimals =
+                this.step != Math.floor(this.step)
+                    ? this.step.toString().split('.')[1].length
```
Rajdeepc, line 500: If `this.step` is not a floating-point number, calling split('.')[1] on it will result in an error. Check first if `this.step` is a floating point number or not.

packages/number-field/src/NumberField.ts
blunteshwar: Cecking if value is less than `min` or more than `max` is handled in `valueWithLimits` function, which is being called on line 491 at the beginning of `validateInput`

packages/number-field/src/NumberField.ts
blunteshwar: I am using a ternary operator where if `this.step` is not a floating-point number, then ` this.step != Math.floor(this.step)` will evaluate to `false` and hence `digitsAfterDecimals` will be assigned `0`. (Line 501)

Westbrook (commented):
This is looking much better! One more question: now that we're using Number Format, how would we be served by attempting to normalize this workflow in `protected get numberFormatter(): NumberFormatter {` rather than constructing a new formatter on every validation?

blunteshwar:
> new formatter on every validation?

I believe we have to create a new formatter object on every validation beacause we need the number of decimal digits after the `step` to pass into the formatter's properties like this `const formatter = new Intl.NumberFormat('en-US', {
                maximumFractionDigits: digitsAfterDecimals,
            });`
  Here `digitsAfterDecimals` is the no. of digits after decimal in `step`. Hence if a user changes the value of `step` for a number-field we need a new formatter right away!

packages/number-field/src/NumberField.ts
```diff
@@ -495,7 +495,17 @@ export class NumberField extends TextfieldBase {
...
+                    : 0;
+            const formatter = new Intl.NumberFormat('en-US', {
+                maximumFractionDigits: digitsAfterDecimals,
+            });
```
Westbrook, line 504: This news a NumberFormat for _every_ validation. It really only needs to do that for each time the value of `step` changes. In the case that we keep a _second_ formatter like this, we likely want to cache it's value in some way. Similarly, we could add this logic to the larger construction in `protected get numberFormatter()` and only remove it's cache when updating `step` which would allow us to manage a single formatter.

blunteshwar: Another approach that comes to my mind is to cache the formatter and only create a new one when `digitsAfterDecimals` changes, I can leverage a private field to store the formatter and another one to store the last known value of `digitsAfterDecimals`. Then, in the getter for the formatter, I can check if `digitsAfterDecimals` has changed and create a new formatter if it has changed.

`private _formatter: Intl.NumberFormat | null = null;
private _lastDigitsAfterDecimals: number | null = null;

protected get formatter(): Intl.NumberFormat {
    if (this._formatter === null || this._lastDigitsAfterDecimals !== this.digitsAfterDecimals) {
        this._formatter = new Intl.NumberFormat('en-US', {
            maximumFractionDigits: this.digitsAfterDecimals,
        });
        this._lastDigitsAfterDecimals = this.digitsAfterDecimals;
    }
    return this._formatter;
}`
THOUGHTS?

Westbrook: I'm fine with having two formatters if you see that as the right approach here, as long as both are cached appropriately. We should ensure that they both use the `NumberFormatter` helper rather than `Intl.NumberFormat` directly so that there isn't confusion down the line.

packages/number-field/src/NumberField.ts
```diff
@@ -495,7 +495,14 @@ export class NumberField extends TextfieldBase {
         // Step shouldn't validate when 0...
         if (this.step) {
             const min = typeof this.min !== 'undefined' ? this.min : 0;
-            const moduloStep = (value - min) % this.step;
+            this.digitsAfterDecimal =
```
Westbrook, line 498: Can we cauterize this work to `change.get('step')` in one of the lifecycle callbacks? I'm not super excited about the `this._lastDigitsAfterDecimal !== this.digitsAfterDecimal` check, can we leverage the existing cache management mechanics?

blunteshwar: Sure!

packages/number-field/src/NumberField.ts
blunteshwar: I had to use 2 formatters because when I was combining the logic in a single formatter, there were a lot of vrt's failing when the number field had units like percentage or dollar etc, because that formatter was called twice once inside `validateInput` and once to maintain the units.

packages/number-field/src/NumberField.ts
```diff
@@ -510,6 +512,7 @@ export class NumberField extends TextfieldBase {
                     value -= this.step;
                 }
             }
+            value = parseFloat(this.valueFormatter.format(value));
```
Westbrook, line 515: Is it possible to only format at the end to reduce the chances of fidelity loss?

packages/number-field/src/NumberField.ts
```diff
@@ -560,8 +563,24 @@ export class NumberField extends TextfieldBase {
...
+                { maximumFractionDigits: this.digitsAfterDecimal }
+            );
+        }
+
+        return this._valueFormatter;
+    }
     private _numberFormatter?: NumberFormatter;
     private _numberFormatterFocused?: NumberFormatter;
+    private _valueFormatter!: NumberFormatter;
+    private digitsAfterDecimal: number = 0;
```
Westbrook, line 583: Do we need to keep this on the class, or can we use cache clearing to trigger the use of `valueFormatter` as a getter and contain this value therein?

blunteshwar: Yeah, we can totally do that

packages/number-field/src/NumberField.ts
```diff
@@ -685,6 +703,18 @@ export class NumberField extends TextfieldBase {
...
+        if (changes.has('step')) {
+            if (this.step) {
+                const digitsAfterDecimal =
+                    this.step != Math.floor(this.step)
+                        ? this.step.toString().split('.')[1].length
+                        : 0;
+                this._valueFormatter = new NumberFormatter(
+                    this.languageResolver.language,
+                    { maximumFractionDigits: digitsAfterDecimal }
+                );
```
Westbrook, line 715: We shouldn't need to populate the cache here, if we evict it, the use of `this.valueFormatter` later will automatically populate it on demand.

blunteshwar: we are populating cache here because when ever the value of `step` changes we have to create a new instance of NumberFormatter. 
If instead of line 708-715 i just write `this.valueFormatter` this will not update the value of `decimalAfterDigits`.
For instance lets assume initially `this.step` was 0.1 and thus `decimalAfterDigits` was 1 and hence `this._valueFormatter` was initialised with `maximumFractionDigits: 1`. Now user updates `this.step` to 0.001, so `decimalAfterDigits` will become 3, now we have to again initialise `this._valueFormatter` with updated value of `decimalAfterDigits`. But if don't do so and just write `this.valueFormatter` instead of line 708-715, this will return `this._valueFormatter` with old value of `maximumFractionDigits`

blunteshwar: Done!

Westbrook (approved):
Thanks for working through this. LGTM!

PR #4228, opened 2024-03-28
performance: added memory tests in packages

Rajdeepc:
## Objective

The objective of this RFC is to propose a standardised approach for adding memory tests to each package in the component library using [Chrome's Performance API](https://developer.mozilla.org/en-US/docs/Web/API/Performance/measureUserAgentSpecificMemory), specifically the `measureUserAgentSpecificMemory()` method.

refs #3722

## Background

Memory leaks in web applications can lead to performance issues and degraded user experience. By adding memory tests to each package in the component library, we can proactively identify and fix potential memory leaks, ensuring the overall stability and performance of the library.

### Implementation of Memory Tests

1. The memory tests will be implemented using the `measureUserAgentSpecificMemory()` method from the Chrome Performance API.
2. The memory test for each package will consist of the following steps:
        - Create a test scenario that triggers the component's functionality.
        - Measure the memory usage before and after the test scenario.
        - Calculate the memory usage difference (DOM and JavaScript) to identify any memory leaks.

### Test Structure

1. Each memory test will be structured as a separate test case, following a standardised format.
2. The test case will include the following components:
        - Setup: Prepare the environment for the test.
        - Execution: Trigger the component's functionality.
        - Measurement: Measure the memory usage before and after the test.
        - Assertion: Verify that the memory usage difference is within an acceptable range.

### Test Helper Function

1. A test helper function named `usedHeapMB()` will be used to measure the memory usage.
2. The `usedHeapMB()` function will:
        - Call `measureUserAgentSpecificMemory()` to get the memory sample.
        - Parse the memory sample to extract the memory usage breakdown (DOM, JavaScript).
        - Return the memory usage breakdown in megabytes.

### Integration with Testing Framework

1. The memory tests will be integrated into the existing testing framework `@open-wc/testing` used for testing the SWC component library.
2. The memory tests will be executed as part of the automated test suite to ensure they are run regularly and consistently.

For more details on this experiment please go through this [RFC](https://paper.dropbox.com/doc/RFC-Adding-Memory-Tests-for-Component-Library-Packages--CMJMs9I58fIJUkqq87IGZ8KGAg-Q6ANtQLuEi5a8EkdGns1U)

## Related Issue:
Building on top of https://github.com/adobe/spectrum-web-components/pull/3721

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

[...]

PR #4264, opened 2024-04-12
chore: added memory tests for missing packages

### LINKED ISSUE #3722: [testing] Abstract and apply memory usage testing across the repo
### Code of conduct

### Description of issue

Building off of #3721, abstract the usage of `measureUserAgentSpecificMemory()` to a test helper, and apply that helper across all of the packages in the project.

Address leaks, etc., as they are uncovered.

TarunAdobe:
Added memory tests for remaining packages.

## Objective

fixes #3722

## Background

Memory leaks in web applications can lead to performance issues and degraded user experience. By adding memory tests to each package in the component library, we can proactively identify and fix potential memory leaks, ensuring the overall stability and performance of the library.

## Related Issue:
Building on top of https://github.com/adobe/spectrum-web-components/pull/4228

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/action-button/test/action-button-memory.test.ts
+++ b/packages/action-button/test/action-button-memory.test.ts
@@ -9,12 +9,7 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import { fixture } from '@open-wc/testing';
 import { m as BlackActionButton } from '../stories/action-button-black.stories.js';
-import { ActionButton } from '@spectrum-web-components/action-button';
 import { testForMemoryLeaks } from '../../../test/testing-helpers.js';
 
-testForMemoryLeaks(
-    async () =>
-        await fixture<ActionButton>(BlackActionButton(BlackActionButton.args))
-);
+testForMemoryLeaks(BlackActionButton(BlackActionButton.args));
--- a/packages/action-button/test/action-button-memory.test.ts
+++ b/packages/action-button/test/action-button-memory.test.ts
@@ -1,0 +1,20 @@
+/*
+Copyright 2023 Adobe. All rights reserved.
+This file is licensed to you under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License. You may obtain a copy
+of the License at http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+import { fixture } from '@open-wc/testing';
+import { m as BlackActionButton } from '../stories/action-button-black.stories.js';
+import { ActionButton } from '@spectrum-web-components/action-button';
+import { testForMemoryLeaks } from '../../../test/testing-helpers.js';
+
+testForMemoryLeaks(
+    async () =>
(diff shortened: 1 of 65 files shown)


packages/action-button/test/action-button-memory.test.ts
```diff
@@ -0,0 +1,20 @@
...
+
+import { fixture } from '@open-wc/testing';
+import { m as BlackActionButton } from '../stories/action-button-black.stories.js';
+import { ActionButton } from '@spectrum-web-components/action-button';
+import { testForMemoryLeaks } from '../../../test/testing-helpers.js';
+
+testForMemoryLeaks(
+    async () =>
+        await fixture<ActionButton>(BlackActionButton(BlackActionButton.args))
+);
```
Westbrook, line 20: Not sure if this is useful or not, but should we signature of this method so it works along the lines of:

```suggestion
testForMemoryLeaks<ActionButton>(BlackActionButton(BlackActionButton.args));
```

I'm not 100% sure the type is even technically required as we're not testing against its actual API here. 

Maybe I'm being overly precious and it's not worrying about, but I'm less excited by the ones blow out to the following by the linter:

```
testForMemoryLeaks(
    async () =>
        await fixture<Underlay>(
            html`
                <sp-underlay></sp-underlay>
            `
        )
);
```

With this change, it would technically be the below:

```
testForMemoryLeaks<Underlay>(html`<sp-underlay></sp-underlay>`);
```

Or even less is the type is unneeded.

🤔

Rajdeepc: I think to me it does improve the readability and conciseness of the test calls.

Rajdeepc: with the current implementation of testForMemoryLeaks adding a type will expect the function to return a value. How do we see if we just update the above to
```
testForMemoryLeaks(html`<sp-underlay></sp-underlay>`);
```

Westbrook (commented):
These changes _look_ good, but we need to make sure CI is passing before we can merge.

Separately, we'll need to file some follow up work to get this test into the component generator, but not reason to block progress on that. Can you also work with the three in progress new component PRs to ensure this test is included therein?

Rajdeepc:
> These changes _look_ good, but we need to make sure CI is passing before we can merge.
> 
> Separately, we'll need to file some follow up work to get this test into the component generator, but not reason to block progress on that. Can you also work with the three in progress new component PRs to ensure this test is included therein?

Yes sure! Also will update the express team to include these memory tests into their alert banner and breadcrumb PRs

Westbrook:
We _can_ merge this, because it is green. However, this causes the Chromium tests to take more the 4 minutes when they were previously taking less and a minute. Do we want to look at other options now or ship this and revisit timing/concurrency/segmentation of tests at a later date?

Rajdeepc:
> We _can_ merge this, because it is green. However, this causes the Chromium tests to take more the 4 minutes when they were previously taking less and a minute. Do we want to look at other options now or ship this and revisit timing/concurrency/segmentation of tests at a later date?

We can ship this. Yes there is a need to optimize the build time here! Its taking forever. Concurrency shards the build but delays the build time. Its better to segregate the memory test and check this in a separate PR, If you think this is fine let's merge this!

Rajdeepc:
@Westbrook Can we land this?

Westbrook (approved):
Let's hit up optimization in a follow up.

:shipit: 🙇🏼

Issue #4268, opened 2024-04-15
[Bug]: self-managed Tooltip creates infinite loop when no focusable trigger is found


Rocss:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-tooltip

### Expected behavior

Given the below code:
```
    <sp-icon-help>
        <sp-tooltip self-managed>Help is on the way</sp-tooltip>
    </sp-icon-help>
```

And given that no focusable element can be found in the parent tree of the `sp-tooltip`, I expect that the focusable element to end up being the document, and that a warning would be thrown when `window.__swc.DEBUG === true` so that I am instructed on how to provide an accessible, focusable trigger for my tooltip.

### Actual behavior

When `window.__swc.DEBUG === true`, the search loop is closed and the trigger is the `document`.
We get the expected warnings in the console.

However, when `window.__swc.DEBUG === false`, nothing is ever returned from the search loop, ending up in an infinite loop.

While this is a problem resulting from an incorrect usage of `self-managed`, developers may ignore the warnings in the console, as it does not create a loop in environments where DEBUG is `true` and end up with the infinite loop in production.

### How can we reproduce this issue?

1. In the code below, replace `window.__swc.DEBUG` with `false` (or directly in the browser console)
2. Open the Tooltip self-managed storybook
3. Remove the focusable element and replace it with a `div`
5. Check console
6. See error

### Sample code that illustrates the problem

```
while (!triggerElement?.matches?.(focusableSelector)) {
            start =
                triggerElement.assignedSlot || (triggerElement as HTMLElement);
            root = start.getRootNode();
            /* c8 ignore next 13 */
            if (window.__swc.DEBUG) {
                if (root === document) {
                    window.__swc.warn(
                        this,
                        `Self managed [${this.localName}](https://app.slack.com/client/E23RE8G4F/$%7Bthis.localName%7D) elements walk up the composed tree to acquire a trigger element. No trigger element was found before the document.`,
                        'https://opensource.adobe.com/spectrum-web-components/components/tooltip#self-managed-overlays',
                        {
                            level: 'high',
                        }
                    );
                    return root as HTMLElement;
                }
            }
            triggerElement = (start.parentElement ||
                (root as ShadowRoot).host ||
                /* c8 ignore next 1 */
                root) as HTMLElement;
        }
        return triggerElement;
```

PR #4269, opened 2024-04-15
fix(tooltip): fix infinite loop in self-managed tooltips

Rocss:
## Description

Reordering some `if` statements to make sure a trigger is always returned for self-managed tooltips.

## Related issue(s)

- https://github.com/adobe/spectrum-web-components/issues/4268

## Motivation and context

Fixes a bug which reproduces only when NOT in DEBUG mode. See details in the issue above.

Not sure how to add a unit test for this, since `triggerElement` is private and I can not check it from outside.

## How has this been tested?

-   [x] window.__swc.DEBUG === false
    1. Modified the `self-managed` story and replaced `sp-action-button` with a `div`
    2. No infinite loop - trigger ends up being the `document`
    3. No warning is thrown in the console
-   [x] window.__swc.DEBUG === true
    1. Modified the `self-managed` story and replaced `sp-action-button` with a `div`
    2. No infinite loop - trigger ends up being the `document`
    4. Warning is thrown in the console

## Types of changes

## Checklist

## Best practices


Westbrook (commented):
Do we have tests that prove that _if_ it finds document it _doesn't_ bind listeners? A Tooltip in this situation should not work in any way, it should only dispatch the Dev Mode warning.

Rocss:
@Westbrook As it is right now it _does_ bind event listeners on the document.
May I suggest that `get triggerElement()` would return `null` if no focusable element was found in the parent chain, instead of `document`?
Looking at the `Overlay` class, it seems like it is allowed to be `null`:
```
Overlay.triggerElement: HTMLElement | VirtualTrigger | null
```

Westbrook (commented):
Do we have tests that prove that _if_ it finds document it _doesn't_ bind listeners? A Tooltip in this situation should not work in any way, it should only dispatch the Dev Mode warning.

If not, we can land this on the presumption it works, as it seems to not be working for you otherwise.

packages/tooltip/test/tooltip.test.ts
```diff
@@ -214,6 +214,19 @@ describe('Tooltip', () => {
...
+    it('self managed does not attach event listeners if no trigger was found', async () => {
+        const documentEventsSpy = spy(document, 'addEventListener');
+        const el = await fixture<HTMLDivElement>(
+            html`
+                <sp-tooltip self-managed>Help text.</sp-tooltip>
+            `
+        );
+
+        await elementUpdated(el);
+        expect(documentEventsSpy.callCount).to.equal(0);
```
Westbrook, line 227: Can we confirm it doesn't attach to anything else, too?
```suggestion
        expect(documentEventsSpy.callCount).to.equal(0);
        expect(el.overlayElement?.triggerElement).to.be.null;
```

Westbrook (commented):
For Lighthouse/a preview URL, your branch name is too long. 🙈 We won't count it against you 😉 

This is a good test, with that one extra expectation and a rebase to `main` we should be good to go here. Thanks for finding tis issue!

Westbrook (approved):
:shipit: Thank you, thank you!

PR #4273, opened 2024-04-16
fix(action-bar): include focus-visible polyfilling

inasirhussain:
## Description
For older browser and UXP, we have to rely on focus-visible polyfill to match behaviour with the modern browser where the corresponding pseudo-selector is supported.

## Related issue(s)
fixes https://github.com/adobe/spectrum-web-components/issues/4279

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/action-bar/src/ActionBar.ts
+++ b/packages/action-bar/src/ActionBar.ts
@@ -14,7 +14,6 @@ import {
     CSSResultArray,
     html,
     SpectrumElement,
-    SizedMixin,
     TemplateResult,
 } from '@spectrum-web-components/base';
 import { property } from '@spectrum-web-components/base/src/decorators.js';
@@ -31,9 +30,7 @@ export const actionBarVariants = ['sticky', 'fixed'];
  * @element sp-action-bar
  * @slot - Content to display with the Action Bar
  */
-export class ActionBar extends SizedMixin(
-    FocusVisiblePolyfillMixin(SpectrumElement)
-) {
+export class ActionBar extends FocusVisiblePolyfillMixin(SpectrumElement) {
     public static override get styles(): CSSResultArray {
         return [actionBarStyles];
     }
--- a/packages/action-bar/src/ActionBar.ts
+++ b/packages/action-bar/src/ActionBar.ts
@@ -14,6 +14,7 @@ import {
     CSSResultArray,
     html,
     SpectrumElement,
+    SizedMixin,
     TemplateResult,
 } from '@spectrum-web-components/base';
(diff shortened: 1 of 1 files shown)


Westbrook:
As a member of Adobe, Inc. you should be able to self serve access to this repo and make this PR as a branch, rather than a fork, and enjoy our full CI coverage. Would you be able to set that up and reopen this PR?

packages/action-bar/src/ActionBar.ts
```diff
@@ -14,6 +14,7 @@ import {
     CSSResultArray,
     html,
     SpectrumElement,
+    SizedMixin,
```
Westbrook, line 17: There are no t-shirt values for Action Bar. Please remove.

packages/action-bar/src/ActionBar.ts
```diff
@@ -23,13 +24,16 @@ import '@spectrum-web-components/button/sp-close-button.js';
...
+import { FocusVisiblePolyfillMixin } from "@spectrum-web-components/shared/src/focus-visible.js";
 export const actionBarVariants = ['sticky', 'fixed'];
 
 /**
  * @element sp-action-bar
  * @slot - Content to display with the Action Bar
  */
-export class ActionBar extends SpectrumElement {
+export class ActionBar extends SizedMixin(
+    FocusVisiblePolyfillMixin(SpectrumElement)
```
Westbrook, line 35: Seems reasonable for the close button. Would be great if UXP prioritized adding this, as I'm sure there are lots of little places where UI like this falls through the cracks and keeping the polyfill code for something that has been fixed in browser for so long is questionable on our end.

Westbrook (commented):
Simplify to only the needed changes and we can adopt this.

In the future, if you were able to work from a branch rather than a fork you'll get more complete CI coverage of these sorts of changes.

packages/action-bar/src/ActionBar.ts
Westbrook:
```suggestion
```

packages/action-bar/src/ActionBar.ts
```diff
@@ -23,13 +24,16 @@ import '@spectrum-web-components/button/sp-close-button.js';
...
+) {
```
Westbrook, line 36:
```suggestion
export class ActionBar extends FocusVisiblePolyfillMixin(SpectrumElement) {
```

PR #4284, opened 2024-04-19
chore: added figma design and badges addons to storybook

Rajdeepc:
## Description

Badge for deprecated components. With this we can control tags like Migrated, Spectrum-two etc on component level

Screenshot 2024-04-19 at 5 46 27 PM

Design Tab
Screenshot 2024-04-19 at 5 19 26 PM

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices


Westbrook:
This seems like a cool addition, however, this is not my favorite: 
IMG_6689

Maybe we could work with the design team on an open source entry point to these design files?

Rajdeepc:
> This seems like a cool addition, however, this is not my favorite: IMG_6689
> 
> Maybe we could work with the design team on an open source entry point to these design files?

Yes you are right. I will talk to design and see if I find a public url without auth.

castastrophe:
In the short term, you can toggle add-ons based on environment variable so you could show the design add-on only for dev mode.

Rajdeepc:
> * Storybook

Yes for the time being I have added this addon for `dev mode` only but definitely we should add this as a public url in future.

Rajdeepc:
@Westbrook does this make sense to include this in the current state of main branch considering we will be deleting the deprecated items in SWC 1.0 release?

projects/css-custom-vars-viewer/package.json
```diff
@@ -48,9 +48,11 @@
     },
     "devDependencies": {
         "@custom-elements-manifest/analyzer": "^0.9.0",
+        "@geometricpanda/storybook-addon-badges": "^2.0.2",
         "@open-wc/eslint-config": "^12.0.0",
         "@open-wc/testing": "^3.2.0",
         "@storybook/addon-a11y": "^7.5.0",
+        "@storybook/addon-designs": "^8.0.0",
```
Westbrook, line 55: Why here?

yarn.lock
```diff
@@ -5386,9 +5411,83 @@
   resolved "https://registry.yarnpkg.com/@spectrum-css/vars/-/vars-9.0.8.tgz#6af3bcdace903b8461f5fcd4c9aa23e70128a456"
   integrity sha512-rGfd7jqXOdR69bEjrRP58ynuIeJU0czPfwQvzhtCzg7jKVukV+efNHqrs086sC6xutB3W4TF71K/dZMr3oyTyg==
 
+"@spectrum-web-components/base@^0.40.5":
```
Westbrook, line 5414: Why these changes?

Westbrook (commented):
Do we also want to do `iconset` here?

I think with this cauterized to dev time, we can move forward with this. Couple of notes internally and then we can land it.

yarn.lock
Rajdeepc: These should be transitive depedencies. If you run a yarn install on `main` branch you can see these changes

Westbrook: I don't think they belong. Do you have a `react` directory locally?

PR #4285, opened 2024-04-22
feat(contextual-help): add contextual help pattern

Rocss:
## Description

Adds `@spectrum-web-components/contextual-help`, as per https://spectrum.adobe.com/page/contextual-help/

## Related issue(s)

-  https://github.com/adobe/spectrum-web-components/discussions/4281

## Motivation and context

Contextual help is a pattern used widely across Adobe apps. Even if at its core it is a simple component, the lack of a reusable component around this pattern results in different teams implementing different versions of it.
Differences in implementation differ not only on the visual level, but also on the accessibility and mobile responsiveness level. Current implementations are some of them lacking keyboard navigation support, screen reader support, and not taking into consideration user experience on mobile devices.

## How has this been tested?

-   [ ] _Test case 1_
    1. Go [here](https://contextual-help--spectrum-web-components.netlify.app/storybook/?path=/story/contextual-help--default)
    2. Open/close the Contextual help component
    3. UI respects the spectrum-css styling
-   [ ] _Test case 2_
    1. Go [here](https://contextual-help--spectrum-web-components.netlify.app/storybook/?path=/story/contextual-help--default)
    2. Toggle between `info`/`help` variants
    3. Observe the icon is changing from informative to help
-   [ ] _Test case 3_
    1. Go [here](https://contextual-help--spectrum-web-components.netlify.app/storybook/?path=/story/contextual-help--default)
    2. Change the `placement` prop using the Storybook Controls Addon
    3. Observe the popover is positioned based on the `placement` prop
-   [ ] _Test case 4_
    1. Go [here](https://contextual-help--spectrum-web-components.netlify.app/storybook/?path=/story/contextual-help--default)
    2. Switch to responsive mode -> choose a mobile view
    3. Observe the Contextual help is opening a dialog on mobile, instead of a popover
-   [ ] _Test case 5_
    1. Go [here](https://contextual-help--spectrum-web-components.netlify.app/storybook/?path=/story/contextual-help--default)
    2. Check out controlled behaviour by toggling the `open` control

Screenshot 2024-04-22 at 11 28 57

## Types of changes

## Checklist

## Best practices


Rocss:
I'm marking it as ready for review even if unit tests are incomplete, because I am looking for a high level review of this current implementation, and any changes would make the tests obsolete.

I found this component to be a good use case for the `slottable-request` pattern. However, if this seems overkill given the small contents of the popover I can get rid of it.

I'm looking for comments on the `width` / `max-width` dilema: https://github.com/adobe/spectrum-web-components/discussions/4281#discussioncomment-9186262, as well as on the mobile behaviour.

packages/contextual-help/src/ContextualHelp.ts
```diff
@@ -0,0 +1,205 @@
...
+        }
+
+        import('@spectrum-web-components/popover/sp-popover.js');
+
+        const template = html`
+            <sp-popover class="popover">
+                <section>
+                    ${this.headline &&
+                    html`
+                        <h2 class="heading">${this.headline}</h2>
```
Westbrook, line 121: I don't see any design examples without a heading: https://spectrum.adobe.com/page/contextual-help/ Should it really be possible to ship this without one?

Westbrook, line 121: A user should have control over the heading level leveraged here. They should also have the ability to render HTML into this content.

Likely we want something like:
```
<div class="heading"> <!-- holds style delivery -->
  <slot name="heading"> <!-- normalize this name to that in other elements -->
    <h2>${this.heading}</h2> <!-- holds default headline level but reverts default heading styles -->
  </slot>
</div>
```
Or something.

packages/contextual-help/src/ContextualHelp.ts
```diff
@@ -0,0 +1,205 @@
...
+                    `}
+                    <slot></slot>
+                    <div class="link">
```
Westbrook, line 124: Styling a wrapper means that even when a consumer does not provide a link the padding/margin managing it will be present. Can we point rules for this onto the slotted element, which will only apply if one is present?

packages/contextual-help/src/ContextualHelp.ts
```diff
@@ -0,0 +1,205 @@
...
+                        <slot name="link"></slot>
+                    </div>
+                </section>
+            </sp-popover>
+        `;
+
+        render(template, event.target as HTMLElement);
```
Westbrook, line 131: Can we gate between mobile and not at `template` in this line, rather than in line 199?

Should end up with less code AND in that case someone goes back and forth between mobile and not the render won't need to add/remove the event listener for the method ID changing.

packages/contextual-help/src/ContextualHelp.ts
```diff
@@ -0,0 +1,205 @@
...
+     * The `offset` property accepts either a single number, to
+     * define the offset of the Popover along the main axis from
+     * the action button, or 2-tuple, to define the offset along the
+     * main axis and the cross axis.
+     */
+    @property({ type: Number })
+    public offset: number | [number, number] = 0;
+
+    @state()
+    overlayOpen = false;
```
Westbrook, line 93: You probably want to normalize on `open` and make it an `@property()` so that consuming apps can declare that the help is open when needed.

packages/contextual-help/src/ContextualHelp.ts
```diff
@@ -0,0 +1,205 @@
...
+                          ></sp-icon-help-outline>
+                      `
+                    : html`
+                          <sp-icon-info-outline
+                              slot="icon"
+                          ></sp-icon-info-outline>
+                      `}
+            </sp-action-button>
+            <sp-overlay
+                trigger="trigger@click"
```
Westbrook, line 188: Moving to `open` likely means moving to `.triggerElement=...` instead and `?open=${this.open}` which them means you might want to leverage `DependencyManagerController` as seen in `Picker`.

packages/contextual-help/src/ContextualHelp.ts
```diff
@@ -0,0 +1,205 @@
...
+                placement=${ifDefined(actualPlacement)}
+                type=${this.isMobile.matches ? 'modal' : 'auto'}
+                receives-focus="true"
+                .offset=${this.offset}
+                @sp-opened=${() => {
+                    this.overlayOpen = true;
+                }}
+                @sp-closed=${() => {
+                    this.overlayOpen = false;
+                }}
```
Westbrook, line 198: Please don't use anonymous methods in event listeners.

packages/contextual-help/src/ContextualHelp.ts
```diff
@@ -0,0 +1,205 @@
...
+                          ></sp-icon-info-outline>
+                      `}
+            </sp-action-button>
+            <sp-overlay
+                trigger="trigger@click"
+                placement=${ifDefined(actualPlacement)}
+                type=${this.isMobile.matches ? 'modal' : 'auto'}
+                receives-focus="true"
+                .offset=${this.offset}
+                @sp-opened=${() => {
```
Westbrook, line 193: `sp-opened` technically _late_ for this. The `slottable-request` timing is likely more appropriate for knowing when the Overlay is "open".

packages/contextual-help/src/contextual-help.css
```diff
@@ -0,0 +1,17 @@
...
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+
+@import url('./spectrum-contextual-help.css');
+
+#trigger[is-active] {
+    background-color: var(--spectrum-actionbutton-background-color-down);
+}
```
Westbrook, line 17: Is this the same as applying the `active` attribute and allowing the Action Button to do this on its own?

packages/contextual-help/src/spectrum-contextual-help.css
```diff
@@ -0,0 +1,92 @@
...
+    --spectrum-contextual-help-content-spacing: var(--spectrum-spacing-100);
+    --spectrum-contextual-help-link-spacing: var(--spectrum-spacing-300);
+    --spectrum-contextual-help-heading-size: var(
+        --spectrum-contextual-help-title-size
+    );
+    --spectrum-contextual-help-heading-color: var(--spectrum-heading-color);
+    --spectrum-contextual-help-body-color: var(--spectrum-body-color);
+}
+
+.spectrum--large .popover {
```
Westbrook, line 26: This seems like it doesn't belong... It _should_ be processed out on it's own, but it may exist outside of our current CSS processing. Can you confirm the selector in the origin CSS?

packages/contextual-help/stories/contextual-help.stories.ts
```diff
@@ -0,0 +1,171 @@
...
+export default {
+    title: 'Contextual Help',
+    component: 'sp-contextual-help',
+    args: {
+        label: '',
+        variant: 'info',
+        headline: undefined,
+        placement: undefined,
+    },
+    argTypes: {
```
Westbrook, line 36: We need to update our generated files, but can you move this to match https://github.com/adobe/spectrum-web-components/blob/main/packages/accordion/stories/args.ts

packages/contextual-help/stories/contextual-help.stories.ts
```diff
@@ -0,0 +1,171 @@
...
+                'left',
+                'left-start',
+                'left-end',
+            ],
+        },
+    },
+};
+
+export const Default = (args: StoryArgs): TemplateResult => {
+    return html`
```
Westbrook, line 112: We need to update our generated files, but can you move this to match https://github.com/adobe/spectrum-web-components/blob/main/packages/accordion/stories/template.ts

Then as much sharing of `Template` as you can between the stories would be great.

Westbrook (commented):
You can confirm the args and template sharing by seeing the demo published at the top of the docs site after the change.

packages/contextual-help/test/contextual-help.test.ts
```diff
@@ -0,0 +1,98 @@
...
+    testForLitDevWarnings(
+        async () =>
+            await fixture<ContextualHelp>(
+                html`
+                    <sp-contextual-help headline="Permission required">
+                        Your admin must grant you permission before you can
+                        create a segment.
+                    </sp-contextual-help>
+                `
+            )
```
Westbrook, line 35: Once you have a `template.ts` file, it would be great to generate the test context from there.

packages/contextual-help/src/ContextualHelp.ts
```diff
@@ -0,0 +1,205 @@
...
+                              slot="icon"
+                          ></sp-icon-help-outline>
+                      `
+                    : html`
+                          <sp-icon-info-outline
+                              slot="icon"
+                          ></sp-icon-info-outline>
+                      `}
+            </sp-action-button>
+            <sp-overlay
```
Westbrook, line 187: We may also want to look at the `trigger` directive here, as it lists as _slightly_ more performant that leveraging the element.

packages/contextual-help/src/ContextualHelp.ts
```diff
@@ -0,0 +1,205 @@
...
+                type=${this.isMobile.matches ? 'modal' : 'auto'}
+                receives-focus="true"
+                .offset=${this.offset}
+                @sp-opened=${() => {
+                    this.overlayOpen = true;
+                }}
+                @sp-closed=${() => {
+                    this.overlayOpen = false;
+                }}
+                @slottable-request=${this.isMobile.matches
```
Westbrook, line 199: Do you want to also offer the option for a consumer to lazily render the content of this overlay?

packages/contextual-help/src/ContextualHelp.ts
Rocss: You're right, I guess I was influenced by the other examples I've seen in Express.

Rocss: I actually did not use a slot _because_ of the fact that users have too much flexibility on the heading here and can actually use other things besides heading elements, such as `p` or `div`. I wanted to kind of enforce this one to be a heading. I can of course replace it with a slot.

packages/contextual-help/src/ContextualHelp.ts
Rocss: I was not aware of this directive. I think it is really cool that is handles the `slottable-request` event for you.
A thing I did not found straight-forward is the open/close events. Is there a better way to know when it opens/closes events besides adding event listeners on the `ContextualHelp` class?
I'd need to know of open/close states for the button's `active` state. Would you consider a good addition to add `active` attribute on the trigger as a built-in feature of the `trigger` directive?

packages/contextual-help/src/spectrum-contextual-help.css
Rocss: Here it is: https://github.com/adobe/spectrum-css/blob/main/components/contextualhelp/index.css#L25 I can not find other components with this `spectrum--large` class. Maybe the `--spectrum-contextual-help-content-spacing` variable should be moved in the `large-vars` and `medium-vars`?

packages/contextual-help/src/ContextualHelp.ts
Westbrook: This is a really difficult balance to tread. We've been asked in the past by the accessibility team to ensure that patterns like this can accept multiple levels of heading to ensure the content is accessible, even if we haven't always gotten back to the related fixes 🙈. If there were other practical approaches to ensuring default delivery AND supporting varying levels of headlines, any new insight into the conversation that you might have would be greatly appreciated.

strictness vs support vs flexibility 😓

packages/contextual-help/src/ContextualHelp.ts
Rocss: This does not seem like the kind of component you interact with way too many times, as its content does not change often.
So I am thinking it would be a good practice to lazy load its content. On the other hand, its content is indeed modest. What is your opinion on this?

packages/contextual-help/src/ContextualHelp.ts
Rocss: As a consumer, the first thing I looked for was a `handleOpen` and `handleClose` callback, as `slottable-request` event is part of the internal works of this directive and I don't know if as a consumer I should be aware of that.

`sp-open`/`sp-close` events are fine, but are actually a bit delayed, because of the CSS transitions that have dispatched I guess. So the question with `handleOpen` / `handleClose` would be if the timing of calling these callbacks should be on the `handleSlottableRequest` method, or a bit late, on `sp-open` and `sp-closed`.

In my case as you mentioned `sp-open`/`sp-close` are a bit late :(

Westbrook: Yes, `sp-open` and `sp-close` are not "correct for this situation.

If we went with callbacks, they would _likely_ happen at `slottable-request` time, because that represents the _outside_ of the open/close transition, rather than only the end of each transition. But, it could be argued that this belongs at the _beginning_ of each transition, which is when the state actually changes. 🤔 

Relying on the state feels nicer than relying on a knowable part of the lifecycle, which is why I like the `TriggerController`, but there is much to think about here.

Rocss: > that best I can think of with the established API is to use specific handing of slottable-request to toggle open which is rendered to active.

Regarding this, I don't think this works, as the event is attached on the `sp-overlay`, which I don't have access to.

Westbrook: Indeed. So we can ship once and edit later leveraging `<sp-overlay>` directly, or we can hold this for an expansion of the directives capabilities. Do you have a preference on that?

packages/contextual-help/src/ContextualHelp.ts
Westbrook: I think we'd want it at some point. It is a performance optimization, so I'm OK with excluding it for the initial release, but prefer to take your lead as you're the one actually doing the work.

packages/contextual-help/src/ContextualHelp.ts
Westbrook: I kind of like that as an option, though the work to do so is non-trivial. Might be worth creating an issue to come back to that at a later date?

Rocss: Sure, I'll create an issue for this. Do you see this being useful for other components as well, such as Dialogs?

packages/contextual-help/src/ContextualHelp.ts
Rocss: Depending on when exactly would the directive be expanded. If it's more than 1-2 releases, then I'd ship it as it is and come back when this behaviour is available.

packages/contextual-help/src/ContextualHelp.ts
Rocss: I'd ship it as it is for a component which _theoretically_ should have just some text inside.

Rocss:
@Rajdeepc as per office hours discussion, I removed the `directive` usage and went back to `sp-overlay`, as it was better suited for my component.

packages/contextual-help/src/ContextualHelp.ts
Rajdeepc: Rather than a dev mode warning I would like to get it surfaced up to the documentation site to have more visibility

packages/contextual-help/src/spectrum-contextual-help.css
Rajdeepc: @Rocss Can you please confirm this with @pfulton

TarunAdobe:
Can you look into why these examples don't work for this component?
image

Rajdeepc:
@Rocss Can you make sure the VRTs are passing and the storybook controls are working as expected before we can do a final review and land this!

Rocss:
> Can you look into why these examples don't work for this component? image


@TarunAdobe seems to be due to the type of the controls: `radio`, `number`, and `text` which do not map well on `Picker`. I can convert the `radio` one into `select`, but the `number` and `text` I don't think that those can be converted. Do we have a way to exclude them from the docs, but not the storybook?

packages/contextual-help/src/spectrum-contextual-help.css
Rocss: @Rajdeepc  Since I'm surfacing a dialog on mobile, can we exclude this CSS?

TarunAdobe:
@Rocss Hmm seems like you're right! Thanks for pointing it out. We will take this up in a separate PR as this needs some work around how we create these demos.

packages/contextual-help/src/ContextualHelp.ts
TarunAdobe: @Rocss Updates on this? Do you think using `DependencyManagerController` here would benefit the component?

packages/contextual-help/test/contextual-help.test.ts
```diff
@@ -0,0 +1,72 @@
...
+        const heading = headingSlot.assignedElements()[0].textContent;
+        expect(heading).to.equal('Permission required');
+
+        const closed = oneEvent(el, 'sp-closed');
+        await sendKeys({
+            press: 'Escape',
+        });
+        await closed;
+        await nextFrame();
+        await nextFrame();
```
TarunAdobe, line 67: Why do we need to wait for a few frames here?

Rajdeepc:
@Rocss Would it be okay to surface up a `contextual-help-memory.test.ts` with this PR so that we can triage any memory leaks in the code similar to [this](https://github.com/adobe/spectrum-web-components/blob/main/packages/accordion/test/memory.test.ts)

Rajdeepc:
> @Rocss Hmm seems like you're right! Thanks for pointing it out. We will take this up in a separate PR as this needs some work around how we create these demos.

Should be fixed by #4512

packages/contextual-help/src/spectrum-contextual-help.css
Rajdeepc: Yes!

packages/contextual-help/test/contextual-help.test.ts
Rocss: Waiting for DOM clean up to complete, similar to [here](https://github.com/adobe/spectrum-web-components/blob/7a86d8c449e4cd8c8b94992139aa09a7656dd071/packages/overlay/test/overlay-directive.test.ts#L194). The test fails without it because it still finds the popover in the DOM.

blunteshwar:
Looks like  vrts's are failing!

Rocss:
@blunteshwar I observe this, however, the failing tests are not related to this PR, and the error seems to be `Error: There was no baseline image to compare against.`. Is this happening on any other PR?

TarunAdobe (approved):
LGTM!!! We will work on adding a `dependency` manager (if feasible and required) in a separate PR.

Rajdeepc (approved):
Thanks for bringing this in!

PR #4196, opened 2024-03-20
chore: use lit@3.0

Westbrook:
## Description
Lit@3.0 is coming. It's focus is on removing support for out-of-date browsers, and cleaning up deprecations internal to their APIs, with the intent of including the smallest possible breaking changes.

This clarifies any breaking/required changes for the library to get updated to it.

fixes #3773

## How has this been tested?
-   [ ] _Test case 1_
    1. All tests pass
-   [ ] _Test case 2_
    1. Storybook/the docs site are usable

## Types of changes
-   [x] Chore (minor updates related to the tooling or maintenance of the repository, does not impact compiled assets)

## Checklist
-   [x] I have signed the [Adobe Open Source CLA](http://opensource.adobe.com/cla.html).
-   [x] My code follows the code style of this project.
-   [ ] If my change required a change to the documentation, I have updated the documentation in this pull request.
-   [x] I have read the **[CONTRIBUTING](<(https://github.com/adobe/spectrum-web-components/blob/main/CONTRIBUTING.md)>)** document.
-   [ ] I have added tests to cover my changes.
-   [x] All new and existing tests passed.

PR #4292, opened 2024-04-23
chore: update to lit@3.0

Westbrook:
## Description
Updates to `lit@3.0` and actually works! This may be worth holding for 1.0, if we can actually see 1.0 coming in this quarter, but 3.0 has been the default version of Lit for a while now and there are teams being held back by not being able to start greenfield projects due to our lack of support here previously.

Leverages the `^2.0 || ^3.0` syntax for versions, but at some point a bandaid pull is appropriate here.

## Related issue(s)

- fixes https://github.com/adobe/spectrum-web-components/issues/3773
- fixes #4196

## How has this been tested?
-   [ ] _Test case 1_
    1. No functional changes
    2. Make sure everything works exactly like it did
    3. This is why we have tests!

## Types of changes

## Checklist


packages/overlay/src/Overlay.ts
```diff
@@ -565,10 +565,7 @@ export class Overlay extends OverlayFeatures {
                 `${this.tagName.toLowerCase()}-${randomID()}`
             );
         }
-        if (
-            changes.has('open') &&
-            (typeof changes.get('open') !== 'undefined' || this.open)
-        ) {
+        if (changes.has('open') && (this.hasUpdated || this.open)) {
```
Rajdeepc, line 568: if we already trying to find 'open' attributes from the PropertyValues do we also need a check on `this.open` ?

packages/overlay/test/index.ts
```diff
@@ -47,6 +46,7 @@ export const runOverlayTriggerTests = (type: string): void => {
                                     justify-content: center;
                                 }
                             </style>
+                            <input type="text" />
```
Rajdeepc, line 49: is this deliberate?

projects/documentation/scripts/build-ts.js
```diff
@@ -72,7 +72,13 @@ async function main() {
                 },
             }),
         ],
-        external: ['@spectrum-web-components/*'],
+        external: [
+            '@spectrum-web-components/*',
+            'lit-html',
```
Rajdeepc, line 77: Curious to know why the lit packages are added here?

projects/example-project-rollup/package.json
```diff
@@ -29,7 +29,7 @@
   },
   "devDependencies": {
     "@open-wc/building-rollup": "^3.0.2",
-    "@open-wc/eslint-config": "^12.0.0",
+    "@open-wc/eslint-config": "^12.0.3",
```
Rajdeepc, line 32: is this updated required in this PR?

Rajdeepc:
@Westbrook Do you think its a good idea to point out the pros of upgrading to `lit: ^3.x.x` to the consumers here in this PR?

Westbrook:
What sort of benefits do you think users might be interested in? This is a dependency of our code which is allowed to live next to alternate versions of itself, so it should be _generally_ transparent to consumers.

projects/documentation/scripts/build-ts.js
Westbrook: Because without it we've been shipping multiple copies to our docs site:

image

packages/overlay/test/index.ts
Westbrook: Yes, the test was giving issues getting into "focus visible", and this makes it easier to enforce that by allowing the test to `Tab` from the `input` to the test element.

packages/overlay/src/Overlay.ts
Westbrook: This specific says, when `open` has changed, but NOT when open has changed from `undefined` (its initial value) to `false` (its default value). That was the if only returns true when it's not the first update or the end user applied `open === true` in the DOM of the element before it was upgraded.

Rajdeepc (approved):

lschierer:
awesome!! thanks for this work

PR #4296, opened 2024-04-24
docs: update font delivery for better LCP

Westbrook:
Reduce LCP by about 0.5s by relying on a local copy of the Abode Clean font(s) rather than on TypeKit.

## How has this been tested?
-   [ ] _Test case 1_
    1. Compare this the following webpagetest resutls
    2. Previously: [2.147s](https://www.webpagetest.org/results.php?test=240425_BiDcP9_CS5&end=visual)
    3. Now: [1.701s](https://www.webpagetest.org/result/240425_BiDc2D_D8C/)
-   [ ] _Test case 2_
    1. See the Lighthouse results from this branch
    2. Previously: [98](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1714072155436-48718.report.html)
    3. Now: [99](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1714072156094-62164.report.html)


Rajdeepc:
Are you planning to test this in slow 3g?

Westbrook:
I'm specifically test why this is the result of the LCP results in slow 4G lighthouse:

image

This never seems to repaint, based on:

image

So I'm confused and poking around.

TarunAdobe (commented):
I'm curious as to why we had  this `typekit` directory in `tests/visual` earlier cz I couldn't find any test file referencing to it :(

Westbrook:
> I'm curious as to why we had this `typekit` directory in `tests/visual` earlier cz I couldn't find any test file referencing to it :(

A historical oddity. At one point in the past in order to support taking more stable visual regression testing we were including fonts for those tests locally, but when we stopped doing so it never got cleaned up completely. 🧹 

This does that AND leverages those files for something!

TarunAdobe (approved):
Looks perfect now! Thanks for clarification.

PR #4340, opened 2024-04-29
fix(number-field): select full value when using Tab to enter a field with a unit

### LINKED ISSUE #3282: [Feat]: Select all the contents of the spectrum number field when doing a tab switch.
### Code of conduct

### Impacted component(s)

Spectrum number field

In spectrum number field element, I see that if the style is ‘unit’ and we use Tab key to switch to the Number field elements, contents of the field does not get selected. If the unit is not there then switching with Tab does select the whole field.

This should be enabled, because as a user I would expect that the tab switch always selects the contents of the field for me to edit those easily. We did receive this feedback from one of Adobe AI-Web users on the prerelease channel.

### Mockups or screenshots

Tab select into field with units.

tab-select-into-field-with-units

Tab select into field without units.

tab-select-into-field-without-units

_No response_

Westbrook:
## Description
Keyboard based focus of Number Fields with units should _select_ all of the available content.

## Related issue(s)
- fixes #3282

## How has this been tested?
-   [ ] _Test case 1_
    1. Go [here](https://number-field-selection--spectrum-web-components.netlify.app/storybook/?path=/story/number-field--pixels)
    2. Use `Tab` to focus the Number Field, see that all of the content is selected
-   [ ] _Test case 2_
    1. Go [here](https://number-field-selection--spectrum-web-components.netlify.app/storybook/?path=/story/number-field--pixels)
    2. Click to focus the Number Field
    3. See that there is no selection in the Number Field

## Types of changes

## Checklist

### DIFF
--- a/packages/number-field/src/NumberField.ts
+++ b/packages/number-field/src/NumberField.ts
@@ -430,6 +430,17 @@ export class NumberField extends TextfieldBase {
         });
     }
 
+    private hasRecentlyReceivedPointerDown = false;
+
+    protected override handleInputElementPointerdown(): void {
+        this.hasRecentlyReceivedPointerDown = true;
+        this.updateComplete.then(() => {
+            requestAnimationFrame(() => {
+                this.hasRecentlyReceivedPointerDown = false;
+            });
+        });
+    }
+
     protected override handleInput(event: Event): void {
         if (this.isComposing) {
             event.stopPropagation();
@@ -758,6 +769,15 @@ export class NumberField extends TextfieldBase {
                 }
             }
             this.inputElement.inputMode = inputMode;
+        }
+        if (
+            changes.has('focused') &&
+            this.focused &&
+            !this.hasRecentlyReceivedPointerDown &&
+            !!this.formatOptions.unit
+        ) {
+            // Normalize keyboard focus entry between unit and non-unit bearing Number Fields
+            this.setSelectionRange(0, this.displayValue.length);
         }
     }
 }
--- a/packages/textfield/src/Textfield.ts
+++ b/packages/textfield/src/Textfield.ts
@@ -270,6 +270,8 @@ export class TextfieldBase extends ManageHelpText(
         this.focused = !this.readonly && false;
     }
 
+    protected handleInputElementPointerdown(): void {}
+
     protected renderStateIcons(): TemplateResult | typeof nothing {
         if (this.invalid) {
             return html`
@@ -355,6 +357,7 @@ export class TextfieldBase extends ManageHelpText(
                 .value=${live(this.displayValue)}
                 @change=${this.handleChange}
                 @input=${this.handleInput}
+                @pointerdown=${this.handleInputElementPointerdown}
                 @focus=${this.onFocus}
                 @blur=${this.onBlur}
                 ?disabled=${this.disabled}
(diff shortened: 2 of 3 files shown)


packages/textfield/src/Textfield.ts
```diff
@@ -355,6 +357,7 @@ export class TextfieldBase extends ManageHelpText(
                 .value=${live(this.displayValue)}
                 @change=${this.handleChange}
                 @input=${this.handleInput}
+                @pointerdown=${this.handleInputElementPointerdown}
```
Rajdeepc, line 360: This is interesting! Is this a deliberate to create a null event?

Rajdeepc: OH my bad. This is surfaced up in NumberField

packages/number-field/src/NumberField.ts
```diff
@@ -430,6 +430,17 @@ export class NumberField extends TextfieldBase {
         });
     }
 
+    private hasRecentlyReceivedPointerDown = false;
+
+    protected override handleInputElementPointerdown(): void {
+        this.hasRecentlyReceivedPointerDown = true;
+        this.updateComplete.then(() => {
```
Rajdeepc, line 437: Won't this cause a delay in the event to complete?

packages/number-field/src/NumberField.ts
```diff
@@ -759,5 +770,15 @@ export class NumberField extends TextfieldBase {
...
             this.inputElement.inputMode = inputMode;
         }
+        if (
+            changes.has('focused') &&
+            this.focused &&
+            !this.hasRecentlyReceivedPointerDown &&
+            !!this.formatOptions.unit
+        ) {
+            // Normalize keyboard focus entry between unit and non-unit bearing Number Fields
+            this.setSelectionRange(0, this.displayValue.length);
```
Rajdeepc, line 780: Do we need to check if displayValue has a length or not right?

packages/number-field/src/NumberField.ts
Westbrook: This is a reaction to the event having been triggered. Because it is in the promise (`updateComplete`) it will release the thread during the event and wait till the promise releases before reengaging the workflow.

packages/number-field/src/NumberField.ts
Westbrook: `displayValue` is always a string, and even if it's length is `0` the outcome here will still be correct.

packages/textfield/src/Textfield.ts
Westbrook: Yeah, it's an annoyance of the class extension system and templating that this is required but the idea is definitely that by default Textfield does nothing for `pointer` down events and then in NumberField it does something specific. It implies the benefit of a alternative composition techniques within these render methods, but that's a huge alternative project with unclear benefits and risks.

packages/number-field/src/NumberField.ts
```diff
@@ -759,5 +770,15 @@ export class NumberField extends TextfieldBase {
...
+            this.hasRecentlyReceivedPointerDown = false;
```
blunteshwar, line 781: Is it necessary to again assign `this.hasRecentlyReceivedPointerDown = false;`, when it was already false when it was checked on line 776?

Westbrook: Great question. I feel like that was required at some point, but it _shouldn't_ do anything in this version. Will confirm.

Westbrook: Removed.

Rajdeepc (approved):

Issue #4384, opened 2024-05-02
[Bug]: Combobox throws errors when options with numeric values are provided


Rocss:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-combobox

### Expected behavior

Given that I have a list of options with numeric values, when I navigate via keyboard through the options, the list should scroll and no errors should be thrown.

### Actual behavior

An error is being thrown in the console and the list does not scroll.

### Screenshots

https://github.com/adobe/spectrum-web-components/assets/13311865/770c2ef4-95f6-4a9a-933d-2008ba388719`

### What browsers are you seeing the problem in?

Chrome

### How can we reproduce this issue?

1. Go here: https://opensource.adobe.com/spectrum-web-components/storybook/index.html?path=/story/combobox--controlled
2. Navigate the options list using the keyboard.
3. Check console.
4. Observe the error thrown in the console.

### Sample code that illustrates the problem

The Combobox code does
```
        const activeEl = this.shadowRoot.querySelector(
            `#${this.activeDescendant?.value}`
        ) as HTMLElement;
```
but for instances where `value` starts with a digit `querySelector` on IDs that start with a digit is not a valid selector.

Maybe querying by the `value` attribute or querying by `[id="${this.activeDescendant?.value}"]` would work better here?

### Logs taken while reproducing problem

`Uncaught DOMException: Failed to execute 'querySelector' on 'DocumentFragment': '#55' is not a valid selector.` is being thrown

Westbrook:
In theory, if we change:
```
 const activeEl = this.shadowRoot.querySelector(
            `#${this.activeDescendant?.value}`
        ) as HTMLElement;
```
to
```
 const activeEl = this.shadowRoot.getElementById(
            this.activeDescendant?.value
        ) as HTMLElement;
```
This would no longer throw and no practical changes would be needed in the template?

Westbrook:
There's the edge case where it tries to get an `activeDescendant` that doesn't exist, but I'm not sure how practical actually hitting the code path is...

Rocss:
`getElementById` works too (even with whitespaces, which is another edge).

Another thing I noticed here is that when I select an option using the keyboard, the `change` event is not triggered.

PR #4405, opened 2024-05-07
fix(combobox): allow numeric values and trigger change event on keybo…

Rocss:
…ard selection

## Description

This PR aims to fix two issues when navigating in the Combobox using the keyboard.

## Related issue(s)

- https://github.com/adobe/spectrum-web-components/issues/4384

## Motivation and context

1. When using numeric values for Combobox options (representing IDs or quantities), an error is thrown in the console and the list does not scroll
2. When pressing "Enter" after navigating to an option, the `change` event is not triggered, as it is triggered on mouse interactions.

## How has this been tested?

-   [ ] _Test case 1_
    1. Go [here](https://rocss-numeric-combobox-fix--spectrum-web-components.netlify.app/storybook/?path=/story/combobox--controlled)
    3. Click on the Combobox input
    4. Press Arrow-Down
    5. Continue pressing Arrow-Down/Up to navigate in the options list
    6. Observe that no errors are thrown in the browser console
-   [ ] _Test case 2_
    1. Go [here](https://rocss-numeric-combobox-fix--spectrum-web-components.netlify.app/storybook/?path=/story/combobox--controlled)
    3. Click on the Combobox input
    4. Press Arrow-Down
    5. Press Enter
    6. The `change` event is triggered, `event.target.value` is `55`

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/combobox/src/Combobox.ts
+++ b/packages/combobox/src/Combobox.ts
@@ -108,9 +108,12 @@ export class Combobox extends Textfield {
     }
 
     private scrollToActiveDescendant(): void {
-        const activeEl = this.shadowRoot.querySelector(
-            `#${this.activeDescendant?.value}`
-        ) as HTMLElement;
+        if (!this.activeDescendant) {
+            return;
+        }
+        const activeEl = this.shadowRoot.getElementById(
+            this.activeDescendant.value
+        );
         if (activeEl) {
             activeEl.scrollIntoView({ block: 'nearest' });
         }
@@ -211,6 +214,7 @@ export class Combobox extends Textfield {
             return;
         }
         this.value = this.activeDescendant.itemText;
+        this.handleChange();
     }
 
     public filterAvailableOptions(): void {
--- a/packages/combobox/src/Combobox.ts
+++ b/packages/combobox/src/Combobox.ts
@@ -213,7 +213,6 @@ export class Combobox extends Textfield {
         if (!this.activeDescendant) {
             return;
         }
-        this.value = this.activeDescendant.itemText;
 
         const activeEl = this.shadowRoot.getElementById(
             this.activeDescendant.value
--- a/packages/combobox/src/Combobox.ts
+++ b/packages/combobox/src/Combobox.ts
@@ -214,7 +214,13 @@ export class Combobox extends Textfield {
             return;
         }
         this.value = this.activeDescendant.itemText;
-        this.handleChange();
+
+        const activeEl = this.shadowRoot.getElementById(
+            this.activeDescendant.value
+        );
+        if (activeEl) {
+            activeEl.click();
+        }
     }
 
     public filterAvailableOptions(): void {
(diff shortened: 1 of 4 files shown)


packages/combobox/src/Combobox.ts
```diff
@@ -211,6 +214,7 @@ export class Combobox extends Textfield {
             return;
         }
         this.value = this.activeDescendant.itemText;
+        this.handleChange();
```
Westbrook, line 217: It looks like `selectDescendant()` is _only_ used when the `Enter` key is pressed. I wonder if we can get the same, but slightly better, simply by `this.activeDescendant.click()`. In theory, this would trigger the lifecycle of the `<sp-menu>` and hit all of the appropriate `change`, etc. pathways ensuing all selections were "the same".

packages/combobox/test/combobox-a11y.test.ts
```diff
@@ -354,4 +354,31 @@ describe('Combobox accessibility', () => {
 
         expect(el.open).to.be.false;
     });
+    it('manages keyboard navigation and selection', async () => {
```
Westbrook, line 357: Can we include testing for numeric values, too?

Rocss: Sure! 😄  More than what I already added in `packages/combobox/test/combobox.data.test.ts`? 
Or a combination of numeric values + selection testing?

packages/combobox/src/Combobox.ts
Rocss: `this.activeDescendant.click()` does not really work, since it is defined as `MenuItem | ComboboxOption`, but what can be done and works is similar to what's inside `scrollToActiveDescendant`, but instead of scroll you click.

packages/combobox/src/Combobox.ts
```diff
@@ -211,6 +214,13 @@ export class Combobox extends Textfield {
             return;
         }
         this.value = this.activeDescendant.itemText;
+
+        const activeEl = this.shadowRoot.getElementById(
+            this.activeDescendant.value
+        );
+        if (activeEl) {
+            activeEl.click();
+        }
```
Westbrook, line 223: Does this change also trigger the code in line 216 to happen in `handleInput`, as well?

packages/combobox/test/combobox-a11y.test.ts
Westbrook: Oh, somehow I didn't see that test. I think we just need to press `Enter` on something there to confirm it's fully hit the `getElementById()` path, and then this will be good to go!

packages/combobox/src/Combobox.ts
Rocss: Ah indeed.
It does not trigger `handleInput`, buuuut it does trigger `handleMenuChange` which does `this.value = selected?.itemText || '';` which is same thing as line 216.

packages/combobox/test/combobox-a11y.test.ts
Rocss: I combined them into one test since it was 90% the same, just the data was different. 🙈

Westbrook (approved):
Awesome! Gonna catch up `main` and then merge when Ci is green. Many thanks 🙇🏼

Issue #4407, opened 2024-05-07
[Bug]: Checkbox and Overlay are missing lit dependencies in 0.42.3 release


jcmitch:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Checkbox, Overlay

### Expected behavior

Should be able to install components without error or without needing to include additional dependencies. 

### Actual behavior

Error messages when installing.

```
CheckboxMixin.d.ts:1:33 - error TS2307: Cannot find module 'lit' or its corresponding type declarations.

1 import { ReactiveElement } from 'lit';
                                  ~~~~~

InteractionController.d.ts:1:41 - error TS2307: Cannot find module 'lit' or its corresponding type declarations.

1 import type { ReactiveController } from 'lit';
                                          ~~~~~

slottable-request-directive.d.ts:2:32 - error TS2307: Cannot find module 'lit/async-directive.js' or its corresponding type declarations.

2 import { AsyncDirective } from 'lit/async-directive.js';
```

### How can we reproduce this issue?

1. Attempt to install 0.42.3 outside of an environment which includes lit

Westbrook:
Closed by #4416

PR #4416, opened 2024-05-08
fix(base): move lit imports to base

jcmitch:
Fixing issue with various lit imports introduced in 0.42.3 release.

## Description

Both Checkbox and Overlay included imports directly from lit while neither package has lit as a direct dependency. Moving these imports to pull from base as this seems to be the established standard. 

## Related issue(s)

- https://github.com/adobe/spectrum-web-components/issues/4407

## Motivation and context

Unable to install the latest release without needing to add additional dependencies. 

## How has this been tested?

Successfully builds

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/checkbox/src/CheckboxMixin.ts
+++ b/packages/checkbox/src/CheckboxMixin.ts
@@ -10,8 +10,11 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 
-import { ReactiveElement } from 'lit';
-import { html, TemplateResult } from '@spectrum-web-components/base';
+import {
+    html,
+    ReactiveElement,
+    TemplateResult,
+} from '@spectrum-web-components/base';
 import {
     property,
     query,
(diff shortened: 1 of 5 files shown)


tools/base/src/index.ts
```diff
@@ -13,3 +13,4 @@ governing permissions and limitations under the License.
 export * from './Base.js';
 export * from './sizedMixin.js';
 export * from 'lit';
+export * from 'lit/async-directive.js';
```
Westbrook, line 16: Great to see this in the repro!

Please move this to a `async-directive.ts` file that mirrors `directive.ts` so that tree shaking is less required to exclude this code from a project.

Then I think we're good to ship this!

jcmitch: Done

Westbrook (commented):
It's not fully clear that the CI fails for this, but can you run `yarn` with you changes and then commit the updates, then we'll be good to merge this.

Apologies if I've said similar before (or if I end up needing to say similar again). 😬 👍🏼

Westbrook (approved):
This looks great. I'll catch this branch up to `main` and if we're still green :shipit:

Issue #4415, opened 2024-05-08
[Bug]: Slider values are truncated and divided by 1000 - when using values > 1000


spdev3000:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Slider

### Expected behavior

When an editable slider is used with values > 1000 - those values should be displayed without truncation

```
<sp-slider 
        label="Slider with values > 1000" 
        min=0
        max=5000
        step="1"
        editable
        hide-stepper
      ></sp-slider>
``` 

### Actual behavior

Using an editable slider with values > 1000 those values will be truncated by 1000 and rendered into input field

### Screenshots

![Screenshot 2024-05-08 at 11 06 26](https://github.com/adobe/spectrum-web-components/assets/29090947/0947a4b9-bb63-4e48-b085-207b5102c30d)

https://github.com/adobe/spectrum-web-components/assets/29090947/c970c5ef-9359-4193-8951-9262adfa6ef5

### How can we reproduce this issue?

1. Go to https://studio.webcomponents.dev/edit/MhXptBuyqtzL3DXAK3Cj/src/index.ts?p=stories
2. Click on slider and adjust values to be greater than 999
3. See input field renders values divided by 1000, so 1 instead of 1000 and 2 instead of 2000 are rendered, etc.

spdev3000:
Probably was introduced with: https://github.com/adobe/spectrum-web-components/pull/4263

Westbrook:
@blunteshwar please look at this with priority!

spdev3000:
Thanks!

PR #4417, opened 2024-05-08
fix(number-field): handles values greater than 1000

blunteshwar:
## Description

This PR addresses an issue where large numbers were being incorrectly formatted or rounded in the valueFormatter of the component. The problem was occurring due to the default behavior of the NumberFormatter which was using grouping for large numbers. This caused values greater than 1000 to be formatted incorrectly.

### Changes Made:

Added the `useGrouping: false` option to the NumberFormatter instantiation in the valueFormatter method.
This change ensures that large numbers are formatted without grouping, preventing the rounding or truncation issues observed previously.

## Related issue(s)

- #4415 

## Motivation and context

This fix improves the accuracy of number formatting for large numbers in the component, ensuring that they are displayed correctly to the user.

## How has this been tested?
Tested the fix with various large numbers and confirmed that they are now formatted correctly without any rounding or truncation.
Checked that the formatting behavior for numbers less than 1000 remains unchanged.

-   [ ] _Test case 1_
    1. Go [here](https://blunteshwar-truncating-values--spectrum-web-components.netlify.app/storybook/?path=/story/number-field--quiet&args=value:995;min:4;max:3400;step:1)
    2. Increment the value beyond 1000.
-   [ ] _Test case 2_
    1. Go [here](https://blunteshwar-truncating-values--spectrum-web-components.netlify.app/storybook/?path=/story/number-field--quiet&args=value:995;min:4;max:3400;step:1)
    2. enter any value above 1000

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/number-field/test/number-field.test.ts
+++ b/packages/number-field/test/number-field.test.ts
@@ -146,9 +146,7 @@ describe('NumberField', () => {
             expect(el.valueAsString).to.equal('-2.4');
             expect(el.focusElement.value).to.equal('-2.4');
         });
-    });
-    describe('it handles values greater than 1000', () => {
-        it('correctly handles values greater than 1000 with step=1', async () => {
+        it('correctly handles max values greater than 1000 with step=1', async () => {
             const el = await getElFrom(
                 Default({
                     step: 1,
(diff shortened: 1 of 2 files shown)


packages/number-field/test/number-field.test.ts
```diff
@@ -147,6 +147,31 @@ describe('NumberField', () => {
             expect(el.focusElement.value).to.equal('-2.4');
         });
     });
+    describe('it handles values greater than 1000', () => {
+        it('correctly handles values greater than 1000 with step=1', async () => {
```
Westbrook, line 151: Can we bump this test up into the `describe('Step', () => {` block so we don't need the new `describe`, please.

Rajdeepc (approved):

spdev3000:
Thanks for the fast fix!

PR #4422, opened 2024-05-08
ci: run unit and memory tests separately

Westbrook:
## Description
Separate unit and memory tests for CI time efficiencies.

@Rajdeepc @TarunAdobe it looks like we were previously skipping all of the memory tests in CI, because CI runs against non-memory tooling Chrome by default which skips those tests. This move puts the memory tests back into the workflow and four of them seem to be failing. Would yall prefer to work against this branch or to see them skipped herein so you can PR them back on, in a passing state, in individual PRs, later?

I'll also test whether we can run memory tests with concurrency in this setup.

## How has this been tested?
-   [ ] _Test case 1_
    1. [Go to CI](https://app.circleci.com/pipelines/github/adobe/spectrum-web-components/21745/workflows/c6de2dcf-9667-44dc-916d-baf2663a04f7)
    2. See that [test-chromium](https://app.circleci.com/pipelines/github/adobe/spectrum-web-components/21745/workflows/c6de2dcf-9667-44dc-916d-baf2663a04f7/jobs/567560) and [test-chromium-memory](https://app.circleci.com/pipelines/github/adobe/spectrum-web-components/21745/workflows/c6de2dcf-9667-44dc-916d-baf2663a04f7/jobs/567579) are now separate jobs
    3. See that allowing concurrency in Chrome while unit testing gets the job [under 4 minutes](https://app.circleci.com/pipelines/github/adobe/spectrum-web-components/21745/workflows/c6de2dcf-9667-44dc-916d-baf2663a04f7/jobs/567560) with appropriately cached
    4. See that the memory tests have [some failures](https://app.circleci.com/pipelines/github/adobe/spectrum-web-components/21745/workflows/c6de2dcf-9667-44dc-916d-baf2663a04f7/jobs/567579)

## Types of changes

## Checklist

### DIFF
--- a/web-test-runner.utils.js
+++ b/web-test-runner.utils.js
@@ -17,7 +17,6 @@ import fg from 'fast-glob';
 
 export const chromium = playwrightLauncher({
     product: 'chromium',
-    concurrency: 1,
     createBrowserContext: ({ browser }) =>
         browser.newContext({
             ignoreHTTPSErrors: true,
@@ -27,6 +26,29 @@ export const chromium = playwrightLauncher({
 
 export const chromiumWithMemoryTooling = playwrightLauncher({
     product: 'chromium',
+    createBrowserContext: ({ browser }) =>
+        browser.newContext({
+            ignoreHTTPSErrors: true,
+            permissions: ['clipboard-read', 'clipboard-write'],
+        }),
+    launchOptions: {
+        headless: false,
+        args: [
+            '--js-flags=--expose-gc',
+            '--headless=new',
+            /**
+             * Cause `measureUserAgentSpecificMemory()` to GC immediately,
+             * instead of up to 20s later:
+             * https://web.dev/articles/monitor-total-page-memory-usage#local_testing
+             **/
+            '--enable-blink-features=ForceEagerMeasureMemory',
+        ],
+    },
+});
+
+export const chromiumWithMemoryToolingCI = playwrightLauncher({
+    product: 'chromium',
+    concurrency: 1,
     createBrowserContext: ({ browser }) =>
         browser.newContext({
             ignoreHTTPSErrors: true,
--- a/web-test-runner.utils.js
+++ b/web-test-runner.utils.js
@@ -48,7 +48,7 @@ export const chromiumWithMemoryTooling = playwrightLauncher({
 
 export const chromiumWithMemoryToolingCI = playwrightLauncher({
     product: 'chromium',
-    concurrency: 1,
+    concurrency: 2,
     createBrowserContext: ({ browser }) =>
         browser.newContext({
             ignoreHTTPSErrors: true,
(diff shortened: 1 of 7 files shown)


Rajdeepc:
@Westbrook I checked these failures and it seems like there can be small memory leaks happening on these elements. Lets skip these 4 from here and we can take it up in separate PRs. I will add 4 internal tickets to track these.

web-test-runner.utils.js
```diff
@@ -47,6 +46,29 @@ export const chromiumWithMemoryTooling = playwrightLauncher({
     },
 });
 
+export const chromiumWithMemoryToolingCI = playwrightLauncher({
+    product: 'chromium',
+    concurrency: 1,
```
Westbrook, line 51: @Rajdeepc while you're working through the new VRT failures here, can you test whether the separation of unit and memory tests allows us to leverage concurrency to get the memory test into the same general time scale as the unit/list/preview tasks? If we could get this one under 5 minutes as well, that would be great!

Rajdeepc:
Adding `ifDefined` directive to the style attributes fixes these errors. Every time the element is rendered, new string is getting created for the style attribute, even if the value has not changed. This is leading to memory leaks.

Westbrook:
> Adding `ifDefined` directive to the style attributes fixes these errors. Every time the element is rendered, new string is getting created for the style attribute, even if the value has not changed or assigned. This is leading to memory leaks.

Doesn't seem that it _ever_ creates the style string in that context: https://separate-memory-and-unit-testing--spectrum-web-components.netlify.app/storybook/?path=/story/color-area--default Would it be better to cache the string or to use a `styleMap()` in attempt to deliver the component correct and yet not leak?

Rajdeepc:
> > Adding `ifDefined` directive to the style attributes fixes these errors. Every time the element is rendered, new string is getting created for the style attribute, even if the value has not changed or assigned. This is leading to memory leaks.
> 
> Doesn't seem that it _ever_ creates the style string in that context: https://separate-memory-and-unit-testing--spectrum-web-components.netlify.app/storybook/?path=/story/color-area--default Would it be better to cache the string or to use a `styleMap()` in attempt to deliver the component correct and yet not leak?

Adding a `styleMap` not caching or abstracting the value rather I found out that it is leaking some unrelated memory which is a thing to check. But to make up for the lost VRTs, `ifDefined` has been replaced with `styleMap`

Rajdeepc:
@Westbrook I have passed all the tests, fixing the Swatch memory leak. Will take up `color-area, color-wheel and color-slider` in a separate PR and unblock this going to `main`. Feel free to add your thoughts so that we can land this!

web-test-runner.utils.js
Rajdeepc: Screenshot 2024-05-10 at 4 27 52 PM

Rajdeepc (approved):

PR #4442, opened 2024-05-10
chore(theme): S2 foundations

TarunAdobe:
## Summary
This PR introduces changes to the Spectrum Web Components (SWC) project to facilitate the transition from Spectrum 1 (S1) to Spectrum 2 (S2) CSS. It includes the addition of a new index file `(component/{bridge,spectrum,express}/index.css)` in the `@spectrum-css/tokens@14.1.0-alpha.3` package to assist in loading component-specific CSS. Additionally, it updates the SWC library to process the bridge file (referred to as legacy-overrides in SWC for now) and import it into the components.

## Changes Made

1.`@spectrum-css` added a new bridge file in the `@spectrum-css/tokens@14.1.0-alpha.3` package to aid in loading component-specific CSS from the tokens module directly.
2. Updated the `generate-token` script to include component's system levels in `global-vars.css`
3. Updated the `process-spectrum` script to create a `legacy-override` css file from the `@spectrum-css/tokens.../bridge.css`

Note: Investigated and identified VRT failures due to changes in global CSS tokens in the new `sp-action-button` from the CSS side. For example, `--mod-actionbutton-content-color-default` now points to `gray-25` instead of `gray-50 `for the selected state, which might be a breaking change for S1.

## Next Steps
Ensure there are no breaks to the existing code before merging.

## Motivation and context
This PR is important to bring in S1/Express core tokens with S2 overrides keeping in mind
- can be reversed via a theme (back-portable to S1/express)
- change without removing APIs
- change without impacting layout

## How has this been tested?

-  VRTs off the `main` branch should just work as it is !!

## Types of changes

## Checklist

## Best practices


scripts/generate-tokens-wrapper.js
```diff
@@ -134,13 +134,77 @@ const processTokens = (srcPath, tokensDir) => {
...
+            css
+        );
+    } catch (er) {}
 };
 
 const processPackages = async (srcPath, tokensDir, index) => {
+    if (tokensDir === 'tokens') {
+        const packagename = tokenPackages[index];
+
+        let componentLevelTokensPath = path.join(
```
castastrophe, line 149: Question: instead of hardcoding to the node_modules path should this leverage require.resolve?

Rajdeepc: Good catch!

scripts/generate-tokens-wrapper.js
```diff
@@ -134,13 +134,77 @@ const processTokens = (srcPath, tokensDir) => {
...
+        );
+        const spectrumPath = path.join(
+            componentLevelTokensPath,
+            'spectrum',
+            packagename + '.css'
+        );
+        // check if spectrumPath exists
+        if (fs.existsSync(spectrumPath)) {
+            let spectrum = fs.readFileSync(spectrumPath, 'utf8');
+            spectrum = removeImporantComments(targetHost(spectrum));
```
castastrophe, line 167:
```suggestion
            spectrum = removeImportantComments(targetHost(spectrum));
```

Small spelling note ^

castastrophe: Oh question - I'm assuming this function removes the copyright which you're also adding to files - why not just keep the copyright and strip duplicates after merge?  The `postcss-discard-comments` plugin offers a `removeAllButFirst` option which is a nice way to preserve the first copyright in the file.

blunteshwar (approved):

blunteshwar:
LGTM!

tools/styles/package.json
```diff
@@ -112,10 +113,10 @@
...
     },
     "devDependencies": {
-        "@spectrum-css/commons": "^10.0.0",
+        "@spectrum-css/commons": "^11.0.0-s2-foundations.9",
         "@spectrum-css/expressvars": "^3.0.9",
-        "@spectrum-css/tokens": "^14.0.0",
-        "@spectrum-css/tokens-v2": "npm:@spectrum-css/tokens@^14.0.0-next.3",
+        "@spectrum-css/tokens": "^@spectrum-css/tokens@14.3.1",
+        "@spectrum-css/tokens-v2": "npm:@spectrum-css/tokens@^15.0.0-s2-foundations.9",
         "@spectrum-css/typography": "^6.1.0",
```
nikkimk, line 120: @TarunAdobe should this be `"@spectrum-css/typography": "^7.0.0-s2-foundations.9",`

tasks/process-spectrum.js
```diff
@@ -169,7 +169,8 @@ async function processComponent(componentPath) {
      */
     for await (const conversion of conversions) {
         const sourcePath = require.resolve(conversion.inPackage);
```
nikkimk, line 171: Should this be `const sourcePath = require.resolve(conversion.inPackage).replace('index.css', 'index-base.css');`?

packages/accordion/src/system-overrides.css
```diff
@@ -0,0 +1,320 @@
+/*
```
nikkimk, line 1: Shouldn't this be imported by by `accordion.css`?

TarunAdobe: yess!!! thnx for the catch

tools/styles/package.json
TarunAdobe: This was creating some errors back when I tried... Will revisit this again tommoorrow. Thank you for reminding

tasks/process-spectrum.js
TarunAdobe: Updated this... However after changing it I am seeing some un-expected UI changes. Tried to trace down why but haven't figured out properly yet.

nikkimk: The system-overrides to be imported after each component's index-base css. Otherwise the UI will be off.

TarunAdobe: I updated it thnx. Please verify.

TarunAdobe (commented):
@nikkimk Should we be importing the overrides in the components like this or should we keep this task tightly coupled to the css files only as we do for other component overrides? 
For ex. in `accordion` we import the overrides in `accordion.css` file instead of adding those in the `Accordion.ts` file.
for icons we don't have custom css files and that's why I understand you've added these overrides in the `.ts` files itself. But this would affect the performance if I understand correctly as we are importing multiple css files in the component. Instead in the earlier case since we were importing only one css file for each component this was a better performant system. What are your thoughts on this?

packages/accordion/src/AccordionItem.ts
```diff
@@ -22,6 +22,7 @@ import { Focusable } from '@spectrum-web-components/shared/src/focusable.js';
 import { when } from '@spectrum-web-components/base/src/directives.js';
 import '@spectrum-web-components/icons-ui/icons/sp-icon-chevron100.js';
 import chevronIconStyles from '@spectrum-web-components/icon/src/spectrum-icon-chevron.css.js';
+import chevronIconOverrides from '@spectrum-web-components/icon/src/icon-chevron-overrides.css.js';
```
TarunAdobe, line 26: @nikkimk Should we be importing the overrides in the components like this or should we keep this task tightly coupled to the css files only as we do for other component overrides? 
For ex. in `accordion` we import the overrides in `accordion.css` file instead of adding those in the `Accordion.ts` file.
for icons we don't have custom css files and that's why I understand you've added these overrides in the `.ts` files itself. But this would affect the performance if I understand correctly as we are importing multiple css files in the component. Instead in the earlier case since we were importing only one css file for each component this was a better performant system. What are your thoughts on this?

Rajdeepc: Let's not remove anything in this PR, better we should create a parallel PRs for removals.

Issue #4472, opened 2024-05-21
[Bug]: sp-table-head-cell with sortable cannot be activated by keyboard


jnurthen:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-table-head-cell

### Expected behavior

Should be able to be sorted using the keyboard

### Actual behavior

neither space nor enter will sort the column

### What browsers are you seeing the problem in?

Firefox, Chrome, Safari, Microsoft Edge

### How can we reproduce this issue?

1. Go to https://opensource.adobe.com/spectrum-web-components/storybook/index.html?path=/story/table-virtualized--virtualized
2. Tab until you get to the sortable column title
3. Press enter 
4. Nothing happens
5. Press space
6. page scrolls

PR #4473, opened 2024-05-21
fix: add keyboard handlers to sp-table-cell-head

jnurthen:
## Description

Adds keyboard handlers to sp-table-cell-head
## Related issue(s)

#4472

## Motivation and context

Keyboard handlers needed for accessibility.

## How has this been tested?

-   [x] _Test case 1_
    1. Go to [Storybook](https://jnurthen-issue4472--spectrum-web-components.netlify.app/storybook/)
    2. Tab to Column Title (sortable)
    3. Press Enter
    4. sort activates
    5. Press and release space
    6. sort activates

## Types of changes

## Checklist

## Best practices


packages/table/src/TableHeadCell.ts
```diff
@@ -58,6 +58,45 @@ export class TableHeadCell extends SpectrumElement {
...
+        switch (code) {
+            case 'Space':
+                this.removeEventListener('keyup', this.handleKeyup);
+                this.click();
+                break;
+            /* c8 ignore next 2 */
+            default:
+                break;
+        }
+    }
```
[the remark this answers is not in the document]
jnurthen: honestly I copied this from swatch which was doing the same thing. Happy to simplify if that makes sense

packages/table/src/TableHeadCell.ts
```diff
@@ -58,6 +58,45 @@ export class TableHeadCell extends SpectrumElement {
...
+        switch (code) {
+            case 'Space':
+                event.preventDefault();
+                this.addEventListener('keyup', this.handleKeyup);
+                break;
+            /* c8 ignore next 2 */
+            default:
+                break;
+        }
+    }
```
Westbrook, line 72: Keydown, in previous consumptions, in this way was meant to allow for `this.active = true` while the `Space` key was doing to better mimic the interaction you get with a `<button>` element. This element currently includes styles for this interaction, but not the property to power those styles. We probably should...

Can you add this to the beginning of the class:
```
public active = false;
```

`this.active = true;` here. And `this.active = false;` to the `handleKeyup()` logic? Then we should be good to go here.

packages/table/src/TableHeadCell.ts
```diff
@@ -58,6 +61,47 @@ export class TableHeadCell extends SpectrumElement {
...
+            case 'NumpadEnter':
+                this.click();
+                break;
+            /* c8 ignore next 2 */
+            default:
+                break;
+        }
+    }
+
+    protected handleKeyup(event: KeyboardEvent): void {
```
blunteshwar, line 91: Why is 'space' handled in `handleKeyup` and `handleKeydown` and not directly in `handleKeypress` like `enter`

jnurthen: Space and enter work slightly differently on native buttons.
Enter activates on the press event, space activates on the up event. This is intended to mimic that. A user should be able to press and hold space on a button, tab away from the button and then release space without activating it.

blunteshwar: Understood!

blunteshwar (approved):

packages/table/src/TableHeadCell.ts
Rajdeepc: @jnurthen Can you please make sure this is added as requested above!

Rajdeepc (changes requested):
One small addition needs to go in to the definition of this class. Can you make sure it is added before we can land this!

packages/table/src/TableHeadCell.ts
jnurthen: I think I did in https://github.com/adobe/spectrum-web-components/pull/4473/commits/550b46a5320f5baf36887dd2e9e6957a5ef6f062

blunteshwar (approved):

Rajdeepc (approved):
LGTM!

PR #4491, opened 2024-05-23
build: ensure that token files are constructed from empty

Westbrook:
Clear the `global-vars.css` files before appending to them.


Westbrook (commented):
Reviewed by @TarunAdobe in person.

PR #4516, opened 2024-05-29
chore(progress-bar): updated tabindex on firstupdated to allow for screen-readers

Rajdeepc:
## Description

Updated `tabindex=0` on `firstUpdated` lifecycle to attach a tab stop to the element so that it can be picked up by the screen readers.

## Related issue(s)

-

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/progress-bar/src/ProgressBar.ts
+++ b/packages/progress-bar/src/ProgressBar.ts
@@ -119,6 +119,9 @@ export class ProgressBar extends SizedMixin(
         if (!this.hasAttribute('role')) {
             this.setAttribute('role', 'progressbar');
         }
+        if (!this.hasAttribute('tabindex')) {
+            this.setAttribute('tabindex', '0');
+        }
     }
 
     protected override updated(changes: PropertyValues): void {


packages/progress-bar/src/ProgressBar.ts
```diff
@@ -119,6 +119,9 @@ export class ProgressBar extends SizedMixin(
         if (!this.hasAttribute('role')) {
             this.setAttribute('role', 'progressbar');
         }
+        if (!this.hasAttribute('tabindex')) {
```
blunteshwar, line 122: This effectively sets a default tabindex of 0 when the attribute is not present. However, I have a question: what if tabindex is set to -1 initially?

In such cases, this logic will not alter the tabindex, leaving it at -1, which excludes the element from the tab order. Should we consider handling this scenario explicitly to ensure consistent keyboard accessibility?

blunteshwar:
Can you also add a relevant test for this.

packages/progress-bar/src/ProgressBar.ts
Rajdeepc: To create a tab stop on this from tabIndex = -1 it has to be in a roving tab index. Otherwise this progress bar will not find itself in a screen-reader!

blunteshwar: Why can't we skip this check `if (!this.hasAttribute('tabindex'))` and directly assign `progress-bar` a `tabIndex=0` by using just `this.setAttribute('tabindex', '0');`

blunteshwar:
I noticed that some sections of this PR are not fully completed. Could you please fill out all the necessary details and ensure that everything is properly documented?

packages/progress-bar/src/ProgressBar.ts
blunteshwar: Understood!

blunteshwar (approved):

nikkimk:
What would be the use case to make a progressbar focusable given that progressbars are not interactive, and that `aria-busy` and `aria-live` are better ways to communicate updates to the element?

Rajdeepc:
Closing this since aria-live should take care of this!
https://www.magentaa11y.com/checklist-web/progress/

PR #4528, opened 2024-05-31
chore(search): added logic to hold search value on escape

Rajdeepc:
## Background
The current implementation uses `type="search"` on the input element, causing the search value to be cleared whenever the `ESC` key is pressed. This default behavior is not always desirable.

## Objective
To provide an API that allows users to determine whether the search value should be retained when the `ESC` key is pressed. This will be achieved through a new property `holdValueOnEscape`.

## Implementation Details
Property Introduction: A new Boolean property `holdValueOnEscape` is introduced. When set to `true`, the `type `attribute of the input element is set to `'text'`, preventing the default behavior of clearing the input value on `ESC`.

Default Behavior: By default, `holdValueOnEscape` is true, ensuring the value is retained when `ESC` is pressed.
Conditionally Setting Type: In the `firstUpdated` lifecycle method, the type attribute of the input element is conditionally set to 'text' based on the `holdValueOnEscape` property.

## Related issue(s)

-

## How has this been tested?

-   [ ] Storybook
    1. Go [here](https://feature-search-escape--spectrum-web-components.netlify.app/storybook/?path=/story/search--hold-value-on-escape)
    2. Type something on the Searchbox
    3. Press ESC
    4. See the value doesn't reset
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/search/src/Search.ts
+++ b/packages/search/src/Search.ts
@@ -55,6 +55,9 @@ export class Search extends Textfield {
     @property()
     public override placeholder = 'Search';
 
+    @property({ type: Boolean })
+    public holdValueOnEscape!: boolean;
+
     @query('#form')
     public form!: HTMLFormElement;
 
@@ -72,6 +75,9 @@ export class Search extends Textfield {
 
     private handleKeydown(event: KeyboardEvent): void {
         const { code } = event;
+        if (code === 'Escape' && this.holdValueOnEscape) {
+            return;
+        }
         if (!this.value || code !== 'Escape') {
             return;
         }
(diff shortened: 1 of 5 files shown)


packages/search/src/Search.ts
```diff
@@ -72,6 +75,9 @@ export class Search extends Textfield {
 
     private handleKeydown(event: KeyboardEvent): void {
         const { code } = event;
+        if (code === 'Escape' && this.holdValueOnEscape) {
+            return;
+        }
```
blunteshwar, line 80: Rather than two `if` conditions can we achieve this using a single if condition like this
```
if(code === 'Escape' && !this.holdValueOnEscape)
{
    this.reset();
}
return;
```

blunteshwar (changes requested):
Can you look into the failing VRTs

TarunAdobe (approved):

Issue #4531, opened 2024-06-03
[Bug]: Inconsistent behavior of sp-number-field in Safari on iOS


mihaiudvuleanu:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-number-field

### Expected behavior

The number grouping separator `,` should not be interpreted as a decimal point. For example if i edit a number displayed as "1,000" and I delete the last 0 digit, the result should be "100".

### Actual behavior

While editing a number, the grouping separator is interpreted as a decimal point and, in the above scenario, the result is 1.

### Screenshots

![number-field](https://github.com/adobe/spectrum-web-components/assets/171576787/7ea55bb4-1bef-412e-8bb1-408491d6b623)

### What browsers are you seeing the problem in?

Safari / iOS 17.5.1

### How can we reproduce this issue?

1. Navigate to https://opensource.adobe.com/spectrum-web-components/components/number-field/
2. Edit the first editable box on the page (from Sizes), by deleting the last digit (try to change it from 1024 to 102)
3. Hit Enter, now I'm expecting to see 102 in the text box
4. Instead, it displays 1.02 (the grouping separator was interpreted as a decimal point)

rubencarvalho:
Fixed by https://github.com/adobe/spectrum-web-components/pull/4571

PR #4552, opened 2024-06-10
[DO NO MERGE]: fix(number-field): separator interpreted as decimal

mizgaionutalexandru:
## Description

Updated the special `convertValueToNumber` case for iPhones to correctly identify user's intent of using a decimal point.
Also changed the `,` or `.` that's being replaced with the website language specific decimal character from the first occurrence left-to-right to right-to-left. (e.g. for 1,234,56 previously the value would've been 1.23456, now it is 1,234.56 to better reflect user's intent).

## Related issue(s)
- https://github.com/adobe/spectrum-web-components/issues/4531

## Motivation and context

This solves the problem of a comma used as a separator in the number-field being interpreted as a decimal point when the user didn't want that. These changes came as a necessity since iOS keyboard shows either `,` or `.` based on device's settings.

## How has this been tested?

-   [x] _Test case 1_
    1. Open storybook on an iPhone
    2. Select the number-field component
    3. Enter a value that would have a separator, such as 1000
    4. Blur the number-field
    5. Remove a digit from the input's value
    6. Blur the number-field (the value should be 100)
-   [x] _Test case 2_
    1. Open storybook on an iPhone
    2. Select the number-field component
    3. Enter a value that would have a separator and a decimal, such as 1000.5
    4. Blur the number-field
    5. Remove/Add a digit from/to the input's value
    6. Blur the number-field
-   [x] _Test case 3_
    1. Open storybook on an iPhone
    2. Select the number-field component
    3. Enter a value that would have a separator and a decimal, such as 1000.5
    4. Blur the number-field
    5. Remove/Add a digit and a comma/dot from/to the input's value
    6. Blur the number-field

Below you can find videos of some of these testings done on two different types of keyboard.

https://github.com/adobe/spectrum-web-components/assets/67201262/ca9e1781-ff1f-437a-8139-2bfa4c4df148

https://github.com/adobe/spectrum-web-components/assets/67201262/9486c881-2e98-4284-b538-a833f2ea9ba2

## Types of changes

## Checklist

## Best practices


Rajdeepc:
Can you please add a test here!

packages/number-field/test/number-field.test.ts
```diff
@@ -117,6 +117,40 @@ describe('NumberField', () => {
             expect(el.focusElement.value).to.equal('13 377 331');
         });
     });
+    xit('correctly interprets decimal point on iPhone', async () => {
```
mizgaionutalexandru, line 120: I created a test with some common usecases but haven't been able to run it since I couldn't find a way to mock isIphone.

mizgaionutalexandru:
This approach seems to fix the issue in most cases. That being said, I've found a corner case caused by the following flow:

1. user clicks on number-field with a non-decimal value and with groupings (e.g. 3,000,000)
2. user types `,`/`.` (based on iOS device language settings) and a number - boolean isIntentDecimal is set to true
3. user deletes said `,`/`.` and number - boolean still true
4. user blurs number-field, now the value becomes 3,000.0

Maybe there is a comparison to be made so that we can reset isIntentDecimal? We should also consider the position of said `,`/`.` to verify that the character deleted was the one added by the user, not a grouping/decimal separator that already existed in the string value.
I think this PR solves most problems with this caveat. Maybe we should consider a long-term solution in another issue/PR, be the approach explained above or others. How can we best approach this?

blunteshwar (approved):

blunteshwar:
Some tests are failing. Can you have a look at them. Other than that LGTM!

mizgaionutalexandru:
Looks like the Lighthouse CI check problem is similar with this one: https://github.com/adobe/spectrum-web-components/pull/4269#pullrequestreview-2028457037

I think the performance check is also related given the fact that the storybook link is not generated correctly.

Rajdeepc:
@mizgaionutalexandru I can see you have polluted the commit history. I see 4 commits from main are being logged here! Can you please close this PR and create a fresh PR with these changes?
Thank you!

Issue #4559, opened 2024-06-12
[Feat]: Fill-start in slider should respect 0-values as well as normalization function


spdev3000:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Silder

### Description of the requested feature

1. Currently `fill-start` attribute only allows to set values !== 0, but sometimes clients have sliders with min<0 , so for example: `min="-100", max="200"` - so we need a possibility to define `fill-start="0"` to not start from centered track position
2. Also currently the fill-start calculation doesn't respect any provided normalization function, so we'd like to calculate all fill-related values (offset + length) by a given normalization function of model of active handle 

### Mockups or screenshots

![Screenshot 2024-06-12 at 18 05 40](https://github.com/adobe/spectrum-web-components/assets/29090947/9f36c876-300d-4c5a-808b-c7d4abebc6a3)

najikahalsema:
fixed via #4573

PR #4560, opened 2024-06-12
feat: Silder: adjust fillStart calculation for value=0 and normalization function

spdev3000:
This PR adds a way to use fill-start="0" as reference point as well as respecting a provided normalization function to correctly calculate fill offset and length.

## Description

First check, if we really just have the attribute `fill-start` provided with no number value, if so, use centered point of track like before.
For all other number values, use normalization function of model of active handle to calculate real fill-offset and fill-length.

## Related issue(s)

#4559 

## Motivation and context

2 current problems:
1. Currently fill-start attribute only allows to set values !== 0, but sometimes clients have sliders with min<0 , so for example: min="-100", max="200" - so we need a possibility to define fill-start="0" to not start from centered track position
2. Also currently the fill-start calculation doesn't respect any provided normalization function, so we'd like to calculate all fill-related values (offset + length) by a given normalization function of model of active handle

## How has this been tested?

2 new test cases have been added into slider.stories.ts as well as into slider.test.ts 

## Types of changes

## Checklist

## Best practices


packages/slider/src/Slider.ts
```diff
@@ -545,12 +546,13 @@ export class Slider extends SizedMixin(ObserveSlotText(SliderHandle, ''), {
     protected override willUpdate(changed: PropertyValues): void {
         if (changed.has('value') && changed.has('fillStart')) {
             this._cachedValue = Number(this.value);
-            if (this.fillStart) {
-                this.centerPoint = Number(this.fillStart);
-            } else {
+            // Test if fill-start is set without a value
+            if (this.getAttribute('fill-start') === '') {
```
spdev3000, line 550: TBH I don't see any other option to check, if `fill-start` is provided, but without any value. As soon as I check on the property `fillStart` it converts both empty string and `0` to the number `0`, where I cannot decide, was it empty or really the value `0`.
Any ideas to improve?

Rajdeepc: How about you check for the type of `fillStart`? It would have a value if it is not type of a `boolean`

spdev3000: `typeof this.fillStart` always returns `number` and not `boolean` as lit maps that attribute always to a number property.

Rajdeepc:
@spdev3000 Can you also add some documentation around it! That would be very helpful to let the consumers know about the flexibility to use the min-max ranges from -ve minimum values.

spdev3000:
will be continued as #4573

PR #4573, opened 2024-06-20
feat: Silder: adjust fillStart calculation for value=0 and normalization function

Rajdeepc:
This PR adds a way to use fill-start="0" as reference point as well as respecting a provided normalization function to correctly calculate fill offset and length.

## Description

First check, if we really just have the attribute `fill-start` provided with no number value, if so, use centered point of track like before.
For all other number values, use normalization function of model of active handle to calculate real fill-offset and fill-length.

## Related issue(s)

#4559 

## Motivation and context

2 current problems:
1. Currently fill-start attribute only allows to set values !== 0, but sometimes clients have sliders with min<0 , so for example: min="-100", max="200" - so we need a possibility to define fill-start="0" to not start from centered track position
2. Also currently the fill-start calculation doesn't respect any provided normalization function, so we'd like to calculate all fill-related values (offset + length) by a given normalization function of model of active handle

## How has this been tested?

-   [x] Storybook
    Try [here](https://fix-slider-negative-range--spectrum-web-components.netlify.app/storybook/?path=/story/slider--fill-start-with-negative-min-range)

## Types of changes

## Checklist

## Best practices


spdev3000 (dismissed):
lgtm

spdev3000 (approved):

PR #4583, opened 2024-07-02
chore: surface changelogs to the doc site

TarunAdobe:
## Description

This PR surfaces the changelogs for every component to the doc site so that consumers can check what breaking changes or features were added in which release and plan their updates in a better way.

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_
    1. Go here
    2. Do this

image

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/projects/documentation/scripts/copy-component-docs.js
+++ b/projects/documentation/scripts/copy-component-docs.js
@@ -299,8 +299,8 @@ async function main() {
     for await (const mdPath of globby.stream(
         `${projectDir}/(packages|tools)/**/*.md`
     )) {
-        await processREADME(mdPath);
-        await processChangelog(mdPath);
+        processREADME(mdPath);
+        processChangelog(mdPath);
     }
 }
 
--- a/projects/documentation/scripts/copy-component-docs.js
+++ b/projects/documentation/scripts/copy-component-docs.js
@@ -88,11 +88,18 @@ export async function processChangelog(mdPath) {
     );
 
     // Replace major version headings: from single # to ##
+    // Updated to handle optional link
     changelogContent = changelogContent.replace(
-        /^# \[(\d+\.\d+\.\d+)\]\((.*?)\) \((.*?)\)/gm,
-        '## [$1]($2) ($3)'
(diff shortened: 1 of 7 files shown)


Rajdeepc:
Great Start! Thanks for doing this!
Found few formatting issues though which seems to be trivial but let's fix this up!
Screenshot 2024-07-08 at 3 02 58 PM

Also let's make sure we communicate the users about no changelogs as seen in `sp-accordion-item`

Screenshot 2024-07-08 at 3 02 45 PM

projects/documentation/scripts/copy-component-docs.js
```diff
@@ -233,6 +300,7 @@ async function main() {
         `${projectDir}/(packages|tools)/**/*.md`
     )) {
         await processREADME(mdPath);
+        await processChangelog(mdPath);
```
Rajdeepc, line 303: Why this needs to be synchronous? Do we need to wait for processREADME to complete?

Rajdeepc (changes requested):

projects/documentation/src/router.ts
```diff
@@ -18,16 +18,25 @@ tabs.addEventListener('change', (event: Event) => {
     const { selected } = target;
     const { pathname } = location;
     const isAPI = pathname.search('api') > -1;
+    const isChangelog = pathname.search('changelog') > -1;
     switch (selected) {
         case 'api': {
             if (isAPI) return;
-            const dest = (pathname + '/api/').replace('//a', '/a');
+            const dest =
+                pathname.replace(/\/api\/?|\/changelog\/?/, '/') + 'api/';
```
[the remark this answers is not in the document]
TarunAdobe: yessss sure.

projects/documentation/scripts/copy-component-docs.js
TarunAdobe: no we don't. i've made it synchronous now.

Rajdeepc (approved):
Thanks for bringing this in!

Issue #4590, opened 2024-07-08
[Bug]: Initial Focus Issue with Tabs Component in Spectrum Web Components


jianliao:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Tabs

### Expected behavior

**Preconditions:**
1. Tabs component has an initial selected tab set.
2. TabsOverflow is enabled to allow horizontal scrolling when tabs exceed the available width.

**Expected Behavior:**
1. The initial selected tab should be highlighted as selected.
2. Horizontal scrolling arrow buttons should be displayed correctly if tabs exceed the width.
3. Tabs component should not auto-focus

### Actual behavior

1. All UI elements render correctly.
2. Upon initialization, the Tabs component automatically grabs the current focus.
3. If the Tabs component is placed at the bottom of a long page, the page scrolls to the bottom on load, focusing on the Tabs component.

### Screenshots

https://spectrum.corp.adobe.com/page/action-bar/

https://github.com/adobe/spectrum-web-components/assets/1207520/5c81086e-2197-4b65-9c4d-45d8a2d8086f

### What browsers are you seeing the problem in?

Chrome

### How can we reproduce this issue?

https://studio.webcomponents.dev/edit/Sw2YyLG4GfgevvzkYbm8/src/index.ts?p=stories

### Sample code that illustrates the problem

```html
<div class="container" style="max-width: 250px">
    <sp-tabs-overflow>
      <sp-tabs selected="1" size="m">
        <sp-tab label="Tab 1" value="1"></sp-tab>
        <sp-tab label="Tab 2" value="2"></sp-tab>
        <sp-tab label="Tab 3" value="3"></sp-tab>
        <sp-tab label="Tab 4" value="4"></sp-tab>
        <sp-tab label="Tab 5" value="5"></sp-tab>
        <sp-tab label="Tab 6" value="6"></sp-tab>
        <sp-tab label="Tab 7" value="7"></sp-tab>
        <sp-tab label="Tab 8" value="8"></sp-tab>
        <sp-tab-panel value="1">Content for Tab 1</sp-tab-panel>
        <sp-tab-panel value="2">Content for Tab 2</sp-tab-panel>
        <sp-tab-panel value="3">Content for Tab 3</sp-tab-panel>
        <sp-tab-panel value="4">Content for Tab 4</sp-tab-panel>
        <sp-tab-panel value="5">Content for Tab 5</sp-tab-panel>
[...]

PR #4592, opened 2024-07-09
fix(tabs): prevent vertical auto scroll

Rocss:
## Description

Possible fix for https://github.com/adobe/spectrum-web-components/issues/4590

The problem with `scrollIntoView` is that we can not specify the direction of scroll, therefore it scrolls vertically as well, and the function is called all the time, even if you don't actually need to scroll.

This implementation provides an alternative, with validations before scrolling, but unfortunately a lot of checks needed (RTL / LTR, take into account the space between the Tabs and the arrows).

## Related issue(s)

- https://github.com/adobe/spectrum-web-components/issues/4590

## Motivation and context

Replaces `scrollIntoView` with calculations based of scroll offsets.
A follow-up of https://github.com/adobe/spectrum-web-components/pull/4032.

## How has this been tested?

-   [ ] _Test case 1_ All VRTs should pass, as we want to keep this change transparent for the consumers.
-   [ ] _Test case 2_
    1. Go [here](https://rocss-4590--spectrum-web-components.netlify.app/storybook/?path=/story/tabs-overflow--autoscroll)
    2. Using the `selected` control, enter the number of a tab which is not visible
    3. The component should automatically scroll to it
-   [ ] _Test case3_
    1. Go [here](https://rocss-4590--spectrum-web-components.netlify.app/storybook/?path=/story/tabs-overflow--autoscroll)
    2. Using the inspector add to the `.container` class a `margin-top of 1000px`, or whatever value is needed so that the tabs are no longer into view
    3. Using the `selected` control, change the `selected` value, you can choose a number from 1 to 20
    4. The page does not **vertically** scroll
-   [ ] _Test case 4_
    1. Go [here](https://rocss-4590--spectrum-web-components.netlify.app/storybook/?path=/story/tabs-overflow--autoscroll)
    2. Change from LTR to RTL
    3. Using the `selected` control, enter the number of a tab which is not visible
    3. The component should automatically scroll to it

## Types of changes

## Checklist

## Best practices


packages/tabs/src/Tabs.ts
```diff
@@ -263,7 +302,7 @@ export class Tabs extends SizedMixin(Focusable, { noDefaultSize: true }) {
         super.updated(changedProperties);
 
         if (changedProperties.has('selected')) {
-            this.scrollToSelection();
+            this.maybeScrollToSelection();
```
Rajdeepc, line 305: I like the way you added humour to the names but can we change this to something like checkScrollToSelection() instead?

packages/tabs/src/Tabs.ts
```diff
@@ -245,16 +245,55 @@ export class Tabs extends SizedMixin(Focusable, { noDefaultSize: true }) {
...
+
+        return scrollTarget;
+    }
+
+    public async maybeScrollToSelection(): Promise<void> {
         if (!this.enableTabsScroll || !this.selected) {
             return;
         }
 
         await this.updateComplete;
```
Rajdeepc, line 282: Why this is required here?

packages/tabs/src/Tabs.ts
```diff
@@ -245,16 +245,55 @@ export class Tabs extends SizedMixin(Focusable, { noDefaultSize: true }) {
...
+
+        // If the tab is already visible, the necessary scroll factor is 0.
+        let scrollTarget = 0;
+
+        if (selectionEnd > viewportEnd) {
+            // Selection is on the right side, not visible.
+            const nextTab = this.tabs[index + (this.dir === 'rtl' ? -1 : 1)];
+            scrollTarget = nextTab
+                ? nextTab.offsetLeft - this.tabList.offsetWidth
+                : viewportEnd;
```
Rajdeepc, line 263: We can modularize this functions for better testing approach!

packages/tabs/src/Tabs.ts
```diff
@@ -245,16 +245,55 @@ export class Tabs extends SizedMixin(Focusable, { noDefaultSize: true }) {
...
+        } else if (selectionStart < viewportStart) {
+            // Selection is on the left side, not visible.
+            const prevTab = this.tabs[index + (this.dir === 'rtl' ? 1 : -1)];
+            const leftmostElement =
+                this.dir === 'rtl' ? -this.tabList.offsetWidth : 0;
+            scrollTarget = prevTab
+                ? prevTab.offsetLeft + prevTab.offsetWidth
+                : leftmostElement;
```
Rajdeepc, line 271: Same here!

Rajdeepc (changes requested):
Good Approach! Can you also add some storybook tests here so that we can surface it up to the VRTs

packages/tabs/src/Tabs.ts
Rocss: Outside of this class you mean?

Rajdeepc: Yes!

packages/tabs/src/Tabs.ts
Rocss: I reverted it to the initial name if that's okay, I think it would be breaking if I change the name of public methods?

packages/tabs/src/Tabs.ts
Rocss: I added some modifications here, found a more straight forward approach to account for tabs margins, let me know if you think it is better.

packages/tabs/src/Tabs.ts
Rocss: here I am waiting for `this.tabs` to be populated, i observed that unit tests failed without it because of this array being empty :( open to any other suggestion

Rajdeepc: If we execute the next set of instructions only if `this.tabs` is truthy, will that work for you! This await will be flaky!

Rocss: Unfortunately no, the feature will not work at all without this await

packages/tabs/src/Tabs.ts
Rocss: Would you suggest where woult be a good place for it?

Rajdeepc: You can create a helper method here also! We will think where can we best put this later on!

blunteshwar:
Can you update this to main and also some VRTs are failing.

Rajdeepc:
Closing this due to commit override, merging [here](https://github.com/adobe/spectrum-web-components/pull/4613)

Issue #3232, opened 2023-05-23
[Feat]: Dropzone filled+dragged state


mlogsdon18:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Dropzone

### Description of the requested feature

There is a new state of the dropzone component for when it is filled and dragged. The design for this can be seen in Spectrum CSS docs and I've attached the design mock-up. 

### Mockups or screenshots

![Screenshot 2023-05-23 at 3 21 17 PM](https://github.com/adobe/spectrum-web-components/assets/99203545/2e12b0b6-f2eb-41f6-9738-3523e2aaaa0c)

### Implementation notes or ideas

There is already a "dragged" state so that state would just need to be updated whenever the component is also filled, meaning some type of attribute to identify the component as filled would need to be added as well.

najikahalsema:
completed via https://github.com/adobe/spectrum-web-components/pull/4617

PR #4617, opened 2024-07-23
feat: add filled state to dropzone component

rubencarvalho:
## Description

There is a new design for the dropzone component when it is in a "filled state". This PR adds this functionality to the component and respective documentation.

## Related issue(s)

- https://github.com/adobe/spectrum-web-components/issues/3232

## Motivation and context

Catching up the implementation with the most up to date designs.
 
## How has this been tested?

-   [x] _Check the new `filled` state_
    1. Go to [Storybook](https://feat-filled-dropzone--spectrum-web-components.netlify.app/storybook/?path=/story/dropzone--filled)
    2. Toggle the `isFilled` control.
-   [ ] _Added integration test for the `filled` attribute_

## Screenshots

Screenshot 2024-07-25 at 14 28 24

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/dropzone/README.md
+++ b/packages/dropzone/README.md
@@ -30,7 +30,7 @@ import { Dropzone } from '@spectrum-web-components/dropzone';
 
 ```html
 <sp-dropzone id="dropzone-1" style="width: 400px;">
-    <sp-illustrated-message heading="Drag and Drop Your File">
+    <sp-illustrated-message heading="Drag and drop your file">
         <svg
             xmlns="http://www.w3.org/2000/svg"
             viewBox="0 0 150 103"
@@ -70,7 +70,7 @@ When a file is dragged over the `<sp-dropzone>` element, it will display with th
 
 ```html
 <sp-dropzone id="dropzone" dragged style="width: 400px;">
-    <sp-illustrated-message heading="Drag and Drop Your File">
+    <sp-illustrated-message heading="Drag and drop your file">
         <svg
             xmlns="http://www.w3.org/2000/svg"
             viewBox="0 0 150 103"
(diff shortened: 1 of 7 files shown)


Rajdeepc:
@rubencarvalho Can you update the hash and pull this up to `main` so that I can merge this for this release?

packages/dropzone/README.md
```diff
@@ -104,6 +104,22 @@ When a file is dragged over the `<sp-dropzone>` element, it will display with th
...
+
+```html
+<style>
+    sp-action-button {
+        margin-block-end: 32px;
+    }
+</style>
+<sp-action-button draggable="true">Drag me!</sp-action-button>
+<sp-dropzone id="dropzone" filled>
+    <div>Drop file here to replace</div>
```
Rajdeepc, line 119: Does this text change to surface up the notification that 'your file has been dropped'?

packages/dropzone/stories/dropzone.stories.ts
```diff
@@ -106,6 +120,23 @@ Dragged.args = {
...
+export const Filled = (args: StoryArgs): TemplateResult => {
+    return html`
+        <style>
+            sp-action-button {
+                margin-block-end: 32px;
+            }
+        </style>
+        <sp-action-button draggable="true">Drag me!</sp-action-button>
+        <sp-dropzone id="dropzone" ?filled=${args.isFilled}>
+            <div>Drop file here to replace</div>
```
Rajdeepc, line 132: Also here is this sufficient to let the user know that the drop is succesful?

Rajdeepc (changes requested):

packages/dropzone/stories/dropzone.stories.ts
```diff
@@ -106,24 +120,50 @@ Dragged.args = {
...
+    return html`
+        <sp-dropzone id="dropzone" ?filled=${args.isFilled}>
+            Filled dropzone
+        </sp-dropzone>
+    `;
+};
+Filled.args = {
+    isFilled: true,
+};
+
```
TarunAdobe, line 133: Probably nitpicking but can we make sure that all our stores look same? i.e., like the other stories of dropzone i'd like it to follow the same design with illustrated message and whatever else. 
Also on the side since we try to (or atleast in theory we do) use the same template for all our stories to show different component states, it would be better if we can reuse the same template without rewriting the same code in stories. Not a blocker. Just wanted to get your opinion on making our codebase better.

TarunAdobe (changes requested):

TarunAdobe:
Also @rubencarvalho you probably need to rebase this to main and update the golden image hash once again so that ci stops getting angry. Rest everything looks great. Thanks for working on it.

packages/dropzone/stories/dropzone.stories.ts
rubencarvalho: That is a great point, and I think we should really consider a holistic approach to this (maybe outside the scope of this PR). But let me elaborate: since we use Storybook for both component demos and VRT, we need to balance showcasing component functionality with isolating each story's features to test specific functionalities. For example, if we change the `sp-action-button` or the `sp-illustrated-message`, it will cause `sp-dropzone` screenshots to fail. Is this acceptable? Maybe 🤷  However, ideally, we should have an additional layer of component demos that are independent of VRT so Storybook could focus solely on component testing in isolation.
I will follow-up on this.

packages/action-menu/src/ActionMenu.ts
```diff
@@ -113,8 +113,6 @@ export class ActionMenu extends ObserveSlotPresence(
                 class="button"
                 size=${this.size}
                 @blur=${this.handleButtonBlur}
-                @click=${this.handleActivate}
-                @pointerdown=${this.handleButtonPointerdown}
```
blunteshwar, line 117: Why did we remove these ?

rubencarvalho: Bad rebase - should be good now! :)

packages/dropzone/stories/dropzone.stories.ts
Rajdeepc: This is a great discussion thread! We should definitely talk more on this. Currently the storybook config are tightly coupled to docs site demos and we are tightly coupled here! We are trying to create some obfuscated demos which will cater for Desktop and Mobile versions to be represented in docs site and are independent of storybook. 
[This](https://m3.material.io/components/dialogs/guidelines#7e98ae4b-9b0f-4fc0-93c8-e811efffb952) looks like a good and informative way to represent component behaviours:

Rajdeepc (approved):

PR #4627, opened 2024-07-29
test: rectify failing tests

blunteshwar:
## Description

Some tests were failing on main branch on local. This PR will aim at rectifying those failing tests.

## Types of changes

## Checklist

## Best practices


tools/grid/test/grid.test.ts
```diff
@@ -51,7 +51,8 @@ describe('Grid', () => {
 
         expect(el.tabIndex).to.equal(0);
 
-        el.focus();
+        await sendKeys({ press: 'Tab' });
+        await sendKeys({ press: 'Tab' });
```
rubencarvalho, line 56: Can we have an assertion or comment here to understand what these Tabs are doing?

tools/grid/test/grid.test.ts
```diff
@@ -77,7 +78,8 @@ describe('Grid', () => {
...
```
rubencarvalho, line 83: Same here 😄

rubencarvalho (changes requested):
Thank you for fixing these :)

tools/grid/test/grid.test.ts
blunteshwar: Instead of calling focus() function, we are focusing using Tab key

rubencarvalho: can we add an assertion to check if the element is indeed focused?

blunteshwar: The code after line 56 is checking for exactly the same!

rubencarvalho (approved):

PR #4646, opened 2024-08-05
chore: downgrade lightning css to keep fallback css properties

Rajdeepc:
## Description
Downgrade lightning-css to v19.x.x to support keeping fallback properties.
In the latest version minification is removing/tree shaking the fallback css properties from the bundle.

## Related issue(s)

-

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_

    1. Go here
    2. Do this

-   [ ] Did it pass in iPad?

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/package.json
+++ b/package.json
@@ -167,7 +167,7 @@
         "husky": "^9.0.10",
         "latest-version": "^9.0.0",
         "lerna": "^8.1.2",
-        "lightningcss": "^1.24.0",
+        "lightningcss": "^1.19.0",
         "lit": "^2.5.0 || ^3.1.3",
         "lit-analyzer": "^2.0.3",
         "lit-html": "^2.4.0 || ^3.1.3",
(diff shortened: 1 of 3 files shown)


TarunAdobe:
why do we need this in main right now?

Rajdeepc:
> Details

There is a bug raised by the Expess team where `v1.24.x` is trimming fallback properties in `modal.css`

package.json
```diff
@@ -167,7 +167,7 @@
         "husky": "^9.0.10",
         "latest-version": "^9.0.0",
         "lerna": "^8.1.2",
-        "lightningcss": "^1.24.0",
+        "lightningcss": "^1.19.0",
```
rubencarvalho, line 170: If we want to pin the version, we should remove the `^`.

Rajdeepc: its not needed the bug is above this minor version!

yarn.lock
rubencarvalho, line 1: After pinning the version, could we regenerate this lockfile?

package.json
rubencarvalho: But `^1.19.0` allows for versions up to `2.0.0`. Is this fine?

TarunAdobe:
I love how it's no longer messing up the css property ordering and at the same time I have no idea why it was doing so in the first place. Good work. LGTM!

TarunAdobe (approved):

devongovett:
Hi, I maintain Lightning CSS. Would you please file an issue for whatever you ran into? https://github.com/parcel-bundler/lightningcss/issues ideally you could include a small reproduction on https://lightningcss.dev/playground/index.html

Issue #4647, opened 2024-08-06
[Feat]: Support iPad Mini in sp-tray delivery 


Rajdeepc:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Action menu, Picker, Menu

### Description of the requested feature

Currently in iPad Mini 6, in landscape mode the `sp-action-menu /sp-picker/ menu` appears to be rendered from the bottom like a tray, instead of opening as a `sp-menu`. 
It doesn't happen on  IPad Pro or IPad 10 (Can only reproduced it on physical devices or in the Safari IOS Simulator). 

### Mockups or screenshots

![image](https://github.com/user-attachments/assets/f85ae0a2-39df-46c3-bc78-0059259460a1)
![image](https://github.com/user-attachments/assets/c7acdd99-2b4b-4039-928e-110a9893f723)

### Implementation notes or ideas

Bug in SWC due to fixed height settings in  https://github.com/adobe/spectrum-web-components/blob/eb86aa131095470d7d1c5863e183b16883bb22a2/tools/reactive-controllers/src/MatchMedia.ts#L16

Possible solution:
Horizon approach:
https://git.corp.adobe.com/Horizon/hz/blob/main/platform/common/browser-detection-utils/src/ScreenBreakpoint.ts#L29 (edited)

PR #4656, opened 2024-08-07
fix(reactive-controllers): update IS_MOBILE condition for MatchMediaController

TarunAdobe:
## Description

MatchMediaController would now `window.screen.availHeight` and `window.screen.availWidth` in conjunction with`window.matchMedia` to detect a mobile device accurately. 

## Related issue(s)

#4647 

## Motivation and context

Inconsistency in Picker, Action-Menu opening as `popover` vs `tray` in iPad devices after changing orientation.

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_

    1. Go here
    2. Do this

-   [ ] Did it pass in iPad?

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/tools/reactive-controllers/src/MatchMedia.ts
+++ b/tools/reactive-controllers/src/MatchMedia.ts
@@ -13,7 +13,7 @@ import type { ReactiveController, ReactiveElement } from 'lit';
 
 export const DARK_MODE = '(prefers-color-scheme: dark)';
 export const IS_MOBILE =
-    '(max-width: 700px) and (hover: none) and (pointer: coarse), (max-height: 700px) and (hover: none) and (pointer: coarse)';
+    '(max-width: 743px) and (hover: none) and (pointer: coarse)';
 
 export class MatchMediaController implements ReactiveController {
     key = Symbol('match-media-key');


rubencarvalho (changes requested):
Let us dive deeper into avoiding attaching an event listener to each component, i.e., let's explore the singleton approach.

tools/reactive-controllers/src/MatchMedia.ts
```diff
@@ -27,23 +27,45 @@ export class MatchMediaController implements ReactiveController {
...
+        if (this.mediaMatches === event.matches) return;
+        this.mediaMatches = event.matches;
+        this.host.requestUpdate(this.key, !this.mediaMatches);
+    }
+
+    protected updateMatches(): void {
+        // check for window.screen.availWidth and window.screen.availHeight to confirm that the device is a mobile device
+        const newMatches =
+            this.mediaMatches &&
+            (window.screen.availWidth < 768 || window.screen.availHeight < 768);
```
lazd, line 65: Very interesting! Wondering how this works around issues with `mediaMatches`?

lazd (commented):
Agreed that this needs to be done as a singleton, if this controller is on every element, we don't want thousands of listeners.

tools/reactive-controllers/src/MatchMedia.ts
```diff
@@ -15,35 +15,153 @@ export const DARK_MODE = '(prefers-color-scheme: dark)';
...
-    protected media: MediaQueryList;
+    protected isMobileScreenQuery!: MediaQueryList;
+    protected smallScreenQuery!: MediaQueryList;
+    protected largeScreenQuery!: MediaQueryList;
+    protected orientationQuery!: MediaQueryList;
+
+    protected mediaQueryList!: MediaQueryList;
 
-    constructor(host: ReactiveElement, query: string) {
+    protected screenSize: 'small' | 'large' = 'small';
```
rubencarvalho, line 37: Could we create some `enum`s for `screenSize` and `orientation`?

tools/reactive-controllers/src/MatchMedia.ts
```diff
@@ -15,35 +15,153 @@ export const DARK_MODE = '(prefers-color-scheme: dark)';
...
-        this.matches = this.media.matches;
-        this.onChange = this.onChange.bind(this);
+
+        // if the query is provided then we use it to determine the matches
+        if (query) {
+            this.mediaQueryList = window.matchMedia(query);
+            this.matches = this.mediaQueryList.matches;
+            this.mediaQueryList.addEventListener(
+                'change',
+                this.onChange.bind(this)
```
rubencarvalho, line 50: nit: I believe our code style typically uses arrow functions for binding `this`

tools/reactive-controllers/src/MatchMedia.ts
```diff
@@ -15,35 +15,153 @@ export const DARK_MODE = '(prefers-color-scheme: dark)';
...
 
     protected host: ReactiveElement;
 
-    protected media: MediaQueryList;
+    protected isMobileScreenQuery!: MediaQueryList;
+    protected smallScreenQuery!: MediaQueryList;
+    protected largeScreenQuery!: MediaQueryList;
+    protected orientationQuery!: MediaQueryList;
+
+    protected mediaQueryList!: MediaQueryList;
```
rubencarvalho, line 35: nit: could we have a more descriptive name for this one? The name `mediaQueryList` confuses me a bit when reading through the code. Maybe something like `customMediaQueryList` would be clearer that this is used when a custom media query is passed?

tools/reactive-controllers/src/MatchMedia.ts
```diff
@@ -15,35 +15,153 @@ export const DARK_MODE = '(prefers-color-scheme: dark)';
 export const IS_MOBILE =
     '(max-width: 700px) and (hover: none) and (pointer: coarse), (max-height: 700px) and (hover: none) and (pointer: coarse)';
 
+export const MOBILE_SCREEN_QUERY = '(hover: none) and (pointer: coarse)';
+export const SMALL_SCREEN_QUERY = '(max-width: 743px)';
+export const LARGE_SCREEN_QUERY = '(min-width: 744px)'; // iPad Mini 6 max dimension, want it to be a 'tablet' using Desktop UI
+export const ORIENTATION_QUERY = '(orientation:portrait)';
+
 export class MatchMediaController implements ReactiveController {
     key = Symbol('match-media-key');
```
rubencarvalho, line 24: What is this `key` used for?

tools/reactive-controllers/src/MatchMedia.ts
```diff
@@ -15,35 +15,153 @@ export const DARK_MODE = '(prefers-color-scheme: dark)';
...
-        this.host.requestUpdate(this.key, !this.matches);
+    protected onChange(): void {
+        // if the query is provided
+        if (this.mediaQueryList) {
+            this.matches = this.mediaQueryList.matches;
+            this.host.requestUpdate();
+            return;
+        }
+
+        const screenSize = this._determineScreenSize();
```
rubencarvalho, line 151: Could we encapsulate this "check if is mobile" logic into a dedicated method?

```ts
private _checkIsMobile(): boolean {
  const screenSize = this._determineScreenSize();
  const orientation = this._determineScreenOrientation();
  const isMobileScreen = this.isMobileScreenQuery.matches;

  // if the screen size is small or the orientation is landscape and the device is mobile
  return isMobileScreen && (screenSize === screenSize.small || orientation === orientation.landscape);
}
```

and then

```ts
this.matches = _checkIsMobile()
```

Rajdeepc (approved):
This is good to unblock for now! Let's think through a context level approach in achieving this in a follow up PR!

PR #4667, opened 2024-08-12
feat(calendar): keyboard navigation

mizgaionutalexandru:
## Description

Mainly this PR adds the keyboard navigation functionality to the Calendar's grid of days. As per [WAI keyboard support](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/datepicker-dialog/#kbd_label) for date picker dialogs, the selected day is the only one tabbable from the grid. If there is no day selected the fallback will be the current day and the first day of the displayed month. After the said day is in focus, the following keys and functions apply.

|Key| Action |
| :---:  | :---  |
| Space/ Enter|Selects the focused date|
|Up/ Down Arrow|Moves the focus to the same day of the previous/next available week. An "available week" is a week that has at least one day available. If the said date is disabled, the focus is moved to the closest available day with respect to the navigation direction (a minimum focus movement difference).|
|Left/ Right Arrow|Moves the focus to the previous/next available day. It can go from one week to another if the focus is currently on the first available day of the week|

If the user changes the focus out from the grid and back in, his previous location should be the one tabbable.

Also, this PR addresses some code suggestions/improvements such as:
- removing the unnecessary _"spectrum-Calendar-*"_ from class names
- small performance gains when parsing date table cell properties such as "isToday" or "isSelected"
- passing a single method for every date cell for click events (see previous [comment](https://github.com/adobe/spectrum-web-components/pull/3327#discussion_r1274260137))
- don't dispatch change event when the user selects the previously selected date
- allow clicks only in the date cell radius shown by the styles for consistent UX

While creating tests for this new functionality I also overwritten old tests. Some tests have a _TODO_ comment and will be addressed in the next PR to account for the new value API of the components.

## Related issue(s)

## Motivation and context

Not having keyboard navigation the Calendar wouldn't be keyboard accessible.

## How has this been tested?
There can be many use-cases and hopefully most of them are covered in unit tests. As per the behaviour explained above, you can apply said key functions by using the component's storybook.

-   [ ] _Test case_
    1. Go to the [Calendar's storybook](https://imizga-calendar-keyboard--spectrum-web-components.netlify.app/storybook/?path=/story/calendar--default)
    2. Click somewhere on the storybook's iframe (where the component is displayed)
    3. Press _Tab_ to navigate through the Previous/Next month buttons as well as the selected/initial keyboard navigable day
    4. Use your arrow keys (_up/down/left/right_) to move through the calendar's days
    5. Use _Space/Enter_ to select the current day
    6. Play around with the mentioned keys, including _Tab_

-   [ ] Did it pass in iPad?

https://github.com/user-attachments/assets/dc998e94-515e-4d1e-b843-c1256d60ece4

## Types of changes

## Checklist

## Best practices


packages/calendar/test/calendar.test.ts
```diff
@@ -10,23 +10,67 @@ OF ANY KIND, either express or implied. See the License for the specific languag
 governing permissions and limitations under the License.
 */
 import { elementUpdated, expect, fixture, html } from '@open-wc/testing';
-import sinon, { spy } from 'sinon';
-
+import { sendKeys, sendMouse } from '@web/test-runner-commands';
 import { testForLitDevWarnings } from '../../../test/testing-helpers.js';
-import { Calendar } from '@spectrum-web-components/calendar';
-
+import { spy } from 'sinon';
```
mizgaionutalexandru, line 15: I found this neccessary for checking the component's change event. In SWC it seems like sinon is used a lot, even in those testing-helpers, but here I need just the spy functionality.

Related to: https://github.com/adobe/spectrum-web-components/pull/3856#discussion_r1705118930

packages/calendar/test/calendar.test.ts
```diff
@@ -10,23 +10,67 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+        Date.now = () => fixedTime;
+    });
+
+    beforeEach(async () => {
+        element = await fixtureElement();
+        await elementUpdated(element);
+    });
+
+    after(() => {
+        Date.now = originalDateNow;
```
mizgaionutalexandru, line 72: If sinon is used in these tests, should we also use a sandbox to stub the `Date.now` function?

packages/calendar/stories/calendar.stories.ts
```diff
@@ -62,114 +98,41 @@ const renderCalendar = (
             `.story-container-${randomId}`
         );
 
-        if (container) {
-            render(story, container as HTMLElement);
-        }
+        if (container) render(story, container as HTMLElement);
```
mizgaionutalexandru, line 101: Can this be done in a different manner or should we stick to `requestAnimationFrame`?
It seems like without it the component renders only the header (month navigation buttons and the current month indicator) and not the grid of days. Does this hide a problem in the component's lifecycle? 🤔

Rocss:
I found the following edge case while manual testing this feature:

1. Go to https://imizga-calendar-keyboard--spectrum-web-components.netlify.app/storybook/?path=/story/calendar--min-and-max-dates
2. Using the keyboard, go ahead and select a different date besides today, for example, tomorrow
3. Now using the arrows in the header, go to next month
4. Return
5. Try to select a day again => No longer working, calendar days are no longer into tab order.

packages/calendar/stories/calendar.stories.ts
mizgaionutalexandru: It seems like the issue comes from the interaction of `sp-theme` and `LanguageResolutionController`. Explicitly that the controller's `hostConnected` is called before the `sp-theme`'s constructor, causing the `sp-language-context` event to be dispatched before the event listener for said event is added in `Theme.ts`. 
This looks like a hack-ish workaround for a more general problem that I recreated outside of the storybook as well: https://studio.webcomponents.dev/edit/29O0MXJBQtaMF3CgjhhW/src/index.ts?p=stories. Here you can see that even if the `sp-theme`'s `lang` attribute is changed, the language of the `number-field` coming from the controller didn't change.
More investigation is needed here because I don't yet know how this interaction can be fixed. I would be happy to get any help or suggestions.

mizgaionutalexandru:
> I found the following edge case while manual testing this feature:

Thank you for checking this out! I think the issue is fixed now, could you double check please?

packages/calendar/stories/calendar.stories.ts
mizgaionutalexandru: Two possible solutions that I've found are:
1. Set `bundled` to `false` in the `swcThemeDecoratorWithConfig` ([here](https://github.com/adobe/spectrum-web-components/blob/v0.46.0/projects/story-decorator/decorator.ts#L35)) - I'm not aware of all the implications and where is this variable set other than the default value.
2. Add one more call of `resolveLanguage` in the `LanguageResolutionController` using the `hostUpdate` (making sure to call this only for the first update). For environments other than the storybook this one extra call should not be a problem because `Theme` checks if the `contextConsumer` was already subscribed ([here](https://github.com/adobe/spectrum-web-components/blob/main/tools/theme/src/Theme.ts#L646)). 

Personally I'm leaning more towards the second solution. What do you think about this @Rajdeepc ?

Rajdeepc: Do you need more discussion on this @mizgaionutalexandru . I think adding a `@state()` decorator will reconcile the changes!

Rajdeepc:
I would also like to add the below functionality.
1. Go to https://imizga-calendar-keyboard--spectrum-web-components.netlify.app/storybook/?path=/story/calendar--min-and-max-dates
2. Tab to one date. Then using navigation keys try to go to another date for next/previous month. This is not working.
3. Users should be able to freely move to any date in any month by navigation keys

packages/calendar/stories/calendar.stories.ts
mizgaionutalexandru: Yes, I would love to clarify this because there are 2 different topics. The `state()` decorator works for 'normal app environments' let's say, and that's how the storybook decorator is implemented but due to the fact that in **storybook** the story function gets called before the `sp-theme`'s constructor gets the chance to add the event listener that is needed for the language resolver.
I found that those 2 approaches would fix the issue in the storybook environment.

mizgaionutalexandru:
> 2. Tab to one date. Then using navigation keys try to go to another date for next/previous month. This is not working.
> 3. Users should be able to freely move to any date in any month by navigation keys

Thank you for your input on this one!
Just to be clear, if you're talking about adding a new keyboard shortcut to be able to navigate to the previous month more easily like `Page Up` here, I can see that working although the RFC does not include this.

image

If you're saying that unavailable/disabled days should be navigable, I think the current approach that aligns with React Spectrum is better, this way the constraints the user has makes him pick an available day, as per the RFC. You can check out the React Spectrum approach [here](https://react-spectrum.adobe.com/react-spectrum/Calendar.html#validation).
Also, worth mentioning is that using the DateTimePicker the user will be able to type-in an unavailable day, this PR is only for the Calendar component.

What are your thoughts on this?

packages/calendar/src/Calendar.ts
```diff
@@ -311,46 +383,193 @@ export class Calendar extends SpectrumElement {
...
+            }
+            case 'Space':
+            case 'Enter': {
+                this.handleDaySelect(event);
+                break;
+            }
+        }
+
+        if (this.currentDate.month !== initialMonth)
+            this.setWeeksInCurrentMonth();
```
mizgaionutalexandru, line 529: This call, as well the ones from `handlePreviousMonth` and `handleNextMonth` will be removed from here and added in a special check inside `willUpdate`, I think this makes more sense and it's less error prone.

Rajdeepc (approved):

Issue #4579, opened 2024-06-26
[Bug]: `change` event is not triggered for double byte characters on Safari


Rocss:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-number-field

### Expected behavior

Given that I enter a double-byte number in a `sp-number-field`, I expect that a `change` event is triggered when I press `enter` or on blur with the correct value, which should be the equivalent of what I entered, in single byte characters.
This works well on Chrome, but not on Safari.

### Actual behavior

`change` event is not triggered on `enter` keypress, only the `input` event
`input` and `change` events are not triggered correctly on blur, event.target.value is `NaN`

### Screenshots

https://github.com/adobe/spectrum-web-components/assets/13311865/ed3312f9-fb95-4e2e-b9d7-dbc2e0b8a982

### What browsers are you seeing the problem in?

Safari

### How can we reproduce this issue?

1. Open Safari browser (I used v17.5)
2. Go to https://opensource.adobe.com/spectrum-web-components/storybook/index.html?path=/story/number-field--default
3. In your system settings, select "Japanese" as the IME
4. Type any number in the `sp-number-field` component
5. Observe behaviour on `enter` keypress and blur
6. 

### Sample code that illustrates the problem

The problem seems to be the fact that when entering double-byte characters, `sp-number-field` converts the characters to single byte using a map, and then sets the input value with the new and processed number.
Seems like `input` does not trigger a `change` event in Safari when you set the value of the input programatically.
I could not find a workaround for that.

PR #4672, opened 2024-08-16
fix(number-field): update IME change detection

rubencarvalho:
## Description
This PR resolves an issue in Safari where multi-byte characters were being improperly converted, resulting in `NaN` and updates the function responsible for triggering the `change` event to ensure it behaves consistently across all supported browsers.

## Related issue(s)

- fixes https://github.com/adobe/spectrum-web-components/issues/4579

## Motivation and context

`sp-number-field` + Safari + IME were in need of some love.

## How has this been tested?

-   [ ] Storybook actions
    1. Open the Safari browser
    2. Go to the deployed/locahost `/storybook/index.html?path=/story/number-field--default`
    3. In your system settings, select "Japanese" as the IME
    4. Type any number in the `sp-number-field` component
    5. Observe the `actions` tab on `Enter` keypress and `blur` for the expected behaviour

-   [x] Did it pass in iPad (simulator)?

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/number-field/src/NumberField.ts
+++ b/packages/number-field/src/NumberField.ts
@@ -63,8 +63,17 @@ export const remapMultiByteCharacters: Record<string, string> = {
     '％': '%',
     '＋': '+',
     ー: '-',
+    一: '1',
+    二: '2',
+    三: '3',
+    四: '4',
+    五: '5',
+    六: '6',
+    七: '7',
+    八: '8',
+    九: '9',
+    零: '0',
 };
-
 const chevronIcon: Record<string, (dir: 'Down' | 'Up') => TemplateResult> = {
     s: (dir) => html`
         <sp-icon-chevron50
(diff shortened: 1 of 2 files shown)


rubencarvalho:
https://developer.apple.com/forums/thread/698078

packages/number-field/src/NumberField.ts
```diff
@@ -214,7 +228,13 @@ export class NumberField extends TextfieldBase {
     private valueBeforeFocus: string = '';
     private isIntentDecimal: boolean = false;
 
-    private convertValueToNumber(value: string): number {
+    private convertValueToNumber(inputValue: string): number {
+        // Normalize full-width characters to their ASCII equivalents
```
rubencarvalho, line 232: Without normalizing this value (remapping the multi-byte characters)  before the input manipulation, we would get `NaN`.

packages/number-field/test/number-field.test.ts
```diff
@@ -502,6 +502,18 @@ describe('NumberField', () => {
             el.value = 52;
             expect(changeSpy.callCount).to.equal(0);
         });
+        it('handles IME input correctly and dispatches change event', async () => {
```
rubencarvalho, line 505: This test covers both the `NaN` case and the `change` event.

packages/number-field/src/NumberField.ts
```diff
@@ -175,20 +184,25 @@ export class NumberField extends TextfieldBase {
     private _trackingValue = '';
     private lastCommitedValue?: number;
 
-    private setValue(value: number = this.value): void {
-        this.value = value;
+    private setValue(newValue: number = this.value): void {
+        // Capture previous value for accurate IME change detection
```
rubencarvalho, line 188: For IME input, the value conversion process would cause `this.value` and `this.lastCommitedValue` to be identical at the time of the comparison, even though a change had happened. So I added `previousValue` to hold the value of `this.lastCommitedValue` before the update to `this.value`, which fixed it! :)

TarunAdobe (approved):

Issue #4109, opened 2024-03-01
[Feat]: sp-number-field repopulate field after deletion


cdvu1:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-number-field

### Description of the requested feature

Currently, removing the number in the number field leaves the number field blank. It would be nice if the number field could repopulate to either the previous value, or the default value (ex. 0).

TarunAdobe:
Hello @cdvu1. I discussed the idea of having a `default-value` for the number-field component with the team. We felt that since the `number-field` component is a controllable by the consuming application, having a reset to `default-value`behaviour can be easily created by listening to the `@change` event on the number-field component. I have added an example of the same in our doc site in this [PR](https://github.com/adobe/spectrum-web-components/pull/4698). Let me know if this helps :)

PR #4698, opened 2024-08-27
chore(number-field): added default-value example to the doc site

TarunAdobe:
## Description

Expand `number-field` doc.

## Related issue(s)

- #4109

## Motivation and context

Currently, removing the number in the number field leaves the number field blank. It would be nice if the number field could repopulate to either the previous value, or the default value (ex. 0).

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/number-field/README.md
+++ b/packages/number-field/README.md
@@ -229,3 +229,41 @@ If the user types a value that is between two steps and blurs the input, the val
     step="3"
 ></sp-number-field>
 ```
+
+## Default value
+
+The number-field component doesn't manage a default value by itself. This means that consumers can set the value of the number-field as an empty string by clearing the input. If we want the number-field to reset to a `default-value` when the user clears the input, we can listen for the `change` event on the number-field component and set its value to the desired `default-value` if the input is empty.
+
+```html-live
+<sp-field-label for="default">
+    Default value of this number field is 42
+</sp-field-label>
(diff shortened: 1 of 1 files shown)


blunteshwar (changes requested):
Can you shake off the VRTs so that we are good to go

packages/number-field/README.md
```diff
@@ -229,3 +229,41 @@ If the user types a value that is between two steps and blurs the input, the val
     step="3"
 ></sp-number-field>
 ```
+
+## Default value
+
+The number-field component doesn't manage a default value by itself. This means that consumers can set the value of the number-field as an empty string by clearing the input. If we want the number-field to reset to a `default-value` when the user clears the input, we can listen for the `change` event on the number-field component and set its value to the desired `default-value` if the input is empty.
```
Rajdeepc, line 235:
```suggestion
The `<sp-number-field>` component doesn't manage a default value by itself. This means that consumers can set the value of the number-field as an empty string by clearing the input. If we want the number-field to reset to a `default-value` when the user clears the input, we can listen for the `change` event on the number-field component and set its value to the desired `default-value` if the input is empty.
```

Rajdeepc:
> Can you shake off the VRTs so that we are good to go

VRT's are good. I am adding a fix for the `chromedriver` to update to the latest version to align with the browser version

Rajdeepc (approved):

Issue #4700, opened 2024-08-27
[Bug]: Tabs overflow exceed tabs limit


mizgaionutalexandru:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Tabs Overflow

### Expected behavior

The tabs movements should stop when there are no items/tabs left to show.

### Actual behavior

The user can still tap the chevron icon and a weird UI flickering is happening.

### Screenshots

https://github.com/user-attachments/assets/4f122e8b-91fc-49f0-85c7-19d9957359a9

### What browsers are you seeing the problem in?

Chrome, Safari

### How can we reproduce this issue?

1. Go to [Tabs Overflow Storybook](https://opensource.adobe.com/spectrum-web-components/storybook/index.html?path=/story/tabs-overflow--compact) using an iOS device (iPad, iPhone etc.)
3. Scroll/use the arrow to move to "Tab Item 20"
4. Select it
5. Use the chevron icon to go back to "Tab Item 1"
6. Observe that you can still act upon the icon after the item is reached
7. Click on it and see the weird flickering

### Severity

SEV 3

PR #4722, opened 2024-09-04
fix(tabs): scroll exceeding tabs limit

mizgaionutalexandru:
## Description

This PR aims to address said problem by limitting the delta a consumer can pass to the `scrollTabs` public method, used in `sp-tabs-overflow`.

## Related issue(s)

- https://github.com/adobe/spectrum-web-components/issues/4700

## Motivation and context

Currently on iOS devices the user can still act upon the "left/right" chevron icons to scroll through the tabs even when there are no more tabs to scroll to. Also, an empty space is visible when scrolling to the right/left most tab.

## How has this been tested?

-   [ ] _Test case 1_
    1. Go to the tabs overflow storybook using an iOS device
    2. Click on the "right" icon to scroll to the right until you reach the last tab item
    3. Observe that the scroll stops right at the last item, without any empty spaces and the icon disappears
-   [ ] _Test case 2_
    1. Repeat _test case 1_
    2. Click on the "left" icon to scroll to the left until you reach the first item
    3. Observe that the scroll stops right at the first item, without any empty spaces and the icon disappears
-   [ ] _Test case 3_
    1. Go to the tabs overflow storybook using an iOS device
    2. Set the direction to 'rtl'
    3. Click on the "left" icon to scroll to the left until you reach the last tab item
    4. Observe that the scroll stops right at the last item, without any empty spaces and the icon disappears
-   [ ] _Test case 4_
    1. Repeat _test case 3_
    2. Click on the "right" icon to scroll to the right until you reach the first item
    3. Observe that the scroll stops right at the first item, without any empty spaces and the icon disappears

-   [x] Did it pass in iPad?

https://github.com/user-attachments/assets/eb0cd502-b928-4949-bd48-acec6209a966

## Types of changes

## Checklist

## Best practices


mizgaionutalexandru:
> Great Start! Would you also be interested in creating a test case for this so that we can surface this up and make sure that we get non-breaking changes.

I've been trying to get that working but wasn't able to create a test for it. I tried to focus the `nextButton` and use `sendKeys` to simulate multiple clicks and then checking the scroll position of the `#list` with `scrollLeft` but I got some random numbers that weren't reliable. Do you have any idea on how can i approach this?

Rajdeepc:
> > Great Start! Would you also be interested in creating a test case for this so that we can surface this up and make sure that we get non-breaking changes.
> 
> I've been trying to get that working but wasn't able to create a test for it. I tried to focus the `nextButton` and use `sendKeys` to simulate multiple clicks and then checking the scroll position of the `#list` with `scrollLeft` but I got some random numbers that weren't reliable. Do you have any idea on how can i approach this?
> 
> Also, I'm not aware how I can make a test specific for iOS, where the problem is. Are there any guidelines in SWC for this type of situations?

I am working with the playwright team to work on a solution to enable touch specific renderers which will help us test mobile specific touch events.
Currently it is still not support but you can create mobile viewport emulation as described [here](https://playwright.dev/docs/emulation#devices)

packages/tabs/test/tabs-overflow.test.ts
```diff
@@ -169,6 +179,74 @@ describe('TabsOverflow', () => {
...
+        await elementUpdated(el);
+        await setViewport({ width: 360, height: 640 });
+        await nextFrame();
+
+        const tabsOverflow = el.querySelector(
+            'sp-tabs-overflow'
+        ) as TabsOverflow;
+
+        expect(tabsOverflow['overflowState'].canScrollLeft).to.be.true;
+        expect(tabsOverflow['overflowState'].canScrollRight).to.be.false;
```
mizgaionutalexandru, line 237: Used `overflowState` for assertions due to the fact the `waitUntil` that allows scroll behavior in unit tests needs a specific condition to stop. Haven't found a way to test for a 'bad actor' like behaviour to see if the scroll can exceed the expected limit.
I'm happy to hear any thoughts/alternatives to this

Rajdeepc (changes requested):

packages/tabs/src/TabsOverflow.ts
```diff
@@ -116,11 +116,12 @@ export class TabsOverflow extends SizedMixin(SpectrumElement) {
         }
     }
 
+    private scrollFactor = 0.5;
```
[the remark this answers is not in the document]
mizgaionutalexandru: I just extracted the value to a private property so that it can be used in unit tests. I didn't modify it and I'm not currently aware of the reason behind its exact value.

packages/tabs/src/Tabs.ts
```diff
@@ -235,12 +235,26 @@ export class Tabs extends SizedMixin(Focusable, { noDefaultSize: true }) {
         return this.rovingTabindexController.focusInElement || this;
     }
 
+    private limitDelta(min: number, max: number) {
```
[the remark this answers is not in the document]
mizgaionutalexandru: Thank you for pointing that out. Indeed this method especially seemed somewhat ambiguous. I changed it's implementation so that it does the same thing but in a cleaner manner. What are your thoughts on this?

Rajdeepc:
Please pull this up to `main` so that the VRTs are happy!

packages/tabs/src/Tabs.ts
```diff
@@ -235,12 +235,35 @@ export class Tabs extends SizedMixin(Focusable, { noDefaultSize: true }) {
         return this.rovingTabindexController.focusInElement || this;
     }
 
+    private limitDeltaToInterval(min: number, max: number) {
```
rubencarvalho, line 238: nit: this can be a one liner

`Math.max(min, Math.min(max, delta));`

packages/tabs/test/tabs-overflow.test.ts
```diff
@@ -169,6 +179,74 @@ describe('TabsOverflow', () => {
...
+
+        await scrollToEnd(el, LEFT_BUTTON_SELECTOR, 'ltr');
+
+        expect(tabsOverflow['overflowState'].canScrollLeft).to.be.false;
+        expect(tabsOverflow['overflowState'].canScrollRight).to.be.true;
+    });
+
+    xit('should scroll up to the last item and back in RTL', async () => {
```
rubencarvalho, line 216: Would it work and make sense to let these tests run, even without setting the user agent?

packages/tabs/test/tabs-overflow.test.ts
```diff
@@ -362,3 +440,83 @@ describe('calculateScrollTargetForLeftSide', () => {
...
+    if (iteration > times) return;
+
+    const distanceToReach = distanceToReachInIteration(iteration);
+
+    await sendKeys({ press: 'Enter' });
+    await elementUpdated(elementToUpdate);
+    await waitUntil(
+        () =>
+            Math.ceil(Math.abs(elementToScroll.scrollLeft)) -
+                Math.abs(distanceToReach) ===
```
rubencarvalho, line 468: Is this condition reliable enough? I'm just wondering if there’s any small discrepancy due to floating-point rounding. Alternatively we could check if it's less than a tolerance small enough, e.g. `< 0.5`. 
(but if this is not a problem, feel free to ignore!)

packages/tabs/test/tabs-overflow.test.ts
```diff
@@ -362,3 +440,83 @@ describe('calculateScrollTargetForLeftSide', () => {
...
+        elementToScroll,
+        distanceToReachInIteration,
+    } = options;
+    if (iteration > times) return;
+
+    const distanceToReach = distanceToReachInIteration(iteration);
+
+    await sendKeys({ press: 'Enter' });
+    await elementUpdated(elementToUpdate);
+    await waitUntil(
```
rubencarvalho, line 465: I'm wondering if it would make sense to wrap this in a `try`, `catch` block, so we could log error messages. Otherwise, if the test times out, it might be harder to pinpoint what is breaking.

packages/tabs/src/Tabs.ts
```diff
@@ -235,12 +235,35 @@ export class Tabs extends SizedMixin(Focusable, { noDefaultSize: true }) {
...
+     */
     public scrollTabs(
         delta: number,
         behavior: ScrollBehavior = 'smooth'
     ): void {
+        if (delta === 0) return;
+
+        const { scrollLeft, clientWidth, scrollWidth } = this.tabList;
+        const dirLimit = scrollWidth - clientWidth - Math.abs(scrollLeft);
+
```
rubencarvalho, line 259: Should we round any of these values?

packages/tabs/src/Tabs.ts
mizgaionutalexandru: I did it this way previously but I think this approach is better for readability purposes.

packages/tabs/test/tabs-overflow.test.ts
mizgaionutalexandru: We can run the tests if we remove the `setUserAgent` part, this is how I created them in the first place. I think it would make sense to run these for desktop as well, to cover the scenario.

The thing with this approach is that they wouldn't fail even if this bug is reproducible on iOS and I think is good to signal in some way that these tests must be enabled once we have a way to run them on iPad/iOS/Android. I'm thinking to enable them in the current tests contexts and commenting the `setUserAgent` part, along with a comment like this:
```ts
// TODO: run on iPhone as per https://github.com/adobe/spectrum-web-components/pull/4722
````
What do you think, do you have any suggestions on this?

File: packages/tabs/test/tabs-overflow.test.ts (hunk 6 above)
Comment (mizgaionutalexandru) [reply packages/tabs/test/tabs-overflow.test.ts:468]: I understand your concern. From my testing sessions I found the rounding up with `Math.ceil` to be enough.

File: packages/tabs/test/tabs-overflow.test.ts (hunk 7 above)
Comment (mizgaionutalexandru) [reply packages/tabs/test/tabs-overflow.test.ts:465]: The `scroll to ${distanceToReach}` message has this purpose. If one of these tests fails due to the `waitUntil`, a message similar to the one bellow will appear in the console. I think this is enough to pinpoint the exact point where something breaks.
```bash
❌ TabsOverflow > should scroll up to the last item and back in LTR
      Error: Timeout: scroll to 100
```

packages/tabs/src/Tabs.ts
mizgaionutalexandru: I don't think so. Some rounding happens in the `scrollState` [getter](https://github.com/adobe/spectrum-web-components/blob/6760ec283ad190f45f9639e636953e90ea562385/packages/tabs/src/Tabs.ts#L254). I think that's the right place and I currently don't see the value in duplicating/adding it if it's not necessary. It's best to scroll exactly how the device/screen allows, according to this `dirLimit`.

packages/tabs/test/tabs-overflow.test.ts
rubencarvalho: Nice! I agree that is enough.

packages/tabs/test/tabs-overflow.test.ts
rubencarvalho: Great!

packages/tabs/test/tabs-overflow.test.ts
rubencarvalho: Yes, I think that approach is good. While this bug is iOS-only, this test is running in a mobile viewport which still helps with our testing coverage 👍

rubencarvalho (approved):

Issue #4675, opened 2024-08-20
[Bug]: [Ipad] Blue focus ring appear on the button after opening/closing ActionMenu component.


manassinghadobe:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Action Menu

### Expected behavior

Blue Focus Ring should not appear Action Menu Component on Click

### Actual behavior

After opening and closing action menu, blue focus ring appear on the action button

https://github.com/user-attachments/assets/393b5870-58b0-4c1f-bb3c-6fcc99e5b115

https://github.com/user-attachments/assets/31728198-2ef9-43a8-93d2-bfb2c475a4fa

### How can we reproduce this issue?

3. Scroll to '....'
4. Check console
5. See error

### Sample code or abstract reproduction which illustrates the problem

The issue is reproducible on Spectrum component page of `Action Menu` but not able to repro in [playground](https://studio.webcomponents.dev/edit/xHqNjZPTQnMiC8INZ5fh/src/index.ts?p=stories)

### Severity

None

PR #4724, opened 2024-09-05
fix(picker): added a custom class to make `:focus-visible` styles consistent across all browsers

TarunAdobe:
## Description

Added a custom `remove-focus-ring-safari-hack` class that gets applied when the picker (that is opened using a mouse click) closes after making a selection in a mobile environment using webkit.

This class unsets the focus-ring properties that we apply for the `:focus-visible` pseudo class. 

This is needed so that we don't get focus-ring styles on making a selection in an action-menu in a IOS device.

## Related issue(s)

- #4675

## Motivation and context

Blue Focus Ring should not appear on the action-menu and picker on opening and closing in IOS mobile devices.

## How has this been tested?

-   I opened [storybook](https://fix-blue-ring-action-menu--spectrum-web-components.netlify.app/storybook/?path=/story/action-menu--default) and manually tested the following scenarios.

On IOS mobile devices 

Test 1 - 
- [x] Click on the action-menu and make a selection.
- [x] There is no focus-ring styles on the action-menu anymore.

Test 2 - 
- [x] Click on the action-menu and click outside to close it.
- [x] There is no focus-ring styles on the action-menu anymore.

Test 3 - 
- [x] Click on the action-menu and click outside to close it.
- [x] There is no focus-ring styles on the action-menu anymore.
- [x] Use tab and then shift-tab to focus back on the action using the keyboard.
- [x] The focus-ring appears on the action-menu as expected.

On All Desktop Devices

Test 1 - 
- [x] Click on the action-menu and make a selection.
- [x] There is no focus-ring styles on the action-menu.

Test 2 - 
- [x] Tab into the action-menu.
- [x] The focus-ring is shown as expected.
- [x] Use keyboard navigation to make a selection after opening the action menu using "enter" key.
- [x] The focus-ring is shown as expected.

## Types of changes

## Checklist

## Best practices


packages/picker/src/MobileController.ts
```diff
@@ -28,6 +29,16 @@ export class MobileController extends InteractionController {
         this.preventNextToggle = this.open ? 'yes' : 'no';
     }
 
+    public override handleButtonFocus(event: FocusEvent): void {
+        super.handleButtonFocus(event);
+
+        // if the focus comes from a closing the menu, we need to make sure to
+        // unset the :focus-visible state to avoid the focus ring on safari mobile
+        if (isWebKit() && !this.open) {
+            this.target.blur();
```
rubencarvalho, line 38: Are we sure that `.blur()` is the desired approach here? From my understanding, using .blur() will remove focus from the element entirely, but our intention seems to be to keep the element focused while only removing the visual indication (the focus ring). Or how does it work in the other browsers?
Can we check where the focus goes with this implementation? Does it impact keyboard navigation?

packages/picker/src/MobileController.ts
```diff
@@ -41,12 +57,8 @@ export class MobileController extends InteractionController {
...
             { signal }
         );
-        this.target.addEventListener(
-            'focus',
-            (event: FocusEvent) => this.handleButtonFocus(event),
-            {
-                signal,
-            }
-        );
+        this.target.addEventListener('focusout', () => this.handleFocusOut(), {
```
Rajdeepc, line 60: How are these events removed?

Rajdeepc (changes requested):

packages/picker/src/MobileController.ts
TarunAdobe: All of the eventListeners that I'm adding have an argument `signal` which binds them to the abortController and in the `InteractionController.ts` we abort all those signals in the hostDisconnected method.

packages/picker/src/MobileController.ts
```diff
@@ -26,6 +27,21 @@ export class MobileController extends InteractionController {
 
     public override handlePointerdown(): void {
         this.preventNextToggle = this.open ? 'yes' : 'no';
+        if (isWebKit()) {
+            this.target.classList.add('remove-focus-ring-safari-hack');
```
rubencarvalho, line 31: nit: we could put this class name in a `const`

rubencarvalho (commented):
This solution is actually ingenious! 😄
Do you think we could try and cover this with a simple test? I believe we can mock the `isWebKit` and check the computed style of the `::after` pseudo element. Let me know if you want to sync on this 👍

Rajdeepc (approved):
I am not sure you can surface up a test for this! But this is good!

packages/action-button/src/action-button.css
```diff
@@ -22,6 +22,38 @@ governing permissions and limitations under the License.
...
+    color: var(
+        --highcontrast-actionbutton-content-color-default,
+        var(
+            --mod-actionbutton-content-color-default,
+            var(--spectrum-actionbutton-content-color-default)
+        )
+    );
+}
+
+:host(.remove-focus-ring-safari-hack:focus-visible)::after {
```
Rajdeepc, line 53: if you add `:after` does it catch in the pre-commit check?

Rajdeepc:
> This solution is actually ingenious! 😄 Do you think we could try and cover this with a simple test? I believe we can mock the `isWebKit` and check the computed style of the `::after` pseudo element. Let me know if you want to sync on this 👍

Something on this front.
```js
const isWebKitMacOS = isWebKit() && detectOS() === 'Mac OS';
```

packages/picker/src/MobileController.ts
TarunAdobe: do i keep this const here in this file only? also any naming suggestions?

packages/action-button/src/action-button.css
```diff
@@ -22,6 +22,38 @@ governing permissions and limitations under the License.
     pointer-events: none !important;
 }
 
+/* remove-focus-ring-safari-hack is a custom css class that we are using to 
+unset the focus ring styles in safari. This is a temporary fix because safari 
+sets the :focus-visible pseudo class on the parent element when the dialog is closed.
+*/
+:host(.remove-focus-ring-safari-hack:focus-visible) {
+    background-color: var(
```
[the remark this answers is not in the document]
TarunAdobe: because our spectrum-css file updates these properties on :focus-visible.

.stylelintrc.json
```diff
@@ -5,7 +5,7 @@
         "header/header": ["config/license.js", {}],
         "length-zero-no-unit": [true, { "ignore": "custom-properties" }],
         "selector-type-no-unknown": [true, { "ignore": ["custom-elements"] }],
-        "selector-pseudo-element-colon-notation": ["single", {}],
+        "selector-pseudo-element-colon-notation": null,
```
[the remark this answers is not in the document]
TarunAdobe: I agree. Not sure what error I was getting earlier by using ::after but it seems to work with :after now. So reverted it back.

TarunAdobe:
Note: I have to add tests to this and also make sure that it works for picker and other related components (if any)

packages/picker/src/MobileController.ts
TarunAdobe: done!

rubencarvalho (approved):

PR #4726, opened 2024-09-05
feat: new value and constraints API for Calendar and DateTimePicker

mizgaionutalexandru:
Many changes from this PR relate to one another, hence the extent of it.

### New value API 
```js
type DateValue = CalendarDate | CalendarDateTime | ZonedDateTime
type DateTimePickerValue = DateValue 

// and for the Calendar because different types allow for separate further developing
type CalendarValue = DateValue
```
- The components accept multiple types but will convert the types that have time/time zone to `CalendarDate` using the `toCalendarDate` funtion provided by the internationalized/date library. 
- The components also have a `clear` method that clears its value, being a more declarative way of saying `value = undefined`

```js
import {CalendarDate} from '@internationalized/date';

<sp-calendar
    .value=${new CalendarDate(2020, 2, 14)}
    .min=${new CalendarDate(2020, 0, 1)}
    .max=${new CalendarDate(2025, 0, 1)}
></sp-calendar>

 // or the DateTimePicker
<sp-date-time-picker
    .value=${new CalendarDate(2020, 2, 14)}
    .min=${new CalendarDate(2020, 0, 1)}
    .max=${new CalendarDate(2025, 0, 1)}
>
    <sp-help-text slot="negative-help-text">
        Select a date between 2020 and 2025.
    </sp-help-text>
</sp-date-time-picker>
```
- If the DateTimePicker's set properties have different types the component will try to convert to the most specific one. 
For example, if `value` is of type `CalendarDate` but the `min` constraint is of type `CalendarDateTime`, the `value` will be converted to `CalendarDateTime` with a time of 00.00 or 12 AM. If said date's time doesn't comply with the min constraint it should get the minimum available time from it. 
- If the transformed dates don't comply with the min and max constraints the pre-selected value will be ignored and a warning will be shown in debug mode. 

### Removed TextField and InputSegments from the inheritance chain
- It looks like the `Textfield` class was extended only for styling purposes. This PR removes that from the inheritance chain and the styles were be added manually from spectrum-css. This surfaced some problems with the `valid` variant and it was deleted for now since no requirement mentioned it but it was included in the RFC because it was previously available when the `TextField` was extended. This can be prio'ed later on if required.
- The `InputSegments` class seems too big and handles a lot of logic that is specific for the `DateTimePicker`. _Currently the logic was moved to the `DateTimePicker` file and will be extracted after some changes like the new value API are complete. An idea is to use something similar to a `ReactiveController` or simply a `InputManager` class that can partially handle some logic for the segments and be instantiated in the `DateTimePicker`._

### Tests structure for DateTimePicker
- These will be created in future PRs but this one includes the one for the new features introduced as well as some for the basic behavior of value changing of the segments. Many of them will fail and their exact specifics will be clarified and bugs will be solved in future PRs.

### Others
- Calendar's `getDatesInWeek` method now gets executed only when needed instead of every update, using `currentMonthDates` as a 2D array

## Related issue(s)

## How has this been tested?

Many changes are code refactoring, improvements and API changes. The changes should be covered by tests as well as the storybook.

## Types of changes

## Checklist

## Best practices


packages/date-time-picker/src/DateTimePicker.ts
```diff
@@ -38,122 +93,602 @@ import '@spectrum-web-components/popover/sp-popover.js';
...
+                    </div>
+                </sp-popover>
+            </sp-overlay>
+        `;
+    }
+
+    private handleChange(event: Event): void {
+        event.stopPropagation();
+        this.isCalendarOpen = false;
+        const calendarValue = (event.target as Calendar).value as CalendarDate;
```
mizgaionutalexandru, line 348: Feels wrong that a type assertion is needed here. This is needed because the Calendar component allows its value to be set to more than one type but it's internally converted to `CalendarDate`. 

If a consumer uses other types, i.e. ZonedDateTime, and uses the Calendar component, i see 2 options:
1. He converts that value to `CalendarDate` before passing it to the Calendar component
```ts
import {toCalendarDate} from '@internationalized/date';
<sp-calendar
    .value=${toCalendarDate(myZonedDateTime)}
    @change=${(event: Event) => {
          // Now when he gets that value it will always be of type CalendarDate
          console.log((event.target as Calendar).value)
     }}
></sp-calendar>


```
2. As of now, he can dirrectly pass any DateValue type (`CalendarDate | CalendarDateTime | ZonedDateTime`), but a type assertion is needed when reading the value

Personally, now I'm leaning more towards the second option, as it aligns more with the meaning of the Calendar component.

packages/date-time-picker/src/DateTimePicker.ts
```diff
@@ -38,122 +93,602 @@ import '@spectrum-web-components/popover/sp-popover.js';
...
+
+        if (!this.includesTime) this.value = calendarValue;
+        else {
+            const hour = this.segments.hour?.value ?? 0;
+            const minute = this.segments.minute?.value ?? 0;
+            const second = this.segments.second?.value ?? 0;
+            const time = new Time(hour, minute, second);
+            this.value = toCalendarDateTime(calendarValue, time);
```
mizgaionutalexandru, line 356: I think here the value can also be a `ZonedDateTime` so a check needs to be added to know exactly into what type to convert the `CalendarDate`.

packages/calendar/src/Calendar.ts
```diff
@@ -169,16 +161,80 @@ export class Calendar extends SpectrumElement {
...
+                    `<${this.localName}> expects the 'min' to be less than 'max'. Please ensure that 'min' property's date is earlier than 'max' property's date.`,
+                    'https://opensource.adobe.com/spectrum-web-components/components/calendar' // TODO: update link
+                );
+                this.min = undefined;
+                this.max = undefined;
+            }
+        }
 
-        if (changedProperties.has('max')) this.setMaxCalendarDate();
+        if (changesValue && this.value) {
```
mizgaionutalexandru, line 201: If only the interval gets changed by the consumer, the component's value should still be checked against said interval. This applies for the DateTimePicker as well and will be modified in an upcoming PR with unit tests for it.
 
```suggestion
if (this.value)
```

packages/date-time-picker/test/date-time-picker.test.ts
```diff
@@ -10,30 +10,1908 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+                    await sendKeys({ press: 'ArrowUp' });
+                    await elementUpdated(element);
+                    monthSegment.focus();
+                    await sendKeys({ press: 'ArrowUp' });
+                    await sendKeys({ press: 'ArrowUp' });
+                    await elementUpdated(element);
+
+                    expect(yearSegment.innerText).to.equal(`2022`); // Common year in the Gregorian calendar
+                    expect(monthSegment.innerText).to.equal(`02`);
+                    expect(daySegment.innerText).to.equal(`28`);
```
mizgaionutalexandru, line 369: For incrementing/ decrementing/ typing in/ clearing the segments when a date is defined (the DateTimePicker has a value), the tests should also check for a correct update of said value. These will be added in a follow-up.

packages/date-time-picker/src/DateTimePicker.ts
```diff
@@ -38,122 +93,602 @@ import '@spectrum-web-components/popover/sp-popover.js';
...
+        const dateTimes = dateValuesDefined.filter(isCalendarDateTime);
+        if (dateTimes.length > 0) return dateTimes[0];
+
+        const dates = dateValuesDefined.filter(isCalendarDate);
+        return dates[0];
+    }
+
+    private currentDate: ZonedDateTime = now(this.timeZone);
+
+    private convertToMostSpecificDateValue(): void {
```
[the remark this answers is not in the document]
mizgaionutalexandru: This will apply in the next PR about precision updates.

packages/date-time-picker/src/DateTimePicker.ts
mizgaionutalexandru: Fixed in the next PR about precision updates.

packages/calendar/src/Calendar.ts
mizgaionutalexandru: Fixed in the next PR.

PR #4744, opened 2024-09-13
chore(reactive-controllers): updates lit dependency

jnjosh:
## Description

When we updated to support Lit 2 or Lit 3 we left Reactive Controllers behind, only supporting 3. This has created downstream issues in consumers that don't support Lit 3.

## How has this been tested?

- [ ] Need some help in how to best test this.

## Checklist

-   [-] I have added tests to cover my changes.
-   [-] All new and existing tests passed.

## Best practices


rubencarvalho (approved):
I took a note to add downstream dummy consumers with different constraints to dry-run our upcoming releases. This would then be caught.
Thank you for fixing!

TarunAdobe (approved):
I love such PRs!!

Rajdeepc (changes requested):
Awesome! But sadly this will not work. `reactive-controllers` will need to be pointed to `^3.1.3` since `@spectrum-web-components/reactive-controllers 0.42.2 depends on lit ^3.1.3` which means `^2.5.0 || ^3.1.3` will not be not true in this case.

graynorton:
> `reactive-controllers` lit dependency needs to be pointed to `^3.1.3`

@Rajdeepc, can you provide any detail on why the `@spectrum-web-components/reactive-controllers` package won't work with Lit `2.5.0`?

Rajdeepc:
> > `reactive-controllers` lit dependency needs to be pointed to `^3.1.3`
> 
> @Rajdeepc, can you provide any detail on why the `@spectrum-web-components/reactive-controllers` package won't work with Lit `2.5.0`?

My initial findings is that with `lit: 2.5.0`, storybook `7.x` is not supported under @storybook/types. Still digging into it!

graynorton:
> My initial findings is that with `lit: 2.5.0`, storybook `7.x` is not supported under @storybook/types. Component preview breaks under S7.

@Rajdeepc Let's try to sync up next week, maybe in the internal discussions meeting on Tuesday. I tried a bunch of things and couldn't identify any failures as a result of making this change, so would be good to compare notes. Thanks!

caseyisonit:
@graynorton @Rajdeepc what was the outcome of the convo mentioned above? Do we still support Lit2? If so we should get this in

rubencarvalho:
This got fixed by https://github.com/adobe/spectrum-web-components/pull/5105. Closing.

Issue #4733, opened 2024-09-10
[Bug]: Can't trigger change event on breadcrumbs via keyboard


rickharris:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-breadcrumbs

### Expected behavior

When focused on a breadcrumb item, I would expect that hitting enter (or possibly spacebar as well, though I'm not sure what the correct behavior for the sp-breadcrumb-item role would be) would trigger the `change` event on the sp-breadcrumbs element.

### Actual behavior

The `change` event is only triggered on the `sp-breadcrumbs` element when a breadcrumb item is clicked, but not when pressing enter or spacebar while a breadcrumb item has focus.

### Screenshots

https://github.com/user-attachments/assets/ac364adf-ee65-4162-aa94-f734836514bf

In this screen recording you can see that the change event is triggered when a breadcrumb item is clicked. After clicking through items, I focus a breadcrumb item and then press both enter and spacebar several times, and the event is not triggered.

### How can we reproduce this issue?

1. Go to https://studio.webcomponents.dev/edit/kmx7QJCb3pDgMh5caVtZ/src/index.ts?p=stories
2. Click on a breadcrumb item to see the event handler working
3. Focus on one of the breadcrumb items and press enter and spacebar
4. No event is triggered for keyboard events

### Sample code or abstract reproduction which illustrates the problem

https://studio.webcomponents.dev/edit/kmx7QJCb3pDgMh5caVtZ/src/index.ts?p=stories

### Severity

SEV 3

I've labeled this as SEV 3, but it could possibly be a SEV 2 depending on accessibility standards. It doesn't seem possible to navigate via the breadcrumb without a mouse.

najikahalsema:
Fixed via #4769

PR #4769, opened 2024-09-18
fix(breadcrumbs): trigger change event on breadcrumbs via keyboard

Rocss:
## Description

Due to `undefined` href attribute, breadcrumb items without hrefs would not trigger change events on keyboard Enter press.
I added a similar fix which is used for `sp-sidenav-item`, which is a placeholder href value.

## Related issue(s)
Fix for #4733

## How has this been tested?

-   [ ] _Test case 1_
    1. Go [here](https://4733--spectrum-web-components.netlify.app/storybook/?path=/story/breadcrumbs--default)
    2. Using the keyboard, navigate through the breadcrumb items and press "Enter"
    3. In the "Actions" addon, you should see the change event was triggered.

-   [ ] Did it pass in Mobile? - N/A
-   [ ] Did it pass in iPad? - N/A

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/breadcrumbs/src/BreadcrumbItem.ts
+++ b/packages/breadcrumbs/src/BreadcrumbItem.ts
@@ -82,17 +82,22 @@ export class BreadcrumbItem extends LikeAnchor(Focusable) {
         }
     }
 
+    protected handleKeyDown(event: KeyboardEvent): void {
+        if (event.key === 'Enter' || event.keyCode === 13) {
+            this.handleClick(event);
+        }
+    }
+
     protected renderLink(): TemplateResult {
         return html`
             <a
                 id="item-link"
-                href=${ifDefined(
-                    !this.isLastOfType ? this.href || '#' : undefined
-                )}
+                href=${ifDefined(!this.isLastOfType ? this.href : undefined)}
                 tabindex=${0}
                 aria-current=${ifDefined(
                     this.isLastOfType ? 'page' : undefined
                 )}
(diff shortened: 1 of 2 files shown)


Rajdeepc (approved):

rubencarvalho:
Would it work to handle the keydown event and trigger the same action as the click event when the `Enter` key is pressed?

pseudo-code:

```ts
<breadcrumb-item @keydown=${this.handleKeyDown}>

protected handleKeyDown(event: KeyboardEvent): void {
    if (event.key === 'Enter' || event.keyCode === 13) {
        this.handleClick(event);
    }
}

```

packages/breadcrumbs/src/BreadcrumbItem.ts
```diff
@@ -86,7 +86,9 @@ export class BreadcrumbItem extends LikeAnchor(Focusable) {
         return html`
             <a
                 id="item-link"
-                href=${ifDefined(!this.isLastOfType ? this.href : undefined)}
+                href=${ifDefined(
+                    !this.isLastOfType ? this.href || '#' : undefined
```
rubencarvalho, line 90: Could this lead to a jump to the top of the page when `"#"` is set?

Rajdeepc: Good catch! I have updated the changes. Let' see if we can land this for tomorrow's release.

blunteshwar (approved):

Issue #4783, opened 2024-09-25
[Bug]: decimal point missing on iPad soft keyboard on a number field with decimals


mizgaionutalexandru:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

NumberField

### Expected behavior

The numbered keyboard on iPad (the keyboard after you "pinch" it) is missing the decimal point when the number field uses decimals.

### Actual behavior

The numbered keyboard on iPad (the keyboard after you "pinch" it) **should show** the decimal point when the number field uses decimals.

### Screenshots

![simulator_screenshot_5FBAC497-7BF4-467D-952A-AAB8EDD748A9](https://github.com/user-attachments/assets/7fc39563-d025-4c0f-a4b8-645e44e5f944)

### What browsers are you seeing the problem in?

Safari

### How can we reproduce this issue?

1. Go to the decimals numberfield storybook ([here](https://opensource.adobe.com/spectrum-web-components/storybook/index.html?path=/story/number-field--decimals))
2. Click on the input itself to see the iPad's keyboard
3. Pinch the keyboard to minimize it
4. Observe that the keyboard doesn't have a decimal point, only numbers

### Severity

SEV 3

PR #4784, opened 2024-09-25
fix(number-field): show decimal on iPad minimized keyboard

mizgaionutalexandru:
## Description

This PR changes the inputMode of the `NumberField`'s `input` to `decimal` on iPad so the minimized keyboard can show the decimal point. 

## Related issue(s)

- https://github.com/adobe/spectrum-web-components/issues/4783

## Motivation and context

Currently an iPad user can't use the minimized keyboard to type in decimal values (i.e. 2.5).

## How has this been tested?

-   [ ] _Test case 1_
    1. Go to the decimals numberfield storybook ([here](https://imizga-i-pad-os-decimal-missing--spectrum-web-components.netlify.app/storybook/?path=/story/number-field--decimals))
    2. Click on the input itself to see the iPad's keyboard
    3. Pinch the keyboard to minimize it
    4. Observe that the keyboard now shows a decimal point
  
-   [ ] _Test case 2_
    1. Go to an editable slider's storybook ([here](https://imizga-i-pad-os-decimal-missing--spectrum-web-components.netlify.app/storybook/?path=/story/number-field--decimals))
    2. Click on the input itself to see the iPad's keyboard
    3. Pinch the keyboard to minimize it
    4. Observe that the keyboard now shows a decimal point

-   [x] Did it pass in iPad?

simulator_screenshot_A78C42F2-2EBC-452B-8564-CE398B04C020

simulator_screenshot_6D7D8B36-1359-4B82-B172-619A0627EF13

## Types of changes

## Checklist

## Best practices


TarunAdobe (commented):
is it possible for us to add a test for this fix so that we don't mess it up in the future?

mizgaionutalexandru:
> is it possible for us to add a test for this fix so that we don't mess it up in the future?

Currently all the tests I know in SWC that are device specific (for example [this iPhone decimals on number-field](https://github.com/adobe/spectrum-web-components/blob/e14d0827cd5a190e63b14418b8dd89fe2cab49ac/packages/number-field/test/number-field.test.ts#L120)) are skipped, so it wouldn't catch the issue if it comes up again. I don't yet see a reason to create these tests just for the sake of it.
If it is required and you see the value in it, I can add some tests similar to the mentioned one, to check the component's `inputMode`. How do you feel about this @TarunAdobe, am I missing something?

packages/number-field/src/NumberField.ts
```diff
@@ -802,30 +803,21 @@ export class NumberField extends TextfieldBase {
         }
 
         if (changes.has('min') || changes.has('formatOptions')) {
-            let inputMode = 'numeric';
             const hasNegative = typeof this.min !== 'undefined' && this.min < 0;
```
rubencarvalho, line 806: What happens if `min` is not defined but the current value is negative?, would it make sense to add this to the condition?

rubencarvalho:
> > is it possible for us to add a test for this fix so that we don't mess it up in the future?
> 
> Currently all the tests I know in SWC that are device specific (for example [this iPhone decimals on number-field](https://github.com/adobe/spectrum-web-components/blob/e14d0827cd5a190e63b14418b8dd89fe2cab49ac/packages/number-field/test/number-field.test.ts#L120)) are skipped, so it wouldn't catch the issue if it comes up again. I don't yet see a reason to create these tests just for the sake of it. If it is required and you see the value in it, I can add some tests similar to the mentioned one, to check the component's `inputMode`. How do you feel about this @TarunAdobe, am I missing something?

Maybe this is outside the scope of this PR, but I'm curious—why are these tests being skipped? Is there something specific preventing us from unskipping them?

packages/number-field/src/NumberField.ts
mizgaionutalexandru: Currently that's not taken into account but I see some value in it. I also see some concerns after a minimal PoC - if we do this at (every) value update it might confuse users to show different keyboards based on the component's value. Therefore I really am not sure if we want something like this.

Do you want to further investigate/ talk more about this in a new issue?
I also opened a [new one](https://github.com/adobe/spectrum-web-components/issues/4788) for a visual bug I found during this investigation.

mizgaionutalexandru:
> Maybe this is outside the scope of this PR, but I'm curious—why are these tests being skipped? Is there something specific preventing us from unskipping them?

AFAIK so far the project didn't focus on mobile and there were very few cases where setting the [user agent](https://github.com/adobe/spectrum-web-components/blob/e14d0827cd5a190e63b14418b8dd89fe2cab49ac/packages/number-field/test/number-field.test.ts#L1232) was neccessary. As I see from the `@web/test-runner-playwright` docs we can create different contexts, for an Iphone for example, as per [these](https://github.com/adobe/spectrum-web-components/blob/main/web-test-runner.utils.js) contexts created in SWC, but I haven't yet found a way to use specific contexts for only a subset of tests.  Is there any chance you can help me with this? there are a lot of unknowns for me here.

packages/number-field/src/NumberField.ts
rubencarvalho: >  I also see some concerns after a minimal PoC - if we do this at (every) value update it might confuse users to show different keyboards based on the component's value

Right, one approach could be to implement this as a flag at the component level, which remains fixed once toggled. I don’t have a strong preference here, but it might be helpful to emphasize in [our documentation](https://opensource.adobe.com/spectrum-web-components/components/number-field/#minimum-and-maximum-values) the importance of setting the minimum value explicitly. Currently, if negative numbers are allowed but no minimum value is set, users on iOS won’t see the minus key on the keyboard.

> Do you want to further investigate/ talk more about this in a new issue?
Not sure, maybe we can find a compromise with either one (or both?) of the approaches above.

> I also opened a https://github.com/adobe/spectrum-web-components/issues/4788 for a visual bug I found during this investigation.

Oof! Thanks for finding and reporting it!

rubencarvalho:
> > Maybe this is outside the scope of this PR, but I'm curious—why are these tests being skipped? Is there something specific preventing us from unskipping them?
> 
> AFAIK so far there were very few cases where setting the [user agent](https://github.com/adobe/spectrum-web-components/blob/e14d0827cd5a190e63b14418b8dd89fe2cab49ac/packages/number-field/test/number-field.test.ts#L1232) was neccessary (but playwright doesn't seem to support it). As I see from the `@web/test-runner-playwright` docs we can create different contexts, for an Iphone for example, as per [these](https://github.com/adobe/spectrum-web-components/blob/main/web-test-runner.utils.js) contexts created in SWC, but I haven't yet found a way to use specific contexts for only a subset of tests. Is there any chance you can help me with this? there are a lot of unknowns for me here.
> 
> I've encountered this issue in [this PR](https://github.com/adobe/spectrum-web-components/pull/4722) as well.
> 
> Nevertheless I will create an iPad test similar to [these](https://github.com/adobe/spectrum-web-components/blob/e14d0827cd5a190e63b14418b8dd89fe2cab49ac/packages/number-field/test/number-field.test.ts#L1232), maybe we will enable iPhone/iPad/Android contexts or something similar in the future.

Thanks for the context! 👍 I am OK with keeping this outside the scope of this PR for now. I’ve noted it down, and it’s also implicitly part of our ongoing and upcoming efforts to improve mobile support in the future.

TarunAdobe (approved):

packages/number-field/src/NumberField.ts
mizgaionutalexandru: > Not sure, maybe we can find a compromise with either one (or both?) of the approaches above.

What if we use something like this instead:
```suggestion
            const hasNegative = typeof this.min === 'undefined' || this.min < 0;
```

This way I think it aligns more with the current documentation saying:
> If a valid range is known ahead of time, it is a good idea to provide it to `<sp-number-field>` so it can optimize the experience. For example, when the minimum value is greater than or equal to zero, it is possible to use a numeric keyboard on iOS rather than a full text keyboard (necessary to enter a minus sign).

Now we always assume that the user can type in negative values (which he would've gotten to using the decrement button multiple times), easing the experience and opening the keyboard that allows the most flexibility. 
This way we also don't need "hack-ish" flags or workarounds. If a consumer knows that he wants only positive values, he can simply set the min to zero and the keyboard will adapt and not try to force-show a minus sign.

Does this make sense for you, @rubencarvalho ?

rubencarvalho: I agree with your approach, and it does align better with the documentation!
But since we are assuming the number field can have negative numbers by default unless the `min` is greater than `0`, what do you think about re-writing the condition to be more explicit:
```ts
const hasOnlyPositive = typeof this.min !== 'undefined' && this.min >= 0;
```

mizgaionutalexandru: Indeed, I think going for default `false` values is better for boolean variables and using `hasOnlyPositives` it would be false by default (with no properties provided by the consumer). 
I updated the logic. Thank you for pointing this out! 🙏🏻

rubencarvalho (approved):
Thank you for the patience! :D LGTM!

Issue #4786, opened 2024-09-26
[Bug]: menu opened stops page keyboard shortcuts from working


mizgaionutalexandru:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Menu

### Expected behavior

An opened sp-menu should allow application keyboard shortcuts.

### Actual behavior

An opened sp-menu **doesn't** allow application keyboard shortcuts, stopping the propagation of _almost_ all keydown events.

### How can we reproduce this issue?

1. Go to [the menu storybook](https://opensource.adobe.com/spectrum-web-components/storybook/index.html?path=/story/menu-submenu--submenu)
2. Press "V" (a storybook keyboard shortcut), observe that nothing happens.
3. Close the menu
4. Press "V" again
5. See that the keyboard shorcut now works.

### Sample code or abstract reproduction which illustrates the problem

The problem seems to be related to the interaction between the menu and rovingTabIndex controller as per this PR: https://github.com/adobe/spectrum-web-components/pull/3780/files#r1661806320.

### Severity

SEV 2

PR #4793, opened 2024-10-01
fix(menu): allow keyboard events to propagate from opened menu

TarunAdobe:
## Description

An opened sp-menu should allow application keyboard shortcuts.

## Related issue(s)

- #4786

## Motivation and context

An opened sp-menu doesn't allow application keyboard shortcuts, stopping the propagation of almost all keydown events.

## How has this been tested?

-   Added a test and it passed ;)

-   [ ] Did it pass in iPad?

## Types of changes

## Checklist

## Best practices


blunteshwar (approved):
LGTM! 🔥

Rajdeepc (approved):

rubencarvalho (changes requested):
Let us further discuss if this is the most adequate approach to solving this issue. We are diverging from the native `select` element behaviour which, when opened, captures focus and directs keyboard events specifically towards interacting with its options (even though we currently have not implemented the "interacting with its options" part).

Can we check if consumers can solve this on their side by adding an event listener in the capture phase instead of relying solely on the bubbling phase?

lazd:
> Let us further discuss if this is the most adequate approach to solving this issue. We are diverging from the native `select` element behaviour which, when opened, captures focus and directs keyboard events specifically towards interacting with its options (even though we currently have not implemented the "interacting with its options" part).

Good point, I didn't realize `<select>` eats all keyboard events when opened, even listeners added with `capture: true`! However, I can still use keyboard shortcuts such as `Command W` and `Command +` when focused on and interacting with a `<select>`.

> Can we check if consumers can solve this on their side by adding an event listener in the capture phase instead of relying solely on the bubbling phase?

This would work and was actually my first inclination. I suppose we should speak with accessibility and ask what they think the best behavior is here.

rubencarvalho:
Hello @lazd, can you confirm if you could solve this on your side?

lazd:
Hey @rubencarvalho, unfortunately, we're unable to use a capture listener on our side to work around this bug.

This behavior in SWC was actually [introduced 4 months ago in a feature PR that was adding support for action menus in action bars](https://github.com/adobe/spectrum-web-components/pull/3780/files#r1656403552), apparently to prevent a potential bug:

image

It was never intentionally added to model behavior after `<select>`, and this is clearly an unintended side effect of that change.

I suggest we merge this PR after testing that the original but this change was meant to address is no longer happening. Thank you!

caseyisonit:
@rubencarvalho @nikkimk @TarunAdobe  can we confirm if the menu omnibus accessibility work addressed this issue? or what conversations do we need to have to update the status on this PR?

rubencarvalho:
For this particular case, we agreed that consumers should handle it on the application level (resolved in Slack thread).
There is a similar issue opened here about other components: https://github.com/adobe/spectrum-web-components/issues/5099

Issue #4805, opened 2024-10-07
[Bug]: Text suggestions appear for `sp-number-field` when using Japanese/Simplified Chinese/Traditional Chinese IME


Rocss:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-number-field

### Expected behavior

If only digits are allowed in the field then suggestions should not appear for the same when enter inputs using IMEs.

### Actual behavior

 Although a text is not allowed to be input, Text candidates appear.

### Screenshots

![Screenshot 2024-10-07 at 15 20 17](https://github.com/user-attachments/assets/9c102a81-0531-4779-97ba-a4fff4f81368)

### What browsers are you seeing the problem in?

Chrome

### How can we reproduce this issue?

1. Go to https://opensource.adobe.com/spectrum-web-components/components/number-field/
2. Make sure to use Japanese IME (you can find this in the Keyboard settings)
3. Try to enter some text (ex. 'あい'in JP) into the numeric input box.
4. Observe the text.

### Sample code or abstract reproduction which illustrates the problem

Maybe this is due to the `input type="text"`?

### Severity

Minor

Rocss:
@rubencarvalho I prepared a fix for this on `rocss/4805` branch, but looking at how native `input type=number` behaves, I am not sure it should be fixed in the first place? The[ HTML input type number](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number) allows non-numeric characters when using Japanese/Chinese IME. However, looking at the brothers next door, React Spectrum restricts to numeric characters.
Thoughts?

rubencarvalho:
I agree with your observation, and we typically align with web standards. However, a key aspect of `<input type="number>` is that it _should_ have built-in validation to reject non-numerical entries. Given this expectation, I believe we are still in the clear here in enforcing numeric-only input, even when using an IME. Thank you for contributing the fix 👍

PR #4817, opened 2024-10-14
fix(number-field): allow only numeric characters for Japanese/Chinese IME

Rocss:
## Description

Text candidates appear for characters by inputting them with Japanese/Simplified Chinese/Traditional Chinese IME for numbers in respective fields where only digits are allowed.

## Related issue(s)

- #4805 

## Motivation and context

I think users should not be allowed to type "love" for example in an `sp-number-field`, when using Japanese/Chinese IME.
We do this check for english keyboard, but it seems it was missed to internationalised keyboards. 
Internal tracking: CCEX-145560

## How has this been tested?

-   [ ] _Test case 1_ Repeat for Safari/Chrome/Firefox
    1. Go [here](https://rocss-4805--spectrum-web-components.netlify.app/components/number-field/)
    2. Make sure to use Japanese IME (you can find this in the Keyboard settings)
    3. Try to enter some text (ex. 'あい'in JP) into the numeric input box.
    4. You should not be allowed to enter characters that are not numeric

-   [ ] Did it pass in iPad?

## Types of changes

## Checklist

## Best practices


packages/number-field/test/inputs.test.ts
```diff
@@ -466,5 +466,18 @@ describe('NumberField - inputs', () => {
...
+
+            el.focusElement.focus();
+            el.dispatchEvent(new CompositionEvent('compositionstart'));
+            await sendKeys({ type: 'あい' });
+
+            await elementUpdated(el.focusElement);
+            await nextFrame();
+
+            expect(el.focusElement.value).to.equal('100');
+            el.dispatchEvent(new CompositionEvent('compositionend'));
```
rubencarvalho, line 480: Could we add another assertion after the `compositionend` event to ensure the value remains unchanged once composition is complete?
```suggestion
            el.dispatchEvent(new CompositionEvent('compositionend'));
                        expect(el.focusElement.value).to.equal('100');
```

packages/number-field/src/NumberField.ts
```diff
@@ -496,6 +496,18 @@ export class NumberField extends TextfieldBase {
 
     protected override handleInput(event: InputEvent): void {
         if (this.isComposing) {
+            // If user actually types a new character.
+            if (event.data) {
```
Rajdeepc, line 500: If event.data is null, it could lead to unexpected behavior. For example, pressing backspace or delete can cause event.data to be empty.

Rajdeepc: An early return would be more tactical!
```
if (event.data === null || event.data === '') {
    return;
}
```

packages/number-field/src/NumberField.ts
```diff
@@ -496,6 +496,18 @@ export class NumberField extends TextfieldBase {
...
+                // Don't allow non-numeric characters even in composing mode.
+                const partialValue = this.convertValueToNumber(event.data);
+
+                if (Number.isNaN(partialValue)) {
```
Rajdeepc, line 504: Does this allow pasting inputs?

packages/number-field/src/NumberField.ts
```diff
@@ -496,6 +496,18 @@ export class NumberField extends TextfieldBase {
...
+                    this.inputElement.value = this.indeterminate
+                        ? indeterminatePlaceholder
+                        : this._trackingValue;
```
Rajdeepc, line 507: Resetting the value to `this._trackingValue` can be confusing! Can we show some validation feedback instead of a reset?

Rajdeepc (changes requested):

packages/number-field/src/NumberField.ts
Rocss: `if (event.data)` also guards for `null` and empty string, and after this this block is a `return` at line 511.
Wouldn't it work the same?

packages/number-field/src/NumberField.ts
Rocss: I followed the same logic as on english keyboard (line 547), we are resetting here because you should not be able to type non-numeric numbers, so we're removing what you typed.

packages/number-field/src/NumberField.ts
Rocss: yep

packages/number-field/src/NumberField.ts
blunteshwar: Makes Sense!

Rocss: good catch, fixed 👍  can you try again?

Rocss:
@Rajdeepc @blunteshwar @rubencarvalho any more feedback here?
I'd like to close this ticket this sprint if possible, thanks!

rubencarvalho (approved):

PR #4837, opened 2024-10-18
chore: added icon mapping logic

Rajdeepc:
## Description

Uses  [icons-name-mapping.csv](https://adobe.sharepoint.com/:x:/s/SpectrumOrgLeadership/ETBKE8VJXu5FtgJkBFwPO9oBP5WOyjui3-0ceJUTGNpRUg?e=uPxh0j) to map icon names between s1 and s2.

## Related issue(s)

-

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_

    1. Go here
    2. Do this

-   [ ] Did it pass in iPad?

## Types of changes

## Checklist

## Best practices


packages/icons-workflow/bin/build.js
```diff
@@ -19,6 +19,20 @@ import prettier from 'prettier';
 import Case from 'case';
 import { fileURLToPath } from 'url';
 
+const systemsIconMapping = {
```
Rajdeepc, line 22: So are we hardcoding the mapping for now? or this object can be dynamic?

packages/icons-workflow/bin/build.js
```diff
@@ -216,8 +230,11 @@ async function buildIcons(icons, tag, iconsNameList) {
         // check if the icon is present in the other version
         let otherVersionIconImport = defaultIconImport;
 
-        if (iconsNameList.includes(ComponentName)) {
-            const alternateTag = tag === 'icons' ? 'icons-s2' : 'icons';
+        const alternateTag = tag === 'icons' ? 'icons-s2' : 'icons';
+
+        if (systemsIconMapping[ComponentName]) {
```
Rajdeepc, line 235: Lets add some documentation block here!

packages/iconset/stories/icons-demo.ts
```diff
@@ -182,11 +234,29 @@ export class IconsDemo extends SpectrumElement {
...
                   (icon) => icon.name.toLowerCase().search(this.search) !== -1
               )
             : this.icons;
+        if (this.spectrumVersion === 2) {
+            // don't show icons from s1IconsList
+            matchingIcons = matchingIcons.filter((icon) => {
+                let iconName = icon.name;
+                iconName = iconName.replace(/\s/g, '');
+                iconName = iconName.toLowerCase();
+                return iconsList['s2'].includes(iconName);
```
Rajdeepc, line 248: Instead of repeating this logic in if and else let's write it this way!
```js
const filterIcons = (icon, spectrumType) => {
    let iconName = icon.name;
    iconName = iconName.replace(/\s/g, '');
    iconName = iconName.toLowerCase();
    return iconsList[spectrumType].includes(iconName);
}
```
// call like filterIcons(icon, 's1')

packages/iconset/stories/iconsList.json
```diff
@@ -0,0 +1 @@
...hapes","shareandroid","share","shoppingcart","shuffle","similar","slideshow","socialnetwork","sortdown","sortup","sort","stampclone","starfilled","star","stepbackward","stepforward","stickynote","strokedotted","strokesolid","strokewidth","switchvertical","switch","table","tag","target","temperature","template","textadd","textaligncenter","textalignjustifylastcenter","textalignjustifylastleft","textalignjustifylastright","textalignjustify","textalignleft","textalignright","textbold","textcapsall","textcapssmall","textitalic","textnumbers","textparagraph","textsize","textstrikethrough","textsubscript","textsuperscript","textunderline","textvariablefontsettings","text","thumbdown","thumbup","toggle","touchonefingerswipeleftright","transcript","transformdistort","transformgeneric","transformperspective","transformskew","transformwarp","translate","tutorials","unlink","undo","unlinkhoriz","unlinkvertical","uploadtocloud","upload","useradd","useravatar","useredit","userfollowing","usergroup","userlock","user","userslock","vectordraw","video","viewgrid","viewlist","viewtransparency","visibilityoff","visibility","volumeoff","volumeone","volumetwo","webnavbar","webpage","zoomin","zoomout"]}
```
Rajdeepc, line 1: Curious to know how did you generate this json?

packages/action-bar/stories/template.ts
```diff
@@ -37,9 +37,9 @@ export const Template = ({
                           <sp-icon-edit slot="icon"></sp-icon-edit>
                       </sp-action-button>
                       <sp-action-button slot="buttons" label="Share">
-                          <sp-icon-share-light
+                          <sp-icon-share
```
Rajdeepc, line 40: Is this done to match the mapping of S1 and S2?

packages/iconset/stories/iconsList.json
rubencarvalho, line 1: TypeScript doesn’t support JSON imports out of the box unless you explicitly enable them :)  
We need to add `resolveJsonModule: true` to our `tsconfig`.

packages/icons-workflow/bin/build-icons-mapping.js
rubencarvalho, line 1: Will we need this for icons-ui as well?

packages/icons-workflow/bin/build-icons-mapping.js
```diff
@@ -0,0 +1,67 @@
...
+
+    // save icons mapping to a json file
+    fs.writeFileSync(
+        path.join(__dirname, './icons-mapping.json'),
+        JSON.stringify(mapping),
+        'utf8'
+    );
+};
+
+await saveIconsMapping();
```
rubencarvalho, line 67: I won't make it blocking, but it would be great to have a mock `.csv` file with the expected structure to be able to test this.

packages/icons-workflow/bin/build-icons-mapping.js
Rajdeepc: Probably yes!

packages/icons-workflow/bin/build-icons-mapping.js
```diff
@@ -0,0 +1,67 @@
...
+import { fileURLToPath } from 'url';
+
+import { getComponentName } from './build.js';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+
+// builds and saves the icons mapping to a json file
+const saveIconsMapping = async () => {
+    // check if there is file with .csv extension in the bin folder
+    const files = fs.readdirSync(path.join(__dirname, './'));
```
rubencarvalho, line 25: We're using `fs` sync methods in an async function, which might block the event loop unnecessarily

packages/icons-workflow/bin/build-icons-mapping.js
```diff
@@ -0,0 +1,67 @@
...
+    rows.forEach((row) => {
+        // if the third column says "name change"
+        if (row.split(',')[2] === 'name change') {
+            const columns = row.split(',');
+
+            const oldName = getComponentName(columns[0]);
+            const newName = getComponentName(columns[1]);
+
+            // skip if the oldName is 123
+            if (oldName === '123') {
```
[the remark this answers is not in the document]
Rajdeepc: These could be added while testing. This should be coming as a part of the package!

packages/icons-workflow/bin/build-icons-mapping.js
```diff
@@ -0,0 +1,67 @@
...
+import { getComponentName } from './build.js';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+
+// builds and saves the icons mapping to a json file
+const saveIconsMapping = async () => {
+    // check if there is file with .csv extension in the bin folder
+    const files = fs.readdirSync(path.join(__dirname, './'));
+    let data = '';
+    files.forEach((file) => {
```
rubencarvalho, line 27: nit: we could use `for ... of` instead of `forEach`. this would allow for early termination when we've found the `.csv` file we need, instead of iterating over all files

packages/icons-workflow/bin/build-icons-mapping.js
```diff
@@ -0,0 +1,67 @@
...
+        if (file.endsWith('.csv')) {
+            data = fs.readFileSync(path.join(__dirname, file), 'utf8');
+        }
+    });
+
+    if (!data) {
```
rubencarvalho, line 33: Could we handle all potential errors, including when reading the file itself? While unlikely, file reads can fail due to various reasons (e.g., permissions issues, file encoding problems)

packages/icons-workflow/bin/build-icons-mapping.js
```diff
@@ -0,0 +1,67 @@
...
+        throw new Error(
+            'No csv file found in the src folder. Please make sure to have the csv file in the bin folder'
+        );
+    }
+
+    const rows = data.split('\n');
+    const mapping = {};
+    rows.forEach((row) => {
+        // if the third column says "name change"
+        if (row.split(',')[2] === 'name change') {
```
rubencarvalho, line 43: Could we start by moving the definition `const columns = row.split(',');` up?

nit: also if we were to use `for (const row of rows)` here, we could control the flow nicely, for example continuing `if (columns.length < 3 || !columns[2])`. But I am OK with the current approach!

packages/icons-workflow/bin/build-icons-mapping.js
```diff
@@ -0,0 +1,67 @@
...
+                return;
+            }
+
+            mapping[oldName] = newName;
+            mapping[newName] = oldName;
+        }
+    });
+
+    // save icons mapping to a json file
+    fs.writeFileSync(
```
rubencarvalho, line 60: Same suggestion as above:
```suggestion
    // Asynchronously save the icons mapping to a JSON file
    const outputPath = path.join(__dirname, './icons-mapping.json');
    try {
        await fs.promises.writeFile(outputPath, JSON.stringify(mapping), 'utf8');
        console.log('Icons mapping saved to icons-mapping.json');
    } catch (error) {
        throw new Error(`Error saving the icons mapping: ${error.message}`);
    }
```

Here, I used [fs.promises](https://nodejs.org/api/fs.html#fspromiseswritefilefile-data-options).

packages/icons-workflow/bin/build-icons-mapping.js
```diff
@@ -0,0 +1,67 @@
...
+            // skip if the oldName is 123
+            if (oldName === '123') {
+                return;
+            }
+
+            mapping[oldName] = newName;
+            mapping[newName] = oldName;
+        }
+    });
+
```
rubencarvalho, line 58: nit, nice to have: 

```suggestion
if (Object.keys(mapping).length === 0) {
  console.warn('No valid name changes found in the CSV file.');
}

```

packages/icons-workflow/bin/build.js
```diff
@@ -82,6 +84,40 @@ let manifestListings = `\r\nexport const iconManifest = [\r\n`;
...
+
+    if (ComponentName === 'TextStrikeThrough') {
+        ComponentName = 'TextStrikethrough';
+    }
+    if (ComponentName === 'UnLink') {
+        ComponentName = 'Unlink';
+    }
+
+    return ComponentName;
+};
```
rubencarvalho, line 119: I feel like this could be simplified...

We could use some regex for removing the beginning or end of the string and use a replacements map to make the code less repetitive.

```suggestion
export const getComponentName = (i) => {
    let id = path
        .basename(i, '.svg') 
        .replace(/^(S2_Icon_|S_)/, '')  
        .replace(/(_20_N|_22x20_N|_18_N@2x)$/, ''); 

    if (i.startsWith('Ad')) {
        id = i.replace(/^Ad/, '') + 'Advert';
    }

    const replacements = {
        'UnLink': 'Unlink',
        'TextStrikeThrough': 'TextStrikethrough',
        'github': 'GitHub'
    };

    return Case.pascal(replacements[id] || id);
};
```

packages/icons-workflow/bin/build.js
rubencarvalho, line 1: Also, on lines https://github.com/adobe/spectrum-web-components/pull/4837/files#diff-9cbbc8842323aa2e451a5e384f83bc11bfa026a19e09bb96de1e1fde0be1d877R46-R60, this could also be very much streamlined:

```ts
const ensureDirectoryExists = (dirPath) => {
    if (!fs.existsSync(dirPath)) {
        fs.mkdirSync(dirPath);
    }
};

const directories = [
    `${rootDir}packages/icons-workflow/src`,
    `${rootDir}packages/icons-workflow/src/icons`,
    `${rootDir}packages/icons-workflow/src/icons-s2`,
    `${rootDir}packages/icons-workflow/src/elements`,
    `${rootDir}packages/icons-workflow/icons`,
];

directories.forEach(ensureDirectoryExists);
```

packages/iconset/stories/icons-demo.ts
```diff
@@ -182,11 +234,21 @@ export class IconsDemo extends SpectrumElement {
         this.updateSearch(event);
     }
     private renderSearch(): TemplateResult {
-        const matchingIcons = this.search
+        let matchingIcons = this.search
```
rubencarvalho, line 237:
```suggestion
        let matchingIcons = this.search?.trim()
```

packages/iconset/stories/icons-demo.ts
```diff
@@ -182,11 +234,21 @@ export class IconsDemo extends SpectrumElement {
...
             ? this.icons.filter(
                   (icon) => icon.name.toLowerCase().search(this.search) !== -1
```
rubencarvalho, line 239:
```suggestion
                  (icon) => icon.name.toLowerCase().includes(this.search.toLowerCase())
```

packages/iconset/stories/icons-demo.ts
```diff
@@ -182,11 +234,21 @@ export class IconsDemo extends SpectrumElement {
...
               )
             : this.icons;
+
+        const iconVersion = this.spectrumVersion === 2 ? 's2' : 's1';
+        // Filter out icons that are not in the current version
+        matchingIcons = matchingIcons.filter((icon) => {
+            let iconName = icon.name;
+            iconName = iconName.replace(/\s/g, '');
+            iconName = iconName.toLowerCase();
```
rubencarvalho, line 248:
```suggestion
            const iconName = icon.name.replace(/\s/g, '').toLowerCase();
```

packages/icons-workflow/bin/build-icons-mapping.js
```diff
@@ -0,0 +1,67 @@
...
+import fs from 'fs';
+import path from 'path';
+import { fileURLToPath } from 'url';
+
+import { getComponentName } from './build.js';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+
+// builds and saves the icons mapping to a json file
```
rubencarvalho, line 22:
```suggestion
/*
     * You can run this script by running the `build-icons-mapping ` on `packages/icons-workflow/package.json`,
     * which generates the icons mapping from an internal CSV file.
     * This is a manual step and should only be run when the icons team provides updates.
     * Place the CSV file under the `packages/icons-workflow/bin` folder and run the script to generate the new JSON mapping file.
     * This JSON mapping should be committed to the repository.
     */

rubencarvalho (approved):
Awesome work! 👍 We still have a couple of open threads, please check if you want to address them or if you want to dismiss them.

packages/icons-workflow/bin/build-icons-mapping.js
TarunAdobe: Great Point... If we do this then would we be pushing it in the pr?

PR #4868, opened 2024-10-28
fix(menu): prevent sp-menu-item text-align cascading

mizgaionutalexandru:
## Description

Previous to the Overlay V2, the `text-align` property of action-menu, or similar components, did not cascade into the `sp-menu-items`. This PR brings back that behavior by setting the `text-align: initial` on `sp-menu-item`. 

PS: This is the same behavior as in React Spectrum. As per Spectrum guidelines I haven't yet seen a use-case for other types of aligment. If they are needed, we can provide a `--mod-*` type css variable.
There is also an on-going discussion on a spectrum-css [issue](https://github.com/adobe/spectrum-css/issues/3058) but this missalignment in behavior comes from the SWC's approach to Overlay so I think this is a good place to fix.

## Related issue(s)

- N/A

## Motivation and context

Fixes the text-align cascading into the overlay menu-items.

## How has this been tested?

-   [ ] _Test case 1_
    1. Open [ActionMenu Storybook](https://imizga-text-align--spectrum-web-components.netlify.app/storybook/?path=/story/action-menu--default)
    2. Place a `text-align: center` on a parent of the ActionMenu (e.g. sp-theme) 
    3. Click the ActionMenu
    4. Observe that the text inside the menu-items is centered

-   [x] Did it pass in iPad?

Without the change:
without the fix

After the change:
with the fix

## Types of changes

## Checklist

## Best practices


rubencarvalho (commented):
Thanks for the contribution! Could you look into adding a test for this? Either a new Storybook story or a check for the computed text-align value on the menu item in the scenario where it was breaking would probably work well - just to ensure it doesn’t break again. We can continue discussing with CSS to see if this should eventually be handled on their end, but for now, I’m comfortable having this solution in SWC.

TarunAdobe (approved):
LGTM!

PR #4875, opened 2024-10-29
fix(date-time-picker): ux bug fixes and improvements

mizgaionutalexandru:
## Description

Bugs fixed:
- deleting a year of a defined date makes the calendar no longer display the current one
- the calendar selects a year in the 1900s when the DTP has a year with 2- digits 
- calendar's previous month button increases the year if the year is 1
- the year can't be deleted on safari without changing the focus back and forth on the year segment
- Bengali(India) year is not formatted on increment/decrement showing arabic numerals
- warnings were shown outside of dev-mode

UX improvements:
- Calendar month navigation buttons are now disabled when reaching the end of an era
- Calendar now always displays the initial month in a place where dates are available
- Calendar now closes when the disabled attribute is changed 
- The year segment no longer has a min-width

Also, some methods were moved around inside the Calendar.ts source file to improve readability.

## Related issue(s)

## How has this been tested?

Unit tests.

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/calendar/src/Calendar.ts
+++ b/packages/calendar/src/Calendar.ts
@@ -214,7 +214,7 @@ export class Calendar extends SpectrumElement {
                     'https://opensource.adobe.com/spectrum-web-components/components/calendar' // TODO: update link
                 );
                 this.value = undefined;
-            } else this.currentDate = this.value as CalendarDate;
+            }
         }
     }
 
@@ -233,6 +233,7 @@ export class Calendar extends SpectrumElement {
         if (changesDates) {
             this.convertToCalendarDates();
             this.checkDatePropsCompliance(changesMin || changesMax);
+            this.currentDate = (this.value as CalendarDate) || this.today;
         }
 
         const previousMonth = changedProperties.get('currentDate');
--- a/packages/calendar/src/Calendar.ts
+++ b/packages/calendar/src/Calendar.ts
@@ -191,7 +191,16 @@ export class Calendar extends SpectrumElement {
         if (changesDates) {
             this.convertToCalendarDates();
             this.checkDatePropsCompliance(changesMin || changesMax);
-            this.currentDate = (this.value as CalendarDate) || this.today;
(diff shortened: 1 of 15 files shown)


packages/calendar/src/Calendar.ts
```diff
@@ -233,12 +191,22 @@ export class Calendar extends SpectrumElement {
...
+            if (this.value) this.currentDate = this.value as CalendarDate;
+            else {
+                const isTodayNonCompliant = this.isNonCompliantDate(this.today);
+
+                if (isTodayNonCompliant) {
+                    if (this.min) this.currentDate = this.min as CalendarDate;
+                    else if (this.max)
+                        this.currentDate = this.max as CalendarDate;
+                } else this.currentDate = this.today;
+            }
```
AndreiBaicu26, line 203: Consider moving these nested ifs in a single method to improve readability. 
Something like 
```

    setDate() {
        if (this.value){
            this.currentDate = this.value as CalendarDate;
            return;
        } 

        if(this.isNonCompliantDate(this.today) && this.min) {
            this.currentDate = this.min as CalendarDate;
            return;
        }

        if(this.isNonCompliantDate(this.today) && this.max) {
            this.currentDate = this.max as CalendarDate;
            return;
        }

        this.currentDate = this.today;
    }
```

packages/date-time-picker/src/date-time-picker.css
```diff
@@ -130,6 +130,8 @@ sp-popover .popover-content {
 .literal-segment {
     white-space: pre;
     user-select: none;
+
+    /* -webkit-user-select: none; */
```
AndreiBaicu26, line 134: Should be deleted ?

mizgaionutalexandru: Uncommented [here](https://github.com/adobe/spectrum-web-components/pull/4962/files#diff-0f4bef552114d643584f8ef0a08fb745b8555778f6e4c90f3ac0275e161df866).

PR #4912, opened 2024-11-07
fix(overlay): safari :focus-visible inconsistency when using overlay type modal

mizgaionutalexandru:
## Related issue(s)

-

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_

    1. Go here
    2. Do this

-   [ ] Did it pass in iPad?

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/overlay/src/HoverController.ts
+++ b/packages/overlay/src/HoverController.ts
@@ -11,13 +11,16 @@ governing permissions and limitations under the License.
 */
 
 import { conditionAttributeWithId } from '@spectrum-web-components/base/src/condition-attribute-with-id.js';
+import { isWebKit } from '@spectrum-web-components/shared';
 import { randomID } from '@spectrum-web-components/shared/src/random-id.js';
-
+import { noop } from './AbstractOverlay.js';
 import {
     InteractionController,
     InteractionTypes,
+    lastInteractionType,
 } from './InteractionController.js';
-import { noop } from './AbstractOverlay.js';
+
+export const SAFARI_FOCUS_RING_CLASS = 'remove-focus-ring-safari-hack';
 
 const HOVER_DELAY = 300;
 
@@ -32,15 +35,33 @@ export class HoverController extends InteractionController {
 
     pointerentered = false;
 
+    handleKeyup(event: KeyboardEvent): void {
+        if (isWebKit() && (event.code === 'Tab' || event.code === 'Escape')) {
+            this.open = true;
+            this.removeSafariFocusRingClass();
+        }
+    }
+
     handleTargetFocusin(): void {
         if (!this.target.matches(':focus-visible')) {
             return;
         }
+
+        if (
+            isWebKit() &&
+            this.target[lastInteractionType] === InteractionTypes.click
+        ) {
+            this.target.classList.add(SAFARI_FOCUS_RING_CLASS);
+            return;
+        }
+
         this.open = true;
         this.focusedin = true;
+        this.removeSafariFocusRingClass();
     }
 
     handleTargetFocusout(): void {
+        this.removeSafariFocusRingClass();
         this.focusedin = false;
         if (this.pointerentered) return;
         this.open = false;
@@ -139,6 +160,11 @@ export class HoverController extends InteractionController {
         this.abortController = new AbortController();
         const { signal } = this.abortController;
         this.target.addEventListener(
+            'keyup',
+            (event) => this.handleKeyup(event),
+            { signal }
+        );
+        this.target.addEventListener(
             'focusin',
             () => this.handleTargetFocusin(),
             { signal }
@@ -178,5 +204,13 @@ export class HoverController extends InteractionController {
             () => this.handleHostPointerleave(),
             { signal }
         );
+    }
+
+    private removeSafariFocusRingClass(): void {
+        if (
+            isWebKit() &&
+            this.target.classList.contains(SAFARI_FOCUS_RING_CLASS)
+        )
+            this.target.classList.remove(SAFARI_FOCUS_RING_CLASS);
     }
 }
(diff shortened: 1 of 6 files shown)


packages/overlay/src/HoverController.ts
```diff
@@ -11,13 +11,16 @@ governing permissions and limitations under the License.
...
-
+import { noop } from './AbstractOverlay.js';
 import {
     InteractionController,
     InteractionTypes,
+    lastInteractionType,
 } from './InteractionController.js';
-import { noop } from './AbstractOverlay.js';
+
+export const SAFARI_FOCUS_RING_CLASS = 'remove-focus-ring-safari-hack';
```
mizgaionutalexandru, line 23: I wonder where can we best define this const if we're going to use it in multiple places in the project. (currently it is also defined in [`MobileController`](https://github.com/adobe/spectrum-web-components/blob/3d93d530a7c9691753338f82a1486ccc8abac10f/packages/picker/src/MobileController.ts#L18))

packages/overlay/test/overlay-trigger-hover-click.test.ts
```diff
@@ -270,4 +276,84 @@ describe('Overlay Trigger - Hover and Click', () => {
...
+                    type: 'click',
+                    position: [0, 0],
+                },
+            ],
+        });
+        await closed;
+
+        // This fails but when manually tested it works in the browser
+        // in the test it shows that the tooltip is open (overlayTrigger.open === 'hover')
+        // expect(overlayTrigger.open).to.be.undefined;
```
mizgaionutalexandru, line 320: Would appreciate any feedback/ideas on why would this be flaky, couldn't yet figure out why this would fail if manually tested it works.

packages/overlay/test/overlay-trigger-hover-click.test.ts
```diff
@@ -270,4 +276,84 @@ describe('Overlay Trigger - Hover and Click', () => {
...
+
+        const opened = oneEvent(trigger, 'sp-opened');
+        await sendKeys({ press: 'Enter' });
+        await opened;
+        const closed = oneEvent(trigger, 'sp-closed');
+        await sendKeys({ press: 'Escape' });
+        await closed;
+
+        // Manually testing this in the browser doesn't fail without the handleKeyup method but it should
+        // expect(overlayTrigger.open).to.equal('hover');
```
mizgaionutalexandru, line 352: Same thing here as in the previous test.

Rajdeepc:
@mizgaionutalexandru Thanks for doing this! Shaping up well! By any chance were you able to test this in a touch interface?

packages/overlay/src/HoverController.ts
TarunAdobe: maybe define and export it from the `InteractionController`?

mizgaionutalexandru: I think that is a good idea, unfortunately there are 2 `InteractionController`s, one for the Overlay and one for the Picker, resulting in the same double declaration problem.

mizgaionutalexandru:
> @mizgaionutalexandru Thanks for doing this! Shaping up well! By any chance were you able to test this in a touch interface?

Yes, I have tested this on mobile and tablet simulators as well as a real iPad device. It seems to work just fine. I don't know why the written tests fail though.

packages/overlay/test/overlay-trigger-hover-click.test.ts
TarunAdobe: iinnnteresting

packages/overlay/src/HoverController.ts
TarunAdobe: I'd still say let's keep this constant in the `InteractionController` for now (separately for both picker nd overlay) cz that is the entry point for the interactions related stuff for both overlay and picker.

packages/overlay/src/HoverController.ts
```diff
@@ -32,15 +34,33 @@ export class HoverController extends InteractionController {
 
     pointerentered = false;
 
+    handleKeyup(event: KeyboardEvent): void {
+        if (isWebKit() && (event.code === 'Tab' || event.code === 'Escape')) {
+            this.open = true;
+            this.removeSafariFocusRingClass();
+        }
+    }
+
```
TarunAdobe, line 43:
```suggestion
    handleKeyup(event: KeyboardEvent): void {
        if (event.code === 'Tab' || event.code === 'Escape') {
            this.open = true;
            if(isWebKit()){
                this.removeSafariFocusRingClass();
            }
        }
    }

```

TarunAdobe (changes requested):
This needs to be updated so as to make sure the `hover` tooltip opens in the following case on other browsers too.
1. Click on the trigger open the popover
2. press escape 
(expect the hover to be opened again)

packages/overlay/test/overlay.test.ts
```diff
@@ -786,6 +790,74 @@ describe('Overlay - type="modal"', () => {
         await close;
         expect(el.open).to.be.null;
     });
+
+    describe.only('maintains consistency of focus ring in different browsers', () => {
```
rubencarvalho, line 794:
```suggestion
    describe('maintains consistency of focus ring in different browsers', () => {
```

rubencarvalho:
If we can just remove that forgotten `.only`, I am good to merge this! 👍

rubencarvalho (approved):
Thank you for this! 😃

TarunAdobe (changes requested):
Tests are still failing when run parallel to other ones. They're only passing if we only run those tests separately. We'll have to take another look :(

rubencarvalho (approved):

TarunAdobe (approved):

PR #4924, opened 2024-11-08
fix(action-group): add null check for slotElement in manageButtons

rubencarvalho:
## Description

Added a guard to the `manageButtons` function in the `ActionGroup` component to handle cases where `slotElement` is `null`. This prevents potential runtime errors when the `MutationObserver` triggers `manageButtons` before the component is fully rendered.

## Related issue(s)

- [SWC-595](https://jira.corp.adobe.com/browse/SWC-595)

## Motivation and context
We have gotten reports from customers that they were seeing this error under some circumstances:
Screenshot 2024-11-08 at 14 48 25

Adding this `null` check allows the component to handle this scenario gracefully.

## How has this been tested?

-   [x] _Added a test case in action-group.test.ts_
    To check it is not evergreen, comment out the newly added safeguard clause.

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/action-group/test/action-group.test.ts
+++ b/packages/action-group/test/action-group.test.ts
@@ -122,7 +122,7 @@ async function multipleSelectedActionGroup(
 }
 
 describe('ActionGroup', () => {
-    it('throws an error if slotElement is null', async () => {
+    it('does not throw an error if slotElement is null', async () => {
         // To verify that this test is not evergreen, you can temporarily disable the safeguard
         // clause in `manageButtons` by commenting out the following lines:
         // if (!this.slotElement) { return; }
--- a/packages/action-group/test/action-group.test.ts
+++ b/packages/action-group/test/action-group.test.ts
@@ -54,6 +54,7 @@ import { sendMouse } from '../../../test/plugins/browser.js';
 import { HasActionMenuAsChild } from '../stories/action-group.stories.js';
 import '../stories/action-group.stories.js';
 import { isWebKit } from '@spectrum-web-components/shared';
+import sinon from 'sinon';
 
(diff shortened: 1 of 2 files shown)


packages/action-group/test/action-group.test.ts
```diff
@@ -121,6 +122,34 @@ async function multipleSelectedActionGroup(
...
+                <sp-action-button value="first">First</sp-action-button>
+                <sp-action-button value="second">Second</sp-action-button>
+            </sp-action-group>
+        `);
+
+        // Stub the slotElement getter to return null
+        const slotElementStub = sinon.stub(el, 'slotElement').get(() => null);
+
+        // Call manageButtons and expect a TypeError
+        expect(() => Reflect.get(el, 'manageButtons').call(el)).not.to.throw(
```
Rajdeepc, line 140: What is Reflect?

Rajdeepc (changes requested):

packages/action-group/test/action-group.test.ts
rubencarvalho: You can check out more details [here on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect/get), but basically, I’m using `Reflect.get` as a hack to access the private `manageButtons` method on `el` in the test.
I tried a few different approaches to directly test `manageButtons`, including adding a subclass just for testing to expose these private methods (a pattern we could actually explore more if we want to start getting more isolated unit tests for private methods). Reflect.get, however, sounded like a good compromise solution.
I am happy to try out other ideas though!

Rajdeepc (approved):
Lets land this! Thank you

Issue #4689, opened 2024-08-23
[Bug]: OverlayTrigger can get stuck in a render loop causing page crash


jcmitch:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

OverlayTrigger

### Expected behavior

OverlayTrigger should render as expected regardless of quick successive updates to slot content.

### Actual behavior

Been struggling with an issue related to OverlayTrigger for awhile so posting here to see if anyone more familiar with how it works might have some ideas of what might be happening.

We have an app that uses OverlayTrigger but when rendering it gets in an infinite render loop causing the page to crash. This only happens when upgrading to the newer (not really "new" anymore) overlay approach in SWC. I've been able to track it down to a few places in the SWC code. Basically, it gets stuck in a loop of update, render, updated that never exits. I've been able to place logs in the SWC code to see this.

The update method from lit provides the map of changed properties and it seems that sometimes it is empty and sometimes it has hoverContent (which is the trigger I'm attempting to use). It alternates between these two states but sometimes will see a few empty objects before hoverContent pops in again. This does happen on initial render where logs in my code (outside of SWC) show 3 quick successive renders as everything gets setup. It does quickly settle to what appears to be a stable state on my end but by this point overlay-trigger is stuck in the loop.

The loop seems like it could be caused by the fact that render calls [renderHoverOverlay](https://github.com/adobe/spectrum-web-components/blob/main/packages/overlay/src/OverlayTrigger.ts#L274), which calls [renderSlot](https://github.com/adobe/spectrum-web-components/blob/main/packages/overlay/src/OverlayTrigger.ts#L212), which listens slotchange event and when handled calls [handleSlotContent](https://github.com/adobe/spectrum-web-components/blob/main/packages/overlay/src/OverlayTrigger.ts#L182), which sets [hoverContent](https://github.com/adobe/spectrum-web-components/blob/main/packages/overlay/src/OverlayTrigger.ts#L128), which triggers [update](https://github.com/adobe/spectrum-web-components/blob/main/packages/overlay/src/OverlayTrigger.ts#L155), which then triggers render, which seems to start to loop again. It seems like if the content in the "hover" slot isn't changing then it shouldn't be looping or at least only loop for the 3 initial renders I would expect to cause this update.
[...]

PR #4930, opened 2024-11-11
fix(overlay): returning only the first element for slot assignedElements

zolar:
Only return first element form `assignedElements` instead of an array of elements.

## Description

This PR reverses a helper method that was used in the previous version of the `OverLayTrigger`. In that version, the `slotchange` handler uses `assignedNodes` instead of `assignedElements`. I think using `assignedElements` renders the slot so fast, that our dialog component has not completed loading, leading to forcing a reload, in which it does goes threw the same process.

https://github.com/adobe/spectrum-web-components/blob/a532ff8a410abeefb54d9638a2316ae82570566e/packages/overlay/src/OverlayTrigger.ts#L454

## Related issue(s)
  

[[Bug]: OverlayTrigger can get stuck in a render loop causing page to crash](https://github.com/adobe/spectrum-web-components/issues/4689)

## Motivation and context

Resolves an issue related to endless render loop.

## How has this been tested?

-   [ ] Did it pass in iPad?

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/overlay/src/OverlayTrigger.ts
+++ b/packages/overlay/src/OverlayTrigger.ts
@@ -74,22 +74,22 @@ export class OverlayTrigger extends SpectrumElement {
     public receivesFocus: 'true' | 'false' | 'auto' = 'auto';
 
     @state()
-    private clickContent: HTMLElement[] = [];
+    private clickContent: HTMLElement | undefined;
 
     private clickPlacement?: Placement;
 
     @state()
-    private longpressContent: HTMLElement[] = [];
+    private longpressContent: HTMLElement | undefined;
 
     private longpressPlacement?: Placement;
 
     @state()
-    private hoverContent: HTMLElement[] = [];
+    private hoverContent: HTMLElement | undefined;
 
     private hoverPlacement?: Placement;
 
     @state()
-    private targetContent: HTMLElement[] = [];
+    private targetContent: HTMLElement | undefined;
 
     @query('#click-overlay', true)
     clickOverlayElement!: Overlay;
@@ -100,8 +100,12 @@ export class OverlayTrigger extends SpectrumElement {
     @query('#hover-overlay', true)
     hoverOverlayElement!: Overlay;
 
-    private getAssignedElementsFromSlot(slot: HTMLSlotElement): HTMLElement[] {
-        return slot.assignedElements({ flatten: true }) as HTMLElement[];
+    private getAssignedElementsFromSlot(
+        slot: HTMLSlotElement
+    ): HTMLElement | undefined {
+        return slot.assignedElements({ flatten: true })[0] as
+            | HTMLElement
+            | undefined;
     }
 
     private handleTriggerContent(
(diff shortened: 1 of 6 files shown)


Rajdeepc:
We would like to see you use case here considering this is very specific. Can you add a video or a repro to show how does this solve your issue?

zolar:
> We would like to see you use case here considering this is very specific. Can you add a video or a repro to show how does this solve your issue?

@Rajdeepc Yes, I can add a video or we can do another debug session so I can walk threw it with you or you team?

zolar:
@Rajdeepc This is a video of the component not rendering. The component container stays blank and you can see the console output going forever. It will eventually crash the browser. 

https://github.com/user-attachments/assets/dacc1d68-6b89-4f7b-b62e-147ab35cfb1c

Rajdeepc:
> @Rajdeepc This is a video of the component not rendering. The component container stays blank and you can see the console output going forever. It will eventually crash the browser.
> 
>  Sequence.low.02.mp4

This seems to be a condition where the entire component is re-rendering multiple times due to state or property changes. Can you do a repro of your use case by any means?

zolar:
> > @Rajdeepc This is a video of the component not rendering. The component container stays blank and you can see the console output going forever. It will eventually crash the browser.
> > Sequence.low.02.mp4
> 
> This seems to be a condition where the entire component is re-rendering multiple times due to state or property changes. Can you do a repro of your use case by any means?

@Rajdeepc 
This issue is only repoing in Express app. We have an example / development application in which we do not see this issue. I think that the issue has to do with Express being so resource intensive. I think this is a race condition in which the Dialog is not completely finished rendering when the `slothandler` attaches to the slot content, which forces it to `rerender`. I think that using he `nodes.find` delays in just enough that the Dialog completes rendering. The issue can also be fixed using the current `slot.assignedElements` and a zero wait promise `new Promise(resolve => setTimeout(resolve, 0));`. 

Express does use Service Workers but its still very resource heavy and I think its effecting the rendering of our component. We also use a lot of OverlayTriggers that might effect render changes also.

zolar:
> > @Rajdeepc This is a video of the component not rendering. The component container stays blank and you can see the console output going forever. It will eventually crash the browser.
> > Sequence.low.02.mp4
> 
> This seems to be a condition where the entire component is re-rendering multiple times due to state or property changes. Can you do a repro of your use case by any means?

@Rajdeepc This only happens in Express, it does not repo in our development at or any other environment that I know of. There maybe a late state change that is happening but its not from a map or any other iterator that I could find. I think its just a normal state change or its being delayed by the slag in the process.

Rajdeepc (changes requested):
Can you please make sure that the PR is clean from any css related changes on radio and search!

Rajdeepc:
@zolar Also please check the tests in `overlay-trigger-longpress.test.ts` line 222 which needs to flatten to
```js
            await waitUntil(() => {
                const localName = (
                    this.el as unknown as { targetContent: HTMLElement }
                ).targetContent?.localName;
```

zolar:
Closing this PR since this does not fix the root cause.

Issue #3272, opened 2023-06-02
[Bug]: sp-number-field pressing increment/decrement buttons using NVDA moves focus to the text field


jnurthen:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-number-field

### Expected behavior

When I press the increment or decrement buttons focus should stay on the button. I can't press it more than once if focus moves

### Actual behavior

Focus currently moves to the text field

### What browsers are you seeing the problem in?

Chrome, Microsoft Edge

### How can we reproduce this issue?

1. Go to https://opensource.adobe.com/spectrum-web-components/components/number-field/
2. Turn on NVDA 
3. arrow through the page (down arrow) until you get to a number field
4. Press down arrow one more time to get to the increment button
5. press space or enter 
Note focus moves to the text field and forms mode gets enabled

najikahalsema:
Related to #3274 #3271

PR #4933, opened 2024-11-12
chore(number-field): added aira-hidden attribute for inc/dec buttons

blunteshwar:
## Description

Added aria-hidden attribute for increment and decrement buttons in sp-number-field

## Related issue(s)

#3272 
-

## Motivation and context

Earlier the increment and decrement buttons didn't have aria-hidden attribute which was making sp-number-field less accessible. It also allowed NVDA/ screen readers to pseudo-focus in increment and decrement buttons which should not happen as it will add un-necessary tab stops. Thus by adding `aria-hidden="true"` the problem was solved and sp-number-field became more accessible.

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_

    1. Go here
    2. Do this

-   [ ] Did it pass in iPad?

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/number-field/src/NumberField.ts
+++ b/packages/number-field/src/NumberField.ts
@@ -748,6 +748,7 @@ export class NumberField extends TextfieldBase {
                               block="start"
                               class="button step-up"
                               aria-describedby=${this.helpTextId}
+                              aria-hidden="true"
                               label=${'Increase ' + this.appliedLabel}
                               size=${this.size}
                               tabindex="-1"
@@ -765,6 +766,7 @@ export class NumberField extends TextfieldBase {
                               block="end"
                               class="button step-down"
                               aria-describedby=${this.helpTextId}
+                              aria-hidden="true"
                               label=${'Decrease ' + this.appliedLabel}
                               size=${this.size}
                               tabindex="-1"


Rajdeepc:
Can we add a video of a screen reader test here to make sure we are not suppressing any other accessibility contracts!

blunteshwar:
> * Documentation Site
PFA the screen recording for the same!


https://github.com/user-attachments/assets/8023ffb9-539a-4246-bf30-66b2da420e33

Rajdeepc:
@nikkimk Can you do a round of review here before I merge this to main!

Rajdeepc (approved):

packages/number-field/src/NumberField.ts
```diff
@@ -748,6 +748,7 @@ export class NumberField extends TextfieldBase {
                               block="start"
                               class="button step-up"
                               aria-describedby=${this.helpTextId}
```
nikkimk, line 750: Since this is `aria-hidden`, we do not need `aria-describedby`

packages/number-field/src/NumberField.ts
```diff
@@ -765,6 +766,7 @@ export class NumberField extends TextfieldBase {
                               block="end"
                               class="button step-down"
                               aria-describedby=${this.helpTextId}
```
nikkimk, line 768: Since this is aria-hidden, we do not need aria-describedby

packages/number-field/test/number-field.test.ts
```diff
@@ -1752,40 +1744,4 @@ describe('NumberField', () => {
             await clickBySelector(el, '.step-down');
         });
     });
```
nikkimk, line 1746: We should probably have tests for the following:

1. That the buttons don't appear in the a11ySnapshot
2. That they cannot receove keyboard focus

nikkimk (changes requested):
Looking good. I recommend a couple of additional changes.

nikkimk (approved):
Looks good to me once it is up-to-date and tests passing

Rajdeepc (approved):

PR #4938, opened 2024-11-18
refactor(card): clarify component's slot API

mizgaionutalexandru:
## Description

This PR addresses APIs misunderstandings regarding the `sp-card` component's slots. Previously used slots `preview` and `cover-photo` are now merged into a single one named `image`. The styles are applied exactly as before, based on the `variant` and `horizontal` properties, through using the according id that maps to spectrum CSS styles.
Also, now the `quiet` variant is included in the style overrides, and a new `--mod-*` CSS variable was added to allow consumers to reliably obtain the desired behaviour (the added story reflects the usage of this token).
Some CSS was also changed to remove deprecated overrides and improve readability. 

## Related issue(s)

- N/A

## Motivation and context

1. The component public API [states that](https://github.com/adobe/spectrum-web-components/blob/5cf5d34645bf9494ebd20f64c42d1619523d2d84/packages/card/src/Card.ts#L45) the `cover-photo` slot should be used with the `quiet` variant. The consumer should not need to determine which slot to use with each variant.
2. [Spectrum-CSS uses](https://opensource.adobe.com/spectrum-css/card.html#undefined:~:text=A%20quiet%20card%20for%20an%20image.) `preview` with the `quiet` variant
3. SWC's [code examples](https://github.com/adobe/spectrum-web-components/blob/v1.0.0/packages/card/README.md#quiet) use `preview` with the `quiet` variant
4. The component, while placing the user in a position to make the correct decision about what slot to use with a specific variant, allows for both slots to be used at the same time but [doesn't display the cover-photo anyway](https://github.com/adobe/spectrum-web-components/blob/5cf5d34645bf9494ebd20f64c42d1619523d2d84/packages/card/src/card.css#L48)
5. Looks like the `quiet` variant was included in the `width: 100%` at a previous point in time but the change to not include it anymore wasn't documented too much. There is a need to come back to the previous behaviour (to include the `quiet` variant in the full width CSS). It seems like a `--mod-*` is the best common ground to allow consumers all behaviours related to using an image and how to position it inside the asset, be that cover, contain, scale-down etc. ([discussion context](https://cclight.slack.com/archives/CESK60MQD/p1727720298481989))

## How has this been tested?

No functional changes were added. All the slots usage in the project have been updated. This being said, all unit tests should pass and VRTs should show changes only for the quiet variant's initial width. A new story was added to reflect the usage of the newly added token.

## Types of changes

## Checklist

## Best practices


packages/card/src/card.css
```diff
@@ -44,19 +37,15 @@ slot[name='description'] {
...
 
-#preview + #cover-photo {
-    display: none;
-}
-
 #cover-photo ::slotted(*),
-:host(:not([variant='quiet'])) #preview ::slotted(*) {
-    width: 100%;
+#preview ::slotted(*) {
+    inline-size: 100%;
```
mizgaionutalexandru, line 42: Should include the `quiet` variant [as per spectrum-css](https://github.com/adobe/spectrum-web-components/blob/5cf5d34645bf9494ebd20f64c42d1619523d2d84/packages/card/src/spectrum-card.css#L567).

Rajdeepc:
This will become a breaking change! Will bring this up in our office hours!

packages/card/src/Card.ts
```diff
@@ -19,30 +20,30 @@ import {
...
-import '@spectrum-web-components/asset/sp-asset.js';
-
-import { Checkbox } from '@spectrum-web-components/checkbox/src/Checkbox';
+import {
+    ifDefined,
+    when,
+} from '@spectrum-web-components/base/src/directives.js';
 import '@spectrum-web-components/checkbox/sp-checkbox.js';
-import '@spectrum-web-components/popover/sp-popover.js';
+import { Checkbox } from '@spectrum-web-components/checkbox/src/Checkbox';
```
Rajdeepc, line 32:
```suggestion
import { Checkbox } from '@spectrum-web-components/checkbox';
```

packages/card/src/Card.ts
```diff
@@ -19,30 +20,30 @@ import {
...
+import headingStyles from '@spectrum-web-components/styles/heading.js';
+import cardStyles from './card.css.js';
 
 /**
  * @element sp-card
  *
  * @fires change - Announces a change in the `selected` property of a card
- * @slot preview - This is the preview image for Gallery Cards
- * @slot cover-photo - This is the cover photo for Default and Quiet Cards
+ * @slot image - This is the image of the card
```
Rajdeepc, line 46:
```suggestion
 * @slot image - Image content
```

packages/card/src/Card.ts
```diff
@@ -207,7 +204,7 @@ export class Card extends LikeAnchor(
         this.addEventListener('pointercancel', handleEnd);
     }
 
-    protected get renderHeading(): TemplateResult {
+    protected renderHeading(): TemplateResult {
```
Rajdeepc, line 207: I dont see any computation logic here so why change to a callable function instead of keeping it as a getter?

Rajdeepc (changes requested):
We need to create a section called `BREAKING CHANGE` to surface up this change to the consumers. Let's talk in the office hours and get some more opinions from other product teams. I feel there is a lot of usage of the Card and the asset components in multiple products.

rubencarvalho:
Thank you for your contribution here! I’m fully on board with this proposal. Just one thing: we’ve been following a practice of providing a migration period where we log developer warnings to inform users and allow them time to migrate while maintaining backward compatibility.
Could we log a descriptive message recommending the new `image` slot approach while still keeping the existing `cover-photo` and `preview` slots for now? We could then follow up with another PR to remove all deprecated functionality and tests. This intermediary step has proven to be very useful in the past :)

packages/card/src/Card.ts
mizgaionutalexandru: I think the naming fits a method better than a getter due to the fact that it contains a verb. Would you prefer another naming pattern?

PR #4960, opened 2024-11-27
[DRAFT] Make icons great again!!

TarunAdobe:


Rajdeepc:
This is a good start but let's create a RFC on this first, we need to talk to the product teams also to make sure we are building for them and not for the component library. We can connect this PR on the RFC also.

caseyisonit:
@TarunAdobe are we still making icons great or has this lost wind and we should close the PR?

TarunAdobe:
> @TarunAdobe are we still making icons great or has this lost wind and we should close the PR?

We can close this PR. This was meant for the demo purposes only and that purpose is fulfilled. We are surely going to make icons great again, just probably not here.

PR #4982, opened 2024-12-09
feat: screen reader support for Calendar and DateTimePicker

mizgaionutalexandru:
## Description

This PR aims to offer screen reader support for the most common use-cases of these components.

### The calendar:
- Announces when a date is selected
- Announces when the month is changed
- Announces when the user navigates using Arrows or VO + Arrows
- Announces the month and year when going into the component
- Announces the type of action on previous/next month buttons
- New `labels` property to override english-default labels

### The DateTimePicker:
- Announces the current focused segment as a stepper with the correct type translated according to the locale ('year'/'month' etc.)
- Announces if the current focused segment is Empty
- Announces the new value of the segment when it gets changed 
- Announces the type of button 'Calendar'
- Announces the spelled out date value (or 'Empty') when going into the component
- New `labels` property to override english-default labels

## Related issue(s)

## Motivation and context

Screen reader is one of the basic testings - the minimum level expected for any product of Adobe. 

## How has this been tested?

mainly Chrome with macOS VoiceOver (cmd + F5)

-   [ ] _Test case 1_
    1. Go to the Calendar storybook 
    2. Tab to the Calendar component
    3. Observe that the month and year get announced
    4. Tab to the days grid and use VO + Arrows to navigate through the dates
    5. Observe that the VoiceOver announces the full dates for each day in the grid
    6. Select a date using VO + space
    7. Observe that the new selected value is announced
-   [ ] _Test case 2_
    1. Go to the Calendar storybook
    2. Tab to the next/previous month buttons
    3. Observe that the action is announced ('Previous'/'Next')
    4. Change the month using VO + space
    5. Observe that the new month and year get announced
-   [ ] _Test case 3_
    1. Go to the DateTimePicker storybook
    2. Tab to the DateTimePicker component
    3. Observe that the value gets announced (or 'Empty')
    4. Tab through the segments (year/month/hour etc.)
    6. Observe that the type of the segment gets announced along with the value
    7. Change the value of the segment using Arrow Up/ Arrow Down or Type In
    8. Observe that the new value gets announced

-   [ ] Did it pass in iPad?

## Types of changes

-   [ x New feature (non-breaking change which adds functionality)

## Checklist

## Best practices

### DIFF
--- a/packages/calendar/src/Calendar.ts
+++ b/packages/calendar/src/Calendar.ts
@@ -133,14 +133,14 @@ export class Calendar extends SpectrumElement {
     private currentMonthDates: CalendarDate[][] = [];
 
     @state()
-    protected set isDateFocusIntent(value: boolean) {
+    private set isDateFocusIntent(value: boolean) {
         if (this._isDateFocusIntent === value) return;
 
         this._isDateFocusIntent = value;
         this.requestUpdate('isDateFocusIntent', !value);
     }
 
-    protected get isDateFocusIntent(): boolean {
+    private get isDateFocusIntent(): boolean {
         return this._isDateFocusIntent;
     }
     private _isDateFocusIntent: boolean = false;
@@ -200,7 +200,10 @@ export class Calendar extends SpectrumElement {
             changedProperties.has('currentDate') &&
             (!previousDate || !isSameMonth(previousDate, this.currentDate));
 
-        if (changesMonth) this.setCurrentMonthDates();
+        if (changesMonth) {
+            this.setCurrentMonthDates();
+            this.setAttribute('aria-label', this.monthAndYear);
+        }
     }
 
     override updated(changedProperties: PropertyValues): void {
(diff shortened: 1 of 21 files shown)


packages/calendar/src/Calendar.ts
```diff
@@ -286,36 +304,29 @@ export class Calendar extends SpectrumElement {
...
         return html`
             <div class="header" @focusin=${this.resetDateFocusIntent}>
-                <!--
-                 * TODO: Attribute 'role="heading"' removed, due to error 'The "heading" role requires the attribute
-                 * "aria-level"'
-                -->
-                <div
+                <h2
                     class="title"
                     aria-live="assertive"
```
mizgaionutalexandru, line 319: This seems to not work on Safari with VoiceOver, it always announces the initial month-year of the Calendar. 
I think a general approach similar to the [React Spectrum's Live Announcer](https://github.com/adobe/react-spectrum/blob/main/packages/%40react-aria/live-announcer/src/LiveAnnouncer.tsx) might fix this issue.

packages/date-time-picker/src/DateTimePicker.ts
```diff
@@ -541,6 +559,15 @@ export class DateTimePicker extends ManageHelpText(
...
                 role="spinbutton"
+                aria-valuenow=${ifDefined(segment.value)}
+                aria-valuemin=${segment.minValue}
+                aria-valuemax=${segment.maxValue}
+                aria-label=${segment.label}
+                aria-valuetext=${ifDefined(
+                    segment.value !== undefined
+                        ? segment.formatted
+                        : this.labels.empty
+                )}
```
mizgaionutalexandru, line 570: With this approach it seems like on Chrome it doesn't always read the whole value of the segment but instead the VoiceOver announces "selection replaced {number}". Also, deleting the segment's value makes the VO announce the content of the segment (----) instead of "Empty".
Also, on Safari the VO announces "beginning of text" instead of the segment's value when they change using the ArrowUp/Down, and when trying to use the VO + ArrowUp/Down to change the stepper's value it ocasionally crashed the webpage.

I'm currently exploring ways of fixing these issues. In the meantime, could someone please suggest some ideas? Thank you!

packages/calendar/src/Calendar.ts
```diff
@@ -200,20 +212,26 @@ export class Calendar extends SpectrumElement {
             changedProperties.has('currentDate') &&
             (!previousDate || !isSameMonth(previousDate, this.currentDate));
 
-        if (changesMonth) this.setCurrentMonthDates();
+        if (changesMonth) {
+            this.setCurrentMonthDates();
+            this.setAttribute('aria-label', this.monthAndYear);
```
mizgaionutalexandru, line 217: I wonder if this is the best way to make the screen reader announce the current month-year when focusing in the Calendar. (the same method was applied for the DateTimePicker as well)

packages/calendar/src/Calendar.ts
Rocss: I think we would want to use `polite`, not `assertive`, so that if any other things are read, the screen reader would wait for those to finish first before reading this one

packages/date-time-picker/src/DateTimePicker.ts
Rocss: @nikkimk ^

mizgaionutalexandru:
Will merge this as is. 
Left concers will be addressed in the `feature branch -> main` PR or in a follow-up issue/PR.

PR #4999, opened 2024-12-19
DRAFT: Make process-spectrum more robust

graynorton:
## Description

Make `tasks/process-spectrum` more robust to changes in CSS theme overrides.

*Making this a draft for now, as it may need discussion.*

## Related issue(s)

-

## Motivation and context

When I first got the SWC CSS files in my combined monorepo proof-of-concept building directly from the CSS workspaces, I found that a bunch of the styling was visibly broken in Storybook. After poking around a bit, I determined that this was because bunch of the components were referencing `--system-foo-bar-baz` custom properties that weren't actually defined. Looking further, it seemed like:

- The CSS at the tip of s2-foundations-redux (which was where I'd pulled from) no longer defined a bunch of theme-specific variables that previously been present in the released versions of the CSS components that SWC had been referencing (presumably because they were no longer necessary—yay!)

- The logic in SWC's `tasks/process-spectrum script` wasn't well equipped to deal with the case where previously existing CSS theme files were removed. Specifically, `process-spectrum` avoided writing `*-overrides.css` files for non-existent or empty theme files. This seems logical enough, but causes problems because it means that "stale" `*-overrides.css` files are left behind when the corresponding theme files go away.

I tweaked the logic in `process-spectrum` to unconditionally write `*-overrides.css files`, and that seemed to take care of the problem.

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_

    1. Go here
    2. Do this

-   [ ] Did it pass in iPad?

## Types of changes

## Checklist

## Best practices


tasks/process-spectrum.js
```diff
@@ -627,30 +627,17 @@ async function processComponent(componentPath) {
...
-                }
+                // Note: We write the overrides file even if it's empty.
+                // This is to ensure that we don't end up with stale overrides
+                // files in the case where the bridge file previously contained
+                // overrides but no longer does.
+                writeMachineGeneratedSourceFile(overridesPath, code);
+            } else {
+                // For the same reason, we write an empty file if the bridge file
+                // doesn't exist (in case it previously did).
+                writeMachineGeneratedSourceFile(overridesPath, '');
```
rubencarvalho, line 640: Should we delete the file entirely if it’s empty instead of writing an empty file? 💭

tasks/process-spectrum.js
```diff
@@ -818,9 +805,14 @@ async function processComponent(componentPath) {
...
-            `/*
+        writeMachineGeneratedSourceFile(outputPath, code);
+    }
+}
+
+function writeMachineGeneratedSourceFile(outputPath, code) {
+    fs.writeFileSync(
+        outputPath,
+        `/*
 Copyright 2023 Adobe. All rights reserved.
```
rubencarvalho, line 816: nit: We can bump this to 2025 😄 🥳

tasks/process-spectrum.js
graynorton: Good question! That was my first instinct as well, but then I recalled that these generated `foo-override.css` files are imported by manually maintained `foo.css` files.

If we allow the override files to be routinely created / deleted, we would either need to start having the script modify the currently-manually-maintained files to update the imports, OR have the script output some list of deleted override files and depend on ourselves to make the corresponding edits.

Neither of those solutions seemed great, and since most of the components do in fact have overrides files, it seemed like unconditionally creating the overrides was the least-bad option. I'm open to being convinced otherwise, though!

graynorton: Of course, I'm hoping that all of this stuff can be made to go away in our post-1.x setup!

caseyisonit:
@graynorton @rubencarvalho do we want to clean up this PR and get it reviewed or can we close this?

rubencarvalho:
Closing - it got cherry-picked in https://github.com/adobe/spectrum-web-components/pull/5004

PR #5001, opened 2024-12-23
docs: updated documentation for Calendar and DateTimePicker components

mizgaionutalexandru:
## Description

Updating the documentation of the new components: Calendar and DateTimePicker according to implemented features for the initial release.

Also, the storybook was updated to include the locale picker using the chromatic globals settings.

## Related issue(s)

## Motivation and context

Ease of use.

## How has this been tested?

Main changes has been tested in Storybook and on the local docs server.

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/calendar/README.md
+++ b/packages/calendar/README.md
@@ -34,7 +34,7 @@ By default, the `<sp-calendar>` element will display the current month with emph
 ## Value
 
 A pre-selected date value can be provided to the `<sp-calendar>` element and it will dictate the displayed month of the element.
-The element's `value` property accepts multiple date types (`CalendarDate`, `CalendarDateTime` and `ZonedDateTime`), but when read the property will always be of type `CalendarDate`.
+The element's `value` property accepts dates of `DateValue` type (`CalendarDate | CalendarDateTime | ZonedDateTime`), but when read the property will always be of type `CalendarDate`.
 More about these types and when to use each one can be found on the [`@internationalized/date` page](https://react-spectrum.adobe.com/internationalized/date/index.html).
 
 To clear the user-selected value, the public `clear` method can be used.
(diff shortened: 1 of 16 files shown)


packages/calendar/README.md
```diff
@@ -23,20 +25,115 @@ import { Calendar } from '@spectrum-web-components/calendar';
...
+<sp-calendar id="min-max-calendar"
+    .value=${new CalendarDate(2020, 2, 14)}
+></sp-calendar>
+```
+
+## Minimum and maximum values
+
+The `min` and `max` properties can be used to limit the selected value to a specific range. Ranges can be open, by providing only one of the mentioned properties, or closed, when both `min` and `max` are provided. These properties are also of `DateValue` type, but will always be of `CalendarDate` type when read, similar to `value`.
+
+If the provided closed interval is not a valid one (e.g `min > max`), the assignment of those properties gets ignored. The same way, if a pre-selected value doesn't comply with the interval, it gets ignore and the element will behave as it wouldn't have a `value` provided.
```
AndreiBaicu26, line 54:
```suggestion
If the provided closed interval is not a valid one (e.g `min > max`), the assignment of those properties gets ignored. The same way, if a pre-selected value doesn't comply with the interval, it gets ignored and the element will behave as it wouldn't have a `value` provided.
```

packages/calendar/src/Calendar.ts
```diff
@@ -78,20 +78,20 @@ export class Calendar extends SpectrumElement {
      * The selected date in the calendar. If defined, this also indicates where the calendar opens.
      * If not, the calendar opens at the current month.
      */
-    @property({ attribute: false })
-    value?: CalendarValue;
+    @property({ type: Object })
```
AndreiBaicu26, line 81: Can this be set to `DateValue` ?

AndreiBaicu26 (approved):

PR #5002, opened 2024-12-24
feat: Calendar and DateTimePicker components

mizgaionutalexandru:
## Description

Calendar and DateTimePicker components with functionality for the initial release. More about what each component can be seen in their documentation and storybook.

These components are the main ones that benefit from the Storybook locale picker, apart from `NumberField`. Therefore, this PR also includes the addition of said locale picker using chromatic globals.

## Related issue(s)

## How has this been tested?

- Functionality is covered by unit tests
- Layout can be compared with Spectrum CSS([Calendar](https://opensource.adobe.com/spectrum-css/?path=/docs/components-calendar--docs) and [DateTimePicker](https://opensource.adobe.com/spectrum-css/?path=/docs/components-date-picker--docs)). Multiple PRs along the way allowed VRTs to check the progress/changes

## Types of changes

## Checklist

## Best practices


Rajdeepc:
Hi @mizgaionutalexandru is this design approved with a separate clear button?

Screenshot 2025-01-08 at 1 47 26 PM

mizgaionutalexandru:
> Hi @mizgaionutalexandru is this design approved with a separate clear button?

Hi, @Rajdeepc! No, I just wanted to show case a use-case extension that consumers can apply.

Rajdeepc (changes requested):

packages/calendar/src/Calendar.ts
```diff
@@ -0,0 +1,756 @@
...
+                    size="s"
+                    label=${this.labels.previous}
+                    class="prevMonth"
+                    data-test-id="prev-btn"
+                    ?disabled=${this.isPreviousMonthDisabled}
+                    @click=${this.handlePreviousMonth}
+                >
+                    <div slot="icon">
+                        <slot name="prev-icon">
+                            <sp-icon-chevron-left></sp-icon-chevron-left>
```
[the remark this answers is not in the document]
mizgaionutalexandru: I'm not sure that I fully understand your question. As of now, consumers will be able to change the icons for the "previous month" and "next month" action buttons.

projects/story-decorator/decorator.ts
```diff
@@ -32,26 +32,29 @@ export const swcThemeDecoratorWithConfig =
     ({ bundled } = { bundled: true }) =>
     (
         story: () => TemplateResult,
-        context: import('@storybook/csf').StoryContext<any, any>
+        context: import('@storybook/csf').StoryContext
```
caseyisonit, line 35: Why did we remove the type check here over improving them from `any`?

projects/story-decorator/decorator.ts
```diff
@@ -32,26 +32,29 @@ export const swcThemeDecoratorWithConfig =
...
+        // Needed so that the `sp-language-context` event can be caught by decorator's `sp-theme`
+        requestAnimationFrame(() => {
+            document.documentElement.setAttribute('lang', 'en-US');
+            const decorator = document.querySelector(
+                'sp-story-decorator'
+            ) as HTMLElement;
+            render(bundled ? story() : html``, decorator);
+        });
+
+        const isGlobalSettingDefined = (setting: string): boolean => {
```
caseyisonit, line 46: nice clean up

packages/calendar/.npmignore
caseyisonit, line 1: I dont believe we suppress these on a component level, can you provide context for needing this ignore file?

packages/calendar/exports.json
caseyisonit, line 1: This file is not part of our component architecture pattern. Can you explain what this does and why it should be included? (this goes for both components but I don't like duplicating comments)

packages/calendar/src/calendar-overrides.css
caseyisonit, line 1: We should have way more auto generated CSS styles based on their documentation: https://opensource.adobe.com/spectrum-css/?path=/docs/components-calendar--docs

packages/calendar/.npmignore
Rajdeepc: Thanks for catching this. The PR needs some cleanup!

projects/story-decorator/decorator.ts
mizgaionutalexandru: Generally, using `any` would cause typescript not to typecheck things. As Ruben said, removing this allows typescript to infer the default generic types. I think this works for now and if in the future we will need some more specific types we should investigate that rather than using `any`.

projects/story-decorator/decorator.ts
```diff
@@ -32,26 +32,29 @@ export const swcThemeDecoratorWithConfig =
     ({ bundled } = { bundled: true }) =>
```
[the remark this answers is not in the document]
mizgaionutalexandru: I haven't touched this line in this PR, but as far as I understand you'd like to have a type declared here, right?

If so, I think your suggestion could work if we add a default value for the whole object as well, to not have to provide an empty object whenever the function is called.
`({ bundled = true }: { bundled?: boolean } = {}) =>`
Or, this way:
`({ bundled }: { bundled?: boolean } = { bundled: true })`

What do you think? Should i make this change in this PR?
Thanks!

packages/date-time-picker/src/DateTimePicker.ts
```diff
@@ -0,0 +1,873 @@
...
+        if (this.readonly) return nothing;
+
+        return html`
+            <sp-picker-button
+                ?open=${this.isCalendarOpen}
+                ?quiet=${this.quiet}
+                ?invalid=${this.invalid}
+                ?disabled=${this.disabled}
+                @click=${() => (this.isCalendarOpen = true)}
+                label=${this.labels.calendar}
```
spdev3000, line 441: pls add size here to allow auto-adapting to different main sizes also of the sp-picker-button:
```suggestion
                label=${this.labels.calendar}
                size=${this.size}
```

packages/date-time-picker/src/date-time-picker.css
```diff
@@ -0,0 +1,166 @@
...
+.input-content::-webkit-scrollbar {
+    display: none;
+    width: 0;
+    height: 0;
+}
+
+.literal-segment,
+.editable-segment {
+    display: inline-block;
+    height: 100%;
```
spdev3000, line 113: This css value needs to be removed (or adapted) in order to be fully aligned with `size="l"` of the DateTimePicker.

packages/date-time-picker/src/DateTimePicker.ts
```diff
@@ -0,0 +1,874 @@
...
+    public static override get styles(): CSSResultArray {
+        return [styles];
+    }
+
+    /**
+     * The selected date of the component. If defined, this also indicates where the calendar opens.
+     * If not, the calendar opens at the current month, and placeholder values are shown.
+     */
+    @property({ type: Object })
+    public value?: DateValue;
```
spdev3000, line 114: Is it possible to set those properties `value` ,  `min`, `max` (also) as JS Date object to avoid having outside type dependencies to `@internationalized/date` ?

mizgaionutalexandru: It is _possible_, but personally I'm against it since JS Date objects are not that reliable, especially when dealing with multiple timezones and daylight savings. It was agreed upon from the beginning to use [Adobe's internationalized/date library](https://react-spectrum.adobe.com/internationalized/date/index.html) instead.
Since we would use this library in the components themselves anyway, I don't see that much of a problem adding this as a dependency where the component is used, ensuring consistency across what consumers expect from the component and what it actually does.
This can be explored as a possibility but I don't see it as a blocker for release.
Are there any other problems you have encountered using these components?

PR #5004, opened 2025-01-03
build: token processing updates

castastrophe:
## Description

Main file for review: `scripts/generate-tokens-wrapper.js`

* Clean the `tools/styles/(tokens|tokens-v2)` folders before generating asssets
  * Why? This was not being done previously leading to outdated assets being retained in the library and causing invalid token output.
* Using require to fetch the packages from node_modules rather than specifying the location to better facilitate linking
* Remove the `custom-*.css` assets from import because that content is already concatenated in the `(dark|light|medium|large)-vars.css` files.
* Rename the `component-vars.css` in the `tools/styles/tokens*` directories to `system-theme-bridge.css` for improved clarity.

### Along for the ride

- Removed [chalk](https://www.npmjs.com/package/chalk) dependency in favor of the smaller [colors](https://www.npmjs.com/package/colors) package
- `tasks/process-spectrum.js` removed the chalk reference and as such, also had to resolve the lint errors in order to push up that change (removed unused variables reported and replaced ts-ignore with ts-expect-error)
- `tasks/css-tools.js`

## Motivation and context

By sourcing the system theming from the components, it lets us ensure the component theming provided matches the version of the component the SWC library supports. Previously, CSS was copying the theming assets from the component packages and bundling them in the token package to streamline their use by SWC. Now that we're making this update though, we can harden that pipeline to ensure the correct system mappings are being used and allowing SWC the freedom to rely on older versions of some components than others might be at while still advancing the tokens library.

## How has this been tested?

-   [ ] _Expect to see no changes in visual regression testing_
    1. Though what assets are being shipped for theming has changed, the content of those assets has not changed and so we expect to see no visual change.

## Types of changes

## Checklist

## Best practices


TarunAdobe (approved):
This looks good!! Thanks for working on it.

package.json
```diff
@@ -143,7 +143,7 @@
         "@webcomponents/webcomponentsjs": "^2.8.0",
         "alex": "^11.0.1",
         "cem-plugin-module-file-extensions": "^0.0.5",
-        "chalk": "^5.0.1",
```
Rajdeepc, line 146: is this removal intentional?

TarunAdobe: Yesss! We are not using this package anymore.

tasks/process-spectrum.js
```diff
@@ -385,7 +374,8 @@ async function processComponent(componentPath) {
                         }
                     }
                 });
-                // @ts-ignore
+
+                // @ts-expect-error - host is defined
```
rubencarvalho, line 378: if `host` is defined, why do we need to check for `if (host)`?

tasks/css-tools.js
```diff
@@ -26,22 +56,15 @@ const wrapCSSResult = (content) => {
...
+const licensePath = path.resolve(__dirname, '..', 'config', 'license.js');
+let header = '';
+if (fs.existsSync(licensePath)) {
+    header = fs.readFileSync(licensePath, 'utf8');
+    header = header.replace('<%= YEAR %>', new Date().getFullYear());
 }
 
-header = header.replace('<%= YEAR %>', new Date().getFullYear());
-
 export const processCSS = async (cssPath) => {
```
rubencarvalho, line 66: Feel free to ignore if it is out of scope, but could we check in `build-css.js` if there is any benefit if we parallelize calling this?

pseudo-code:
`Promise.all(cssPaths.map((cssPath) => processCSS(cssPath)));`

packages/tags/package.json
```diff
@@ -74,8 +74,8 @@
         "@spectrum-web-components/shared": "^1.0.1"
     },
     "devDependencies": {
-        "@spectrum-css/tag": "^10.0.0-s2-foundations.15",
-        "@spectrum-css/taggroup": "^6.0.0-s2-foundations.15"
+        "@spectrum-css/tag": "10.0.0-s2-foundations.15",
```
rubencarvalho, line 77: General question, as this seems to have changed on all components, why are we pinning the versions now?

rubencarvalho:
Is there a possibility of incorporating the changes from https://github.com/adobe/spectrum-web-components/pull/4999 into this PR? Or are these things that should not get addressed together?

scripts/generate-tokens-wrapper.js
```diff
@@ -135,124 +168,195 @@ const targetHost = (css) => {
...
+ * to be used by the sp-theme
+ * @param {'spectrum'|'spectrum-two'} spectrumVersion
+ * @param {Object} options
+ * @param {boolean} options.clean - Whether to clean the output directory before writing
+ * @returns {Promise}
  */
-const processTokens = (srcPath, tokensDir) => {
-    let css = fs.readFileSync(srcPath, 'utf8');
-    const fileName = srcPath.split(path.sep + 'css' + path.sep).at(-1);
+export async function generateTokensWrapper(
```
rubencarvalho, line 179: nit: I don’t have a clear idea of how long these operations usually take. Could we use console.time() to measure it? This might help us start understanding what takes how much time in our build chain.

scripts/generate-tokens-wrapper.js
```diff
@@ -135,124 +168,195 @@ const targetHost = (css) => {
...
-                        outputType,
-                        'global-vars.css'
-                    ),
-                    content
+    const promises = themeTokens.map(async (fileName) =>
+        fsp
+            .readFile(path.join(tokenSourcePath, fileName), 'utf8')
+            .then(async (content) => {
+                /* If the file is empty, return */
+                if (!content) {
```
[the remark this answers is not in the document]
castastrophe: It's honestly just a best practice to safely exit here but since we're using fg to fetch the files, they will exist. It's very unlikely we'd publish empty files from CSS but if we do, it shouldn't block SWC moving forward without it (because presumably VRTs in CSS passed even with that empty file).

scripts/generate-tokens-wrapper.js
castastrophe: I like the idea a lot - we do this in CSS for our script output as well. It's a bit out of scope here I feel because it's important when setting up the console times that you kick off the timer at the start and close the timer right when it ends but the current structure of this file would make that challenging (having to pass the console key around for instance to ensure it is reported accurately).

tasks/css-tools.js
castastrophe: Okay, going to flag this one as out-of-scope since we're not really making functional changes to css-tools at this time, just a little clean-up.

tasks/process-spectrum.js
castastrophe: Maybe a better comment is that the host variable _does_ exist because the ts error has to do with host not existing.

castastrophe: Turns out no comment needed at all!

.circleci/config.yml
```diff
@@ -14,7 +14,7 @@ parameters:
     # 3. Commit this change to the PR branch where the changes exist.
     current_golden_images_hash:
         type: string
-        default: a095ed1ef88686fdb24936dad763598ac7981193
+        default: 93ad92a2fc031db0277fee635434955e438e94a9
```
castastrophe, line 17: VRTs showed a fix in the checkbox which is now picking up the correct corner radius for Spectrum 2 🥳

tasks/css-tools.js
rubencarvalho, line 66: We have been on a "mission" of improving some inline documentation on our code, I'm wondering if we could add a JSDoc comment here, something along the lines of:

```suggestion
/**
 * Processes a CSS file using lightningcss, minifies it, and outputs a TypeScript module.
 * The output module includes license headers and wraps the CSS in a template literal.
 * 
 * @param {string} cssPath - Path to the CSS file to process
 * @returns {Promise<void>} A promise that resolves when processing is complete
 * 
 */
export const processCSS = async (cssPath) => {
```

rubencarvalho (approved):
Thank you for landing this! I left a small comment about a JSDoc, but it’s non-blocking.

PR #5012, opened 2025-01-09
fix(OverlayTrigger): delay update for overlay trigger to avoid infinite loop

jcmitch:
## Description

Overlay Trigger has potential to get into an infinite loop on initial render. This PR adds a delay using `requestAnimationFrame` to avoid this issue. As discussed in person this does not solve the root cause but finding that will require a deeper dive by the core SWC team. 

## Related issue(s)

- #4689 

## Motivation and context

Resolves an issue related to endless render loop.

## How has this been tested?

Manual testing in Adobe Express

-   [ ] Did it pass in iPad?

N/A

## Types of changes

## Checklist

## Best practices


jcmitch:
@rubencarvalho can you clarify the jump? I could see how this change could introduce a jump but I wasn't able to see one at all. If it is a problem we could also look to see if restricting the delay to initial render would still fix the issue. 

Here is a video of what I'm seeing comparing Storybook on main and this branch for the two stories you called out. 

https://github.com/user-attachments/assets/b955f1cd-fba8-48ef-9438-dd45de8dee2c

caseyisonit:
@jcmitch can you include some testing guidance that section seems to have been removed and only manually tested in Adobe Express. Also we will need to add automated testing to cover this in CI so we cover regression.

rubencarvalho:
This got solved by https://github.com/adobe/spectrum-web-components/pull/5046, issue https://github.com/adobe/spectrum-web-components/issues/4689 is closed.

Issue #4820, opened 2024-10-14
[Feat]: Allow using popover menu style on mobile


adixon-adobe:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-action-menu

### Description of the requested feature

We've been using the action menu component in a header on our site, and sometimes it only has a few items. We've found the mobile tray that shows up at the bottom to be confusing and often too far away from the button you clicked.

It'd be really nice to have an option to disable the tray -- I'd expect the normal popover menu to work well on mobile anyways.

### Implementation notes or ideas

Just a simple `disable-mobile-tray` attribute or something similar would do the trick. [The code](https://github.com/adobe/spectrum-web-components/blob/f86ef998b06299642b493cfff363b8a7eb14f4c9/packages/picker/src/Picker.ts#L621) looks like it'd be pretty simple to modify, but may also affect other components.

najikahalsema:
According to Spectrum guidelines, this behaviour is permitted. We just need to think about the API name, so a small RFC would be good here. That way we can abstract that pattern to other components if necessary.

adixon-adobe:
I've integrated this with SWC 1.1.1, but the [documented](https://opensource.adobe.com/spectrum-web-components/components/action-menu/api/#attributes%20and%20properties_force-popover) attribute name `force-popover` doesn't work. You need to do `forcepopover`. As long as this is still open, seems worth tracking. I know there are more general issues with touch/menus being fixed now that prevented me from checking this on the current doc site.

rubencarvalho:
This has been shipped under the `force-popover` property in `1.3.0.`.

PR #5041, opened 2025-01-21
feat(picker): add forcePopover property

rubencarvalho:
## Description

This PR introduces functionality for `sp-picker` and `sp-action-menu` to force a popover to render over the default tray. 

- introduces `forcePopover` attribute 

## Related issue(s)

- [Issue 4820](https://github.com/adobe/spectrum-web-components/issues/4820)

## Motivation and context

The `Action Menu` forces the menu to load in a tray at small viewports and consumers need a way to override that functionality so the menu is always a popover at any screen size. ActionMenu extends PickerBase thus affecting both ActionMenu and Picker components.

## How has this been tested?

-   [ ] _Test case 1_
    1. Go to the `Force Popover On Mobile` story in the Picker component
    2. Open Chrome DevTools (or equivalent).
    3. Toggle the Device Toolbar (the phone/tablet icon).
    4. Select a device preset (e.g. iPhone 12).
    5. Chrome will set user-agent strings, simulate touch, and adjust DPI.
    6. Reload the page
    7. Click the Picker 1 and see a tray
    8. Click the Picker 2 and see a popover

-   [ ] _Test case 2_
    1. Go to the `Force Popover On Mobile` story in the Action Menu component
    2. Open Chrome DevTools (or equivalent).
    3. Toggle the Device Toolbar (the phone/tablet icon).
    4. Select a device preset (e.g. iPhone 12).
    5. Chrome will set user-agent strings, simulate touch, and adjust DPI.
    6. Reload the page
    7. Click the Action Menu and see a tray

-   [ ] Did it pass in iPad?

## Types of changes

## Checklist

-   [ x] All new and existing tests passed.

## Best practices


Rajdeepc:
@rubencarvalho @caseyisonit Are we waiting on something to make this ready for review? Please let me know if I can help here?

packages/action-menu/test/action-menu-responsive.test.ts
```diff
@@ -0,0 +1,177 @@
...
+                            boundingRect.x + boundingRect.width / 2,
+                            boundingRect.y + boundingRect.height / 2,
+                        ],
+                    },
+                ],
+            });
+
+            await opened;
+
+            const tray = el.shadowRoot.querySelector('sp-tray');
```
rubencarvalho, line 92: Checking that tray exists is probably enough, but could we also verify that the popover is null?
```suggestion
            const tray = el.shadowRoot.querySelector('sp-tray');
            const popover = el.shadowRoot.querySelector('sp-popover');
            
            expect(tray).to.not.be.null;
            expect(popover).to.be.null; 

```

packages/action-menu/test/action-menu-responsive.test.ts
```diff
@@ -0,0 +1,177 @@
...
+            await nextFrame();
+            await nextFrame();
+
+            const opened = oneEvent(el, 'sp-opened');
+            el.open = true;
+            await opened;
+
+            const popover = el.shadowRoot.querySelector('sp-popover');
+
+            expect(popover).to.not.be.null;
```
rubencarvalho, line 109: And the same check for tray to be null here?

packages/action-menu/test/action-menu-responsive.test.ts
```diff
@@ -0,0 +1,177 @@
...
+            await setViewport({ width: 701, height: 640 });
+            // Allow viewport update to propagate.
+            await nextFrame();
+            await nextFrame();
+
+            const opened = oneEvent(el, 'sp-opened');
+            el.open = true;
+            await opened;
+
+            const popover = el.shadowRoot.querySelector('sp-popover');
```
rubencarvalho, line 172: And same here, could we check for tray to be null

packages/action-menu/test/action-menu-responsive.test.ts
```diff
@@ -0,0 +1,177 @@
...
+        beforeEach(async () => {
+            el = await actionMenuFixture({ forcePopover: true });
+            await elementUpdated(el);
+        });
+
+        it('is a Popover in mobile', async () => {
+            /**
+             * This is a hack to set the `isMobile` property to true so that we can test the MobileController
+             */
+            el.isMobile.matches = true;
```
rubencarvalho, line 123: I can see that this approach works, but it might not be the ideal way of doing this. Directly modifying properties like `isMobile.matches` can lead to hard-to-debug tests and potential side effects. For example, this manual assignment carries over to other tests since it isn’t reset in an `afterEach`, nor is it being properly stubbed (e.g., using a library like sinon).

I’m not entirely sure of the inner workings of the `MatchMediaController` to determine what is retriggering the logic and resetting `isMobile.matches`. In the line 132 below, where we use `await setViewport({ width: 360, height: 640 });`, it would make sense to me that it would turn `isMobile.matches` to `false` because it would react to the viewport change and attempt to match `hover` and `pointer: coarse`, which like you commented, are not present in this test setup - so it is not 100% clear to me why it works 🤷 
We have a bigger "mobile testing strategy" planned so I won't make this blocking, but I still wanted to call it out!

packages/picker/README.md
```diff
@@ -459,6 +459,28 @@ When in pending state, `<sp-picker>` elements will not respond to click events a
 </sp-picker>
 ```
 
+## Force Popover on Mobile Devices
+
+When on a mobile device, `<sp-picker>` elements will render an `<sp-tray>` to the bottom of the viewport by default. To override this functionality so that the picker remains an `<sp-popover>` in mobile, you can simply add the attribute `forcePopover` to the `<sp-picker>`. To see this functionality in action, load this page from your mobile device or use Chrome DevTools (or equivalent) and select a mobile device once the Device Toolbar (the phone/tablet icon) is active.
```
rubencarvalho, line 464: nit: This is a very nice technical explanation, but could we also provide guidance focusing on the UX aspects of it?
e.g., taking some inspiration from https://spectrum.adobe.com/page/tray/#Trays-or-popovers? and https://spectrum.adobe.com/page/popover/#Popovers-or-trays?

packages/picker/src/Picker.ts
```diff
@@ -96,6 +96,14 @@ export class PickerBase extends SizedMixin(Focusable, { noDefaultSize: true }) {
     @property({ type: Boolean, reflect: true })
     public invalid = false;
 
+    /**
+     * Forces the Picker to render as a popover on mobile instead of a tray.
+     *
+     * @memberof PickerBase
```
rubencarvalho, line 102: Using `@memberof` is a nice touch! While our current documentation strategy doesn’t currently pick this up, it’s something we could consider leveraging in the future.

packages/picker/test/picker-responsive.test.ts
```diff
@@ -108,4 +114,69 @@ describe('Picker, responsive', () => {
...
+            await setViewport({ width: 701, height: 640 });
+            // Allow viewport update to propagate.
+            await nextFrame();
+            await nextFrame();
+
+            const opened = oneEvent(el, 'sp-opened');
+            el.open = true;
+            await opened;
+
+            const popover = el.shadowRoot.querySelector('sp-popover');
```
rubencarvalho, line 177: last nit: test for tray 😄

packages/action-menu/test/action-menu-responsive.test.ts
Rajdeepc: Currently there is no way to test for mobile emulation in playwright. I already have a request open to the playwright team which I think is not prioritized at their end. I agree once we move to a broader work on the mobile we need to rethink how we should start mobile emulations.

packages/action-menu/README.md
```diff
@@ -272,13 +272,26 @@ When `selects` is set to `single`, the `<sp-action-menu>` element will maintain
 </sp-action-menu>
 ```
 
-## Accessibility
+## Force Popover on Mobile Devices
 
-An `<sp-action-menu>` parent will ensure that the internal `<sp-menu>` features a role of `listbox` and contains children with the role `option`. Upon focusing the `<sp-action-menu>` using `ArrowDown` will also open the menu while throwing focus into first selected (or unselected when none are selected) menu item to assist in selecting of a new value.
+When on a mobile device, `<sp-action-menu>` elements will render an `<sp-tray>` to the bottom of the viewport by default. To override this functionality so that the picker remains an `<sp-popover>` in mobile, you can simply add the attribute `forcePopover` to the `<sp-action-menu>`. To see this functionality in action, load this page from your mobile device or use Chrome DevTools (or equivalent) and select a mobile device once the Device Toolbar (the phone/tablet icon) is active.
```
Rajdeepc, line 277: This is a bit contradicting to what the API contract is all about. We are not overriding functionality but rather we are giving the users to choose the visual ui of popover in mobile devices. Rather I would say now there is an option to switch between different modes of popover UI in mobile. Tray or Popover.
Let me know your thoughts!

rubencarvalho: This goes in hand with my comment above! 👍

packages/action-menu/test/action-menu-responsive.test.ts
caseyisonit: yeah so this is the rough workaround Tarun put in place. It works to get the right functionality to test but like the reasons Rajdeep stated, the query selector with `(hover: none) and (pointer: coarse)` makes it impossible to check without an emulator that's why you have to use dev tools to see this in storybook too.

caseyisonit (approved):

Rajdeepc (approved):

PR #5046, opened 2025-01-27
chore: always render sp-overlay

rubencarvalho:
## Related issue(s)

-

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_

    1. Go here
    2. Do this

-   [ ] Did it pass in iPad?

## Types of changes

## Checklist

## Best practices


packages/overlay/src/OverlayTrigger.ts
```diff
@@ -27,29 +27,59 @@ import type { Placement } from '@floating-ui/dom';
...
+     * we can avoid:
+     * 1. Extra renders from unnecessary slot reparenting
+     * 2. Potential infinite render loops during content detection
+     * 3. Race conditions during slot assignment
+     *
+     * By only returning overlay wrappers for explicitly declared content types,
+     * we minimize unecessary DOM nodes, operations and ensure a more stable rendering behavior.
+     */
+    @property({ attribute: 'triggered-by' })
+    triggeredBy?: TriggeredByType;
```
Rajdeepc, line 82: This will not surface up in the API tab. You need to make it public!

packages/overlay/overlay-trigger.md
```diff
@@ -117,6 +117,40 @@ The delivery of hover content can be customized via the `placement` attribute. H
 </overlay-trigger>
 ```
 
+### Performance optimization
```
Rajdeepc, line 120: I don't know whether this should be called Perf Optimization but we can specify or educate the users why we should use this?

packages/overlay/src/OverlayTrigger.ts
rubencarvalho: fixed!

Rajdeepc (approved):

PR #5060, opened 2025-02-03
fix(picker): stop the click events from reaching the elements below picker-tray

### LINKED ISSUE #5042: [Bug]: (iOS) Tapping on a button in the sp-menu will also tap on a button directly underneath it
### Code of conduct

### Impacted component(s)

sp-menu, sp-picker

### Expected behavior

While using the `sp-picker` in iOS Safari, tapping on a button in the `sp-menu` should **not** also tap on a button directly underneath it. This seems related to a [previously fixed issue](https://github.com/adobe/spectrum-web-components/issues/4461); however, checking out version SWC 47.0.1 does not resolve the issue in the Express app and mobile web.

### Actual behavior

While using the `sp-picker` in iOS Safari, tapping on a button in the `sp-menu` will also tap on a button directly underneath it.

### Screenshots

Two different examples of this problem occurring:

Image

https://github.com/user-attachments/assets/9114ae2c-51ee-42e0-9df3-4f0e5971a97f

Safari

**On Express mobile app:**
1. Open Editor
2. Go to "Share" to share any asset. This will open up the Share Sheet dialog.
3. Input username and tap Return
4. Click on the permission dropdown [Can edit/comment].
5. Try to tap on the "Can comment" option at the same location as 'Cancel' or 'Invite to edit/comment' button text are located.
6. Observed - the click event is propagated to the buttons ["Cancel" and "invite to edit"]behind the panel.

**On mobile web, which may be easier to troubleshoot and test for developers and QE:**
1. Install xcode > Settings > Platform > Install iOS 16.4 Simulator
2. Open Simulator for iPhone 14 Pro
3. Navigate to new.express.adobe.com (prod)
4. Go to "Share" to share any asset. This will open up the Share Sheet dialog.
5. Input a username you'd like to share with, and tap return
6. Click on the permission dropdown (Can edit/comment)
7. Try to tap on the "Can comment" option at the same location as overlaying the button text, "Cancel"
8. Notice the Cancel button is clicked instead of the "Can Comment" and the overlay is closed.

Note: To reproduce the issue, the clicking of the overlaying button text needs to align directly above the button text underneath it (not the button itself, but the text within the button).

We were able to implement a temporary workaround where we addEventListener for `touchstart` and preventDefault and stopPropagation whenever the sp-menu is visible. Once it is closed, we removeEventListener.

_No response_

### Severity

None

_No response_

TarunAdobe:
## Description

Stop the click events to reach the element below tray when we open a picker in iOS <17

## Related issue(s)

- fixes #5042

## Motivation and context

* Prevents the event leaking out of the dialog to the element below it.
* This is a workaround for a bug in iOS <17 where the event leaks out of the dialog to the element below it. 

## How has this been tested?

- Go [here](https://picker-background-click--spectrum-web-components.netlify.app/storybook/?path=/story/picker--background-click-test&globals=color:dark) in iOS<17 
    - [ ] Open the picker and make a selection on the tray.
    - [ ] There should not be any console log from the button getting clicked.

-   Manually Tested. See the attached Video below for before and After:

### Before
[screen-capture.webm](https://github.com/user-attachments/assets/57520995-d025-4042-8c00-988df3519ddf)

### After
[screen-capture (1).webm](https://github.com/user-attachments/assets/d5b9d014-e866-4ece-b496-f260a780247b)

-   [x] Did it pass in iPad?

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/overlay/src/OverlayDialog.ts
+++ b/packages/overlay/src/OverlayDialog.ts
@@ -54,6 +54,7 @@ export function OverlayDialog<T extends Constructor<AbstractOverlay>>(
                     if (!targetOpenState) {
                         const close = (): void => {
                             el.removeEventListener('close', close);
+                            this.cleanupIOSEventManagement();
                             finish(el, index);
                         };
                         el.addEventListener('close', close);
@@ -88,6 +89,7 @@ export function OverlayDialog<T extends Constructor<AbstractOverlay>>(
                         return;
                     }
                     this.dialogEl.showModal();
+                    this.setupIOSEventManagement();
                 };
             const finish = (el: OpenableElement, index: number) => (): void => {
                 if (this.open !== targetOpenState) {
--- a/packages/overlay/src/AbstractOverlay.ts
+++ b/packages/overlay/src/AbstractOverlay.ts
@@ -338,4 +338,38 @@ export class AbstractOverlay extends SpectrumElement {
         overlay.placement = options.placement;
         overlay.willPreventClose = !!options.notImmediatelyClosable;
     }
+
+    private iosEventPropagationController?: AbortController;
+
+    protected setupIOSEventManagement(): void {
+        // This is a workaround for a bug in iOS <17 where the event leaks out of the dialog to the element below it.
+        // Issue - https://github.com/adobe/spectrum-web-components/issues/5042
+        this.iosEventPropagationController = new AbortController();
+        this.dialogEl.addEventListener(
+            'pointerup',
+            (event) => event.stopPropagation(),
+            {
+                capture: true,
+                signal: this.iosEventPropagationController.signal,
+            }
+        );
+        this.dialogEl.addEventListener(
+            'touchend',
+            (event) => event.stopPropagation(),
+            {
+                capture: true,
+                signal: this.iosEventPropagationController.signal,
+            }
+        );
+    }
+
+    protected cleanupIOSEventManagement(): void {
+        this.iosEventPropagationController?.abort();
+        this.iosEventPropagationController = undefined;
+    }
+
+    override disconnectedCallback(): void {
+        this.cleanupIOSEventManagement();
+        super.disconnectedCallback();
+    }
 }
--- a/packages/picker/stories/picker.stories.ts
+++ b/packages/picker/stories/picker.stories.ts
@@ -694,3 +694,29 @@ custom.args = {
     open: true,
 };
 custom.decorators = [isOverlayOpen];
+
+export const BackgroundClickTest = (): TemplateResult => {
+    return html`
+        <div style="display: flex; flex-direction: column;">
+            <sp-picker size="l">
+                <sp-menu-item value="option-1">Deselect</sp-menu-item>
+                <sp-menu-item value="option-2">Select Inverse</sp-menu-item>
+            </sp-picker>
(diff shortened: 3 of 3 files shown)


TarunAdobe:
Since this issue happens only in specific version(s) of iOS combined with the fact that we don't have a good testing infrastructure setup for mobile - I couldn't figure out a way to write tests for this which won't pass always. So I attached a before and after video in the description. 
Apart from this I did test and verified manually as well as through our tests that everything else related to the menu and picker behaviour remains exactly the same!

packages/overlay/src/OverlayDialog.ts
```diff
@@ -163,6 +163,16 @@ export function OverlayDialog<T extends Constructor<AbstractOverlay>>(
                     finish(el, index)
                 );
             });
+
+            /**
+             * Prevents the event leaking out of the dialog to the element below it.
+             * This is a workaround for a bug in iOS <17 where the event leaks out of the dialog to the element below it. 
+             * Issue - https://github.com/adobe/spectrum-web-components/issues/5042
+             */
+            this.dialogEl.addEventListener('pointerup', (event) => {
```
rubencarvalho, line 172: is `pointerup` enough? Have we considered adding `touchend`?

packages/overlay/src/OverlayDialog.ts
rubencarvalho, line 1: Does `OverlayDialog` only get instantiated once, or is there a chance that the dialog may be re-instantiated? If so, should we define the event handler as a class method and attach/detach it?

packages/overlay/src/OverlayDialog.ts
TarunAdobe: Either of the two `pointerup` or `touchend` was working fine and similar to the other one in simulators... Didn't test it on a real device. Should I stop the propagation for both?

packages/overlay/src/OverlayDialog.ts
TarunAdobe: Yess! Thanks for the catch. Totally slipped my mind that we need to avoid attaching the event listener multiple times and clean it properly. Will update the code to do that. 
I was thinking of having the logic of attaching and removing the listeners abstracted out and then will call that function to attach the listener when the dialog opens and remove it when it closes?

packages/overlay/src/OverlayDialog.ts
rubencarvalho: I’m wondering if there’s any downside to having `touchend` as a “safe fallback”... I don’t think it should be necessary, but without real device testing, it is hard for me to say 🤷

TarunAdobe: I don't particularly see any other downside than the fact that in the worst case we're adding 1 extra event listener....so

rubencarvalho (approved):
LGTM 🚢

caseyisonit:
@TarunAdobe should we add in a test to confirm this behavior?

TarunAdobe:
> @TarunAdobe should we add in a test to confirm this behavior? or is this a device emulation issue?

I made a comment on this PR abt this that got lost between all the bot comments but TLDR; Since this issue happens only in specific version(s) of iOS combined with the fact that we don't have a good testing infrastructure setup for mobile - I couldn't figure out a way to write tests for this which always end up passing. And so it'd be useless :(

PR #5075, opened 2025-02-06
DRAFT: Update Storybook

graynorton:
*Note*: work has been handed off to @castastrophe; see updates below:

## Work in progress summary

Goal: a minimal migration to latest Storybook leveraging Vite.

### Completed

- Migrated over assets and styles from Spectrum CSS which leverages the same Vite tooling
- Updated dependencies in package.json
- Added vite-plugin-cem which generates the custom element manifest on the fly
- Temporarily removed unused add-ons (badges)
- Set up the vite config to mimic the original tooling as much as possible (rollup is used in Vite so those ported nicely)
- Initial ESBuild set-up in Storybook, updated core build scripts to be extensible for this purpose

### Todo

- Local path alias'ing
- Migrating the webpack configurations as necessary (Vite leverages rollup and _not_ webpack)
- Configuring the esbuild to process the assets for Storybook correctly

### Notes to self

- The '@rollup/plugin-json plugin should no longer be needed (JSON support built into Vite)
- Need to remove workspace dependencies in the Spectrum CSS preview package so we can just add that as a dependency and use those doc blocks directly

## Description

Update Storybook to latest, use `@storybook/web-components-vite`

## Related issue(s)

-

## Motivation and context

Having issues firing Chromatic on the outdated version of Storybook; attempting this upgrade in order allow Chromatic VRT runs for SWC.

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_

    1. Go here
    2. Do this

-   [ ] Did it pass in iPad?

## Types of changes

## Checklist

## Best practices


.storybook/main.js
```diff
@@ -0,0 +1,51 @@
...
+governing permissions and limitations under the License.
+*/
+const config = {
+    stories: [
+        '../packages/*/stories/*.stories.js',
+        '../tools/*/stories/*.stories.js',
+    ],
+
+    addons: [
+        getAbsolutePath("@storybook/addon-links"),
```
castastrophe, line 20:
```suggestion
        "@storybook/addon-links",
```

We should be able to reference the package by string name rather than the full path. This prevents us having to maintain a hardcoded pathing infrastructure.

.storybook/main.js
```diff
@@ -0,0 +1,51 @@
...
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+const config = {
+    stories: [
+        '../packages/*/stories/*.stories.js',
+        '../tools/*/stories/*.stories.js',
+    ],
```
castastrophe, line 17:
```suggestion
    stories: [
		{
			directory: '../packages',
			files: '*/stories/*.@(stories.js|mdx)',
			titlePrefix: 'Components',
		},
		{
			directory: "./tools",
			files: '*/stories/*.@(stories.js|mdx)',
			titlePrefix: 'Tools',
		},
    ],
```

Would it be helpful to sort the imports into the Component & Tools categories?

.storybook/main.js
```diff
@@ -0,0 +1,51 @@
...
+        // Conditionally add the addon-designs addon temporarily
+        // https://storybook.js.org/addons/@storybook/addon-designs/
+        ...(process.env.NODE_ENV === 'development'
+            ? ['@storybook/addon-designs']
+            : []),
+        getAbsolutePath("@geometricpanda/storybook-addon-badges"),
+        '@chromatic-com/storybook'
+    ],
+
+    framework: "@storybook/web-components-vite",
```
castastrophe, line 32: I noticed we don't have `rootDir`, `staticDirs`, `core`, or any `viteFinal` settings here. This is what we use in CSS:

```
	core: {
		disableTelemetry: true,
		disableWhatsNewNotifications: true,
		builder: "@storybook/builder-vite",
	},
```

package.json
```diff
@@ -1,437 +1,437 @@
 {
-    "name": "@adobe/spectrum-web-components",
```
castastrophe, line 2: Do you know why the spacing might have changed in this asset?

castastrophe (commented):
Ran through the code updates and this is looking great so far. I'll pull it down next and see if there are any rendering issues.

package.json
graynorton: Good catch—I have no idea, but fixed now!

.storybook/main.js
graynorton: Yeah, I have no idea why `getAbsolutePath()` was being used here, but I went ahead and removed it.

.storybook/main.js
graynorton: Sure! I went ahead and did this.

.storybook/main.js
graynorton: @castastrophe `disableTelemetry` and `disableWhatsNewNotifications` definitely sound good!

My understanding was that setting the framework to `@storybook/web-components-vite` caused `builder-vite` to be used by default. Is that not the case?

castastrophe: I'm sure you're right, it's not explicit in the docs so I followed instructions from here as well just to be sure I leveraged the right tooling: https://storybook.js.org/docs/builders/vite

castastrophe: Are there any Vite configs you think could be relevant or useful for the web components?

package.json
```diff
@@ -57,7 +57,7 @@
         "prepare": "husky",
         "prepublish:react": "yarn build:react && sed -i \"\" \"s/react/# react/g\" .gitignore && git commit -am \"Commit React Wrappers\" --no-verify",
         "prestorybook": "wireit",
-        "prestorybook:build": "cem analyze --outdir storybook/",
+        "prestorybook:build": "cem analyze --outdir .storybook/",
```
castastrophe, line 60: Pre commands should no longer be automatically working in yarn 4 but I think this isn't erroring yet because wireit is holding it together. Not something we need to consider right now but it might worth clarifying and combining required commands so we don't try to run `yarn storybook:build` for example and expect cem to run (but it won't because only `yarn storybook` is a wireit command I believe.

.storybook/main.js
graynorton: Nothing is top of mind; I don’t consider myself to be an expert in either Vite or Storybook, TBH, but I think unless there are things you know of that you’d like to add, we can plan to circle back and make improvements later.

pfulton:
@castastrophe @graynorton just dropping in on this PR to ask (or strongly suggest!) that, for now, we not run the Chromatic GH Action _unless_ the `run_vrt` label is added. Meaning, that in a GitHub PR context, it would _not_ run by default, and you would need to explicitly add the label to invoke it. We'll want to think about running it by default at some future time, but we should err on the side of on-demand running right now until we get a handle on how many snapshots it'll be taking (there are a lot of stories in this Storybook, and 1 story === (at least) 1 snapshot.

Always happy to discuss further in Slack or on a call :-)

package.json
caseyisonit: same brain wavelengths! also to note we added the `enableScripts: true` to the `yarnrc.yml` to enforce the old behavior of running pre and post :) we added an @todo in a comment above it. this came up as in issue in the 1.1.0 release.

caseyisonit:
updated the circleci config to resolve preview PRs

castastrophe:
> @castastrophe @graynorton just dropping in on this PR to ask (or strongly suggest!) that, for now, we not run the Chromatic GH Action _unless_ the `run_vrt` label is added. Meaning, that in a GitHub PR context, it would _not_ run by default, and you would need to explicitly add the label to invoke it. We'll want to think about running it by default at some future time, but we should err on the side of on-demand running right now until we get a handle on how many snapshots it'll be taking (there are a lot of stories in this Storybook, and 1 story === (at least) 1 snapshot.

Yes to all! This is set up as a  workflow_dispatch so it will only run if you explicitly run it from GitHub Actions and it doesn't even run on a label or tag.

Issue #5081, opened 2025-02-07
[Bug]: Picker does not keep up with changes to icons in its menu items


estollnitz:
### Code of conduct

- [x] I agree to follow this project's code of conduct.

### Impacted component(s)

Picker

### Expected behavior

When there's a change to the `src` attribute of an `sp-icon` used within a menu item, the selected item displayed by the picker should update to reflect the modified icon.

### Actual behavior

Picker displays a stale icon associated with the selected menu item when the icon has changed to a new `src` value. The menu items shown in the popover are correct; only the copy shown in the picker button itself is stale.

### Screenshots

![Image](https://github.com/user-attachments/assets/48d2f1f3-d6f9-4f90-800b-36274acb2c03)

### What browsers are you seeing the problem in?

Firefox, Chrome, Safari, Microsoft Edge

### How can we reproduce this issue?

See the sample code below. It displays repro steps that are easy to follow.

### Sample code or abstract reproduction which illustrates the problem

https://studio.webcomponents.dev/edit/wF3pii8z9xOpdkyufNqu/src/index.ts?p=stories

### Severity

SEV 3

estollnitz:
I think the issue stems from MenuItem caching a clone of its icon the first time that Picker requests the `itemChildren` getter here: https://github.com/adobe/spectrum-web-components/blob/main/packages/menu/src/MenuItem.ts#L196. The result gets put into MenuItem's `_itemChildren` cache, and that cache is not invalidated when I update the `src` attribute of an `sp-icon`. The cache is only invalidated when the MutationController sees a change to `characterData` or `childList` or `subtree` and then calls `breakItemChildrenCache` here: https://github.com/adobe/spectrum-web-components/blob/main/packages/menu/src/MenuItem.ts#L218.

estollnitz:
I confirmed that calling the `breakItemChildrenCache` method of each MenuItem whenever I update its icon will indeed cause the Picker to display the correct icon. I wrote this hacky workaround to call the protected `breakItemChildrenCache` method:
```
    const menuItems = this.shadowRoot.querySelectorAll('sp-menu-item')
    for (const menuItem of menuItems) {
      (menuItem as any).breakItemChildrenCache();
    }
```
Of course, this is just a workaround — I should not have to call a protected method of a class I don't own.

PR #5088, opened 2025-02-12
fix(picker): update picker when menu item icons change

estollnitz:
## Description

Picker displays a clone of the icon and text from the selected MenuItem. That clone is updated only when selection changes or when certain changes occur in the MenuItems themselves. Before the changes in this PR, the cloned icon failed to update if the selected MenuItem contained an `<sp-icon>` element whose `src` attribute was updated dynamically.

## Related issue(s)

#5081: Picker does not keep up with changes to icons in its menu items

## Motivation and context

Dynamically updating the `src` attribute of an `<sp-icon>` element is certainly a corner case. However, it's a useful scenario when the icons are (for example) displaying thumbnails of layers in an Illustrator file, and those thumbnails need to update as the user edits their vector artwork. Just as the text displayed in the Picker needs to match the text in the selected MenuItem, the icons need to match as well.

## How has this been tested?

1. I added a Storybook story within the Picker category (in `picker.stories.ts`), called "Dynamic Icons," that illustrates the problem. Without the fix, clicking on the "Change icons" button causes the Picker icon to get out of sync with the selected MenuItem icon. With the fix, the icons stay in sync.
2. I added a unit test to `packages/picker/test/index.ts`, based on the same story. It fails without the fix, and passes with the fix — using Chromium, Firefox, and Webkit.

-   [ ] Did it pass in iPad?

I don't know how to test on mobile or iPad, but I'm hopeful that the same tests would work.

## Types of changes

## Checklist

## Best practices


Rajdeepc:
@estollnitz This does solve your problem but since we are dealing with a composite API so we would need to test this in many areas. It might not work with this approach.

Rajdeepc (approved):

Issue #4775, opened 2024-09-19
[Feat]: Extend ComboboxOption to support disabled state of menu items


spdev3000:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

Combobox

### Description of the requested feature

As a user I'd like to disable certain items in a combobox so I can see that those values/items exists, but aren't selectable.

### Mockups or screenshots

<img width="199" alt="Screenshot 2024-09-19 at 18 17 11" src="https://github.com/user-attachments/assets/397666bf-3108-47f3-96a2-abceb3473369">

### Implementation notes or ideas

Extend Combobox options with boolean param: `disabled` , see: https://github.com/adobe/spectrum-web-components/blob/a2b6bea7142930d8143d9ca887e2c56ffa71a669/packages/combobox/src/Combobox.ts#L45
such as
```
export type ComboboxOption = {
    value: string;
    itemText: string;
    disabled: boolean;
};
``` 

and map that disabled state to related menu-item. Check if selection process will still work and disabled menu item won't be selected when user presses keyUp/keyDown.

PR #5104, opened 2025-02-19
feat(combobox): allow support for disabled items

TarunAdobe:
## Description

Allow combobox to support having `disabled` items.

## Related issue(s)

- #4775 

## Motivation and context

As a user I'd like to disable certain items in a combobox so I can see that those values/items exists, but aren't selectable.

## How has this been tested?

-   [ ] Go [here](https://ttomar-combobox-disabled-items--spectrumwc.netlify.app/storybook/?path=/story/combobox--has-disabled-items)
    1. Open the combobox
    2. Try selecting the disabled item using keyboard or mouse
 

## Types of changes

## Checklist

## Best practices


packages/combobox/src/Combobox.ts
```diff
@@ -215,20 +216,36 @@ export class Combobox extends Textfield {
         const activeIndex = !this.activeDescendant
             ? -1
             : this.availableOptions.indexOf(this.activeDescendant);
-        const nextActiveIndex =
-            (this.availableOptions.length + activeIndex + 1) %
-            this.availableOptions.length;
-        this.activeDescendant = this.availableOptions[nextActiveIndex];
+        let nextActiveIndex = activeIndex;
+        do {
```
Rajdeepc, line 220: Before looping, can we check if all items are disabled and exit early:
```js
if (this.availableOptions.every(option => option.disabled)) {
    return; 
}
```

packages/combobox/src/Combobox.ts
```diff
@@ -215,20 +216,36 @@ export class Combobox extends Textfield {
...
+            nextActiveIndex =
+                (this.availableOptions.length + nextActiveIndex + 1) %
+                this.availableOptions.length;
+            // Break if we've checked all options to avoid infinite loop
+            if (nextActiveIndex === activeIndex) break;
```
Rajdeepc, line 225: This was needed to prevent an infinite state. Thanks

packages/combobox/src/Combobox.ts
```diff
@@ -215,20 +216,36 @@ export class Combobox extends Textfield {
...
     public activatePreviousDescendant(): void {
         const activeIndex = !this.activeDescendant
             ? 0
             : this.availableOptions.indexOf(this.activeDescendant);
-        const previousActiveIndex =
-            (this.availableOptions.length + activeIndex - 1) %
-            this.availableOptions.length;
-        this.activeDescendant = this.availableOptions[previousActiveIndex];
+        let previousActiveIndex = activeIndex;
+        do {
```
Rajdeepc, line 238: Same as above we can exit early if all items are disabled.

Rajdeepc (changes requested):

packages/combobox/src/Combobox.ts
TarunAdobe: if (this.availableOptions.every(option => option.disabled)) {
    return; 
}

this would be equivalent of looping no?

Rajdeepc: Yes! Good question. `.every` is a single pass operation so if it finds one single non-disable items it will break away from the loop whereas `do while` loop will move to the next item till it completes the entire loop. 
`.every` is O(n) complexity :)
So we should keep both.

packages/combobox/src/Combobox.ts
```diff
@@ -587,6 +605,11 @@ export class Combobox extends Textfield {
             }
         }
         if (changed.has('options') || changed.has('optionEls')) {
+            // if all options are disabled, set combobox to disabled
+            if (this.options?.every((option) => option.disabled)) {
+                this.disabled = true;
+            }
+
```
TarunAdobe, line 612: @Rajdeepc I added this... Now we don't need to check the same thing during the navigation.

rubencarvalho: I am not sure we should do this or recommend customers do it... 🤔

Rajdeepc: If all options are disabled and if we allow the customers to disable the Combo-box we need to think of these following cases.
1. Users may expect to open the menu to understand why they can’t select anything.
2. Disabling the entire combobox might not be what customers expect
Let's check with design on this!

Rajdeepc: Nit: We can come back on this and move forward with the above solutions @TarunAdobe. Let's put a note out to the design team and add a follow up PR to tackle this?

packages/combobox/src/Combobox.ts
```diff
@@ -110,7 +111,7 @@ export class Combobox extends Textfield {
      * An array of options to present in the Menu provided while typing into the input
      */
     @property({ type: Array })
-    public options?: ComboboxOption[];
+    public options?: (ComboboxOption | MenuItem)[];
```
Rajdeepc, line 114: This will not be expanded by jsdoc to let users know what are ComboboxOption.
Screenshot 2025-02-20 at 7 01 11 PM

rubencarvalho: If there is too much effort to fix this, I don't think this should be blocking. This is a common issue across our codebase that will need to be tackled holistically.

TarunAdobe: I actually have no idea how to fix this bwahaha. so

packages/combobox/test/combobox.test.ts
```diff
@@ -917,6 +918,85 @@ describe('Combobox', () => {
...
+
+            const opened = oneEvent(el, 'sp-opened');
+            el.click();
+            await opened;
+
+            const menuItems = el.shadowRoot.querySelectorAll('sp-menu-item');
+            expect(menuItems[2].disabled).to.be.true;
+        });
+        it('disabled items should not be focusable using keyboard', async () => {
+            const el = await withDisabledItemsFixture();
```
Rajdeepc, line 934: dont initialize the el in all the test cases. You can globally define `el` and use it in the below test cases. Saves you lines of code.

Rajdeepc (approved):

Issue #5117, opened 2025-02-24
[Bug][Safari]: Slider label hides when slider triggers related overlay and pointer is moved out


spdev3000:
### Code of conduct

- [x] I agree to follow this project's code of conduct.

### Impacted component(s)

Slider, Overlay

### Expected behavior

Slider label should stay visible when user hovers over slider and triggers related sp-overlay

### Actual behavior

Slider label dismisses when user moves pointer out of slider and that closes related overlay

### Screenshots

<img width="421" alt="Image" src="https://github.com/user-attachments/assets/048e68f0-7ed5-4cb0-85b2-aec896314957" />

### What browsers are you seeing the problem in?

Safari

### How can we reproduce this issue?

1. Go to https://studio.webcomponents.dev/edit/p8NIOWGap5gajrjn3EDH/src/index.ts?p=stories in Safari 18.3
2. Hover over slider 
3. Overlay appears, now move pointer out of slider
4. See Slider label hides

### Severity

SEV 3

najikahalsema:
SWC-694

PR #5118, opened 2025-02-24
fix(slider): make label not disappear when using overlays in Safari

rubencarvalho:
## Description

This PR addresses a Safari-specific issue where slider labels would disappear when a related hover-overlay is triggered and the user moves the pointer away from the slider. The problem was occurring both when using the `label` attribute and when using the default slot for text labels.

The fix ensures that the label element remain visible in the DOM regardless of hover state changes, which prevents Safari's hover state behavior from causing the disappearing label issue.

Additionally, this PR marks the default slot as deprecated, because we don't need two APIs for doing the same thing, so we want to encourage users to use the label property instead of the slot for text labels.

## Related issue(s)

- [SWC-694](https://jira.corp.adobe.com/browse/SWC-694)
- Fixes https://github.com/adobe/spectrum-web-components/issues/5117

## Motivation and context

This issue appears to be related to how Safari handles projecting slot content across shadow DOM boundaries (the sp-slider slot inside the sp-field-label) when combined with overlays creating different stacking contexts. Safari seems particularly sensitive to these interactions compared to other browsers. 🤷 

I explored an alternative fix specifically for the slotted content case, which involved moving the slot outside of sp-field-label, hiding its visibility, copying its assigned node content to a state variable, and then rendering that variable instead. However, this approach was overly complex for what's essentially a niche use case that we're deprecating anyway. The current solution is much simpler while still fixing the core issue.

## How has this been tested?

I added unit test for the deprecation warning, added a storybook story that demonstrates the fixed behavior in Safari and verified that the label remains visible in Safari when hovering over slider and moving pointer away.

Also tested across Firefox and Chrome to ensure consistent behavior.

In Safari:
-   [ ] _Go to this new [With Popover story](https://ruben-slider-label-fix--spectrumwc.netlify.app/storybook/?path=/story/slider--with-popover)_
    1. Hover over the multiple Sliders
    2. Verify how "Label in slot" disappears (not fixed, reason described above)
    3. Verify how "Label in attribute" does not disappear (fixed)

-   [x] Did it pass in iPad?

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/slider/src/Slider.ts
+++ b/packages/slider/src/Slider.ts
@@ -44,7 +44,7 @@ export const variants = ['filled', 'ramp', 'range', 'tick'];
 /**
  * @element sp-slider
  *
- * @slot - text label for the Slider
+ * @slot - @deprecated Text label for the Slider. Use the `label` property instead.
  * @slot handle - optionally accepts two or more sp-slider-handle elements
  */
 export class Slider extends SizedMixin(ObserveSlotText(SliderHandle, ''), {
@@ -231,6 +231,16 @@ export class Slider extends SizedMixin(ObserveSlotText(SliderHandle, ''), {
     public override connectedCallback(): void {
         super.connectedCallback();
         this.handleController.hostConnected();
+
+        // Deprecation warning for default slot when content is provided
+        if (window.__swc.DEBUG && this.textContent?.trim()) {
+            window.__swc.warn(
+                this,
+                `The default slot for text label in <${this.localName}> has been deprecated and will be removed in a future release. Use the "label" property instead.`,
+                'https://opensource.adobe.com/spectrum-web-components/components/slider/',
+                { level: 'deprecation' }
+            );
+        }
     }
 
     public override disconnectedCallback(): void {
--- a/packages/slider/src/Slider.ts
+++ b/packages/slider/src/Slider.ts
@@ -265,8 +265,8 @@ export class Slider extends SizedMixin(ObserveSlotText(SliderHandle, ''), {
                     @click=${this.handleLabelClick}
                     size=${this.size}
                 >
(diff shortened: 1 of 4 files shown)


spdev3000 (approved):
lgtm

packages/slider/test/slider.test.ts
```diff
@@ -1378,100 +1378,6 @@ describe('Slider', () => {
         await elementUpdated(el);
         expect(el.values).to.deep.equal({ a: 10, b: 10, c: 10 });
     });
-    it('warns in Dev Mode when `min="previous"` is leveraged on first handle', async () => {
```
rubencarvalho, line 1381: Did not remove these, just moved them below to be all inside the same `describe('dev mode')` block

packages/slider/src/Slider.ts
```diff
@@ -265,8 +275,8 @@ export class Slider extends SizedMixin(ObserveSlotText(SliderHandle, ''), {
                     @click=${this.handleLabelClick}
                     size=${this.size}
                 >
-                    ${this.slotHasContent ? nothing : this.label}
-                    <slot>${this.label}</slot>
+                    <span>${this.label}</span>
```
Rajdeepc, line 278: we should render this is `this.slotHasContent`. Now with this implementation `this.label` will always be rendered?

Rajdeepc: FYI A slider may or maynot have a label!

Rajdeepc (changes requested):

packages/slider/src/Slider.ts
rubencarvalho: updated it! 👍

Rajdeepc (approved):

PR #5137, opened 2025-02-28
chore: fix smoke test to run before merge

caseyisonit:
## Description

Currently, we have a smoke test running against the documentation PRODUCTION site testing the search functionality. It only runs on merges to main so you don't know there's a failing test until production builds and someone posts a PR. A smoke test should run against the Branch documentation preview site that is generated. 

This PR fixes:
- the search functionality for locally served docs site
- the tests are cleaned up to use more accurate targeting and the correct menu functionality from the omnibus PR
- moved a smoke test to our pull-release action
- refactored script to use a CI URL or fall back to localhost for debugging 
- moved browser checks to own file to run in parallel 

## Related issue(s)

- Jira 689

## Motivation and context

Stabilizing our CI

## How has this been tested?

-   [ ] All tests pass in CI

## Checklist

## Best practices


.github/workflows/browser-tests.yml
```diff
@@ -0,0 +1,114 @@
...
+
+            - name: Archive Firefox tachometer results
+              uses: actions/upload-artifact@v4
+              with:
+                  name: tachometer-results-firefox
+                  path: |
+                      tach-results.firefox.*.json
+                      tachometer.firefox-ran.txt
+
+    compare-chrome:
```
rubencarvalho, line 44: Looking at this, it seems the jobs are exactly the same.
Could we try defining a [matrix](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow) for this?
It could probably reduce this file by half and the tests run in parallel 😃

.github/workflows/browser-tests.yml
```diff
@@ -0,0 +1,114 @@
...
+            - name: Checkout PR branch
+              uses: actions/checkout@v4
+
+            - name: Setup Job and Install Dependencies
+              uses: ./.github/actions/setup-job
+
+            - name: Retrieve tachometer results archives
+              uses: actions/download-artifact@v4
+              with:
+                  name: tachometer-results-firefox
```
rubencarvalho, line 100: I believe you can omit the `name`. If you don't specify it, all artifacts should get downloaded to their respective folders, so you can save the step below.

projects/documentation/searchIndex.json
rubencarvalho, line 1: This should not be committed if we're generating it dynamically. Maybe we could add it to the `.gitignore`?

.github/workflows/test.yml
```diff
@@ -24,14 +26,25 @@ jobs:
             - name: Setup Job and Install Dependencies
               uses: ./.github/actions/setup-job
 
+            - name: Extract branch name
+              shell: bash
+              run: echo "BRANCH=$(npx slugify-cli ${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}})" >> $GITHUB_OUTPUT
+              id: extract_branch
+
+            - name: Extract doc preview url
+              run: echo "DOC_PREVIEW_URL=https://${{ steps.extract_branch.outputs.branch }}--spectrumwc.netlify.app/" >> $GITHUB_OUTPUT
```
rubencarvalho, line 35: You can put this directly in a `GITHUB_ENV` instead and that will save you the steps of redefining the output as a local `ENV` in each step you need to access it. Same with the `BRANCH` value above.
Putting it as a global env variable means it's accessible in all scripts, not just in the dependent steps.

```suggestion
              run: echo "DOC_PREVIEW_URL=https://${{ steps.extract_branch.outputs.branch }}--spectrumwc.netlify.app/" >> $GITHUB_ENV
```

.gitignore
```diff
@@ -92,6 +92,7 @@ test/**/*.js.map
 test/**/*.d.ts
 !test/visual/**/*.js
 test/visual/test.js
+test-results
```
rubencarvalho, line 95: What command / script generates this?

rubencarvalho:
This is awesome! 🚀 I left some small comments 😃

projects/documentation/searchIndex.json
rubencarvalho: BTW, the Lunr script that generates this seems to take a negligible amount of time to run on the first startup, so I think we’re fine keeping it out.

.github/workflows/browser-tests.yml
caseyisonit: Can that be a separate ticket? This was scoped to fixing smoke

rubencarvalho: I am fine with this - this is not required

TarunAdobe (approved):
I likey!

test/visual/test.ts
```diff
@@ -68,6 +97,53 @@ async function testReady(test: StoryDecorator, retry = 0): Promise<void> {
...
+    // Force a layout computation
+    root.getBoundingClientRect();
+
+    // First frame to process attribute changes
+    await nextFrame();
+
+    // Wait for all animations to complete
+    try {
+        const elements = Array.from(root.querySelectorAll('*'));
+        const animations = elements.flatMap((el) => {
```
rubencarvalho, line 116: Instead of querying all elements, you can pass `{subtree: true}` to `el.getAnimations({subtree: true})`. This would catch animations from pseudo-elements and the shadow DOM. In the current case, the shadow DOM is not being traversed.

rubencarvalho (changes requested):

test/visual/test.ts
```diff
@@ -68,6 +97,53 @@ async function testReady(test: StoryDecorator, retry = 0): Promise<void> {
...
+            try {
+                return el.getAnimations ? el.getAnimations() : [];
+            } catch (error) {
+                return [];
+            }
+        });
+
+        if (animations.length > 0) {
+            await Promise.all(
+                animations.map((a) => a.finished.catch(() => undefined))
```
rubencarvalho, line 126: `=> undefined` means `Promise.all` never rejects so the outer catch is redundant because nothing will be logged.

rubencarvalho (approved):

PR #5142, opened 2025-03-03
WIP | feat: update css to latest

castastrophe:
## Description

This is a full upgrade of all packages to use the latest published CSS from the @spectrum-css project.

### Why update everything at once?

It is often easier to see changes in components that influence one-another (i.e., --mod passthrough styles from parent components) if we bump all components together.

That does not mean that all components need to be merged via this PR but rather, that all components can be viewed in unison via this PR.

## Related issue(s)

Most related issues exist in Jira but this is a nice view of the scope of updates we are still needing in order to be up-to-date with the latest fast-follows design updates and CSS theming optimizations (these optimizations have removed _hundreds_ of extemporaneous custom properties.

## How has this been tested?

-   [ ] Strongly recommend relying heavily on the VRT results and sanity-checking the Storybook pages by clicking through some of the more problematic components (especially those with nested SWC components or styles from multiple CSS packages).
-  [ ] Strongly suggest design review this Storybook for their most important components and validate the S2 versions are as they wish. This PR is up-to-date with the latest fast-follows that have been merged.

-   [ ] Did it pass in iPad?

## Types of changes

-   [x] Bug fix (the semver designation as this update bumps the _preview_ theme for S2 which is not the foundational theme in this release as-yet).
-   [x] Breaking _S2 design_ change (this does not translate to the semver used but is an important distinction)

## Checklist

## Best practices


.prettierrc.yaml
```diff
@@ -6,3 +6,7 @@ trailingComma: es5
 bracketSpacing: true
 arrowParens: always
 htmlWhitespaceSensitivity: ignore
+overrides:
```
castastrophe, line 9: Prettier's line wrapping for CSS does not have exceptions for var functions; adding whitespace or newlines to var functions cannot be minified safely downstream (and many minifiers won't) because whitespace is valid in a CSS var function. By increasing the line-length for CSS, we prevent invalid newlines being adding into var functions.

package.json
```diff
@@ -66,9 +67,6 @@
         "process-spectrum": "wireit",
         "publish:react": "yarn changeset publish --no-git-tag --tag snapshot --no-push",
         "push-to-remote": "git add . && git commit -m \"chore: release new versions #publish\" && git push",
-        "spectrum-css": "wireit",
-        "spectrum-tokens": "wireit",
-        "spectrum-vars": "wireit",
```
castastrophe, line 71: These commands don't exist in wireit and haven't for a long while - seems safe to remove the alias'es right?

package.json
```diff
@@ -12,6 +12,7 @@
         "analyze": "lit-analyzer \"{packages,tools}/*/src/**/!(*.css).ts\"",
         "build": "wireit",
         "build:clear-cache": "rimraf packages/*/tsconfig.tsbuildinfo && rimraf tools/*/tsconfig.tsbuildinfo",
+        "build:component:css": "node ./tasks/process-spectrum.js",
```
castastrophe, line 15: I wanted the ability to force-run the build only for specific components as a means of validating updates to the spectrum-config files.

package.json
```diff
@@ -97,16 +95,14 @@
         "vrt:preview": "yarn wds --config test/visual/wds-vrt.config.js",
         "vrt:quick-link": "yarn netlify deploy --alias=vrt --dir=projects/vrt-quick-link"
     },
-    "peerDependencies": {
-        "common-tags": "^1.8.0"
-    },
```
castastrophe, line 102: This implies consumers checking out the monorepo root would need to install common-tags locally for themselves in order to be able to build this project but I don't believe that is the case. Can we remove this? I don't think the monorepo root has any required peerDeps.

PR #5160, opened 2025-03-06
fix(overlay): manual overlay last item should close on pressing escape

Rajdeepc:
## Description

`sp-overlay` with type="manual" should close on pressing `ESC` key. When the last item is on overlay stack we are triggering the close method on esc key event.

## Related issue(s)

- SWC-699

## Motivation and context

PSWeb working on an implementation test where they are closing overlays with type="manual". Our current logic is an anti-pattern on what they are expecting and also this is bug from our side.

## How has this been tested?

-   [ ] Last item on Overlay Stack dismissed on pressing `ESC`
    1. Go [here](https://rajdeep-manual-overlay-escape--spectrumwc.netlify.app/components/overlay/#manual)
    2. Click on Open Manual overlay button
    3. Press ESC and see the overlay dissmisses

-   [ ] Did it pass in iPad?

## Types of changes

## Checklist

## Best practices


packages/overlay/test/overlay-element.test.ts
```diff
@@ -1065,9 +1065,9 @@ describe('sp-overlay', () => {
                     press: 'Escape',
                 });
 
-                await aTimeout(200);
-
-                expect(el.open).to.be.true;
+                await elementUpdated(el);
```
rubencarvalho, line 1068: Can we change the test title to reflect the fixed behavior?
```
            it('does not close when pressing `Escape`', async () => {
```

packages/overlay/src/OverlayStack.ts
```diff
@@ -120,12 +120,12 @@ class OverlayStack {
             event.preventDefault();
             return;
         }
-        if (supportsPopover) return;
         if (last?.type === 'manual') {
             // Manual Overlays should not close on "light dismiss".
```
rubencarvalho, line 124:
```suggestion
            // Manual overlays should close on "Escape" key, but not when losing focus or interacting with other parts of the page.
```

blunteshwar:
can you add some details in `How has this been tested?` section of pr description?

rubencarvalho (approved):
Nice! Thanks for this 🚀

PR #5200, opened 2025-03-16
chore: improve css import results; optimize precommit runs

castastrophe:
### Pre-commit improvements

Previous the husky pre-commit task was running analysis and linting on files whether they were present in the commit or not. This update adds a specificity to the pre-commit tool by ensuring commands are run _only_ on those files being modified by that commit.

Most relevantly, `yarn analyze` was running the lit-analyzer on 5,840 files every single commit even if no changes were made to lit-based assets.

### CSS import improvements

Prettier's line wrapping for CSS does not have exceptions for var functions; adding whitespace or newlines to var functions cannot be minified safely downstream (and many minifiers won't) because whitespace is valid in a CSS var function. By increasing the line-length for CSS, we prevent invalid newlines being adding into var functions.

This PR adds a spectrum-*.css exception to prettier to allow for up-to-500 characters per line to prevent aggressive line wrapping that will break that downstream minification.

It also swaps stylelint configurations to only lean on the header plugin by default and adds the stylelint-config settings for files _not_ being imported from Spectrum CSS. Spectrum CSS already processes the assets according to rigorous Stylelint configurations and these were stripping important vendor prefixes from the results.

## How has this been tested?

**Setup**: Before testing the husky pre-commit hooks, add `exit 1` to the bottom of the pre-commit script. [Docs](https://typicode.github.io/husky/how-to.html#testing-hooks-without-committing). Remove this once testing is complete and do not commit that change.

- [ ] To validate the update to line 6 of .husky/pre-commit (eslint), update any `*.ts` file in the project, `git add . && git commit -m "chore: testing pre-commit hook`. Expect it to run eslint against that file (but no others).
- [ ] To validate the update to line 7 of .husky/pre-commit (lit-analyzer), update any `packages/*/src/**/*.ts` file in the project, `git add . && git commit -m "chore: testing pre-commit hook`. Expect it to run lit-analyzer against that file (but no others).
- [ ] To validate the update to line 8 of .husky/pre-commit (linting css), update any `packages/**/*.css` file in the project, `git add . && git commit -m "chore: testing pre-commit hook`. Expect it to run stylelint against that file (but no others).
- [x] We can already validate the prettier and stylelint updates by reviewing the committed `packages/*/src/spectrum-*.css` files which now have significantly fewer line breaks inside of var functions.
- [ ] To validate no regressions in how stylelint formats non-imported CSS assets, make a change to `packages/accordion/src/accordion.css` to remove the copyright and add `-webkit-animation: alternate;` to the `:host` selector. Run `yarn lint:css` and expect to see:

```sh
packages/accordion/src/accordion.css
  1:1  ✖  Header not found                              header/header
  6:5  ✖  Expected empty line before declaration        declaration-empty-line-before
  6:5  ✖  Unexpected vendor-prefix "-webkit-animation"  property-no-vendor-prefix
```

## Types of changes

## Checklist

-   **N/A**  I have added tests to cover my changes.

## Best practices

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,8 +1,13 @@
 STAGED_FILES_TO_LINT=$(git diff --name-only --cached --diff-filter=d -- "*.ts" "*.js")
+STAGED_FILES_TO_ANALYZE=$(git diff --name-only --cached --diff-filter=d -- "{packages,tools}/*/src/**/!(*.css).ts")
+STAGED_CSS_FILES=$(git diff --name-only --cached --diff-filter=d -- "{packages,tools}/**/*.css")
 VERSION_FILE=$(dirname "$0")/../tools/base/src/version.js
-yarn eslint -f pretty $STAGED_FILES_TO_LINT
-yarn analyze
-yarn lint:css
+
+[[ -z "$STAGED_FILES_TO_LINT" ]] || yarn eslint -f pretty $STAGED_FILES_TO_LINT
+[[ -z "$STAGED_FILES_TO_ANALYZE" ]] || yarn lit-analyzer $STAGED_FILES_TO_ANALYZE
+[[ -z "$STAGED_CSS_FILES" ]] || yarn stylelint $STAGED_CSS_FILES
+
 yarn pretty-quick --staged
+
 yarn genversion --es6 --semi $VERSION_FILE
 git add $VERSION_FILE
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -1,27 +1,33 @@
 {
     "plugins": ["stylelint-header"],
-    "extends": ["stylelint-config-standard"],
+    "extends": [],
     "rules": {
-        "header/header": ["config/license.js", {}],
-        "length-zero-no-unit": [true, { "ignore": "custom-properties" }],
-        "selector-type-no-unknown": [true, { "ignore": ["custom-elements"] }],
-        "selector-pseudo-element-colon-notation": ["single", {}],
-        "custom-property-pattern": "^_?([a-z][a-z0-9]*)(-[a-z0-9]+)*$",
-        "no-duplicate-selectors": null,
-        "selector-class-pattern": null,
-        "no-descending-specificity": null,
-        "declaration-block-no-redundant-longhand-properties": null
+        "header/header": ["config/license.js", {}]
     },
     "overrides": [
         {
             "files": [
-                "packages/**/src/spectrum-*.css",
-                "tools/**/src/spectrum-*.css",
-                "tools/styles/**/*.css"
+                "!packages/**/src/spectrum-*.css",
+                "!tools/**/src/spectrum-*.css",
+                "!tools/styles/**/*.css"
             ],
-            "extends": [],
+            "extends": ["stylelint-config-standard"],
             "rules": {
-                "header/header": ["config/license.js", {}]
+                "header/header": ["config/license.js", {}],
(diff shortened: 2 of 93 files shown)


package.json
```diff
@@ -66,9 +66,6 @@
         "process-spectrum": "wireit",
         "publish:react": "yarn changeset publish --no-git-tag --tag snapshot --no-push",
         "push-to-remote": "git add . && git commit -m \"chore: release new versions #publish\" && git push",
-        "spectrum-css": "wireit",
-        "spectrum-tokens": "wireit",
-        "spectrum-vars": "wireit",
```
castastrophe, line 71: These commands don't exist in wireit and haven't for a long while - seems safe to remove the alias'es right?

package.json
```diff
@@ -97,9 +94,6 @@
         "vrt:preview": "yarn wds --config test/visual/wds-vrt.config.js",
         "vrt:quick-link": "yarn netlify deploy --alias=vrt --dir=projects/vrt-quick-link"
     },
-    "peerDependencies": {
-        "common-tags": "^1.8.0"
-    },
```
castastrophe, line 102: This implies consumers checking out the monorepo root would need to install common-tags locally for themselves in order to be able to build this project but I don't believe that is the case. Can we remove this? I don't think the monorepo root has any required peerDeps.

package.json
```diff
@@ -385,7 +379,7 @@
             ]
         },
         "process-spectrum": {
-            "command": "node ./scripts/spectrum-vars.js && node ./tasks/process-spectrum.js && node ./scripts/generate-tokens.js && yarn lint:css --fix && pretty-quick --pattern \"{packages,tools}/**/*.css\" && pretty-quick --pattern \"packages/dialog/src/spectrum-dialog.css\"",
+            "command": "node ./scripts/spectrum-vars.js && node ./tasks/process-spectrum.js && node ./scripts/generate-tokens.js && yarn lint:css --fix && pretty-quick --pattern \"{packages,tools}/**/*.css\"",
```
castastrophe, line 382: I removed the additional pretty-quick run against spectrum-dialog because it's already included in the pattern for the previous pretty-quick run.

.husky/pre-commit
```diff
@@ -1,8 +1,13 @@
 STAGED_FILES_TO_LINT=$(git diff --name-only --cached --diff-filter=d -- "*.ts" "*.js")
+STAGED_FILES_TO_ANALYZE=$(git diff --name-only --cached --diff-filter=d -- "{packages,tools}/*/src/**/!(*.css).ts")
```
Rajdeepc, line 2: This is nice! :)

castastrophe: What do you mean by "globbing through packages"? Git diff is the _only_ way, as far as I know, to fetch the list of files changed: https://git-scm.com/docs/git-diff

.stylelintrc.json
```diff
@@ -1,27 +1,33 @@
...
-                "tools/styles/**/*.css"
+                "!packages/**/src/spectrum-*.css",
+                "!tools/**/src/spectrum-*.css",
+                "!tools/styles/**/*.css"
             ],
-            "extends": [],
+            "extends": ["stylelint-config-standard"],
             "rules": {
-                "header/header": ["config/license.js", {}]
+                "header/header": ["config/license.js", {}],
```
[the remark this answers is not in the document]
castastrophe: It _should_ inherit but this was me just being safe

caseyisonit (approved):
looks good to me, been tested

Issue #5203, opened 2025-03-17
[Bug]: SP-ACTION-BUTTON href, target="_blank" opening same tab twice when mouse clicked


vikalpgupta:
### Code of conduct

- [x] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-action-button

### Expected behavior

Should open only 1 new tab

### Actual behavior

Two tabs with the same href are getting opened

https://github.com/user-attachments/assets/3aa2f122-78da-4bd2-ada7-80e0a6b5e2ee

### What browsers are you seeing the problem in?

Chrome

### How can we reproduce this issue?

1. Go to https://opensource.adobe.com/spectrum-web-components/components/action-button/
2. Scroll to href, target="_blank" example
3. Click on "Click me" button in href, target="_blank" example
4. This will open multiple new tab with same href

### Sample code or abstract reproduction which illustrates the problem

import '@spectrum-web-components/action-button/sp-action-button.js';

<sp-action-button
    href="https://github.com/adobe/spectrum-web-components"
    target="_blank"
>
    Click me
</sp-action-button>

### Severity

SEV 4

Rajdeepc:
SWC-782

PR #5204, opened 2025-03-17
fix(action-button): remove proxy click in capture phase

Rajdeepc:
## Description

This PR addresses an issue with the `ButtonBase` component where using Command/Meta+click on buttons with an `href` attribute would incorrectly open two browser tabs. The issue was caused by the proxy click mechanism not respecting modifier keys when forwarding the click to the internal anchor element.

### Changes

- Modified the `shouldProxyClick` method to accept an optional MouseEvent parameter
- Added logic to skip the proxy click when modifier keys (Command/Meta, Ctrl, Shift, Alt) are pressed
- Improved accessibility by adjusting ARIA attributes and focus management
- Removed the `hidden` class from the anchor element to improve screen reader compatibility
- Explicitly set `tabindex="-1"` on the anchor element to prevent double focus

## Related issue(s)

-  https://github.com/adobe/spectrum-web-components/issues/5203

## How has this been tested?

-   [ ] Voice Over Test
    1. Go [here](https://rajdeep-remove-proxy-click--spectrum-wc.netlify.app/components/action-button/#handling-events)
    2. Once focus is on the button, voice over will instruct you to use CTRL+Option+Space to click the link.
    3. Press `CTRL+Option+Space` and click the link
    4. See one tab opens

-   [ ] Click Test

    1. Go [here](https://rajdeep-remove-proxy-click--spectrum-wc.netlify.app/components/action-button/#handling-events)
    2. Press Command on Mac or CNTRL in Windows and Click on the button
    3. See one tab opens

-   [ ] Did it pass in iPad?

## Types of changes

## Checklist

## Best practices


packages/button/src/ButtonBase.ts
```diff
@@ -105,8 +105,6 @@ export class ButtonBase extends ObserveSlotText(LikeAnchor(Focusable), '', [
     private shouldProxyClick(): boolean {
         let handled = false;
         if (this.anchorElement) {
-            // click HTML anchor element by proxy
-            this.anchorElement.click();
```
nikkimk, line 109: @Rajdeepc this is breaking the ink in voiceover. To replicate:

1. Using VoiceOver, go to the [example with an HREF in the docs](https://rajdeep-remove-proxy-click--spectrumwc.netlify.app/components/action-button/#handling-events). 
2. Once focus is on the button, voice over will instruct you to use <kbd>CTRL+Option+Space</kbd> to click the link.
3. Nothing happens.

I think removing the proxy Click is absolutely the right direction, but we may need to look at delegating focus of the component to the link or button rendered within the component before removing the proxy.

And since all components that extend focusable and like anchor are built upon the the current pattern there may be some additional refactoring. (I have a JIRA ticket about exploring button refactor if you'd like to know more.)

nikkimk (changes requested):

caseyisonit:
@Rajdeepc can you include testing instructions for this PR?

packages/button/src/ButtonBase.ts
caseyisonit: We will also need to be wary of other components using the Base pattern, which should be included in the PR description and/or testing. 

Perhaps we should include a test with those keyboard commands to simulate non-visual user interaction and ensure the correct behavior persists.

nikkimk (approved):
Nice work! I think the the second test link should have gone to https://rajdeep-remove-proxy-click--spectrum-wc.netlify.app/storybook/?path=/story/action-button--href-with-target instead of https://rajdeep-remove-proxy-click--spectrum-wc.netlify.app/components/action-button/#handling-events, so I tested it there and it worked.

Thanks for fixing this in an accessible way! :)

TarunAdobe (approved):

PR #5221, opened 2025-03-19
PoC | leverage css module imports in components

castastrophe:
## Description

We opened this as a demo of using the native CSS module imports for the web components (reference: https://web.dev/articles/css-module-scripts) and wondering if there are any decent polyfills that could get us the coverage we need to switch over to it. 

This approach would mean in future infrastructure we wouldn't need to convert `*.css` -> `*.ts` -> `*.js` at all. One less thing to ship (and have to maintain as a part of our API too).

## Action items

1. Take a look at this approach in the pull request
2. Click through the Storybook
3. Read up on CSS Module imports
4. Post your thoughts and findings to this PR

### DIFF
--- a/packages/accordion/src/Accordion.ts
+++ b/packages/accordion/src/Accordion.ts
@@ -26,7 +26,7 @@ import { FocusGroupController } from '@spectrum-web-components/reactive-controll
 
 import { AccordionItem } from './AccordionItem.js';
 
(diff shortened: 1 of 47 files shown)


castastrophe:
Found a possible polyfill solution: https://github.com/guybedford/es-module-shims?tab=readme-ov-file#css-modules

PR #5224, opened 2025-03-20
chore: update tachometer reporting to stabilize ci [SWC-767] 

castastrophe:
## Description

When pull requests are opened by tools or utilities, [example](https://github.com/adobe/spectrum-web-components/actions/runs/13961757174), pull request comments cannot be written.

## How has this been tested?

The best way to validate this fix is to merge this update into main and then see if the comment step of the tachometer passes when dependabot opens another pull request.

- [x] Expect to see no regressions to tachometer comments on non-tooling pull requests. Validate that on this pull request [here](https://github.com/adobe/spectrum-web-components/actions/runs/13975871850/job/39129503463?pr=5224). [@castastrophe]

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/.github/workflows/browser-tests.yml
+++ b/.github/workflows/browser-tests.yml
@@ -4,6 +4,10 @@ on:
     pull_request:
         types: [opened, synchronize, reopened]
 
+permissions:
+    contents: read
+    pull-requests: write
+
 jobs:
     test-changed-packages:
         strategy:
@@ -13,7 +17,7 @@ jobs:
 
         # The job will only run if the pull request is from the same repository.
         # Benchmarks can't run on PRs from forked repos due to comment posting restrictions without a GitHub token.
-        if: ${{ github.event.pull_request == null || github.event.pull_request.head.repo.full_name == github.repository }}
+        if: ${{ github.event.pull_request.head.repo.full_name == github.repository }}
         runs-on: ubuntu-22.04
         steps:
             - name: Checkout main
@@ -50,6 +54,7 @@ jobs:
               run: touch tachometer.${{ matrix.browser }}-ran.txt
 
             - name: Archive ${{ matrix.browser }} tachometer results
+              id: upload-artifact
               uses: actions/upload-artifact@v4
               with:
                   name: tachometer-results-${{ matrix.browser }}
@@ -73,11 +78,16 @@ jobs:
 
             - name: Checkout PR branch
               uses: actions/checkout@v4
+              with:
+                  fetch-depth: 2
 
             - name: Setup Job and Install Dependencies
               uses: ./.github/actions/setup-job
 
             - uses: actions/download-artifact@v4
+              with:
+                  pattern: tachometer-results-*
+                  merge-multiple: true
 
             - name: Post Tachometer Performance Comment
               uses: actions/github-script@v7


.github/workflows/browser-tests.yml
```diff
@@ -4,6 +4,10 @@ on:
     pull_request:
         types: [opened, synchronize, reopened]
 
+permissions:
+    contents: read
+    pull-requests: write
```
castastrophe, line 9: Writing a comment requires write access to the pull request.

.github/workflows/browser-tests.yml
```diff
@@ -13,7 +17,7 @@ jobs:
 
         # The job will only run if the pull request is from the same repository.
         # Benchmarks can't run on PRs from forked repos due to comment posting restrictions without a GitHub token.
-        if: ${{ github.event.pull_request == null || github.event.pull_request.head.repo.full_name == github.repository }}
```
castastrophe, line 16: The only trigger for this action is the pull_request event so that value will never be null.

.github/workflows/browser-tests.yml
```diff
@@ -57,17 +62,24 @@ jobs:
...
+                  pattern: tachometer-results-*
+                  merge-multiple: true
 
             - name: Post Tachometer Performance Comment
-              uses: actions/github-script@v7
+              id: comment
+              run: echo "TACHOMETER_COMMENT=$(yarn log:performance | xargs)" >> "$GITHUB_OUTPUT"
+
+            - name: Post Performance to PR
+              uses: marocchino/sticky-pull-request-comment@v2
```
castastrophe, line 81: We use this package over in CSS and it's very effective at posting and maintaining the comment without us having to maintain the code for doing so in our project.

tasks/build-tachometer-comment.js
```diff
@@ -156,16 +156,15 @@ export const buildTachometerComment = () => {
...
 `)
         : 'Currently, no packages are changed by this PR...';
     const firefoxBody = firefoxTables.length
         ? firefoxTables.join(`
 
 `)
         : '';
-    let comment = `## Tachometer results
-`;
+    let comment = '';
```
castastrophe, line 167: The sticky comment utility boosts the title to the GitHub Action so it doesn't need to be set here.

tasks/build-tachometer-comment.js
```diff
@@ -187,4 +186,15 @@ ${body[1]}
     return comment;
 };
 
-console.log(buildTachometerComment());
+export default function main() {
```
castastrophe, line 189: I wrapped the output in this function to make it easier to call this script from the command line.

.github/workflows/browser-tests.yml
```diff
@@ -57,11 +62,16 @@ jobs:
...
               uses: actions/checkout@v4
+              with:
+                  fetch-depth: 2
 
             - name: Setup Job and Install Dependencies
               uses: ./.github/actions/setup-job
 
             - uses: actions/download-artifact@v4
+              with:
+                  pattern: tachometer-results-*
```
castastrophe, line 73: Being more explicit here to ensure we are only downloading the assets we require.

.github/workflows/browser-tests.yml
```diff
@@ -73,11 +78,16 @@ jobs:
 
             - name: Checkout PR branch
               uses: actions/checkout@v4
+              with:
+                  fetch-depth: 2
```
caseyisonit, line 82: What does fetch depth: 2 do? i see above we checkout everything with fetch depth: 0

.github/workflows/browser-tests.yml
```diff
@@ -73,11 +78,16 @@ jobs:
...
 
             - name: Setup Job and Install Dependencies
               uses: ./.github/actions/setup-job
 
             - uses: actions/download-artifact@v4
+              with:
+                  pattern: tachometer-results-*
+                  merge-multiple: true
```
caseyisonit, line 90: does merging multiple change the path or what would the output be for this download? I ask because the logic in the script that is ran to generate the comment is based on the directory structure of this download.

castastrophe: By setting this to true, it ensures that the downloaded artifacts will be in the same directory specified by path. If multiple matches are found and this is not set to true, it would default to putting them in different locations.

castastrophe: For reference: https://github.com/actions/download-artifact

rubencarvalho (approved):

.github/workflows/browser-tests.yml
caseyisonit: thank you for sharing this resoure! nice add

.github/workflows/browser-tests.yml
caseyisonit: im not of a strong opinion so I'm good with two but I think 1 would work too but again zero opinion. this is non blocking for me.

caseyisonit (approved):
checks out for me

Rajdeepc:
@castastrophe I am not sure if this change directly impacted Tachometer to surface up the comments, but can we revisit these changes once again. I see the Tachometer stopped working here. This is the [PR](https://github.com/adobe/spectrum-web-components/pull/5287) where we fixed it.

PR #5249, opened 2025-03-25
ci: disable commit to main branch precommit hook

Rajdeepc:
## Description

- Addresses the need to protect the main branch from direct commits
- Ensures all changes go through proper PR review process

## Motivation and context
Direct commits to the main branch can bypass code review processes and potentially introduce issues into production code. This pre-commit hook enforces best practices by:

- Ensuring all changes go through proper PR review
- Preventing accidental commits to main
- Maintaining code quality through enforced review process
- Supporting team collaboration and code review workflows

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_

    1. Go here
    2. Do this

-   [ ] Did it pass in iPad?

## Types of changes

## Checklist

## Best practices


blunteshwar (approved):
Thanks for doing this!

.husky/pre-commit
```diff
@@ -1,3 +1,14 @@
...
+branch="$(git rev-parse --abbrev-ref HEAD)"
+
+echo ""
+echo "Current Branch: $branch"
+echo ""
+
+if [ "$branch" = "main" ]; then
+  echo "You can't commit directly to the $branch branch"
+  exit 1
+fi
```
castastrophe, line 11: This should not be handled in pre-commit via husky - this is done in the GitHub Project Settings under Branch protection rules

castastrophe (changes requested):

.husky/pre-commit
caseyisonit: Agreed, this should be a branch protection rule

caseyisonit:
I have updated our branch protection rules for main to the following:
Screenshot 2025-03-25 at 12 36 43 PM

We should be able to close this PR now that the protection is in place.

Rajdeepc:
This was already added in branch protection rule. Closing this as this is redundant.

.husky/pre-commit
blunteshwar: True! but I don't think adding this check in pre-commit will do us any harm. Rather it would prevent committing to main even in local. Since some people can still push directly to main irrespective of `locking the branch`, I think it's a good idea to add a check at local level commit.

PR #5308, opened 2025-03-31
fix(overlay): prevent overlay close on document scroll

Rajdeepc:
## Description
This PR fixes an issue where the Picker component closes when the document is scrolled while the picker is open. The fix ensures that pickers remain open during document scrolling, improving the user experience when interacting with pickers in scrollable contexts.

## Changes
- Modified the overlay system to prevent closing pickers on document scroll events
- Added type checking to ensure safe DOM operations

## Related issue(s)

- SWC-731
- https://github.com/adobe/spectrum-web-components/pull/5308

## Motivation and context

This change improves the user experience by:
- Maintaining picker state during document scrolling
- Preventing accidental closures when scrolling through content
- Preserving picker functionality in complex DOM structures

## How has this been tested?

-   [ ] Action Menu
    1. Go [here](https://rajdeep-document-overlay-close--spectrum-w-c.netlify.app/components/action-menu/)
    2. Open a Action Menu
    3. Scroll the document
    4. See the action menu stays open
    5. Click on the document anywhere
    6. Action Menu closes

-   [ ] Picker

    1. Go [here](https://rajdeep-document-overlay-close--spectrum-w-c.netlify.app/components/picker/)
    2. Open a Picker
    3. Scroll the document
    4. See the picker stays open
    5. Click on the document anywhere
    6. Picker closes

-   [x] Did it pass in iPad?

https://github.com/user-attachments/assets/a2a6abd7-f3f3-4fc5-a0cf-4ee3ba59f451

## Types of changes

## Checklist

## Best practices


.changeset/giant-moles-divide.md
```diff
@@ -0,0 +1,5 @@
+---
+'@spectrum-web-components/overlay': minor
+---
+
+prevent overlay close on document scroll
```
castastrophe, line 5: Should this include more details and be formatted since this is the content that will go into the changelog when published? Do we have any best practices about how SWC likes the changelog content to be structured? Maybe it should include the Jira ticket info as well?

packages/overlay/src/OverlayStack.ts
```diff
@@ -34,8 +34,35 @@ class OverlayStack {
         this.document.addEventListener('pointerdown', this.handlePointerdown);
         this.document.addEventListener('pointerup', this.handlePointerup);
         this.document.addEventListener('keydown', this.handleKeydown);
+        this.document.addEventListener('scroll', this.handleScroll, {
```
castastrophe, line 37: Nice use of the standard naming conventions here

packages/picker/test/index.ts
```diff
@@ -1412,6 +1412,41 @@ export function runPickerTests(): void {
...
+                        <sp-menu-item value="option-2">Option 2</sp-menu-item>
+                        <sp-menu-item value="option-3">Option 3</sp-menu-item>
+                    </sp-picker>
+                </div>
+            `);
+
+            const picker = el.querySelector('sp-picker') as Picker;
+            await elementUpdated(picker);
+            await nextFrame();
+            await nextFrame();
```
castastrophe, line 1429: I'm curious what purpose do these duplicate awaits serve? Is there an event or specific state we're waiting on that can be queried for instead to ensure we don't add a new test that might be flaky as a result of an arbitrary wait time?

packages/picker/test/index.ts
```diff
@@ -1412,6 +1412,41 @@ export function runPickerTests(): void {
...
+
+            expect(picker.open).to.be.false;
+
+            const opened = oneEvent(picker, 'sp-opened');
+            picker.click();
+            await opened;
+
+            expect(picker.open).to.be.true;
```
castastrophe, line 1437: Since we're not testing the picker opening action in this test, would it be beneficial  to set up the test with picker already set to open and in this test, just validate that it's in an open state before proceeding?

packages/picker/test/index.ts
```diff
@@ -1412,6 +1412,41 @@ export function runPickerTests(): void {
...
+
+            // Scroll the document
+            if (document.scrollingElement) {
+                document.scrollingElement.scrollTop = 100;
```
castastrophe, line 1441: I'm wondering if the tooling has any event hooks we can connect to that indicate when the scroll action is completed? Might let us reduce the need for the arbitrary awaits afterwards.

castastrophe (changes requested):
A few questions and suggestions here. Just pending the manual test cases in the pull request before approving but the code review portion is complete.

.changeset/giant-moles-divide.md
Rajdeepc: I am working on a branch to surface up a global formatted changelog. Would like to keep the same that we were doing with lerna. Please let me know if you have any ideas on this. Would love to know!

packages/picker/test/index.ts
Rajdeepc: This is a way we use this API to wait for multiple rendering cycles to ensure the DOM has fully updated and any potential asynchronous operations like overlay positioning to have completed.

packages/picker/test/index.ts
Rajdeepc: No, it wouldn't be beneficial in this case. The test is specifically checking that the picker remains open after scrolling, so you need to first verify it opens correctly and then verify it stays open after the scroll event. Setting it to open programmatically would bypass testing the actual opening behavior.

Rajdeepc:
> A few questions and suggestions here. Just pending the manual test cases in the pull request before approving but the code review portion is complete.

Added the test cases and also a working video. Thanks for the review

packages/picker/test/index.ts
caseyisonit: so i have learned this can cause flakiness, did you try elementUpdated(picker)?

castastrophe: Right I understand what purpose it serves but we also have a ton of flaky tests as a result of this kind of pattern which is not awaiting a specific action but an arbitary amount of time.

packages/picker/test/index.ts
castastrophe: It seems to me you need to verify _that it is open_ not that open action can be performed via JS.

packages/picker/test/index.ts
castastrophe: I want us to avoid adding any new tests that would add to our existing list of flaky tests, which is why I raise this as a concern because generic await times don't guarantee the test is ready to be performed.

caseyisonit: calling out the elementUpdated(picker) here too, nextFrame() is the fallback but most cases should work with elementUpdated

Rajdeepc: Thanks. A little bit of context, when multiple overlays or components uses `nextFrame()` simultaneously, the timing of their resolutions might not be predictable, especially during scroll events or rapid state changes. This is where most of the flakiness happens in the virtual triggers. In this context however this is a linear event and I think we can use the `nextFrame()` to safely add the resolution but on the safer side we can definitely and should go with a more robust approach with `waitUntil` like you stated.

.changeset/giant-moles-divide.md
Rajdeepc: We note down all the migration of breaking changes [here](https://opensource.adobe.com/spectrum-web-components/migrations/2024-10-31%20(1.0.0)/)

TarunAdobe (approved):
LGTM!! I don't see any blockers here 🫡

.changeset/giant-moles-divide.md
nikkimk: @Rajdeepc see these recommendations for writing changesets: https://github.com/RedHat-UX/red-hat-design-system/wiki/Changesets

PR #5309, opened 2025-03-31
chore(storybook): minimal upgrade migration

castastrophe:
## Description

This update migrations the SWC Storybook instance off of the `@web/storybook-builder` and `@web/storybook-framework-web-components` and onto the latest breaking change (v8) release of `@storybook/*` and `storybook` packages.

This update does _not_:
- migrate the build off webpack and onto Vite

Had to remove the `maintained node versions` entry to the browsers list config because it was causing the Storybook build to fail:

```sh
Error: For the selected environment is no default script chunk format available:
JSONP Array push can be chosen when 'document' or 'importScripts' is available.
CommonJs exports can be chosen when 'require' or node builtins are available.
Make sure that your 'browserslist' includes only platforms that support these features or select an appropriate 'target' to allow selecting a chunk format by default. Alternatively specify the 'output.chunkFormat' directly.
```

## Related issue(s)

- [SWC-766](https://jira.corp.adobe.com/browse/SWC-766)

## Motivation and context

[Blog post](https://storybook.js.org/blog/storybook-8/)

- Built-in visual testing
- 2-4x faster test builds
- Refreshed desktop UI
- Rebuilt mobile UX
- Removed React dependency for non-React projects (yay!)

## How has this been tested?

-   [ ] _Open the storybook_ (either locally or the one built to this [PR](https://castastrophe-chore-storybook-migration--spectrum-w-c.netlify.app/storybook))
    1. Expect the landing page to now be a docs summary story for the Accordion.
    2. Expect each component and component group to now contain a summary docs page with a view of the default story and a list of the available properties.
    3. Expect no regressions in how theming works via the global knobs at the top
    4. If a story does not contain any defined properties, expect the properties section of the docs page to be missing as in [Action bar](https://castastrophe-chore-storybook-migration--spectrum-w-c.netlify.app/storybook/?path=/docs/action-bar--docs)
-   [ ] _Expect to see no regressions in visual regression tests_
-   [ ] _Expect to see no regressions in functional tests_
-   [ ] `yarn start`: expect no regressions to local Storybook
    - Update a CSS asset and validate the changes refresh live in Storybook
    - Update a TS asset and validate the changes refresh live in Storybook

## Screenshots

Screenshot 2025-04-01 at 10 07 59 AM

## Types of changes

## Checklist

## Best practices


package.json
```diff
@@ -187,19 +190,20 @@
         "pretty-bytes": "^6.1.1",
         "pretty-quick": "^4.1.1",
         "re-template-tag": "^2.0.1",
-        "react": "^18.2.0",
-        "react-dom": "^18.2.0",
+        "react": "^18.3.1",
+        "react-dom": "^18.3.1",
```
castastrophe, line 194: Question - do we need the react dependency for anything other than Storybook? If not, I can remove it from here because Storybook no longer requires React as a peer dep

caseyisonit: Let me confirm but i believe its used for the react wrappers but ill follow up

storybook/preview.js
```diff
@@ -148,4 +165,72 @@ export const parameters = {
     },
 };
 
-export const decorators = [swcThemeDecorator];
+export const decorators = [
```
caseyisonit, line 168: how is this different from the projects/story-decorator? can we have one or the other or do we need both?

storybook/preview.js
```diff
@@ -148,4 +165,72 @@ export const parameters = {
...
+                .direction=${textDirection}
+                ?reduce-motion=${reduceMotion}
+            >
+                ${story(context)}
+            </sp-story-decorator>
+        `;
+    },
+];
+
+export const tags = ['autodocs'];
```
caseyisonit, line 236: can NOT wait for this to go live

web-test-runner.utils.js
```diff
@@ -277,32 +278,3 @@ export const configuredVisualRegressionPlugin = () =>
             );
         },
     });
-
-export function watchSWC() {
```
caseyisonit, line 281: Do you know what this was doing or why its safe to remove this now?

package.json
caseyisonit: so looks like `react` and `react-dom` are used by another storybook that exists in `projects/custom-css-vars-viewer`. I don't know if that's even used anymore or if we can remove that project too? Otherwise it appears safe to remove

Also I think we can remove `next` as a package too.

package.json
```diff
@@ -113,32 +113,35 @@
         "@rollup/plugin-node-resolve": "^15.2.3",
         "@sindresorhus/slugify": "^2.1.1",
         "@spectrum-web-components/eslint-plugin": "file:./linters/eslint",
-        "@storybook/addon-a11y": "^7.5.0",
-        "@storybook/addon-designs": "^8.0.0",
-        "@storybook/addon-essentials": "^7.5.0",
-        "@storybook/addon-links": "^7.5.0",
-        "@storybook/web-components": "^7.5.0",
```
caseyisonit, line 120: looks like this package was removed but this is how we set our custom manifest (the part you commented out in preview.js). Was this intentional or is it not supported in storybook 8?

storybook/preview.js
```diff
@@ -9,15 +9,19 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
...
 import { Locales } from '@spectrum-web-components/story-decorator/src/locales.js';
-import { setCustomElementsManifest } from '@storybook/web-components';
+// import { setCustomElementsManifest } from '@storybook/web-components';
+import DocumentationTemplate from './DocumentationTemplate.mdx';
+import '@spectrum-web-components/story-decorator/sp-story-decorator.js';
 
-const cem = await import('./custom-elements.json', {
-    assert: { type: 'json' },
-});
+// const cem = await import('./custom-elements.json', {
```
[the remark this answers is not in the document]
caseyisonit: This actually helps build our args tables dynamically from the source. By removing this all args tables will default to only the few args defined in an args file. not all components use this pattern though. is there a way we can do a follow up ticket to ensure our args tables are accurate with the available API?

package.json
caseyisonit: Does look like it comes with v8 support: https://www.npmjs.com/package/@storybook/web-components

storybook/preview.js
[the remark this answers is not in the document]
castastrophe: That was what it was supposed to do but none of the args tables were being set by this. When I removed it, it caused no changes to the interface so I suspect it wasn't working the way it was supposed to in the first place.

This is also a WIP branch - not ready for review yet.

package.json
castastrophe: It may say that it's supported but there's some serious issues in that package. When I add the dependency, it adds this to our install logs (yes, it does the same if we bump all packages to 8.6.12 too):

Screenshot 2025-04-14 at 1 38 53 PM

When we add it in anyway despite the logged error, it blocks styles from loading. When I commented out the CEM loading, the arg types still loaded in the stories and the styles rendered. We can debug loading the CEM another way but I'm not sure this package is going to be it.

package.json
castastrophe: I tested out removing next and got a set of these errors so maybe we hold off on that for now:

```
react/icons-ui/next/TripleGripper.ts:13:21 - error TS2307: Cannot find module 'next/dynamic' or its corresponding type declarations.

13 import dynamic from 'next/dynamic';
```

web-test-runner.utils.js
castastrophe: Yes this was forcing rebuild of the assets when updates were made to the source files. The current system handles this automatically and doesn't need us to manually kick this off any more. We can validate this by making changes to an existing asset and seeing it live refresh on the page. I'll add a test case for that when I start writing up validation steps.

storybook/main.js
```diff
@@ -9,14 +9,11 @@ the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTA
...
-import { fromRollup } from '@web/dev-server-rollup';
-import rollupCommonjs from '@rollup/plugin-commonjs';
-import { watchSWC } from '../web-test-runner.utils.js';
 
-/** @type { import('storybook-builder-wds').StorybookConfigWds } */
-const config = {
+import { merge } from 'webpack-merge';
+
+/** @type { import('@storybook/web-components-webpack5').StorybookConfig } */
+export default {
```
pfulton, line 16: A personal nit: how do you feel about including the `disableWhatsNewNotifications: true` setting in the `core` settings (like we do in Spectrum CSS)? Also: what if we set `disableTelemetry: true`?

pfulton (approved):
I pulled down this branch and ran it successfully on my local environment. Everything looked good, functioned as expected, and was speedy!

I love that this enables the AutoDocs stuff!

I dropped one nit comment here that is just a personal preference, but let me know what you think.

Otherwise, I _think_ this should be good to go.

caseyisonit (approved):
LGTM - would love a fast follow to investigate the custom element manifest or automating the args table over the current implementation which is completely manual.

web-test-runner.utils.js
nikkimk: worked for me!

nikkimk (approved):
As always, thanks for making our developer experience so much better!

Issue #5114, opened 2025-02-24
[Bug]: (Mobile) Tapping on sp-menu-item in sp-menu will also tap on element underneath


RuxG:
### Code of conduct

- [x] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-menu, sp-menu-item

### Expected behavior

On mobile, while having the following rendering logic:
1. An `sp-menu` that is rendered conditionally
2. An `sp-checkbox` that is rendered conditionally, in the place of the `sp-menu`, after an `sp-menu-item` is tapped

By tapping on the `sp-menu-item`, the tap event should not propagate to the checkbox underneath.

### Actual behavior

By tapping on the `sp-menu-item`, the tap event is propagated to the checkbox underneath, and it becomes checked right after being rendered.

### What browsers are you seeing the problem in?

Chrome, Safari

### How can we reproduce this issue?

https://github.com/user-attachments/assets/b1e395c8-396e-4222-8ac5-65a42f796d6b

You can reproduce the issue on: https://studio.webcomponents.dev/edit/2oibVYTXHpdxV0KB302m/src/index.ts?p=stories

### Severity

None

najikahalsema:
I marked this bug as SEV2. 

SWC-693

caseyisonit:
@RuxG can you provide a use case for how this might happen in a user experience? This reproduction is very isolated, and we want to better understand how this is implemented in a production app where this behavior would happen. Also what version of Spectrum are you using that this bug is present in?

RuxG:
I am rendering an sp-menu within a panel. This menu displays a list of categories. By clicking on a category within the panel, the main content of the panel is replaced with a list of items from the selected category.  These items in the list can be selected. The list of items renders in the place where the menu was, and somehow the tap on the menu-item propagates to the list of items that is rendered, causing the item to get selected.

PR #5313, opened 2025-04-02
fix(menu): capture touchend on mobile

TarunAdobe:
## Description

This pull request addresses the issue where tapping on an `sp-menu-item` causes the tap event to propagate to an `sp-checkbox` underneath it, resulting in the checkbox being checked unintentionally. The fix involves capturing the `touchend` event on the `sp-menu` to prevent this propagation.

On mobile, while having the following rendering logic:

- An `sp-menu` that is rendered conditionally
- An `sp-checkbox` that is rendered conditionally, in the place of the `sp-menu`, after an `sp-menu-item` is tapped

By tapping on the `sp-menu-item`, the tap event should not propagate to the checkbox underneath.

## Related issue(s)

- [Issue #5114](https://github.com/adobe/spectrum-web-components/issues/5114) 

## Motivation and context

This change is required to enhance the user experience on mobile devices by ensuring that interactions with the `sp-menu-item` do not inadvertently affect the state of the `sp-checkbox`. This prevents confusion and unintended behavior when users interact with the menu.

## How has this been tested?

-   [ ] **Test case 1**
    1. Go to the following reproduction [link](https://ttomar-menu-replace-checkbox--spectrum-w-c.netlify.app/storybook/?path=/story/menu-item--menu-item-with-checkbox) on a mobile device.
    2. Tap on an `sp-menu-item` and verify that the `sp-checkbox` does not become checked.
    
-   [x] Did it pass in iPad?

_No response_

## Types of changes

## Checklist

-   [x] I have read the **[CONTRIBUTING](https://github.com/adobe/spectrum-web-components/blob/main/CONTRIBUTING.md)** document.
-   [x] I have reviewed the Accessibility Practices for this feature, see: [Aria Practices](https://www.w3.org/TR/wai-aria-practices/)

## Best practices


packages/menu/src/Menu.ts
```diff
@@ -151,10 +151,9 @@ export class Menu extends SizedMixin(SpectrumElement, { noDefaultSize: true }) {
 
     /**
      * whether or not to support pointerdown - drag - pointerup selection strategy
-     * default is true
-     * should be false for mobile to prevent click event being captured behind the menu-tray (cz menu immediately closes on pointerup)
+     * default is false to prevent click/touch event being captured behind the menu-tray in a mobile environment (cz menu immediately closes on pointerup)
      */
-    public shouldSupportDragAndSelect = true;
+    public shouldSupportDragAndSelect = false;
```
caseyisonit, line 156: Changing a default value is considered a breaking change. How is this property being used and are there tests associated with it? Is there another way of doing this fix without changing the default value? Can we listen for device type and then set it? Or how does drag and drop work on mobile? 

Posing questions I’ll help look deeper too

TarunAdobe: This default value should always have been false. 
I shipped this in our last release. 
The place where we used it (i.e. picker) automatically applies the correct one always so this is not breaking anybody.
as for it being public... it's public cause we need to toggle this in picker otherwise we don't document it and its supposed to be internal. 

happy to hear your thoughts

Rajdeepc:
@TarunAdobe Thanks for creating a baseline with this new story. Can you also add a test here?

packages/menu/src/Menu.ts
rubencarvalho: Should this be `protected` instead?

TarunAdobe:
> @TarunAdobe Thanks for creating a baseline with this new story. Can you also add a test here?

Can't do that. This is a fix for mobile devices and our test suite doesn't support that fully :(

Rajdeepc:
> > @TarunAdobe Thanks for creating a baseline with this new story. Can you also add a test here?
> 
> Can't do that. This is a fix for mobile devices and our test suite doesn't support that fully :(

Does this help? https://playwright.dev/docs/api/class-touchscreen

blunteshwar (approved):

packages/menu/src/Menu.ts
```diff
@@ -151,10 +151,9 @@ export class Menu extends SizedMixin(SpectrumElement, { noDefaultSize: true }) {
 
     /**
      * whether or not to support pointerdown - drag - pointerup selection strategy
```
rubencarvalho, line 153:
```suggestion
/**
 * Whether to support the pointerdown-drag-pointerup selection strategy.
 * Defaults to false to prevent click/touch events from being captured
 * behind the menu tray in mobile environments (since the menu closes
 * immediately on pointerup).
 */
```

Rajdeepc (approved):
LGTM!

Issue #5109, opened 2025-02-20
[Bug]: sp-color-field 1.1.2 color swatch does not scroll with text control.


mscherotter:
### Code of conduct

- [x] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-color-field

### Expected behavior

the color swatch and edit field should move together when in a scrollable region.

### Actual behavior

I found a bug with the sp-color-field 1.1.2 Spectrum Web Component when put it in a scrollable region in a sp-tab-panel container, the color swatches do not move with the edit fields.
see attached video
https://github.com/user-attachments/assets/6156b7e4-dec5-47ad-86c3-ccca79b73405

### Screenshots

https://github.com/user-attachments/assets/bcad4039-0cfb-4c88-bc51-dd6ad71f4c16

### What browsers are you seeing the problem in?

Microsoft Edge

### How can we reproduce this issue?

3. Scroll to '....'
4. Check console
5. See error

### Severity

SEV 3

mscherotter:
added sample project that demonstrates the bug.

mscherotter:
Here is a link to the reproduced sample: https://1drv.ms/u/s!AmvW0E0DYMdm4bsydGPKhVyDS9njdQ?e=u90WNB

najikahalsema:
SWC-635

PR #5322, opened 2025-04-03
fix(color-field): color handle positioning within scrollable containers

blunteshwar:
## Description
**Issue**
The color handle in sp-color-field components was displaying in incorrect positions when multiple color fields were placed inside a scrollable container. As the container scrolled, the color handles would appear disconnected from their respective color fields, creating a poor user experience.

**Root Cause**
The issue occurred because absolutely positioned elements (like the sp-color-handle) require a positioned ancestor to establish their coordinate system. Since the position of sp-color-field(host) was by default `static` so without one, they position themselves relative to the nearest positioned ancestor (often the scrollable container itself), causing positioning inconsistencies during scrolling.

**Solution**
Added position: relative to the :host selector in color-field.css
This simple change ensures that each color field component establishes its own positioning context, causing the color handle to position itself relative to its parent component sp-color-field rather than a distant ancestor.

## Related issue(s)
#5109 

-

## How has this been tested?

-   [ ] _Test case 1_
    1. Go [here](https://pvashish-swc-685--spectrum-w-c.netlify.app/storybook/?path=/story/color-field--multiple)
    2. Confirm that the color handles maintain correct positioning while scrolling

-   [ ] Did it pass in iPad?

Before
Screenshot 2025-04-03 at 12 09 49 PM

After
Screenshot 2025-04-03 at 12 23 39 PM

## Types of changes

## Checklist

## Best practices


packages/color-field/src/color-field.css
```diff
@@ -0,0 +1,19 @@
...
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License./
+ */
+
+:host {
+    position: relative;
+}
+
+sp-color-handle {
```
rubencarvalho, line 17: I just noticed `sp-color-handle` already has `position: absolute` set here: https://github.com/adobe/spectrum-web-components/blob/main/packages/color-handle/src/spectrum-color-handle.css#L48

So we can remove this rule and just leave the `position: relative` above 😄

blunteshwar: Done

Rajdeepc (approved):
Verified in Mobile and Desktop.LGTM!

blunteshwar:
Waiting for the new golden Image hash to be generated!

rubencarvalho (approved):

PR #5328, opened 2025-04-04
ci: global changelog generator script

Rajdeepc:
## Description
This PR introduces a dual changelog generation setup in our monorepo:

✅ Per-package changelogs via @changesets/changelog-github

✅ Custom global changelog appended to the root CHANGELOG.md using a Yarn 4-compatible ESM script

## Changes
- Configured .changeset/config.json to use @changesets/changelog-github for individual packages.

- Added a custom ESM script (scripts/add-global-changelog.js) to:

  - Read newly generated changelogs from the .changeset output

  - Format and append only feat: and fix: commits to the root CHANGELOG.md

  - Automatically add diff links (e.g., [1.4.0](compare/v1.3.0...v1.4.0))

Updated package.json scripts to include:
```zsh
"changeset-publish": "yarn prepublishOnly && yarn changeset version && yarn changelog:global && ......."
"changelog:global": "node scripts/add-global-changelog.mjs"
```

## Related issue(s)

- SWC-837

## Motivation and context

Since Changesets doesn’t support a global changelog out of the box, this is a custom script to bring that adds the experience back.

## How has this been tested?

**DO NOT COMMIT OR PUSH ANYTHING FROM TESTING**

-   [ ] _Patch version changelog_
    1. Pull down this PR
    2. Run `yarn prepublishOnly && yarn changelog:global`
    3. Review the output to match all available snapshots (patch)
    4. Check the version number is 1.6.1

-   [ ] _Minor version changelog_
    1. Revert the changelog changes from the previous test
    2. Run `yarn changeset`
    3. Select `sp-alert-banner` package
    5. Select `minor` version
    6. Enter changeset message: `-   **Fixed**: Updated `<sp-alert-banner>` default styles for better contrast [#9999](https://github.com/adobe/spectrum-web-components/pull/9999)`
    7. Run `yarn changelog:global`
    8. Review the output to match all available snapshots (minor and patch)
    9. Check the version number is 1.7.0

-   [ ] _Major version changelog_
    1. Revert the changelog changes from the previous test, not the changeset
    2. Run `yarn changeset`
    3. Select `sp-textfield` package
    5. Select `major` version
    6. Enter changeset message: `-   **Added**: Added `hidden-label` attribute, for use with assistive technologies. (Before: `<sp-textfield label="Full Name"></sp-textfield>` / After: `<sp-textfield hidden-label="Full Name"></sp-textfield>`)`
    7. Run `yarn changelog:global`
    8. Review the output to match all available snapshots (major, minor and patch)
    9. Check the version number is 2.0.0

**UNSTAGE ALL CHANGES AND CHANGESETS**

Screenshot 2025-05-12 at 2 01 07 PM
Screenshot 2025-05-12 at 2 15 02 PM
Screenshot 2025-05-12 at 2 16 25 PM

## Types of changes

## Checklist

## Best practices


scripts/add-global-changelog.js
```diff
@@ -0,0 +1,93 @@
...
+
+import fs from 'fs';
+import { execSync } from 'child_process';
+import { fileURLToPath } from 'url';
+import path from 'path';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const repoUrl = 'https://github.com/adobe/spectrum-web-components';
+
+const pkg = JSON.parse(
```
castastrophe, line 22: A best practice to keep code scoped is to wrap them in a main function and call the function at the end of the file. I think it's a good idea to maintain that best practice here that we see in our other scripts as well.

scripts/add-global-changelog.js
```diff
@@ -0,0 +1,93 @@
...
+    fs.readFileSync(path.resolve(__dirname, '../package.json'), 'utf-8')
+);
+const newVersion = pkg.version;
```
castastrophe, line 25: I think you can skip this abstraction since newVersion is only used in the following line. We also need a check for if the package doesn't load to throw that warning. You're assuming here that pkg.version exists.

scripts/add-global-changelog.js
```diff
@@ -0,0 +1,93 @@
...
+const newTag = `v${newVersion}`;
+const prevTag = execSync('git tag --sort=-creatordate')
+    .toString()
+    .split('\n')
+    .filter(Boolean)
+    .find((tag) => tag !== newTag);
```
castastrophe, line 31: This needs more robust failure captures. I recommend if you're going to use exec for this, separate the command execution (exec is notoriously flaky in node scripts so you  need to account for it failing) from the string parsing. Check that exec returned a string and then run the split, etc.

scripts/add-global-changelog.js
```diff
@@ -0,0 +1,93 @@
...
+
+if (!prevTag) {
+    console.error('No previous tag found.');
+    process.exit(1);
```
castastrophe, line 35: I'm trying to think about what information I would need to debug this error. At the age of this project, there's no change that there aren't previous tags to be found so maybe we want this error to tell us why the exec command couldn't return a value we were expecting? Maybe this should log the git tag command output?

scripts/add-global-changelog.js
```diff
@@ -0,0 +1,93 @@
...
+}
+
+const date = new Date().toISOString().split('T')[0];
```
castastrophe, line 38:
```suggestion
const date = new Date().toLocaleDateString('en-CA', {
    year: 'numeric',
    month: '2-digit',
    day: '2-digit',
}));
```

This should return the result you're wanting without having to do inline array parsing (which can sometimes lead to invalid results or fail when the array isn't in the format we're expecting).

scripts/add-global-changelog.js
```diff
@@ -0,0 +1,93 @@
...
+const compareUrl = `${repoUrl}/compare/${prevTag}...${newTag}`;
```
castastrophe, line 39: Is prevTag allowed to be a pre-tag or is there a requirement that prevTag must be a semver version? It seems like it must be one of the semver releases (not the betas for example) so maybe we can add a comment to that effect?

scripts/add-global-changelog.js
```diff
@@ -0,0 +1,93 @@
...
+const commitLogs = execSync(`git log ${prevTag}..HEAD --pretty=format:"%s|%h"`)
+    .toString()
+    .trim();
```
castastrophe, line 42: This looks like it's returning the commit logs but not the changelog content. Is that what we're wanting to add to the global changelog? It seems like the commit history is less useful now that we've migrated to changesets.

scripts/add-global-changelog.js
```diff
@@ -0,0 +1,93 @@
...
+            features.push(entry);
+        } else if (type === 'fix') {
+            fixes.push(entry);
+        }
+    }
+});
+
+// Skip if nothing relevant
+if (!features.length && !fixes.length) {
+    console.log('🚫 No new feat() or fix() commits to add.');
```
castastrophe, line 67:
```suggestion
    console.log('🚫 No new feat() or fix() commits to add.');
```

Since no new features or fixes isn't necessarily a failure of the script, should we format this more like a success message that it ran successfully but with no changes?

scripts/add-global-changelog.js
```diff
@@ -0,0 +1,93 @@
...
+    if (typeMatch) {
+        const [, type, scope, description] = typeMatch;
+        const entry = `-   **${scope}**: ${description} ([\`${hash}\`](${repoUrl}/commit/${hash}))`;
+        if (type === 'feat') {
+            features.push(entry);
+        } else if (type === 'fix') {
+            fixes.push(entry);
+        }
+    }
+});
```
castastrophe, line 63: I think this can serve a temporary fix but we should really be using the changesets tooling to create this content from it's new source (which is not the commit messages): https://github.com/changesets/changesets/blob/main/docs/modifying-changelog-format.md#writing-changelog-formatting-functions

I think the challenge with this as a sustainable approach is that less and less useful data is present in commit messages and the real value for customers now lives in the changesets files.

scripts/add-global-changelog.js
```diff
@@ -0,0 +1,93 @@
...
+
+// Prepend to existing CHANGELOG.md
+const changelogPath = path.resolve(__dirname, '../CHANGELOG.md');
+const existingChangelog = fs.existsSync(changelogPath)
+    ? fs.readFileSync(changelogPath, 'utf-8')
+    : '';
+
+fs.writeFileSync(
+    changelogPath,
+    `${newEntry.trim()}\n\n${existingChangelog}`,
```
castastrophe, line 90: Wouldn't this push the `# Change log` heading down to below the latest update?

castastrophe (changes requested):
If this is an urgent automation, I think this approach can work as a temporary fix with a few hardening updates to how it's written but long-term we need to be handling this from the Changesets tooling (https://github.com/changesets/changesets/blob/main/docs/modifying-changelog-format.md#writing-changelog-formatting-functions)

scripts/add-global-changelog.js
Rajdeepc: This is a point where I need to discuss. I don't see only a summary of the change is sufficient for the users to check what changes went along. I want to keep the CHANGELOG to follow the same pattern as we were doing during lerna which I feel the users still wants.

Rajdeepc:
> If this is an urgent automation, I think this approach can work as a temporary fix with a few hardening updates to how it's written but long-term we need to be handling this from the Changesets tooling (https://github.com/changesets/changesets/blob/main/docs/modifying-changelog-format.md#writing-changelog-formatting-functions)

Thanks for the info, I deliberately added a custom script and didn't opt for built-in `changeset` tooling since it doesn' expose any API to create a global CHANGELOG which is the focus of this PR.

blunteshwar:
> ling since it doesn' expose any API to create a global CHANGELOG which is the focus of this PR. There has been some discussions around this in the community but this is not something its on a priority radar for the team

I agree, I think the custom script approach here is quite pragmatic. While Changesets does a great job generating per-package changelogs, it doesn’t currently offer a built-in mechanism or public API to aggregate them into a single global changelog file. This script effectively fills that gap without waiting on upstream support.

Relying solely on Changesets tooling would either limit us to fragmented changelogs or require non-trivial workarounds. The custom solution gives us immediate value, full control over formatting, and keeps things transparent for consumers of the monorepo.

Until Changesets supports this natively, this seems like a solid and scalable workaround.

scripts/add-global-changelog.js
```diff
@@ -0,0 +1,181 @@
+/*
```
castastrophe, line 1: Minor nitpick but should we name this file update-global-changelog.js instead?

Rajdeepc: Happy to!

CHANGELOG.md
```diff
@@ -3,18 +3,6 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
-# [1.5.0](https://github.com/adobe/spectrum-web-components/compare/v1.4.0...v1.5.0) (2025-04-15)
```
castastrophe, line 6: Is this in here because the branch needs to be rebased with main?

scripts/update-global-changelog.js
```diff
@@ -0,0 +1,181 @@
...
+
+*/
+
+import fs from 'fs';
+import { execSync } from 'child_process';
+import { fileURLToPath } from 'url';
+import path from 'path';
+import { version as newVersion } from '@spectrum-web-components/base/src/version.js';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
```
castastrophe, line 20: Nice organization here

CHANGELOG.md
Rajdeepc: This is just for test. Once the PR is approved I will remove this generation.

scripts/update-global-changelog.js
```diff
@@ -0,0 +1,181 @@
...
+const repoUrl = 'https://github.com/adobe/spectrum-web-components';
+
+async function createGlobalChangelog() {
+    const newTag = `v${newVersion}`;
```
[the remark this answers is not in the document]
Rajdeepc: You're right - Added the necessary fail-safe check at the beginning of the function to ensure newVersion exists before attempting to use it.

CHANGELOG.md
Rajdeepc: Yes, I was testing the new script.

scripts/update-global-changelog.js
```diff
@@ -0,0 +1,185 @@
...
+import fs from 'fs';
+import { execSync } from 'child_process';
+import { fileURLToPath } from 'url';
+import path from 'path';
+import { version as newVersion } from '@spectrum-web-components/base/src/version.js';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const repoUrl = 'https://github.com/adobe/spectrum-web-components';
+
+async function createGlobalChangelog() {
```
castastrophe, line 23: Since this is a new script, can we include the JSDoc notes on all the new functions just to keep our stuff well documented. Might be good to add a note about when this script needs to be run.

package.json
```diff
@@ -23,7 +23,8 @@
         "build:types": "wireit",
         "build:watch": "wireit",
         "changeset-snapshot-publish": "yarn prepublishOnly && yarn changeset version --snapshot && yarn lint:versions --fix && yarn update-version && yarn changeset publish --no-git-tag --tag snapshot",
-        "changeset-publish": "yarn prepublishOnly && yarn changeset version && yarn install && yarn lint:versions --fix && yarn update-version && yarn changeset publish --no-git-tag && yarn push-to-remote && yarn create-git-tag && yarn postpublish",
+        "changeset-publish": "yarn prepublishOnly && yarn changeset version && yarn changelog:global && yarn install && yarn lint:versions --fix && yarn update-version && yarn changeset publish --no-git-tag && yarn push-to-remote && yarn create-git-tag && yarn postpublish",
```
blunteshwar, line 26: Shouldn’t `yarn changelog:global` be executed before `yarn changeset version`? This is because `yarn changelog:global` reads from changeset files located in the `.changeset` directory. However, after running `yarn changeset version`, all the changesets are removed/deleted, and the changelogs are populated. Consequently, `yarn changelog:global` will no longer be able to read the changeset files in the `.changeset` directory.

blunteshwar (changes requested):

package.json
Rajdeepc: You're absolutely right — excellent observation.

projects/documentation/src/components/styles.css
```diff
@@ -10,26 +10,21 @@ OF ANY KIND, either express or implied. See the License for the specific languag
...
+@import url("@spectrum-web-components/styles/tokens/global-vars.css");
+@import url("@spectrum-web-components/styles/tokens/spectrum/global-vars.css");
+@import url("@spectrum-web-components/styles/tokens/spectrum/custom-vars.css");
+@import url("@spectrum-web-components/styles/tokens/spectrum/system-theme-bridge.css");
+@import url("@spectrum-web-components/styles/src/spectrum-heading.css");
+@import url("@spectrum-web-components/styles/src/spectrum-body.css");
+@import url("@spectrum-web-components/styles/src/spectrum-code.css");
+@import url("@spectrum-web-components/opacity-checkerboard/src/spectrum-opacity-checkerboard.css");
+@import url("./inline-alert.css");
+@import url("./fonts.css");
```
blunteshwar, line 22: Why did we add this `url` here?

blunteshwar (approved):

projects/documentation/src/components/styles.css
Rajdeepc: These files came in from the `main`.

scripts/update-global-changelog.js
```diff
@@ -0,0 +1,242 @@
...
+
+    // Preserve the header if it exists in the current changelog
+    let headerText = '';
+    const headerMatch = existingChangelog.match(
+        /^(# Change Log\n\n[\s\S]+?(?=\n\n# \[))/
+    );
+    if (headerMatch) {
+        headerText = headerMatch[1];
+        existingChangelog = existingChangelog.substring(headerMatch[0].length);
+    } else if (existingChangelog.startsWith('# Change Log')) {
```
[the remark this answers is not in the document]
Rajdeepc: In both cases of if and else, the code:

1. Extracts the header text into headerText
2. Removes the header from existingChangelog
3. Later uses these to reconstruct the changelog with the new version entry

The main difference is that the first case expects version entries to follow the header, while the second case handles a changelog that might be empty or just have a header.

rubencarvalho (approved):

PR #5365, opened 2025-04-10
chore: update beta release workflow to use snapshots

rubencarvalho:
## Related issue(s)

-

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_

    1. Go here
    2. Do this

-   [ ] Did it pass in iPad?

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/.github/workflows/beta-release.yml
+++ b/.github/workflows/beta-release.yml
@@ -21,13 +21,11 @@ jobs:
                   git config --global user.email "support+actions@github.com"
                   git config --global user.name "github-actions-bot"
 
-            - name: Update package versions for beta release
+            - name: Update package versions for beta snapshot release
               env:
                   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
               run: |
-                  yarn changeset pre enter beta
-                  # Apply the changeset with specific beta tag
-                  yarn changeset version
+                  yarn changeset version --snapshot
                   yarn lint:versions --fix
                   yarn update-version
 
@@ -37,12 +35,12 @@ jobs:
               run: |
                   echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }}" > ~/.npmrc
 
-            - name: Publish beta release
+            - name: Publish beta snapshot release
               env:
                   NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
               run: |
                   git add .
-                  git commit -am "chore: publish beta version"
+                  git commit -am "chore: publish beta snapshot version"
                   yarn prepublishOnly
-                  yarn changeset publish --no-git-tag
+                  yarn changeset publish --no-git-tag --tag beta
                   git reset --hard HEAD^


.changeset/config.json
```diff
@@ -12,5 +12,9 @@
     "access": "public",
     "baseBranch": "main",
     "updateInternalDependencies": "patch",
-    "ignore": []
+    "ignore": [],
+    "snapshot": {
+        "useCalculatedVersion": true,
```
rubencarvalho, line 17: When `changesets version --snapshot` is used, the default behavior is to use `0.0.0` as the base version for the snapshot release.

Setting `useCalculatedVersion: true` will change the default behavior and will use the calculated version, based on the changeset files.

.changeset/config.json
```diff
@@ -12,5 +12,9 @@
...
+        "prereleaseTemplate": "{tag}.{datetime}"
```
rubencarvalho, line 18: This configures the suffix for the snapshot releases, using a template with placeholders.
I decided for `tag.datetime`, which seems to be a common pattern. But we could have used:
- `{tag}` - the name of the snapshot tag, as specified in `--snapshot something`
- `{commit}` - the Git commit ID
- `{timestamp}` - Unix timestamp of the time of the release
- `{datetime}` - date and time of the release (14 characters, for example, `20211213000730`)

> Note: because we are using `{tag}` as placeholder, we can't call `--snapshot` with an empty tag name - this will result in error.

Rajdeepc (approved):
LGTM! Thanks for bringing this in!

.changeset/config.json
castastrophe: What's nice too is I think the tag.datetime syntax matches how Garth did the snapshots for foundations in the spectrum-tokens repo too! Cross-team synchronicity!

.github/workflows/beta-release.yml
```diff
@@ -21,13 +21,11 @@ jobs:
                   git config --global user.email "support+actions@github.com"
                   git config --global user.name "github-actions-bot"
 
-            - name: Update package versions for beta release
+            - name: Update package versions for beta snapshot release
               env:
                   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
               run: |
-                  yarn changeset pre enter beta
```
castastrophe, line 28: When do you activate the beta mode state if you don't do it here? I didn't see this being done elsewhere -- or is it wrapped in a command now?

.changeset/config.json
caseyisonit: @rubencarvalho what do you think about the tag being the PR number or the number is prepended the custom tag?This would make it easier to understand which beta snapshots are associated with PR fixes.

.github/workflows/beta-release.yml
caseyisonit: @rubencarvalho +1 on needing more clarity here

blunteshwar (approved):

.changeset/config.json
castastrophe: @caseyisonit There's no asset for PR number because there's no guarantee every tag will be connected to a PR but could use the commit.

.github/workflows/beta-release.yml
caseyisonit: @castastrophe through my pairing with piyush on his other PR I now have a bit more context on this that I'm happy to share. 

So essentially Rubens PR is suggesting using snapshot releases as a work around to the beta releases since running beta releases against main is warned against by changesets and we don't use a backup branch. 

> Warning: If you decide to do prereleases from the main branch of your repository, without having a branch for your last stable release without the prerelease changes, you will block other changes until you are ready to exit prerelease mode. We thoroughly recommend only running prereleases from a branch other than the main branch.

The other issue we were having is that beta releases weren't counting correctly due to leaving prerelease when it is merged. the pre.json doesn't persist so it wont override the latest version in the package.json. 

I think this is a good temporary solution for providing releases to consumers to confirm the changes work.

Hopefully this helps make more sense out of what is being suggested.

rubencarvalho: Hey! This is a great question — I should’ve added a clarifying comment here. There are 3 main reasons:
1. We do not keep a living `beta` branch nor do we sync the newly generated beta version back to any other branch for that matter. As a result, when this workflow runs, it always calculates the same first version, e.g.: 1.5.0-beta.**0**
2. By not using the prerelease mode and instead using snapshots, we can generate a unique identifier (timestamp-based), which avoids this issue without the overhead of managing a dedicated branch or modifying our existing workflows (that said, I do think there’s value in exploring a `beta` branch strategy or something along those lines in the future.)
3. There is a somewhat scary warning on the changesets documentation about using prerelease mode in `main`:
Screenshot 2025-04-11 at 22 03 24

.github/workflows/beta-release.yml
```diff
@@ -37,12 +35,12 @@ jobs:
...
+            - name: Publish beta snapshot release
               env:
                   NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
               run: |
                   git add .
-                  git commit -am "chore: publish beta version"
+                  git commit -am "chore: publish beta snapshot version"
                   yarn prepublishOnly
-                  yarn changeset publish --no-git-tag
+                  yarn changeset publish --no-git-tag --tag beta
```
caseyisonit, line 45:
```suggestion
                  yarn changeset publish --no-git-tag --tag beta-${GITHUB_SHA::8}
```

This will append the 8-digit SHA of the merge commit going in to main. this is provided TO changesets and would be in the {tag} position in the template. I tested this in #5447 to make sure the tag would compile correctly. Screenshot of this suggestions output.

Screenshot 2025-05-06 at 5 04 47 PM

caseyisonit: applied this to the wrong line. this tag is associated with NPM tags, moving this suggestion up to line 28

.github/workflows/beta-release.yml
```diff
@@ -21,13 +21,11 @@ jobs:
...
 
-            - name: Update package versions for beta release
+            - name: Update package versions for beta snapshot release
               env:
                   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
               run: |
-                  yarn changeset pre enter beta
-                  # Apply the changeset with specific beta tag
-                  yarn changeset version
+                  yarn changeset version --snapshot beta
```
caseyisonit, line 28:
```suggestion
                  yarn changeset version --snapshot beta--${GITHUB_SHA::8}
```

.github/workflows/publish-docs-site.yml
```diff
@@ -11,7 +11,7 @@ jobs:
         name: Build & publish site
         runs-on: ubuntu-latest
         # Run the job if manually triggered or if the commit message includes '#publish' & the check suite has passed or if the commit message includes 'docs' and the check suite has passed
-        if: github.event_name == 'workflow_dispatch' || contains(github.event.head_commit.message, '#publish') || contains(github.event.head_commit.message, 'docs')
+        if: github.event_name == 'workflow_dispatch' || contains(github.event.head_commit.message, '#publish') || contains(github.event.head_commit.message, 'docs:') || contains(github.event.head_commit.message, 'docs(')
```
caseyisonit, line 14: more explicit to avoid a commit message that has the word docs in the message. this ensures it is of the commit type docs

.github/workflows/urls-smoke-test.yml
```diff
@@ -1,4 +1,4 @@
-name: Links, and Smoke Tests
+name: Preview URLs and Smoke Tests
```
caseyisonit, line 1:
```suggestion
name: Review URLs and Smoke Tests
```

PR #5414, opened 2025-04-29
chore: lock playwright version [swc-829]

castastrophe:
## Related issue(s)

-

## How has this been tested?

-   [ ] _Test case 1_
    1. Go here
    2. Do this
-   [ ] _Test case 2_

    1. Go here
    2. Do this

-   [ ] Did it pass in iPad?

## Types of changes

## Checklist

## Best practices

### DIFF
--- a/packages/number-field/test/inputs.test.ts
+++ b/packages/number-field/test/inputs.test.ts
@@ -14,7 +14,6 @@ import { html } from '@spectrum-web-components/base';
 import { elementUpdated, expect, nextFrame } from '@open-wc/testing';
 import { getElFrom } from './helpers.js';
 import { createLanguageContext } from '../../../tools/reactive-controllers/test/helpers.js';
-import { shouldPolyfill } from '@formatjs/intl-numberformat/should-polyfill.js';
 
 import '@spectrum-web-components/number-field/sp-number-field.js';
 import { remapMultiByteCharacters } from '@spectrum-web-components/number-field';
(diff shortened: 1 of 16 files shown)


packages/number-field/test/inputs.test.ts
```diff
@@ -28,23 +27,6 @@ import {
 import { sendKeys } from '@web/test-runner-commands';
 
 describe('NumberField - inputs', () => {
-    before(async () => {
```
castastrophe, line 31: CanIUse indicates that the international number field has been available in our supported evergreen - 2 so we can remove the polyfills now

tools/reactive-controllers/src/LanguageResolution.ts
```diff
@@ -19,7 +19,18 @@ export const languageResolverUpdatedSymbol = Symbol(
 
 export class LanguageResolutionController implements ReactiveController {
     private host: ReactiveElement;
-    language = document.documentElement.lang || navigator.language;
+    get language(): string {
+        return document.documentElement.lang || navigator.language;
```
castastrophe, line 23: This is fetching the root-level language but shouldn't we also be honoring if the language attribute is set on a container too?

caseyisonit: this change is what causes a majority of the number field failing tests, I would like to revert it since its out of scope for this specifically and capture this idea somehow for when we refactor number field.

caseyisonit: ok looks like you made some edits since i started and ill retest

castastrophe: Those tests were failing before as well - I added this change to resolve the first round of issues I saw. You can see the details outlined in the Jira ticket.

> The first error I identified was on NumberField which threw an error that locale was invalid on every instance. I fixed this according to similar debugging advice I found online by converting any manually provided locales with underscores to dashes.

caseyisonit: thats odd, when i reverted it the tests passed. I'm going to do a fresh clone again to make sure it wasn't something weird on my local. thank you for this context and apologies I missed that from the ticket.

PR #5430, opened 2025-05-05
tooling: update pre-commit hook to stop failure propagation of no-import rule of missing dependencies

Rajdeepc:
## Description

This PR updates the `pre-commit` hook to reliably lint, analyze, and format only the staged files using ESLint, Lit Analyzer, Stylelint, and Prettier. The script has been rewritten as a standalone shell script with strict error handling using `set -e`. It ensures that the commit process is blocked if any linting or analysis tool fails. Additionally, `genversion` is executed and the resulting file is re-staged for inclusion in the commit.

## Related issue(s)

* N/A

## Motivation and context

Previously, the `pre-commit` hook used a `bash << EOF` heredoc block, which caused failures in expected linting behavior. Errors from linters were not blocking commits as expected due to the lack of strict failure propagation in the heredoc context. This update resolves that by using a standard shell script with correct error handling and scoped linting to staged files only.

## How has this been tested?

Manually tested in a local development environment using the following steps:

* [x] *Test case 1*

  1. Make a change to a `.ts` file that introduces an ESLint error.
  2. Attempt to commit.
  3. Verify that the commit is blocked with the lint error shown.

* [x] *Test case 2*

  1. Make changes to `.css` and `.ts` files without lint issues.
  2. Attempt to commit.
  3. Verify that the commit goes through and the version file is updated and re-staged.

* [x] *Test case 3*

  1. Remove `@spectrum-web-components/reactive-controller` from the dependencies in `packages/icon/package.json`.
  2. Attempt to commit.
  3. Verify that the commit  is blocked with the lint error shown.
  4. `import/no-extraneous-dependencies` rule is blocking the commit due to missing dependencies

* [x] Did it pass in Desktop?

* [x] Did it pass in Mobile?

* [x] Did it pass in iPad?

N/A

## Types of changes

- [x] Chore (minor updates related to the tooling or maintenance of the repository, does not impact compiled assets)

## Checklist

- [ ] If my change required a change to the documentation, I have updated the documentation in this pull request.
- [ ] I have read the **[CONTRIBUTING](<(https://github.com/adobe/spectrum-web-components/blob/main/CONTRIBUTING.md)>)** document.


packages/icon/src/IconBase.ts
```diff
@@ -33,7 +33,6 @@ export class IconBase extends SpectrumElement {
     public static override get styles(): CSSResultArray {
         return [iconStyles];
     }
-
```
castastrophe, line 36: Did you mean to remove this whitespace? It seems useful to delineate the two properties.

packages/icon/package.json
```diff
@@ -88,7 +88,8 @@
     ],
     "dependencies": {
         "@spectrum-web-components/base": "1.6.0",
-        "@spectrum-web-components/iconset": "1.6.0"
+        "@spectrum-web-components/iconset": "1.6.0",
+        "@spectrum-web-components/reactive-controllers": "1.6.0"
```
castastrophe, line 92: good catch here

.husky/pre-commit
```diff
@@ -1,16 +1,16 @@
-#!/usr/bin/env bash
-bash << EOF
-    STAGED_FILES_TO_LINT=$(git diff --name-only --cached --diff-filter=d -- "*.ts" "*.js")
-    STAGED_FILES_TO_ANALYZE=$(git diff --name-only --cached --diff-filter=d -- "{packages,tools}/*/src/**/!(*.css).ts")
-    STAGED_CSS_FILES=$(git diff --name-only --cached --diff-filter=d -- "{packages,tools}/**/*.css")
-    VERSION_FILE=$(dirname "$0")/../tools/base/src/version.js
+#!/bin/bash
```
castastrophe, line 1: Yeah I've noticed this issue as well, however here's the documentation around why we added the bash wrapper: https://typicode.github.io/husky/how-to.html#bash

We're not currently supporting Windows so it seemed a fine compromise. The if syntax however _is_ Bash syntax so removing that would be a potential issue. What do you suggest?

castastrophe: I was thinking we could migrate all these rules into a lint-staged config and just run the `yarn lint-staged` command here instead of all this logic. What are your thoughts? Was playing around with that here: https://github.com/adobe/spectrum-web-components/pull/5393

castastrophe (changes requested):
Left a few questions! Would suggest we reword the changeset before merging too in order for it to read more like a sentence than a commit message.

packages/icon/src/IconBase.ts
Rajdeepc: Ah! I was testing the linting with some additional fail checks here. Should be removed. Thanks for catching this

.husky/pre-commit
Rajdeepc: I like the idea of migrate these in a module in a lint-staged config. More scalable. Let me update this here and test. Please feel free to add your thoughts too!

lint-staged.config.js
```diff
@@ -0,0 +1,46 @@
...
+    '!(*.css|*.ts)': [
+        'prettier --cache --no-error-on-unmatched-pattern --ignore-unknown --log-level silent --write',
+    ],
+
+    // Generates and add version file
+    // This will always re-run and stage version.js
+    'tools/base/src/version.js': () => {
+        console.log('Generating version file...');
+        execSync('yarn genversion --es6 --semi tools/base/src/version.js');
+        return 'git add tools/base/src/version.js';
```
castastrophe, line 44:
```suggestion
    'tools/base/src/version.js': (file) => [
        `genversion --es6 --semi ${file}`
        `git add ${file}`,
    ],
```

This approach will still run them in parallel but it avoids using the exec function which is really problematic in scripts due to it's inconsistent way of returning warnings and information from the console.

lint-staged.config.js
```diff
@@ -0,0 +1,46 @@
...
+Unless required by applicable law or agreed to in writing, software distributed under
+the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+OF ANY KIND, either express or implied. See the License for the specific language
+governing permissions and limitations under the License.
+*/
+import { execSync } from 'child_process';
+
+export default {
+    // Runs ESLint with fix on all staged JS/TS files
+    '*.{js,ts}': (files) => {
```
castastrophe, line 17: So lint-staged is great because it actually shows you the files it's running against so you don't need to echo the file list. It also enables parallelization so unless the command requires it, you don't have to pass the files in manually.

For example:

```
"*.js": ["eslint", "prettier --write"]
```

lint-staged.config.js
```diff
@@ -0,0 +1,46 @@
...
+    '{packages,tools}/*/src/**/!(*.css).ts': (files) => {
+        console.log('Running lit-analyzer on:', files.join(' '));
+        return [`lit-analyzer ${files.join(' ')}`];
+    },
+
+    // Runs stylelint with fix on all staged CSS files
+    '{packages,tools}/**/*.css': (files) => {
+        console.log('Running stylelint on:', files.join(' '));
+        return [`stylelint --fix ${files.join(' ')}`];
+    },
```
castastrophe, line 32:
```suggestion
    '*.css': ['stylelint --fix'],
```

Since stylelint has rules about which css files it will run against, we can pass any css file from the list and it will only run if it matches the rules. This lets us manage the CSS rules from the stylelint config instead of having to replicate it here too.

package.json
```diff
@@ -41,6 +41,7 @@
         "icons": "wireit",
         "icons:ui": "wireit",
         "icons:workflow": "wireit",
+        "lint-staged": "lint-staged",
```
castastrophe, line 44: I'm not sure we benefit from this abstraction, can we look at just using the migration work I already put together? 

https://jira.corp.adobe.com/browse/SWC-855

https://github.com/adobe/spectrum-web-components/pull/5393

packages/.eslintrc.json
```diff
@@ -6,13 +6,26 @@
...
     ],
     "rules": {
+        "import/no-extraneous-dependencies": [
+            "error",
+            {
+                "devDependencies": false,
+                "optionalDependencies": false,
+                "peerDependencies": false
+            }
+        ],
```
Rajdeepc, line 28: Added a import dependency rule to check for missing direct dependencies.

Rajdeepc:
Resolved in `main`

PR #5457, opened 2025-05-07
chore: add conditional check to preview release

caseyisonit:
## Description

This adds a conditional check to the `preview-release` workflow to only run on commit types `feat`,`fix`, or `!:`. The `!:` will capture any breaking change.

## Motivation and context

- Reduce the number of preview releases published
- Only release functional testable code 

## Related issue(s)

-   relates to PR #5365

## Author's checklist

---

## Reviewer's checklist

-   [ ] The conditional logic covers all semver edge cases that should release a preview version

### DIFF
--- a/.github/workflows/preview-release.yml
+++ b/.github/workflows/preview-release.yml
@@ -8,7 +8,8 @@ on:
 jobs:
     release:
         runs-on: ubuntu-latest
-
+        # Run the job if manually triggered or if the commit message includes a commit type of fix or feat
+        if: contains(github.event.head_commit.message, 'fix:') || contains(github.event.head_commit.message, 'fix(') || contains(github.event.head_commit.message, 'feat:') || contains(github.event.head_commit.message, 'feat(')
         steps:
             - name: Checkout PR branch
               uses: actions/checkout@v4
--- a/.github/workflows/preview-release.yml
+++ b/.github/workflows/preview-release.yml
@@ -8,8 +8,8 @@ on:
 jobs:
     release:
         runs-on: ubuntu-latest
-        # Run the job if manually triggered or if the commit message includes a commit type of fix or feat
-        if: contains(github.event.head_commit.message, 'fix:') || contains(github.event.head_commit.message, 'fix(') || contains(github.event.head_commit.message, 'feat:') || contains(github.event.head_commit.message, 'feat(')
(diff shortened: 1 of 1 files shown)


.github/workflows/preview-release.yml
```diff
@@ -8,7 +8,8 @@ on:
 jobs:
     release:
         runs-on: ubuntu-latest
-
+        # Run the job if manually triggered or if the commit message includes a commit type of fix or feat or breaking change
+        if: contains(github.event.head_commit.message, 'fix:') || contains(github.event.head_commit.message, 'fix(') || contains(github.event.head_commit.message, 'feat:') || contains(github.event.head_commit.message, 'feat(') || contains(github.event.head_commit.message, '!:')
```
caseyisonit, line 12: using the `:` and `(` ensures it is a commit type and not just a random word in the message

nikkimk (approved):
Lawless genius twirls mustache (LGTM)

graynorton (commented):
This looks like it will work, and I'm not necessarily opposed to going with this approach, but it bothers me a little bit to be relying on conventional commits for the publish-or-not logic, since we have moved in the direction of having Changesets be our source of truth for representing release-worthy changes.

What if, instead of conditionalizing the logic inside the job and inspecting commits, we were to only run the job when a push to main includes changes beneath the `.changesets` directory, something like this:

```
on:
  push:
    branches:
      - main
    paths:
      - '.changesets/**'
```

I think we could also add `workflow_dispatch:` to the `on:` block if we wanted to allow manually triggering the action from the GitHub UI.

caseyisonit:
> I think we could also add `workflow_dispatch:` to the `on:` block if we wanted to allow manually triggering the action from the GitHub UI.

we provide the `changsets-snapshot-publish` command for this use case since there wouldn't be a git sha associated from a workflow_dispatch

graynorton:
> we provide the `changsets-snapshot-publish` command for this use case since there wouldn't be a git sha associated from a workflow_dispatch

Ah, I hadn't thought about the missing sha issue!

The use case I had in mind was not an one-off snapshot release (which I agree `changesets-snapshot-publish` covers), but one where something had gone wrong and an automated preview build had been skipped or failed, and we wanted to essentially say "make an official preview release, but trust me and skip the changesets check". In this case, I imagined that the sha from the `HEAD` of `main` would be used, but I didn't think about where it would come from.

Anyway, it may be that we don't need to worry about that case.

Issue #5515, opened 2025-06-05
[Bug]: Swc Cards are  very much touch sensitive for click events, because of which scroll behavior as well causing card click triggers.


kusumadevadiga:
### Code of conduct

- [x] I agree to follow this project's code of conduct.

### Impacted component(s)

swc-react/card

### Expected behavior

Only click behavior should trigger onClick action and, scroll or simple touch should not cause click events.
ref: https://stackblitz.com/edit/vitejs-vite-xqza59ed?file=src%2Fmy-element.ts 

### Actual behavior

Swc cards are sending click events for a scroll .

### How can we reproduce this issue?

Open https://www.stage.adobe.com/community on Mobile device or any touch device
Scroll bottom for adobe events section or behance card section
Try to scroll events card from right to left
Cards gets opened in new tab which is click action. Ideally, cards should be scrolled and do not trigger click action

Or 

1. go to https://stackblitz.com/edit/vitejs-vite-xqza59ed?file=src%2Fmy-element.ts 
2. try to scroll

### Severity

None

### Logs taken while reproducing problem

https://stackblitz.com/edit/vitejs-vite-xqza59ed?file=src%2Fmy-element.ts 

### Would you like to track this issue in Jira?

- [x] Yes, please tell me the ticket number!

Rajdeepc:
SWC-913

PR #5521, opened 2025-06-06
fix(card): replaced the time-based click detection with a distance-based approach

Rajdeepc:
## Description
Replaced the time-based click detection with a distance-based approach:
- Added tracking of initial touch coordinates (X and Y)
- Introduced a 10px movement threshold to distinguish between clicks and scrolls
- Added `pointermove` event handling to detect actual scrolling
- Improved event listener cleanup

## Motivation and context
On mobile Chrome (both Android and iOS), scrolling on `sp-card` components would inadvertently trigger click events. This was caused by the timing-based click detection (200ms threshold) in the pointer event handling, which could misinterpret quick scrolls as clicks. This issue did not affect Safari on mobile devices.

### Root Cause
The `pointerdown` handler is capturing all touch starts, and then manually measuring the duration of the interaction. If the duration is less than `200ms`, we are triggering a `.click()` on the card:

```ts
const start = +new Date();
const handleEnd = (): void => {
    const end = +new Date();
    if (end - start < 200) {
        this.click();
    }
```
The problem is: mobile Chrome often doesn't distinguish well between taps and touch scrolls unless there's actual movement. So if the user slightly touches and scrolls (even without crossing a threshold), it registers as a tap and triggers `.click()`.

### Technical Details
- Changed `handlePointerdown` to use `PointerEvent` instead of generic `Event`
- Added `startX`, `startY`, and `isScrolling` properties to track touch state
- Introduced `SCROLL_THRESHOLD` constant (10px) for movement detection
- Added proper cleanup of all event listeners
- Maintained existing behavior for anchor elements and desktop browsers

## Related issue(s)

-   fixes SWC-913
-   fixes https://github.com/adobe/spectrum-web-components/issues/5515

---

https://github.com/user-attachments/assets/861d2082-e78b-49c6-aa29-d75e82751962

## Author's checklist

---

## Reviewer's checklist

### Manual review test cases

-   [ ] Scroll Issue test in Chrome in Android

    1. Go [here](https://rajdeep-card-chrome-mobile-bug--spectrum-wc.netlify.app/storybook/?path=/story/card--scroll-test)
    2. Open Chrome in Android
    3. Change to Mobile View in an actual simulator or device
    4. Scroll through the card list
    5. Check the click event doesn't fire anymore while scrolling

-   [ ] Scroll Issue test in Chrome in iOS
    1. Go [here](https://rajdeep-card-chrome-mobile-bug--spectrum-wc.netlify.app/storybook/?path=/story/card--scroll-test)
    2. Open Chrome in iOS
    3. Change to Mobile View in an actual simulator or device
    4. Scroll through the card list
    5. Check the click event doesn't fire anymore while scrolling

### Device review


packages/card/src/Card.ts
```diff
@@ -186,21 +186,46 @@ export class Card extends LikeAnchor(
         }
     }
 
-    private handlePointerdown(event: Event): void {
+    /**
+     * Handles pointer down events on the card element.
```
castastrophe, line 190: Love the addition of these comments!

packages/card/src/Card.ts
```diff
@@ -186,21 +186,46 @@ export class Card extends LikeAnchor(
...
         const hasAnchor = path.some(
             (el) => (el as HTMLElement).localName === 'a'
         );
         if (hasAnchor) return;
-        const start = +new Date();
-        const handleEnd = (): void => {
-            const end = +new Date();
-            if (end - start < 200) {
+        // Record the time and initial position of the pointerdown event
+        const startTime = +new Date();
```
castastrophe: Wow! No idea what `+new Date()` did! TIL! That said, since it's a bit of a hack since JS is loosely typed, should we replace this with `Number(new Date())` and/or perhaps a comment that this is converting the date to a numeric string?

packages/card/src/Card.ts
```diff
@@ -186,21 +186,46 @@ export class Card extends LikeAnchor(
...
+        const startX = event.clientX;
+        const startY = event.clientY;
```
castastrophe, line 203: Do we have (or need) any error handling for if the event object is undefined?

packages/card/stories/card.stories.ts
```diff
@@ -342,7 +343,7 @@ export const smallQuiet = (args: StoryArgs): TemplateResult => {
     return html`
         <div>
             <sp-card
-                size=${args.size}
+                size=${ifDefined(args.size)}
```
castastrophe, line 346: Is the default size medium or is default considered `:not([size])`?

packages/card/test/card.test.ts
```diff
@@ -57,95 +43,92 @@ describe('card', () => {
...
-            `
-        );
+                    <sp-menu>
+                        <sp-menu-item>Deselect</sp-menu-item>
+                        <sp-menu-item>Select Inverse</sp-menu-item>
+                        <sp-menu-item>Feather...</sp-menu-item>
+                        <sp-menu-item>Select and Mask...</sp-menu-item>
+                        <sp-menu-divider></sp-menu-divider>
+                        <sp-menu-item>Save Selection</sp-menu-item>
+                        <sp-menu-item disabled>Make Work Path</sp-menu-item>
```
castastrophe, line 110:
```suggestion
                        <sp-menu-item>Deselect</sp-menu-item>
                        <sp-menu-item>Select inverse</sp-menu-item>
                        <sp-menu-item>Feather...</sp-menu-item>
                        <sp-menu-item>Select and mask...</sp-menu-item>
                        <sp-menu-divider></sp-menu-divider>
                        <sp-menu-item>Save selection</sp-menu-item>
                        <sp-menu-item disabled>Make work path</sp-menu-item>
```

I know this is just an indent update but should we take the opportunity to fix the sentence case here anyway?

packages/card/test/card.test.ts
```diff
@@ -57,95 +43,92 @@ describe('card', () => {
...
-                    <div slot="description">10/15/18</div>
-                    <div slot="footer">Footer</div>
-                </sp-card>
-            `
-        );
+        const el = await fixture<Card>(html`
+            <sp-card variant="gallery" heading="Card Heading" subheading="JPG">
+                <img
+                    slot="preview"
+                    src="https://picsum.photos/532/192"
```
castastrophe, line 125: A future suggestion, I've seen CI issues with using API services to render images, I think we should consider having a few local assets for testing.

castastrophe (approved):
No blockers but a few questions and suggestions added.

packages/card/src/Card.ts
rubencarvalho: +1 on TIL! 😄 Even though I like it, I think the most idiomatic approach here is to rely on `event.timeStamp` for measuring the delta between the two events.

packages/card/src/Card.ts
rubencarvalho: These events are bound natively without any argument, so the browser always passes an `Event` object as the first argument. But I agree it wouldn’t hurt to wrap this logic in a `try`/`finally` block (with the event listener removals in the `finally`). That way, if any line ever throws (for any reason?), at least we don't throw and still clean up the dangling handlers.

packages/card/src/Card.ts
Rajdeepc: Going with the appreciated choice of the people.

packages/card/test/card.test.ts
Rajdeepc: Will add this in a follow up PR. Not a part of this change

packages/card/test/card.test.ts
Rajdeepc: Sure will take this as a part of the clean up!

packages/card/stories/card.stories.ts
Rajdeepc: SizedMixin is configured with noDefaultSize: true, which means there is no default size,  it's considered :not([size]) when no size is specified

packages/card/src/Card.ts
```diff
@@ -198,13 +198,13 @@ export class Card extends LikeAnchor(
...
         // Record the time and initial position of the pointerdown event
-        const startTime = +new Date();
+        const startTime = event.timeStamp;
         const startX = event.clientX;
         const startY = event.clientY;
 
         // Define the handler for when the pointer interaction ends
         const handleEnd = (endEvent: PointerEvent): void => {
-            const endTime = +new Date();
+            const endTime = event.timeStamp;
```
rubencarvalho: `endEvent.timeStamp`

rubencarvalho (approved):

Issue #5541, opened 2025-06-13
[Bug]: Layout thrashing in ColorArea component


Rajdeepc:
### Code of conduct

- [x] I agree to follow this project's code of conduct.

### Impacted component(s)

ColorArea

### Expected behavior

ColorArea should not leak memory

### Actual behavior

There are style recalculations happening in `render` method accessing `getboundingClientRect` during render.

### Screenshots

![Image](https://github.com/user-attachments/assets/73bce776-af85-489d-88b5-8e9107b02eea)

### What browsers are you seeing the problem in?

Chrome, Firefox, Safari

### How can we reproduce this issue?

1. Go to https://opensource.adobe.com/spectrum-web-components/components/color-area/#standard
2. Open Memory profiling in Chrome
3. Take a snapshot of heap memory usage
4. Check image below
5. See memory leaks

### Sample code or abstract reproduction which illustrates the problem

```ts
protected override render(): TemplateResult {
    const { width = 0, height = 0 } = this.boundingClientRect || {};
    // ...
    style=${`transform: translate(${
        (this.isLTR ? this.x : 1 - this.x) * width
    }px, ${height - this.y * height}px);`}
```

```ts
this.observer = new (window as unknown as WithSWCResizeObserver).ResizeObserver(
    (entries: SWCResizeObserverEntry[]) => {
        for (const entry of entries) {
            this.boundingClientRect = entry.contentRect;
        }
        this.requestUpdate();  // This triggers a full re-render
    }
);
```

### Severity

None

### Would you like to track this issue in Jira?

- [ ] Yes, please tell me the ticket number!

najikahalsema:
SWC-927

PR #5550, opened 2025-06-18
fix(colorarea): remove layout thrashing and cache selectors

tjgupta:
Remove layout invalidation from ColorArea and cache selectors

## Description

This change removes layout invalidation/recalculation from the ColorArea when adjusting the current color, either with the mouse or with keyboard input. The change also caches the selectors for the queries so they don't need to be recalculated every time they're accessed.

## Motivation and context

The layout thrashing had some negative performance impact, most noticeable in Safari on Intel based Macs.

## Related issue(s)

-   fixes https://github.com/adobe/spectrum-web-components/issues/5541
-   resolves [SWC-927]

**Before:**
Screenshot 2025-06-17 at 11 27 45 PM

**After:**
Screenshot 2025-06-18 at 11 12 54 AM

---

## Author's checklist

---

## Reviewer's checklist

### Manual review test cases

    1. Go [here](url)
    2. Do this action
    3. Expect this result

    1. Go [here](url)
    2. Do this action
    3. Expect this result

### Device review


packages/color-area/src/ColorArea.ts
```diff
@@ -562,18 +556,6 @@ export class ColorArea extends SpectrumElement {
...
-                's',
-                this.inputX.valueAsNumber * 100
-            );
-        }
-        if (this.y !== this.inputY.valueAsNumber) {
-            this.colorController.color.set(
-                'v',
-                (1 - this.inputY.valueAsNumber) * 100
-            );
-        }
```
tjgupta: I'm not entirely clear on why we're doing this here again, as we already set these in the various setters above. Also, the y value is setting the value differently than above (line 574), which seems incorrect. I don't see any side effects to removing this altogether.

caseyisonit:
@tjgupta can you please include detailed testing instructions? Also should this cause a visual change at all, I'm seeing a large number of VRT failures? 

caveat: please don't update the golden hash until a second reviewer gives a thumbs up if visual changes are expected

caseyisonit:
Also can you added an automated test if possible to cover this code?

Rajdeepc:
Thanks for your contribution @tjgupta . Can you please add a script to set x and y rapidly like below.
```ts
const colorArea = document.querySelector('sp-color-area');
for (let i = 0; i <= 1; i += 0.01) {
    colorArea.x = i;
    colorArea.y = 1 - i;
}
```
Let's observe for layout (reflow/paint) spikes.

packages/color-area/src/ColorArea.ts
Rajdeepc: Removing these lines delegates full responsibility for color state updates to the x and y property setters which introduces a tighter coupling between property updates and state synchronisation. 
If, for any reason, the input elements and the internal color state become unsynchronised, the component may exhibit inconsistent behavior—potentially leading to visual regressions. 
This constitutes a breaking change in the component's internal state management logic and can have both visual and functional regressions.Your tests are failing bcz of the removal of this block.

We would need to a thorough audit of the all the state updates to ensure we are covering every aspects. We can add this as a enhancement but I would not block this PR due to this optimisation.

Rajdeepc (changes requested):

packages/color-area/src/ColorArea.ts
tjgupta: I removed this part of the change for now, but as we started to discuss separately, I think it's worth considering whether it makes sense to have roughly the same state stored and updated in multiple places (and accidentally also in multiple ways).

In my opinion it's better to have only one representation of the color value stored in one piece of reactive state, and one way to update it. This state should be used to render the handle position (rather than querying the DOM) for best performance. The same state could also be used to derive the x/y values for the a11y range inputs.

I suspect (only a theory as of yet) that the fact that the color is represented both as a color in the ColorController, and sort of also as x and y coordinates in this ColorArea component (and that they're updated multiple times in different ways) are a source of some strange behavior, especially on slower browsers/devices or when the browser is under some additional load. See the end of this video for what I mean...

https://github.com/user-attachments/assets/8f123067-790a-43ab-96d0-143c7bdca164

packages/color-area/src/ColorArea.ts
```diff
@@ -65,7 +65,7 @@ export class ColorArea extends SpectrumElement {
     @property({ type: String, attribute: 'label-y' })
     public labelY = 'luminosity';
 
-    @query('.handle')
+    @query('.handle', true)
```
Rajdeepc, line 68: The second argument to @query sets the required option. If the element is not found, it will throw.
If .handle, [name="x"], or [name="y"] elements are missing from the DOM at any point, your code will throw, potentially breaking the component.

packages/color-area/src/ColorArea.ts
```diff
@@ -126,57 +126,51 @@ export class ColorArea extends SpectrumElement {
...
-            this.inputX.value = x.toString();
-            this.colorController.color.set(
-                's',
-                this.inputX.valueAsNumber * 100
-            );
-        } else {
-            this.colorController.color.set('s', x * 100);
-        }
+        this._x = newValue;
+        this.colorController.color.set('s', newValue * 100);
```
Rajdeepc, line 144: If inputX or inputY had any extra browser-level validation logic it will be bypassed. UI and internal state may get out of sync if something else updates the input elements directly.
Can you add tests to confirm that the component and the UI never get out of sync.

packages/color-area/src/ColorArea.ts
```diff
@@ -126,57 +126,51 @@ export class ColorArea extends SpectrumElement {
...
         this.requestUpdate('x', oldValue);
     }
 
+    private _y = 1;
```
Rajdeepc, line 148: I see the public getter returns values derived from this.colorController.color.hsv. 
If this.colorController.color.set() is called elsewhere, the cache (_x/_y) and the color model may become inconsistent. You may have two sources of truth for the same value.

Instead consider always deriving x/y from the color model and using the setters only to mutate the model. You may not need _x and _y at all

packages/color-area/src/ColorArea.ts
```diff
@@ -126,57 +126,51 @@ export class ColorArea extends SpectrumElement {
...
     @property({ type: Number })
     public get y(): number {
         return this.colorController.color.hsv.v / 100;
     }
 
     public set y(y: number) {
-        if (y === this.y) {
+        const oldValue = this._y;
+        // Restrict y to the nearest step
+        const newValue = Math.round(y / this.step) * this.step;
```
Rajdeepc, line 158: Confirm that this rounding logic aligns with user expectations and the behavior of any input controls in the UI

packages/color-area/src/ColorArea.ts
```diff
@@ -126,57 +126,51 @@ export class ColorArea extends SpectrumElement {
...
+    private _x = 1;
+
     @property({ type: Number })
     public get x(): number {
         return this.colorController.color.hsv.s / 100;
     }
 
     public set x(x: number) {
-        if (x === this.x) {
+        const oldValue = this._x;
```
Rajdeepc, line 137: Why are taking cached values, why not take directly from the controller?

Rajdeepc (changes requested):

packages/color-area/src/ColorArea.ts
tjgupta: The second argument is setting the `cache` option. This is so that it doesn't have to query the DOM each time we access `this.handle`. Is there a condition you're concerned about where this element won't be in the DOM?

tjgupta:
I'm having quite a lot of difficulty understanding, running, and fixing the tests, as well as a bit of difficulty understanding all the cases that this component handles (in my opinion it probably has too large of an API surface area and too many places storing the same state). I'm going to close this PR for now due to time constraints, but feel free to reopen if you're able to fix this performance issue.

PR #5583, opened 2025-07-01
fix: overlay flaky test

Rajdeepc:
## Description

Enhanced the focus handling in the test with more robust timing and retry logic:

1. Increased timeout duration: Extended the initial wait time from 50ms to 100ms to allow more time for click events to register
2. Added retry mechanism: Implemented a retry loop that attempts to focus the textfield up to 3 times if the initial click doesn't work
3. Extended waitUntil timeouts: Increased the waitUntil timeout from 1000ms to 2000ms for the first focus check and from 500ms to 1000ms for the second focus check to accommodate slower CI environments

## Changes

- packages/overlay/test/overlay-trigger-extended.test.ts:

  - Keep the programmatic focus fallback for the initial focus test (before overlay opens) since we just need to establish that the textfield is focusable
  - Remove the programmatic focus fallback after the overlay is closed since we specifically want to verify that:
    - The overlay is truly gone
    - The textfield is actually clickable/not occluded
    - The user can interact with the textfield normally

## Motivation and context
The `overlay-trigger-extended.test.ts` was experiencing intermittent failures in CI environments where clicking on textfield elements wasn't reliably focusing them. The test would sometimes fail because the textfield wasn't receiving focus after mouse clicks, causing the waitUntil assertions to timeout.

## Related issue(s)

-   fixes SWC 473

---

## Author's checklist

---

## Reviewer's checklist

### Device review

### DIFF
--- a/packages/overlay/test/overlay-trigger-extended.test.ts
+++ b/packages/overlay/test/overlay-trigger-extended.test.ts
@@ -251,13 +251,8 @@ describe('Overlay Trigger - extended', () => {
 
         await sendMouseTo(textfield, 'click');
 
-        // If click didn't work, try programmatic focus
-        if (document.activeElement !== textfield) {
-            await textfield.focus();
-        }
+        // verify the textfield is focused and actually clickable
 
-        // verify the textfield is focused
-        // and that textfield is no longer occluded
         await waitUntil(
             () => document.activeElement === textfield,
             `clicking focuses textfield again (active element is ${document.activeElement?.tagName})`,


packages/overlay/test/overlay-trigger-extended.test.ts
```diff
@@ -174,32 +174,40 @@ describe('Overlay Trigger - extended', () => {
...
-        });
+
+        // Wait a bit for the click to register
+        await aTimeout(50);
+
+        // If click didn't work, try programmatic focus
+        if (document.activeElement !== textfield) {
+            textfield.focus();
+            await aTimeout(50);
+        }
```
TarunAdobe: if we're adding this programmatic focus... do we still need to increase the timeout???🤔

Rajdeepc: Yes focus events are asynchronous in nature. even though it is called synchronously here the focus change can take a moment to propagate through the DOM. This 50ms time gives the browser event loop to process focus change before we check `document.activeElement` again.

caseyisonit: If focus is asynchronous shouldn’t we just await it?

packages/overlay/test/overlay-trigger-extended.test.ts
```diff
@@ -174,32 +174,40 @@ describe('Overlay Trigger - extended', () => {
...
-            connected: textfield.isConnected,
-            visible: textfield.offsetParent !== null,
-            disabled: textfield.disabled,
-            tabIndex: textfield.tabIndex,
-        });
+
+        // Wait a bit for the click to register
+        await aTimeout(50);
+
+        // If click didn't work, try programmatic focus
```
rubencarvalho, line 198: Why wouldn't the click work? Is the timeout above not enough? 😞

packages/overlay/test/overlay-trigger-extended.test.ts
```diff
@@ -247,12 +255,21 @@ describe('Overlay Trigger - extended', () => {
 
         await sendMouseTo(textfield, 'click');
 
+        // Wait a bit for the click to register
+        await aTimeout(50);
+
+        // If click didn't work, try programmatic focus
```
rubencarvalho: Before the overlay is opened, I think it’s totally fine to use programmatic focus to assert that the textfield is focused. But doing the same after the overlay is open feels like it could cause the last assertion (below) to always pass, even if the overlay is still occluding the textfield.

packages/overlay/test/overlay-trigger-extended.test.ts
rubencarvalho: Looking at this, given that we're already using `waitUntil` below, this and the other timeouts are redundant? If the focus state propagates slowly, waitUntil will catch it.

Rajdeepc: Yes of course that's one way of handling it.

packages/overlay/test/overlay-trigger-extended.test.ts
Rajdeepc: So while the click should work most of the time, having this fallback pattern makes the test more robust across different browsers and CI environments without sacrificing the intent to test real user interactions first.

There are many times `click` might not work since we are using `sendMouseTo()` which creates synthetic mouse events which are not exactly same as browser events and another case which I saw is during race conditions.
The click events are async - it is sent through executeServerCommand() which the browser needs time to process, update the DOM and the focus state which different browsers handle differently.


This is a fall back strategy what you are seeing. 

```ts
// Try click first
 await sendMouseTo(textfield, 'click');
 
 // If click didn't work, use programmatic focus
 if (document.activeElement !== textfield) {
     await textfield.focus();
 }
 
 // Wait for either method to take effect
 await waitUntil(
     () => document.activeElement === textfield,
     `clicking focuses textfield...`,
     { timeout: 1000 }
 );
   
```

packages/overlay/test/overlay-trigger-extended.test.ts
Rajdeepc: Good point @rubencarvalho We are unfortunately bypassing the occlusion check.

PR #5604, opened 2025-07-14
chore: fix flaky overlay occlude test

caseyisonit:
## Related issue(s)

-   fixes [Issue Number]

---

## Author's checklist

---

## Reviewer's checklist

### Manual review test cases

    1. Go [here](url)
    2. Do this action
    3. Expect this result

    1. Go [here](url)
    2. Do this action
    3. Expect this result

### Device review

### DIFF
--- a/packages/overlay/test/overlay-trigger-extended.test.ts
+++ b/packages/overlay/test/overlay-trigger-extended.test.ts
@@ -214,18 +214,6 @@ describe('Overlay Trigger - extended', () => {
                 />
             </div>
         `);
-        // // Set up the overlay trigger, button, and popover for the test
-        // const { overlayTrigger, button, popover } = await initTest();
-
-        // // Create a textfield that will be placed behind the overlay
-        // const textfield = document.createElement('input');
-        // textfield.type = 'text';
-        // textfield.tabIndex = 0;
-        // textfield.style.position = 'relative';
-        // textfield.style.zIndex = '1';
-
-        // Insert the textfield into the DOM after the overlay trigger
-        // overlayTrigger.insertAdjacentElement('afterend', textfield);
 
         // Get a reference to the textfield element
         const textfield = el.querySelector('#textfield') as HTMLInputElement;
--- a/packages/overlay/test/overlay-trigger-extended.test.ts
+++ b/packages/overlay/test/overlay-trigger-extended.test.ts
@@ -18,13 +18,13 @@ import {
     nextFrame,
     waitUntil,
 } from '@open-wc/testing';
-import '@spectrum-web-components/overlay/overlay-trigger.js';
-import { OverlayTrigger } from '@spectrum-web-components/overlay';
-import '@spectrum-web-components/button/sp-button.js';
 import { Button } from '@spectrum-web-components/button';
-import '@spectrum-web-components/popover/sp-popover.js';
-import { Popover } from '@spectrum-web-components/popover';
+import '@spectrum-web-components/button/sp-button.js';
 import '@spectrum-web-components/dialog/sp-dialog.js';
+import { OverlayTrigger } from '@spectrum-web-components/overlay';
+import '@spectrum-web-components/overlay/overlay-trigger.js';
+import { Popover } from '@spectrum-web-components/popover';
+import '@spectrum-web-components/popover/sp-popover.js';
 import { sendMouse } from '../../../test/plugins/browser.js';
 import { fixture, sendMouseTo } from '../../../test/testing-helpers.js';
 import { overlayClosed, overlayOpened } from './overlay-testing-helpers.js';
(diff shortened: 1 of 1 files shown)


packages/overlay/test/overlay-trigger-extended.test.ts
```diff
@@ -172,92 +172,142 @@ describe('Overlay Trigger - extended', () => {
...
             document.activeElement,
             `textfield cannot be clicked`
         ).to.not.equal(textfield);
 
+        // Close the overlay
         overlayTrigger.open = undefined;
 
-        await overlayClosed(overlayTrigger.clickOverlayElement, 300);
+        // Wait for the overlay to be fully closed
+        overlayClosed(overlayTrigger.clickOverlayElement, 300);
```
Rajdeepc: If `overlayClosed` is async (returns a promise), removing await means your test will not actually wait for the overlay to close before running the next lines. This can cause race conditions and flaky tests.

caseyisonit: i didnt mean to remove it, thanks for catching

packages/overlay/test/overlay-trigger-extended.test.ts
```diff
@@ -172,92 +172,142 @@ describe('Overlay Trigger - extended', () => {
...
         await overlayOpened(overlayTrigger.clickOverlayElement, 400);
 
-        // click the textfield
+        // Attempt to click the textfield while the overlay is open
         await waitUntil(
             async () => await sendMouseTo(textfield, 'click'),
-            `textfield clicked again`,
-            { timeout: 300 }
+            `textfield clicked again`
         );
```
Rajdeepc, line 284: `waitUntil` expects a function that returns a boolean or a promise(boolean) which indicates whether the wait condition is satisfied. Here, `sendMouseTo` returns Promise<void>, not a boolean. So your condition always "passes" as soon as the promise resolves (which is immediately after the click is simulated), not when the textfield is focused.
Continue once the focus is actually on the textfield.
```ts
await sendMouseTo(textfield, 'click');
await waitUntil(
    () => document.activeElement === textfield,
    `textfield should be focused after clicking`
);
```

Rajdeepc (dismissed):

packages/overlay/test/overlay-trigger-extended.test.ts
```diff
@@ -172,92 +172,142 @@ describe('Overlay Trigger - extended', () => {
...
-        if (document.activeElement !== textfield) {
-            await textfield.focus();
-        }
+        expect(textfield.tabIndex, 'textfield is focusable').to.be.equal(0);
 
+        // Focus the textfield by clicking it (simulates user interaction)
         await waitUntil(
-            () => document.activeElement === textfield,
-            `clicking focuses textfield (active element is ${document.activeElement?.tagName})`,
-            { timeout: 1000 }
```
nikkimk, line 206: Without a timeout, these tests will still fail but not give us good error messages to tell us which step failed when it timed out. Instead I recommend increasing the timeout time.

packages/overlay/test/overlay-trigger-extended.test.ts
nikkimk: I'd suggest something like:

```
await sendMouseTo(textfield, 'click');
await waitUntil(
    () => document.activeElement === textfield,
    `textfield should be focused after clicking`,
    { timeout: 1000 }
);
```

nikkimk (changes requested):

packages/overlay/test/overlay-trigger-extended.test.ts
caseyisonit: @nikkimk 1000 is the default timeout for waitUntil https://open-wc.org/docs/testing/helpers/#waituntil

@Rajdeepc can you point me to the docs where you are seeing the return type needs to be a boolean because based on the code in web-test-runner waitUntil actually accepts `unknown | Promise<unknown>` hence why I was setting it up like this. See the screenshot attached for the code in node_modules.

Screenshot 2025-07-15 at 1 13 46 PM

nikkimk (approved):
Lyrics garbled terrible musical

packages/overlay/test/overlay-trigger-extended.test.ts
caseyisonit: im marking this as resolved since the helper does expect a promise and this is working much more reliably

cdransf (approved):

packages/overlay/test/overlay-trigger-extended.test.ts
Rajdeepc: @caseyisonit From the docs: `waitUntil(conditionFn, message?, options?)`

- conditionFn: A function that returns a truthy value or a Promise that resolves to a truthy value.
- message: (optional) Error message to show if condition never becomes true.
- options: (optional) { interval, timeout } in milliseconds.

`waitUntil` is not a wrapper to "try repeatedly until this action succeeds" — it's for polling a condition until it becomes true. This is not useful until sendMouse returns a condition. If this works there must be some underlying code that is satisfying the condition.

Issue #5606, opened 2025-07-15
[Bug]: Picker closes on scrolling on a touch device and selects element instead of scrolling


yfloresdev:
### Code of conduct

- [x] I agree to follow this project's code of conduct.

### Impacted component(s)

picker, menu-item

### Library version

1.4.0 and latest

### Expected behavior

on touch screen devices like iPad and mobile, you should be able to scroll through menu items in a picker dropdown.

### Actual behavior

If you attempt to scroll through menu items in a picker dropdown, the item you first touched when starting to scroll is instead selected and the picker closes.

### Screenshots

https://github.com/user-attachments/assets/c89d110f-1b07-4b96-bb02-a87081de959f

https://github.com/user-attachments/assets/a69d4f79-b4d0-4c49-8aed-8b9d8a329653

### What browsers are you seeing the problem in?

Chrome, Safari

### How can we reproduce this issue?

1. Go to any picker with a large menu item list on touch screen device big enough to not trigger a tray ex. iPad, Samsung galaxy s22
2. open picker and attempt to scroll
3. See error

also repros with simulator and chrome dev tool settings

### Sample code or abstract reproduction which illustrates the problem

https://stackblitz.com/edit/vitejs-vite-r724rwig?file=src%2Fmy-element.ts open on iPad or use chrome dev tool with iPad settings or Xcode simulator 

### Severity

SEV 3

### Would you like to track this issue in Jira?

- [x] Yes, please tell me the ticket number!

Rajdeepc:
SWC 975

caseyisonit:
@yfloresdev was the reproduced on an actual device or only the simulator? I just tried on my phone in both chrome and safari with force-popover on and was able to scroll and select an option without trouble. I'm working on updating my iPad right now to test on my iPad and will update my comment when I've confirmed.

PR #5616, opened 2025-07-21
fix(menu): iPad scrolling issue in picker dropdown

Rajdeepc:
## Summary

Fixes an issue where scrolling through menu items in a picker dropdown on iPad would accidentally select the first touched item and close the picker. This was caused by touch events being interpreted as selection events during scroll gestures.

## Problem

When users attempted to scroll through menu items in a picker dropdown on iPad, the item they first touched when starting to scroll would be selected instead of allowing the scroll gesture to complete. This made the picker unusable on iPad devices.

## Solution

- Added touch event handling to distinguish between scroll gestures and selection gestures:
- Touch tracking: Added properties to track touch start position, time, and scroll state
- Scroll detection: Implemented logic to detect when a touch gesture is a scroll (vertical movement > threshold within time limit)
- Selection prevention: Modified handlePointerBasedSelection to prevent selection when scrolling is detected
- Memory leak prevention: Properly bound and cleaned up touch event listeners in disconnectedCallback

## Technical details

- Added `touchStartY`, `touchStartTime`, `isScrolling` properties for scroll detection
- Implemented `handleTouchStart`, `handleTouchMove`, `handleTouchEnd` methods
- Used `passive: true` for touch event listeners to improve scrolling performance
- Added scroll `threshold (10px)` and time `threshold (300ms)` for scroll detection
- Modified `handlePointerBasedSelection` to check `isScrolling` state before processing selection

## Related issue(s)

-   fixes https://github.com/adobe/spectrum-web-components/issues/5606

Fix

https://github.com/user-attachments/assets/191dc536-cef1-4a4d-b4b5-fa87c3b1a87d

---

## Author's checklist

---

## Reviewer's checklist

### Manual review test cases

    1. Go [here](url)
    2. Do this action
    3. Expect this result

    1. Go [here](url)
    2. Do this action
    3. Expect this result

### Device review

### DIFF
--- a/packages/menu/src/Menu.ts
+++ b/packages/menu/src/Menu.ts
@@ -72,6 +72,13 @@ export class Menu extends SizedMixin(SpectrumElement, { noDefaultSize: true }) {
 
     protected rovingTabindexController?: RovingTabindexController<MenuItem>;
 
+    // Add touch tracking properties for iPad scroll handling
+    private touchStartY = 0;
+    private touchStartTime = 0;
+    private isScrolling = false;
+    private scrollThreshold = 10; // pixels
+    private scrollTimeThreshold = 300; // milliseconds
+
     /**
      * label of the menu
      */
@@ -400,6 +407,17 @@ export class Menu extends SizedMixin(SpectrumElement, { noDefaultSize: true }) {
         this.addEventListener('pointerup', this.handlePointerup);
         this.addEventListener('sp-opened', this.handleSubmenuOpened);
         this.addEventListener('sp-closed', this.handleSubmenuClosed);
+
+        // Add touch event listeners for iPad scroll detection
+        this.addEventListener('touchstart', this.handleTouchStart, {
+            passive: true,
+        });
+        this.addEventListener('touchmove', this.handleTouchMove, {
+            passive: true,
+        });
+        this.addEventListener('touchend', this.handleTouchEnd, {
+            passive: true,
+        });
     }
 
     /**
@@ -443,6 +461,38 @@ export class Menu extends SizedMixin(SpectrumElement, { noDefaultSize: true }) {
         }
     }
 
+    // Add touch event handlers for iPad scroll detection
+    private handleTouchStart(event: TouchEvent): void {
+        if (event.touches.length === 1) {
+            this.touchStartY = event.touches[0].clientY;
+            this.touchStartTime = Date.now();
+            this.isScrolling = false;
+        }
+    }
+
+    private handleTouchMove(event: TouchEvent): void {
+        if (event.touches.length === 1) {
+            const currentY = event.touches[0].clientY;
+            const deltaY = Math.abs(currentY - this.touchStartY);
+            const deltaTime = Date.now() - this.touchStartTime;
+
+            // If we've moved enough vertically and it's been a short time, we're scrolling
+            if (
+                deltaY > this.scrollThreshold &&
+                deltaTime < this.scrollTimeThreshold
+            ) {
+                this.isScrolling = true;
+            }
+        }
+    }
+
+    private handleTouchEnd(): void {
+        // Reset scrolling state after a short delay
+        setTimeout(() => {
+            this.isScrolling = false;
+        }, 100);
+    }
+
     // if the click and pointerup events are on the same target, we should not
     // handle the click event.
     private pointerUpTarget = null as EventTarget | null;
@@ -475,6 +525,11 @@ export class Menu extends SizedMixin(SpectrumElement, { noDefaultSize: true }) {
     private async handlePointerBasedSelection(event: Event): Promise<void> {
         // Only handle left clicks
         if (event instanceof MouseEvent && event.button !== 0) {
+            return;
+        }
+
+        // Prevent selection if we're currently scrolling (iPad fix)
+        if (this.isScrolling) {
             return;
         }
 
@@ -925,6 +980,14 @@ export class Menu extends SizedMixin(SpectrumElement, { noDefaultSize: true }) {
     }
 
     public override disconnectedCallback(): void {
+        // Clean up touch event listeners to prevent memory leaks
+        this.removeEventListener(
+            'touchstart',
+            this.handleTouchStart.bind(this)
+        );
+        this.removeEventListener('touchmove', this.handleTouchMove.bind(this));
+        this.removeEventListener('touchend', this.handleTouchEnd.bind(this));
+
         this.cachedChildItems = undefined;
         this.selectedItems = [];
         this.selectedItemsMap.clear();
--- a/packages/picker/test/index.ts
+++ b/packages/picker/test/index.ts
@@ -618,6 +618,50 @@ export function runPickerTests(): void {
             expect(el.open, 'open?').to.be.false;
             expect(el.value).to.equal(thirdItem.value);
         });
+        it('does not select items when scrolling on iPad', async () => {
+            // Simulate iPad touch scrolling behavior
+            const el = await pickerFixture();
+            await elementUpdated(el);
+
+            // Open the picker
+            el.open = true;
+            await elementUpdated(el);
+
+            // Store initial value
+            const initialValue = el.value;
+            const touchStartEvent = new Event('touchstart', {
+                bubbles: true,
+                composed: true,
+            });
+
+            // Simulate touch move (scrolling gesture)
+            const touchMoveEvent = new Event('touchmove', {
+                bubbles: true,
+                composed: true,
+            });
+
+            // Simulate touch end on fourth item (after scrolling)
+            const touchEndEvent = new Event('touchend', {
+                bubbles: true,
+                composed: true,
+            });
+
+            // Dispatch touch events in sequence
+            el.optionsMenu.dispatchEvent(touchStartEvent);
+            await nextFrame();
+
+            el.optionsMenu.dispatchEvent(touchMoveEvent);
+            await nextFrame();
+
+            el.optionsMenu.dispatchEvent(touchEndEvent);
+            await nextFrame();
+
+            // The value should not have changed because we were scrolling, not selecting
+            expect(el.value).to.equal(
+                initialValue,
+                'Value should not change during scroll gesture'
+            );
+        });
         it('opens/closes multiple times', async () => {
             expect(el.open, 'open?').to.be.false;
             const boundingRect = el.button.getBoundingClientRect();


packages/menu/src/Menu.ts
```diff
@@ -72,6 +72,13 @@ export class Menu extends SizedMixin(SpectrumElement, { noDefaultSize: true }) {
 
     protected rovingTabindexController?: RovingTabindexController<MenuItem>;
 
+    // Add touch tracking properties for iPad scroll handling
+    private touchStartY = 0;
+    private touchStartTime = 0;
+    private isScrolling = false;
+    private scrollThreshold = 10; // pixels
```
blunteshwar, line 118: Regarding `scrollThreshold` and `scrollTimeThreshold`: It's great that this addresses the iPad scrolling issue. Could you provide a bit more context on how the scrollThreshold and scrollTimeThreshold values were determined? Are they based on empirical testing, or are there any common best practices for these values that could be referenced? Just want to ensure they're robust enough to cover various scroll speeds and nuances on iPad.

Rajdeepc: Yes the `10px` and `300ms` values were determined through empirical testing on various iPad models. One thing we can do is to make these values configurable but I don't think that is something we can do if we get reports of lags or jitter. Tested this bug in iPad Pro, Air, iPad mini in simulator should cover the most of the blast radius of this issue.

packages/menu/src/Menu.ts
```diff
@@ -400,6 +407,17 @@ export class Menu extends SizedMixin(SpectrumElement, { noDefaultSize: true }) {
...
         this.addEventListener('sp-closed', this.handleSubmenuClosed);
+
+        // Add touch event listeners for iPad scroll detection
+        this.addEventListener('touchstart', this.handleTouchStart, {
+            passive: true,
+        });
+        this.addEventListener('touchmove', this.handleTouchMove, {
+            passive: true,
+        });
+        this.addEventListener('touchend', this.handleTouchEnd, {
```
caseyisonit: there is already a touchend event listener above this. should that be removed in favor of this?

packages/menu/src/Menu.ts
```diff
@@ -925,6 +980,14 @@ export class Menu extends SizedMixin(SpectrumElement, { noDefaultSize: true }) {
     }
 
     public override disconnectedCallback(): void {
+        // Clean up touch event listeners to prevent memory leaks
```
caseyisonit: why are we removing these and not the others?

packages/menu/test/menu.test.ts
```diff
@@ -707,4 +707,249 @@ describe('Menu', () => {
...
+        const el = await fixture<Menu>(html`
+            <sp-menu selects="single">
+                <sp-menu-item value="1">Item 1</sp-menu-item>
+                <sp-menu-item value="2">Item 2</sp-menu-item>
+            </sp-menu>
+        `);
+        await elementUpdated(el);
+
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const menu = el as any;
```
caseyisonit: we should avoid using the `any` type, I believe this should be Menu base on the signature above

packages/menu/test/menu.test.ts
```diff
@@ -707,4 +707,249 @@ describe('Menu', () => {
...
+        const firstItem = el.querySelector('sp-menu-item') as MenuItem;
+
+        // Manually set scrolling state to simulate iPad scrolling
+        menu.isScrolling = true;
+
+        // Try to select an item while scrolling
+        const clickEvent = new MouseEvent('click', { button: 0 });
```
caseyisonit: this is using a mouse event in an anticipated touch experience, I don't this this is the correct check we want.

packages/menu/test/menu.test.ts
```diff
@@ -707,4 +707,249 @@ describe('Menu', () => {
...
+        // Verify not scrolling initially
+        expect(menu.isScrolling).to.be.false;
+
+        // Try to select an item when not scrolling
+        firstItem.click();
+
+        // Wait for the selection to be processed
+        await elementUpdated(el);
+        await elementUpdated(firstItem);
+        await nextFrame();
```
caseyisonit: this shouldnt be needed after an elementUpdated returns. Can you explain why this is needed?

packages/picker/test/index.ts
```diff
@@ -618,6 +618,50 @@ export function runPickerTests(): void {
...
+            const el = await pickerFixture();
+            await elementUpdated(el);
+
+            // Open the picker
+            el.open = true;
+            await elementUpdated(el);
+
+            // Store initial value
+            const initialValue = el.value;
+            const touchStartEvent = new Event('touchstart', {
```
caseyisonit: should we use oneEvent to stub the event so it doesn't introduce side effects?

packages/menu/test/menu.test.ts
```diff
@@ -707,4 +707,249 @@ describe('Menu', () => {
         await nextFrame();
         expect(el.selected).to.deep.equal(['3']);
     });
+
+    it('prevents selection when scrolling is detected', async () => {
+        const el = await fixture<Menu>(html`
+            <sp-menu selects="single">
```
caseyisonit: there arent enough items to simulate scrolling, I'm wondering if these tests should be present in picker versus just a flat menu?

packages/menu/src/Menu.ts
nikkimk: If we only want these event listeners to exist while attached to the DOM and to remove them during the `disconnected` phase of the lifecycle, then we should have been adding them on the `connected` part of the lifecycle instead of the `constructor`.

Right now, if you were to remove the picker from the DOM and add it back, the listeners would not be re-applied.

packages/menu/src/Menu.ts
nikkimk: In the code comments, we need more of an explanation about what this logic does and why it works, so that we avoid future questions on it and regressions. I can't imagine someone else supporting this a year down the road and knowing exactly how to fix it.

packages/menu/test/menu.test.ts
```diff
@@ -707,4 +707,249 @@ describe('Menu', () => {
         await nextFrame();
         expect(el.selected).to.deep.equal(['3']);
     });
+
+    it('prevents selection when scrolling is detected', async () => {
```
nikkimk: Why are these tests on menu when the issue is with Picker's dropdown?

nikkimk (changes requested):

Rajdeepc:
> For the tests in menu, I'm struggling to understand how its accurately testing scrolling with only two menu items and the menu isn't in a container. These tests feel a little false positive but would like to understand better how they are working.
> 
> I've left a number of comments and questions for you. :)

Good point Casey, those were working since the mouse events are not trying to simulate a touch Event. I checked the repo and also the web-test-runner docs it lacks support for `TouchEvent` as it doesnt support Node based API execution in headless environment. I refactored the test to test the actual prevention logic instead.

castastrophe:
@Rajdeepc Is there also a Jira ticket associated with this update?

caseyisonit:
So i was learning more about the touch events because I haven't used them before and re-read the issue filed. The support request lists Chrome and Safari for the bug and I noticed in the API browser compatibility table touch events aren't supported in safari. I think this solution works for chrome but will need something else for safari. 

https://developer.mozilla.org/en-US/docs/Web/API/Touch_events#api.touchevent

caseyisonit:
i just tried to reproduce the bugs on both chrome and safari on my iPhone with force-popover on from the production URL and I'm able to scroll and select items just fine with no jitter and thrashing.

Rajdeepc:
> @Rajdeepc Is there also a Jira ticket associated with this update?

SWC-975

Rajdeepc:
> So i was learning more about the touch events because I haven't used them before and re-read the issue filed. The support request lists Chrome and Safari for the bug and I noticed in the API browser compatibility table touch events aren't supported in safari. I think this solution works for chrome but will need something else for safari.
> 
> https://developer.mozilla.org/en-US/docs/Web/API/Touch_events#api.touchevent

Good research! iPads run Safari on iOS, which has full support for touch events, including the TouchEvent constructor and touchstart / touchend listeners.
The MDN table you saw likely refers to Safari on macOS (desktop), which does not support TouchEvent, but that does not apply to iPads. You can check here on the support of iOS: https://caniuse.com/touch

.changeset/four-lions-enjoy.md
```diff
@@ -0,0 +1,5 @@
+---
+'@spectrum-web-components/menu': minor
+---
+
+**Fixed** : Fix iPad scrolling issue in picker dropdown where scrolling through menu items would accidentally select the first touched item and close the picker.
```
castastrophe, line 5: Would it be valuable to include a summary of _how_ the issue was fixed so that customers can get a sense from reading this about how the change might impact them or what considerations they should use in validating the fix?

packages/menu/src/Menu.ts
```diff
@@ -72,6 +72,51 @@ export class Menu extends SizedMixin(SpectrumElement, { noDefaultSize: true }) {
...
+     * 2. On touchmove: Calculate vertical movement and time elapsed
+     * 3. If movement > threshold AND time < threshold: Mark as scrolling
+     * 4. On touchend: Reset scrolling state after a delay
+     * 5. During selection: Prevent selection if scrolling is detected
+     *
+     * This prevents the common iPad issue where users accidentally select menu
+     * items while trying to scroll through the menu content.
+     */
+    private touchStartY = 0;
+    private touchStartTime = 0;
```
castastrophe: Is it possible for these to be undefined initially as a means of indicating `isScrolling` rather than having a separate `isScrolling` property?

Alternatively, `isScrolling` could be simply a getter that checks if the touchStartY and touchStartTime are undefined. Either way, I do think the `isScrolling` state should be a getter and a public-facing API so that apps can query the component for it's current scrolling state.

packages/menu/src/Menu.ts
```diff
@@ -72,6 +72,51 @@ export class Menu extends SizedMixin(SpectrumElement, { noDefaultSize: true }) {
...
+    private isScrolling = false;
+
+    /**
+     * Minimum vertical movement (in pixels) required to trigger scrolling detection.
+     * Small movements (like tapping) won't trigger scrolling, but intentional
+     * scroll gestures will.
+     */
+    private scrollThreshold = 10; // pixels
```
castastrophe, line 118: This one should be customizable by applications so they can adjust sensitivity to their own use-cases.

packages/menu/src/Menu.ts
```diff
@@ -72,6 +72,51 @@ export class Menu extends SizedMixin(SpectrumElement, { noDefaultSize: true }) {
...
+     * Maximum time (in milliseconds) for a movement to be considered scrolling.
+     * Quick movements within this timeframe are likely intentional scrolls,
+     * while slower movements are more likely taps or selections.
+     */
+    private scrollTimeThreshold = 300; // milliseconds
+
+    /**
+     * Public getter/setter for scrolling state
+     */
+    public get scrolling(): boolean {
```
castastrophe: I think the naming here is off. This should be the `isScrolling` API for customers and we don't necessarily need a duplicative static `isScrolling` boolean if we use the logic I proposed above to check if the start variables are undefined.

packages/menu/src/Menu.ts
```diff
@@ -443,6 +496,62 @@ export class Menu extends SizedMixin(SpectrumElement, { noDefaultSize: true }) {
...
+    /**
+     * Handles touchend events for iPad scroll detection.
+     *
+     * Resets the scrolling state after a short delay (100ms) to allow for
+     * any final touch events to be processed. This delay prevents immediate
+     * state changes that could interfere with the selection logic.
+     */
+    private handleTouchEnd(): void {
+        // Reset scrolling state after a short delay
+        setTimeout(() => {
```
castastrophe, line 586: This is where we can set the start variables to undefined instead of manually updating the `isScrolling` boolean. It will also ensure our start values are reset appropriately when we're done so that they're not returning old values when used.

castastrophe (changes requested):
A few notes on how we can optimize the data used to determine `isScrolling` -> I think if we leverage the start values of the component as a means of tracking that scrolling has started and when scrolling ends, we reset those same values to undefined, then we can rely on those values as a means of telling us if we are actively scrolling.

packages/menu/test/menu.test.ts
```diff
@@ -680,31 +679,205 @@ describe('Menu', () => {
...
+        expect(middleItem.selected).to.be.true;
+        expect(el.value).to.equal('15');
+
+        // Test that the component can be disconnected without errors
+        el.remove();
+    });
+
+    it('provides isScrolling getter for public API', async () => {
+        const menuItems = Array.from(
+            { length: 5 },
```
TarunAdobe: this test is not repetitive as the first test you added is already testing all of this

packages/menu/test/menu.test.ts
```diff
@@ -680,31 +679,205 @@ describe('Menu', () => {
...
+
+        // Verify that selection is prevented during scroll
+        expect(middleItem.selected).to.be.false;
+        expect(el.value).to.equal('1'); // Should still be the first item
+
+        // Reset scrolling state
+        el.isScrolling = false;
+        expect(el.isScrolling).to.be.false;
+
+        // Now try to select the item again (should work since scrolling is reset)
```
TarunAdobe, line 831: this feels super forced... is there a way that we can simulate real scrolling instead of using these variables cz tbh using this variables would ofcs work.

packages/menu/src/Menu.ts
```diff
@@ -72,6 +72,53 @@ export class Menu extends SizedMixin(SpectrumElement, { noDefaultSize: true }) {
...
+    private touchStartY: number | undefined = undefined;
+    private touchStartTime: number | undefined = undefined;
+    private isCurrentlyScrolling = false;
+
+    /**
+     * Minimum vertical movement (in pixels) required to trigger scrolling detection.
+     * Small movements (like tapping) won't trigger scrolling, but intentional
+     * scroll gestures will.
+     */
+    private scrollThreshold = 10; // pixels
```
TarunAdobe: is there a standard based on which we're setting this value? cz otherwise it might end up conflicting with some teams

TarunAdobe (changes requested):

PR #5641, opened 2025-07-28
fix(color-wheel): debug step functionality and keyboard navigation support

blunteshwar:
## Description

The step attribute for the color wheel was not functional. Modifying the step value had no effect on keyboard navigation. 
The issue was occurring because the step value was resolving to 1 irrespective of the step value provided by user inside `set altered` function.
This fix ensures that the step value is properly calculated and respected during keyboard navigation for the color wheel.

## Motivation and context
The step attribute for color-wheel was not functional at all. Changing the step value had no effect on keyboard navigation. 

## Related issue(s)
SWC-1022

-   fixes [Issue Number]

---

## Author's checklist

---

## Reviewer's checklist

### Manual review test cases

    1. Go [here](https://swcpreviews.z13.web.core.windows.net/pr-5641/docs/storybook/?path=/story/color-wheel--step-size)
    2. Click on the color handle for ay color wheel and navigate using arrow keys.
    3. Observe how step size is respected.

### Device review

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,7 +22,7 @@ parameters:
     # 3. Commit this change to the PR branch where the changes exist.
     current_golden_images_hash:
         type: string
-        default: dd2217ab088b6a78c6d95aa4751396726f24fa0f
+        default: e919c253a7cb75ee16c0ef680b50a032236f7275
     wireit_cache_name:
         type: string
         default: wireit


packages/color-wheel/test/color-wheel.test.ts
```diff
@@ -291,6 +291,164 @@ describe('ColorWheel', () => {
 
         expect(el.value).to.equal(0);
     });
+    it('respects custom step value with keyboard navigation', async () => {
+        const el = await fixture<ColorWheel>(html`
+            <sp-color-wheel step="10"></sp-color-wheel>
```
Rajdeepc, line 296: Can you also create a test for fractional step?

blunteshwar: Done!

blunteshwar:
Please update Image hash!

.circleci/config.yml
```diff
@@ -22,7 +22,7 @@ parameters:
     # 3. Commit this change to the PR branch where the changes exist.
     current_golden_images_hash:
         type: string
-        default: dd2217ab088b6a78c6d95aa4751396726f24fa0f
+        default: 5c3fb4c424ec590766e3a88c0018deef36210165
```
caseyisonit: can you reset this golden hash so we can review the VRTs before merge?

blunteshwar: Done!

packages/color-wheel/src/ColorWheel.ts
```diff
@@ -85,13 +85,16 @@ export class ColorWheel extends Focusable {
         this.colorController.color = color;
     }
 
-    private get altered(): number {
-        return this._altered;
-    }
+    private _baseStep = 1; // Preserves user's value
```
Rajdeepc: `_baseStep` can be typed explicitly as number. We want to avoid accidental assignment of non numeric values

packages/color-wheel/src/ColorWheel.ts
```diff
@@ -394,6 +397,12 @@ export class ColorWheel extends Focusable {
         this.addEventListener('blur', this.handleBlur);
     }
 
+    protected override willUpdate(changed: PropertyValues<this>): void {
+        if (changed.has('step')) {
+            this._baseStep = this.step;
```
Rajdeepc, line 402: If a user sets step="0" or step="-1", what happens? Can you add guards against these cases.

packages/color-wheel/src/ColorWheel.ts
blunteshwar: Great Catch

packages/color-wheel/src/ColorWheel.ts
blunteshwar: When user sets `step=-1` it will behave exactly as it should by changing the direction of up & down navigartion, which is inverted by default but it will change when step is negative. When user will set `step=0` (although its very rare a user will do this) the component will behave exactly as expected.
In number-field as well when we set step =0 or step=-1 it behaves similarly.
So I don't thing we should prevent user from entering negative/zero values for step.

Rajdeepc (approved):

PR #5654, opened 2025-07-31
docs(link): refactor link docs and add a11y

aramos-adobe:
## Related issue(s)

-   addresses [SWC-388]

---

## Author's checklist

---

## Reviewer's checklist

### DIFF
--- a/packages/link/README.md
+++ b/packages/link/README.md
@@ -1,6 +1,6 @@
-## Description
+## Overview
 
-An `<sp-link>` allow users to navigate to a different location. They can be presented in-line inside a paragraph or as a standalone text.
+An `<sp-link>` allows users to navigate to a different location. They can be presented in-line inside a paragraph or as a standalone text.
 
 ### Usage
 
@@ -24,16 +24,14 @@ When looking to leverage the `Link` base class as a type and/or for extension pu
 import { Link } from '@spectrum-web-components/link';
 ```
 
-## Example
+### Options
 
 <!-- prettier-ignore -->
 ```html
 This is an <sp-link href="#">example link</sp-link>.
 ```
 
-## Variants
-
-### Standard links
+#### Standard links
 
 Standard links are blue and should be used to call attention to the link or for when the blue color won’t feel too overwhelming in the experience.
 
@@ -42,16 +40,7 @@ Standard links are blue and should be used to call attention to the link or for 
 This is a <sp-link href="#">standard link</sp-link>.
 ```
 
-### Disabled links
-
-Disabled links are blue and should not propagate any events and they are not focussable.
-
-<!-- prettier-ignore -->
-```html
-This is a <sp-link disabled href="#">disabled link</sp-link>.
-```
-
-### Secondary links
+#### Secondary links
 
 The secondary variant is the same color as the paragraph text inline of which it appears. Its subdued appearance is optimal for when the primary variant is too overwhelming, such as in blocks of text with several references linked throughout.
 
@@ -60,7 +49,7 @@ The secondary variant is the same color as the paragraph text inline of which it
 This is a <sp-link href="#" variant="secondary">secondary link</sp-link>.
 ```
 
-### Static colored links
+#### Static colored links
 
 When a link needs to be placed on top of a colored background or a visual it may be appropriate to ship it with a static color, regardless of the theme settings with which it is delivered. Leverage the `static-color` attribute with its `white` or `black` values to ensure the delivery is the same in all contexts.
 
@@ -99,7 +88,7 @@ When a link needs to be placed on top of a colored background or a visual it may
 </sp-tab-panel>
 </sp-tabs>
 
-### Quiet links
+#### Quiet links
 
 All links can have a quiet style, which means they don’t have an underline. This style should only be used when the placement and context of the link is explicit enough that a visible underline isn’t necessary. Quiet links are less accessible, so they should not be used for links that are essential to the experience. These are commonly used in website footers, where there are several lists of links that are shortcuts to other pages.
 
@@ -118,7 +107,16 @@ All links can have a quiet style, which means they don’t have an underline. Th
 </div>
 ```
 
-### Download attribute
+#### Disabled links
+
+Disabled links are blue and should not propagate any events and they are not focussable.
+
+<!-- prettier-ignore -->
+```html
+This is a <sp-link disabled href="#">disabled link</sp-link>.
+```
+
+#### Download attribute
 
 The download attribute on an `<a>` tag prompts a user to download a link as opposed to navigating to it. This attribute has been carried forward to `<sp-link>` to function the same.
 
@@ -129,3 +127,18 @@ with standard `<a>` [rules](https://developer.mozilla.org/en-US/docs/Web/HTML/El
 ```html
 This is a <sp-link download="myfile.txt" href="#">download link</sp-link>.
 ```
+
+### Accessibility
+
+#### Best Practices
+
+- Use links in body copy.
+- Identify the target of each link directly in the link text to communicate context and set clear expectations about where the link will go.
+- Be mindful of link placement and language, and create experiences that are inclusive of users navigating with screen readers.
+- Ensure labels for links are descriptive and specific.
+
+#### Keyboard Interaction
+
+- `Tab`: Move focus to the next focusable element.
+- `Enter`: Activate the link.
+- `Shift + F10` (Optional): Open the context menu for the link.


packages/link/README.md
```diff
@@ -99,7 +88,7 @@ When a link needs to be placed on top of a colored background or a visual it may
 </sp-tab-panel>
 </sp-tabs>
 
-### Quiet links
+#### Quiet links
 
 All links can have a quiet style, which means they don’t have an underline. This style should only be used when the placement and context of the link is explicit enough that a visible underline isn’t necessary. Quiet links are less accessible, so they should not be used for links that are essential to the experience. These are commonly used in website footers, where there are several lists of links that are shortcuts to other pages.
```
Rajdeepc:
```suggestion
All links can have a quiet style, which means they don’t have an underline. This style should only be used when the placement and context of the link is explicit enough that a visible underline isn’t necessary. 
Quiet links are less accessible because users may not recognise them as links. Use only when context and placement make their purpose unmistakable, and avoid using quiet links for critical navigation.
```

packages/link/README.md
```diff
@@ -118,7 +107,16 @@ All links can have a quiet style, which means they don’t have an underline. Th
 </div>
 ```
 
-### Download attribute
+#### Disabled links
+
+Disabled links are blue and should not propagate any events and they are not focussable.
```
Rajdeepc:
```suggestion
Disabled links are blue and should not propagate any events and they are not focusable.
```

packages/link/README.md
```diff
@@ -24,16 +24,14 @@ When looking to leverage the `Link` base class as a type and/or for extension pu
...
 
-## Example
+### Options
 
 <!-- prettier-ignore -->
 ```html
 This is an <sp-link href="#">example link</sp-link>.
 ```
 
-## Variants
```
Rajdeepc, line 34: Can you tell me why Variants section is removed? The below are all the link variants

packages/link/README.md
```diff
@@ -118,7 +107,16 @@ All links can have a quiet style, which means they don’t have an underline. Th
...
 ```
 
-### Download attribute
+#### Disabled links
+
+Disabled links are blue and should not propagate any events and they are not focussable.
+
+<!-- prettier-ignore -->
+```html
+This is a <sp-link disabled href="#">disabled link</sp-link>.
```
Rajdeepc, line 123: For the “disabled link” example, can you also clarify in text that the link is not focusable nor clickable, so users know what to expect.

packages/link/README.md
```diff
@@ -129,3 +127,18 @@ with standard `<a>` [rules](https://developer.mozilla.org/en-US/docs/Web/HTML/El
...
 ```
+
+### Accessibility
+
+#### Best Practices
+
+- Use links in body copy.
+- Identify the target of each link directly in the link text to communicate context and set clear expectations about where the link will go.
+- Be mindful of link placement and language, and create experiences that are inclusive of users navigating with screen readers.
+- Ensure labels for links are descriptive and specific.
```
Rajdeepc: Can we add these also to expand the best practices guidelines?

1. A note about sufficient color contrast for links, especially for users with low vision.
2. Mention skip links or landmark navigation for complex pages.
3. If the component supports ARIA attributes, list them explicitly.

packages/link/README.md
aramos-adobe: I just removed it to follow the template structure so I swapped it with `Options`. I'll revert it back to `Variants`

Rajdeepc: Thank you!

packages/link/README.md
```diff
@@ -24,16 +24,14 @@ When looking to leverage the `Link` base class as a type and/or for extension pu
 import { Link } from '@spectrum-web-components/link';
 ```
 
-## Example
+### Variants
```
Rajdeepc, line 27:
```suggestion
## Variants
```

Rajdeepc (changes requested):

packages/link/README.md
```diff
@@ -1,39 +1,39 @@
...
 ```html
 This is an <sp-link href="#">example link</sp-link>.
 ```
 
-## Variants
-
-### Standard links
+<sp-tabs selected="standard" auto label="Variant Options">
+<sp-tab value="standard">Standard</sp-tab>
+<sp-tab-panel value="standard">
```
rise-erpelding: I'm curious about why this variant is called "standard" when the next one is "secondary." It feels like "primary" would align more closely with the "secondary" variant, it's also language that matches the [design docs](https://spectrum.adobe.com/page/link/#Primary-variant).

I think if the goal is to try to mirror the language seen in the design docs, updating this would be a good idea. You can also say in the tab panel copy that it's the default variant, and if there's a compelling reason to keep the word "standard" in here, we can also say it's the standard, default variant.

But I also see that there isn't a specific `variant="primary"` so I wonder if that's why we don't use that language in web components?

packages/link/README.md
```diff
@@ -1,39 +1,39 @@
...
 ```
 
 When looking to leverage the `Link` base class as a type and/or for extension purposes, do so via:
 
-```
+```ts
 import { Link } from '@spectrum-web-components/link';
 ```
 
-## Example
```
rise-erpelding, line 27: If we're taking out the Example heading, should we take out the example link below too? It feels a little redundant to have an example and then the tab panel with the variants right below it:

image

packages/link/README.md
```diff
@@ -42,25 +42,43 @@ Standard links are blue and should be used to call attention to the link or for
...
+All links can have a quiet style, which means they don’t have an underline. This style should only be used when the placement and context of the link is explicit enough that a visible underline isn’t necessary.
+Quiet links are less accessible because users may not recognise them as links. Use only when context and placement make their purpose unmistakable, and avoid using quiet links for critical navigation.
 
 <!-- prettier-ignore -->
 ```html
-This is a <sp-link href="#" variant="secondary">secondary link</sp-link>.
+<p>This is a <sp-link quiet href="#">quiet standard link</sp-link>.</p>
+<p>This is a <sp-link quiet variant="secondary" href="#">quiet secondary link</sp-link>.</p>
+<div
+    style="background-color: #0f797d; padding: 15px 20px; display: inline-block;"
```
rise-erpelding: I haven't seen that there's a standard for this particularly, but using spectrum custom properties for colors is something we can do here.

I see all sorts of background colors being used in the static examples on this docs site, so I don't think there's a _right_ way to do it, but IMO this feels like a good spot to use `var(--spectrum-docs-static-black-background-color)`. (It'd be cool to bring over the gradient we have in the CSS `spectrum-two` branch eventually.)

packages/link/README.md
```diff
@@ -42,25 +42,43 @@ Standard links are blue and should be used to call attention to the link or for
...
+>
+    <p style="color: rgb(240, 240, 240);">
```
rise-erpelding: I don't know if I have a good recommendation here since it's a quiet link, I think 240 is ok if you want a different color than the link, if there's specific color variable that aligns with that, that'd be better.

packages/link/README.md
```diff
@@ -42,25 +42,43 @@ Standard links are blue and should be used to call attention to the link or for
...
+        over a background.
+    </p>
+</div>
 ```
 
-### Static colored links
+</sp-tab-panel>
+</sp-tabs>
+
+#### Static colored links
```
rise-erpelding, line 76: Colors could be updated to use spectrum vars here too, `--spectrum-white` and `--spectrum-black` are often used for content color on static color backgrounds, and we can use the `--spectrum-docs-static-black-background-color` (a similar one exists for white) for backgrounds (looks like those custom props correspond to the background colors that were already here).

packages/link/README.md
```diff
@@ -129,3 +139,22 @@ with standard `<a>` [rules](https://developer.mozilla.org/en-US/docs/Web/HTML/El
...
+
+### Accessibility
+
+#### Best Practices
+
+- Use links in body copy.
+- Identify the target of each link directly in the link text to communicate context and set clear expectations about where the link will go.
+- Be mindful of link placement and language, and create experiences that are inclusive of users navigating with screen readers.
+- Ensure labels for links are descriptive and specific.
+- Implement skip links to improve navigation for keyboard and screen reader users when necessary especially the page has many sections and lengthy scroll.
```
rise-erpelding:
```suggestion
- Implement skip links to improve navigation for keyboard and screen reader users when necessary, especially when the page has many sections and lengthy scroll.
```

packages/link/README.md
```diff
@@ -129,3 +139,22 @@ with standard `<a>` [rules](https://developer.mozilla.org/en-US/docs/Web/HTML/El
 ```html
 This is a <sp-link download="myfile.txt" href="#">download link</sp-link>.
 ```
+
+### Accessibility
+
+#### Best Practices
+
+- Use links in body copy.
```
rise-erpelding: Is this paraphrased from the [design guidelines](https://spectrum.adobe.com/page/link/#Use-links-in-body-copy)?

> Links are for usage in body copy and are not appropriate in titles. Consider using a different component if you're looking for a larger or more prominent call to action, such as a button.

It would be nice to explain that they're for use in body copy and not in titles, for a little further explanation.

packages/link/README.md
```diff
@@ -129,3 +139,22 @@ with standard `<a>` [rules](https://developer.mozilla.org/en-US/docs/Web/HTML/El
...
+- Identify the target of each link directly in the link text to communicate context and set clear expectations about where the link will go.
+- Be mindful of link placement and language, and create experiences that are inclusive of users navigating with screen readers.
```
rise-erpelding: A little extra context might be helpful here too, for instance something like:

```suggestion
- Be mindful of link placement and language, and create experiences that are inclusive of users navigating with screen readers, who may navigate links without their surrounding language.
```

Or, it might be safer to copy some of what we already have in the design guidelines rather than paraphrasing.

packages/link/README.md
```diff
@@ -129,3 +139,22 @@ with standard `<a>` [rules](https://developer.mozilla.org/en-US/docs/Web/HTML/El
...
+- Ensure labels for links are descriptive and specific.
```
rise-erpelding: The design docs use the term "link text" over "labels for links," it might be good to align there.

This bullet sort of relates to the bullet point above, but could also be separate.

There's also this section about [avoiding generic language](https://spectrum.adobe.com/page/link/#Avoid-generic-language) that we could take from. We could fit something like this below into the bullet point here:

> It’s more accessible and inclusive to write link text as unique descriptions of the navigational target or function.

packages/link/README.md
```diff
@@ -129,3 +139,22 @@ with standard `<a>` [rules](https://developer.mozilla.org/en-US/docs/Web/HTML/El
...
+- Implement skip links to improve navigation for keyboard and screen reader users when necessary especially the page has many sections and lengthy scroll.
+- For links that open in a new tab, add `target="_blank"`, `rel="noopener noreferrer"` and a UI icon to the link.
+- Add `aria-label` or `aria-labelledby` to links for additonal context. Links can be more concise to lessen visible noise, but adding these attributes can help make the purpose of the link more clear.
+- Ensure strong color contrast between the link and its background. For users with low vision,consider using `7:1 ratio` for critical links and `21:1 ratio` for severe vision impairments.
```
rise-erpelding: nit, but these ratios don't need to be code-formatted, they're not on the design docs:

image

packages/link/README.md
```diff
@@ -129,3 +134,22 @@ with standard `<a>` [rules](https://developer.mozilla.org/en-US/docs/Web/HTML/El
 ```html
 This is a <sp-link download="myfile.txt" href="#">download link</sp-link>.
 ```
+
+### Accessibility
+
+#### Best Practices
+
+- Use links in body copy and not in titles. For larger projects, consider using a [button](/components/button/) instead.
```
rise-erpelding: For "larger calls to action," maybe?

```suggestion
- Use links in body copy and not in titles. For larger calls to action, consider using a [button](/components/button/) instead.
```

packages/link/README.md
```diff
@@ -144,14 +139,14 @@ This is a <sp-link download="myfile.txt" href="#">download link</sp-link>.
...
-- Be mindful of link placement and language, and create experiences that are inclusive of users navigating with screen readers.
-- Ensure labels for links are descriptive and specific.
-- Implement skip links to improve navigation for keyboard and screen reader users when necessary especially the page has many sections and lengthy scroll.
+- Be mindful of link placement and language, and create experiences that are inclusive of users navigating with screen readers, who may navigate links without their surrounding language.
+- It’s more accessible and inclusive to write link text as unique descriptions of the navigational target or function.
+- Implement skip links to improve navigation for keyboard and screen reader users when necessary, especially when the page has many sections and lengthy scroll.
 - For links that open in a new tab, add `target="_blank"`, `rel="noopener noreferrer"` and a UI icon to the link.
 - Add `aria-label` or `aria-labelledby` to links for additonal context. Links can be more concise to lessen visible noise, but adding these attributes can help make the purpose of the link more clear.
```
rise-erpelding: Oop I missed this the first time around. There's a typo here, "additonal" is spelled wrong.

Also, re: `aria-label`, it looks like the way that this works is if you want to pass an `aria-label` to the link, you add a `label` property in `<sp-link>`? It doesn't seem like adding `aria-label` to `<sp-link>` will pass that to the `<a>`. Thus, we'd want to say somewhere in here that you'd use `label` to set `aria-label`, I think.

I'm a bit more confused about `aria-labelledby` though, it doesn't seem like there's a way to set it on `<sp-link>` that would pass it down to the `<a>` tag... I would have thought it would need to be set on the anchor tag in order to be accessible? I'm wondering if that's a possible issue that needs addressing, unless I'm misunderstanding how `aria-labelledby` works in web components?

aramos-adobe: I think what I was trying to say is that `aria-label` and `aria-labeldescribedby` is only used for screen reader purposes. Sometimes designers would make links short so adding more context, `aria-labeldescribedby` you can add very descriptive text while `aria-label` is just the text written for `sp-link` for accessibility

PR #5683, opened 2025-08-08
[DRAFT] gnorton/swc-1032-do-not-sort-export-conditions

graynorton:
## Related issue(s)

-   fixes [Issue Number]

---

## Author's checklist

---

## Reviewer's checklist

### Manual review test cases

    1. Go [here](url)
    2. Do this action
    3. Expect this result

    1. Go [here](url)
    2. Do this action
    3. Expect this result

### Device review

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -17,6 +17,15 @@
             "files": ["*.json"],
             "parser": "jsonc-eslint-parser",
             "rules": {
+                "jsonc/sort-keys": ["warn"],
+                "notice/notice": "off"
+            }
+        },
+        {
+            "extends": ["plugin:jsonc/recommended-with-jsonc"],
+            "files": ["package.json"],
+            "parser": "jsonc-eslint-parser",
+            "rules": {
                 "jsonc/sort-keys": [
                     "warn",
                     {
@@ -65,14 +74,13 @@
                             "cpu",
                             "publishConfig"
                         ],
-                        "pathPattern": ".*" // Hits the all properties
+                        "pathPattern": "^$" // Top-level properties
                     },
                     {
                         "order": { "type": "asc" },
-                        "pathPattern": ".*"
+                        "pathPattern": "^(?!exports\\[).*" // All other properties except export conditions
                     }
-                ],
-                "notice/notice": "off"
+                ]
             }
         },
         {
--- a/packages/.eslintrc.json
+++ b/packages/.eslintrc.json
@@ -135,6 +135,15 @@
             "files": ["*.json"],
             "parser": "jsonc-eslint-parser",
             "rules": {
+                "jsonc/sort-keys": ["warn"],
+                "notice/notice": "off"
+            }
+        },
+        {
+            "extends": ["plugin:jsonc/recommended-with-jsonc"],
+            "files": ["package.json"],
+            "parser": "jsonc-eslint-parser",
+            "rules": {
                 "jsonc/sort-keys": [
                     "warn",
                     {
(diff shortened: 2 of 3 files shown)


.eslintrc.json
```diff
@@ -16,6 +16,15 @@
             "extends": ["plugin:jsonc/recommended-with-jsonc"],
             "files": ["*.json"],
             "parser": "jsonc-eslint-parser",
+            "rules": {
+                "jsonc/sort-keys": ["warn"],
+                "notice/notice": "off"
```
Rajdeepc, line 21: Can you give me some context on this change?

packages/.eslintrc.json
```diff
@@ -183,14 +192,13 @@
                             "cpu",
                             "publishConfig"
                         ],
-                        "pathPattern": ".*" // Hits the all properties
+                        "pathPattern": "^$" // Top-level properties
                     },
                     {
                         "order": { "type": "asc" },
-                        "pathPattern": ".*"
+                        "pathPattern": "^(?!exports\\[).*" // All other properties except export conditions
```
Rajdeepc: Can we document these decisions too so that future maintainers know why certain keys aren't sorted?

.eslintrc.json
graynorton: Sure! What's happening here is that I've lifted our general-purpose JSON-linting behavior...

- By default, sort keys alphabetically
- Don't check for inclusion of our copyright notice, because that notice is only intended for JS/TS files

...into an initial rule, which applies to all JSON files by default.

This general rule is followed by a rule targeting only `package.json` files, which covers `package.json`-specific needs:

- Sort top-level keys in a custom order
- DON'T sort export conditions alphabetically

packages/.eslintrc.json
graynorton: Sure, I'll add the reasoning to the comment—good catch

graynorton: Updated to include a more thorough comment. Thanks!

Rajdeepc (approved):
LGTM!! Thanks for the detailed write-up and explanation.

packages/.eslintrc.json
```diff
@@ -134,6 +134,15 @@
             "extends": ["plugin:jsonc/recommended-with-jsonc"],
             "files": ["*.json"],
             "parser": "jsonc-eslint-parser",
+            "rules": {
```
castastrophe, line 137: You didn't need to repeat these settings because eslint configs inherit.

packages/.eslintrc.json
```diff
@@ -183,14 +192,18 @@
                             "cpu",
                             "publishConfig"
                         ],
-                        "pathPattern": ".*" // Hits the all properties
+                        "pathPattern": "^$" // Top-level properties
```
castastrophe, line 195: Json does not allow inline commenting unless you're using json5 which we are not.

packages/.eslintrc.json
graynorton: I believe that because we have specified `"root": true` in the ESLint configs for `packages` and `tools`, inheritance from configs in ancestor directories is short-circuited, making these effectively standalone configs.

(I actually discovered this because, when working on this issue, I first applied a config change to the root config only and was confused to find it didn't apply to files beneath the `packages` directory.

As we've discussed in Slack, this doesn't seem necessary or desirable, since—at least as far as I know—there's no good reason for us not to share most of the same rules throughout the project. That's definitely something I'd like us to change in a follow-up PR.)

In any case, the repetition here was not added in this PR; the JSON config was already present in all three files, so this PR just updates it in all three places.

packages/.eslintrc.json
graynorton: I wondered about this, too, as there were already comments present in our configs. I looked it up, and it turns out that both ESLint 8.x and 9.x support JS-style comments in their JSON config files:

https://eslint.org/docs/v8.x/use/configure/configuration-files#configuration-file-formats

Issue #5620, opened 2025-07-21
[Bug]: button-base aria label not updating on change


SavkaTaras:
### Code of conduct

- [x] I agree to follow this project's code of conduct.

### Impacted component(s)

ButtonBase.ts

### Library version

1.7.0

### Expected behavior

Should update on change.

### Actual behavior

Not updating after firstUpdated lifecycle

### What browsers are you seeing the problem in?

Chrome

### How can we reproduce this issue?

3. Scroll to '....'
4. Check console
5. See error

### Severity

None

### Would you like to track this issue in Jira?

- [ ] Yes, please tell me the ticket number!

Rajdeepc:
SWC-1039

najikahalsema:
Closed as duplicate of #5675

PR #5674, opened 2025-08-06
fix: base-button, update aria label on updated event. This also updated attr on slotchange event

tsavka-adobe:
## Description
Currently the `aria-label` attribute gets set on `firstUpdate`. This causes an issue when the button is rendered and then the prop is updated. The change is not getting reflected. This makes sure the change actually gets reflected.

Update:
`aria-label` gets removed on `slotChange` event as well. This PR sets `aria-label` on any change based on `label` prop value.

## Motivation and context

Adobe Expres A11Y critical bug

## Related issue(s)

-   fixes [[5620](https://github.com/adobe/spectrum-web-components/issues/5620)]

## Screenshots (if appropriate)

---

## Author's checklist

-   [ ] I have read the **[CONTRIBUTING](<(https://github.com/adobe/spectrum-web-components/blob/main/CONTRIBUTING.md)>)** and **[PULL_REQUESTS](<(https://github.com/adobe/spectrum-web-components/blob/main/PULL_REQUESTS.md)>)** documents.
-   [ ] I have reviewed at the Accessibility Practices for this feature, see: [Aria Practices](https://www.w3.org/TR/wai-aria-practices/)
-   [ ] I have added automated tests to cover my changes.
-   [ ] I have included a well-written changeset if my change needs to be published.
-   [ ] I have included updated documentation if my change required it.

---

## Reviewer's checklist

-   [ ] Includes a Github Issue with appropriate flag or Jira ticket number without a link
-   [ ] Includes thoughtfully written changeset if changes suggested include `patch`, `minor`, or `major` features
-   [ ] Automated tests cover all use cases and follow best practices for writing
-   [ ] Validated on all supported browsers
-   [ ] All VRTs are approved before the author can update Golden Hash

### Manual review test cases

-   [ ] _Descriptive Test Statement_

    1. Go [here](url)
    2. Do this action
    3. Expect this result

-   [ ] _Descriptive Test Statement_
    1. Go [here](url)
    2. Do this action
    3. Expect this result

### Device review

-   [ ] Did it pass in Desktop?
-   [ ] Did it pass in (emulated) Mobile?
-   [ ] Did it pass in (emulated) iPad?

PR #5691, opened 2025-08-12
fix(button): update aria-label when label changes

mizgaionutalexandru:
## Description

This PR modifies the aria-label update logic of the `ButtonBase` in order to correctly update when the `label` property changes during its lifetime, gating the buttons that use the `PendingStateController` to not override their aria-label manipulation.

Unit tests were added to validate these usecases in the future. All existing tests pass.

Implementation based on https://github.com/adobe/spectrum-web-components/pull/5674

## Motivation and context

Currently when the `label` property changes, the `aria-label` of the button components do not reflect this change.
This issue was introduced in 0.48.0 with the new PendingStateController, hence this PR fixes a regression.

## Related issue(s)

-   fixes https://github.com/adobe/spectrum-web-components/issues/5620

https://github.com/user-attachments/assets/bb921063-1924-400c-8370-dab44decb53a

---

## Author's checklist

---

## Reviewer's checklist

### Manual review test cases

-   [ ] _Validate `aria-label` updates when `label` updates_

    1. Go to [the button storybook](https://swcpreviews.z13.web.core.windows.net/pr-5691/docs/storybook/?path=/story/button-accent-fill--default)
    2. Inspect the element and get its aria-label
    3. Modify the element's label
    4. Get the element's new aria-label
    5. Observe the updated aria-label

-   [ ] _Validate `aria-label` values when element's content changes_

    1. Repeat test 1
    4. Change the element's textContent/innerText/innerHTML
    5. Get the element's aria-label
    6. Observe that it stays the same

### Device review


packages/button/src/ButtonBase.ts
```diff
@@ -220,18 +220,26 @@ export class ButtonBase extends ObserveSlotText(LikeAnchor(Focusable), '', [
...
 
+    private isPendingState(): boolean {
+        return (
+            'pending' in this && (this as { pending: boolean }).pending === true
+        );
+    }
+
+    private updateAriaLabel(): void {
+        if (this.label) {
+            this.setAttribute('aria-label', this.label);
```
Rajdeepc, line 243: If a user sets `aria-label` directly after the component has set it via the label property, the component might override the user's intent here. I would be checking if `aria-label` is set by the user or not before we trigger component logic.
```ts
private _userSetAriaLabel = false;

connectedCallback() {
    super.connectedCallback();
    // Detect if aria-label was set by the user before component logic runs
    this._userSetAriaLabel = this.hasAttribute('aria-label');
}

attributeChangedCallback(name: string, oldValue: string | null, newValue: string | null) {
    if (name === 'aria-label' && oldValue !== newValue) {
        this._userSetAriaLabel = true;
    }
    super.attributeChangedCallback(name, oldValue, newValue);
}

private updateAriaLabel(): void {
    if (this._userSetAriaLabel) return; // Don't override user intent
    if (this.label) {
        this.setAttribute('aria-label', this.label);
    } else {
        this.removeAttribute('aria-label');
    }
}
```

packages/button/src/ButtonBase.ts
```diff
@@ -220,18 +220,26 @@ export class ButtonBase extends ObserveSlotText(LikeAnchor(Focusable), '', [
         }
     }
 
+    private isPendingState(): boolean {
+        return (
+            'pending' in this && (this as { pending: boolean }).pending === true
```
Rajdeepc: duck-typing is fragile here and is not explicit. Two suggestions either you check for instance and cast only if instance is PendingState or you can add a clear comment saying that `pending` is a property injected here.

packages/button/src/ButtonBase.ts
```diff
@@ -243,6 +251,11 @@ export class ButtonBase extends ObserveSlotText(LikeAnchor(Focusable), '', [
             this.manageAnchor();
         }
 
+        // Don't override PendingStateController's aria-label changes
+        if (changed.has('label') && !this.isPendingState()) {
+            this.updateAriaLabel();
```
Rajdeepc, line 269: Both `update` and `updated` might call `updateAriaLabel` for the same change. Can you call `updateAriaLabel` in only one lifecycle or just make sure it runs once per change. 
Do you have any specific reason to call this in `update` method?

packages/button/src/ButtonBase.ts
```diff
@@ -243,6 +251,11 @@ export class ButtonBase extends ObserveSlotText(LikeAnchor(Focusable), '', [
             this.manageAnchor();
         }
 
+        // Don't override PendingStateController's aria-label changes
```
Rajdeepc:
```suggestion
       // Do not update aria-label if component is in pending state,
       // as PendingStateController may manage it for accessibility.

```

Rajdeepc (changes requested):

packages/button/src/ButtonBase.ts
mizgaionutalexandru: So we want to allow both `label` and `aria-label` on the component at the same time? Isn't `label` used just for aria-label purposes, [as per docs](https://opensource.adobe.com/spectrum-web-components/components/button/api/#attributes%20and%20properties_label:~:text=An%20accessible%20label%20that%20describes%20the%20component.%20It%20will%20be%20applied%20to%20aria%2Dlabel%2C%20but%20not%20visually%20rendered.) and code?
If the consumer doesn't set a `label` but sets an `aria-label`, the component will have the desired aria-label.

packages/button/src/ButtonBase.ts
mizgaionutalexandru: This surfaces multiple conflicts since the Button component has `label` and the `PendingStateController` that modify the `aria-label`. Maybe a long term major change would be to get rid of the `label` prop altogether 🤔 or if we keep it we need a library level way to determine if it's only for aria-label purposes or not, in order to update the `PendingStateController` accordingly.
For now I gated the updates according to the pending state to happen either in `update` or `updated`.

packages/button/src/ButtonBase.ts
mizgaionutalexandru: Updated this to use an `instanceof` check here instead. Thank you for pointing this out!

packages/button/src/ButtonBase.ts
Rajdeepc: Cool thats something we would like to do in future. Do you mind validating this in screen readers?

packages/button/src/ButtonBase.ts
Rajdeepc: Theoretically we generally don’t want both to apply at the same time. The implementation should respect a consumer’s explicit `aria-label` and only use label as a fallback for accessibility. This will avoid race conditions and user confusion. 
I think we should document that `aria-label` always takes priority over `label`.
  - If both are set, use `aria-label`.
  - If only label is set, set `aria-label` to label.
  - If neither is set, no `aria-label` is present.

mizgaionutalexandru: I agree that this _might_ be a good improvement, even though there is no need for a consumer to provide both `label` and `aria-label`. Currently, on main, if `label` and `aria-label` are set, the `label` will override the `aria-label`. That being said, I feel like adding more is out of scope and we're looking forward to see this fix merged. Can we keep this in mind and act on it in another PR if it truly is something we want?

packages/button/src/ButtonBase.ts
mizgaionutalexandru: Sure, I added a screen recording, validating the manual test cases with a screen reader. Let me know if I should add anything else.

jnurthen (commented):
IMO we should not add support for consumers using `aria-label` directly on SWC.
`aria-label` is an aria global so it forces the component itself to be exposed as an accessible node. If in the future we wanted to change the implementation to instead use a button within the shadow DOM we wouldn't be able to do so.

mizgaionutalexandru:
> IMO we should not add support for consumers using `aria-label` directly on SWC. `aria-label` is an aria global so it forces the component itself to be exposed as an accessible node. If in the future we wanted to change the implementation to instead use a button within the shadow DOM we wouldn't be able to do so.

This PR does not "add support for aria-label". It only fixes an inconsistency bug in that area. 
Am I missing something here? @jnurthen

.changeset/hot-clouds-burn.md
```diff
@@ -0,0 +1,5 @@
+---
+'@spectrum-web-components/button': patch
+---
+
+**Fixed** aria-label updates when `label` property changes.
```
castastrophe, line 5: Should we add a note here about the updates to the pending state check as well? I don't see that reflected in the log.

packages/button/src/ButtonBase.ts
```diff
@@ -220,18 +224,34 @@ export class ButtonBase extends ObserveSlotText(LikeAnchor(Focusable), '', [
...
+            this.pendingStateController instanceof PendingStateController
+        );
+    }
+
+    private isPendingState(): boolean {
+        return this.isHostWithPendingState() && this.pending === true;
+    }
+
+    private updateAriaLabel(): void {
+        if (this.label) {
```
castastrophe, line 242: Should we explicitly check for undefined instead because empty quotes would be a valid label (`label=""`)? cc @nikkimk is that allowed on buttons or am I just thinking of images that use the empty label string sometimes?

packages/button/src/ButtonBase.ts
castastrophe: I think keeping that change as out of scope makes sense. I'd like to see us create a follow-up ticket to cover the suggestion though. @Rajdeepc would you be able to take care of that and include your code suggestions?

packages/button/test/button.test.ts
```diff
@@ -423,6 +423,45 @@ describe('Button', () => {
             expect(el.getAttribute('aria-label')).to.equal('clickable');
         });
 
+        it('updates aria-label when label changes', async () => {
```
castastrophe, line 426: Thanks for adding this test case!

packages/button/test/button.test.ts
```diff
@@ -423,6 +423,45 @@ describe('Button', () => {
...
+            const el = await fixture<Button>(html`
+                <sp-button label="Initial Label">Button</sp-button>
```
castastrophe:
```suggestion
                <sp-button label="Initial label">Button</sp-button>
```

(Super minor nit - we try to keep with content standards for Spectrum which includes Sentence case and not title case. It's a label though so it's really not that important.)

packages/button/test/button.test.ts
```diff
@@ -423,6 +423,45 @@ describe('Button', () => {
...
+        it('preserves aria-label when slot content changes', async () => {
+            const el = await fixture<Button>(html`
+                <sp-button label="Test Label">Initial Content</sp-button>
+            `);
+
+            await elementUpdated(el);
+            expect(el.getAttribute('aria-label')).to.equal('Test Label');
+
+            // Change the slot content
+            el.textContent = 'Updated Content';
```
castastrophe:
```suggestion
            el.textContent = 'Updated content';
```

Same minor nit here; not ultra important since it's a test case but just a good habit to be in.

.changeset/hot-clouds-burn.md
mizgaionutalexandru: I added it just now. Please let me know if this works or if you have any other suggestions

packages/button/test/button.test.ts
mizgaionutalexandru: Changed it, thank you for mentioning!

packages/button/src/ButtonBase.ts
mizgaionutalexandru: Good question, I haven't thought of that since this check was also used previously. 🤔 
My take is that we don't want to allow consumers to use an empty string as an aria-label, they can just skip using it altogether. I know something like this is used with the `alt` attribute of images though.

.changeset/hot-clouds-burn.md
castastrophe: Thank you!

packages/button/src/ButtonBase.ts
Rajdeepc: Yes I agree with both of you! Let's create a follow up ticket and unblock this PR for now.

Rajdeepc:
Tracked in https://github.com/adobe/spectrum-web-components/pull/5730

PR #5699, opened 2025-08-18
docs(swatchgroup): refactor docs, add a11y section

aramos-adobe:
## Description

Documentation audit for the swatch group component. 

## Related issue(s)

-   docs SWC-409

## Author's checklist

---


packages/swatch/swatch-group.md
```diff
@@ -116,11 +118,13 @@ import { SwatchGroup } from '@spectrum-web-components/swatch';
 </sp-tab-panel>
 </sp-tabs>
 
-## Density
+#### Density
 
 The `density` attribute/property is not required and when applied accepts the values of `compact` or `spacious`.
```
rise-erpelding: If no `density` is specified, it just picks a density in between compact and spacious, is that correct? That'd be a great thing to add in here, and an example within the tabs below. The design docs lean on the idea of compact/regular/spacious density so it'd be nice to echo that here.

packages/swatch/swatch-group.md
```diff
@@ -217,6 +233,27 @@ When available, the value of `selected` will be constructed from the `value` att
...
+
+#### Best practices
+
+- Ensure swatches have sufficient color contrast for visibility.
+- Verify that swatches are appropriately labeled for screen readers.
+
+#### Keyboard navigation
 
-An `<sp-swatch-group>` element can be modified by the following attributes/properties to customize its delivery as desired for your use case: `border`, `disabled`, `mixedValue` (accepted as the `mixed-value` attribute), `nothing`, `rounding`, `shape`, and `size`. Use these in concert with each other for a variety of final visual deliveries. Applying a value for one of these attributes/properties to an `<sp-swatch-group>` element will have it forward the value to all of the `<sp-swatch>` elements that are a direct child of the group, overriding any value that may be applied directly to those children.
+- `Tab`: Move focus to the next focusable element
```
rise-erpelding, line 337: Could we also talk about how the component controls the tab index? Similar to how it's presented either in [tags](https://opensource.adobe.com/spectrum-web-components/components/tags/#accessibility) or in [tab panel](https://opensource.adobe.com/spectrum-web-components/components/tab-panel/#accessibility).

packages/swatch/swatch-group.md
```diff
@@ -217,6 +233,27 @@ When available, the value of `selected` will be constructed from the `value` att
...
+
+```html
+<sp-swatch-group border="light" rounding="full" shape="rectangle" size="l">
+    <sp-swatch color="var(--spectrum-blue-500)"></sp-swatch>
+    <sp-swatch color="var(--spectrum-indigo-500)"></sp-swatch>
+    <sp-swatch color="var(--spectrum-purple-500)"></sp-swatch>
+</sp-swatch-group>
+```
+
+### Accessibility
```
rise-erpelding, line 323: Another thing that could be mentioned here is that it looks like [swatch group sets the role on the swatch](https://github.com/adobe/spectrum-web-components/blob/main/packages/swatch/src/SwatchGroup.ts#L270-L276) depending on whether it's single select, multi select, or not selectable.

It'll also set the role on the swatch group if it's in a select mode (either to `group` or `radiogroup`).

packages/swatch/swatch-group.md
```diff
@@ -7,23 +7,25 @@ An `<sp-swatch-group>` group is a grouping of `<sp-swatch>` elements that are re
...
 
-```
+```js
 import { SwatchGroup } from '@spectrum-web-components/swatch';
 ```
 
-## Sizes
+### Options
+
+#### Sizes
```
rise-erpelding, line 28: Could be worth noting that the `size` property "only affects the size of each individual swatch, not the spacing between them" according to the [design docs](https://spectrum.adobe.com/page/swatch-group/#Size).

packages/swatch/swatch-group.md
```diff
@@ -144,15 +150,20 @@ The `density` attribute/property is not required and when applied accepts the va
 </sp-swatch-group>
 ```
 
-## Selection
+</sp-tab-panel>
+</sp-tabs>
+
+#### Selection
 
 An `<sp-swatch-group>` element can carry a selection of a `single` swatch or of `multiple` swatches. Then the `selects` property is set to one of these values, the `selected` property will surface an array the represents the string values that have been selected in the UI.
```
rise-erpelding: Not sure "single" and "multiple" need code formatting in this context.

```suggestion
An `<sp-swatch-group>` element can carry a selection of a single swatch or of multiple swatches. Then the `selects` property is set to one of these values, the `selected` property will surface an array the represents the string values that have been selected in the UI.
```

But I also see that the documentation means to say you can set it to either of those values, so maybe a rewording would be better?

```suggestion
An `<sp-swatch-group>` element can be configured to allow either single or multiple swatch selection. When the `selects` property is set to `single` or `multiple`, the `selected` property will contain an array of string values representing the currently selected swatches in the UI.
```

packages/swatch/swatch-group.md
```diff
@@ -144,15 +150,20 @@ The `density` attribute/property is not required and when applied accepts the va
...
 
 When the value of `selected` is updated via user input, the `change` event will be dispatched on the `<sp-swatch-group>` element to announce that interaction. Calling `preventDefault()` on the `chagne` event will prevent both the `<sp-swatch-group>` and the `<sp-swatch>` that initiated the `change` interaction from updating their `selected` values.
```
rise-erpelding: A misspelling of "change" here:

```suggestion
When the value of `selected` is updated via user input, the `change` event will be dispatched on the `<sp-swatch-group>` element to announce that interaction. Calling `preventDefault()` on the `change` event will prevent both the `<sp-swatch-group>` and the `<sp-swatch>` that initiated the `change` interaction from updating their `selected` values.
```

packages/swatch/swatch-group.md
```diff
@@ -144,15 +150,20 @@ The `density` attribute/property is not required and when applied accepts the va
...
 
 The value of `selected` can also be provided directly from the `<sp-swatch>` children. Child `<sp-swatch>` elements with their own `selected` attribute will be gathered and merged with any other selection data on the `<sp-swatch-group>` parent to populate `selected`.
 
-### Single
+<sp-tabs selected="single" auto label="Selection Attribute Options">
+<sp-tab value="single">Single</sp-tab>
+<sp-tab-panel value="single">
 
 The `selected` property is always represented as an array, and as such an application leveraging an `<sp-swatch-group>` element can apply more than one selection, regardless of the vaue of `selects`, however all future interactions will force the interace to a single selection.
```
rise-erpelding: This part was a worded in a way that confused me. Suggesting:

```suggestion
The `selected` property is always an array, so applications can programmatically set multiple selections regardless of the `selects` value. However, user interactions will respect the `selects` property and enforce either single or multiple selection behavior accordingly.
```

packages/swatch/swatch-group.md
```diff
@@ -163,16 +174,16 @@ The `selected` property is always represented as an array, and as such an applic
 >
     <sp-swatch color="var(--spectrum-blue-500)"></sp-swatch>
     <sp-swatch color="var(--spectrum-indigo-500)"></sp-swatch>
-    <sp-swatch color="var(--spectrum-purple-500)" selected></sp-swatch>
```
rise-erpelding, line 166: Oh I see. I guess we had multiple `selected`s on the swatches in this group to illustrate that, at least initially if it's programmatically set that way, the array can still have multiple elements even in single select mode, even though you can only actually select one swatch.

I'm on board with removing this nuance in the example. It really doesn't feel much like a feature to me. Plus the `selected` array switches from two elements back to just one once you select a different swatch.

packages/swatch/swatch-group.md
```diff
@@ -181,15 +192,20 @@ The `selected` property is always represented as an array, and as such an applic
...
+    <sp-swatch color="var(--spectrum-blue-500)" selected></sp-swatch>
     <sp-swatch color="var(--spectrum-indigo-500)"></sp-swatch>
-    <sp-swatch color="var(--spectrum-purple-500)"></sp-swatch>
+    <sp-swatch color="var(--spectrum-purple-500)" selected></sp-swatch>
     <sp-swatch color="var(--spectrum-fuchsia-500)"></sp-swatch>
     <sp-swatch color="var(--spectrum-magenta-500)"></sp-swatch>
 </sp-swatch-group>
-<div>Selected: [ ]</div>
+<div>
+    Selected: [ "var(--spectrum-blue-500)", "var(--spectrum-purple-500)" ]
```
rise-erpelding, line 220: Would these arrays look better with code formatting? I'm on the fence; not sure I have a huge preference either way.

packages/swatch/swatch-group.md
```diff
@@ -181,15 +192,20 @@ The `selected` property is always represented as an array, and as such an applic
...
+</div>
 ```
 
+</sp-tab-panel>
+</sp-tabs>
+
 ### Value
 
 When available, the value of `selected` will be constructed from the `value` attributes/properties of the child `<sp-swatch>` elements. This can be useful when you would like the swatch data needs to correlate to a hash rather than the raw color string.
```
rise-erpelding: I got a little lost in the last sentence, I think "like" and "needs" are maybe conflicting. Suggesting:

```suggestion
When available, the `selected` property will be populated with the `value` attributes/properties of the child `<sp-swatch>` elements. This is useful when you need the swatch data to correlate to a hash or identifier rather than the raw color string.
```

packages/swatch/swatch-group.md
```diff
@@ -116,11 +118,13 @@ import { SwatchGroup } from '@spectrum-web-components/swatch';
 </sp-tab-panel>
 </sp-tabs>
 
-## Density
+#### Density
 
 The `density` attribute/property is not required and when applied accepts the values of `compact` or `spacious`.
 
-### Compact
+<sp-tabs selected="compact" auto label="Density Attribute Options">
```
marissahuysentruyt, line 127: What do you think about including "regular" or "default" with the density tabs? Then we could say something to the effect of "if a value for `density` isn't defined, the default spacing will render.' (that sentence needs help, but you get my point!) It feels like it makes sense to show the 3 options next to each other.

packages/swatch/swatch-group.md
```diff
@@ -217,6 +233,27 @@ When available, the value of `selected` will be constructed from the `value` att
 <div>Selected: [ "color-2", "color-1", "color-3" ]</div>
 ```
 
-## Swatch modifying attributes
+### Swatch modifying attributes
+
+An `<sp-swatch-group>` element can be modified by the following attributes/properties to customize its delivery as desired for your use case: `border`, `disabled`, `mixedValue` (accepted as the `mixed-value` attribute), `nothing`, `rounding`, `shape`, and `size`. Use these in concert with each other for a variety of final visual deliveries. Applying a value for one of these attributes/properties to an `<sp-swatch-group>` element will have it forward the value to all of the `<sp-swatch>` elements that are a direct child of the group, overriding any value that may be applied directly to those children. For example:
+
+```html
+<sp-swatch-group border="light" rounding="full" shape="rectangle" size="l">
```
rise-erpelding: The `rounding="full"` and `shape="rectangle"` conflict with each other here, because you can't have both full rounding (circular swatches) and rectangular swatches. It might be worth considering other ways to display this, or an additional example/examples. I do really like that there's an example included now though. It'd be cool to see some of these properties being applied to the individual swatches so we can see how the swatch group overrides them.

packages/swatch/swatch-group.md
marissahuysentruyt: I have a couple of ideas here that might need some massaging, but I think we clear some things up. 

First, I think maybe we should use the term "selection mode." I've seen that in React, and that makes it a little clearer (at least to me) what exactly we're talking about- that we have a "single selection" mode where a user can only select 1 swatch, while in "multiple selection" mode a user can select more than 1. 

Next, I think we need to clarify that `selected` is both on the `sp-swatch` based on the `change` event, and on the `sp-swatch-group`. I'll leave another comment that sort of rearranges some things in this paragraph to make it a little clearer which things are which.

marissahuysentruyt: There's a typo in `change` (it's `chagne` right now.

packages/swatch/swatch-group.md
```diff
@@ -144,15 +150,20 @@ The `density` attribute/property is not required and when applied accepts the va
...
+</sp-tab-panel>
+</sp-tabs>
+
+#### Selection
 
 An `<sp-swatch-group>` element can carry a selection of a `single` swatch or of `multiple` swatches. Then the `selects` property is set to one of these values, the `selected` property will surface an array the represents the string values that have been selected in the UI.
 
 When the value of `selected` is updated via user input, the `change` event will be dispatched on the `<sp-swatch-group>` element to announce that interaction. Calling `preventDefault()` on the `chagne` event will prevent both the `<sp-swatch-group>` and the `<sp-swatch>` that initiated the `change` interaction from updating their `selected` values.
 
 The value of `selected` can also be provided directly from the `<sp-swatch>` children. Child `<sp-swatch>` elements with their own `selected` attribute will be gathered and merged with any other selection data on the `<sp-swatch-group>` parent to populate `selected`.
```
marissahuysentruyt:
```suggestion
An `<sp-swatch-group>` element has two selection modes. The `selects` property can be set to either `single`, indicating a user can choose up to one swatch, or `multiple`, indicating a user can choose more than one swatch. 

When the `selects` property is set to one of these values, the `selected` property of `<sp-swatch-group>` will surface an array the represents the string values that have been selected in the UI. This `selected` property can be populated in two ways: though user input, or directly from `<sp-swatch>` children.

When the value of `selected` is updated via user input, the `change` event will be dispatched on the `<sp-swatch-group>` element to announce that interaction. Calling `preventDefault()` on the `change` event will prevent both the `<sp-swatch-group>` and the `<sp-swatch>` that initiated the `change` interaction from updating their `selected` values.

The value of the `selected` property also can be provided directly from the `<sp-swatch>` children. Child `<sp-swatch>` elements with their own `selected` attribute will be gathered and merged with any other selection data on the `<sp-swatch-group>` parent to populate `selected`.
```

packages/swatch/swatch-group.md
marissahuysentruyt, line 166: good find!

packages/swatch/swatch-group.md
marissahuysentruyt: There's a typo with "value:" regardless of the _value_ of `selects`...

packages/swatch/swatch-group.md
```diff
@@ -217,6 +233,27 @@ When available, the value of `selected` will be constructed from the `value` att
 <div>Selected: [ "color-2", "color-1", "color-3" ]</div>
 ```
 
-## Swatch modifying attributes
+### Swatch modifying attributes
```
marissahuysentruyt, line 254: We need examples of each of these attributes! They don't quite all work together, (like you can't have a rectangle swatch, but also have full rounding because full rounding makes it a circle 🤔 ), so this might be a good place for a set of tabs again. 

I am imagining tabs for: 
- shape="rectangle"
- rounding="full"
- rounding="none"

Screenshot 2025-08-19 at 5 34 06 PM

- border="light" 
- border="none"

Screenshot 2025-08-19 at 5 34 49 PM

are any of those specific to the swatch component, as opposed to the swatch _group_ component? The only one that I think might be swatch-specific would be the border property.

packages/swatch/swatch-group.md
```diff
@@ -217,6 +233,27 @@ When available, the value of `selected` will be constructed from the `value` att
...
+
+An `<sp-swatch-group>` element can be modified by the following attributes/properties to customize its delivery as desired for your use case: `border`, `disabled`, `mixedValue` (accepted as the `mixed-value` attribute), `nothing`, `rounding`, `shape`, and `size`. Use these in concert with each other for a variety of final visual deliveries. Applying a value for one of these attributes/properties to an `<sp-swatch-group>` element will have it forward the value to all of the `<sp-swatch>` elements that are a direct child of the group, overriding any value that may be applied directly to those children. For example:
```
marissahuysentruyt: Does SWC support all of these attributes on the swatch group? Maybe I'm adding it wrong, but if I put `disabled` on the sp-swatch-group nothing happens. Same with mixed value and nothing- I don't see stories for those use cases either. 

Disabled, nothing, and mixedValue all work on individual swatches, but _not_ for swatch group, so maybe we leave those off?

packages/swatch/swatch-group.md
```diff
@@ -217,6 +233,27 @@ When available, the value of `selected` will be constructed from the `value` att
...
+    <sp-swatch color="var(--spectrum-purple-500)"></sp-swatch>
+</sp-swatch-group>
+```
+
+### Accessibility
+
+#### Best practices
+
+- Ensure swatches have sufficient color contrast for visibility.
+- Verify that swatches are appropriately labeled for screen readers.
```
marissahuysentruyt: Can we expand on this any further? I'm assuming they would be labeled with `aria-label`?

marissahuysentruyt (changes requested):
Nice work. I think my biggest comment would be to add more of examples of the other modifying attributes! I listed out some ideas and left some screenshots from the CSS repo, that I think we should make sure we capture here!

packages/swatch/swatch-group.md
aramos-adobe: I like that idea!

packages/swatch/swatch-group.md
aramos-adobe: @rise-erpelding, Good catch on this!

packages/swatch/swatch-group.md
```diff
@@ -144,17 +166,24 @@ The `density` attribute/property is not required and when applied accepts the va
...
-## Selection
+</sp-tab-panel>
+</sp-tabs>
+
+#### Selection mode
+
+An `<sp-swatch-group>` element has two selection modes. The `selects` property can be set to either `single`, indicating a user can choose up to one swatch, or `multiple`, indicating a user can choose more than one swatch.
 
-An `<sp-swatch-group>` element can carry a selection of a `single` swatch or of `multiple` swatches. Then the `selects` property is set to one of these values, the `selected` property will surface an array the represents the string values that have been selected in the UI.
+When the `selects` property is set to one of these values, the `selected` property of `<sp-swatch-group>` will surface an array the represents the string values that have been selected in the UI. This `selected` property can be populated in two ways: though user input, or directly from `<sp-swatch>` children.
```
marissahuysentruyt: Ah shoot- I might have suggested a typo 😩 Sorry about that.

It should be "through user input."

packages/swatch/swatch-group.md
```diff
@@ -217,6 +251,65 @@ When available, the value of `selected` will be constructed from the `value` att
...
+```html
+<sp-swatch-group shape="rectangle">
+    <sp-swatch color="var(--spectrum-blue-500)"></sp-swatch>
+    <sp-swatch color="var(--spectrum-indigo-500)"></sp-swatch>
+    <sp-swatch color="var(--spectrum-purple-500)"></sp-swatch>
+</sp-swatch-group>
+```
+
+</sp-tab-panel>
+<sp-tab value="rounding">Rounding</sp-tab>
```
marissahuysentruyt, line 273: Do you think it is useful to show the `rounding="none"` option? According to [the design guidelines](https://spectrum.adobe.com/page/swatch-group/#Corner-rounding-in-swatch-groups), that's actually the preferred/default corner rounding for swatch groups. We haven't shown that anywhere in the docs, however.

Along the same lines, do you think it would be useful to show the `border="none"` option? I was just thinking we should show examples of all of the available values for those attributes, since it's not super clear just from the API table.

packages/swatch/swatch-group.md
```diff
@@ -144,17 +166,24 @@ The `density` attribute/property is not required and when applied accepts the va
...
+The value of the `selected` property also can be provided directly from the `<sp-swatch>` children. Child `<sp-swatch>` elements with their own `selected` attribute will be gathered and merged with any other selection data on the `<sp-swatch-group>` parent to populate `selected`.
 
-### Single
+<sp-tabs selected="single" auto label="Selection Attribute Options">
+<sp-tab value="single">Single</sp-tab>
+<sp-tab-panel value="single">
 
-The `selected` property is always represented as an array, and as such an application leveraging an `<sp-swatch-group>` element can apply more than one selection, regardless of the vaue of `selects`, however all future interactions will force the interace to a single selection.
+The `selected` property is always an array, so applications can programmatically set multiple selections regardless of the value of `selects`. However, user interactions will respect the `selects` property and enforce either single or multiple selection behavior accordingly.
```
rise-erpelding: Looking over this again, I'm wondering if this might fit better outside of `<sp-tabs>`?

Just because before it was showing a single select example that had 2 elements in the array because they were programmatically placed there, and now we have a single select example that only is programmatically selecting 1 swatch so therefore will _always_ have 1 element in the array.

Still cool with this example as is, but the explanation no longer fits the example so probably doesn't need to be in the tab panel, even though it's useful information.

packages/swatch/swatch-group.md
rise-erpelding: image

It's unfortunate that the import of the border/rounding/shape types doesn't seem to work properly for the API table, since we end up with just the name of the type (for instance `SwatchBorder`) instead of the actual values it can take.

Either way, I think the swatch docs can do a lot of the heavy lifting for showing all the different features and options for a swatch and we don't necessarily need to repeat all of that, and it might be fine to just link to swatch somewhere around here instead of describing all the options again.

What I _would_ want to show/tell for swatch group though would include:
* No rounding since the design docs state that that's the preference within a group
* Something about the border but I can't remember off the top of my head what's the most relevant within a swatch group... is the preference light border? Does it depend on the swatch's color contrast?
* The text here talks about how you can put a property on the swatch group and it overrides any properties applied to the swatch, so I think that's what I'd want to see in the code snippet(s). If we set one rounding on the swatch group, and another on the swatch, the one on the group should take precedence. If we set a size on the swatch group, and another on the swatch, the one on the group would take precedence, etc.

packages/swatch/swatch-group.md
aramos-adobe: Ah okay I understand what you mean now !

marissahuysentruyt (approved):
offlinegranny-hellyeah

rise-erpelding (approved):

PR #5703, opened 2025-08-20
chore: initial POC of prefixed tag names

benjamind:
## Description

This PR explores a POC for adding prefixing of spectrum-web-component custom element registrations, allowing us to avoid versioning conflicts (at the expense of payload).

## Related issue(s)

-   fixes [Issue Number]

---

## Author's checklist

---

## Reviewer's checklist

### Manual review test cases

    1. Go [here](url)
    2. Do this action
    3. Expect this result

    1. Go [here](url)
    2. Do this action
    3. Expect this result

### Device review

### DIFF
--- a/packages/action-bar/src/ActionBar.ts
+++ b/packages/action-bar/src/ActionBar.ts
@@ -12,12 +12,12 @@
 
 import {
     CSSResultArray,
-    html,
-    SpectrumElement,
+    staticHtml as html,
+    LitElement,
+    SpectrumMixin,
     TemplateResult,
 } from '@spectrum-web-components/base';
 import { property } from '@spectrum-web-components/base/src/decorators.js';
-import '@spectrum-web-components/popover/sp-popover.js';
 import '@spectrum-web-components/action-group/sp-action-group.js';
 import '@spectrum-web-components/button/sp-close-button.js';
 import '@spectrum-web-components/field-label/sp-field-label.js';
@@ -26,11 +26,15 @@ import { ifDefined } from '@spectrum-web-components/base/src/directives.js';
 import { FocusVisiblePolyfillMixin } from '@spectrum-web-components/shared/src/focus-visible.js';
 export const actionBarVariants = ['sticky', 'fixed'];
 
+import { Popover } from '@spectrum-web-components/popover/src/Popover.js';
+
 /**
  * @element sp-action-bar
  * @slot - Content to display with the Action Bar
  */
-export class ActionBar extends FocusVisiblePolyfillMixin(SpectrumElement) {
+export class ActionBar extends FocusVisiblePolyfillMixin(
+    SpectrumMixin(LitElement, 'action-bar')
+) {
     public static override get styles(): CSSResultArray {
         return [actionBarStyles];
     }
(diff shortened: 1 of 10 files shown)


benjamind:
Initial tachometer results show this is actually faster than the original registration which I find hard to believe, but may be due to my having updated to the latest playwright install.

```
┌─────────────┬───────────────────────┐
│   Benchmark │ action-bar:basic-test │
├─────────────┼───────────────────────┤
│     Browser │ chrome-headless       │
│             │ 139.0.0.0             │
├─────────────┼───────────────────────┤
│ Sample size │ 50                    │
└─────────────┴───────────────────────┘

┌─────────┬────────────┬───────────────────┬─────────────────┬─────────────────┐
│ Version │ Bytes      │          Avg time │       vs remote │             vs  │
├─────────┼────────────┼───────────────────┼─────────────────┼─────────────────┤
│ remote  │ 614.21 KiB │ 15.17ms - 15.35ms │                 │          slower │
│         │            │                   │        -        │         2% - 3% │
│         │            │                   │                 │ 0.24ms - 0.47ms │
├─────────┼────────────┼───────────────────┼─────────────────┼─────────────────┤
│ <none>  │ 590.44 KiB │ 14.84ms - 14.97ms │          faster │                 │
│         │            │                   │         2% - 3% │        -        │
│         │            │                   │ 0.24ms - 0.47ms │                 │
└─────────┴────────────┴───────────────────┴─────────────────┴─────────────────┘
```

graynorton (commented):
This general approach looks promising to me!

I wanted to see how it might look end-to-end, so I created a Lit Playground example to try it out:

https://lit.dev/playground/#gist=039ebbad69ce634ff3cf6d4d14c7df01

In doing so, I found that there were a couple of issues:

* We need to provide `StaticValue`s to render into Lit templates in the tag-name position; strings aren't sufficient
* The logic for rendering dependent tag names needs to combine the `prefix` from the current component instance with the `tagName` from the dependency's constructor; as written in this PR, it gets both from the base constructor, so doesn't utilize the custom prefix

In the process of addressing these, I ended up making a bunch of other experimental tweaks, primarily in the interest of improving dev experience.

Take a look at the Playground and see what you think. In particular, I ended up going down a path that tried to adhere closely to Lit's template-safety best practices, so requires use of a tagged template literal when defining `prefix` and `tagName`. This is responsible for a bit of complexity in the implementation, so I could imagine deciding that plain strings + `unsafeStatic` are good enough for our use case, but I wanted to see what the more rigorous approach would look like.

packages/action-bar/src/ActionBar.ts
```diff
@@ -97,7 +101,7 @@ export class ActionBar extends FocusVisiblePolyfillMixin(SpectrumElement) {
 
     public override render(): TemplateResult {
         return html`
-            <sp-popover ?open=${this.open} id="popover">
+            <${Popover.tag} ?open=${this.open} id="popover">
```
castastrophe, line 104: Love this! We used to do something similar at Red Hat at one point though we used the tag attribute for the CMS rather than internal component referencing.

jcmitch:
This seems very useful as a replacement for our existing solution. The only additional callout I'd add is that in our implementation we currently not only have to rewrite tag names for registration but there are a number of selectors scattered throughout the repo which use tag names. Ideally we would just move away from tag based selectors. 
https://github.com/adobe/spectrum-web-components/blob/v1.6.0/packages/dialog/src/DialogWrapper.ts#L81

PR #5718, opened 2025-09-03
feat(badge): s2 styling and render brought in

castastrophe:
## Related issue(s)

-   fixes [Issue Number]

---

## Author's checklist

---

## Reviewer's checklist

### Manual review test cases

    1. Go [here](url)
    2. Do this action
    3. Expect this result

    1. Go [here](url)
    2. Do this action
    3. Expect this result

### Device review


second-gen/packages/swc/components/progress-circle/ProgressCircle.ts
```diff
@@ -46,29 +47,51 @@ export class ProgressCircle extends ProgressCircleBase {
...
-            this.makeRotation(
-                -180 + (180 / 50) * Math.max(this.progress - 50, 0)
-            ),
-        ];
-        const masks = ['Mask1', 'Mask2'];
+        // SVG strokes are centered, so subtract half the stroke width from the radius to create an inner stroke.
+        const radius = `calc(50% - ${this.strokeWidth / 2}px)`;
+
         return html`
-            <slot @slotchange=${this.handleSlotchange}></slot>
```
castastrophe, line 57: For the sake of simplifying this for S2, I've regressed some of this functionality by removing the slot and leveraging our SVG approach. Would love to have a larger conversation around this for S2 but this seemed sufficient for the base requirement of getting the S2 styles and rendering into the component.

second-gen/packages/core/components/progress-circle/ProgressCircle.base.ts
```diff
@@ -34,10 +34,11 @@ export abstract class ProgressCircleBase extends SizedMixin(SpectrumElement, {
 
     /**
      * Static color variant for use on different backgrounds.
-     * When set to 'white', the component uses white styling for dark backgrounds.
+     * When set to 'white', the component uses white styling for images with a dark tinted background.
+     * When set to 'black', the component uses black styling for images with a light tinted background.
      */
     @property({ reflect: true, attribute: 'static-color' })
-    public staticColor?: 'white';
+    public staticColor?: 'white' | 'black';
```
castastrophe: This is an example where S1 supports only static color white and S2 has styles for static color white & black. What's the current thinking on handling that?

second-gen/packages/core/components/progress-circle/ProgressCircle.base.ts
```diff
@@ -46,6 +47,24 @@ export abstract class ProgressCircleBase extends SizedMixin(SpectrumElement, {
     @property({ type: Number })
     public progress = 0;
 
+    /**
+     * Stroke width for the progress circle.
+     */
+    @property({ type: Number, reflect: false })
+    public get strokeWidth(): number {
```
castastrophe: This seems a nicer API than the current but would be good to have a discussion re: handling the evolution of the APIs between S1 & S2. S2 progress circle _could_ have a new strokeWidth customization and eschew the current makeRotation function in favor of letting CSS do the calculations (sans transformation).

second-gen/packages/swc/components/badge/Badge.ts
```diff
@@ -52,16 +54,36 @@ export class Badge extends BadgeBase {
...
+                        typeof this.size !== 'undefined',
+                    [`spectrum-Badge--${this.variant}`]:
+                        typeof this.variant !== 'undefined',
+                    [`spectrum-Badge--subtle`]: this.subtle,
+                    [`spectrum-Badge--outline`]: this.outline,
+                    [`spectrum-Badge--fixed-${this.fixed}`]:
+                        typeof this.fixed !== 'undefined',
+                })}
+            >
+                ${when(
```
castastrophe, line 101: I updated the ternary to use `when` instead but let me know if there's a reason not to use that here.

second-gen/packages/swc/components/badge/Badge.ts
```diff
@@ -52,16 +54,36 @@ export class Badge extends BadgeBase {
...
+                                [`spectrum-Badge-icon`]: true,
+                                [`spectrum-Badge-icon--no-label`]:
+                                    !this.slotHasContent,
+                            })}
+                        >
+                            <slot name="icon"></slot>
+                        </div>
+                    `
+                )}
+                <div class="spectrum-Badge-label">
```
castastrophe, line 115: I tested adding this class to the slot itself and it didn't render padding correctly. Maybe just a case of needing the appropriate display settings but I didn't dig into it too much. Would be good to have a standard "best practice" in place for whether we want to wrap slots to apply styles or style slot containers directly.

second-gen/packages/core/components/badge/Badge.base.ts
```diff
@@ -37,6 +37,11 @@ export const BADGE_VARIANTS = [
     'green',
     'cyan',
     'blue',
+    'pink',
```
castastrophe: Another question here on handling new variant options. S1 badge has variants but not styles for these additional options. How do we want to handle extending this and adding the new options for S2-only?

second-gen/packages/core/components/badge/Badge.base.ts
```diff
@@ -49,8 +54,11 @@ export type FixedValues = (typeof FIXED_VALUES)[number];
 
 /**
  * @element sp-badge-base
- * @slot - The text label to display in the badge.
- * @slot icon - The icon to display in the badge.
+ * @property {BadgeVariant} variant - The variant of the badge.
+ * @property {boolean} subtle - Whether the badge is subtle.
+ * @property {boolean} outline - Whether the badge is outlined.
```
castastrophe: Subtle & outline are new to S2

second-gen/packages/core/components/badge/Badge.base.ts
```diff
@@ -61,26 +69,14 @@ export abstract class BadgeBase extends SizedMixin(
...
     public variant: BadgeVariant = 'informative';
 
-    @property({ reflect: true })
-    public get fixed(): FixedValues | undefined {
-        return this._fixed;
-    }
+    @property({ type: Boolean, reflect: true })
+    public subtle: boolean = false;
 
-    public set fixed(fixed: FixedValues | undefined) {
```
castastrophe, line 69: This seems to me to be how the @property works by default. I don't see any regressions when I simplify this but let me know if I've missed a nuance here.

second-gen/packages/swc/components/progress-circle/stories/progress-circle.stories.ts
```diff
@@ -139,7 +139,7 @@ export const Indeterminate: Story = {
 export const StaticWhite: Story = {
     render: () => html`
         <div
-            style="background-color: rgba(0,0,0,0.4); padding: 24px; display: flex; gap: 24px; align-items: center;"
+            style="background: linear-gradient(45deg, rgb(64 0 22), rgb(14 24 67)); padding: 24px; display: flex; gap: 24px; align-items: center;"
```
castastrophe: Added our pretty gradients here for static white/black which match how design demos these styles. They're meant to be on visually busy backgrounds like gradients or images so that drives home that guidance in Storybook as well.

second-gen/packages/swc/components/progress-circle/stories/progress-circle.stories.ts
```diff
@@ -189,9 +216,3 @@ export const IndeterminateStaticWhite: Story = {
         </div>
     `,
 };
-
-export const WithSlottedContent: Story = {
```
castastrophe, line 193: I removed the slot for the purposes of my S2 update but we can always bring this back.

second-gen/packages/swc/components/progress-circle/ProgressCircle.ts
```diff
@@ -11,12 +11,18 @@
...
 
 import { CSSResultArray, html, TemplateResult } from 'lit';
-import { ifDefined } from 'lit/directives/if-defined.js';
+import { classMap } from 'lit/directives/class-map.js';
 
 import { ProgressCircleBase } from '@swc/core/components/progress-circle';
 
 import progressCircleStyles from './progress-circle.css';
 
+function capitalize(str?: string): string {
```
castastrophe: I kind of liked this utility living separately but let me know how you want to handle this in the SWC infrastructure.

second-gen/packages/core/components/badge/Badge.base.ts
```diff
@@ -16,13 +16,16 @@ import { SizedMixin, SpectrumElement } from '@swc/core/shared/base';
 import { ObserveSlotPresence } from '@swc/core/shared/observe-slot-presence';
 import { ObserveSlotText } from '@swc/core/shared/observe-slot-text';
 
-export const BADGE_VARIANTS = [
+export const BADGE_VARIANTS_SEMANTIC = [
```
castastrophe, line 19: Separating out the semantic variants from the color variants helps us in Storybook (and possibly down the road) because only semantic variants support outline styling.

second-gen/packages/core/components/badge/Badge.base.ts
```diff
@@ -37,11 +40,11 @@ export const BADGE_VARIANTS = [
     'green',
     'cyan',
     'blue',
-    'pink',
```
castastrophe: Since these don't exist in S1, I removed these additions and moved them into the S2 component file as an extension of the existing base.

second-gen/packages/core/components/badge/Badge.base.ts
```diff
@@ -55,8 +58,6 @@ export type FixedValues = (typeof FIXED_VALUES)[number];
 /**
  * @element sp-badge-base
  * @property {BadgeVariant} variant - The variant of the badge.
- * @property {boolean} subtle - Whether the badge is subtle.
```
castastrophe: This only exists in S2's new styles so I'm testing out moving these into the S2-only component.

second-gen/packages/swc/components/badge/Badge.ts
```diff
@@ -11,16 +11,39 @@
...
 import { when } from 'lit/directives/when.js';
 
-import { BadgeBase } from '@swc/core/components/badge';
+import {
+    BADGE_VARIANTS_COLOR as BADGE_VARIANTS_COLOR_BASE,
+    BADGE_VARIANTS_SEMANTIC,
+    BadgeBase,
+} from '@swc/core/components/badge';
+
+export const BADGE_VARIANTS_COLOR = [
```
castastrophe, line 24: This re-exports the const with the addition of the new colors.

second-gen/packages/swc/components/badge/Badge.ts
```diff
@@ -11,16 +11,39 @@
...
+    'brown',
+    'cinnamon',
+    'silver',
+] as const;
+
+export const BADGE_VARIANTS = [
+    ...BADGE_VARIANTS_SEMANTIC,
+    ...BADGE_VARIANTS_COLOR,
+] as const;
+export type BadgeVariant = (typeof BADGE_VARIANTS)[number];
```
castastrophe, line 37: I'm not sure if I need to redefine this since technically it's the same but the content of the BADGE_VARIANTS has changed so I went ahead and did that.

second-gen/packages/core/components/badge/Badge.base.ts
```diff
@@ -49,8 +57,9 @@ export type FixedValues = (typeof FIXED_VALUES)[number];
 
 /**
  * @element sp-badge-base
- * @slot - The text label to display in the badge.
```
castastrophe, line 52: Since this doesn't have a render, should we define these only on the render layer?

second-gen/packages/core/components/progress-circle/ProgressCircle.base.ts
```diff
@@ -46,6 +47,24 @@ export abstract class ProgressCircleBase extends SizedMixin(SpectrumElement, {
...
+        return this._strokeWidth;
+    }
+
+    public set strokeWidth(customWidth: unknown) {
+        if (customWidth && customWidth instanceof Number) {
+            this._strokeWidth = customWidth as number;
+        }
+
+        this._strokeWidth = this.size === 's' ? 2 : this.size === 'l' ? 4 : 3;
+    }
```
aramos-adobe:
```js
    public set strokeWidth(customWidth: unknown) {
        if (customWidth && customWidth instanceof Number) {
            this._strokeWidth = customWidth as number;
            this._customStrokeWidth = true;
        } else {
            this._strokeWidth =
                this.size === 's' ? 2 : this.size === 'l' ? 4 : 3;
            this._customStrokeWidth = false;
        }
    }
```

I like the addition of customWidth, not sure how we'd introduce this concept though. 

For now, I'm just exploring how to get the svg to fit within the parent container.

second-gen/packages/core/components/progress-circle/ProgressCircle.base.ts
```diff
@@ -46,6 +47,24 @@ export abstract class ProgressCircleBase extends SizedMixin(SpectrumElement, {
...
+
+    private _strokeWidth: number = 2;
```
aramos-adobe:
```js
    private _strokeWidth: number = 3;
    private _customStrokeWidth: boolean = false;
```

Again, maybe you have a better idea, but trying to figure out why the changes weren't updating.

second-gen/packages/core/components/progress-circle/ProgressCircle.base.ts
aramos-adobe, line 1: in `protected override update` I added this to update the stroke width per shirt size:

```js
// Update stroke width when size changes (if no custom width was set)
        if (changes.has('size') && !this._customStrokeWidth) {
            // Force recalculation by calling the setter
            this.strokeWidth = undefined;
        }
```

aramos-adobe:
@castastrophe The S2 addition of these are sooooo lovely!! I'd love to contribute to this!!

second-gen/packages/core/components/progress-circle/ProgressCircle.base.ts
castastrophe: Yes! Great points. We need to be accounting for those changes so we can trigger repaint.

second-gen/packages/swc/components/badge/Badge.ts
```diff
@@ -28,14 +53,16 @@ export type { BadgeVariant, FixedValues } from '@swc/core/components/badge';
  * @since 1.0.0
  * @status stable
  * @github https://github.com/adobe/spectrum-web-components/tree/main/...
- * @figma https://www.figma.com/design/...
+ * @figma https://www.figma.com/design/Mngz9H7WZLbrCvGQf3GnsY/S2-%2F-Desktop?node-id=36806-6551
```
marissahuysentruyt, line 56: At one point, we talked about not sharing Figma links. Are we concerned at all with committing this link to the repo? I know we have it in CSS, but I thought that ended up being sort of a no-no. Or was that just for in PR descriptions? Or am I misremembering (which could totally be)?

second-gen/packages/swc/components/badge/stories/badge.stories.ts
```diff
@@ -10,81 +10,233 @@
...
+
+/** @todo Pull this up into a decorator for all stories to leverage */
+const CONTAINER = (content: TemplateResult<1>[]) =>
+    html`<div
+        style=${styleMap({
+            display: 'flex',
+            gap: 'var(--spectrum-spacing-200)',
+            'flex-wrap': 'wrap',
+            'justify-content': 'center',
+            // Used 80ch because that's generally considered the maximum readable width for text in a web page.
```
marissahuysentruyt, line 42: 👏 ❤️ :accessibility:

second-gen/packages/swc/components/badge/Badge.ts
castastrophe: Great question! I'm not sure. I know we don't post Jira links publicly but I thought since the Figma assets are behind a login, it was okay to include them in Storybook. @najikahalsema do you know?

second-gen/packages/swc/components/badge/Badge.ts
castastrophe: @rubencarvalho Do you know how that works in Typescript? Would it calculate the value of BadgeVariant at the time of use based on the new definition of BADGE_VARIANTS?

second-gen/packages/core/components/badge/Badge.base.ts
castastrophe, line 19: @rubencarvalho Initially I started updating these assets per the S2 guidelines before I realized this is the base class being imported into S1 components as well. Would it make sense to hoist this folder to sit between the first- and second-get folders so it's clear that it's a shared resource?

second-gen/packages/core/components/progress-circle/ProgressCircle.base.ts
castastrophe: @rubencarvalho @graynorton Should this property be moved entirely to the separate base classes? staticColor white only for S1 and staticColor white & black for S2?

second-gen/packages/core/components/progress-circle/ProgressCircle.base.ts
castastrophe: I might remove this entirely. After talking with @aramos-adobe, it doesn't make sense to allow strokeWidth to vary because it corresponds to the input step and since this represents a percent out of 100, we likely only want to support single steps.

second-gen/packages/swc/components/progress-circle/ProgressCircle.ts
castastrophe: Moved over to PR 5743 focused on progress circle.

second-gen/packages/core/components/progress-circle/ProgressCircle.base.ts
castastrophe: Moved over to PR 5743 focused on progress circle.

second-gen/packages/core/components/badge/Badge.base.ts
graynorton: Yes, this seems like a good idea!

graynorton: I think the key here is just going to be to make it clear (via docs and other means) that the S1/S2 distinction is not the same as the 1st-gen/2nd-gen distinction.

The `core` library is in 2nd-gen because it, along with 2nd-gen SWC, is going to use our 2nd-gen tooling and infrastructure, and because we have a sense that the `core` layer may prove useful even after the 1st-gen to 2nd-gen transition is complete. (If it does not, we can remove it when we eventually stop supporting 1st-gen.)

As a layer of the architecture that explicitly supports both S1 and S2, I think it's actually OK if `core` ends up including some type definitions that are S1- and S2-aware, allowing us to manage the differences in one place. I've done some further exploration of what this might look like that I'll put in a separate PR.

second-gen/packages/swc/components/badge/Badge.ts
graynorton: 👍 Since the 2nd-gen `render()` method is conceptually free to diverge from the 1st-gen method in whatever way is useful (provided there is no impact on the component API), there's no reason not to opt for `when` here.

second-gen/packages/swc/components/badge/Badge.ts
graynorton: Yes, it looks like adding a rule to the component styles that sets `display: block;` on the slot also works.

I think in general it would be better to avoid systematically wrapping slots for styling purposes.

second-gen/packages/core/components/badge/Badge.base.ts
graynorton: I think it's better to define them here on the base class, since these slots are effectively part of the component architecture dictated by the base class (i.e., any concrete implementation building on this base class will need to include the slots), and the base class includes the mixin functionality associated with the slots.

second-gen/packages/core/components/badge/Badge.base.ts
graynorton: At a glance, I think you're right—at least, I can't immediately see why this property would need a bespoke getter and setter.

That said, the aim of the Barebones milestone is to avoid any changes not strictly related to migrating these components to the new architecture, so I think the best course of action here would be to add a `@todo` comment rather than making a change in this PR.

second-gen/packages/swc/components/badge/Badge.ts
graynorton: Yes, we definitely need to define a constant representing the full set of S2 colors. As I mentioned in an earlier comment, I've done some additional exploration into patterns for managing differences in S1 & S2 types, which I'll put in a separate PR (to land after this one).

second-gen/packages/swc/components/badge/Badge.ts
graynorton: Yes, as far as TypeScript is concerned, the type defined in this module is completely different from the type defined in the `core` module, though they share the same name, so it would in fact need to be exported here.

Again, though, I'll make a separate follow-on PR to propose some patterns for managing S1 and S2 types.

second-gen/packages/swc/components/badge/Badge.ts
graynorton: I think, since we're not going to be releasing any S2 components for a while (even in preview form), we might as well take the conservative approach and not include any real Figma links for the time being.

graynorton (approved):
@castastrophe Thank you for blazing the trail here and getting the S2 Badge rendering and styling migrated!

The patterns you explored around the nuances of the factoring between the abstract base components in `core` and the concrete rendering components in 1st- and 2nd-gen SWC, and for managing type differences between the S1 and S2 components, were also very informative. 🙏

As I mentioned in a few specific review comments, what we'd like to do is go ahead and merge this PR into the `barebones` branch (along with your work on ProgressCircle), after which I'll make a follow-on PR proposing some general patterns for use on the remaining Barebones components.

I've gone ahead and approved this PR; no need to make any further changes on this branch as far as I'm concerned.

PR #5730, opened 2025-09-15
fix(pending-state): correct reflection of aria in pending controller

caseyisonit:
## Description

This PR fixes aria-label handling issues in button components and improves accessibility for pending state controllers. The changes ensure that aria-labels are properly managed and preserved across component state changes, providing better screen reader support.

### Key Changes:
**Button Component (@spectrum-web-components/button):**

- Fixed timing of aria-label updates to occur after slot content changes are processed, preventing race conditions
- Added label property support for programmatic aria-label control in Storybook
- Added comprehensive tests for aria-label behavior during content changes
- Moved aria-label management logic to execute after slot content processing

**PendingState Controller (@spectrum-web-components/reactive-controllers):**

- Improved aria-label caching logic to better handle dynamic label changes
- Changed progress circle from aria-valuetext to aria-label for better accessibility compliance
- Enhanced caching mechanism to preserve user-set aria-labels during pending states
- Added more robust logic for determining when to cache aria-labels

## Motivation and context

The previous implementation had timing issues, allowing aria-label updates to occur before slot content changes were fully processed, leading to inconsistent accessibility behavior. Additionally, the pending state controller was using `aria-valuetext` instead of `aria-label` for progress indicators, which is not the recommended approach for accessibility.

These changes ensure that:

- Screen readers receive consistent and accurate aria-label information
- User-set aria-labels are properly preserved during component state changes
- Progress indicators follow accessibility best practices
- Dynamic content changes don't interfere with aria-label management

## Related issue(s)

-   fixes [5620]
-   resolves SWC-1039

---

## Author's checklist

---

## Reviewer's checklist

### Manual review test cases

- [ ] Button aria-label timing test

1. Go to the button Storybook stories
2. Set a label property on a button with slot content
3. Change the slot content dynamically
4. Verify that the aria-label remains consistent and doesn't get overridden

- [ ] Pending state aria-label preservation test

1. Go to any component that uses the PendingState controller (like buttons with pending states)
2. Set a custom aria-label on the component
3. Trigger the pending state
4. Verify that the custom aria-label is cached and restored when pending state ends

- [ ] Progress circle accessibility test

1. Go to components with pending states
2. Use a screen reader to navigate to the progress indicator
3. Verify that the progress circle announces properly with aria-label instead of aria-valuetext

- [ ] Dynamic content changes test

1. Create a button with both slot content and a label property
2. Dynamically change the slot content multiple times
3. Verify that the aria-label from the label property is preserved throughout

### Device review

### DIFF
--- a/packages/progress-circle/stories/progress-circle.stories.ts
+++ b/packages/progress-circle/stories/progress-circle.stories.ts
@@ -110,7 +110,7 @@ export const inButton = ({
             margin-block: 0;
         }
     </style>
-    <sp-button static-color="black">
+    <sp-button variant="black" style="color: white">
         <sp-progress-circle
             progress="53"
             static-color="white"


tools/reactive-controllers/src/PendingState.ts
```diff
@@ -73,17 +73,27 @@ export class PendingStateController<T extends HostWithPendingState>
...
             this.host.setAttribute('aria-label', pendingLabel || 'Pending');
-        } else if (!pending || disabled) {
+        } else {
             // Restore the cached `aria-label` if it exists
             if (this.cachedAriaLabel) {
                 this.host.setAttribute('aria-label', this.cachedAriaLabel);
-            } else if (!pending) {
-                // If no cached `aria-label` and not `pending`, remove the `aria-label`
+            } else {
                 this.host.removeAttribute('aria-label');
```
Rajdeepc, line 112: You are now always removing `aria-label` If there is no cached label. Do you intend this to be removed even if it was never set?  Maybe check for if this exists first?

tools/reactive-controllers/src/PendingState.ts
```diff
@@ -73,17 +73,27 @@ export class PendingStateController<T extends HostWithPendingState>
         const { pending, disabled, pendingLabel } = this.host;
         const currentAriaLabel = this.host.getAttribute('aria-label');
 
-        if (pending && !disabled && currentAriaLabel !== pendingLabel) {
-            // Cache the current `aria-label` to be restored when no longer `pending`
+        // If the current `aria-label` is different from the pending label, cache it
+        // or if the cached `aria-label` is different from the current `aria-label`, cache it
+        if (
+            (!this.cachedAriaLabel &&
```
Rajdeepc: If this.cachedAriaLabel is undefined, isn't the second branch (this.cachedAriaLabel !== currentAriaLabel) always true? Also this is quite complex to read. Can you abstract these into a helper function, added a small snippet below if you want to take reference. Just making sure the main method is more readable
```ts
if (shouldCacheAriaLabel(this.cachedAriaLabel, currentAriaLabel, pendingLabel)) {
    this.cachedAriaLabel = currentAriaLabel;
}
...
function shouldCacheAriaLabel(cached, current, pending) {
    return (
        (!cached && current && current !== pending) ||
        (cached !== current && current && current !== pending)
    );
}
```

Rajdeepc (changes requested):

tools/reactive-controllers/src/PendingState.ts
caseyisonit: this is for when an element is using its text content as the aria label. this is removing it from the host, the host will not always have a set aria label so we would be honoring the initial render.

tools/reactive-controllers/src/PendingState.ts
caseyisonit: the first half is checking if there is a cached label at all:
1. Save > Pending > Save

the second half is checking if the button label changes after pending that the cached label gets updated with the new label:

2. Save > Pending > Continue

This keeps everything in sync if the button goes to pending again. I believe the second half is actually solving the consumer needs of the issue filed.

I like the suggestion for the helper function, i will add that in.

.changeset/seven-nails-wave.md
```diff
@@ -0,0 +1,20 @@
+---
+'@spectrum-web-components/button': patch
+'@spectrum-web-components/reactive-controllers': patch
```
Rajdeepc: nit: Will this be read with our current changelog script? Would you prefer adding separate changeset for each package instead?

packages/button/stories/template.ts
```diff
@@ -9,7 +9,7 @@
  * OF ANY KIND, either express or implied. See the License for the specific language
  * governing permissions and limitations under the License.
  */
-
+/* eslint-disable import/no-extraneous-dependencies */
```
rubencarvalho: Is this needed?

packages/button/stories/index.ts
```diff
@@ -9,6 +9,7 @@
  * OF ANY KIND, either express or implied. See the License for the specific language
  * governing permissions and limitations under the License.
  */
+/* eslint-disable import/no-extraneous-dependencies */
```
rubencarvalho: Same as below 😛 Is this needed?

tools/reactive-controllers/src/PendingState.ts
```diff
@@ -58,6 +58,7 @@ export class PendingStateController<T extends HostWithPendingState>
                       id="loader"
                       size="s"
                       indeterminate
+                      aria-label=${pendingLabel}
```
rubencarvalho: Blocking: This change adds an `aria-label` to the progress-circle. Previously, only the host element had an `aria-label`. When the component is in `pending` state, this results in two labels being announced, causing the screen reader to read the label twice. Can you confirm?

rubencarvalho (changes requested):
Nice work! I left one blocking comment: could we confirm the double label behavior?

tools/reactive-controllers/src/PendingState.ts
caseyisonit: I will confirm. This resolves the incorrect implementation causing the dev warning on every button

rubencarvalho: Screenshot 2025-09-22 at 15 59 59

I couldn’t get VO to announce it twice, but the accessibility tree does show it. I’m not sure how NVDA or other screen readers would handle this.

caseyisonit: I just ran this and it only reads pending once from the host

caseyisonit: filing a github issue to resolve the aria and dev warning issues with pending state as a separate scope

.changeset/seven-nails-wave.md
caseyisonit: there are other changesets with mulitple packages, i dont think it should be an issue

rubencarvalho: I think what Rajdeep means is, even though this works, our change log script does not pick up multiple packages (only the first one).

tools/reactive-controllers/src/PendingState.ts
```diff
@@ -73,17 +75,37 @@ export class PendingStateController<T extends HostWithPendingState>
         const { pending, disabled, pendingLabel } = this.host;
         const currentAriaLabel = this.host.getAttribute('aria-label');
 
-        if (pending && !disabled && currentAriaLabel !== pendingLabel) {
-            // Cache the current `aria-label` to be restored when no longer `pending`
+        function shouldCacheAriaLabel(
+            cached: string | null,
+            current: string | null,
+            pending: string | undefined
+        ): string | boolean | null {
```
rubencarvalho: Are the possible return values accurate? `string | boolean | null`?

tools/reactive-controllers/src/PendingState.ts
```diff
@@ -73,17 +75,37 @@ export class PendingStateController<T extends HostWithPendingState>
...
+            return (
+                (!cached && current && current !== pending) ||
```
rubencarvalho: `current && ...` treats `""` as `false`. Is an empty string invalid? Or do we need to treat `null`/`undefined` differently from an empty string?

rubencarvalho: If an empty string, maybe we can add a test to cover that?

tools/reactive-controllers/src/PendingState.ts
```diff
@@ -47,8 +47,10 @@ export class PendingStateController<T extends HostWithPendingState>
 
     public cachedAriaLabel: string | null = null;
     /**
-     * Renders the pending state UI.
+     * Renders the pending state UI. The aria-valuetext is needed for Firefox
      * @returns A TemplateResult representing the pending state UI.
+     *
+     * @TODO: [SWC-1255] This should now be using the progress-circle component. It should be using an animated SVG icon with correct role and aria.
```
rubencarvalho: nit: 
```suggestion
     * @TODO: [SWC-1255] This should not be using the progress-circle component. It should be using an animated SVG icon with correct role and aria.
```

(also, this todo is very prescriptive of the solution, even though I don't think we have complete clarity on the path forward her)

tools/reactive-controllers/src/PendingState.ts
```diff
@@ -47,8 +47,10 @@ export class PendingStateController<T extends HostWithPendingState>
 
     public cachedAriaLabel: string | null = null;
     /**
-     * Renders the pending state UI.
+     * Renders the pending state UI. The aria-valuetext is needed for Firefox
```
rubencarvalho: > The aria-valuetext is needed for Firefox

I'm curious, why?

tools/reactive-controllers/src/PendingState.ts
caseyisonit: i can change the language

tools/reactive-controllers/src/PendingState.ts
caseyisonit: because firefox reads indeterminate as 50 and value text will override that for screen readers, again this is part of the accessibility issue with using pending-circle

tools/reactive-controllers/src/PendingState.ts
rubencarvalho: current !== "" && ...

tools/reactive-controllers/src/PendingState.ts
rubencarvalho: role = "" 
aria-hidden="true"

tools/reactive-controllers/src/PendingState.ts
caseyisonit: got it to return boolean only

tools/reactive-controllers/src/PendingState.ts
caseyisonit: i removed the check of 'current' being present and added parentheses to aid in readability. this also allowed the return type to be boolean.

tools/reactive-controllers/src/PendingState.ts
caseyisonit: moving forward with `role="presentation"` as that is more semantically correct for the usage.

packages/progress-circle/src/ProgressCircle.ts
```diff
@@ -119,10 +119,11 @@ export class ProgressCircle extends SizedMixin(SpectrumElement, {
 
         if (window.__swc.DEBUG) {
             if (
-                !this.label &&
-                !this.getAttribute('aria-label') &&
-                !this.getAttribute('aria-labelledby') &&
-                !this.slotEl.assignedNodes().length
+                (!this.label ||
```
rubencarvalho: I needed to re-read this a couple of times 😛 If I got it right, we are saying "if this element is of role progressbar but it has no text label, no aria-label, no aria-labelledby, and no slotted label content, then we throw the warning".

What do you think of making it a bit more clear (pseudo-code):

```ts
hasAccessibleName() {
    return Boolean(
        this.label ||
        this.getAttribute('aria-label') ||
        this.getAttribute('aria-labelledby') ||
        this.slotEl.assignedNodes().length
    );
}
```

and then

```ts
if (!this.hasAccessibleName() && this.getAttribute('role') === 'progressbar') {
```

rubencarvalho (approved):
Nice! 👍 I think this is the best path forward! I just left one 1 small nit comment.

.changeset/seven-nails-wave.md
```diff
@@ -0,0 +1,19 @@
...
+---
+
+**Fixed** aria-label handling in button components and pending state controller.
+
+**Button component changes:**
+
+- Fixed timing of aria-label updates to occur after slot content changes are processed
+- Added `label` property support for programmatic aria-label control
```
rubencarvalho: Is this accurate?

rubencarvalho:
Please check why the stories are loading differently from prod: https://opensource.adobe.com/spectrum-web-components/storybook/iframe.html?globals=system%3Aspectrum-two&args=&id=picker-pending--s&viewMode=story

PR: https://swcpreviews.z13.web.core.windows.net/pr-5730/docs/storybook/iframe.html?globals=system%3Aspectrum-two&args=&id=picker-pending--s&viewMode=story

In the PR link I can't see the pending state when opening the story in a new tab.

rubencarvalho:
Also, unrelated to your changes, but when we have a pending picker there is no VO indication that you can't open it (but it is effectively "disabled")

.changeset/seven-nails-wave.md
caseyisonit: no thank you for checking, i need to update this with the new changes

nikkimk (changes requested):
When I tested with voice over on a button with a slotted label and an `aria-label` only the `aria-label` is read. I don't think what is read should differ from what is on screen. This fixes the initial problem, but there are some a11y pitfalls with allowing this. 

Can we make sure the docs recommends not using `label` or `aria-label` if they are supplying visible text via the slot?

packages/progress-circle/src/ProgressCircle.ts
caseyisonit: sure we can do that

rubencarvalho:
> When I tested with voice over on a button with a slotted label and an `aria-label` only the `aria-label` is read. I don't think what is read should differ from what is on screen. This fixes the initial problem, but there are some a11y pitfalls with allowing this.
> 
> Can we make sure the docs recommends not using `label` or `aria-label` if they are supplying visible text via the slot?

This is a good call! But as per spec, `aria-label` _does_ take precedence over slotted text, so what you saw is expected... I guess we can try and rely on our docs to steer consumers away from combining a slot label with aria-label (maybe only use it for icon-only cases? or for cases where multiple buttons in a page have the same label and you need to provide SR users a bit more context to differentiate the buttons). And a dev-time warning could help too!

But I agree this is out-of-scope for this PR and better tackled on the Button refactor! 😃

nikkimk (approved):

rubencarvalho (approved):
:feels_good:

PR #5733, opened 2025-09-16
fix(sp-picker): s2 theme update - no borders for picker

iuliauta:
## Description

Picker border color should be hidden in S2 theme

## Motivation and context

Fix the regressions found after changing from the express theme to the spectrum-two theme.

## Related issue(s)

-   fixes SWC-1074
- CCEX-230904

Spectrum 2
Screenshot 2025-09-16 at 16 16 53

Spectrum 1
Screenshot 2025-09-16 at 16 17 13

Express
Screenshot 2025-09-16 at 16 17 30

---

## Author's checklist

---

## Reviewer's checklist

### Manual review test cases

    1. Go [picker component](https://swcpreviews.z13.web.core.windows.net/pr-5733/docs/components/picker/)
    2. Select Spectrum 2 theme.
    3. Observe that the picker component do not have any borders anymore (the same as the Express theme).

### Device review


.circleci/config.yml
```diff
@@ -22,7 +22,7 @@ parameters:
     # 3. Commit this change to the PR branch where the changes exist.
     current_golden_images_hash:
         type: string
-        default: 5cb7586aae24a30d566d452d4e0b3c934fe3b9f0
+        default: 545766d1fd28704c1920e13b3fb8bf24f9c7dbba
```
Rajdeepc: We usually update the hash once the PR is approved to make sure we are not overriding any unintended visual changes to `main`

iuliauta: Did not know, sorry! I reverted it right now, thank you!

tools/styles/tokens-v2/system-theme-bridge.css
```diff
@@ -338,14 +338,14 @@
...
-    --system-picker-border-color-disabled: var(--spectrum-gray-300);
-    --system-picker-border-width: var(--spectrum-border-width-100);
+    --system-picker-border-color-default: transparent;
+    --system-picker-border-color-default-open: transparent;
+    --system-picker-border-color-hover: transparent;
+    --system-picker-border-color-hover-open: transparent;
+    --system-picker-border-color-active: transparent;
+    --system-picker-border-color-key-focus: transparent;
+    --system-picker-border-color-disabled: transparent;
+    --system-picker-border-width: 0px;
```
marissahuysentruyt: I think this border is going to need to stay border-width-100. From a CSS standpoint, we're hooking into this border width specifically in high contrast mode/`forced-colors: active`. 

Screenshot 2025-09-17 at 9 50 56 AM
Screenshot 2025-09-17 at 9 50 33 AM

So it remains transparent in a non-high contrast setting, but then we define a `--highcontrast-picker-border-color-default` color.

iuliauta: Thank you for pointing it out, I was not aware of its need for the high contrast case! I reverted the border-width change.

tools/styles/tokens-v2/system-theme-bridge.css
```diff
@@ -345,7 +345,7 @@
     --system-picker-border-color-active: transparent;
     --system-picker-border-color-key-focus: transparent;
     --system-picker-border-color-disabled: transparent;
-    --system-picker-border-width: 0px;
+    --system-picker-border-width: var(--spectrum-border-width-100);
```
rise-erpelding: I was also noticing that the label text was positioned further up within the picker before, so reverting back to `border-width-100` also solves that issue!

But I'm wondering if this border-width is desired for the invalid variant, or do we want to adjust maybe to `border-width-200`?
image

marissahuysentruyt: @iuliauta definitely knows best, but the border-width-200 is in the specs for picker in S2. This suggestion makes sense to me!

iuliauta: Yes, it seems that I should use `--system-picker-border-width: var(--spectrum-border-width-200);` for s2. I will update it, thank you for all the details, I was not aware of the label positioning, but it is good information for me.

rise-erpelding (approved):
Thanks @iuliauta, this looks great!

One thing I want to call out that I think is out of scope for this work, is that there's a diff in the VRTs for the position of the text (just for L, it looks like), where there's more spacing above the text where we increased the border width, so it sits lower within the Picker.

image

I'm wondering if part of the issue is that we're not calculating the border-width out of the spacing like we do with [button](https://github.com/adobe/spectrum-web-components/blob/main/packages/button/src/spectrum-button.css#L76), for instance doing something like:

```
--spectrum-picker-spacing-edge-to-text: calc(var(--spectrum-component-edge-to-text-100) - var(--spectrum-picker-border-width));
```

This would help keep the spacing consistent even if the border width changes.

All of that said though, since the fix would need to happen outside of the theme files, I think it's out of scope, but I wanted to explain what I think might be happening here.

iuliauta:
> * Spectrum | Light | Medium | LTR



> Thanks @iuliauta, this looks great!
> 
> One thing I want to call out that I think is out of scope for this work, is that there's a diff in the VRTs for the position of the text (just for L, it looks like), where there's more spacing above the text where we increased the border width, so it sits lower within the Picker.
> 
> image
> I'm wondering if part of the issue is that we're not calculating the border-width out of the spacing like we do with [button](https://github.com/adobe/spectrum-web-components/blob/main/packages/button/src/spectrum-button.css#L76), for instance doing something like:
> 
> ```
> --spectrum-picker-spacing-edge-to-text: calc(var(--spectrum-component-edge-to-text-100) - var(--spectrum-picker-border-width));
> ```
> 
> This would help keep the spacing consistent even if the border width changes.
> 
> All of that said though, since the fix would need to happen outside of the theme files, I think it's out of scope, but I wanted to explain what I think might be happening here.

Thank you for pointing it out! I tried to see how this change will influence the themes and it seems that it impacts spectrum-1 theme and I suggest to address this improvement in another PR, unrelated to CSS themes PRs (as you already mentioned in your message).

castastrophe (approved):

PR #5767, opened 2025-09-29
fix(picker): add picker attribute to nested popover and styles to constrain to picker width

cdransf:
## Description

This PR adds styles to the `popover.css` to enforce a fixed width when the popover is used within the picker component. This aligns with the design documentation and the picker component styles.

## Motivation and context

Without this change the rendered picker options will grow in width without wrapping for longer labeld.

## Related issue(s)

-   fixes SWC-562

## Author's checklist

---

### Manual review test cases

1. Navigate to the [picker docs](https://swcpreviews.z13.web.core.windows.net/pr-5767/docs/components/picker/)
2. Open the picker
3. Edit one of the options in the inspector, changing the label text to something much longer than the default.
4. See that the label wraps rather than causing the dropdown to grow horizontally.

### Device review

### DIFF
--- a/packages/picker/src/picker.css
+++ b/packages/picker/src/picker.css
@@ -14,6 +14,8 @@
 @import url("./picker-overrides.css");
 
 :host {
+    --mod-popover-inline-size: var(--mod-picker-inline-size, var(--spectrum-picker-inline-size));
+
     display: inline-flex;
     vertical-align: top;
 
--- a/packages/picker/stories/picker.stories.ts
+++ b/packages/picker/stories/picker.stories.ts
@@ -431,6 +431,7 @@ export const iconValue = (args: StoryArgs): TemplateResult => {
             label="Pick an action"
             icons="only"
             value="2"
+            style="--mod-picker-inline-size: 100px;"
         >
             <sp-menu-item value="1">
                 <sp-icon-edit slot="icon"></sp-icon-edit>
@@ -459,6 +460,7 @@ export const iconsOnly = (args: StoryArgs): TemplateResult => {
             @change=${handleChange(args)}
             label="Pick an action"
             value="3"
+            style="--mod-picker-inline-size: 100px;"
         >
             <sp-menu-item value="1">
                 <sp-icon-edit slot="icon" label="Edit"></sp-icon-edit>
--- a/packages/picker/stories/picker.stories.ts
+++ b/packages/picker/stories/picker.stories.ts
@@ -430,7 +430,6 @@ export const iconValue = (args: StoryArgs): TemplateResult => {
             @change=${handleChange(args)}
             label="Pick an action"
             icons="only"
-            style="width: 100px"
             value="2"
         >
             <sp-menu-item value="1">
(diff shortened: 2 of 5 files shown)


packages/popover/src/popover.css
```diff
@@ -78,3 +78,7 @@
 :host([tip][placement]) #tip {
     height: auto;
 }
+
+:host([picker]) {
```
castastrophe: I'm wondering if there's a CSS-only way to approach this without adding an attribute. Do you think we could add a modifier to Popover at the host level?

This is what we have in popover.css now:
```
:host {
    min-width: min-content;
    max-height: 100%;
    max-width: 100%;
    clip-path: none;
}
```

Maybe we create a modifier for width and then in the picker.css file, we can override that and have it cascade down into the popover?

packages/picker/src/picker.css
```diff
@@ -14,6 +14,8 @@
 @import url("./picker-overrides.css");
 
 :host {
+    --mod-popover-width: var(--mod-picker-inline-size, var(--spectrum-picker-inline-size));
```
castastrophe: Love the naming here - should the new modifier use inline size or do you think we should keep it as width?

cdransf: Good callout! I changed it to `--mod-popover-inline-size` so that it's more consistent with the picker custom properties and the `inline-size` property set in `popover.css`. ✨

rubencarvalho:
Changes look good to me, however I'm wondering if [this](https://swcpreviews.z13.web.core.windows.net/pr-5767/spectrum-light-medium-ltr/review/#PickerStories/iconsOnly.png) is expected / desired?

packages/picker/src/picker.css
```diff
@@ -14,6 +14,8 @@
 @import url("./picker-overrides.css");
 
 :host {
+    --mod-popover-inline-size: var(--mod-picker-inline-size, var(--spectrum-picker-inline-size));
```
rubencarvalho, line 17: Just a quick question: what’s the reasoning behind prefixing this variable with `--mod`? Couldn’t it have been kept as a “non-mod” / internal-only variable in this case?

cdransf: I'd be ok to change it provided there's a preferred/existing pattern so that we're not introducing anything new. 🤔

castastrophe: @rubencarvalho Mods are how we handle "passthroughs" in S1 CSS. It's the only way to ensure an empty hook. Spectrum prefixed properties are not guaranteed to be empty and might require more specificity to override.

packages/picker/stories/picker.stories.ts
```diff
@@ -430,8 +430,8 @@ export const iconValue = (args: StoryArgs): TemplateResult => {
             @change=${handleChange(args)}
             label="Pick an action"
             icons="only"
-            style="width: 100px"
             value="2"
+            style="--mod-picker-inline-size: 100px;"
```
castastrophe, line 434: Nice!

packages/picker/src/picker.css
cdransf: > I'd be ok to change it provided there's a preferred/existing pattern so that we're not introducing anything new. 🤔

I'll stick with the mod scheme for now as that's the established pattern for S1 and we'll have new options in S2.

rubencarvalho (approved):
:shipit:

PR #5785, opened 2025-10-06
feat: introduce dual-generation monorepo architecture

rubencarvalho:
## Description

This PR restructures the Spectrum Web Components repository into a dual-generation monorepo to support side-by-side development of 1st-gen and 2nd-gen components. This architectural change is foundational to the project's strategic objectives and enables the team to pursue multiple goals in parallel without disruption.

### Key changes:

- **Moved existing codebase** into `1st-gen/` workspace, preserving all existing functionality
- **Created `2nd-gen/` workspace** with initial package structure for `@spectrum-web-components/core` and `@adobe/swc`
- **Established shared Core library** (`@spectrum-web-components/core`) containing abstract, non-rendering base classes that both generations can extend
- **Updated root configuration** for Yarn 4 workspace management, delegating build/test/lint scripts to generation-specific workspaces
- **Refactored CI/CD** (CircleCI and GitHub Actions) to support the new workspace structure
- **Added comprehensive contributor documentation** covering architecture, migration paths, and development workflows

## Motivation and context

This change implements the "side-by-side development" strategy outlined in our project objectives, providing two critical advantages:

**Isolation:** 1st-gen and 2nd-gen workspaces are separated, allowing 2nd-gen to be built iteratively from the ground up without risking breakage to 1st-gen. Disruptive changes (tooling overhauls, Spectrum CSS unification, full-fidelity Spectrum 2 implementation) are confined to `2nd-gen/`, while 1st-gen continues operating "as-is".

**Colocation:** Both generations live in the same repository and share core component logic through abstract base classes in `@spectrum-web-components/core`. This means:
- Non-disruptive improvements (accessibility, bug fixes, features) can be shared immediately across both generations
- Customers benefit continuously throughout the transition
- Migration from 1st-gen to 2nd-gen will be smoother due to shared API and behavior

This architectural foundation enables the team to work efficiently on multiple strategic objectives in parallel:
- Unifying Spectrum CSS & SWC
- Building clean tooling and infrastructure
- Delivering full-fidelity Spectrum 2
- Improving accessibility
- Continually enhancing components

See `CONTRIBUTOR-DOCS/03_project-planning/01_objectives-and-strategy.md` for detailed strategic context.

## Related issue(s)
- Part of the broader 2nd-gen SWC initiative

-   fixes [Issue Number]

## Screenshots (if appropriate)

---

## Author's checklist

[...]

PR #5798, opened 2025-10-10
feat(divider): migrate to second-gen styles and architecture

rise-erpelding:
## Description

This PR migrates the next-gen Divider component to Spectrum 2 styles and updates Storybook accordingly.

It also introduces a new Storybook decorator for handling static color backgrounds.

## Motivation and context

This PR migrates the Divider component from first-gen to the second-gen architecture, following the established pattern of separating core component logic from Spectrum-specific implementation.

**Changes:**
- Migrated CSS to align with Spectrum CSS structure
- Added type definitions for size and static color variants
- Created comprehensive Storybook stories with new static color decorator

### What changed

**First-gen:**
- `first-gen/packages/divider/src/Divider.ts` - Remains unchanged for backward compatibility

**Core (Base):**
- `second-gen/packages/core/components/divider/Divider.base.ts` - Updates shared functionalities
- `second-gen/packages/core/components/divider/Divider.types.ts` - Creates types file for size and static color variants
- `second-gen/packages/core/components/divider/index.ts` - Imports and exports types

**Second-gen:**
- `second-gen/packages/swc/components/divider/Divider.ts` - Updates render method to accommodate classes
- `second-gen/packages/swc/components/divider/divider.css` - Adds CSS from `spectrum-css`
- `second-gen/packages/swc/components/divider/stories/divider.stories.ts` - Adds comprehensive story coverage
- `second-gen/packages/swc/.storybook/decorators/static-color-background.ts` - Creates decorator for static color backgrounds
- `second-gen/packages/swc/.storybook/preview.ts` - Registers the static color decorator

## Related issue(s)

-   SWC-1259

image

---

## Author's checklist

---

## Reviewer's checklist

-   [ ] Includes thoughtfully written changeset if changes suggested include `patch`, `minor`, or `major` features (**Note**: I did not add a changeset, I wasn't sure that this was set up for barebones and didn't see any for progress circle or badge, but I'm happy to add one if I'm mistaken!)

### Manual review test cases

**Run second-gen Storybook:**
```bash
yarn workspace @swc/components storybook
```

**Storybook functionality:**
- [x] All controls work correctly on Default story
- [x] Static color control applies appropriate backgrounds via decorator
- [x] Size control accepts string values (`s`, `m`, `l`) to change size
- [x] Vertical control toggles orientation correctly

**Documentation:**
- [x] All stories render correctly on Docs page
- [x] Story descriptions are clear and informative
- [x] Examples demonstrate all component variants (sizes, orientations, static colors) - see [component analysis docs](https://github.com/adobe/spectrum-web-components/blob/2nd-gen-component-analysis/migration-roadmap/divider.md) for a clearer picture of what changes for Spectrum 2

**Static color decorator:**
- [x] Works in Default story when using controls to set `static-color` (as seen in Divider and Progress Circle)
- [x] Works in StaticBlack and StaticWhite stories on Docs page (applies to Divider only), and only on those stories

**First-gen verification (use the PR preview or run locally):**
```bash
yarn start
# or
yarn docs:start
```
- [x] First-gen Divider remains unchanged and functional (VRT failures should be the same as on #5785)

**Code review checklist:**
- [x] TypeScript types are correct and properly exported
- [x] CSS follows second-gen patterns
- [x] No linter errors
- [x] Proper separation of concerns between core and SWC packages

### Device review

### DIFF
--- a/second-gen/packages/swc/components/divider/stories/divider.stories.ts
+++ b/second-gen/packages/swc/components/divider/stories/divider.stories.ts
@@ -76,7 +76,7 @@ export const Default: Story = {
 };
 
 /**
- * The small divider is the default size. This is used to divide similar components such as table rows, action button groups, and components within a panel.
+ * The small divider is used to divide similar components such as table rows, action button groups, and components within a panel.
  *
  * The medium divider is used for dividing subsections on a page, or to separate different groupings of components such as panels, rails, etc.
  *


second-gen/packages/swc/.storybook/decorators/static-color-background.ts
rise-erpelding, line 1: Keeping this separate so it can be rolled back easily if we don't want to add it at this time. I noticed that the static color controls on progress circle didn't apply a background making the static white invisible, so this would address that as well.

second-gen/packages/swc/components/divider/stories/divider.stories.ts
```diff
@@ -10,26 +10,173 @@
...
+/*
+ * @todo This is properly configuring the Select, but the control doesn't
+ * seem to work; need to investigate.
+ */
+
+// argTypes.size = {
+//     ...argTypes.size,
+//     control: { type: 'select' },
+//     options: Divider.VALID_SIZES,
+// };
```
rise-erpelding, line 36: This is the same situation as with Badge and Progress Circle, copying this over from those components.

second-gen/packages/swc/components/divider/divider.css
```diff
@@ -107,16 +78,33 @@ hr {
...
+    );
+    margin-block: var(--mod-divider-vertical-margin, 0);
     block-size: var(--mod-divider-vertical-height, 100%);
-    margin-block: var(--mod-divider-vertical-margin);
-    align-self: var(--mod-divider-vertical-align);
+    align-self: var(--mod-divider-vertical-align, flex-start);
+}
+
+/* windows high contrast mode */
+@media (forced-colors: active) {
```
rise-erpelding, line 84: If you test WHCM or forced-colors (even adjusting the Rendering in Chrome will reveal this), you'll see the divider on the docs page looks something like this:
image

I did investigate this, and it looks like it's coming from Storybook styling that's adding an 8px transparent border that shows up in forced colors. We might want to remove that eventually, but that feels a little out of scope for now. And the border isn't visible when you view the default story.

second-gen/packages/core/components/divider/Divider.base.ts
```diff
@@ -15,13 +15,20 @@ import { property } from 'lit/decorators.js';
...
+
 /**
  * @element swc-divider
  */
 export abstract class DividerBase extends SizedMixin(SpectrumElement, {
-    validSizes: ['s', 'm', 'l'],
+    validSizes: DIVIDER_VALID_SIZES,
     noDefaultSize: true,
 }) {
+    static readonly STATIC_COLORS = DIVIDER_STATIC_COLORS;
```
rubencarvalho: We’re still missing the contributing documentation that will clarify these nuances (SWC-1234). However, in this specific case, we don’t need to add this static variable. It is only needed to handle cases where these values differ from S1/S2 and we want to throw dev warnings (e.g., deprecation, invalid value) at runtime. By making these values class properties, we can access them through the constructor and retrieve the correct value being loaded, whether it’s from the first or second generation. Hopefully, we can make this information clear enough in the aforementioned documentation! 😛

second-gen/packages/swc/.storybook/decorators/static-color-background.ts
rubencarvalho: This is great! I think we should definitely keep it for now. It is one of those things we can always revisit once we dive deeper into our Storybook setup.

rubencarvalho:
Also, that screenshot in the description is _:chefkiss:_

first-gen/packages/divider/package.json
```diff
@@ -58,6 +58,8 @@
     ],
     "types": "./src/index.d.ts",
     "dependencies": {
+        "@spectrum-web-components/base": "1.8.0",
+        "@spectrum-web-components/shared": "1.8.0",
```
rubencarvalho: I am not sure we need the `shared` dependency 🤔 

```suggestion
```

second-gen/packages/swc/.storybook/decorators/static-color-background.ts
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+import { html } from 'lit';
+import { makeDecorator } from '@storybook/preview-api';
+import type { DecoratorFunction } from '@storybook/types';
+
+/**
+ * Static color background settings - matching spectrum-css gradients
+ */
+const staticColorSettings = {
```
castastrophe, line 20: Thanks so much for adding this! After this merges, I could go back into progress circle and update it to use this. 👯

second-gen/packages/core/components/divider/Divider.types.ts
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import type { ElementSize } from '@swc/core/shared/base';
+
+export const DIVIDER_VALID_SIZES: ElementSize[] = ['s', 'm', 'l'] as const;
```
castastrophe, line 15: Nice job here!

second-gen/packages/swc/components/divider/Divider.ts
```diff
@@ -11,18 +11,42 @@
...
+    // ────────────────────
+    //     API OVERRIDES
+    // ────────────────────
+
     public static override styles: CSSResultArray = [styles];
 
     protected override render(): TemplateResult {
-        return html``;
+        return html`
+            <div
```
castastrophe, line 40: 😍

second-gen/packages/swc/components/divider/divider.css
```diff
@@ -107,16 +78,33 @@ hr {
             var(--spectrum-divider-background-color)
         )
     );
-    overflow: visible;
+
+    &:not(&.spectrum-Divider--vertical) {
```
castastrophe:
```suggestion
    &:not(.spectrum-Divider--vertical) {
```

I don't think the amper used inside the not selector is going to give the result you are looking for here.

second-gen/packages/core/components/divider/Divider.types.ts
```diff
@@ -0,0 +1,18 @@
...
+export const DIVIDER_STATIC_COLORS = ['white', 'black'] as const;
```
Rajdeepc, line 16: How do you feel separating the types and the runtime values in `.consts.ts` and `.types.ts` files?

second-gen/packages/swc/components/divider/divider.css
```diff
@@ -107,16 +78,33 @@ hr {
...
+.spectrum-Divider--vertical {
     inline-size: var(
         --mod-divider-thickness,
         var(--spectrum-divider-thickness)
     );
-    block-size: 100%;
+    min-block-size: var(
+        --mod-divider-block-minimum-size,
+        var(--spectrum-divider-block-minimum-size)
+    );
```
castastrophe:
```suggestion
    min-block-size: var(--spectrum-divider-block-minimum-size);
```

Would you mind going through here and cleaning up the mods? Since we're not supporting mods in S2, the idea was to remove them in Spectrum CSS before we bring over the code but we haven't had a chance to do that yet.

second-gen/packages/swc/components/divider/divider.css
castastrophe: Hmm, is this something we can override with forced specificity in the Storybook styles?

second-gen/packages/swc/components/divider/stories/divider.stories.ts
```diff
@@ -11,25 +11,144 @@
...
     component: 'swc-divider',
+    args,
+    argTypes,
+    render: (args) => template(args),
+    parameters: {
+        actions: {
+            handles: events,
+        },
+    },
+    tags: ['migrated'],
```
castastrophe, line 58: Yay!

second-gen/packages/swc/components/divider/Divider.ts
```diff
@@ -11,18 +11,42 @@
...
+                class=${classMap({
+                    ['spectrum-Divider']: true,
+                    [`spectrum-Divider--size${this.size?.toUpperCase()}`]:
+                        typeof this.size !== 'undefined',
```
Rajdeepc: You can also use `this.size != null` for both null and undefined.

second-gen/packages/swc/components/divider/divider.css
```diff
@@ -12,85 +12,56 @@
...
-    );
-    --spectrum-divider-background-color-static-black: var(
-        --spectrum-transparent-black-300
-    );
-}
-
-hr {
-    border: none;
-    margin: 0;
-}
```
Rajdeepc, line 28: Can you please double check this is not a breaking change?

rise-erpelding: I don't see this component rendering an `<hr>`, it renders an empty template for first-gen (and then our `div` wrapper with the classes on it for second-gen), so it doesn't seem like those styles were doing anything there, unless I'm missing something.

I will say that I see the [CSS implementation](https://spectrumcss.z13.web.core.windows.net/pr-2352/index.html?path=/story/components-divider--default) optionally using `<hr>` though (and while the styles do set `border: none`, the margin isn't corrected for the horizontal divider), so I'm not sure if SWC once used `<hr>` as well and this is remaining from that.

second-gen/packages/swc/components/divider/divider.css
rise-erpelding: Nice catch! I had copied over my CSS without cleaning it up (thus the not removing of the mods). Do you have a preference on nesting/no nesting, for instance with something like:

```css
.spectrum-Divider--staticWhite {
	--spectrum-divider-background-color: var(--spectrum-transparent-white-200);

	&.spectrum-Divider--sizeL {
		--spectrum-divider-background-color: var(--spectrum-transparent-white-800);
	}
}
```

It didn't look like badge and progress circle CSS were using it, so I will be removing it here as well (relying more on what's in the dist index.css), but let me know if that's too much!

second-gen/packages/swc/components/divider/divider.css
rise-erpelding: Was this what you meant? I added some styles to a new preview-head.html file in f0d76075c41d553366bc96108882082bb7c50628 but the class selector needs to be tripled in order to increase the specificity enough to override.

second-gen/packages/core/components/divider/Divider.types.ts
rise-erpelding: Done! 1c40f784a742ef717c783a87262e1c863f678358

second-gen/packages/swc/components/divider/stories/divider.stories.ts
marissahuysentruyt: Could it be related to us not defining the `size` property on the component? What I understand from Cursor is that we get the argTypes using getStorybookHelpers, but getStorybookHelpers doesn't bind the size attribute to the component in the "regular" `template(args)` render we're trying to use.

So because divider extends the `SizedMixin()`, but we're only pulling the argTypes from divider specifically, is `getStorybookHelpers` _not_ seeing the `size` property? 

While I was experimenting with this in status light, if I use a different render method that binds the `size` property:
```
render: (args) =>
        html`<swc-status-light .size=${args.size} variant=${args.variant}
            >${args['default-slot']}</swc-status-light
        >`,
```

I think I can get the commented out code to work as we're expecting. Not sure if that's something we want to do right now though.

graynorton: @marissahuysentruyt Thank you for digging into this! I think the solution you came up with is a good workaround, and one we may ultimately end up adopting, but I'd like to hold off for now so we can troubleshoot our Custom Element Manifest tooling pipeline and figure out why mixin-defined properties aren't propagating as we'd like them to.

second-gen/packages/core/components/divider/Divider.base.ts
graynorton: @rise-erpelding: I discussed this briefly offline with @rubencarvalho, and we agreed that for now we'd like to consistently apply the pattern where we hang structured Spectrum data-type arrays off of the component's constructor.

Let's do this even in cases where the data type doesn't vary from S1 to S2, and even in cases where we don't have a specific need for the structured types at runtime.

At minimum, doing this consistently means we should essentially never have to separately import the structured type arrays into a component's stories—we can always just import the constructor and access the type arrays from there.

The fundamental reasoning behind this is to minimize cognitive load for component authors and maximize consistency.

~~In the case of Divider, this would mean defining static readonly properties on the base component for both VALID_SIZES and STATIC_COLORS.~~

In the case of Divider, this would mean defining a static readonly property on the base component for `STATIC_COLORS`. (It's not necessary to define one for `VALID_SIZES`, since `SizedMixin` does that for you.)

Let me know if any of this isn't clear!

second-gen/packages/core/components/divider/Divider.types.ts
graynorton: @Rajdeepc @rise-erpelding While I'm definitely open to evolving in this direction as we work our way through more components and refine our approach to defining structured Spectrum data types, I'd prefer to keep it simple for now by maintaining just one file.

My reasoning here is that I could see us evolving in various directions from this initial starting point, so doing this refactor now seems premature.

second-gen/packages/core/components/divider/Divider.base.ts
```diff
@@ -32,7 +35,23 @@ export abstract class DividerBase extends SizedMixin(SpectrumElement, {
...
+            !(DIVIDER_STATIC_COLORS as readonly string[]).includes(value)
+        ) {
+            // Silently ignore invalid values
+            this._staticColor = undefined;
+            return;
+        }
+        this._staticColor = value;
+    }
+
+    private _staticColor?: DividerStaticColor;
```
graynorton: A couple of things:

1) As a matter of course, I'd prefer that we not include functional changes as part of the migration PRs; I'd rather that, when we come across a potential fix or improvement like this during migration, we add a `@todo` and come back to it in a separate PR.

2) When we do need to do runtime validation against structured Spectrum type data, I'd like to consistently access that data off of the component's constructor. Accessing it directly from module scope like this means that, if things were to change down the road and this particular data type started varying from one concrete implementation to another, this logic would not be accessing the correct data.

second-gen/packages/swc/components/divider/Divider.ts
```diff
@@ -11,18 +11,42 @@
...
+
 /**
- * @element sp-divider
+ * @element swc-divider
  */
 export class Divider extends DividerBase {
+    // ────────────────────
+    //     API OVERRIDES
+    // ────────────────────
+
```
graynorton, line 35: I think this should actually be `RENDERING & STYLING`, not `API OVERRIDES`.

second-gen/packages/swc/components/divider/stories/divider.stories.ts
```diff
@@ -11,25 +11,147 @@
...
 import { html } from 'lit';
-import type { Meta, StoryObj } from '@storybook/web-components';
+import type { Meta, StoryObj as Story } from '@storybook/web-components';
+import { getStorybookHelpers } from '@wc-toolkit/storybook-helpers';
+
+import { Divider } from '@swc/components/divider';
+import {
+    DIVIDER_STATIC_COLORS,
+    DIVIDER_VALID_SIZES,
+} from '@swc/core/components/divider';
```
graynorton: These are the imports we can avoid if we systematically hang the structured Spectrum data type arrays off of the component's constructor.

Side note: You can actually already avoid the import of `DIVIDER_VALID_SIZES`, because `SizedMixin` itself already hangs `VALID_SIZES` off of the constructor.

second-gen/packages/core/components/divider/Divider.base.ts
```diff
@@ -15,13 +15,20 @@ import { property } from 'lit/decorators.js';
...
+
 /**
  * @element swc-divider
  */
 export abstract class DividerBase extends SizedMixin(SpectrumElement, {
-    validSizes: ['s', 'm', 'l'],
+    validSizes: DIVIDER_VALID_SIZES,
     noDefaultSize: true,
 }) {
+    static readonly STATIC_COLORS = DIVIDER_STATIC_COLORS;
```
graynorton: Sorry, missed this earlier, but we should include a doc block and mark it as `@internal`, and explicitly type it to be `readonly string[]`, as in this example from Badge:

```
    /**
     * @internal
     */
    static readonly FIXED_VALUES: readonly string[] = FIXED_VALUES;
```

In cases like this, where it isn't being overridden, the explicit typing isn't strictly necessary, but I'd like to do it as a matter of consistency, as TypeScript does complain if we don't explicitly type and then override in a subclass.

Since it's late in the afternoon and you're on PTO starting tomorrow, I'm going to go ahead and approve the PR rather than request another change; we can always add a commit or edit after merging.

second-gen/packages/swc/components/divider/divider.css
castastrophe: Perfect!

PR #5800, opened 2025-10-10
feat(status-light)!: migrate status-light to s2 styles and second-gen architecture

marissahuysentruyt:
## Description

This PR migrates the status-light component to the second-generation architecture, creating a shared base class that both first-gen and second-gen implementations extend. This enables the component to work in both Spectrum 1 and Spectrum 2 contexts while sharing common logic.

### Key changes

- **Created shared base class** (`@swc/core/components/status-light/StatusLight.base.ts`) that contains common logic for both generations
- **Created shared types file** (`StatusLight.types.ts`) defining variants for S1 and S2:
  - S1 variants: semantic (neutral, info, positive, negative, notice, accent) + color (fuchsia, indigo, magenta, purple, seafoam, yellow, chartreuse, celery, cyan)
  - S2 variants: semantic (neutral, info, positive, negative, notice) + color (all S1 colors plus pink, turquoise, brown, cinnamon, silver)
- **Refactored first-gen** (`first-gen/packages/status-light/src/StatusLight.ts`) to extend the shared base class
  - Maintains S1-specific `disabled` property (deprecated in S2)
  - Uses S1-specific variant types
  - Added deprecation warnings for S1 types/exports
- **Implemented second-gen** (`second-gen/packages/swc/components/status-light/StatusLight.ts`) extending the shared base class
  - Uses S2-specific variant types
  - Removed `disabled` property (not supported in Spectrum 2)
  - Added S2 Spectrum CSS styles
- **Enhanced stories** with comprehensive examples for all variants and sizes
- **Fixed variant selectors** for accent and cyan variants
- **Added size support** with sizeMap decorator for consistent sizing across scales
- **Added tests** for second-gen status light and a test for deprecation warnings in first-gen

## Motivation and context

This migration is part of the broader effort to support Spectrum 2 design system while maintaining backward compatibility with Spectrum 1. By creating a shared base class, we reduce code duplication and ensure consistent behavior across both generations while allowing each to have generation-specific features and styling.

## Related issue(s)

- SWC-1260

**New stories for S2/second-gen:**
Screenshot 2025-10-13 at 2 32 45 PM
Screenshot 2025-10-13 at 2 32 38 PM
Screenshot 2025-10-13 at 2 32 17 PM

---

## Author's checklist

---

## Reviewer's checklist

### Manual review test cases

- Pull down the repo to run locally.

#### First-gen status-light functionality

1. Navigate to the first-gen status-light Storybook
2. Verify all S1 variants render correctly:
    - semantic: neutral, info, positive, negative, notice
    - color: fuchsia, indigo, magenta, purple, seafoam, yellow, chartreuse, celery
3. Verify the missing S2 variants have been added
    - semantic: accent
    - color: cyan
    _NOTE:_ these missing variants were an [identified gap during the component analysis](https://github.com/adobe/spectrum-web-components/blob/2nd-gen-component-analysis/migration-roadmap/status-light.md)
5. Verify `disabled` state still works correctly and properly adds or removes `aria-disabled`
6. There should be no VRT diffs for the status light examples that were originally supported in Storybook. The only VRT diffs would include the missing accent and cyan variants.

#### Second-gen status-light functionality

1. Navigate to the second-gen status-light Storybook
2. Verify all S2 variants render correctly:
    - semantic: neutral, info, positive, negative, notice (accent was deprecated)
    - color: fuchsia, indigo, magenta, purple, seafoam, yellow, chartreuse, celery, cyan, pink, turquoise, brown, cinnamon, silver (5 new non-semantic colors)
5. Expect visual appearance matches Spectrum 2 design specifications

#### Shared base class validation

1. Verify both implementations properly extend the shared base class
2. Expect no TypeScript errors and proper type narrowing for each generation
3. Verify variant validation warnings appear in dev mode for invalid variants
    - First-gen: Try adding an extra S2-only story to the S1 storybook. For example, add: `<sp-status-light size="s" variant="brown">brown</sp-status-light>`
    - Confirm that a browser warning appears in the console:

Screenshot 2025-10-13 at 2 24 48 PM

Second-gen: Try adding an extra S1-only story to the S2 storybook. For example, add: 

```
export const Disabled: Story = {
    render: () => html`
      <swc-status-light variant="accent" disabled>disabled</swc-status-light>`
};
```

Confirm that a browser warning appears in the console: 
  
Screenshot 2025-10-13 at 2 44 11 PM

Replace the variant value with something valid for S2, then reconfirm the new `disabled` console warning appears in the browser:

Screenshot 2025-10-13 at 2 28 27 PM

### Device review

- [ ] Did it pass in Desktop?

### DIFF
--- a/second-gen/packages/core/components/status-light/StatusLight.base.ts
+++ b/second-gen/packages/core/components/status-light/StatusLight.base.ts
@@ -111,13 +111,6 @@ export abstract class StatusLightBase extends SizedMixin(SpectrumElement, {
                         ],
                     }
                 );
-            } else if (this.hasAttribute('disabled') && 'disabled' in this) {
-                // Set aria-disabled when no warning is shown
-                if (!this.hasAttribute('aria-disabled')) {
-                    this.setAttribute('aria-disabled', 'true');
-                } else {
-                    this.removeAttribute('aria-disabled');
-                }
             }
         }
     }


first-gen/packages/status-light/src/StatusLight.ts
```diff
@@ -15,15 +15,82 @@ import {
...
+     */
+    @property({ type: String, reflect: true })
+    public override variant: StatusLightVariantS1 = 'info';
+
+    // ───────────────────────
+    //     API ADDITIONS
+    // ───────────────────────
+
+    /**
+     * @todo This property is deprecated in S2, and can be removed once we are no longer maintaining S1.
```
marissahuysentruyt: Whoops- should this be `@deprecated` instead?

first-gen/packages/status-light/src/spectrum-status-light.css
```diff
@@ -122,7 +124,7 @@
     background-color: var(--mod-statuslight-semantic-neutral-color, var(--spectrum-statuslight-semantic-neutral-color));
 }
 
-.spectrum-StatusLight--accent:before {
+:host([variant="accent"]):before {
```
marissahuysentruyt: The 2 updates in the first-gen CSS correspond to the gaps that were identified in the component docs. Is it still in scope for me to correct these missing variants? It obviously introduces VRT diffs, so I'm not sure where or if or when we want to introduce those diffs.

first-gen/packages/status-light/stories/status-light.stories.ts
```diff
@@ -19,6 +19,7 @@ export default {
 };
 
 export const s = (): TemplateResult => html`
+    <sp-status-light size="s" variant="accent">accent</sp-status-light>
```
marissahuysentruyt: Same question here- we found SWC was missing the accent and cyan variants that are defined for S1. Is it within the scope of this ticket to show those now for the first-gen stories?

first-gen/packages/status-light/test/status-light.test.ts
```diff
@@ -41,4 +42,41 @@ describe('Status Light', () => {
         expect(el.hasAttribute('aria-disabled')).to.be.true;
         expect(el.getAttribute('aria-disabled')).to.equal('true');
     });
+
+    describe('dev mode warnings', () => {
```
marissahuysentruyt, line 46: I had Cursor help me with this- do we want tests for the dev warnings at all? Feedback on this is appreciated either way!

second-gen/packages/swc/components/status-light/stories/status-light.stories.ts
```diff
@@ -10,88 +10,216 @@
...
+        'Changes the color of the status dot. The variant list includes both semantic and non-semantic options.',
+    type: { name: 'string', required: true },
+    table: {
+        type: { summary: 'string' },
+        defaultValue: { summary: 'info' },
+        category: 'Component',
+    },
+    control: { type: 'select' },
+    options: StatusLight.VARIANTS,
+};
```
marissahuysentruyt, line 31: Feel free to tell me to remove this and just follow what Badge is doing! This was the way I found that worked to populate more of the control table, but maybe that's not something we want to dive into right now. (and also this approach didn't really work with the default slot arg, sooo...it's not fool-proof 😆)

second-gen/packages/swc/components/status-light/stories/status-light.stories.ts
```diff
@@ -10,88 +10,216 @@
...
+                        >${sizeMap(size)}</swc-status-light
+                    >
+                `
+            )
+        ),
+    tags: ['!dev'],
+};
+
+// ────────────────────────
+//    HELPER FUNCTIONS
```
marissahuysentruyt, line 154: I pulled `capitalize()` and `CONTAINER()` from the other migrations, and then created a size map function for the labels in the size stories. This was totally personal preference, so I'm not offended if we don't like it and want to drop it.

Screenshot 2025-10-14 at 3 50 23 PM

second-gen/packages/swc/components/status-light/test/status-light.test.ts
marissahuysentruyt, line 1: Cursor pulled in the badge test patterns and created this status light test file. Is this something appropriate for the migration, or would you like me to separate this out into a different PR?

second-gen/packages/swc/components/status-light/stories/status-light.stories.ts
```diff
@@ -10,88 +10,216 @@
...
+ * @todo This is properly configuring the Select, but the control doesn't
+ * seem to work; need to investigate.
+ *
+ * We may have to explicitly bind the args to the component (particularly
+ * helpful for the size property) so the Storybook controls work as expected.
+ * 
+ * i.e. render: (args) =>
+        html`<swc-status-light .size=${args.size} variant=${args.variant}
+            >${args['default-slot']}</swc-status-light
+        >`,
```
marissahuysentruyt, line 43: If this comment is too much/overkill, let me know. I did document this idea in https://github.com/adobe/spectrum-web-components/pull/5798, as well.

second-gen/packages/swc/components/status-light/stories/status-light.stories.ts
```diff
@@ -10,88 +10,216 @@
...
+
+const meta: Meta = {
     title: 'Status light',
+    component: 'swc-status-light',
+    argTypes,
+    parameters: {},
+    args: {
+        ['default-slot']: 'Status light',
+        variant: 'info',
+        size: 'm',
```
marissahuysentruyt: Does this conflict with the `noDefaultSize` arg in `StatusLightBase` (for `SizedMixin`)? I don't have a good handle on the SizedMixin, but there is a default size defined in the design specs.

Maybe I need to do a Cursor dive into SizedMixin!

first-gen/packages/status-light/src/spectrum-status-light.css
graynorton: Interesting! Without digging too deeply, it looks to me like these were just cases where our (now-retired) automatic conversion of Spectrum CSS styles to SWC styles was failing—i.e., these were just bugs. Does that seem right to you?

As a matter of principle, my preference would be to split these fixes out into a separate PR, with a changeset describing the fixes.

first-gen/packages/status-light/src/StatusLight.ts
graynorton: Yes, I think we should just go ahead and deprecate this, with an appropriate customer-facing message, similar to the ones for the other deprecations in this file.

first-gen/packages/status-light/stories/status-light.stories.ts
graynorton: Yeah—per my earlier comment, I'd prefer that this and the corresponding fixes be pulled out into a separate PR.

first-gen/packages/status-light/test/status-light.test.ts
graynorton: This seems like a good idea to me, but open to dissenting opinions (cc @rubencarvalho).

first-gen/packages/status-light/src/spectrum-status-light.css
castastrophe: I think it makes sense personally but maybe we can also open PRs to fix it on main?

first-gen/packages/status-light/test/status-light.test.ts
castastrophe: I think we do test for dev mode on other components so I think that makes a lot of sense to add here too. Check out this reference for an example: https://github.com/adobe/spectrum-web-components/blob/ee5fdd504584b81a27b3486787234de22f996943/first-gen/packages/alert-banner/test/alert-banner.test.ts#L193

second-gen/packages/swc/components/status-light/stories/status-light.stories.ts
castastrophe: Love it!

second-gen/packages/core/components/status-light/StatusLight.base.ts
```diff
@@ -9,52 +9,115 @@
...
+                    }
+                );
+            } else if (this.hasAttribute('disabled') && 'disabled' in this) {
+                // Set aria-disabled when no warning is shown
+                if (!this.hasAttribute('aria-disabled')) {
+                    this.setAttribute('aria-disabled', 'true');
+                } else {
+                    this.removeAttribute('aria-disabled');
+                }
             }
```
graynorton, line 114: This case is interestingly different from the similar case in Badge, and I think we should do something slightly different here.

In Badge, we're doing validation of a feature introduced in S2 (`outline`), which is expected to carry on into the foreseeable future, so it seemed to make sense to put that validation logic in the base class rather than in the 2nd-gen SWC subclass. 

In this case, we have two bits of logic related to a feature being _removed_ in S2 (`disabled`):

- Validation to warn when customers try to apply `disabled` in an S2 context
- S1-specific logic for managing state related to the `disabled` feature

I'd argue that the latter should just live in the S1-specific subclass, where it will eventually fade away into oblivion without further action from us.

The former can stay here, at least for now. @rubencarvalho and I have been discussing the fact that, before going much further, we should decide on an overall philosophy regarding runtime validation of 1st-gen-to-2nd-gen changes (how much of it we want to do, and in what cases).

second-gen/packages/swc/components/status-light/stories/status-light.stories.ts
graynorton: Our vision for where this is going is that pretty much everything populating the table will flow from our code (including JSDocs) all the way through to Storybook, with little to no need for manual intervention in the `*.stories.ts` files.

To the extent that's not happening yet, I'd like to see if we can diagnose and fix issues in our tooling pipeline.

Meanwhile, I'd like to see us take a light touch, making minor tweaks as needed to unblock / ease testing during development, but otherwise not striving to manually fill gaps between what's currently showing and where we'd eventually like to get.

Since we're not expecting customers to be referring to the 2nd-gen Storybook for some time yet, the rough edges shouldn't hurt anyone, and their continued presence should help incentivize us to address the root-cause issues in the tool chain! 😁

If I look at what you've done here and compare it to the equivalent of what's being done in Badge:

```
argTypes.variant = {
    ...argTypes.variant,
    control: { type: 'select' },
    options: StatusLight.VARIANTS,
};
```

...they seem essentially equivalent in terms of behavior, and your improvement to the description can be achieved by editing the JSDoc in component implementation, so I'd like to see us fall back to the lighter version. Does that make sense? Let me know if I'm missing anything significant.

(Ultimately, I'm hoping we don't even have to do this much augmentation—I think we **should** be able to dial in our tooling to the point where it all "just works.")

second-gen/packages/swc/components/status-light/stories/status-light.stories.ts
graynorton: The `getStorybookHelpers` function is supposed to automatically handle the bindings, but it's not working as expected for certain properties; in the case of `size`, I think it's probably because properties from mixins are not propagating as they should through the Custom Element Manifest generation process.

I see this falling into the same category I was discussing in my previous comment—our preferred option is to see if we can make our tooling "just work", so we don't have to manually bind templates in our stories. If we try that and fail, then I think we can fall back to something like this.

second-gen/packages/swc/components/status-light/stories/status-light.stories.ts
graynorton: This is a good question! `SizedMixin` is not too complex, so the implications of `noDefaultSize` should be easy enough to reason about. Not having thought through the nuances, and without knowing the history, it's hard to say why SWC would be applying `noDefaultSize` if the design does, in fact, call for a default size, but I'm hopeful we'll be able to make sense of it.

Meanwhile, I would guess that there's no issue with having a default size from a Storybook perspective, even if there are legitimate technical reasons why `noDefaultSize` is the right choice from a `SizedMixin` point of view.

That said, for the sake of consistency in the way we're overriding args in our Barebones stories, I'd prefer that we insert these lines above the declaration of `meta`:

```
args['default-slot'] = 'Status light';
args.size = 'm';
```

...and then just refer to `args` in the `meta` declaration, rather than define it inline, e.g.:

```
const meta: Meta = {
    /* other props... */
    args,
};

second-gen/packages/swc/components/status-light/test/status-light.test.ts
graynorton: Technically speaking, tests weren't required for these Barebones component migrations.

I am neutral on whether we take these as part of this PR or pull them out and consider testing more holistically across the Barebones components.

@rubencarvalho, thoughts?

graynorton (changes requested):
This looks really good—nice work! 😊

Most of my comments are about places where you've gone above and beyond Barebones requirements in ways that I think we may either want to pull out into separate PRs or just hold off on for right now—let's discuss in the comment threads.

first-gen/packages/status-light/src/StatusLight.ts
marissahuysentruyt: This should now be updated 👍 

https://github.com/adobe/spectrum-web-components/pull/5800/commits/6528bc63b1614a0c8d646fa5a1d28762f0df11c4

first-gen/packages/status-light/src/spectrum-status-light.css
marissahuysentruyt: There were 2 commits related to the implementation gaps that are now reverted (one for [the CSS](https://github.com/adobe/spectrum-web-components/pull/5800/commits/7e0108db43f345ba5aef5cb8ba61dbbafbc531f3) and then one for the [first-gen stories](https://github.com/adobe/spectrum-web-components/pull/5800/commits/90512a935113df0f1631dc3265b1f729ee287652)).

I pulled those 2 commits into a separate branch based on main. I haven't pushed the new branch up yet, but can if we think that's where the work should merge- just let me know!

second-gen/packages/swc/components/status-light/stories/status-light.stories.ts
marissahuysentruyt: Ah, I see- thanks for this explanation. This will be something I need to keep in the back of my mind going forward, so thank you!

I updated the JSDoc for the variant prop, and then refactored the `argTypes.variant` definition to match badge in this commit: https://github.com/adobe/spectrum-web-components/pull/5800/commits/f5ed8fdc4f8f1e7e96f23df228f708d08d500e60

second-gen/packages/swc/components/status-light/stories/status-light.stories.ts
marissahuysentruyt: Got it- changes were made here! https://github.com/adobe/spectrum-web-components/pull/5800/commits/a4e459a32487955d4946334f4c1d3e413e18d2c6 

Would you like me to make a ticket to look into this `noDefaultSize`? I can also create one to investigate why the size attribute isn't getting bound to the template properly.

Out of curiosity, what's the reason for having the arg overrides (like the `args.size = "m"`) outside of the meta declaration? Is there less risk of overriding something on accident when things are more separated? Just trying to add to my understanding for second-gen! 😊

second-gen/packages/swc/components/status-light/stories/status-light.stories.ts
marissahuysentruyt: I haven't removed the comment yet, but certainly can if you'd rather we don't muddy up the file with too many random thoughts/ideas!

I did, however, change the commented out section for `argTypes.size` to remove all of the extra args I had so that it matched Badge once more. https://github.com/adobe/spectrum-web-components/pull/5800/commits/282eb116cab71dad77ea8d35dcbbc6cd38d7e7ba

second-gen/packages/swc/components/status-light/stories/status-light.stories.ts
graynorton: > Out of curiosity, what's the reason for having the arg overrides (like the args.size = "m") outside of the meta declaration?

TBH, there's not any one super compelling reason—either way works fine, of course. I hadn't really explicitly thought through my rationale, but I think it's something like this:

* My mental model for 2nd-gen Storybook meta info is as described in a couple of earlier comments—I would ultimately like to get our tool chain humming to the point where we just get call `getStorybookHelpers` to get `argTypes` and `args` and plop them into the `meta` declaration without any need to modify them at all—so I think that's why my preference is to reference them that way for now...all of the current overrides / additions are self-contained and can just be deleted if/when we don't need them.
* Whatever we do, it would just be nice to be as consistent as we can across components. I foresee us incrementally evolving our best practices for story authoring (and everything else, for that matter) as 2nd-gen progresses, so imagine us coming back and making multiple refinement passes over each component, and I just think that will be easier to do if we try to minimize arbitrary differences from component to component as we go.

Does that make sense?

graynorton: > Would you like me to make a ticket to look into this noDefaultSize? I can also create one to investigate why the size attribute isn't getting bound to the template properly.

Yes, that would be awesome—thank you!

marissahuysentruyt: Got it- yeah, makes sense! Thank you! I can make those tickets, as well. 🙌

second-gen/packages/core/components/status-light/StatusLight.base.ts
marissahuysentruyt: Yes, ok now I understand. Let me know if this is what you were thinking for now. The disabled logic has been [removed from the base class](https://github.com/adobe/spectrum-web-components/pull/5800/commits/155e58b214af00bb81b04997a64c19bb6bd83bf5) and [moved into the S1 component](https://github.com/adobe/spectrum-web-components/pull/5800/commits/bab9bf4d0eb57c784103bb0b37835761bc29ba49) instead.

graynorton (approved):
Looks great! Awesome work on this, @marissahuysentruyt! 👏 🙌

second-gen/packages/swc/components/status-light/stories/status-light.stories.ts
rubencarvalho: This is great! 👍 We’ll definitely find a place for these utils post-barebones.

second-gen/packages/swc/components/status-light/stories/status-light.stories.ts
rubencarvalho: Yes... at the risk of repeating what was said above - this is one of those things that should _just work™_. We’ll probably want to prioritize researching it, as it may also be masking other issues beyond template binding (like attribute, property, or type propagation).

second-gen/packages/core/components/status-light/StatusLight.base.ts
rubencarvalho: I can’t see the referenced commits in your comment, @marissahuysentruyt, but my initial thought (and I think this probably warrants a focused discussion around these dev warnings holistically post-barebones) is that we shouldn’t let 1st-gen, soon-to-be-removed properties leak into 2nd-gen. That way, when we phase out 1st-gen, there’s no technical debt lingering in 2nd-gen code that needs cleanup.

first-gen/packages/status-light/test/status-light.test.ts
rubencarvalho: This is fine! We’ll probably want to create a dedicated test helper to remove some boilerplate from these dev warnings tests, but this looks good for now 😃

second-gen/packages/swc/components/status-light/test/status-light.test.ts
rubencarvalho: I mostly created the tests for Badge to have something to run the infrastructure on. I tried to keep them as simple and focused as possible, and I think these are also pretty straightforward. This seems fine as a starting point, but we’ll almost certainly revisit it post-barebones.

PR #5821, opened 2025-10-22
chore(tooling): add changesets cli patch to allow yarn npm publish

rubencarvalho:
## Description

Adds a patch for `@changesets/cli` v2.29.7 that enables proper yarn support when publishing packages. The patch modifies two key functions:

1. **`getPublishTool`** - Now detects and returns yarn as the publish tool when yarn is the package manager
2. **`internalPublish`** - Uses `yarn npm publish` command instead of `npm publish` when yarn is detected

This ensures that `workspace:*` protocol references in package.json dependencies are properly replaced with actual version numbers during publishing.

## Motivation and context

Without this patch, `changeset publish` fails to properly replace `workspace:*` references when publishing packages in a yarn workspace. This is because changesets defaults to using `npm publish` even in yarn workspaces, and npm doesn't understand yarn's workspace protocol.

The `yarn npm publish` command properly resolves `workspace:*` to the actual published version numbers before publishing to the registry.

## Related issue(s)

- Related to upstream [changesets/changesets#1449](https://github.com/changesets/changesets/pull/1449) - Similar fix pending merge
- Related to upstream [changesets/changesets#1411](https://github.com/changesets/changesets/issues/1411) - Original issue report
- Related to upstream [changesets/changesets#674](https://github.com/changesets/changesets/pull/674) - Comprehensive rework (stalled)

This patch is a temporary solution until the upstream fix is merged and released.

### DIFF
--- a/patches/@changesets+cli+2.29.7.patch
+++ b/patches/@changesets+cli+2.29.7.patch
@@ -1,0 +1,58 @@
+diff --git a/node_modules/@changesets/cli/dist/changesets-cli.cjs.js b/node_modules/@changesets/cli/dist/changesets-cli.cjs.js
+index 53fc925..4cf6c1b 100644
+--- a/node_modules/@changesets/cli/dist/changesets-cli.cjs.js
++++ b/node_modules/@changesets/cli/dist/changesets-cli.cjs.js
+@@ -634,7 +634,13 @@ async function getPublishTool(cwd) {
+   const pm = await packageManagerDetector.detect({
+     cwd
+   });
+-  if (!pm || pm.name !== "pnpm") return {
++  if (!pm) return {
++    name: "npm"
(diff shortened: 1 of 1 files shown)


Rajdeepc:
@rubencarvalho Can you add a minimum validation step like which yarn version is this tested and which CI jobs were validated? Requesting some context here why this was done?

rubencarvalho:
> @rubencarvalho Can you add a minimum validation step like which yarn version is this tested and which CI jobs were validated? Requesting some context here why this was done?

This is compatible with yarn 4 (our version). I tested this by running:
```bash
yarn changeset version --snapshot nightly
yarn changeset publish --tag nightly --no-git-tag
```

But this will still be re-tested under SWC-1236.

PR #5858, opened 2025-11-06
feat: migrate asset to s2 styles and second-gen architecture

TarunAdobe:
## Description

This PR migrates the next-gen Asset component to Spectrum 2 styles and updates Storybook accordingly.

## Motivation and context

This PR migrates the Asset component from first-gen to the second-gen architecture, following the established pattern of separating core component logic from Spectrum-specific implementation.

**Changes:**
- Migrated CSS to align with Spectrum CSS structure

### What changed

**First-gen:**
- `first-gen/packages/asset/src/Asset.ts` - Remains unchanged for backward compatibility

**Core (Base):**
- `second-gen/packages/core/components/asset/Asset.base.ts` - Updates shared functionalities
- `second-gen/packages/core/components/asset/Asset.types.ts` - Creates types file for size and static color variants
- `second-gen/packages/core/components/asset/index.ts` - Imports and exports types

**Second-gen:**
- `second-gen/packages/swc/components/asset/Asset.ts` - Updates render method to accommodate classes
- `second-gen/packages/swc/components/asset/asset.css` - Adds CSS from `spectrum-css`
- `second-gen/packages/swc/components/asset/stories/asset.stories.ts` - Adds comprehensive story coverage

## Related issue(s)

-   SWC-1261

---

## Author's checklist

---

## Reviewer's checklist

-   [ ] Includes thoughtfully written changeset if changes suggested include `patch`, `minor`, or `major` features (**Note**: I did not add a changeset, I wasn't sure that this was set up for barebones and didn't see any for progress circle or badge, but I'm happy to add one if I'm mistaken!)

### Manual review test cases

**Run second-gen Storybook:**
```bash
yarn workspace @swc/components storybook
```

**Storybook functionality:**
- [x] All controls work correctly on Default story

**Documentation:**
- [x] All stories render correctly on Docs page
- [x] Story descriptions are clear and informative

**First-gen verification (use the PR preview or run locally):**
```bash
yarn start
# or
yarn docs:start
```
- [x] First-gen Asset remains unchanged and functional

**Code review checklist:**
- [x] TypeScript types are correct and properly exported
- [x] CSS follows second-gen patterns
- [x] No linter errors
- [x] Proper separation of concerns between core and SWC packages

### Device review

### DIFF
--- a/.changeset/avatar-accessibility-fix.md
+++ b/.changeset/avatar-accessibility-fix.md
@@ -1,0 +1,7 @@
+---
+'@spectrum-web-components/avatar': patch
+---
+
+**Added**: `is-decorative` attribute to `<sp-avatar>` to allow developers to explicitly mark avatars as decorative. When set, the avatar is hidden from screen readers with `alt=""` and `aria-hidden="true"`.
+
+**Fixed**: Fixed accessibility violation where `<sp-avatar>` rendered an underlying `img` without any `alt` attribute when no `label` was provided. The component now defaults to `alt=""` when neither `label` nor `is-decorative` is provided, and logs a dev mode warning to help developers catch missing accessibility attributes.
--- a/.changeset/tall-moments-raise.md
+++ b/.changeset/tall-moments-raise.md
@@ -1,0 +1,6 @@
+---
+'@spectrum-web-components/overlay': patch
+---
+
+**Fixed**: Third-level submenus now open correctly on Safari.
+Simplified the WebKit layout timing fix in `computePlacement()` to ensure consistent overlay positioning for deeply nested menus.
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,7 +22,7 @@ parameters:
     # 3. Commit this change to the PR branch where the changes exist.
     current_golden_images_hash:
         type: string
-        default: b88c1b90e553e4f212ad60a5d96abc4dc9ed6be0
+        default: 4304495a938a5b7dd3e64e12ebc59dfd5e300c1b
     wireit_cache_name:
         type: string
         default: wireit
--- a/1st-gen/packages/avatar/README.md
+++ b/1st-gen/packages/avatar/README.md
@@ -155,4 +155,37 @@ An avatar in a disabled state shows that an avatar exists, but is not available 
 
 ### Accessibility
 
-The `label` attribute of the `<sp-avatar>` will be passed into the `<img>` element as the `alt` tag for use in defining a textual representation of the image displayed.
+The `<sp-avatar>` component requires proper accessibility attributes to ensure screen readers can appropriately handle the avatar image.
+
+#### Labeled avatars
+
+The `label` attribute of the `<sp-avatar>` will be passed into the `<img>` element as the `alt` attribute for use in defining a textual representation of the image displayed. This is the recommended approach for avatars that convey meaningful information.
+
+```html demo
+<sp-avatar
+    label="John Doe"
+    src="https://picsum.photos/500/500"
+></sp-avatar>
+```
+
+#### Decorative avatars
+
+When an avatar is purely decorative and does not convey meaningful information, use the `is-decorative` attribute to mark it as decorative. This will hide the avatar from screen readers with `alt=""` and `aria-hidden="true"`.
+
+```html demo
+<sp-avatar is-decorative src="https://picsum.photos/500/500"></sp-avatar>
+```
+
+#### Linked avatars
+
+When an avatar has an `href` attribute, it becomes a link and requires an accessible name. Provide a `label` attribute to give the link meaningful text for screen readers.
+
+```html demo
+<sp-avatar
+    label="View John Doe's profile"
+    src="https://picsum.photos/500/500"
+    href="https://adobe.com"
+></sp-avatar>
+```
+
+**Note**: Decorative avatars should typically not be interactive links. If you need a decorative avatar with a link, you must provide a `label` attribute for accessibility compliance.
--- a/1st-gen/packages/avatar/src/Avatar.ts
+++ b/1st-gen/packages/avatar/src/Avatar.ts
@@ -48,6 +48,13 @@ export class Avatar extends LikeAnchor(Focusable) {
     @property()
     public src = '';
 
+    /**
+     * When true, marks the avatar as decorative and hides it from screen readers.
+     * The underlying img will have an empty alt attribute (alt="").
+     */
+    @property({ type: Boolean, reflect: true, attribute: 'is-decorative' })
+    public isDecorative = false;
+
     @property({ type: Number, reflect: true })
     public get size(): AvatarSize {
         return this._size;
@@ -71,11 +78,30 @@ export class Avatar extends LikeAnchor(Focusable) {
 
     private _size = defaultSize;
 
+    /**
+     * Renders the avatar image with appropriate accessibility attributes.
+     * Label takes precedence over isDecorative. When decorative and has href,
+     * aria-hidden is not set so the link remains accessible.
+     */
     protected override render(): TemplateResult {
+        let altValue = '';
+        let ariaHidden: 'true' | undefined;
+        if (this.label) {
+            altValue = this.label;
+            ariaHidden = undefined;
+        } else if (this.isDecorative) {
+            altValue = '';
+            ariaHidden = this.href ? undefined : 'true';
+        } else {
+            altValue = '';
+            ariaHidden = undefined;
+        }
+
         const avatar = html`
             <img
                 class="image"
-                alt=${ifDefined(this.label || undefined)}
+                alt=${altValue}
+                aria-hidden=${ifDefined(ariaHidden)}
                 src=${this.src}
             />
         `;
@@ -93,6 +119,53 @@ export class Avatar extends LikeAnchor(Focusable) {
         super.firstUpdated(changes);
         if (!this.hasAttribute('size')) {
             this.setAttribute('size', `${this.size}`);
+        }
+        this.warnMissingAlt();
+    }
+
+    protected override updated(changes: PropertyValues): void {
+        super.updated(changes);
+        if (
+            changes.has('label') ||
+            changes.has('isDecorative') ||
+            changes.has('href')
+        ) {
+            this.warnMissingAlt();
+        }
+    }
+
+    private warnMissingAlt(): void {
+        if (window.__swc?.DEBUG) {
+            // Warn if neither label nor isDecorative is provided
+            if (!this.label && !this.isDecorative) {
+                window.__swc.warn(
+                    this,
+                    `<${this.localName}> needs either a \`label\` attribute or \`is-decorative\` attribute to be accessible.`,
+                    'https://opensource.adobe.com/spectrum-web-components/components/avatar/#accessibility',
+                    {
+                        type: 'accessibility',
+                        issues: [
+                            'Provide a `label` attribute with meaningful alternative text for the avatar image, or',
+                            'Set `is-decorative` attribute to mark the avatar as decorative (hidden from screen readers).',
+                        ],
+                    }
+                );
+            }
+            // Warn if decorative avatar is used as a link without a label
+            else if (this.isDecorative && this.href && !this.label) {
+                window.__swc.warn(
+                    this,
+                    `<${this.localName}> with \`is-decorative\` and \`href\` requires a \`label\` attribute for the link to be accessible.`,
+                    'https://opensource.adobe.com/spectrum-web-components/components/avatar/#accessibility',
+                    {
+                        type: 'accessibility',
+                        issues: [
+                            'Provide a `label` attribute to give the link an accessible name.',
+                            'Decorative avatars should typically not be interactive links.',
+                        ],
+                    }
+                );
+            }
         }
     }
 }
--- a/1st-gen/packages/avatar/stories/avatar.stories.ts
+++ b/1st-gen/packages/avatar/stories/avatar.stories.ts
@@ -13,7 +13,7 @@ import { html, TemplateResult } from '@spectrum-web-components/base';
 
 import '@spectrum-web-components/avatar/sp-avatar.js';
 import { AvatarSize } from '@spectrum-web-components/avatar';
-import { avatar } from './images';
+import { avatar } from './images.js';
 
 export default {
     component: 'sp-avatar',
@@ -22,6 +22,7 @@ export default {
         disabled: { control: 'boolean' },
         label: { control: 'text' },
         src: { control: 'text' },
+        isDecorative: { control: 'boolean' },
     },
     args: {
         disabled: false,
@@ -35,15 +36,22 @@ interface StoryArgs {
     label?: string;
     src?: string;
     size?: AvatarSize;
+    isDecorative?: boolean;
 }
 
 const Template = ({
     label = 'Place Dog',
     src = avatar,
     size = 100,
+    isDecorative = false,
 }: StoryArgs = {}): TemplateResult => {
     return html`
-        <sp-avatar label=${label} src=${src} size=${size}></sp-avatar>
+        <sp-avatar
+            label=${label}
+            src=${src}
+            size=${size}
+            ?is-decorative=${isDecorative}
+        ></sp-avatar>
     `;
 };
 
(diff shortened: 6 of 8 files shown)


package.json
```diff
@@ -24,7 +24,7 @@
         "postinstall": "husky || true && patch-package",
         "start": "yarn start:1st-gen & yarn start:2nd-gen",
         "start:1st-gen": "yarn workspace @spectrum-web-components/1st-gen start",
-        "start:2nd-gen": "yarn workspace @adobe/swc start",
+        "start:2nd-gen": "yarn workspace @adobe/swc storybook",
```
TarunAdobe: There is no `yarn start` command in 2nd gen swc package

2nd-gen/packages/swc/components/asset/asset.css
```diff
@@ -1,91 +1,49 @@
...
-        --system-asset-folder-background-color
-    );
-    --spectrum-asset-file-background-color: var(
-        --system-asset-file-background-color
-    );
-    --spectrum-asset-icon-outline-color: var(--system-asset-icon-outline-color);
 }
 
-::slotted(*) {
+.spectrum-Asset-image {
```
Rajdeepc, line 26: This is a breaking change for the existing behaviour if consumers are slotting an `` without class. 
You can keep this class with a migration shim.
```css
::slotted() { /* same rules */ }
.spectrum-Asset-image { /* same rules */ }
```

2nd-gen/packages/swc/components/asset/asset.css
```diff
@@ -1,91 +1,49 @@
...
+    fill: var(--spectrum-gray-500);
 }
 
-@media (forced-colors: active) {
-    :host {
-        --highcontrast-asset-icon-background-color: currentcolor;
-    }
+.spectrum-Asset-fileOutline {
+    fill: var(--spectrum-gray-500);
 }
```
Rajdeepc, line 53: Is this intentional? High-contrast users may get incorrect colors or no contrast adjustments.

2nd-gen/packages/swc/components/asset/Asset.ts
```diff
@@ -11,62 +11,89 @@
...
-            return file(this.label);
-        } else if (this.variant === 'folder') {
-            return folder(this.label);
-        }
-        return html` <slot></slot> `;
+        return html`
+            <div
+                class=${classMap({
+                    ['spectrum-Asset']: true,
+                })}
```
Rajdeepc, line 85: only one static class is set you could just set `class="spectrum-Asset"` directly. minor simplification.

2nd-gen/packages/swc/components/asset/asset.css
```diff
@@ -1,91 +1,49 @@
...
-}
-
-.folderBackground {
-    fill: var(--spectrum-asset-folder-background);
 }
 
-.fileBackground {
-    fill: var(--spectrum-asset-file-background);
+.spectrum-Asset-folderBackground {
+    fill: var(--spectrum-gray-200);
```
Rajdeepc, line 40: Do these map to the same value? Can you surface up these changes in the PR documentation too for quick reference by reviewers. Its hard to visualize from here

Rajdeepc (changes requested):
Good start!

2nd-gen/packages/swc/components/asset/Asset.ts
```diff
@@ -11,62 +11,89 @@
...
 `;
 
 /**
+ * Use an asset element to visually represent a file, folder, or image.
+ * File and folder representations center themselves within the available space.
+ * Images are contained to the element’s size and centered.
+ *
  * @element swc-asset
- * @slot - content to be displayed in the asset when an acceptable value for `file` is not present
+ * @slot - content to be displayed when no `variant` is set (typically an <img>)
```
marissahuysentruyt: Right now, these docs are missing the `` tag in the parentheses) in the description in the control table: 

Screenshot 2025-11-11 at 12 29 15 PM

just wrapping it in backticks seems to do the trick.
```suggestion
 * @slot - The content to be displayed when no `variant` is set (typically an `<img>`)
```
Screenshot 2025-11-11 at 12 51 30 PM

2nd-gen/packages/swc/components/asset/Asset.ts
```diff
@@ -11,62 +11,89 @@
...
+            class="spectrum-Asset-folderOutline"
             d="M29,6H18.3c-0.1,0-0.2,0-0.4-0.2l-3.1-3.1C14.4,2.3,13.8,2,13.1,2H3C1.3,2,0,3.3,0,5v22c0,1.6,1.3,3,3,3h26c1.7,0,3-1.4,3-3V9C32,7.3,30.7,6,29,6z M31,27c0,1.1-0.9,2-2,2H3c-1.1,0-2-0.9-2-2V7h28c1.1,0,2,0.9,2,2V27z"
         ></path>
     </svg>
 `;
 
 /**
+ * Use an asset element to visually represent a file, folder, or image.
+ * File and folder representations center themselves within the available space.
+ * Images are contained to the element’s size and centered.
```
marissahuysentruyt: This feels like maybe something we would want to display on the docs page. What do you think about moving this JSDoc to the stories file instead, so that it displays under the "Asset" title?

Screenshot 2025-11-11 at 12 35 13 PM

This is something I stole from the divider stories file. 😊

2nd-gen/packages/core/components/asset/Asset.base.ts
```diff
@@ -10,14 +10,67 @@
...
+
+    /**
+     * The variant of the asset. When not provided, slot content is rendered (e.g., an image).
+     */
     @property({ type: String, reflect: true })
-    public variant: 'file' | 'folder' | undefined;
+    public variant: AssetVariant | undefined;
 
+    /**
+     * Accessible label for the asset’s SVG variant.
```
marissahuysentruyt: Super nit pick: our variants are technically named them "file" or "folder," and not "SVG." Is it odd or confusing to mention an SVG variant? I'll leave it up to you if you want to change it.

```suggestion
* Accessible label for the asset’s file and folder variants.
```

2nd-gen/packages/swc/components/asset/Asset.ts
caseyisonit: i would like to keep the classMap pattern so its easily extensible if it arises... not a deal breaker though

2nd-gen/packages/swc/components/asset/asset.css
marissahuysentruyt: These should be the same values, yes (and they look to be mapped correctly).

Before Cassondra left, she started removing some of the complexity from a chunk of components, asset being one of them. You can see the changes (and remove of the `--spectrum-asset-` variables) here: https://github.com/adobe/spectrum-css/pull/4257/files#diff-3a87a9f21648ac03bf4532993e3cdf48604b9862efd1c0ea68c3f6c2d44b32b3

In a different PR (that does the same and removes mods and some of the `--spectrum` prefixed variables for asset list and miller), Cassondra left her reasoning as to why: https://github.com/adobe/spectrum-css/pull/4260#discussion_r2452924846

Hopefully that's useful! I absolutely agree with @Rajdeepc that we should surface the changes in a changeset. None of these migrations have had a changeset yet. (I asked a similar question in Slack when I was working on status light). Do we have the changeset management situation figured out? I've been out, so it might be and we need a changeset!

2nd-gen/packages/swc/components/asset/asset.css
marissahuysentruyt: This was removed in the same PR I mentioned above, where we were removing the mods for asset. Personally, I like this removal after talking with @5t3ph during the app frame side nav WHCM stuff. I don't think this `currentcolor` is necessary and just adds to the CSS we're shipping for no super strong reason. We can rely on the default handling of high contrast instead of defining this icon background color. 

This is with the high contrast variable. I don't think anything is really being added with this query, but I could be wrong.
Screenshot 2025-11-11 at 3 21 35 PM
Screenshot 2025-11-11 at 3 19 40 PM

package.json
marissahuysentruyt: Is this the start command you were looking for? Or does this do something else? 

https://github.com/adobe/spectrum-web-components/blob/9a936540b599d1482f8ed131223be578ca3fb5ec/2nd-gen/package.json#L24

2nd-gen/packages/swc/components/asset/asset.css
Rajdeepc: Thanks for the explanation. Looks fine.

2nd-gen/packages/swc/components/asset/Asset.ts
Rajdeepc: ideally there is no performance impact here since it just adds a tiny internal `Set` to render the class list but yea I’m fine either way unless there’s a strong preference to simplify.

2nd-gen/packages/swc/components/asset/asset.css
Rajdeepc: Please surface up a changeset on this, we will check how the tool is scanning it.

marissahuysentruyt:
@TarunAdobe I had one more thing! 

Looking at the title of this PR, I'm curious if this work is a `feat` as opposed to a `chore`, since it's the migration into second-gen. Should we update this PR title to reflect that? Is my feat vs. chore logic even accurate? Asset is such a small component so I'm not sure if it's really a "breaking" change or not.

package.json
TarunAdobe: Shoot yes!! I think i need to update `start:2nd-gen` to `yarn workspace @spectrum-web-components/2nd-gen start`

2nd-gen/packages/swc/components/asset/asset.css
TarunAdobe: but can we not break this for spectrum-two?

2nd-gen/packages/swc/components/asset/Asset.ts
TarunAdobe: niceee... sure

2nd-gen/packages/swc/components/asset/asset.css
Rajdeepc: we should try not to treat S2 as a chance to introduce breaking changes unless they’re really necessary. These can have a big impact on adoption for product teams, so it’s worth being mindful about when and why we make them. If a change truly adds more value than the potential disruption, we can definitely go for it

2nd-gen/packages/swc/components/asset/asset.css
```diff
@@ -1,91 +1,49 @@
...
-    --spectrum-asset-folder-outline: var(
-        --mod-asset-icon-outline-color,
-        var(--spectrum-asset-icon-outline-color)
-    );
-    --spectrum-asset-file-outline: var(
-        --mod-asset-icon-outline-color,
-        var(--spectrum-asset-icon-outline-color)
-    );
-
+.spectrum-Asset {
```
marissahuysentruyt, line 18: The CSS here looks great- matches the S2 code we have in the `spectrum-two` branch. 👍 🥳

My only question is do we need a selector for `:host`, giving it the `display: flex` property? I believe the default display for `:host` is inline.

marissahuysentruyt:
```suggestion
:host {
display: flex;
}

.spectrum-Asset {
    display: flex; // we would also need this here for the `align-items` & `justify-content`
    align-items: center;
	justify-content: center;
	inline-size: 100%;
	block-size: 100%;
}

File: 2nd-gen/packages/core/components/asset/Asset.base.ts [hunk 10]
```diff
@@ -10,14 +10,67 @@
...
 
+import { PropertyValues } from 'lit';
 import { property } from 'lit/decorators.js';
 
 import { SpectrumElement } from '@spectrum-web-components/core/shared/base/index.js';
 
+import { ASSET_VARIANTS, type AssetVariant } from './Asset.types.js';
+
+/**
+ * @slot - The content to render when no `variant` is provided (typically an <img> element)
```
marissahuysentruyt: There is another JSDoc comment in swc/components/asset/Asset.ts about the `@slot` that is almost identical to this. Is there a reason to have both? It looks like the content in the `swc` directory's Asset.ts file is the one being used in the control table when storybook renders. 

I'm not opposed to keeping both of them, but I was just curious if we need both, or if some other components _should_ have both.

2nd-gen/packages/swc/components/asset/Asset.ts
```diff
@@ -11,62 +11,85 @@
...
+ * @example
+ * <swc-asset style="block-size: 128px">
+ *   <img class="spectrum-Asset-image" src="example.png" alt="Example image" />
+ * </swc-asset>
+ *
+ * @example
+ * <swc-asset variant="file" style="min-inline-size: 150px; block-size: 128px"></swc-asset>
+ *
+ * @example
+ * <swc-asset variant="folder" style="min-inline-size: 150px; block-size: 128px"></swc-asset>
```
marissahuysentruyt: For these examples, do we want to encourage setting inline styles with the `style` attribute? Or is that more often what consumers would do? Maybe we could show just one example with inline styles.

I don't think this is a blocker, but I just noticed each one had inline styles.

2nd-gen/packages/swc/components/asset/stories/asset.stories.ts
```diff
@@ -11,34 +11,83 @@
...
-// };
+// ───────────────
+//    STORIES
+// ───────────────
+
+args['default-slot'] = IMAGE_PLACEHOLDER_STRING();
 
 export const Default: Story = {
-    render: (args) => html` <swc-asset variant="${args.variant}"></swc-asset> `,
+    render: (args) => template(args),
```
marissahuysentruyt: Just for consistency, I believe the preference Gray was trying to establish is to include this `render` in the `Meta` object. So this line would move up under `argTypes` (line 42).

Then, we can clean up each story by just redefining args: 

```suggestion
// @ line 42
render: (args) => template(args),
... // the rest of the metadata and the default export...

export const Default: Story = {
    args: {
        variant: undefined,
    },
};

export const File: Story = {
    args: {
        variant: 'file',
    },
    tags: ['!dev'],
};

export const Folder: Story = {
    args: {
        variant: 'folder',
    },
    tags: ['!dev'],
};
```

That does however, remove the inline styles you have on the stories, but are those inline styles necessary? Does this so acceptable to you?

marissahuysentruyt (changes requested):
I went through it more in depth, and I have a few more questions for you. It still looks really great!

I'd like 2 changes I think: 
1. I think the big thing for me ended up being in the stories file, just keeping the render method in the `meta` to align with how the other migrated components have it set up currently.
2. Pretty sure we need a `:host` selector in the CSS that also has the `display: flex` property.

2nd-gen/packages/core/components/asset/Asset.base.ts
Rajdeepc: The base class should focus on implementation details and internal APIs, while the concrete classes should provide user-facing documentation and examples. 
I agree here that this can be removed since implementation in `swc/Asset.ts` is the appropriate place for customer-facing documentation that gets rendered in Storybook controls.

2nd-gen/packages/swc/components/asset/asset.css
Rajdeepc: Good catch @marissahuysentruyt ! Both display values serve different purposes and are both necessary.

2nd-gen/packages/swc/components/asset/stories/asset.stories.ts
Rajdeepc: The inline styles aren't strictly necessary if the component's default sizing is providing a reasonable display in Storybook. Since file/folder icons should size themselves based on their container, I feel removing them should be fine. The stories will just show the icons at their natural/intrinsic size.

marissahuysentruyt (approved):
swedishchefkiss

looks wonderful! I also verified that we get the console warnings in the browser: 

Screenshot 2025-11-18 at 8 45 12 AM

I wanted to note that I'm actually seeing the alt text render for the default asset variant, with [the image](https://cdn2.thecatapi.com/images/d4i.jpg), right now. I believe that's because [cloudflare is down](https://www.cloudflarestatus.com/) this morning, though. We should make sure the image loads once cloudflare is back up.

Issue #5715, opened 2025-09-02
[Bug]: Scrolling picker inside of another overlay closes it on touch devices


lehelen19:
### Code of conduct

- [x] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-picker, sp-overlay

### Library version

1.7.0

### Expected behavior

Picker should not close when scrolling its contents on a touch device.

### Actual behavior

The menu item inside of the picker is selected and the picker closes.

CCEX bug: https://jira.corp.adobe.com/browse/CCEX-229974

### Screenshots

https://github.com/user-attachments/assets/ff676c05-9e39-42be-9f56-22d9eef93555

### How can we reproduce this issue?

1. Go to https://stackblitz.com/edit/vitejs-vite-euegaejj?file=package.json,src%2Fmy-element.ts
2. In dev tools, use the iPad Air dimensions
3. Click on the button to open the overlay
4. Click on the picker
5. Try to drag the picker content to scroll and observe how it closes on pointer up

### Severity

SEV 2

### Would you like to track this issue in Jira?

- [x] Yes, please tell me the ticket number!

lehelen19:
Should be fixed by https://github.com/adobe/spectrum-web-components/pull/5616

lehelen19:
Still unable to scroll on iPad.

https://github.com/user-attachments/assets/01647081-572d-4b88-80fc-eb1d7858c5f2

Rajdeepc:
Hi @lehelen19 Can you please reverify on the latest build. I cannot see this issue. `iPad Pro 13-inch M4 iOS 18.5`

https://github.com/user-attachments/assets/2e96c379-2225-49ec-952e-41ad52a95050

lehelen19:
Yes, I am still seeing this problem. The StackBlitz automatically uses the latest version.

https://github.com/user-attachments/assets/bc4a5554-a6ff-4bed-8b06-5863827015e8

We are also still seeing it in Express on the 1.9.0 SWC build

najikahalsema:
SWC-1334

Rajdeepc:
@lehelen19 I would need your help to test this in your iPad if you can find one for yourself before I can make this change ready for review.  I did one round on a loaner device and it is working fine with this change.
**Root cause**: Safari’s layout engine required an extra frame for menu rendering after scrollIntoView.
**Fix**: Added a 1-frame requestAnimationFrame delay in `PlacementController.computePlacement()` (excluding submenus).

[...]

PR #5868, opened 2025-11-10
fix(picker): scrolling issue inside overlays in Safari iPad

Rajdeepc:
## Description
Picker menus inside overlays on Safari/iOS couldn't scroll to the bottom after selecting an item and reopening. The menu would appear truncated (90px height instead of the full 947px), making lower items inaccessible.

### Root cause
A timing issue in `PlacementController`: the overlay's height measurement occurred before the menu's scrollIntoView operation completed. Safari's layout engine needed an additional frame to fully render the menu after scroll repositioning, but computePlacement() was measuring prematurely, capturing the collapsed/transitional state rather than the final rendered height.

### Fix
Added a 1-frame requestAnimationFrame delay in PlacementController.computePlacement() specifically for:

- Safari/WebKit browsers (detected via user agent)
- Overlays containing sp-menu elements
- Excludes submenus (which need immediate positioning for hover responsiveness)

This allows Safari's layout engine to complete `scrollIntoView` and stabilize the menu's rendered height before the overlay calculates positioning and applies size constraints.

### Why Safari-specific
Chrome's layout engine completes `scrollIntoView` synchronously within the same frame, while Safari/WebKit requires an additional frame for the layout to settle. This is a known Safari rendering behavior, not a bug.

#### **PlacementController Timing Fix** (`PlacementController.ts`)
Adds a 1-frame delay in PlacementController for Safari browsers when rendering picker menus inside overlays, allowing scrollIntoView to complete before height measurement. Excludes submenus to maintain hover responsiveness.

```typescript
async computePlacement(): Promise<void> {
    const { options, target } = this;
    
    await (document.fonts ? document.fonts.ready : Promise.resolve());
    
    // NEW: Wait for content to fully render and settle
     //Safari/iOS-specific fix: Add small delay for picker menus to allow scrollIntoView to complete
        const isSafari =
            /WebKit/.test(navigator.userAgent) &&
            !/Chrome/.test(navigator.userAgent);

        // Check if this is a submenu overlay (slot="submenu")
        // Submenus need immediate positioning for hover responsiveness
        const isSubmenu = Array.from(this.host.elements).some(
            (el) => el.getAttribute?.('slot') === 'submenu'
        );

        if (isSafari && !isSubmenu) {
            const hasMenu = Array.from(this.host.elements).some(
                (el) =>
                    el.tagName === 'SP-MENU' || el.querySelector?.('sp-menu')
            );

            if (hasMenu) {
                // Wait 1 frame for Safari layout to settle after scrollIntoView
                await new Promise((resolve) => requestAnimationFrame(resolve));
            }
        }
}
```

## Motivation and context

- Users cannot access all menu items after the first selection
- Affects pickers in modal dialogs, popovers, and other overlay contexts
- **Reproducible in Safari (desktop and iPad)** ⚠️
- Chrome works correctly (but fix is still needed for consistency)

## Related issue(s)

-   fixes [SWC-1334]
-   fixes https://github.com/adobe/spectrum-web-components/issues/5715

---

## Author's checklist

---

## Reviewer's checklist

### Manual review test cases

-   [x] Safari and iPad Test

    1. Go [here](https://swcpreviews.z13.web.core.windows.net/pr-5868/docs/storybook/?path=/story/picker--picker-ini-pad-safari)
    2. Open this in iPad
    3. Click on Open Overlay button
    4. Scroll and select any item
    5. close and open again and scroll

## Breaking Changes

None. This is a bug fix with no API changes.

### Device review


Rajdeepc:
Working fix.

https://github.com/user-attachments/assets/32d5dc06-0180-4d5d-9861-0514fbd18e51

1st-gen/packages/menu/src/Menu.ts
caseyisonit, line 1: Can you explain what the double bangs were doing and why removing them fixes this?

1st-gen/packages/overlay/test/overlay.test.ts
caseyisonit, line 1: we should not need to be mocking user agents in these tests, we already run them for safari, chrome, and firefox. we should be using our browser checks `isWebkit()` for specific browser needs in tests. this potentially is introducing an internal tool spec that you would have to be aware of to change as the browsers update

caseyisonit: There's also a lot of repetitive code in each test that could be moved to the beforeEach. Actually looking at this closer, these all look identical in test assertions. this functionality should already be covered in other menu tests. can you explain the purpose and need of these tests so i can better understand?

1st-gen/packages/picker/stories/picker.stories.ts
caseyisonit, line 1: Do we need this story? feels odd to have a device and browser specific story when we cant control the canvas

rise-erpelding: I can see the value of wanting a story to be able to validate, wondering if it would work to repurpose one of the other stories. If we add more menu items to `PickerInModalOverlay` instead would we still be able to confirm whether or not the bug exists there?

Rajdeepc: @caseyisonit This story has been tagged to dev mode only and excluded from the VRTs. I created this for development and testing purpose. Will remove once review is done
@rise-erpelding  Yes you can play with the items in the DOM.

1st-gen/packages/menu/src/Menu.ts
Rajdeepc: This is not the fix. Apologies if this has surfaced up as the fix scenario. They're functionally the same in this context. The `!!` pattern is useful when you need to store or return an explicit boolean, here its not the case. No impact on the current logic or any regressions.

1st-gen/packages/overlay/test/overlay.test.ts
Rajdeepc: You're correct on both fronts. These tests are an anti-pattern. I added these initially to achieve branch coverage on the Safari-specific conditional path in `PlacementController.computePlacement()`.  If this optimisation regresses, our existing overlay positioning assertions will fail on actual WebKit, which is the signal that matters. I agree that this is giving false confidence and I just checked yes we already have a validation strategy. I have excluded the entire Safari specific block from the coverage reporting.

rise-erpelding (approved):
Trying to replicate the bug pre-fix in Storybook to confirm that this fix works was a little tricky for me, on an iPad even after adding additional menu items to the pickers within an overlay, I couldn't get Storybook to replicate the same behavior that I saw in the StackBlitz reproduction.

That said, this does seem to improve the scrolling behavior overall (I can scroll to the bottom where I couldn't before) so I'm optimistic that this might address it!

PR #5871, opened 2025-11-10
chore: modernize and unify TypeScript configuration

caseyisonit:
### Description

This PR modernizes TypeScript configuration across both 1st-gen and 2nd-gen packages by upgrading to TypeScript 5.9.2 and consolidating shared compiler options into a unified base configuration file.

**Key changes:**

- Upgraded TypeScript from 5.3.3 to 5.9.2 for improved type checking and modern language features
- Created `tsconfig.base.json` at repository root containing shared compiler options
- Refactored `1st-gen/tsconfig.json` and `2nd-gen/tsconfig.json` to extend the base configuration
- Unified module resolution strategy to `"bundler"` for modern tooling compatibility (Vite, esbuild)
- Standardized ES2018 target across both generations
- Fixed type errors exposed by stricter TypeScript 5.9 checking
- Added missing type definitions for build tools

### Motivation and context

The TypeScript configuration was duplicated across 1st-gen and 2nd-gen with slight inconsistencies, making maintenance difficult. TypeScript 5.3.3 was outdated and missing improvements from newer versions. This consolidation reduces duplication, ensures consistent type checking, and positions the project for modern build tooling.

### Related issue(s)

- Part of ongoing TypeScript modernization effort
- Improves developer experience and build tooling compatibility

### Author's checklist

### Reviewer's checklist

### Manual review test cases

- [x] **Verify 1st-gen packages build successfully** [@cdransf]
    1. Run `yarn build` in 1st-gen directory
    2. Verify no TypeScript errors
    3. Expect successful build output for all packages

- [x] **Verify 2nd-gen packages build successfully** ﻿﻿[@cdransf]
    1. Run `yarn build` in 2nd-gen directory
    2. Verify no TypeScript errors
    3. Expect successful build output for all packages

- [x] **Verify IDE TypeScript integration** [@cdransf]
    1. Open a component file in VSCode
    2. Check that type hints and autocomplete work correctly
    3. Expect no TypeScript errors in IDE

- [x] **Verify cross-generation imports** [@cdransf]
    1. Check 1st-gen components that import from `@spectrum-web-components/core/*`
    2. Verify path mapping resolves correctly
    3. Expect no module resolution errors

### Device review

- [x] Did it pass in Desktop?


CONTRIBUTOR-DOCS/01_contributor-guides/10_typescript-configuration.md
rubencarvalho, line 1: nit: I am a bit conflicted about if we need this TypeScript config documentation. Right now, I feel like it may be over-justifying and also mixing some concerns.

CONTRIBUTOR-DOCS/01_contributor-guides/10_typescript-configuration.md
```diff
@@ -0,0 +1,252 @@
...
+last 2 Chrome versions
+last 2 Firefox versions
+last 2 Safari versions
+last 2 iOS versions
+not IE 11
+```
+
+**Node.js requirement**: `>=20.19.0` (defined in root `package.json`)
+
+All of these targets fully support ES2020 features.
```
rubencarvalho: This section seems redundant

CONTRIBUTOR-DOCS/01_contributor-guides/10_typescript-configuration.md
```diff
@@ -0,0 +1,252 @@
...
+└── 2nd-gen/tsconfig.json       ← 2nd-gen: Bundler resolution
+    └── packages/*/tsconfig.json ← Package-specific overrides
+```
+
+**Benefits of this structure:**
+
+- Single source of truth for common settings
+- Clear documentation of intentional differences
+- Easy to update TypeScript version or shared settings
+- Reduced configuration duplication
```
rubencarvalho: This is redundant

CONTRIBUTOR-DOCS/01_contributor-guides/10_typescript-configuration.md
```diff
@@ -0,0 +1,252 @@
...
+|-----------|------------|------------|---------|
+| 1st-gen | `node` | TypeScript compiler | Project references, tsc incremental builds |
+| 2nd-gen | `bundler` | Vite | Modern bundler optimizations, faster HMR |
+
+Both use `module: "ESNext"` for maximum interoperability.
+
+**Why different resolution strategies?**
+
+- **1st-gen** uses the classic `node` resolution because it relies on TypeScript's project references and incremental compilation features
+- **2nd-gen** uses `bundler` resolution to take advantage of Vite's optimized module resolution and Hot Module Replacement (HMR)
```
rubencarvalho: This is redundant and we change 1st-gen to be `bundler` too.

CONTRIBUTOR-DOCS/01_contributor-guides/10_typescript-configuration.md
```diff
@@ -0,0 +1,252 @@
...
+
+```typescript
+import { SpectrumElement, html, css } from '@spectrum-web-components/base';
+```
+
+**Direct core import (advanced use cases):**
+
+```typescript
+import { SpectrumElement } from '@spectrum-web-components/core/shared/base/Base.js';
+```
```
rubencarvalho: This is not accurate. The `@spectrum-web-components/base` in 1st-gen is re-exporting from `core` to maintain backwards compatibility and avoid us having to change all its imports to instead depend on 2nd-gen.

rubencarvalho: The "**Direct core import (advanced use cases):**" is a bit misleading.

CONTRIBUTOR-DOCS/01_contributor-guides/10_typescript-configuration.md
```diff
@@ -0,0 +1,252 @@
...
+This allows TypeScript to resolve cross-generation imports correctly during development and builds.
+
+**Build order requirement**: When building from scratch, 2nd-gen must be built before 1st-gen:
+
+```bash
+yarn build:2nd-gen  # Build core dependency first
+yarn build:1st-gen  # Can now resolve 2nd-gen imports
+```
+
+The top-level `yarn build` command handles this automatically.
```
rubencarvalho: If this is handled automatically, do we need this?

CONTRIBUTOR-DOCS/01_contributor-guides/10_typescript-configuration.md
```diff
@@ -0,0 +1,252 @@
...
+
+### Current version
+
+**TypeScript version**: 5.9.2 (aligned across all workspaces as of November 7, 2025)
+
+All three locations use the same version:
+
+- `1st-gen/package.json`
+- `1st-gen/projects/css-custom-vars-viewer/package.json`
+- `1st-gen/projects/example-project-rollup/package.json`
```
rubencarvalho: I would avoid keeping this "current version" section. It will most likely go out of sync.

CONTRIBUTOR-DOCS/01_contributor-guides/10_typescript-configuration.md
```diff
@@ -0,0 +1,252 @@
...
+
+When updating TypeScript versions, follow these steps to ensure consistency:
+
+**1. Update all three package.json files:**
+
+```bash
+# Update version in all three locations
+# 1st-gen/package.json
+# 1st-gen/projects/css-custom-vars-viewer/package.json
+# 1st-gen/projects/example-project-rollup/package.json
```
rubencarvalho: These paths are example projects we don't really use. Updating TypeScript here is of no real benefit for us.

CONTRIBUTOR-DOCS/01_contributor-guides/10_typescript-configuration.md
```diff
@@ -0,0 +1,252 @@
...
+
+```typescript
+// Source code
+const value = obj?.nested?.prop ?? 'default';
+
+// Compiled output (ES2020)
+const value = obj?.nested?.prop ?? 'default'; // Preserved as-is
+
+// Previously with ES2018 target (down-leveled)
+const value = (_a = (_b = obj === null || obj === void 0 ? void 0 : obj.nested) === null || _b === void 0 ? void 0 : _b.prop) !== null && _a !== void 0 ? _a : 'default';
```
rubencarvalho: I think we should defer this to an official reference. e.g. tc39

CONTRIBUTOR-DOCS/01_contributor-guides/10_typescript-configuration.md
```diff
@@ -0,0 +1,252 @@
...
+
+All three locations use the same version:
+
+- `1st-gen/package.json`
+- `1st-gen/projects/css-custom-vars-viewer/package.json`
+- `1st-gen/projects/example-project-rollup/package.json`
+
+### Upgrading TypeScript
+
+When updating TypeScript versions, follow these steps to ensure consistency:
```
rubencarvalho: If we really want to keep this here, a warning about TS not following semver could be useful.

1st-gen/tsconfig.json
```diff
@@ -1,32 +1,24 @@
...
-        "declaration": true,
         "emitDeclarationOnly": true,
-        "experimentalDecorators": true,
         "importHelpers": true,
         "inlineSources": true,
-        "isolatedModules": true,
-        "lib": ["es2018", "dom", "dom.iterable"],
-        "module": "esNext",
+        // 1st-gen specific: Classic Node resolution for tsc builds
         "moduleResolution": "node",
```
rubencarvalho: I don't see a reason this should stay "node" :)

1st-gen/scripts/cem-plugin-react-wrapper.js
```diff
@@ -85,8 +85,8 @@ function genPackageJson(
             "${dependencyPkgName}"
         ],
         "dependencies": {
-            "@lit/react": "^1.0.2",
-            "${dependencyPkgName}": "^${dependencyPkgVersion}"
+            "@lit/react": "1.0.8",
```
rubencarvalho: Not sure we need to pin this one... It may require customers to juggle their deps...

CONTRIBUTOR-DOCS/01_contributor-guides/10_typescript-configuration.md
```diff
@@ -0,0 +1,252 @@
...
+
+## Unified ES2020 target
+
+As of November 2025, this monorepo uses **ES2020** as the unified JavaScript target across both 1st-gen and 2nd-gen workspaces.
+
+### Why ES2020?
+
+We chose ES2020 for several reasons:
+
+- **Modern browser support**: All browsers in our target list support ES2020 (Chrome 80+, Firefox 74+, Safari 13.1+, Edge 80+)
```
rubencarvalho: If our target browsers are a driving force of this decision, we would could have gone es2022 or higher :)

rubencarvalho (commented):
Looks really good :) thanks for taking a crack at this! I only have some hesitation around the documentation part, otherwise, 🚢 it.

CONTRIBUTOR-DOCS/01_contributor-guides/10_typescript-configuration.md
caseyisonit: im comfortable going with es2022

CONTRIBUTOR-DOCS/01_contributor-guides/10_typescript-configuration.md
caseyisonit: honestly im cool killing this document all together, i intitally was trying to document the decision but it does seem redundant returning to this after the break.

CONTRIBUTOR-DOCS/01_contributor-guides/10_typescript-configuration.md
caseyisonit: i didnt feel confident changing 1st-gen to bundler because it uses vite and we dont have that configured in 1st-gen, maybe we can chat and you can help me understand the risks better

1st-gen/tsconfig.json
caseyisonit: i didnt feel comfortable changing how we resolve modules in 1st-gen. do we know if any teams are using commonJS?

1st-gen/scripts/cem-plugin-react-wrapper.js
caseyisonit: i thought we were pinning all dependencies for security?

Rajdeepc: @rubencarvalho We need to validate this first with the generated apps. I will talk to few people.

rubencarvalho: @caseyisonit This is not a dependency in our repository, this is a template literal to generate the `package.json` file for each of the React wrapper components.

1st-gen/tsconfig.json
rubencarvalho: Changing `moduleResolution` in our project does not affect downstream consumers. I raised this since it felt in scope of the ticket, as it touches TypeScript configuration.
Still, it’s worth a sanity check by validating that the emitted `.d.ts` files remain valid.

CONTRIBUTOR-DOCS/01_contributor-guides/10_typescript-configuration.md
rubencarvalho: Sooo... should we? 😛

1st-gen/scripts/cem-plugin-react-wrapper.js
caseyisonit: DUHHHHHHH ok that makes sense, ill add back the ^

.changeset/card-keyboard-handler-fix.md
```diff
@@ -0,0 +1,5 @@
+---
+'@spectrum-web-components/card': patch
+---
+
+**Fixed** keyboard event handler to prevent unintended fall-through behavior in switch cases. Added missing `break` statements after Space key and Enter key handlers to ensure proper event handling flow.
```
Rajdeepc: Not sure if you missed removing this? I don't see a keyboard handler fix in this PR.

1st-gen/packages/card/src/Card.ts
```diff
@@ -147,9 +147,12 @@ export class Card extends LikeAnchor(
                     event.preventDefault();
                     break;
                 }
+                break;
+
             case 'Enter':
             case 'NumpadEnter':
                 this.click();
+                break;
```
Rajdeepc, line 155: Is this breaking any existing logic?

1st-gen/tools/reactive-controllers/src/ColorController.ts
```diff
@@ -686,11 +683,11 @@ export class ColorController {
      * Converts the current color to the specified format.
      *
      * @param format - The desired color format. It can be a string representing one of the valid formats
-     * ('srgb', 'hsva', 'hsv', 'hsl', 'hsla') or a ColorSpace object.
+     * ('srgb', 'hsva', 'hsv', 'hsl', 'hsla') or a ColorSpace.Space object.
      * @returns The color object in the specified format.
      * @throws Will throw an error if the provided format is not a valid string format.
      */
-    getColor(format: string | ColorSpace): ColorObject {
+    getColor(format: string | ColorSpace.Space): ColorObject {
```
Rajdeepc, line 690: This is a **breaking API change**. Even though it's a type-only change, consumers may need to update their code. The changeset doesn't mention this also.

2nd-gen/packages/core/components/badge/Badge.types.ts
```diff
@@ -28,7 +28,7 @@ export const BADGE_VALID_SIZES = [
     'm',
     'l',
     'xl',
-] as const satisfies readonly ElementSize[];
+] as const satisfies ElementSize[];
```
Rajdeepc, line 31: This can effect type inference.

tsconfig.base.json
```diff
@@ -0,0 +1,33 @@
...
+
+        // Decorators (required for Lit)
+        "experimentalDecorators": true,
+        "forceConsistentCasingInFileNames": true,
+        "isolatedModules": true,
+        "lib": ["ES2018", "DOM", "DOM.Iterable"],
+
+        // Module settings (shared)
+        "module": "ESNext",
+        "moduleResolution": "bundler",
```
Rajdeepc, line 18: You need to document the `moduleResolution: bundler` migration impact. This is a significant breaking change for any downstream tooling that relies on `Node.js-style` module resolution.

.vscode/settings.json
```diff
@@ -10,6 +10,35 @@
         "editor.wordWrapColumn": 120,
         "editor.wordWrap": "bounded"
     },
+    "[typescript]": {
+        "editor.defaultFormatter": "esbenp.prettier-vscode",
+        "editor.formatOnSave": true,
```
Rajdeepc: This setting is very opinionated

Rajdeepc (dismissed):

2nd-gen/packages/core/components/badge/Badge.types.ts
caseyisonit: you can not override a readonly type, this is a compilitation error

1st-gen/tools/reactive-controllers/src/ColorController.ts
caseyisonit: The original type was importable; it is also the incorrect type signature that wasn't picked up by our previous TypeScript version. This shouldn't affect consumers, because even if they passed a `ColorSpace`, the method would have thrown an error because it's expecting the `ColorSpace.Space` object. The code in this component is only ever checking for string and doesn't account for the Typed part of the signature.

.changeset/card-keyboard-handler-fix.md
caseyisonit: this is the fix you commented on below about adding in the `break`. this is a patch for making sure the keyboard events dont fall through in the switch

1st-gen/packages/card/src/Card.ts
caseyisonit: no this actually ensures the keyboard events handle their flow correctly and dont fall through. all swtich cases should end with a break to signal the end of the case.

.vscode/settings.json
caseyisonit: i can remove the format on save but we should have our default formatter set

tsconfig.base.json
caseyisonit: so had a convo with Cursor to understand your concerns and this is what its recommendation was:

Your change from "node" to "bundler" is safe and appropriate because:
✅ You're building a library with bundlers (Rollup)
✅ Your output format (type: "module") requires ESM-compatible resolution
✅ The published output structure remains unchanged
✅ Modern consumers use "bundler", "node16", or "nodenext" anyway
✅ The exports field provides broad compatibility
Only legacy consumers (TypeScript < 4.7 with moduleResolution: "node") might need to update their config, but they would have needed to do this anyway when consuming modern ESM packages.

Based on this, there should be no breaking changes and shouldnt effect downstream consumers.

.changeset/card-keyboard-handler-fix.md
caseyisonit: i can remove this changeset if this change doesnt constitute a patch.

CONTRIBUTOR-DOCS/README.md
```diff
@@ -26,6 +26,7 @@
     - [Authoring contributor docs](01_contributor-guides/07_authoring-contributor-docs/README.md)
     - [Patching dependencies](01_contributor-guides/08_patching-dependencies.md)
     - [Accessibility testing](01_contributor-guides/09_accessibility-testing.md)
+    - [TypeScript configuration](01_contributor-guides/10_typescript-configuration.md)
```
rise-erpelding: Hey, these links (there's another one in `CONTRIBUTOR-DOCS/01_contributor-guides/README.md` also) don't go to anything, was there an added markdown file for typescript configuration somewhere that it should link to?

_Edit: I see the resolved comment now where you and @rubencarvalho decided to kill this, so it looks like this just needs to be removed!_

caseyisonit:
```suggestion
```

CONTRIBUTOR-DOCS/01_contributor-guides/README.md
```diff
@@ -21,6 +21,7 @@
     - [AI agent instructions](07_authoring-contributor-docs/01_ai-agent-instructions.md)
 - [Patching dependencies](08_patching-dependencies.md)
 - [Accessibility testing](09_accessibility-testing.md)
+- [TypeScript configuration](10_typescript-configuration.md)
```
caseyisonit, line 24:
```suggestion
- [Accessibility testing](09_accessibility-testing.md)

```

cdransf (approved):
It looks like nearly all feedback has been addressed, all of the validation steps check out and this looks great! ✨

rise-erpelding (approved):

1st-gen/packages/card/src/Card.ts
```diff
@@ -147,9 +147,12 @@ export class Card extends LikeAnchor(
                     event.preventDefault();
                     break;
                 }
+                break;
+
```
rubencarvalho, line 151: Can we confirm this `break;` isn't _breaking_ this 😅 ? It seems like the falltrough might have been expected.
It was working in a way where if `toggles` is `true`, Space should toggle selection and **not** click, but if `toggles` is `false`, Space should behave like Enter and click the card.

2nd-gen/packages/core/components/badge/Badge.types.ts
rubencarvalho: Can you confirm if this a compilation error?
The previous change seems valid to me and actually reflects more closely what we want: "this is a read-only config". Otherwise it infers as a normal array (can be mutated).

tsconfig.base.json
rubencarvalho: This _should_ only affect our internal build. I did a quick check because I do see that we're emitting declarations in 1st-gen (`emitDeclarationOnly`) and I saw no difference vs `main`.
Anyway, to be extra-sure, let's make sure to validate with consumers before the next release via a snapshot build 😄

rubencarvalho: Actually, I did find a difference, but only on the generated `.d.ts` for our CSS modules!

E.g., in `progress-circle`:
- src/progress-circle-overrides.css.d.ts 
- src/progress-circle.css.d.ts 
- src/spectrum-progress-circle.css.d.ts 

in this branch:
```ts
declare const styles: import("lit").CSSResult;
export default styles;
```

in `main`:
```ts
declare const styles: import("@spectrum-web-components/base").CSSResult;
export default styles;
```

I can confirm reverting the setting to "node" produces the same result from `main`.
Now... since these CSS modules are only imported internally by our own components (and not a public consumption surface), this shouldn’t affect customers: it doesn’t change the emitted JS, only the type reference used in the declaration file.

1st-gen/packages/card/src/Card.ts
Rajdeepc: I don't think fallthrough was expected since if Space fell through to Enter, you'd get both a synthetic click event and an anchor click. The `toggleSelected()` method explicitly handles the `!toggles` case.

tsconfig.base.json
Rajdeepc: Good findigs here @rubencarvalho  @caseyisonit .  This is a compile time behaviour and i don't think it effects any published runtime artifacts.

2nd-gen/packages/core/components/badge/Badge.types.ts
Rajdeepc: I am not sure if this will throw error during compilation but please confirm. `as const` is already guaranteeing mutability at the type level.

1st-gen/packages/card/src/Card.ts
rubencarvalho: Yes! It does seems like we might have been having a double click here. I know this is out of scope, but it would be really great to have a super small PR applying this fix and covering it with a test.

2nd-gen/packages/core/components/badge/Badge.types.ts
caseyisonit: the readonly is in the wrong place, i would like to push this to the patterns work in code quality to cover how we override and protect types in our components. this is not breaking and wont effect end users. even using the const this is still assert that ElementSize[] is readonly and it is not, this is the only component where this was applied, all other sizeable components follow this current suggested pattern.

PR #5873, opened 2025-11-11
fix(overlay): default listenerHost to target in overlay-trigger-directive

### LINKED ISSUE #5723: [Bug]: listenerHost can be uninitialized in overlay-trigger-directive
### Code of conduct

### Impacted component(s)

overlay-trigger-directive

### Library version

1.7.0

### Expected behavior

No type errors when [this line](https://github.com/adobe/spectrum-web-components/blob/main/packages/overlay/src/slottable-request-directive.ts#L70) is run.

### Actual behavior

We are seeing errors when an overlay trigger directive's reconnected hook runs and there isn't a [listenerHost set in its overridden update()](https://github.com/adobe/spectrum-web-components/blob/main/packages/overlay/src/overlay-trigger-directive.ts#L82). We should potentially set `this.listenerHost = this.target` as its parent class does.

### Screenshots

Image

_No response_

3. Scroll to '....'
4. Check console
5. See error

_No response_

### Severity

None

_No response_

### Would you like to track this issue in Jira?

- [x] Yes, please tell me the ticket number!

TarunAdobe:
## Description

Fixes OverlayTriggerDirective behavior when used with Lit's cache() directive. The directive now properly cleans up overlay state on disconnect and allows fresh overlay creation on reconnect, preventing stale state and runtime errors.

## Motivation and context

When using Lit's cache() directive with overlay trigger elements, the directive's lifecycle behaves differently than with standard rendering:

**Without `cache()`:**
- Elements are completely destroyed and recreated
- The directive instance is created fresh each time
- `reconnected()` never runs on restoration

**With `cache()`:**
- Elements are hidden (kept in memory) instead of destroyed
- When restored, `disconnected()` and `reconnected()` lifecycle hooks fire
- `reconnected()` attempts to reinitialize by calling `init()`

When a cached trigger element was disconnected and reconnected, the overlay instance remained in a stale state. On reconnect, the overlay reference persisted but:
The overlay DOM element was orphaned or in an inconsistent state
The handleOverlayReady callback wouldn't fire again for the stale overlay
Attempting to reuse the old overlay caused errors and prevented the overlay from opening

Solution:

On disconnected():
- Close the overlay to clean up any open state
- Remove the overlay element from the DOM
- Clear the overlay reference in the strategy by setting _overlay = undefined
On reconnected():

- Do nothing - the overlay reference was cleared, so the next click will create a fresh overlay
- The handleOverlayReady callback will fire again with the new overlay instance
- Everything works as if it's the first time opening

This approach ensures that cached trigger elements always work correctly through multiple hide/show cycles. The strategy's persistent click listeners remain active (as intended), but the overlay is freshly created each time it's needed after a disconnect.

## Related issue(s)

-   fixes #5723 

## Manual Testing Steps

- [ ] Open this [story](https://swcpreviews.z13.web.core.windows.net/pr-5873/docs/first-gen-storybook/?path=/story/overlay-directive--managed-overlay-trigger)
- [ ] Click "Toggle Overlay Render Button" multiple times to hide/show the trigger button
- [ ] Verify there are no console errors or warnings
- [ ] Click "Toggle Popover" button to ensure the overlay still functions correctly

## Author's checklist

-   [x] I have read the **[CONTRIBUTING](https://github.com/adobe/spectrum-web-components/blob/main/CONTRIBUTING.md)** and **[PULL_REQUESTS](https://github.com/adobe/spectrum-web-components/blob/main/PULL_REQUESTS.md)** documents.

---

## Reviewer's checklist


1st-gen/packages/overlay/test/overlay-trigger-directive.test.ts
```diff
@@ -88,4 +97,81 @@ describe('Overlay trigger directive', () => {
...
+
+        // Get the reconnected cached trigger button.
+        cachedTrigger = shadowRoot.querySelectorAll('sp-button')[1] as Button;
+        expect(cachedTrigger, 'cached trigger should exist after toggle').to
+            .exist;
+
+        // Now click the cached trigger to open the overlay.
+        const opened = oneEvent(cachedTrigger, 'sp-opened');
+        cachedTrigger.click();
+        await opened;
```
caseyisonit: can we add some assertions after this to confirm the popover is rendered?

1st-gen/packages/overlay/src/overlay-trigger-directive.ts
```diff
@@ -90,6 +90,7 @@ export class OverlayTriggerDirective extends SlottableRequestDirective {
             this.target = part.element as HTMLElement;
             newTarget = true;
         }
+        this.listenerHost = this.target;
```
caseyisonit: are we sure this is the correct default? im trying to understand the override but the strategies and controller complexity is making my brain warp. can we chat about this in our team meeting thursday?

Rajdeepc: The type error occurs because `this.listenerHost` is undefined when `reconnected(`) → `init()` runs before the overlay is ready.
This happens because the parent class sets `this.listenerHost = this.target` in `update()`, child class overrides `update()` but only sets `this.listenerHost` inside the` async handleOverlayReady` callback. If reconnection happens before the overlay exists → `this.listenerHost` is `undefined` and it will error out.

`this.listenerHost = this.target` is not architecturally correct coz the `slottable-request` [event](https://github.com/adobe/spectrum-web-components/blob/edd1ddaf91fd2db2d6bf367bc1f4cee9866f0542/1st-gen/packages/overlay/src/slottable-request-event.ts#L18) is dispatched from the overlay element with `bubbles: false`. 

Attaching the listener to `this.target` (the trigger button) means it will never receive the event since it doesn't bubble. This breaks the lazy content loading mechanism entirely.

We can override the  `reconnected()` in the child class to guard against premature initialization, its cleaner since it's specific to the child's async overlay setup pattern.

```ts
override reconnected(): void {
    // Only call init() if the overlay is ready
    if (this.listenerHost) {
        this.init();
    }
}
```

TarunAdobe: good job finding a solution @Rajdeepc !!! I updated the change.

1st-gen/packages/overlay/test/overlay-trigger-directive.test.ts
caseyisonit: still missing the assertions for popover being open

.changeset/eight-coats-read.md
```diff
@@ -0,0 +1,5 @@
+---
+'@spectrum-web-components/overlay': patch
+---
+
+Sets this.listenerHost = this.target in overlay-trigger-directive’s overridden update() when a listenerHost is not already defined. This prevents type errors during the reconnected flow when listenerHost is required but unset.
```
caseyisonit: this changeset language needs to be updated to reflect the new solution once we figure out what that is

caseyisonit:
@TarunAdobe can you include the manual testing steps in the PR description?

caseyisonit:
When testing overlay from this link: https://swcpreviews.z13.web.core.windows.net/pr-5873/docs/storybook/?path=/story/overlay-directive--default

The overlay does not open and im receiving this in the console: 
Screenshot 2025-11-17 at 2 25 10 PM

TarunAdobe:
> When testing overlay from this link: https://swcpreviews.z13.web.core.windows.net/pr-5873/docs/storybook/?path=/story/overlay-directive--default
> 
> The overlay does not open and im receiving this in the console: Screenshot 2025-11-17 at 2 25 10 PM

WE TOTALLY DIDN'T KNOW BT THIS IS A PROBLEM IN MAIN TOO!!

TarunAdobe:
@caseyisonit  the error you saw in the story would be fixed in a separate pr #5896

TarunAdobe:
> @TarunAdobe can you include the manual testing steps in the PR description?

I updated the Managed overlay trigger story implementation so that it now caches the overlay trigger instead of replacing it and thus you can verify this fix from that story (check description again).

Rajdeepc:
@lehelen19 Could you help validate this in Hz before we move forward with the rollout? It would also be helpful if you could share your specific use case. We can then add it to our Storybook dev instance to reproduce the scenario and run additional tests on our side. This issue appears to be highly use-case specific, so any additional context will help us ensure full coverage.

1st-gen/packages/overlay/test/overlay-trigger-directive.test.ts
Rajdeepc: @TarunAdobe Either you check for the overlay opened or check if the `sp-opened` is dispatched or not. Either of them will suffice.

lehelen19:
The bug was triggering errors logged in Splunk, so I don't have a specific use case to share

Rajdeepc:
> The bug was triggering errors logged in Splunk, so I don't have a specific use case to share

Thanks for the context. @lehelen19 I want to do a soft test here before we roll out this fix. Can you confirm if this fix works for you or not?

lehelen19:
> Thanks for the context. @lehelen19 I want to do a soft test here before we roll out this fix. Can you confirm if this fix works for you or not?

I don't have a specific test case for this issue. You might be able to remove a component with the overlay directive and then reconnect it to see if you are encountering the error in the console?

We should be able to see the related Splunk error disappear once we get the fix in main

caseyisonit (changes requested):
The story set up is confusing for me on what its actually doing and the differences between the two. Can we set this story up to be more reflective of an actual pattern that would exist in product? can the labels on the buttons be improved for understanding? the popover should also not include a heading, thats not a realistic use case. 

Also the documentation markdown needs to be updated to reflect how and when to use caching for code snippets provided. The caching also requires the lit directive package and should be called out in the docs as well.

TarunAdobe:
> The story set up is confusing for me on what its actually doing and the differences between the two. Can we set this story up to be more reflective of an actual pattern that would exist in product? can the labels on the buttons be improved for understanding? the popover should also not include a heading, thats not a realistic use case.

I have updated the story with a better usecase that is closer to the product requirement. Please check the latest update.

> Also the documentation markdown needs to be updated to reflect how and when to use caching for code snippets provided. The caching also requires the lit directive package and should be called out in the docs as well.

This is a bit nuanced and I don't fully agree if we need to add anything to the documentation as this is a certain pattern that is being used in a certain app. My expectation as a user is that overlay-directive would work for most / all common patterns and we don't all possible usescases of our components either way.

rubencarvalho:
I still see some weirdness 

Uploading Screen Recording 2026-01-08 at 11.39.19.mov…

1st-gen/packages/overlay/stories/overlay-directive.stories.ts
```diff
@@ -290,38 +292,53 @@ insertionOptions.parameters = {
...
-            </sp-button>
+                <p
+                    style="max-width: 600px; color: var(--spectrum-neutral-content-color-default); font-size: 14px; margin: 0;"
+                >
+                    The quick actions button below is wrapped with Lit's
+                    <code>cache()</code>
+                    directive, which preserves the DOM and component state when
+                    toggled off instead of destroying it. This tests that the
+                    overlay trigger properly handles reconnection without
+                    errors.
```
caseyisonit, line 326: Thank you for this update to this story! its a great example now!

1st-gen/packages/overlay/src/overlay-trigger-directive.ts
```diff
@@ -135,6 +135,35 @@ export class OverlayTriggerDirective extends SlottableRequestDirective {
...
+        // Without this cleanup, the old overlay instance would be reused but in a
+        // broken state where handleOverlayReady never fires again.
+        if (this.overlay) {
+            this.overlay.open = false;
+            this.overlay.remove();
+        }
+        // Clear the overlay reference in the strategy so a new one will be created
+        // on the next open, which will trigger handleOverlayReady again.
+        if (this.strategy) {
+            (this.strategy as unknown as { _overlay: undefined })._overlay =
```
caseyisonit: This is overriding a private property in the interaction controller that is typed as `AbstractController`, missing `undefined` thus the need of the typescript. 

There is a public `overlay` property that we can tap into and modify the first return in the setter to `_overlay= undefined` and add `undefined` as an accepted type value to the private property. i believe the getter might also need to accept `undefined` as well.

1st-gen/packages/overlay/src/overlay-trigger-directive.ts
```diff
@@ -136,13 +136,33 @@ export class OverlayTriggerDirective extends SlottableRequestDirective {
...
```
rubencarvalho: Could we consider properly typing `_overlay` as optional to avoid the cast?
Change `private _overlay!: AbstractOverlay;` to `private _overlay?: AbstractOverlay` and update the getter.
And if we really want to be idiomatic, we should add a `public clearOverlay()` method:

```ts
public clearOverlay(): void {
    if (this._overlay) {
        this._overlay.removeController(this);
    }
    this._overlay = undefined;
}
```
and then here we instead just call `this.strategy?.clearOverlay();`. It seems to me like a cleaner approach because it keeps the cleanup logic encapsulated in the controller (where it can also call `removeController()`).

1st-gen/packages/overlay/src/overlay-trigger-directive.ts
```diff
@@ -136,13 +136,33 @@ export class OverlayTriggerDirective extends SlottableRequestDirective {
...
+
+    override reconnected(): void {
+        // If overlay was never created (user never opened it), there's nothing to reconnect.
+        // The overlay and listenerHost are only set when handleOverlayReady fires,
+        // which happens on first open. Without this guard, init() would fail
+        // trying to add a listener to undefined listenerHost.
+        //
+        // If overlay WAS created before disconnect, we cleared the references in disconnected(),
+        // so this.overlay will be undefined and we'll skip init(). A fresh overlay will be
+        // created on the next open via handleOverlayReady.
```
rubencarvalho, line 164: Could we maybe can modify this / add a sentence to call out the intention of the empty reconnected override?

IDK, something along the lines of
```ts
// Intentionally empty: don't call init() since the overlay reference was cleared in disconnected() and will be recreated on next open.
```

1st-gen/packages/overlay/test/overlay-trigger-directive.test.ts
rubencarvalho: I unresolved the thread because I still don't see the assertion here 😅

1st-gen/packages/overlay/src/overlay-trigger-directive.ts
caseyisonit: lol i like this approach over my comment, we were reviewing at the same time

1st-gen/packages/overlay/src/HoverController.ts
```diff
@@ -143,11 +144,13 @@ export class HoverController extends InteractionController {
     }
 
     private prepareContentRelativeDescription(): void {
+        if (!this.overlay) return;
+        const overlay = this.overlay; // Capture for closure
```
rubencarvalho, line 149: Nice! Easy to miss

TarunAdobe: yus the linter cried otherwise i had missed it

Rajdeepc:
@caseyisonit If you get some time today can you please review this. We would like to get this out for PsWeb and give them a alpha release to test

PR #5898, opened 2025-11-18
feat: style field label mixin for textfield, textarea

rise-erpelding:
## Description

This PR updates the new `FieldLabelMixin` to use new styles.

The CSS styling for the slotted label in the shadow DOM now matches the appearance of the standalone `<sp-field-label>` component for textfield and textarea components.

All textfield and textarea VRTs are now passing, confirming visual parity between slotted labels and standalone field label components.

## Motivation and context

Styles the form field mixin to look like `sp-field-label`. Previously, the FormFieldMixin component was created to standardize form field behavior across Spectrum Web Components.

### Issue 1: Field label wrapping

This does change the behavior of the field label. Before, `sp-field-label` was independent and its width was bound to the width of its container, so it could extend beyond the width of the textfield input. You can see this behavior in the [Textfield Grows Only story](https://opensource.adobe.com/spectrum-web-components/storybook/index.html?path=/story/textfield--grows-only):

image

Now, the label's container is `sp-textfield` and therefore its width is restricted to the width of the textfield by default, forcing a longer label to wrap:

image

This is probably the ideal behavior because it's seen in [CSS](https://opensource.adobe.com/spectrum-css/?path=/story/components-text-field--default&args=labelText:Username+that+is+very+long+and+wraps+to+the+next+line+and+is+the+same+width+as+the+input), React, Figma, and the [design guidelines](https://spectrum.adobe.com/page/text-field/#Text-overflow).

It might also be worth noting that help text also behaves this way, where its width is restricted to the width of the input.

If we're worried about how restricting field label's width affects consumers, there are a few possible options that I see to get around this behavior:
- (This one has been implemented for now:) **Offer `--mod-field-label-width`** to allow a way to modify the default width. (Is this a great time to introduce another mod though?) This is what the Textfield Grows Only story is doing by setting the field label width to 400px so it won't wrap in VRTs, but I'm not sure there's a way to scope the width to be 100% of textfield's container. Also worth noting that some of the large VRTs aren't passing because of the same wrapping issue.
- **We _could_ provide a `--mod` to adjust the textfield input instead.** A user could adjust the width of `sp-textfield` with the existing `--mod-textfield-width`. As an example, let's say they want it to be 400px. This expands the width of the field label, input, and help text to 400px. If a user wanted the input to be smaller, we could add a mod that only affects the input, like `--mod-textfield-input-width`, but the help text would be longer than the input, so we'd want a mod for the help text width as well. This feels a bit messy.
- We could **absolutely position the label** and adjust textfield styling around it, but that feels precarious to me and seems like it could potentially create many bugs. I'm happy to look into it if we decide this is a good direction though!

### Issue 2: Side label styling

We're not currently using the mixin with any textfield variants that use the side label, but these would break. The spectrum-css styles for textfield are designed to work with a textfield that includes field label and help text, so bringing some of those in seems like the best approach. No blockers here currently, but noting that this variant won't work in this branch as it is right now.

### Changes

**Architectural CSS changes:**
- Created new `field-label-mixin.css` file that adapts field label styles for use in other components' shadow DOM
- Custom property definitions remain on `:host` (referencing the host component's `size`, `disabled`, etc. attributes)
- Style applications target the `label` selector instead of `:host`, allowing the mixin to style `<label>` elements within other components
- This contrasts with `spectrum-field-label.css` which applies styles to `:host` for the standalone `<sp-field-label>` component
- Both files share the same custom property tokens, ensuring visual parity

## Related issue(s)

-   fixes SWC-1316

---

## Author's checklist

---

## Reviewer's checklist

### Manual review test cases

    1. Go [here](url)
    2. Do this action
    3. Expect this result

    1. Go [here](url)
    2. Do this action
    3. Expect this result

### Device review

### DIFF
--- a/1st-gen/packages/field-label/src/field-label-mixin.css
+++ b/1st-gen/packages/field-label/src/field-label-mixin.css
@@ -98,10 +98,15 @@ label:lang(zh) {
 :host([side-aligned="start"]),
 :host([side-aligned="end"]) {
     column-gap: var(--mod-fieldlabel-side-padding-right, var(--spectrum-field-label-side-padding-right));
-    row-gap: calc(var(--swc-scale-factor) * 12px); /* why? idk, this is how westbrook had it in the stories file */
+    margin-block-end: calc(var(--swc-scale-factor) * 12px); /* mimics row-gap westbrook had in the stories file */
     grid-template-rows: auto auto; /* 2 rows instead of 3 */
-    grid-template-columns: 1fr 2fr; /* questioning if we want 3 columns, that's how css is */
-    inline-size: 100%; /* we need to override the other inline-size */
+    grid-template-columns: 1fr 2fr auto; /* still haven't figured out what the 3rd column is for */
+    inline-size: 100%;
+}
+
+:host([side-aligned="start"]) input,
+:host([side-aligned="end"]) input {
+    inline-size: 240px;
 }
 
 :host([side-aligned="start"]) label,
@@ -109,6 +114,7 @@ label:lang(zh) {
     vertical-align: top;
     margin-block-start: var(--mod-fieldlabel-side-margin-block-start, var(--spectrum-field-label-side-margin-block-start));
     margin-block-end: 0;
+
     /* margin-inline-end is set in the host's column-gap property */
     display: inline-block;
 
@@ -119,6 +125,11 @@ label:lang(zh) {
 
 :host([side-aligned="end"]) label {
     text-align: end;
+}
+
+:host([side-aligned="start"]) ::slotted(sp-help-text),
+:host([side-aligned="end"]) ::slotted(sp-help-text) {
+    grid-area: 2 / 2 / auto / span 1;
 }
 
 @media (forced-colors: active) {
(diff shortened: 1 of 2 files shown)


1st-gen/packages/textfield/stories/textfield.stories.ts
```diff
@@ -66,6 +66,7 @@ export const growsOnly = (): TemplateResult => {
             grows
             id="grows-only"
             placeholder="Does not grow or display incorrectly"
+            style="--mod-field-label-width: 400px;"
```
rise-erpelding, line 69: Is this cheating to make the VRTs pass for this story? Yes. 😆 

After poking around at [CSS](https://opensource.adobe.com/spectrum-css/?path=/story/components-text-field--default&args=labelText:Username+that+is+very+long+and+wraps+to+the+next+line+and+is+the+same+width+as+the+input), React, Figma, and the [design guidelines](https://spectrum.adobe.com/page/text-field/#Text-overflow), I think the expected behavior for the field label is that its width is the same as the text input rather than being completely independent of it.

However, I realize this could potentially cause problems for consumers who expected the field label to behave the other way, so we can optionally provide a way to get around this, if we want.

Happy to have feedback on this.

rise-erpelding:
closing in favor of #5962

Issue #5628, opened 2025-07-22
[Bug]: Modal overlays do not prevent clicks on external elements


lehelen19:
### Code of conduct

- [x] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-overlay

### Library version

1.7.0

### Expected behavior

When a modal overlay is open, we should not be able to click on an element outside of the overlay without first exiting the overlay.

### Actual behavior

When a modal overlay is open, we are able to click on an element outside of the overlay.

### How can we reproduce this issue?

1.4.0: https://stackblitz.com/edit/vitejs-vite-x96vj1wk?file=package.json,src%2Fmy-element.ts

1. Click on the "Open overlay" button to open a modal overlay
2. Hover over the "External" button and notice how the cursor does not change and how you can't click on it immediately

1.7.0: https://stackblitz.com/edit/vitejs-vite-nvgyzymb?file=package.json,src%2Fmy-element.ts
Observe how you can click on the "External" button even when the modal overlay is open.

### Severity

SEV 3

### Would you like to track this issue in Jira?

- [x] Yes, please tell me the ticket number!

rubencarvalho:
Jira ticket created: SWC-1010

Rajdeepc:
@lehelen19 Can you please confirm if the updated [fix](https://swcpreviews.z13.web.core.windows.net/pr-5907/docs/first-gen-storybook/?path=/story/overlay-element--modal-click-blocking) works for you or not!

lehelen19:
@Rajdeepc The environment looks good to me. Thanks for the fix!

PR #5907, opened 2025-11-24
fix(overlay): click-blocking behavior for type="modal" and "page" overlays

Rajdeepc:
## Description

This PR restores the click-blocking behavior for modal and page overlays that was lost when migrating from `dialog.showModal()` to `dialog.showPopover()` for performance reasons. The fix manually implements the click-blocking functionality that `showModal()` provided automatically, ensuring that users cannot interact with elements outside of an open modal overlay.

**Key changes:**
- Added event handlers (`handlePointerdown` and `handleClick`) that intercept pointer and click events in the capture phase
- Implemented logic to detect if clicks/pointer events originate inside modal overlay dialogs using `event.composedPath()`
- Block external clicks by calling `preventDefault()`, `stopPropagation()`, and `stopImmediatePropagation()` when clicks occur outside modal overlays
- Added a transparent backdrop element that helps catch clicks outside modal dialogs
- Refactored duplicate logic into reusable helper methods (`getModalOverlays()` and `isEventInsideModal()`) for better maintainability

## Motivation and context

In version 1.7.0, the overlay implementation migrated from using `dialog.showModal()` to `dialog.showPopover()` for performance improvements. However, `showModal()` provided native click-blocking behavior that prevented users from clicking elements outside the modal overlay, while `showPopover()` does not provide this behavior.

This created a regression where modal overlays no longer blocked external clicks, allowing users to interact with elements behind the modal overlay. This violates accessibility best practices and user expectations for modal dialogs.

**Expected behavior:** When a modal overlay is open, users should not be able to click on elements outside of the overlay without first closing the overlay.

**Previous behavior (1.4.0):** Modal overlays correctly blocked external clicks using `showModal()`.

**Current behavior (1.7.0+):** Modal overlays allowed external clicks, breaking the modal interaction pattern.

This fix restores the expected behavior while maintaining the performance benefits of using `showPopover()`.

## Related issue(s)

- fixes https://github.com/adobe/spectrum-web-components/issues/5628
- fixes SWC-1017

**Before (broken behavior):**
- External button remains clickable when modal overlay is open
- Cursor changes to pointer when hovering over external elements
- Click events reach external elements despite modal being open

DEMO: https://stackblitz.com/edit/vitejs-vite-nvgyzymb?file=package.json,src%2Fmy-element.ts

**After (fixed behavior):**
- External button is not clickable when modal overlay is open
- Cursor does not change when hovering over external elements
- Click events are blocked from reaching external elements

DEMO: https://swcpreviews.z13.web.core.windows.net/pr-5907/docs/first-gen-storybook/?path=/story/overlay-element--modal-click-blocking

### Event handling strategy

The fix uses a two-pronged approach:

1. **Capture phase interception**: Event listeners are attached with `capture: true` to intercept events before they reach their target elements
2. **Path-based detection**: Uses `event.composedPath()` to determine if the event originated inside a modal overlay dialog

### Code structure

- **`getModalOverlays()`**: Helper method that filters the overlay stack to get only open modal/page overlays
- **`isEventInsideModal()`**: Core logic that checks if an event path intersects with any modal overlay dialog
- **`handlePointerdown()`**: Intercepts pointerdown events to block interactions early in the event chain
- **`handleClick()`**: Intercepts click events as a secondary layer of protection
- **`manageModalBackdrop()`**: Creates/removes a transparent backdrop element to help catch external clicks

### Browser compatibility

The implementation uses standard DOM APIs (`composedPath()`, `contains()`, `preventDefault()`, etc.) that are supported across all modern browsers. The fix has been tested and verified to work in:
- Chromium-based browsers (Chrome, Edge)
- Firefox
- WebKit-based browsers (Safari)

## Author's checklist

---

## Reviewer's checklist

### Manual review test cases

-   [x] **Modal overlay click blocking**

    1. Navigate to the Storybook story `overlay-element.stories.ts` → `modalClickBlocking`
    2. Click the "Open overlay" button to open a modal overlay
    3. Try to click the "External" button below the modal
    4. **Expected**: The external button should NOT be clickable. No alert should appear. The cursor should not change to a pointer when hovering over the external button.
    5. Click the button inside the modal overlay
    6. **Expected**: The internal button should work correctly and show an alert
    7. Close the modal overlay (press ESC or click outside if `allow-outside-click` is enabled)
    8. Try clicking the "External" button again
    9. **Expected**: The external button should now be clickable and show an alert

-   [x] **Nested modal overlays**

    1. Navigate to the Storybook story `overlay-element.stories.ts` → `nestedModalOverlays`
    2. Open the outer modal overlay
    3. Open the inner modal overlay
    4. Try clicking elements outside both modals
    5. **Expected**: External clicks should be blocked
    6. Close the inner modal, then try clicking outside
    7. **Expected**: External clicks should still be blocked (outer modal is still open)
    8. Close the outer modal
    9. **Expected**: External elements should now be clickable

-   [x] **Page overlays**

    1. Test with `type="page"` overlays
    2. **Expected**: Page overlays should also block external clicks (same behavior as modal overlays)

-   [x] **Non-modal overlays**

    1. Test with `type="auto"`, `type="hint"`, and `type="manual"` overlays
    2. **Expected**: These overlay types should NOT block external clicks (only modal and page types should block)

-   [x] **Keyboard navigation**

    1. Open a modal overlay
    2. Use Tab key to navigate focus
    3. **Expected**: Focus should be trapped within the modal (existing behavior, not changed by this PR)
    4. Press ESC to close
    5. **Expected**: Modal should close and focus should return appropriately

### Device review

### Automated tests

All existing overlay tests pass (191 tests across Firefox, Chromium, and Webkit):
- `yarn test:focus overlay` - All tests passing ✅

## Breaking changes

None. This is a bug fix that restores expected behavior.

## Additional notes

- The implementation maintains backward compatibility with existing overlay usage
- The fix does not affect non-modal overlay types (auto, hint, manual)
- Focus trapping behavior (existing feature) remains unchanged
- Body scroll blocking (existing feature) remains unchanged

### DIFF
--- a/1st-gen/projects/documentation/e2e/published.spec.ts
+++ b/1st-gen/projects/documentation/e2e/published.spec.ts
@@ -41,12 +41,10 @@ test.describe('search and go', () => {
         await page.waitForTimeout(500); // Allow search to process
         await expect(menuItem).toBeVisible({ timeout: 10000 });
 
-        // Wait for menu item to be fully interactive
-        await page.waitForTimeout(200);
-        await page.keyboard.press('ArrowDown');
-
-        await expect(menuItem).toBeFocused({ timeout: 5000 });
-        await page.keyboard.press('Enter');
+        // Click directly on the target menu item since ArrowDown only moves
+        // to the first result, which may not be the item we're looking for
+        // when multiple matches exist (e.g., searching "base" returns many results)
+        await menuItem.click();
 
         // Wait for navigation to complete
         await page.waitForLoadState('networkidle', { timeout: 15000 });
--- a/1st-gen/tools/bundle/elements.ts
+++ b/1st-gen/tools/bundle/elements.ts
@@ -98,4 +98,5 @@ import '@spectrum-web-components/tooltip/sp-tooltip.js';
 import '@spectrum-web-components/top-nav/sp-top-nav.js';
 import '@spectrum-web-components/top-nav/sp-top-nav-item.js';
 import '@spectrum-web-components/tray/sp-tray.js';
+import '@spectrum-web-components/truncated/sp-truncated.js';
 import '@spectrum-web-components/underlay/sp-underlay.js';
--- a/1st-gen/tools/truncated/README.md
+++ b/1st-gen/tools/truncated/README.md
@@ -1,33 +1,33 @@
-## Description
+## Overview
 
-`<sp-truncated>` renders a line of text, truncating it if it overflows its container. When overflowing, a tooltip is automatically created
-that renders the entire non-truncated content.
+`<sp-truncated>` renders a line of text, truncating it if it overflows its container. When overflowing, a tooltip is automatically created that renders the entire non-truncated content. Additionally, clicking on overflowing text copies the full content to the clipboard and displays a success message.
 
-It is used like a `<span>` to contain potentially-long strings that are important for users to see, even when in small containers, like full
-names and email addresses.
+This component is designed like a `<span>` to contain potentially-long strings that are important for users to see, even when in small containers, such as full names, email addresses, file paths, and other text data that may exceed the available space.
 
 ### Usage
 
 [![See it on NPM!](https://img.shields.io/npm/v/@spectrum-web-components/truncated?style=for-the-badge)](https://www.npmjs.com/package/@spectrum-web-components/truncated)
 [![How big is this package in your project?](https://img.shields.io/bundlephobia/minzip/@spectrum-web-components/truncated?style=for-the-badge)](https://bundlephobia.com/result?p=@spectrum-web-components/truncated)
 
-```
+```bash
 yarn add @spectrum-web-components/truncated
 ```
 
-Import the side effectful registration of `<sp-truncated>` via:
+Import the side effectful registration of `<sp-truncated>` as follows:
 
-```
+```typescript
 import '@spectrum-web-components/truncated/sp-truncated.js';
 ```
 
 When looking to leverage the `Truncated` base class as a type and/or for extension purposes, do so via:
 
-```
+```typescript
 import { Truncated } from '@spectrum-web-components/truncated';
 ```
 
-## Example
+### Anatomy
+
+The basic structure of the truncated component:
 
 ```html
 <sp-truncated>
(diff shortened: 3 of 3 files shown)


TarunAdobe:
I have a question for my understanding... If I navigate to [overlay elements - nested modal overlays story](https://swcpreviews.z13.web.core.windows.net/pr-5907/docs/first-gen-storybook/?path=/story/overlay-element--nested-modal-overlays) and open both outer and inner overlay and then click outside in the empty space, both the overlays close. Is that the expected behaviour?

nikkimk (approved):
Thanks for the detailed PR instructions. It made testing super easy. 

The small nit with [this story](https://swcpreviews.z13.web.core.windows.net/pr-5907/docs/first-gen-storybook/?path=/story/picker--picker-in-modal-overlay) being named `modal` when the type is `auto`.

Rajdeepc:
@rubencarvalho Let me know if I have missed anything on this pattern?

1st-gen/packages/overlay/src/OverlayStack.ts
```diff
@@ -105,13 +113,166 @@ class OverlayStack {
...
         this.lastOverlay = this.stack[this.stack.length - 1];
+
+        // Check for modal overlays and prevent pointerdown outside them
+        // This ensures we block the interaction before click handlers can run
+        if (!this.stack.length) return;
+
+        const modalOverlays = this.getModalOverlays();
+        if (!modalOverlays.length) return;
+
+        const pointerPath = event.composedPath();
```
rubencarvalho, line 170: We already have `this.pointerdownPath = event.composedPath();` above. But I guess if we switch up the order of operations this line may still be needed.

1st-gen/packages/overlay/src/OverlayStack.ts
```diff
@@ -105,13 +113,166 @@ class OverlayStack {
...
+    }
+
+    /**
+     * Cache the `pointerdownTarget` for later testing and prevent clicks outside modal overlays
+     *
+     * @param event {PointerEvent}
      */
     handlePointerdown = (event: Event): void => {
         this.pointerdownPath = event.composedPath();
         this.lastOverlay = this.stack[this.stack.length - 1];
```
rubencarvalho: I think there's an issue here... 🤔 
`this.pointerdownPath` is "cached" before checking if the event should be blocked. When a user clicks outside a modal, the event is correctly blocked, but the path from that blocked interaction is still stored...
If I look down this file, `handlePointerup` runs and it uses this stale path to determine which overlays to close. Won't this eventually close wrong modals?

I guess a simple fix would be to move these lines after the check, so that the path and last overlay are only cached for valid (non-blocked) pointer interactions.

1st-gen/packages/overlay/stories/overlay-element.stories.ts
```diff
@@ -980,3 +981,53 @@ nestedModalOverlays.swc_vrt = {
...
+
+modalClickBlocking.swc_vrt = {
+    skip: true,
+};
+
+modalClickBlocking.parameters = {
+    tags: ['!dev'],
+};
+
+modalClickBlocking.parameters = {
```
rubencarvalho: We're defining the same `.parameters` property twice, the first one will get invalidated.

1st-gen/packages/overlay/src/OverlayStack.ts
```diff
@@ -105,13 +113,166 @@ class OverlayStack {
...
+        }
+        return false;
+    }
+
+    /**
+     * Manage backdrop element for modal/page overlays to block clicks outside.
+     * The backdrop catches all pointer events, and we check if they're inside
+     * the modal dialog before allowing them through.
+     */
+    private manageModalBackdrop(): void {
```
rubencarvalho: I'm curious about the role of the modal backdrop here... Since `handlePointerdown` and `handleClick` are now both using `capture: true` on the document, they should intercept events before they reach any target element. The `isEventInsideModal` check would then block any clicks outside the modal regardless of what's underneath.
Is the backdrop providing something the capturing event listeners don't cover? I might be missing an edge case, but it seems like we could potentially simplify by removing the backdrop and `manageModalBackdrop()` entirely. Would spare us from this DOM manipulation.
Happy to be wrong here, just want to make sure I understand the reasoning! 😛

1st-gen/packages/overlay/test/overlay.test.ts
```diff
@@ -831,6 +831,78 @@ describe('Overlay - type="modal"', () => {
...
+
+        // Open modal overlay
+        const opened = oneEvent(overlay, 'sp-opened');
+        trigger.click();
+        await opened;
+
+        expect(overlay.open).to.be.true;
+
+        // Try to click external button - should be blocked
+        externalButton.click();
```
rubencarvalho, line 880: nit: we're using a programmatic click here. I think it would be nicer to use `sendMouse`. If I understand this correctly, as it currently is, it will bypass `pointerdown`/`pointerup` and only fire the `click` event. Since the blocking logic hooks into both `handlePointerdown` and `handleClick`, I think using `sendMouse` would better reflect a real user interaction here

rubencarvalho (changes requested):
Overall approach looks great, I think we can just double-check a couple of small details. Let me know if you have any questions! 😄

1st-gen/packages/overlay/src/OverlayStack.ts
Rajdeepc: Yes safe to reuse. My bad. We are in the same function call and it is the same event object returning the same path.

1st-gen/packages/overlay/src/OverlayStack.ts
Rajdeepc: Fair point! You're right, and I don't have a strong defense here. The original caching logic predates the modal-blocking feature. When I added the blocking behavior, I inserted the check after the existing cache assignment without considering how it would affect downstream consumers like `handlePointerup`. In hindsight, this should have been restructured the flow when adding the blocking logic.
Good catch. Fixed now by only caching for valid interactions.

1st-gen/packages/overlay/stories/overlay-element.stories.ts
Rajdeepc: Copy-paste error on my part :(

1st-gen/packages/overlay/src/OverlayStack.ts
Rajdeepc: Tested removing the backdrop—clicks on external elements are no longer getting blocked which opens up the bug. The capture phase listeners with `isEventInsideModal()` alone aren't sufficient. The backdrop is needed because it provides a physical barrier that intercepts pointer events before they reach external elements. While I was expecting the capture phase handlers to handle everything, there's an edge case (possibly related to top-layer behavior or inline handlers) where events slip through without the backdrop.
Keeping the backdrop for now. We can investigate further if we want to eliminate the DOM manipulation, but the current approach works reliably.

1st-gen/packages/overlay/test/overlay.test.ts
Rajdeepc: Good suggestion on using `sendMouse`, but it has issues with top layer elements in the test environment. Sticking with `.click()` but did some small robust changes to the test. Added the event listeners explicitly and querying the internal button from the popover after it opens in the top layer which gets the element in its actual rendered position (top layer).

1st-gen/packages/overlay/src/OverlayStack.ts
```diff
@@ -105,15 +110,117 @@ class OverlayStack {
...
+     *
+     * @param event {PointerEvent}
      */
     handlePointerdown = (event: Event): void => {
-        this.pointerdownPath = event.composedPath();
+        // Cache pointerdown path and last overlay for valid interactions (used by handlePointerup)
+        // Also block pointerdown outside modal overlays to prevent click handlers from running
+        if (!this.stack.length) {
+            this.pointerdownPath = event.composedPath();
+            this.lastOverlay = this.stack[this.stack.length - 1];
```
rubencarvalho: If the stack is empty, the condition `!this.stack.length` is `true`
But then `this.stack[this.stack.length - 1]` will be `this.stack[-1] = undefined`.

If the stack is empty there's nothing to manage so we can just return early.

```suggestion
            this.pointerdownPath = event.composedPath();
            this.lastOverlay = this.stack[this.stack.length - 1];
```

rubencarvalho (changes requested):

1st-gen/packages/overlay/stories/overlay-element.stories.ts
```diff
@@ -980,3 +981,50 @@ nestedModalOverlays.swc_vrt = {
...
+        </sp-button>
+        <p style="margin-top: 10px; color: red;">
+            ⚠️ When the modal is open, clicking "External" should be blocked. If
+            you can click it and see the alert, the fix is not working.
+        </p>
+    </div>
+`;
+
+modalClickBlocking.swc_vrt = {
+    skip: true,
```
caseyisonit, line 1024: ❤️

PR #5911, opened 2025-11-27
chore: add storybook-screen-reader plugin

TarunAdobe:
A Storybook addon that simulates screen reader behavior, providing instant audio and visual feedback as you navigate components.

### What it does
Voice Reader: Speaks announcements using Web Speech API
Text Reader: Displays announcements in the addon panel
Tab through your component → hear/see what a screen reader would announce

### How it works
Tracks focus via focusin events
Computes accessible names using dom-accessibility-api (W3C AccName spec)
Maps elements to ARIA roles via aria-query
Announces: role, name, states (checked, expanded, disabled, etc.)

### Implementation
This addon is a shared local package at `2nd-gen/packages/swc/.storybook/addons/screen-reader-addon/`:
- Written in TypeScript `(.ts/.tsx)`
- No build step: Storybook's internal `esbuild` compiles source directly
- Uses Spectrum Web Components: `sp-switch`, `sp-theme`, `sp-textfield`, `sp-help-text`, `sp-field-label`
- Shared across both Storybooks: Used by `1st-gen` and `2nd-gen`
- Theme-aware: Automatically matches Storybook UI theme (light/dark) via luminance detection
- Dependencies: Bundled packages (`aria-query`, `dom-accessibility-api`, `query-selector-shadow-dom`) are installed in the parent package

### Limitations
This is a development aid, not a replacement for real screen reader testing.
Good for: Quick feedback on accessible names, roles, and states
Not for: Final accessibility sign-off (use VoiceOver/NVDA for that)

### Further Reading
See [RFC](https://paper.dropbox.com/doc/Storybook-Screen-Reader-Addon--Cy14IrjIwJMpElRZ~JU4TGYPAg-Cfh4q0XpdU0MEUdU9UGW7) for discussion on ownership, maintenance, and where this should live long-term.

### Testing

First gen: 

Go [here](https://swcpreviews.z13.web.core.windows.net/pr-5911/docs/first-gen-storybook/?path=/story/accordion--open)
-  Click on Screen Reader addon
- Toggle Voice Reader / Text Reader
- Use TAB and ARROW keys to navigate the focus on the component and listen to & read the screen reader annoucements

Note: Check the same storybook screen reader addon [here](https://swcpreviews.z13.web.core.windows.net/pr-5911/docs/second-gen-storybook/?path=/story/components-badge--default&args=dir:!null) in second-gen though it's impossible to verify if it works the same cz second-gen has no focusable component :(


yarn.lock
```diff
@@ -28214,6 +28332,22 @@ __metadata:
   languageName: node
   linkType: hard
 
+"storybook-screen-reader@npm:1.1.0":
+  version: 1.1.0
+  resolution: "storybook-screen-reader@npm:1.1.0"
+  dependencies:
```
caseyisonit: i would want you to lock the versions, and why are we using pre v1 versions of SWC?

2nd-gen/packages/swc/package.json
```diff
@@ -68,11 +68,14 @@
...
         "@vitest/ui": "3.2.4",
         "@wc-toolkit/storybook-helpers": "10.0.0",
+        "aria-query": "^5.3.0",
         "autoprefixer": "10.4.21",
+        "dom-accessibility-api": "^0.7.0",
         "glob": "11.0.3",
         "playwright": "1.53.1",
         "postcss": "8.5.6",
         "postcss-preset-env": "10.4.0",
+        "query-selector-shadow-dom": "^1.0.0",
```
rubencarvalho: Please lock all dependency versions 😄

2nd-gen/packages/swc/.storybook/addons/screen-reader-addon/src/screen-reader/screenReader.ts
```diff
@@ -0,0 +1,970 @@
...
+            this.updateFocusIndicator(currentFocus);
+            this.announceElement(currentFocus);
+        }
+    }
+
+    /**
+     * Stop the screen reader
+     */
+    stop(): void {
+        if (!this.isRunning && !this.storyDocument) {
```
rubencarvalho: Is it possible you mean `OR`, not `AND`?
```suggestion
        if (!this.isRunning || !this.storyDocument) {
```

rubencarvalho: and nit: if you do change this... we need a `storyDocument` cleanup at the end of this stop method.
The `storyDocument` reference is never set to `null`, which could prevent garbage collection 🤔

2nd-gen/packages/swc/.storybook/addons/screen-reader-addon/src/screen-reader/screenReader.ts
```diff
@@ -0,0 +1,970 @@
...
+     * Build announcement string based on role
+     */
+    buildRoleAnnouncement(
+        element: Element,
+        role: string,
+        name: string
+    ): string {
+        const announcements: Record<string, RoleAnnouncementFn> = {
+            link: () => {
+                const visited = element.matches(':visited') ? 'visited ' : '';
```
rubencarvalho: Eheh, this is ingenious, but I think it does not work... Browsers are [super aggressive with their privacy](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:visited#:~:text=Although%20these%20styles%20can%20change%20the%20appearance%20of%20colors%20to%20the%20end%20user%2C%20the%20window.getComputedStyle%20method%20will%20lie%20and%20always%20return%20the%20value%20of%20the%20non%2D%3Avisited%20color.) so this will always return `false`.

2nd-gen/packages/swc/.storybook/addons/screen-reader-addon/src/screen-reader/screenReader.ts
```diff
@@ -0,0 +1,970 @@
...
+            return computeAccessibleDescription(element);
+        } catch {
+            return '';
+        }
+    }
+
+    /**
+     * Generate announcement for an element based on its role and state
+     */
+    announceElement(element: Element | null): void {
```
rubencarvalho: Have we considered adding an `aria-hidden` check to skip elements hidden from the accessibility tree? If I understand this correctly we are now announcing them?

2nd-gen/packages/swc/.storybook/addons/screen-reader-addon/src/screen-reader/screenReader.ts
```diff
@@ -0,0 +1,970 @@
...
+
+    /**
+     * Build announcement string based on role
+     */
+    buildRoleAnnouncement(
+        element: Element,
+        role: string,
+        name: string
+    ): string {
+        const announcements: Record<string, RoleAnnouncementFn> = {
```
rubencarvalho: What about the `disabled` state? Could we append something like `, disabled` if it has `aria-disabled` (I guess if it has `disabled` attribute it is probably not focusable so we don't need to worry about that part? maybe confirm)?

2nd-gen/packages/swc/.storybook/addons/screen-reader-addon/src/screen-reader/screenReader.ts
```diff
@@ -0,0 +1,970 @@
...
+        } catch {
+            return '';
+        }
+    }
+
+    /**
+     * Generate announcement for an element based on its role and state
+     */
+    announceElement(element: Element | null): void {
+        if (!element || element === this.lastAnnouncedElement) {
```
rubencarvalho: Hmmm... State changes on the focused element aren't announced. For example, if a checkbox is toggled, it's the same element, so we're skipping it here.
We could add something along the lines of an `announceStateChange(element, attribute)` method that announces only the new state.
We'd need to update `onMutation` to call it for things like `aria-checked`, `aria-expanded`, `aria-pressed` (can't remember any other) changes on the focused element. BTW, I see that currently only `aria-selected` triggers a re-announcement (line 741).

2nd-gen/packages/swc/.storybook/addons/screen-reader-addon/src/screen-reader/screenReader.ts
TarunAdobe: did not know!!!! :(

2nd-gen/packages/swc/.storybook/addons/screen-reader-addon/src/screen-reader/screenReader.ts
TarunAdobe: you're right... if disabled then no focusable so we no worry but for aria-disabled i should add a skip

2nd-gen/packages/swc/.storybook/addons/screen-reader-addon/src/screen-reader/screenReader.ts
TarunAdobe: Addressed! We now have:

- Mutation observer for aria-checked, aria-pressed, aria-expanded, aria-selected - announces state changes directly
- Change event handler for elements with checked property (like sp-checkbox) - catches native form element patterns

Both paths announce just the state ("checked", "pressed", "expanded") without re-announcing the full element.

1st-gen/packages/radio/src/spectrum-radio.css
```diff
@@ -87,7 +87,6 @@
   position: fixed;
   inset-block-end: 100%;
   inset-inline-end: 100%;
-  clip: rect(1px, 1px, 1px, 1px);
```
rubencarvalho, line 90: Please revert

1st-gen/packages/link/src/spectrum-link.css
```diff
@@ -47,9 +47,9 @@
 }
 
 a {
-  text-decoration-skip: objects;
   color: var(--highcontrast-link-text-color, var(--mod-link-text-color, var(--mod-link-text-color-primary-default, var(--spectrum-accent-content-color-default))));
   text-decoration: underline;
+  text-decoration-skip-ink: auto;
```
rubencarvalho: Please revert

1st-gen/packages/table/src/spectrum-table.css
rubencarvalho, line 1: Please revert

2nd-gen/packages/swc/.storybook/addons/screen-reader-addon/src/components/screen-reader-panel.ts
```diff
@@ -0,0 +1,328 @@
+/**
+ * Copyright 2025 Adobe. All rights reserved.
```
rubencarvalho:
```suggestion
 * Copyright 2026 Adobe. All rights reserved.
```

2nd-gen/packages/swc/.storybook/addons/screen-reader-addon/src/screen-reader/screenReader.ts
```diff
@@ -0,0 +1,1087 @@
+/**
+ * Copyright 2025 Adobe. All rights reserved.
```
rubencarvalho:
```suggestion
 * Copyright 2026 Adobe. All rights reserved.
```

2nd-gen/packages/swc/.storybook/addons/screen-reader-addon/manager.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * Copyright 2025 Adobe. All rights reserved.
```
rubencarvalho:
```suggestion
 * Copyright 2026 Adobe. All rights reserved.
```

rubencarvalho (changes requested):

rubencarvalho (approved):

PR #5917, opened 2025-12-03
chore: two doc blocks to get custom docs added to the template

caseyisonit:
ok I have figured out how to make the doc blocks reusable AND get MDX rendering with functional JSX + Doc blocks

the docblocks have been converted to `Spectrum(Docs|Stories)` and takes in an argument of the component stories imported `* as ComponentStories` for `of` and `tag=[overview, usage, a11y, examples]`. We can easily extend other tags to this pattern if we see needed as we get to more complex components.

We have a super simple pattern for docs now with `component.usage.mdx` and `component.a11y.mdx`. The pattern is also set up for `overview` and `examples` tags, but that can be on an as-needed basis or as we progress in garage week. I used progress-circle for most explorations so check it out. 

i like having the "description" in the subtitle doc block since most of them are short. it would be nice for components to have a quick blurb defining the component and then expand upon it in the description jsdoc above the meta tag in `component.stories.ts`. lets chat more about this.

This still remaining to be solved, the tabs hot linking correctly and the toc rendering the active tab. Also i think ive changed my mind on rendering emtpy tabs. I would like to build in logic that if all doc blocks in a section of the DocumentTemplate.mdx return empty or receive the fallback string (two potential approaches) that the tab and tab panel are removed from the autodocs. as well as if there is only the playground story (i.e. status light, asset) for a component that the examples tab does not render, ive found a cool article that had an interesting approach for an idea. 

Read: [doc blocks and document template blog post](https://evomark.co.uk/development/making-your-own-documentation-template-with-mdx/)

Super stoked the templatizing of docs is working out!! its going to make writing docs soooo much lighter of a lift IMO


2nd-gen/packages/swc/.storybook/DocumentTemplate.mdx
```diff
@@ -4,19 +4,20 @@ import {
...
 import '@spectrum-web-components/tabs/sp-tab-panel.js';
 
 <Meta isTemplate />
 
 <Title />
-
-This template will likely be going away in the future, but for now it's here to help us get started.
+<Subtitle />
 
 <sp-tabs id="component-doc-tabs" size="xl" selected="overview" auto label="Component Documentation">
```
nikkimk, line 22: I'm thinking we might not need the tabs

2nd-gen/packages/swc/.storybook/DocumentTemplate.mdx
```diff
@@ -28,26 +29,18 @@ This template will likely be going away in the future, but for now it's here to
...
+        <SpectrumDocs tag="a11y" />
     </sp-tab-panel>
     <sp-tab-panel value="api" style={{ flexDirection: 'column' }}>
-            <ArgTypes />
+        <ArgTypes />
     </sp-tab-panel>
     <sp-tab-panel value="examples" style={{ flexDirection: 'column' }}>
+        <SpectrumDocs tag="examples" />
         <Stories />
     </sp-tab-panel>
```
nikkimk, line 45: I don't think we need an examples section

2nd-gen/packages/swc/.storybook/DocumentTemplate.mdx
```diff
@@ -28,26 +29,18 @@ This template will likely be going away in the future, but for now it's here to
...
-        hello
-
-        #### fourth level
-
-        bananas my friends
     </sp-tab-panel>
     <sp-tab-panel value="usage" style={{ flexDirection: 'column' }}>
-            variations
-
+        <SpectrumDocs tag="usage" />
```
nikkimk, line 34: I think this is where we put the examples from our old docs

2nd-gen/packages/swc/.storybook/DocumentTemplate.mdx
```diff
@@ -28,26 +29,18 @@ This template will likely be going away in the future, but for now it's here to
         <Description />
         <Primary />
         <Controls />
```
nikkimk, line 31: I would save the controls for last. Just let them see the component and some overall info about when to use.

2nd-gen/packages/swc/.storybook/DocumentTemplate.mdx
```diff
@@ -28,26 +29,18 @@ This template will likely be going away in the future, but for now it's here to
...
-
+        <SpectrumDocs tag="usage" />
     </sp-tab-panel>
     <sp-tab-panel value="accessibility" style={{ flexDirection: 'column' }}>
-            Heres some shit about accessibility
+        <SpectrumDocs tag="a11y" />
     </sp-tab-panel>
     <sp-tab-panel value="api" style={{ flexDirection: 'column' }}>
-            <ArgTypes />
+        <ArgTypes />
```
nikkimk, line 40: I think this is where we have an example with controls and code, so the API is interactive.

PR #5936, opened 2025-12-15
chore(base): remove `dir` management from `SpectrumElement` and `sp-theme`

rubencarvalho:
## Description

This refactor removes the legacy JavaScript-based `dir` (text direction) management from `SpectrumElement` and replaces it with the modern CSS `:dir()` pseudo-class. The previous implementation used a `MutationObserver` to watch for `dir` attribute changes on `document.documentElement` and propagated these to child components. This approach required explicit `dir` attributes on elements and manual registration with `sp-theme`.

The new approach leverages the browser's native `:dir()` CSS pseudo-class, which automatically inherits directionality from the DOM hierarchy without requiring explicit attributes. This reduces JavaScript overhead, simplifies the component architecture, and improves performance.

**Key changes:**
- Removed `isLTR` getter and `dirParent` tracking from `SpectrumMixin`
- Removed `connectedCallback`/`disconnectedCallback` overrides that managed direction registration
- Removed `MutationObserver` and `observedForElements` Set for tracking direction changes
- Added simple `dir` getter using `getComputedStyle(this).direction` for JavaScript access when needed
- Updated CSS files to replace `[dir="ltr"]`/`[dir="rtl"]` attribute selectors with `:dir(ltr)`/`:dir(rtl)`

**Note:**
The only expected VRT failures are due to a fix for a previously unknown `sp-link` RTL issue affecting multiple components.

## Motivation and context

This refactor is in preparation for 2nd-gen components, which will not include this shared direction management logic. By moving to the CSS `:dir()` pseudo-class now, we ensure a consistent approach across both 1st-gen and 2nd-gen components, and reduce the surface area of shared code that needs to be maintained between generations.

The CSS `:dir()` pseudo-class is now well-supported across modern browsers and provides a more performant, declarative approach to handling text direction. It automatically inherits directionality from parent elements in the DOM hierarchy, eliminating the need for manual attribute management and JavaScript-based observation.

## Related issue(s)

-   fixes SWC-1461

---

## Author's checklist

---

## Reviewer's checklist

### Device review


Rajdeepc:
Solid modernization effort, we just cleaned up 4+ years of workaround code.

2nd-gen/packages/core/shared/base/Base.ts
```diff
@@ -126,68 +80,16 @@ export function SpectrumMixin<T extends Constructor<ReactiveElement>>(
...
-        }
     }
     return SpectrumMixinElement;
 }
 
 export class SpectrumElement extends SpectrumMixin(LitElement) {
     static VERSION = version;
+
+    public override get dir(): CSSStyleDeclaration['direction'] {
+        return getComputedStyle(this).direction ?? 'ltr';
```
Rajdeepc, line 92: None of our components read `dir` during construction phase so we are safe but this is a bit expensive. Can you work with some partner teams and check if this is something worth auditing?

rubencarvalho: It is indeed expensive, but it's necessary for components that need JavaScript logic dependent on directionality (from my quick check it's e.g., ColorArea, ColorSlider, ColorWheel, Slider calculations, Menu placement).
But hey, at least this approach is more performant than the previous implementation 😛

Rajdeepc (dismissed):
Sorry I was curious and pulled down the code and did a search on few areas and have some feedback 😄 
1. The `[trackedChildren]`(https://github.com/adobe/spectrum-web-components/blob/ruben/remove-dir/1st-gen/tools/theme/src/Theme.ts#L342) Set and the methods are now dead code. You can remove or keep as no-ops for backward compatibility. 
2. There is a `[dir="rtl"]` in [`spectrum-progress-bar.css`](https://github.com/adobe/spectrum-web-components/blob/ruben/remove-dir/1st-gen/packages/progress-bar/src/spectrum-progress-bar.css#L79). This is a progressive pattern but do you feel we should document this as this is the only css file.
3. I see that there is still a direction inheritance logic in [`Theme.ts`](https://github.com/adobe/spectrum-web-components/blob/ruben/remove-dir/1st-gen/tools/theme/src/Theme.ts#L318). Seems intentional but it is slightly contrasting with changes in the PR which relies on purely CSS `dir()`.
4. A changeset is required marking this as a breaking change.
5. _Just a thought_: How about keeping `isLTR` as a convenience getter for backward compatibility?
```ts
public get isLTR(): boolean {
       return this.dir === 'ltr';
   }
```

2nd-gen/packages/core/shared/base/Base.ts
rubencarvalho: I reached out to PSWeb and they do not support RTL so they are not using `dir` at all.

rise-erpelding (commented):
I played around with several different components, most look great!

One thing I noticed though:

When I play with [Tabs Overflow in Storybook](https://swcpreviews.z13.web.core.windows.net/pr-5936/docs/first-gen-storybook/?path=/story/tabs-overflow--compact), using the control to toggle from LTR to RTL doesn't automatically change the direction of the overflow arrow until you refresh the page. So, if you load the page in LTR, then switch to RTL, the arrow appears on the wrong side and clicking on it has no effect.

My understanding is that this happens because components are not looking out for these changes in `dir` anymore -- but should they? Changing directionality seems kind of edge-case-y and perhaps is more relevant in our Storybook than an actual consumer implementation, but I thought I'd call this out to see if it's something that needs to be addressed.

rise-erpelding:
Realized I hadn't really reviewed the VRTs, so I have a couple of questions about those too:

This RTL behavior with the letters in the link text completely reversing their order seems super unexpected:
image

This RTL on [nested theme ](https://swcpreviews.z13.web.core.windows.net/pr-5936/spectrum-dark-large-rtl/review/#ThemeStories/nestedTheme.png) does seem expected to me (since direction is set at the document level), but I want to confirm that with you since it is a difference:

image

When I opened [Storybook for this Nested Theme](https://swcpreviews.z13.web.core.windows.net/pr-5936/docs/first-gen-storybook/?path=/story/theme--nested-theme), I'm also noticing that when I toggle from LTR to RTL or vice versa that the handles change position and leave a gap where the handle was before (also the value label doesn't match where the handle is):

image

rubencarvalho:
Thank you for such a thorough review! Great catches!!

Let me tell you one by one what I updated:

>When I play with [Tabs Overflow in Storybook](https://swcpreviews.z13.web.core.windows.net/pr-5936/docs/first-gen-storybook/?path=/story/tabs-overflow--compact), using the control to toggle from LTR to RTL doesn't automatically change the direction of the overflow arrow until you refresh the page. So, if you load the page in LTR, then switch to RTL, the arrow appears on the wrong side and clicking on it has no effect.

- Ooh, great catch. There was actually old unusable code on the `TabsOverflow` that I was able to remove: it was checking `changedProperties.has('dir')` in `updated()`, but since dir is no longer a Lit reactive property, that code was completely dead.
Unfortunately I don't think there's a way to reactively re-render the component based on a directionality change... `MutationObserver` only catches attribute changes on the element itself, not inherited `dir` from ancestors or CSS direction rules, and there's no browser API to watch computed styles. That said, I think it'll be very rare that someone dynamically switches direction at runtime. I did find a compromise though: I force a scroll state recalculation when the user interacts with one of the scroll arrows. So if `dir` changed between interactions, the next click will pick up the correct direction and update the button visibility accordingly. It's not perfect but combined with the other existing triggers (resize observer, slot changes, scroll events), I think it covers things pretty well in practice.


> Realized I hadn't really reviewed the VRTs, so I have a couple of questions about those too:
> 
> This RTL behavior with the letters in the link text completely reversing their order seems super unexpected: 


- For this icon story, I re-added the `dir="ltr"` in the `<bdo>` element to make sure the text stays the same in RTL.



> This RTL on [nested theme ](https://swcpreviews.z13.web.core.windows.net/pr-5936/spectrum-dark-large-rtl/review/#ThemeStories/nestedTheme.png) does seem expected to me (since direction is set at the document level), but I want to confirm that with you since it is a difference:
> 

- I agree this is expected. Previously, in this story, the `sp-theme` was hardcoded to be `dir="ltr"`, which I don't see a reason to keep. I removed it so it can react to the document directionality. That's why this ended up as a change 😄 

> When I opened [Storybook for this Nested Theme](https://swcpreviews.z13.web.core.windows.net/pr-5936/docs/first-gen-storybook/?path=/story/theme--nested-theme), I'm also noticing that when I toggle from LTR to RTL or vice versa that the handles change position and leave a gap where the handle was before (also the value label doesn't match where the handle is):

- Another awesome catch! The slider positions its handle and fill using conditional physical CSS properties: `left` in LTR, `right` in RTL, set as inline styles at render time. When the direction toggles, the slider doesn't re-render, so the inline styles become stale (e.g., left: 45% remains even after switching to RTL). I replaced the conditional left/right inline styles with `inset-inline-start` and it works now.

rubencarvalho:
> **Unreachable code**
> 
> * Is this code in [Tabs - around L434](https://github.com/adobe/spectrum-web-components/blob/ruben/remove-dir/1st-gen/packages/tabs/src/Tabs.ts#L434-L435) with `if (changes.has('dir')) {` also unreachable?
> * [`1st-gen/projects/documentation/src/components/layout.ts`](https://github.com/adobe/spectrum-web-components/blob/ruben/remove-dir/1st-gen/projects/documentation/src/components/layout.ts#L519-L521) (lines 519–521)

Another great catch!
- I cleaned up that unreachable code in Tabs.
- The second one on the documentation website needed an update to work as expected 😄 

> **Visual behavior when toggling LTR/RTL (probably non-blocking)** I suspect these are ok as-is because they update as soon as you interact, similar to how Tabs Overflow does now after your fix, but calling out. Like you said it seems like it will be rare for someone to change directionality outside of documentation, so I think these are very likely fine:
> 
> * I see a visual issue with [Tabs](https://swcpreviews.z13.web.core.windows.net/pr-5936/docs/first-gen-storybook/?path=/story/tabs--default) where the selection indicator doesn't move immediately when toggling from LTR to RTL.
> * Similarly I'm also noticing that the color handle on the [color area](https://swcpreviews.z13.web.core.windows.net/pr-5936/docs/first-gen-storybook/?path=/story/color-area--default) doesn't change its position on toggle, although it does move as soon as you try to move the handle (with mouse or keyboard) after you switch sides, which is similar to the behavior now with Tabs Overflow.
> * The same with [Color wheel](https://swcpreviews.z13.web.core.windows.net/pr-5936/docs/first-gen-storybook/?path=/story/color-wheel--default), but again this updates on the first interaction, so it seems non-blocking.

- Yes... Unfortunately I don't think there is an obvious way to do this for now. I did fix it for some components (e.g. Tabs) in the docs / storybook website by firing a `resize` event when switching LTR/RTL modes.
If this ever becomes an issue for our consumers we can revisit it 😄 

> Last thing - this looks like it's missing a changeset.

Added one! 👍

5t3ph (approved):
Happy to approve as is, but _if_ we get feedback that we need to dynamically respond to direction changes, we could look into adding a "theme" [custom property to observe](https://observe.style/) under the `:dir()` selectors that would be inheritable through shadow trees.

Issue #5910, opened 2025-11-27
[Feat]:Support sp-combobox to take both itemText and value, so that duplicate itemTexts can exist in the list


preethialuru:
### Code of conduct

- [x] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-combobox

### Description of the requested feature

When there is a list of items, which has itemText and value. Currently, the combobox only supports the itemText to be sent to change event. So any handling needs to be done using itemText. 

But is the list of items has duplicate items with same itemText and different value (uniqueId, in our case its projectId). In this case, we would need both itemText and value to be sent to the change event. So the handling can be done based on both the itemText and value, even if the itemText is same, we can use value as differentiator. 

So the request here is to, expose an explicit selectedValue property would eliminate reliance on itemText uniqueness and provide a more reliable contract for downstream consumers.

Reference thread: https://cclight.slack.com/archives/CESK60MQD/p1764083843191409

### Would you like to track this issue in Jira?

- [x] Yes, please tell me the ticket number!

nikkimk:
Can you provide an accessible use case for duplicate item text? How would users make the distinction between items with the same item text? What happens when that text is typed in the combobox? Is the first item selected?

PR #5950, opened 2026-01-05
fix: update combobox to handle duplicate items

preethialuru:
## Description

This PR fixes an issue in Combobox component where duplicate itemText values would cause incorrect behavior. When multiple options had the same display text but different underlying values, selecting one would always resolve to the first match.

## Motivation and context

When there is a list of items, which has itemText and value. Currently, the combobox only supports the itemText to be sent to change event. So any handling needs to be done using itemText.

But is the list of items has duplicate items with same itemText and different value (uniqueId, in our case its projectId). In this case, we would need both itemText and value to be sent to the change event. So the handling can be done based on both the itemText and value, even if the itemText is same, we can use value as differentiator.

So the request here is to, expose an explicit selectedValue property would eliminate reliance on itemText uniqueness and provide a more reliable contract for downstream consumers.

https://github.com/adobe/spectrum-web-components/issues/5910

## Related issue(s)

-   fixes [5910]

https://github.com/user-attachments/assets/b769b6d5-448c-4015-892c-e80103647b02

## Author's checklist

---

## Reviewer's checklist

### Manual review test cases

    1. Go [here](url)
    2. Do this action
    3. Expect this result

    1. Go [here](url)
    2. Do this action
    3. Expect this result

### Device review


caseyisonit:
The issue described is an accessibility pattern violation. each item should have a unique value AND label for the drop down. There should never be an instance where two items render the same label with different underlying values.

1st-gen/packages/combobox/stories/combobox.stories.ts
```diff
@@ -284,3 +284,50 @@ controlled.parameters = {
...
+/**
+ * Test story to verify change event includes both value and itemText,
+ * especially for options with duplicate itemText values.
+ */
+export const duplicateItemText = (): TemplateResult => {
+    const optionsWithDuplicates: ComboboxOption[] = [
+        { value: 'val1', itemText: 'Same Display Text' },
+        { value: 'val2', itemText: 'Same Display Text' },
+        { value: 'val3', itemText: 'Unique Text' },
+    ];
```
nikkimk: This example would not be accessible as there is no way to differentiate options to the user. Is there an accessible use case you can provide for this PR instead?

1st-gen/packages/combobox/src/Combobox.ts
```diff
@@ -302,13 +310,34 @@ export class Combobox extends Textfield {
...
+                bubbles: true,
+                composed: true,
+                detail: {
+                    value: this.itemValue,
+                    itemText: this.value,
+                },
+            })
+        );
+    }
+
```
nikkimk, line 335: The combobox documentation should be updated to reflect this.

nikkimk:
We also need to make sure this PR includes a [changeset](https://opensource.adobe.com/spectrum-web-components/guides/writing-changesets/).

1st-gen/packages/combobox/src/Combobox.ts
Rajdeepc: Overriding `handleChange` will suppress base class behaviour too. Do we want this?
This is risky. The base class is doing important housekeeping (state updates, validation, other events). My approach would be to use `super.handleChange()` unless you intentionally want to replace that logic.

1st-gen/packages/combobox/src/Combobox.ts
```diff
@@ -302,13 +310,34 @@ export class Combobox extends Textfield {
...
         );
+        this.itemValue = selected?.value || '';
+        this._menuSelectedValue = selected?.value || '';
         this.value = selected?.itemText || '';
+        this.handleChange();
+        // Clear _menuSelectedValue after the update cycle completes to prevent stale values
+        // when selecting options with the same itemText (where value doesn't change).
+        this.updateComplete.then(() => {
+            this._menuSelectedValue = '';
+        });
```
Rajdeepc: You are setting `_menuSelectedValue` when selecting a menu item and clear it both in shouldUpdate (synchronously) and via `this.updateComplete.then(() => { this._menuSelectedValue = ''; })`.
This is redundant and can mask timing issues. Remove `updateComplete.then(...)` cleanup and keep clearing in `shouldUpdate` right after you use it. Much clearer and ties it to the render/update cycle

1st-gen/packages/combobox/src/Combobox.ts
Rajdeepc: Texfield is already dispatching `change` event at `@change` handler. This is duplicate semantics.
Check here: https://github.com/adobe/spectrum-web-components/blob/63775dd92c603779113199372e85a9365f919147/1st-gen/packages/textfield/src/Textfield.ts#L256

1st-gen/packages/combobox/src/Combobox.ts
```diff
@@ -339,10 +368,15 @@ export class Combobox extends Textfield {
...
-            this.itemValue =
-                this.availableOptions.find(
-                    (option) => option.itemText === this.value
-                )?.value ?? '';
+            if (this._menuSelectedValue) {
+                this.itemValue = this._menuSelectedValue;
+                this._menuSelectedValue = '';
+            } else {
+                this.itemValue =
+                    this.availableOptions.find(
```
Rajdeepc: If availableOptions is stale or lazily updated, you might pick a wrong match. Please ensure `filterAvailableOptions()` is deterministic and up-to-date before `shouldUpdate` runs.

1st-gen/packages/combobox/src/Combobox.ts
preethialuru: I want to dispatch change event with detail like text and value, which is not added in Textfield. These details are needed for the user to handle UI on their side.

Suppose, we have a requirement to store the itemText selected in the text box and pass it to API on a CTA click.

preethialuru:
> I have concerns about the accessibility and usability of items with identical text. Can you provide an example where this is necessary and is accessible.

Hi @nikkimk, we've a requirement where itemText can be duplicate. We cannot rely on just the itemText to handle change event. Please refer to this requirement thread in SWC channel https://cclight.slack.com/archives/CESK60MQD/p1764083843191409

1st-gen/packages/combobox/stories/combobox.stories.ts
preethialuru: I've updated the example to be more accessible.

1st-gen/packages/combobox/src/Combobox.ts
preethialuru: > The combobox documentation should be updated to reflect this.

Can you please let me know where do we need to update the docs?

1st-gen/packages/combobox/src/Combobox.ts
preethialuru: Removed updateComplete.then(...)

1st-gen/packages/combobox/src/Combobox.ts
preethialuru: I've updated PR to use all the options instead of availableOptions.

1st-gen/packages/combobox/src/Combobox.ts
nikkimk: > > The combobox documentation should be updated to reflect this.
> 
> Can you please let me know where do we need to update the docs?

@preethialuru `1st-gen/packages/combobox/README.md`

Add a section just above the accessibility section, something like this:

````
### Behaviors

#### Change Event

When the selection changes, a `change` event is fired. The event detail contains `value` and `itemText` of the selected option.

```html demo
<sp-combobox id="employee" label="Employee">
    <sp-menu-item value="emp-1042">Alex Johnson (Engineering)</sp-menu-item>
    <sp-menu-item value="emp-2187">Alex Johnson (Marketing)</sp-menu-item>
    <sp-menu-item value="emp-3301">Sam Chen (Design)</sp-menu-item>
    <sp-menu-item value="emp-4455">Jordan Lee (Sales)</sp-menu-item>
</sp-combobox>
<script>
    document.getElementById('employee').addEventListener('change', (event) => {
        alert(`${event.detail.value}: ${event.detail.itemText}`);
    });
</script>
```
````

File: 1st-gen/packages/combobox/src/Combobox.ts [hunk 5]
```diff
@@ -302,13 +310,29 @@ export class Combobox extends Textfield {
         const selected = (this.options || this.optionEls).find(
             (item) => item.value === target?.value
         );
+        this.itemValue = selected?.value || '';
+        this._menuSelectedValue = selected?.value || '';
         this.value = selected?.itemText || '';
+        this.handleChange();
```
Rajdeepc, line 316: The `sp-menu` fires its own `change` event with `bubbles: true, composed: true.` That event propagates up through the shadow DOM and reaches consumers listening on `<sp-combobox>`. `event.preventDefault()` does not stop propagation -- it only prevents the browser's default action.
Then `this.handleChange()` dispatches a second `CustomEvent('change', ...)`. Consumers will receive two change events per selection: the menu's original plain Event (no detail) and the new CustomEvent (with detail).
Please call `event.stopPropagation()` to prevent the menu's change from leaking out.

1st-gen/packages/combobox/src/Combobox.ts
Rajdeepc: @preethialuru Your use case is valid but I see ther eis big problem changing the event contract. It effects all change events not just menu selection. If you need `itemValue` we can expose it as a public reactive property rather than changing the event contract. You can add a public @property for the selected option's value:
```ts
@property({ type: String, attribute: 'selected-value' })
public selectedValue = '';
```
Then you can capture it on your app side by
```ts
combobox.addEventListener('change', () => {
    const itemText = combobox.value;           // display text
    const value = combobox.selectedValue;      // underlying ID
    api.save({ id: value, name: itemText });
});
```
This would be non-breaking and will work for all scenarios.

1st-gen/packages/combobox/src/Combobox.ts
```diff
@@ -105,6 +106,13 @@ export class Combobox extends Textfield {
 
     private itemValue = '';
 
+    /**
+     * Tracks the value when an item is selected from the menu dropdown.
+     * This ensures we preserve the exact selected value even when multiple
+     * options have the same itemText.
+     */
+    private _menuSelectedValue: string = '';
```
Rajdeepc, line 114: You dont' need this. You can replace this with a boolean. You are already setting `this.itemValue` in line 313.. Instead add a flag here
```ts
private _valueSetByMenu = false;
```

1st-gen/packages/combobox/src/Combobox.ts
```diff
@@ -302,13 +310,29 @@ export class Combobox extends Textfield {
         const selected = (this.options || this.optionEls).find(
             (item) => item.value === target?.value
         );
+        this.itemValue = selected?.value || '';
+        this._menuSelectedValue = selected?.value || '';
```
Rajdeepc, line 314: No need of creating a shadow copy here. Raise the guard so `shouldUpdate` doesn't overwrite it via itemText lookup.
```ts
this.itemValue = selected?.value || '';
this._valueSetByMenu = true;
```

1st-gen/packages/combobox/src/Combobox.ts
```diff
@@ -339,10 +363,15 @@ export class Combobox extends Textfield {
...
-                )?.value ?? '';
+            if (this._menuSelectedValue) {
+                this.itemValue = this._menuSelectedValue;
+                this._menuSelectedValue = '';
+            } else {
+                const allOptions = this.options || this.optionEls;
+                this.itemValue =
+                    allOptions.find((option) => option.itemText === this.value)
+                        ?.value ?? '';
+            }
```
Rajdeepc, line 374: Check the boolean guard instead of reading a shadow copy of the value
```ts
if (this._valueSetByMenu) {
            // itemValue was already set by handleMenuChange; just lower the guard.
            this._valueSetByMenu = false;
        } else {
            // Value changed from typing or programmatic set -- resolve itemValue
            // by looking up the matching option.
            const allOptions = this.options || this.optionEls;
            this.itemValue =
                allOptions.find((option) => option.itemText === this.value)
                    ?.value ?? '';
        }
```

Rajdeepc:
Hi @preethialuru , Just to let you know that this PR has been moved to a feature request rather than a bug fix. I have added some feedback on the implementation. Once you go through it let me know we can discuss and after talking to the a11y team this is where we stand.

I think we could approve the PR under these three conditions:
1. The examples in docs and storybook  leverage the description slot to differentiate the duplicates.
2. We add guidance under the accessibility documentation to ensure that if they are duplicates, that there is a unique description added.
3. We verify that the screen reader reads them as as different item

Issue #5947, opened 2025-12-22
[Bug]: Modal overlay is hiding when 2nd modal overlay is opened


spdev3000:
### Code of conduct

- [x] I agree to follow this project's code of conduct.

### Impacted component(s)

Overlay

### Library version

1.9.1

### Expected behavior

When we open a second modal overlay (such as a dialog) - the first modal overlay should stay visible "below" the new modal overlay.

### Actual behavior

Currently when we open a 2nd modal overlay (dialog) the first overlay gets auto hidden.

### What browsers are you seeing the problem in?

Chrome

### How can we reproduce this issue?

1. Go to https://stackblitz.com/edit/vitejs-vite-a4u2bta3?file=src%2Fmy-element.ts
2. Click on Open Dialogs button
3. First dialog opens
4. After 1 sec the second modal overlay dialog opens which then hides the first overlay

### Severity

SEV 3

### Would you like to track this issue in Jira?

- [x] Yes, please tell me the ticket number!

Rajdeepc:
@spdev3000 In the original implementation, modal overlays used `<dialog>.showModal()` which natively supports stacking multiple modals on the top layer. When we migrated to the Popover API for performance optimization (avoiding the overhead of dialog modal semantics), we used `popover="auto"` for modal overlays.
The problem is that the browser's Popover API only allows one `popover="auto"` element to be visible at a time. When a second auto-popover opens, the browser automatically closes the first one (light dismiss behavior). This is why sibling modal overlays were closing each other. @rubencarvalho and me have discussed this and to help you bypass this regression we came up with a fix.

1. We have changed modal overlays from `popover="auto"` to `popover="manual"` - This allows multiple modal overlays to be visible simultaneously, restoring the stacking behavior that existed with showModal().
2. Next we added explicit Escape key handling for modal overlays - Since `popover="manual"` doesn't have automatic Escape key dismissal, we now handle it explicitly in the OverlayStack. This ensures only the topmost modal closes when Escape is pressed.
3. To have hover suppression in OverlayTrigger - To prevent hover overlays from immediately opening when focus returns to the trigger after a modal closes (a side effect of the timing differences between auto and manual popovers).

Let me know if there are any additional scenarios we should test before we add this fix? 
I have added a small PR below with these fixes too for validations and testing edge cases if we are missing anything else!
PR: https://github.com/adobe/spectrum-web-components/pull/5951

spdev3000:
[...]

PR #5951, opened 2026-01-05
fix(overlay): siblings overlays stay visible

Rajdeepc:
### Root cause
When we migrated from `<dialog>.showModal()` to the Popover API for performance optimization, modal overlays were configured to use `popover="auto"`. The browser's Popover API only allows one `popover="auto"` element to be visible at a time—opening a second one automatically closes the first (light dismiss behavior).
Additionally, a recent PR [(#5907)](https://github.com/adobe/spectrum-web-components/pull/5907) added click-blocking for modal overlays to replicate `showModal()` behavior. This click-blocking was preventing light dismiss from working when combined with `popover="manual"`.

### Changes

1. `Overlay.ts`

- Changed modal overlays from `popover="auto"` to `popover="manual"`
- Allows multiple modal overlays to be visible simultaneously
- Restores the stacking behavior that existed with `showModal()`
- Updated `needsModalBackdrop` to only apply to page overlays
- Modal overlays rely on `handlePointerup` in OverlayStack for light dismiss
- Page overlays retain the backdrop since they don't have light dismiss

2. `OverlayStack.ts`

- Modified `handlePointerdown` to only block clicks for page overlays
- Allows click events to flow through for modal overlays
- Ensures `pointerdownPath` is cached so `handlePointerup` can handle light dismiss
- Modified handleClick to only block clicks for page overlays
- Same rationale as above
- Added explicit Escape key handling for modal type
- Since `popover="manual"` doesn't have automatic Escape key dismissal, we now handle it explicitly
- Ensures only the topmost modal closes when Escape is pressed

## Related issue(s)

-   fixes https://github.com/adobe/spectrum-web-components/issues/5947

---

## Author's checklist

---

## Reviewer's checklist

### Manual review test cases

    1. Go [here](url)
    2. Do this action
    3. Expect this result

    1. Go [here](url)
    2. Do this action
    3. Expect this result

### Device review

### DIFF
--- a/.changeset/avatar-accessibility-fix.md
+++ b/.changeset/avatar-accessibility-fix.md
@@ -1,0 +1,7 @@
+---
+'@spectrum-web-components/avatar': patch
+---
+
+**Added**: `is-decorative` attribute to `<sp-avatar>` to allow developers to explicitly mark avatars as decorative. When set, the avatar is hidden from screen readers with `alt=""` and `aria-hidden="true"`.
+
+**Fixed**: Fixed accessibility violation where `<sp-avatar>` rendered an underlying `img` without any `alt` attribute when no `label` was provided. The component now defaults to `alt=""` when neither `label` nor `is-decorative` is provided, and logs a dev mode warning to help developers catch missing accessibility attributes.
--- a/.changeset/tall-moments-raise.md
+++ b/.changeset/tall-moments-raise.md
@@ -1,0 +1,6 @@
+---
+'@spectrum-web-components/overlay': patch
+---
+
+**Fixed**: Third-level submenus now open correctly on Safari.
+Simplified the WebKit layout timing fix in `computePlacement()` to ensure consistent overlay positioning for deeply nested menus.
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,7 +22,7 @@ parameters:
     # 3. Commit this change to the PR branch where the changes exist.
     current_golden_images_hash:
         type: string
-        default: b88c1b90e553e4f212ad60a5d96abc4dc9ed6be0
+        default: 4304495a938a5b7dd3e64e12ebc59dfd5e300c1b
     wireit_cache_name:
         type: string
         default: wireit
--- a/1st-gen/packages/avatar/README.md
+++ b/1st-gen/packages/avatar/README.md
@@ -155,4 +155,37 @@ An avatar in a disabled state shows that an avatar exists, but is not available 
 
 ### Accessibility
 
-The `label` attribute of the `<sp-avatar>` will be passed into the `<img>` element as the `alt` tag for use in defining a textual representation of the image displayed.
+The `<sp-avatar>` component requires proper accessibility attributes to ensure screen readers can appropriately handle the avatar image.
+
+#### Labeled avatars
+
+The `label` attribute of the `<sp-avatar>` will be passed into the `<img>` element as the `alt` attribute for use in defining a textual representation of the image displayed. This is the recommended approach for avatars that convey meaningful information.
+
+```html demo
+<sp-avatar
+    label="John Doe"
+    src="https://picsum.photos/500/500"
+></sp-avatar>
+```
+
+#### Decorative avatars
+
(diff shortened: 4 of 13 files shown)


1st-gen/packages/overlay/src/OverlayTrigger.ts
```diff
@@ -181,6 +189,33 @@ export class OverlayTrigger extends SpectrumElement {
...
+            // re-rendered yet when the HoverController tries to open.
+            if (type === 'click' || type === 'longpress') {
+                this.hoverSuppressed = true;
+                if (this.hoverOverlayElement) {
+                    this.hoverOverlayElement.disabled = true;
+                }
+                // The overlay's transition guarantee uses 3 animation frames before
+                // calling finish() which triggers returnFocus(). We need to wait longer
+                // than that before re-enabling hover to prevent it from opening.
+                requestAnimationFrame(() => {
```
spdev3000: This looks a bit wonky and I assume it might bring further timing issues down the road when used in quite complex applications.

Rajdeepc: @spdev3000 Thanks for the feedback. I’ve refactored this to a cleaner approach using a single setTimeout. The 100ms delay reliably covers the close transition and focus return, while remaining short enough to avoid any perceived sluggishness.

This also smooths out edge cases like rapid close/open interactions. Please take another look and let me know if you see any remaining concerns or edge cases we should address.

spdev3000:
looks good for me, one question:
I am wondering, if clicking outside a modal overlay (with popover="manual") also dismisses it within your approach (if "dismissable" is set for example to the dialog component).
And also the other way - if we don't have a dismissable attritube set to dialog - is the user supposed to dismiss the dialog and "Escaping" is being prevented?

Rajdeepc:
> looks good for me, one question: I am wondering, if clicking outside a modal overlay (with popover="manual") also dismisses it within your approach (if "dismissable" is set for example to the dialog component). And also the other way - if we don't have a dismissable attritube set to dialog - is the user supposed to dismiss the dialog and "Escaping" is being prevented?

To your first question this is intentional - modal overlays should trap interaction until explicitly dismissed. This matches the Spectrum design pattern where modals require explicit user action to close. If you need click-outside-to-close behavior, you should use `type="auto"` instead of `type="modal"` (for non-blocking overlays) or implement click-outside logic in your dialog component and programmatically close the overlay.
To your second question, If non-dismissable dialogs should also block Escape, we could add a way for dialog content to communicate its dismissable state to the overlay or check if the overlay content has a dismissable attribute before closing.

spdev3000:
To #1 
SWC currently supports dismissing a modal dialog via click outside a DialogWrapper, see: https://opensource.adobe.com/spectrum-web-components/components/dialog-wrapper/#dismissable
So this should still work.

re #2: Yes we have use cases, where no "cancel" button is rendered inside a dialog and Escape should _not_ dismiss or cancel the dialog, so we'd need such a behavior.

Rajdeepc:
> To #1 SWC currently supports dismissing a modal dialog via click outside a DialogWrapper, see: https://opensource.adobe.com/spectrum-web-components/components/dialog-wrapper/#dismissable So this should still work.
> 
> re #2: Yes we have use cases, where no "cancel" button is rendered inside a dialog and Escape should _not_ dismiss or cancel the dialog, so we'd need such a behavior.

@spdev3000 you can use the overlay type="page" 
https://stackblitz.com/edit/vitejs-vite-mnzgm64e?file=src%2Fmy-element.ts

blunteshwar (approved):

1st-gen/packages/overlay/src/OverlayStack.ts
```diff
@@ -158,34 +158,30 @@ class OverlayStack {
...
+            const pageOverlays = this.stack.filter(
+                (o) => o.open && o.type === 'page'
+            );
+            if (
+                pageOverlays.length > 0 &&
+                !this.isEventInsideModal(pointerPath, pageOverlays)
+            ) {
+                event.preventDefault();
+                event.stopPropagation();
+                event.stopImmediatePropagation();
```
spdev3000, line 180: Not needed?

Issue #5099, opened 2025-02-18
sp-picker and sp-slider do not stop propagation of arrow key events after reacting to those events


rubencarvalho:
### Code of conduct

- [x] I agree to follow this project's code of conduct.

### Impacted component(s)

Picker, Slider

### Description of the requested feature

Opening this issue in behalf of Mukul Sharma.

Picker and Slider components should stop propagation of Arrow keyboard events after reacting to those events.

From the [Slack thread](https://adobedesign.slack.com/archives/CESK60MQD/p1739296909680739) discussion:
While we currently don't have a "blanket policy" for this, in this case, it seems reasonable for the slider to stop propagation since handling arrow key events is intrinsic to its functionality.
For other events we expose, consumers can handle them on their end depending on their use case.

### Implementation notes or ideas

Slider and Picker should stop propagation of Arrow key events

PR #5965, opened 2026-01-15
fix: prevent arrow key events propagation in picker and slider

rubencarvalho:
## Description

This PR adds event.stopPropagation() for arrow key events in the sp-picker and sp-slider components after they react to those events.

## Motivation and context

When sp-picker or sp-slider components are used inside containers that also listen for arrow key events (e.g., carousels, tab panels, grids, or custom navigation handlers), the arrow key events would bubble up and trigger unintended behavior in the parent.
Since handling arrow key events is intrinsic to these components' functionality (navigating options in picker, adjusting values in slider), it's reasonable for them to stop propagation after handling these events. This prevents the "double handling" problem where both the component and its ancestor react to the same keystroke.

## Related issue(s)

- fixes SWC-684
- fixes https://github.com/adobe/spectrum-web-components/issues/5099

---

## Author's checklist

---

## Reviewer's checklist

### Device review


1st-gen/packages/picker/src/Picker.ts
```diff
@@ -997,9 +1003,13 @@ export class Picker extends PickerBase {
         if (openKeys) {
             this.keyboardOpen();
             event.preventDefault();
+            if (arrowKeys) {
+                event.stopPropagation();
+            }
```
blunteshwar, line 1008: Once the picker is opened isn't the event `keydown` event happening on `sp-menu` rather than picket?
Thus once the picker is opened using `openkeys` all the keyboard event will happen on `sp-menu` and this function `handleKeydown` in `Picker.ts` will never be called.
Corect me if I am wrong but I believe we should stop propogation of event in menu rather than picker

1st-gen/packages/picker/src/Picker.ts
```diff
@@ -997,9 +1003,13 @@ export class Picker extends PickerBase {
...
             return;
         }
         event.preventDefault();
+        event.stopPropagation();
```
blunteshwar, line 1012: This method will also be never get called!

1st-gen/packages/slider/src/HandleController.ts
```diff
@@ -472,6 +472,14 @@ export class HandleController {
         const input = event.target as InputWithModel;
         input.model.handle.highlight = true;
         this.requestUpdate();
+        // Stop propagation for arrow keys since slider handles them internally
+        if (
+            ['ArrowUp', 'ArrowDown', 'ArrowLeft', 'ArrowRight'].includes(
+                event.key
+            )
+        ) {
+            event.stopPropagation();
```
blunteshwar, line 481: This totally makes sense here since the event is actually being called on slider!

1st-gen/packages/picker/src/Picker.ts
rubencarvalho: Great call! This is only covering the initial "opening" of the Picker (e.g., on `ArrowDown`). Let me re-think about this...

1st-gen/packages/picker/src/Picker.ts
rubencarvalho: This gets called on `ArrowRight` or `ArrowLeft` when the Picker is closed and we cycle through the options. Nonetheless, I agree this probably needs to live somewhere else... Looking 👀

rubencarvalho:
> For Slider the I could understand the need for `event.stopPropagation()` but for picker I think we should add that to menu instead of picker

OK! I looked at this again and changed it slightly. When we press arrow keys on a Picker or Slider, they “eat” the event instead of letting it bubble up, but the gist of it is:

- Picker (closed) – Arrow keys stop bubbling when you’re cycling the options or open it with the arrow key down.
- Slider – Arrow keys stop bubbling when we're adjusting the value
- Menu – Arrow keys stop bubbling when we're navigating menu items (this is what fixes Picker when it’s open, one of the things you pointed out!)

The thing is, instead of just hardcoding this everywhere, I added a config option to `FocusGroupController` called `stopKeyEventPropagation`. It’s off by default (so we don’t break anything 🤷 ), but Menu opts in. I was afraid of making this the default for all other components, because I’m not really sure about the other use cases, and I see Tabs, Accordion, etc. also use this controller. If we want to, we can opt in whatever component later if we want.

blunteshwar (approved):
LGTM!

Rajdeepc:
> > For Slider the I could understand the need for `event.stopPropagation()` but for picker I think we should add that to menu instead of picker
> 
> OK! I looked at this again and changed it slightly. When we press arrow keys on a Picker or Slider, they “eat” the event instead of letting it bubble up, but the gist of it is:
> 
> * Picker (closed) – Arrow keys stop bubbling when you’re cycling the options or open it with the arrow key down.
> * Slider – Arrow keys stop bubbling when we're adjusting the value
> * Menu – Arrow keys stop bubbling when we're navigating menu items (this is what fixes Picker when it’s open, one of the things you pointed out!)
> 
> The thing is, instead of just hardcoding this everywhere, I added a config option to `FocusGroupController` called `stopKeyEventPropagation`. It’s off by default (so we don’t break anything 🤷 ), but Menu opts in. I was afraid of making this the default for all other components, because I’m not really sure about the other use cases, and I see Tabs, Accordion, etc. also use this controller. If we want to, we can opt in whatever component later if we want.

Arrow keys are overloaded in many composite patterns, and it’s better to be explicit than to silently change semantics

PR #5967, opened 2026-01-16
chore(overlay): optimize backdrop click detection and event handling

rubencarvalho:
## Description

This PR optimizes event handling performance in the overlay system by optimizing backdrop click detection. I added an early return check to skip backdrop detection when no modal/page overlays are open, so we don't need to iterate through the event's `composedPath` on every `pointerup` event.  I also changed `instanceof HTMLElement` to `instanceof HTMLDivElement` for faster filtering when checking for backdrop elements, since backdrops are always `<div>` elements.

I took the opportunity and also eliminated redundant `stopPropagation()` calls that were placed after `stopImmediatePropagation()`, since `stopImmediatePropagation()` already prevents all further event propagation.

## Motivation and context

Some code cleanup feels nice?

## Author's checklist

---

## Reviewer's checklist

### DIFF
--- a/.changeset/fix-safari-voiceover-crash.md
+++ b/.changeset/fix-safari-voiceover-crash.md
@@ -1,0 +1,6 @@
+---
+'@spectrum-web-components/picker': patch
+'@spectrum-web-components/action-menu': patch
+---
+
+**Fixed**: Safari + VoiceOver crash when opening Picker and ActionMenu. The issue was caused by an imperative `render()` call that mutated the DOM during the render cycle, causing Safari to crash while VoiceOver scanned the accessibility tree.
--- a/.changeset/good-humans-trade.md
+++ b/.changeset/good-humans-trade.md
@@ -1,0 +1,5 @@
+---
+'@spectrum-web-components/shared': patch
+---
+
+Improve reliability when composing components and mixins that observe slot presence.
--- a/1st-gen/packages/picker/src/InteractionController.ts
+++ b/1st-gen/packages/picker/src/InteractionController.ts
@@ -50,24 +50,11 @@ export class InteractionController implements ReactiveController {
         if (this._open === open) return;
         this._open = open;
 
-        if (this.overlay) {
-            this.host.open = open;
-            return;
-        }
+        this.host.open = open;
 
-        // When there is no Overlay and `open` is moving to `true`, lazily import/create
-        // an Overlay and apply that state to it.
-        customElements
-            .whenDefined('sp-overlay')
-            .then(async (): Promise<void> => {
-                const { Overlay } = await import(
-                    '@spectrum-web-components/overlay/src/Overlay.js'
-                );
-                this.overlay = new Overlay();
-                this.host.open = true;
-                this.host.requestUpdate();
-            });
-        import('@spectrum-web-components/overlay/sp-overlay.js');
+        if (!this.overlay && this.host.overlayElement) {
+            this.overlay = this.host.overlayElement;
+        }
     }
 
     private _overlay!: AbstractOverlay;
@@ -110,8 +97,6 @@ export class InteractionController implements ReactiveController {
             if (this.preventNextToggle === 'no') {
                 this.open = false;
             } else if (!this.pointerdownState) {
-                // Prevent browser driven closure while opening the Picker
-                // and the expected event series has not completed.
                 this.overlay?.manuallyKeepOpen();
             }
         }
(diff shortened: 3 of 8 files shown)


1st-gen/packages/overlay/src/OverlayStack.ts
```diff
@@ -232,19 +230,29 @@ class OverlayStack {
...
+                (overlay.type === 'modal' || overlay.type === 'page')
         );
-        if (clickedBackdrop) {
-            const topOverlay = this.stack[this.stack.length - 1];
-            // Only modal overlays close on backdrop click.
-            // Page overlays are blocking and should not be light-dismissable.
-            if (topOverlay?.type === 'modal') {
-                this.closeOverlay(topOverlay);
+        if (hasModalOrPageOverlay) {
+            const clickedBackdrop = composedPath.some(
```
Rajdeepc, line 243: `const composedPath = event.composedPath()` is still executed at the top of the handler, the optimisation is effectively lost. Can we move it inside `hasModalOrPageOverlay` to get a performance win?

1st-gen/packages/overlay/src/OverlayStack.ts
```diff
@@ -232,19 +230,29 @@ class OverlayStack {
...
+                (el) =>
+                    el instanceof HTMLDivElement &&
```
Rajdeepc, line 245: I have a little concern here: We are assuming that the backdrop will always be `div` and nothing else. This might not hold long term and will break the backdrop detection in future. Can you use a `element + class` check just to make sure we are detecting how its implemented rather than if its a backdrop or not?
```ts
 const clickedBackdrop = composedPath.some(
        (el) =>
            el &&
            (el as Element).nodeType === Node.ELEMENT_NODE &&
            (el as Element).classList?.contains('modal-backdrop')
    );
```

1st-gen/packages/overlay/src/OverlayStack.ts
```diff
@@ -232,19 +230,29 @@ class OverlayStack {
         this.lastOverlay = undefined;
 
         const lastIndex = this.stack.length - 1;
-        const clickedBackdrop = composedPath.some(
-            (el) =>
-                el instanceof HTMLElement &&
-                el.classList.contains('modal-backdrop')
+
+        // Early optimization: only check for backdrop clicks if modal/page overlays exist
+        // (only they have backdrops). This avoids iterating through composedPath unnecessarily.
```
Rajdeepc:
```suggestion
        // Avoid the cost of event.composedPath() unless we actually have a modal/page
        // overlay open. composedPath() walks and allocates the full event path and can
        // be surprisingly expensive in hot paths.
```

Rajdeepc (commented):
Nice cleanup!
Non blocking: The `HTMLDivElement` check is fine for now but it would be great if you can add a short comment asserting that the backdrops are guaranteed to be `<div>`s and also if you can lock a test with this that would be great. Otherwise you can add a more semantic element + class check for future proofing this.

1st-gen/packages/overlay/src/OverlayStack.ts
rubencarvalho: OK, I nerded out a little bit on this, ran some performance benchmarks and and here are my results:

```
Testing with backdrop at index 5 in a path of 12 elements
Running 100,000 iterations per approach

instanceof Element + hasAttribute       : 16.18ms
nodeType + hasAttribute                 : 19.15ms
Optional chaining                       : 10.02ms
instanceof Element + matches            : 15.50ms
instanceof HTMLDivElement + classList   : 9.41ms
tagName + hasAttribute                  : 15.39ms

--- Results (fastest to slowest) ---
1. instanceof HTMLDivElement + classList   : 9.41ms (100.0% of fastest)
2. Optional chaining                       : 10.02ms (93.9% of fastest)
3. tagName + hasAttribute                  : 15.39ms (61.1% of fastest)
4. instanceof Element + matches            : 15.50ms (60.7% of fastest)
5. instanceof Element + hasAttribute       : 16.18ms (58.2% of fastest)
6. nodeType + hasAttribute                 : 19.15ms (49.1% of fastest)

--- Testing worst case (backdrop NOT in path) ---
1. instanceof HTMLDivElement + classList   : 17.01ms (100.0% of fastest)
2. instanceof Element + hasAttribute       : 23.15ms (73.5% of fastest)
3. Optional chaining                       : 26.70ms (63.7% of fastest)
4. instanceof Element + matches            : 34.63ms (49.1% of fastest)
5. nodeType + hasAttribute                 : 38.58ms (44.1% of fastest)
6. tagName + hasAttribute                  : 42.94ms (39.6% of fastest)
```

So let's go for `instanceof HTMLDivElement + classList`! If the backdrop ever becomes other element other than a DIV, at least 6 tests fail, so we're covered on that front 😄

1st-gen/packages/overlay/src/OverlayStack.ts
rubencarvalho: `event.composedPath()` is already executed in `handlePointerdown (line 168)` and cached in `this.pointerdownPath`. So here it's just a property read, not a function call. I did do the optimization about avoiding iterating through the path array for backdrop detection.

PR #5968, opened 2026-01-16
fix: convert managePresenceObservedSlot to regular method

rubencarvalho:
## Description

Converted `managePresenceObservedSlot` from an arrow function property to a regular method in the `ObserveSlotPresence` mixin. This fixes a runtime error where calling `this.managePresenceObservedSlot()` from the constructor or `MutationController` callback resulted in "this.managePresenceObservedSlot is not a function".

The change aligns the implementation with the pattern used in `observe-slot-text.ts` where `manageTextObservedSlot()` is defined as a regular method.

## Motivation and context

Arrow function properties (`method = () => {}`) are instance properties, not prototype methods. When mixins are nested (e.g., `SizedMixin(ObserveSlotText(ObserveSlotPresence(...)))`), instance properties may not be accessible through the prototype chain, causing the method to be `undefined` when accessed via `this.managePresenceObservedSlot()`.
Regular methods are defined on the prototype and are reliably inherited through mixin chains, making them accessible regardless of how many mixins wrap the base class. This fix ensures consistent behavior and matches the established pattern used in similar mixins like `ObserveSlotText`.

Before:
Screenshot 2026-01-16 at 13 55 28

---

## Author's checklist

---

## Reviewer's checklist

### Manual review test cases

-   [ ] No errors surface in the console

    1. Go [here](https://swcpreviews.z13.web.core.windows.net/pr-5968/docs/second-gen-storybook/?path=/story/components-badge--default&args=dir:!null)
    2. Open your console
    3. Expect no errors

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,7 +22,8 @@ parameters:
     # 3. Commit this change to the PR branch where the changes exist.
     current_golden_images_hash:
         type: string
-        default: 4304495a938a5b7dd3e64e12ebc59dfd5e300c1b
+        default: 53d43cef8f05a527cd662707fd6a7a8e930a6a52
+
     wireit_cache_name:
         type: string
         default: wireit
--- a/1st-gen/packages/menu/src/MenuItem.ts
+++ b/1st-gen/packages/menu/src/MenuItem.ts
@@ -395,7 +395,7 @@ export class MenuItem extends LikeAnchor(
                 this.dependencyManager.loaded}
                 .placement=${this.isLTR ? 'right-start' : 'left-start'}
                 receives-focus="false"
-                .offset=${[-10, -5] as [number, number]}
+                .offset=${[-10, 0] as [number, number]}
                 .type=${'auto'}
                 @close=${(event: Event) => event.stopPropagation()}
                 @slottable-request=${this.handleSlottableRequest}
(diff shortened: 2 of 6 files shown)


Rajdeepc (commented):
This is really nice!
Non-blocking: One small thought for future-proofing: it might be worth adding an automated integration test that reproduces the `nested-mixin `constructor timing where the constructor calls` managePresenceObservedSlot` immediately after `super()`. 

Also this PR needs a changeset to surface this to the downstream customers.

TarunAdobe (approved):

rubencarvalho:
> One small thought for future-proofing: it might be worth adding an automated integration test that reproduces the `nested-mixin `constructor timing where the constructor calls` managePresenceObservedSlot` immediately after `super()`.

I added tests to cover the fix (also added one to `ObserveSlotText`, just in case 😄 ) 👍
This isn’t a public API, so there’s no customer impact. I don’t think it needs a changeset, but let me know if you had something in mind...

Rajdeepc:
> > One small thought for future-proofing: it might be worth adding an automated integration test that reproduces the `nested-mixin `constructor timing where the constructor calls` managePresenceObservedSlot` immediately after `super()`.
> 
> I added tests to cover the fix (also added one to `ObserveSlotText`, just in case 😄 ) 👍 This isn’t a public API, so there’s no customer impact. I don’t think it needs a changeset, but let me know if you had something in mind...

Thanks for clarifying. I wasn't thinking in terms of customer facing API surface as this behavioural change in a core interaction path might be worth surfacing for internal consumers and release notes readers? 😄

Rajdeepc (approved):

PR #5969, opened 2026-01-16
feat(accordion): add level property for controlling title heading

rubencarvalho:
## Description
Adds a `level` property to `<sp-accordion>` that allows developers to specify the semantic heading level (1-6) used for the accordion item heading. This property defaults to `3`, maintaining backward compatibility with the existing `<h3>` heading behavior.

The implementation uses a switch statement to render the appropriate heading tag (`h1` through `h6`) based on the `level` property value, ensuring proper document structure and accessibility.

## Motivation and context

Proper heading hierarchy is essential for accessibility. Screen readers and assistive technologies rely on heading levels to generate navigation outlines and understand page structure. Currently, accordion items always use `<h3>` headings, which may not match the semantic structure of the page where the accordion is used.

This change allows developers to customize the heading level to match their document structure while maintaining a sensible default.

## Related issue(s)

-   fixes PS-180066
-   fixes PS-180061
-   fixes PS-180055
-   fixes PS-180028
-   fixes PS-179997
-   fixes PS-179849
-  fixes SWC-1466

## Screenshots 
Screenshot 2026-01-16 at 14 48 01

---

## Author's checklist

---

## Reviewer's checklist

### Manual review test cases

-   [ ] Verify docs website example

    1. Go [here](https://swcpreviews.z13.web.core.windows.net/pr-5969/docs/first-gen-docs/components/accordion-item/#heading-level)
    2. Inspect the DOM
    3. Expect to see `h2` and `h4` respectively

### Device review

### DIFF
--- a/1st-gen/packages/accordion/src/Accordion.ts
+++ b/1st-gen/packages/accordion/src/Accordion.ts
@@ -110,18 +110,12 @@ export class Accordion extends SizedMixin(SpectrumElement, {
 
     protected override updated(changed: PropertyValues<this>): void {
         super.updated(changed);
-        if (
-            changed.has('size') &&
-            (!!changed.get('size') || this.size !== 'm')
-        ) {
+        if (changed.has('size')) {
             this.items.forEach((item) => {
                 item.size = this.size;
             });
         }
-        if (
-            changed.has('level') &&
-            (!!changed.get('level') || this.level !== 3)
-        ) {
+        if (changed.has('level')) {
             this.items.forEach((item) => {
                 item.level = this.level;
             });


1st-gen/packages/accordion/src/AccordionItem.ts
```diff
@@ -111,20 +118,61 @@ export class AccordionItem extends SizedMixin(Focusable, {
         return chevronIcon[this.size || 'm']();
     };
 
+    private getHeadingLevel(): number {
+        const level = this.level ?? 3;
+        return Math.max(1, Math.min(6, level));
+    }
```
Rajdeepc, line 125: If someone sets `level="auto"` or an otherwise non-numeric string, `Number(this.level)` will become NaN and `Math.max/Math.min` will propagate NaN, which can break the switch logic, falling back to default only because of the default case. Can you make the getter stricter and coerce to an integer like below. This will guarantee the switch always receives a valid interger.
```ts
private getHeadingLevel(): number {
    const raw = Number(this.level);
    if (!Number.isFinite(raw)) return 3;
    const level = Math.trunc(raw);
    return Math.max(1, Math.min(6, level));
}
```

1st-gen/packages/accordion/src/AccordionItem.ts
```diff
@@ -111,20 +118,61 @@ export class AccordionItem extends SizedMixin(Focusable, {
...
-                    @click=${this.onClick}
-                    aria-expanded=${this.open}
-                    aria-controls="content"
-                    ?disabled=${this.disabled}
-                >
-                    ${this.label}
-                </button>
-            </h3>
+            ${this.renderHeading()}
             <div id="content" role="region" aria-labelledby="header">
```
Rajdeepc, line 169: Semantically, the region is being labelled by the button, not the heading. This is not wrong but often the region is labelled by the heading element itself. Do you want to consider whether you prefer to reference the heading element id ("heading") instead though both are valid.

1st-gen/packages/accordion/src/AccordionItem.ts
```diff
@@ -111,20 +118,61 @@ export class AccordionItem extends SizedMixin(Focusable, {
...
+                `;
+            case 6:
+                return html`
+                    <h6 id="heading">${headingContent}</h6>
+                `;
+            default:
+                return html`
+                    <h3 id="heading">${headingContent}</h3>
+                `;
+        }
```
Rajdeepc, line 163: nit: you can use [lit's `unsafeStatic`](https://lit.dev/docs/api/static-html/#unsafeStatic) and you can remove the switch. Might hinder the readability but I am living this upto your choice. I am happy to accept the `switch` though.
```ts
const tag = unsafeStatic(`h${level}`);
....

 return html`
        <${tag} id="heading">
            ${when(this.size, this.renderChevronIcon)}
            .....
        </${tag}>
```

1st-gen/packages/accordion/src/AccordionItem.ts
rubencarvalho: Hmmm… I see what you’re saying. I looked through the codebase, and this isn’t a pattern we follow elsewhere... we generally don’t validate all inputs and instead fail gracefully.
The switch’s default case already handles edge cases like NaN or invalid strings. LMK what you think.

1st-gen/packages/accordion/src/AccordionItem.ts
rubencarvalho: Nice, done!

1st-gen/packages/accordion/src/AccordionItem.ts
rubencarvalho: Good call, I believe this has gone through A11Y review, so I'll leave the interactive button as the label for now.

1st-gen/packages/accordion/src/AccordionItem.ts
rubencarvalho: Oh no 😞  

Actually, 
> Bindings in tag names are not supported.

https://app.circleci.com/pipelines/github/adobe/spectrum-web-components/35331/workflows/3d98caa6-291a-4ae4-b94d-d3df36a0989a/jobs/868487

1st-gen/packages/accordion/src/AccordionItem.ts
rubencarvalho: cc: @nikkimk

1st-gen/packages/accordion/accordion-item.md
```diff
@@ -64,6 +64,21 @@ Individual accordion items can be disabled using the `disabled` attribute. Disab
 </sp-accordion>
 ```
 
+#### Heading level
+
+The `level` attribute controls the semantic heading level (1-6) used for the accordion item heading. This helps maintain proper document structure and accessibility. Defaults to 3:
```
nikkimk: You should not have `<h1>` elements in an accordion as there should only be 1 `<h1>` per page.

1st-gen/packages/accordion/accordion-item.md
```diff
@@ -64,6 +64,21 @@ Individual accordion items can be disabled using the `disabled` attribute. Disab
...
+
+```html
+<sp-accordion>
+    <sp-accordion-item level="2" label="Level 2 Heading">
+        <div>This accordion item uses an h2 heading.</div>
+    </sp-accordion-item>
+    <sp-accordion-item level="4" label="Level 4 Heading">
+        <div>This accordion item uses an h4 heading.</div>
+    </sp-accordion-item>
```
nikkimk: This example wont pass an a11y audit as content of equal importance should be the same heading level. We should have examples of two different `<sp-accordion>` elements with different heading levels.

1st-gen/packages/accordion/src/AccordionItem.ts
```diff
@@ -111,20 +118,61 @@ export class AccordionItem extends SizedMixin(Focusable, {
         return chevronIcon[this.size || 'm']();
     };
 
+    private getHeadingLevel(): number {
+        const level = this.level ?? 3;
+        return Math.max(1, Math.min(6, level));
```
nikkimk: Items of equal importance should be the same heading level. We should have examples of two different `<sp-accordion>` elements with different heading levels.

```suggestion
        return Math.max(2, Math.min(6, level));
```

rubencarvalho: Done!

1st-gen/packages/accordion/accordion-item.md
rubencarvalho: Good call, I added this info to the documentation.

1st-gen/packages/accordion/accordion-item.md
rubencarvalho: Oooh, nice! Thank you for the callout. I created 2 different accordions!

1st-gen/packages/accordion/src/AccordionItem.ts
nikkimk: It depends on the likelihood that accordions will be implemented with a lot of accordion items, be nested, or exist on pages with a lot of existing landmarks.

ARIA landmarks are meant to help users easily navigate the page. To skim content, screenreader users can quickly navigate by heading, by landmark, or by interactive element. If too many landmarks exist on a page, you can render landmark navigation useless. 

Since the heading itself is a way to quickly navigate, I prefer not mark the content as a region.

```
            <div id="content">
```

That said, marking the content with the region and labelling with the button does allow the screen reader user not to switch modes, so we could go with @Rajdeepc's suggestion.

```
            <div id="content" aria-labelledby="buttonID">
```

From https://www.w3.org/WAI/ARIA/apg/patterns/accordion/#wai-ariaroles,states,andproperties:

Optionally, each element that serves as a container for panel content has role [region](https://w3c.github.io/aria/#region) and [aria-labelledby](https://w3c.github.io/aria/#aria-labelledby) with a value that refers to the button that controls display of the panel.

- Avoid using the region role in circumstances that create landmark region proliferation, e.g., in an accordion that contains more than approximately 6 panels that can be expanded at the same time.
- Role region is especially helpful to the perception of structure by screen reader users when panels contain heading elements or a nested accordion.

rubencarvalho: Thanks @nikkimk for the input!! 🙏
I think I'm tempted to keep it as-is with `role="region"` and `aria-labelledby="header"`.
I was reading up on this and if I understand this correctly, if I remove the `role="region"`, the `aria-labelledby` becomes effectively meaningless: a plain `div` has an implicit role of `generic`, which prohibits naming (am I looking at the [right thing](https://w3c.github.io/aria/#generic)?).
I did test with a screen reader and I also liked that it could stay in browse mode. I see you approved the PR, but I still wanted to ack your thorough review!!

rubencarvalho: Unless that with the new information, the region proliferation is more concering in your opinion - in that case, I'm happy to remove it again 😄

1st-gen/packages/accordion/stories/accordion.stories.ts
```diff
@@ -11,7 +11,7 @@
  */
 
 import { TemplateResult } from '@spectrum-web-components/base';
-import { AccordionMarkup } from './';
+import { AccordionMarkup } from './/index.js';
```
Rajdeepc: Is this accidental? Why double slash and index.js?

1st-gen/packages/accordion/src/Accordion.ts
```diff
@@ -110,6 +118,14 @@ export class Accordion extends SizedMixin(SpectrumElement, {
                 item.size = this.size;
             });
         }
+        if (
+            changed.has('level') &&
+            (!!changed.get('level') || this.level !== 3)
+        ) {
```
Rajdeepc: When `level` changes you would still want to sync to items. I feel syncing whenever `level` changes is enough?

Rajdeepc (changes requested):

1st-gen/packages/accordion/src/Accordion.ts
rubencarvalho: ye, I removed this micro-optimization and the one above

Issue #5971, opened 2026-01-19
[Bug]: Picker don't update label if menu-items render or change later


spdev3000:
### Code of conduct

- [x] I agree to follow this project's code of conduct.

### Impacted component(s)

Picker

### Library version

1.9.1

### Expected behavior

Picker should show updated label, when children menu-items render or update later and picker-value has been already set

### Actual behavior

Picker don't show any label, if value has been set, but related menu-items are rendered at a later point

### How can we reproduce this issue?

1. Go to https://stackblitz.com/edit/vitejs-vite-9twlaz7r?file=src%2Fmy-element.ts
2. See content and label of Picker 1
3. Click Button to render sp-menu-items of Picker 2
4. See content and (no) label of Picker 2

### Severity

SEV 4

### Would you like to track this issue in Jira?

- [ ] Yes, please tell me the ticket number!

rubencarvalho:
Jira ticket created: SWC-1472

PR #5972, opened 2026-01-19
fix(picker): preserve selection when menu-items are added after value is set

rubencarvalho:
## Description

This PR fixes an issue where a Picker component with a value set before menu-items are rendered doesn't display the label when menu-items are added later (e.g., when using Lit's `when()` directive for conditional rendering or lazy loading).

The fix preserves the Picker's value when `manageSelection()` runs before menu-items are available, allowing the value to be matched and the label to display correctly once menu-items are added.

## Motivation and context

When a Picker has a `value` attribute set but its menu-items are conditionally rendered or loaded asynchronously, the component's `manageSelection()` method runs before menu-items exist. Previously, this would clear the value immediately, preventing the label from displaying even after menu-items were added.

This is a common pattern in modern web applications where:
- Menu-items are lazy-loaded from an API
- Menu-items are conditionally rendered based on user actions or application state
- Menu-items are dynamically generated based on other component state

The fix ensures the Picker maintains the value until menu-items are available, then correctly matches and displays the label.

## Related issue(s)

-  fixes https://github.com/adobe/spectrum-web-components/issues/5971
-  fixes SWC-1472

## Author's checklist

---

## Reviewer's checklist

### Manual review test cases

-   [ ] _Verify [StackBlitz isolated example](https://stackblitz.com/edit/vitejs-vite-9twlaz7r?file=src%2Fmy-element.ts) now works_

    1. Visit this example story [here](https://swcpreviews.z13.web.core.windows.net/pr-5972/docs/first-gen-storybook/?path=/story/picker--delayed-menu-items)
    2. Click the button to render the menu items
    3. Verify the Picker now shows the correct item (is not empty!)


1st-gen/packages/picker/src/Picker.ts
```diff
@@ -864,8 +864,13 @@ export class PickerBase extends SizedMixin(SpectrumElement, {
...
         } else {
-            this.value = '';
-            this.selectedItem = undefined;
+            // Only clear the value if there are menu-items but none match.
+            // If there are no menu-items yet, they might be added later (e.g., lazy loaded),
+            // so preserve the value to allow matching when items are added.
+            if (this.menuItems.length > 0) {
+                this.value = '';
+                this.selectedItem = undefined;
+            }
```
Rajdeepc: Currently with this change we are assuming items exist and items are meaningful, which might not always be true coz `menu-item` can exist in DOM in incomplete states before upgrade with no `value` or `value` not populated yet during async render. We should only clear only when items exist and at least one item has a real value and none of those values match the picker's value. Thoughts?

rubencarvalho: Nice! I updated the code + tests for this! 👍

1st-gen/packages/picker/src/Picker.ts
```diff
@@ -864,10 +864,12 @@ export class PickerBase extends SizedMixin(SpectrumElement, {
...
         } else {
-            // Only clear the value if there are menu-items but none match.
-            // If there are no menu-items yet, they might be added later (e.g., lazy loaded),
-            // so preserve the value to allow matching when items are added.
-            if (this.menuItems.length > 0) {
+            // Only clear value when items exist with real values but none match.
+            // Preserve value if items are pending (lazy loaded, async render, incomplete upgrade).
+            const hasItemsWithValues = this.menuItems.some(
+                (item) => item.value && item.value.trim() !== ''
+            );
```
Rajdeepc, line 921: This is good but it’s still unsafe in a custom-elements lifecycle. This assumes `item.value` is always present and authoritative. `<sp-menu-item>` can exist in the DOM before it is upgraded. In that state,` item.value` is undefined even if the element already has a value attribute. This will therefore return `false` for items that do have meaningful values at the DOM level. We should check both the property and the attribute, and only care about existence.
```ts
const hasItemsWithValues = this.menuItems.some(
    (item) =>
        (item as any).value != null ||
        item.getAttribute?.('value') != null
);
```

Rajdeepc (changes requested):

1st-gen/packages/picker/src/Picker.ts
rubencarvalho: Nice call! Done!

PR #5976, opened 2026-01-21
fix(table): add accessible labels to table checkboxes for WCAG 4.1.2 compliance

Rajdeepc:
## Description
This PR fixes an accessibility violation where table checkbox inputs were missing accessible labels, causing the axe DevTools "Form elements must have labels" error (WCAG 4.1.2 - Name, Role, Value).

### Changes made:
**`TableCheckboxCell.ts:`**

- Added label property to accept an accessible name for the checkbox
- The label is applied via aria-label directly on the checkbox's internal input element
- Added `updateInputAriaLabel()` method to set the `aria-label` after the checkbox renders
- Fixed `aria-hidden` attribute to only be set when selectsSingle is true (was incorrectly setting `aria-hidden="false"` when false)

**`Table.ts:`**

- Added `selectAllLabel` property (default: "Select All") for the header checkbox's accessible name
- Added itemLabel function property for virtualised tables to extract labels from item data (default: "Select row N")
- Added `getRowCheckboxLabel()` helper that extracts the first `<sp-table-cell>` text content for body row checkboxes
- Updated all checkbox creation points to pass appropriate labels to both new and existing checkboxes

## Motivation and context
Users with screen readers were unable to identify what each checkbox in the table controls because the checkboxes had no accessible names. This is a WCAG 4.1.2 (Level A) violation that affects users:

- Without vision
- With limited vision
- With limited manipulation
- With limited reach and strength
- With limited language, cognitive, and learning abilities

The axe DevTools extension flagged "Form elements must have labels" errors on the table component documentation page.

## Related issue(s)

-   fixes 1189

---

## Author's checklist

---

## Reviewer's checklist

### Manual review test cases
- [x] Verify header checkbox has "Select All" label in multiple selection mode

1. Go to [Table component Storybook](https://swcpreviews.z13.web.core.windows.net/pr-5976/docs/first-gen-storybook/?path=/story/table--selects-multiple)
2. Open Chrome DevTools (F12) → axe DevTools tab
3. Run accessibility scan
4. Expect: No "Form elements must have labels" errors for the header checkbox
5. Inspect the header checkbox element
6. Expect: The checkbox contains <span class="visually-hidden">Select All</span>

- [x] Verify single selection mode checkboxes also have labels

1. Go to [Table component Storybook](https://swcpreviews.z13.web.core.windows.net/pr-5976/docs/first-gen-storybook/?path=/story/table--selects-single)
2. Open Chrome DevTools → axe DevTools tab
3. Run accessibility scan
4. Expect: No "Form elements must have labels" errors (even though checkboxes are visually hidden in single-select mode)

- [x] Verify custom select-all-label attribute works

1. In Storybook or local dev, add select-all-label="Select all items" to an <sp-table>
2. Inspect the header checkbox
3. Expect: The checkbox contains <span class="visually-hidden">Select all items</span>

- [x] Verify visual appearance is unchanged

1. Compare table checkboxes before and after this change
2. Expect: No visible difference in checkbox appearance or layout

### Device review


1st-gen/packages/table/README.md
```diff
@@ -93,6 +93,34 @@ A table consists of the following parts:
...
+            <sp-table-cell>PDF</sp-table-cell>
+        </sp-table-row>
+        <sp-table-row value="row2">
+            <sp-table-cell>Onboarding</sp-table-cell>
+            <sp-table-cell>XLS</sp-table-cell>
+        </sp-table-row>
+    </sp-table-body>
+</sp-table>
+```
+
```
nikkimk, line 123: Thanks for updating the docs with this PR! :)

1st-gen/packages/table/README.md
```diff
@@ -829,7 +865,31 @@ When using row selection:
 
 - `aria-selected` is applied to selectable rows
 - Selection state is announced to screen readers
-- Checkboxes in selection cells are properly labeled
+- Checkboxes in selection cells are properly labeled for screen readers
```
nikkimk:
```suggestion
- Checkboxes in selection cells are properly labelled for screen readers
```

1st-gen/packages/table/src/TableCheckboxCell.ts
```diff
@@ -59,18 +60,52 @@ export class TableCheckboxCell extends SpectrumElement {
     @property({ type: Boolean, reflect: true })
     public emphasized = false;
 
+    /**
+     * The accessible label for the checkbox. For header rows, this defaults to 'Select All'.
+     * For body rows, this should be set to the text content of the first cell in the row.
+     */
+    @property({ type: String })
+    public label = '';
```
marissahuysentruyt, line 68: Because we render an `sp-checkbox`, that already supports a label, I thought maybe we could hook into that instead. But since it's a slot, I don't think we can. 

Maybe there's another discussion to be had about `aria-label`s on sp-checkbox components? (which I understand is odd, since checkboxes should always have a label anyways...). Just something I was thinking of- not blocking.

1st-gen/packages/table/README.md
```diff
@@ -93,6 +93,34 @@ A table consists of the following parts:
 
 The `selects` attribute enables row selection functionality. When `selects="single"`, users can select one row at a time. When `selects="multiple"`, users can select multiple rows and a checkbox column is automatically added to the table header for select all functionality.
 
+#### Accessible checkbox labels
+
+When using selection, checkboxes are automatically given accessible labels for screen readers:
+
+- **Header checkbox**: Uses the `select-all-label` attribute (defaults to "Select All")
```
marissahuysentruyt: 🤏 tiny nit-pick, just to match the "Body row checkboxes" language. Take it or leave it.

```suggestion
- **Header row checkbox**: Uses the `select-all-label` attribute (defaults to "Select All")
- ```

File: 1st-gen/packages/table/README.md [hunk 5]
```diff
@@ -829,7 +865,31 @@ When using row selection:
...
+
+##### Checkbox labeling
+
+All selection checkboxes have accessible labels to comply with WCAG 4.1.2 (Name, Role, Value). The labels are applied via `aria-label` on the checkbox's internal input element.
+
+| Checkbox location           | Default label             | Customization                    |
+| --------------------------- | ------------------------- | -------------------------------- |
+| Header (select all)         | "Select All"              | Use `select-all-label` attribute |
+| Body rows (non-virtualized) | First cell's text content | Automatic                        |
+| Body rows (virtualized)     | "Select row N"            | Use `itemLabel` property         |
```
marissahuysentruyt, line 878: And so just to clarify- are consumers unable to change the accessible label of a body row checkbox in a non-virtualized table? Am I understanding this correctly?

Can I ask why? I don't understand the difference between virtualized and non-virtualized tables, so that could be part of this question as well.

marissahuysentruyt (commented):
Thanks for working through this Rajdeep! I only had a few questions, and agree with Nikki that we probably should add some tests to cover the new methods. Thank you!

1st-gen/packages/table/src/TableCheckboxCell.ts
Rajdeepc: Good point! You're right that `sp-checkbox` uses a slot for its label, which makes it difficult to hook into from the parent component. I tried using slotted content initially, but it triggered the checkbox's label element to display (since `:host(:empty) label { display: none }` no longer applied), which added unwanted margin to the UI.

The current approach of setting `aria-label` directly on the internal input element works, but I agree it would be cleaner if `sp-checkbox` natively supported an `aria-label` property that gets passed to its internal input.

We could create an issue to discuss adding `aria-label` support to `sp-checkbox` for cases where a visible label isn't appropriate or the label needs to be dynamically generated.

1st-gen/packages/table/README.md
Rajdeepc: Currently non-virtualized table body row checkboxes use the first cell's text content automatically and can't be customised.
**The difference is:**
- Virtualized tables use data objects (items), so we can provide an `itemLabel` function to extract labels
- Non-virtualized tables use HTML markup, so there's no data object to extract from
For most cases, using the first cell's text (e.g., file name, row title) provides a meaningful label. However, if customisation is needed, we could add a `label` attribute to `<sp-table-row>` in a follow-up PR.

1st-gen/packages/table/test/table-selects.test.ts
```diff
@@ -331,4 +331,168 @@ describe('Table Selects', () => {
...
+            await headerCheckboxCell.updateComplete;
+
+            expect(headerCheckboxCell.label).to.equal('Select all files');
+            expect(
+                headerCheckboxCell.checkbox.inputElement.getAttribute(
+                    'aria-label'
+                )
+            ).to.equal('Select all files');
+        });
+
```
marissahuysentruyt, line 396: Cursor's suggestion to test dynamically updating `selectAllLabel` (looks like button and checkbox do something similar for properties like disabled/pending or size): 

```suggestion
it('allows select-all-label to be changed dynamically', async () => {
            const el = await fixture<Table>(html`
                <sp-table
                    selects="multiple"
                    select-all-label="Select all items"
                >
                    <sp-table-head>
                        <sp-table-head-cell>Column Title</sp-table-head-cell>
                    </sp-table-head>
                    <sp-table-body>
                        <sp-table-row value="row1">
                            <sp-table-cell>Row Item Alpha</sp-table-cell>
                        </sp-table-row>
                    </sp-table-body>
                </sp-table>
            `);
            await elementUpdated(el);

            const headerCheckboxCell = el.querySelector(
                'sp-table-head sp-table-checkbox-cell'
            ) as TableCheckboxCell;

            await headerCheckboxCell.updateComplete;

            expect(headerCheckboxCell.label).to.equal('Select all items');
            expect(
                headerCheckboxCell.checkbox.inputElement.getAttribute(
                    'aria-label'
                )
            ).to.equal('Select all items');

            el.selectAllLabel = 'Choose all records';

            await elementUpdated(el);
            await headerCheckboxCell.updateComplete;

            expect(headerCheckboxCell.label).to.equal('Choose all records');
            expect(
                headerCheckboxCell.checkbox.inputElement.getAttribute(
                    'aria-label'
                )
            ).to.equal('Choose all records');
        });
```

1st-gen/packages/table/test/table-selects.test.ts
```diff
@@ -331,4 +331,168 @@ describe('Table Selects', () => {
...
+        it('body row checkbox uses fallback label when first cell is empty', async () => {
+            const el = await fixture<Table>(html`
+                <sp-table selects="multiple">
+                    <sp-table-head>
+                        <sp-table-head-cell>Column Title</sp-table-head-cell>
+                    </sp-table-head>
+                    <sp-table-body>
+                        <sp-table-row value="my-row">
```
marissahuysentruyt, line 450: Another idea from Cursor was - would we ever encounter a scenario where this value might be empty? 

Is that an edge case, though? With both the first cell being empty AND the row value being empty?

PR #5982, opened 2026-01-22
fix(picker): voiceover crash fix

Rajdeepc:
## Description

When VoiceOver is enabled in Safari, the Picker component crashes the page. This was introduced by PR #4396 which refactored the picker's interaction logic.

### Root causes

1. Imperative render inside render method - Using render(container, overlay) inside renderOverlay() caused timing issues with Safari's accessibility tree
2. aria-labelledby with dynamic content - Referencing elements whose content changes during selection crashes VoiceOver
3. aria-controls across shadow DOM - Cross-shadow-root ID references confuse Safari's accessibility layer
4. Popover API's beforetoggle event - Fires during accessibility tree updates, causing crashes
5. Menu moving between DOM contexts - Conditionally rendering menu directly vs. via overlay breaks accessibility assumptions

### Solution

`Picker.ts`

- Use declarative <sp-overlay> template instead of imperative render() call that broke accessibility timing
- Replace aria-labelledby with aria-label to avoid crashes from dynamic content references
- Remove aria-controls since cross-shadow-DOM ID references crash VoiceOver
- Remove aria-labelledby from menu for the same cross-shadow-DOM issue
- Use type='manual' overlay to avoid beforetoggle event that crashes during accessibility updates
- Add @sp-closed event handler as safe replacement for @beforetoggle
- Implement custom light-dismiss via handleDocumentClick to restore click-outside-to-close since type='auto' can't be used
- Always render menu via overlay to prevent menu from moving between DOM contexts

`InteractionController.ts`

- Simplify set open() since overlay is now managed declaratively in template
- Empty initOverlay() since overlay setup moved to Picker's template
- Simplify hostUpdated() to just sync internal state without overlay manipulation
- Add optional chaining for safer property access to optionsMenu and button

## Related issue(s)

-   fixes [Issue Number]

## Screenshots (if appropriate)

---

## Author's checklist

-   [ ] I have read the **[CONTRIBUTING](<(https://github.com/adobe/spectrum-web-components/blob/main/CONTRIBUTING.md)>)** and **[PULL_REQUESTS](<(https://github.com/adobe/spectrum-web-components/blob/main/PULL_REQUESTS.md)>)** documents.
-   [ ] I have reviewed at the Accessibility Practices for this feature, see: [Aria Practices](https://www.w3.org/TR/wai-aria-practices/)
-   [ ] I have added automated tests to cover my changes.
-   [ ] I have included a well-written changeset if my change needs to be published.
-   [ ] I have included updated documentation if my change required it.

[...]

PR #5983, opened 2026-01-23
fix(picker): restore overlay declarative rendering to avoid VO crash on Safari

rubencarvalho:
## Description

Fixes a Safari + VoiceOver crash that occurs when opening Picker and ActionMenu components.
The issue was caused by an imperative `render()` call during the component's render cycle, which created a race condition with VoiceOver's accessibility tree scanning.

## Motivation and context

The issue was introduced in July 2024 ([#4396](https://github.com/adobe/spectrum-web-components/pull/4396/)), as part of the strategy refactor. In that change, overlay rendering moved from a declarative approach to an imperative one, which introduced a pretty fundamental Lit anti-pattern.
Before the refactor, renderOverlay() was doing the "right" thing: it simply returned a template that included <sp-overlay> and let Lit handle rendering in a purely declarative way. After the refactor, though, renderOverlay() started calling render() imperatively inside the render path, manually mutating the overlay’s DOM and then returning the overlay instance cast as a TemplateResult. That combination is problematic for a few reasons: render methods in Lit (like in React and other UI libraries) are expected to be pure and side-effect free, but this approach mutates the DOM during render. On top of that, manually rendering into a Lit component bypasses the framework entirely (and the unsafe casting sidesteps type safety altogether :nervous-kek:).
This is also why the crash only happens under very specific conditions with VoiceOver. When VoiceOver is already enabled and the Picker is opened, the Picker’s render() runs and immediately performs that imperative render() call inside renderOverlay(). At the same time, VoiceOver starts scanning the accessibility tree of the newly opened overlay. Because the DOM is being synchronously mutated while VoiceOver is reading it, we end up with a race condition. Safari doesn’t handle that well and crashes due to the accessibility tree being in an unstable state. If VoiceOver is enabled after the overlay is already open, everything is already rendered and stable, so there’s no crash.
Below I tried to make it easier to visualize what is happening:

```ts
  BEFORE (correct):                                                           
  protected renderOverlay(menu: TemplateResult): TemplateResult {             
      const container = this.renderContainer(menu);                           
      return html`                                                            
          <sp-overlay ?open=${this.open}>                                     
              ${container}                                                    
          </sp-overlay>                                                       
      `;                                                                      
  }                                                                           
                                                                              
  AFTER (broken):                                                             
  protected renderOverlay(menu: TemplateResult): TemplateResult {             
      const container = this.renderContainer(menu);                           
      // ❌ Imperative render() call inside render method (this is a big no-no!)                     
      render(container, this.strategy?.overlay as unknown as HTMLElement, {   
          host: this,                                                         
      });                                                                     
      // ❌ Unsafe cast: returning component instance as TemplateResult       
      return this.strategy?.overlay as unknown as TemplateResult;             
  }
```

## Related issue(s)

-   fixes SWC-1488
-  fixes https://github.com/adobe/spectrum-web-components/pull/5983

## Author's checklist

-   [X] I have read the **[CONTRIBUTING](<(https://github.com/adobe/spectrum-web-components/blob/main/CONTRIBUTING.md)>)** and **[PULL_REQUESTS](<(https://github.com/adobe/spectrum-web-components/blob/main/PULL_REQUESTS.md)>)** documents.

---

## Reviewer's checklist

### Manual review test cases

-   [ ] Test Picker VO on Safari 

    1. Visit the [picker documentation page](https://swcpreviews.z13.web.core.windows.net/pr-5983/docs/first-gen-docs/components/picker/) on Safari
    2. Start VoiceOver (e.g., by pressing CMD + F5)
    3. Tab to the picker and open it. 
    4. Expect no page failure or browser error message

### Device review

### DIFF
--- a/.changeset/good-humans-trade.md
+++ b/.changeset/good-humans-trade.md
@@ -1,0 +1,5 @@
+---
+'@spectrum-web-components/shared': patch
+---
+
+Improve reliability when composing components and mixins that observe slot presence.
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,7 +22,8 @@ parameters:
     # 3. Commit this change to the PR branch where the changes exist.
     current_golden_images_hash:
         type: string
-        default: 4304495a938a5b7dd3e64e12ebc59dfd5e300c1b
+        default: 53d43cef8f05a527cd662707fd6a7a8e930a6a52
+
     wireit_cache_name:
         type: string
         default: wireit
--- a/1st-gen/packages/menu/src/MenuItem.ts
+++ b/1st-gen/packages/menu/src/MenuItem.ts
@@ -395,7 +395,7 @@ export class MenuItem extends LikeAnchor(
                 this.dependencyManager.loaded}
                 .placement=${this.isLTR ? 'right-start' : 'left-start'}
                 receives-focus="false"
-                .offset=${[-10, -5] as [number, number]}
+                .offset=${[-10, 0] as [number, number]}
                 .type=${'auto'}
                 @close=${(event: Event) => event.stopPropagation()}
                 @slottable-request=${this.handleSlottableRequest}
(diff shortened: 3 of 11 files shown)


Rajdeepc:
Related to #5982

PR #5992, opened 2026-01-29
fix(slider): add value tooltips and accessibility improvements for multi-handle sliders

Rajdeepc:
## Description
Fixes accessibility issue for multi-handle sliders per WCAG 3.3.2 (Labels or Instructions)
Multi-handle sliders now display value tooltips on hover/focus when value labels are hidden, ensuring users can always identify individual handle values. This addresses the accessibility concern that multi-handle sliders without visible labels make their purpose unclear to visual users.

### Core functionality

- Value tooltips: Multi-handle sliders with `label-visibility="none"` or `label-visibility="text"` now show tooltips displaying the current value for each handle on hover/focus
- Debug warning: Added dev mode warning when multi-handle slider handles are missing label attributes (with optimization to warn only once per slider instance)

### Accessibility enhancements

- Screen reader optimisation: Tooltips use `aria-hidden="true"` since `aria-valuetext` on the input already announces values to assistive technology
- High contrast mode: Value tooltips support Windows High Contrast Mode via forced-colors media query
- Reduced motion: Respects prefers-reduced-motion user preference by disabling transitions
- Disabled state: Tooltips are hidden when the slider is disabled

### RTL support

- Uses CSS logical properties `(inset-inline-start)` for proper right-to-left layout support
- Correct transform direction for RTL layouts

## Motivation and context

An accessibility audit identified that multi-handle sliders in the "Handle slot" section at `opensource.adobe.com/spectrum-web-components/components/slider/` do not have visible labels, making their purpose unclear to visual users.

**Affected users:**

- Users without vision (screen reader users)
- Users with limited vision
- Users with limited manipulation or reach
- Users with cognitive or learning disabilities

WCAG violation: This fails WCAG 3.3.2 Labels or Instructions (Level A), which requires that labels or instructions are provided when content requires user input.

## Related issue(s)

-   fixes SWC-1116

Screenshot 2026-02-02 at 8 17 22 PM

---

## Author's checklist

---

## Reviewer's checklist

### Manual review test cases

-   [x] Screen reader testing (voiceOVer/NVDA) (@rise-erpelding)

    1. Go [here](https://swcpreviews.z13.web.core.windows.net/pr-5992/docs/first-gen-storybook/?path=/story/slider--two-handles-with-value-tooltips)
    2. Tab to multi-handle slider
    3. Expected: Announces handle label and value
    4. Use Arrow keys to change value
    5. Expected: Announces new value

-   [x] Dev mode warning (@rise-erpelding)
    1. Open browser console and add a multi-handler slider without labels
    ```html
    <sp-slider>
       <sp-slider-handle slot="handle" name="min" value="25"></sp-slider-handle>
       <sp-slider-handle slot="handle" name="max" value="75"></sp-slider-handle>
    </sp-slider>
    ```
    2. Expected: Console warning: "Multi-handle sliders require a label attribute"
    3. Warning appears only once per slider instance

### Device review

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,7 +22,7 @@ parameters:
     # 3. Commit this change to the PR branch where the changes exist.
     current_golden_images_hash:
         type: string
-        default: c420a27826be5bc391cddeac8ba3de0872c725eb
+        default: c152abbac4386cb278f987a8eb09f4a8fb5fc59d
 
     wireit_cache_name:
         type: string


1st-gen/packages/slider/src/HandleController.ts
```diff
@@ -710,6 +732,26 @@ export class HandleController {
...
+        if (
+            window.__swc?.DEBUG &&
+            isMultiHandle &&
+            !this._hasWarnedAboutMissingLabels
+        ) {
+            const handlesWithoutLabels = handles.filter(
+                (handle) => handle !== this.host && !handle.label?.length
+            );
+            if (handlesWithoutLabels.length > 0) {
+                this._hasWarnedAboutMissingLabels = true;
```
caseyisonit, line 744: Can you explain why we are using a private variable in the debug pattern? is this a pattern we want to consider adopting widely?

Rajdeepc: Yes. I have used a private flag to ensure that it’s per-instance state, not an API. Using a private label makes it explicit that it’s internal and not something to rely on. 
Its not global since we want the warning once per slider and not once per page load. Private instance state is the most optimised way to do this.
To your question: If Its a good pattern to adopt widely **yes**! but with a consistent helper. We can use this more  by having a  centralize "warn once per instance" in a helper like `warnOnce(this, 'missing-labels', ....)`.
Let me know how you think about this. Happy to know your thoughts.

1st-gen/packages/slider/src/slider.css
```diff
@@ -155,3 +155,78 @@
...
     margin-block-start: 0;
 }
+
+/* Value tooltip for multi-handle sliders when value label is not visible */
+.value-tooltip {
+    position: absolute;
+    bottom: calc(100% + 8px);
+    inset-inline-start: 50%;
+    transform: translateX(-50%);
+    padding: var(--spectrum-global-dimension-size-50, 4px) var(--spectrum-global-dimension-size-100, 8px);
```
rise-erpelding: These `--spectrum-global` prefixed tokens are really old, could we use more modern tokens instead? 

Looking through the tooltip CSS, it looks like many of those custom properties are defined there so they won't work here without adding a dependency on tooltip, but we could use the same global tokens that `--spectrum-tooltip-spacing-inline` and `--spectrum-tooltip-spacing-block-start` + `--spectrum-tooltip-spacing-block-end` rely on (that would be `--spectrum-component-edge-to-text-75`, `--spectrum-component-top-to-text-75`, `--spectrum-component-bottom-to-text-75`).

I see that changes the padding a bit from what it was though -- is that an issue?

1st-gen/packages/slider/src/slider.css
```diff
@@ -155,3 +155,78 @@
...
+    background-color: var(--spectrum-tooltip-background-color, var(--spectrum-global-color-gray-900));
+    color: var(--spectrum-tooltip-text-color, var(--spectrum-global-color-gray-50));
+    font-size: var(--spectrum-global-dimension-font-size-75, 12px);
```
rise-erpelding:
```suggestion
    font-size: var(--spectrum-font-size-75, 12px);
```

1st-gen/packages/slider/src/slider.css
```diff
@@ -155,3 +155,78 @@
...
+    line-height: var(--spectrum-line-height-100, 1.3);
+    border-radius: var(--spectrum-global-dimension-size-50, 4px);
```
rise-erpelding: For this one it looks like `--spectrum-corner-radius-100` would change drastically depending on the theme, for Spectrum 1 it's 4px for medium but it looks like it's 8px in Spectrum 2... but it does align with the rounding seen in `sp-tooltip`.

Do we need the fallback values for these styles?

```suggestion
    border-radius: var(--spectrum-corner-radius-100, 4px);
```

Rajdeepc:
> Confirmed that everything looks good as far as validation steps with screen-reader and the dev mode warning.
> 
> It also looks good in WHCM and RTL!
> 
> Left some comments about the CSS tokens.
> 
> I'm also not familiar enough with `sp-tooltip`, but it seems like there's a reason to use a custom tooltip here instead of `sp-tooltip`?

Great question! I am glad you asked. One alternative I thought about is that we could wrap the input differently or use `sp-tooltip` with explicit `trigger` targeting, but this would require more significant changes to the handle rendering structure and potentially might affect existing styling/behavior. The CSS approach is providing the same visual result with less complexity and better integration with the existing slider architecture. Let me know what do you think?

1st-gen/packages/slider/src/slider.css
```diff
@@ -155,3 +155,79 @@
...
+.value-tooltip {
+    position: absolute;
+    bottom: calc(100% + 8px);
+    inset-inline-start: 50%;
+    transform: translateX(-50%);
+    padding-block-start: var(--spectrum-component-top-to-text-75, 4px);
+    padding-block-end: var(--spectrum-component-bottom-to-text-75, 5px);
+    padding-inline: var(--spectrum-component-edge-to-text-75, 9px);
+    background-color: var(--spectrum-tooltip-background-color, var(--spectrum-global-color-gray-900));
+    color: var(--spectrum-tooltip-text-color, var(--spectrum-global-color-gray-50));
```
rise-erpelding: A couple more `--spectrum-global` tokens here that could use replacing. Also, these `--spectrum-tooltip-` tokens are defined in `sp-tooltip`, I'm not seeing them work here. Would they be expected to work in a different environment (other than Storybook)? If not I'd suggest removing them.

```suggestion
    background-color: var(--spectrum-gray-900);
    color: var(--spectrum-gray-50);
```

1st-gen/packages/slider/src/slider.css
```diff
@@ -155,3 +155,79 @@
...
+/* Tooltip arrow/tip */
+.value-tooltip:after {
+    content: "";
+    position: absolute;
+    top: 100%;
+    inset-inline-start: 50%;
+    transform: translateX(-50%);
+    border-width: 4px;
+    border-style: solid;
+    border-color: var(--spectrum-tooltip-background-color, var(--spectrum-global-color-gray-900)) transparent transparent;
```
rise-erpelding: Same for this border-color, the `--spectrum-tooltip` color isn't defined and the fallback is a `--spectrum-global` token:


```suggestion
    border-color: var(--spectrum-gray-900) transparent transparent;
```

1st-gen/packages/slider/src/HandleController.ts
caseyisonit: oh love the idea of a helper!! thanks for the explanation!

blunteshwar:
Feel free to merge after updating the golden image hash.

PR #5993, opened 2026-01-29
fix(build): replace wildcard exports and downgrade to ES2018 for compatibility

rubencarvalho:
## Description

This PR addresses multiple customer-reported build and compatibility issues by improving module exports and adjusting build targets:
### Export improvements
1. Replaced `core` wildcard exports with explicit named exports in:
- `2nd-gen/packages/core/package.json` - Replaced wildcard patterns with explicit export paths
2. Fixed dependency resolution - Added `@spectrum-web-components/core` as a direct dependency to `@spectrum-web-components/shared`. The shared package imports from core but was only accessing it as a transitive dependency through `@spectrum-web-components/base`, causing module resolution failures in strict environments.

### Build target changes
3. Changed `ES2022` → `ES2018` for better compatibility:
- `2nd-gen/tsconfig.json` - Changed TypeScript target
- `2nd-gen/packages/core/vite.config.js` - Updated build and esbuild targets
- `2nd-gen/packages/swc/vite.config.ts` - Updated build and esbuild targets
4. Added missing dependency - Added `@lit-labs/observers` to `2nd-gen/packages/core` and externalized it in the Vite config

## Motivation and context

Following up on customer-reported build issues discovered while creating snapshot releases and investigating their projects:
- PSWeb: Vitest environment does not support ES2022 build target (production builds work as expected)
- Horizon: Some bundler environments do not support wildcard (*) exports in JavaScript files
- Next.js < v13.1: Does not support wildcard patterns in the package.json exports field
- Module resolution: Strict dependency boundaries in some environments fail when packages rely on transitive dependencies

## Related issue(s)

-   Addresses customer-reported issues from PSWeb, Horizon, and Next.js < v13.1 users

## Author's checklist

---

## Reviewer's checklist

### Manual review test cases

-   [ ] No functional or visual changes expected
-   [ ] Verify 1st-gen components still build and import correctly
1. Run yarn build in the workspace root
2. Check that all 1st-gen components compile without errors
3. Verify no type errors in consuming packages
-   [ ] Verify 2nd-gen components still build and import correctly
1. Run yarn build in 2nd-gen/
2. Check that all components compile with ES2018 target

### DIFF
--- a/.changeset/table-checkbox-accessibility-fix.md
+++ b/.changeset/table-checkbox-accessibility-fix.md
@@ -1,0 +1,5 @@
+---
+'@spectrum-web-components/table': patch
+---
+
+**Fixed**: Fixed accessibility violation (WCAG 4.1.2) where table checkbox inputs were missing accessible labels. The axe DevTools "Form elements must have labels" error is now resolved. The fix sets `aria-label` directly on the checkbox's internal input element.
--- a/1st-gen/packages/table/README.md
+++ b/1st-gen/packages/table/README.md
@@ -93,6 +93,34 @@ A table consists of the following parts:
 
 The `selects` attribute enables row selection functionality. When `selects="single"`, users can select one row at a time. When `selects="multiple"`, users can select multiple rows and a checkbox column is automatically added to the table header for select all functionality.
 
+#### Accessible checkbox labels
+
+When using selection, checkboxes are automatically given accessible labels for screen readers:
+
+- **Header row checkbox**: Uses the `select-all-label` attribute (defaults to "Select All")
+- **Body row checkboxes**: Uses the text content of the first `<sp-table-cell>` in each row
+
+You can customize the header checkbox label using the `select-all-label` attribute:
+
+```html demo
+<sp-table selects="multiple" select-all-label="Select all files">
+    <sp-table-head>
+        <sp-table-head-cell>File name</sp-table-head-cell>
+        <sp-table-head-cell>Type</sp-table-head-cell>
+    </sp-table-head>
+    <sp-table-body>
+        <sp-table-row value="row1">
+            <sp-table-cell>Budget</sp-table-cell>
+            <sp-table-cell>PDF</sp-table-cell>
+        </sp-table-row>
+        <sp-table-row value="row2">
+            <sp-table-cell>Onboarding</sp-table-cell>
+            <sp-table-cell>XLS</sp-table-cell>
+        </sp-table-row>
+    </sp-table-body>
+</sp-table>
+```
+
 ```html demo
 <sp-table selects="multiple" selected='["row1", "row2"]'>
     <sp-table-head>
@@ -474,6 +502,14 @@ Please note that there is a bug when attempting to select all virtualised elemen
 
 By default the `selected` property will surface an array of item indexes that are currently selected. However, when making a selection on a virtualized table, it can be useful to track selection as something other than indexes. To do so, set a custom method for the `itemValue` property. The `itemValue` method accepts an item and its index as arguments and should return the value you would like to track in the `selected` property.
 
+##### Accessible labels for virtualized tables
+
+For accessibility, each checkbox in a virtualized table needs an accessible label. By default, the label is "Select row N" where N is the row number. You can customize this using the `itemLabel` property to provide more meaningful labels based on your item data:
+
+```javascript
+table.itemLabel = (item, index) => item.name || `Select row ${index + 1}`;
+```
+
 ```html-live
 <sp-table
     id="table-item-value-demo"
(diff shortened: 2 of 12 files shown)


1st-gen/tools/base/src/index.ts
```diff
@@ -10,7 +10,13 @@
...
+export { SpectrumElement, SpectrumMixin } from './Base.js';
+export type { SpectrumInterface } from './Base.js';
+export { ElementSizes, SizedMixin } from './sizedMixin.js';
+export type {
+    DefaultElementSize,
+    ElementSize,
+    SizedElementInterface,
+} from './sizedMixin.js';
+export { INPUT_COMPONENT_TAGS, INPUT_COMPONENT_PATTERN } from './constants.js';
 export * from 'lit';
```
Rajdeepc, line 16: This should also cause issues in the same bundler environment

2nd-gen/packages/core/package.json
```diff
@@ -15,29 +15,55 @@
...
-        "./shared/base/*.js": {
-            "types": "./dist/shared/base/*.d.ts",
-            "import": "./dist/shared/base/*.js"
+        "./shared/observe-slot-text.js": {
+            "types": "./dist/shared/observe-slot-text.d.ts",
+            "import": "./dist/shared/observe-slot-text.js"
         }
     },
-    "main": "./dist/index.js",
-    "module": "./dist/index.js",
```
Rajdeepc, line 40: Any customers that are using root imports? This would break consumers if anyone is wiring via `import { something } from '@spectrum-web-components/core'`. 
Why can't we have a barrel `index.ts` which re-exports from subpaths?

2nd-gen/packages/core/package.json
```diff
@@ -15,29 +15,55 @@
...
+            "types": "./dist/shared/base/version.d.ts",
+            "import": "./dist/shared/base/version.js"
+        },
+        "./shared/utilities": {
+            "types": "./dist/shared/utilities/index.d.ts",
+            "import": "./dist/shared/utilities/index.js"
+        },
+        "./shared/utilities/index.js": {
+            "types": "./dist/shared/utilities/index.d.ts",
+            "import": "./dist/shared/utilities/index.js"
```
Rajdeepc, line 68: non-blocking: We should think of automating export map generation here. A script could generate both the directory and file variants automatically in the build pipeline or else this will be hard to maintain.

.changeset/explicit-exports-es2018.md
```diff
@@ -0,0 +1,10 @@
+---
+'@spectrum-web-components/core': patch
+'@spectrum-web-components/shared': patch
+'@spectrum-web-components/base': patch
+---
+
+- **Fixed**: Replaced all wildcard exports with explicit named exports for better bundler compatibility and tree-shaking
+- **Fixed**: Changed build target from ES2022 to ES2018 to support Vitest and other test environments
+- **Fixed**: Added `@spectrum-web-components/core` as direct dependency to `@spectrum-web-components/shared` to resolve module resolution issues in strict dependency environments
+- **Fixed**: Added `@lit-labs/observers` as dependency and externalized it in Vite build config
```
Rajdeepc: Changeset doesn't mention `export { getDeepElementFromPoint } from './get-deep-element-from-point.js';`

2nd-gen/packages/core/package.json
```diff
@@ -15,29 +15,55 @@
...
     "type": "module",
     "exports": {
-        ".": {
-            "types": "./dist/index.d.ts",
-            "import": "./dist/index.js"
-        },
         "./components/*": {
             "types": "./dist/components/*/index.d.ts",
             "import": "./dist/components/*/index.js"
         },
```
Rajdeepc: Is this acceptable in Next < 13.1?

Rajdeepc (changes requested):
blocking:
1. I pulled down your code and found that in `like-anchor.ts` exports `LikeAnchorInterface` but `RenderAnchorOptions` (used in the public `renderAnchor()` method signature) is not exported from `index.ts`. TypeScript consumers won't be able to properly type the options parameter.

2. can you align `@lit-labs/observers` version in `2nd-gen/packages/core` to follow `2.0.2`. This can cause duplicate code in bundles with `1st-gen`
3. Since you downgraded esbuild to 2018, you should use lib to prevent usage of new APIs in `tsconfig.json`. It still shows 2022.
```ts
{
  "compilerOptions": {
    "target": "ES2018",
    "lib": ["ES2018", "DOM", "DOM.Iterable"]
  }
}
```

2nd-gen/packages/core/package.json
rubencarvalho: These lines were redundant because we are not barrel rolling any exports. Currently `@spectrum-web-components/core` is purely internal and so and we have always imported from the full paths.
But to answer your question, we can have a barrel `index.ts`! We just haven't done it 🤷 May happen on the "figure out core folder structure" ticket we'll be working on.

2nd-gen/packages/core/package.json
rubencarvalho: Yes - I agree! This needs figuring out. For now, I tried to only focus on the fix. I assume the folder structure will change and we have a dedicated ticket for it, so I think that will be tackled then.

2nd-gen/packages/core/package.json
rubencarvalho: Good call - I'll expand this!

1st-gen/tools/base/src/index.ts
rubencarvalho: I think it only did in combination with `core` being a dependency and its build target being ES2022. We haven't had any reports of this (+ teams validated this build).

PR #6012, opened 2026-02-06
chore: migrate icon to core and second-gen

TarunAdobe:
## Description

This PR introduces a minimal icon component for 2nd-gen in `@adobe/swc`, following the RFC direction.

It provides a consistent way for 2nd-gen components to render icons via:

- Slotted inline SVG templates

It also adds a small shared icon template catalog under `@adobe/swc/icon/elements` to avoid per-component SVG duplication.

## Motivation and context

This change unblocks migration of icon-dependent components to 2nd-gen without relying on 1st-gen icon packages and global iconset infrastructure.

The goal is to keep icon usage local, explicit, and consistent across 2nd-gen components.

## Related issue(s)

- RFC: minimal icon component for 2nd-gen

N/A

---

## Author's checklist

- [x] I have reviewed accessibility expectations for this feature.
- [x] I have added automated coverage for the icon behavior.
- [x] I have included a changeset if this change needs publishing.
- [x] I have updated relevant Storybook documentation/examples.

---

## Reviewer's checklist

- [x] Scope aligns with RFC for minimal 2nd-gen icon support.
- [x] `swc-icon` behavior is correct for slot/SVG and `src` fallback.
- [x] Shared icon catalog exports are usable (`@adobe/swc/icon/icons`).
- [ ] Automated tests cover expected behavior and pass in CI.
- [ ] VRT/browser validation is completed as required.

### Manual review test cases

- [ ] Review `swc-icon` Storybook docs/examples

  1. Open icon docs in the PR Storybook build.
  2. Verify slotted SVG rendering works (default slot path).
  3. Verify shared template usage examples (`Chevron100Icon`, `AlertIcon`) render correctly.
  4. Inspect DOM to confirm:
     - Slotted SVG receives accessibility attributes as expected

### Device review

- [ ] Desktop
- [ ] Emulated mobile
- [ ] Emulated tablet

### DIFF
--- a/2nd-gen/packages/swc/components/icon/elements/AlertIcon.ts
+++ b/2nd-gen/packages/swc/components/icon/elements/AlertIcon.ts
@@ -13,10 +13,7 @@ import { html, TemplateResult } from 'lit';
 
 export const AlertIcon = (): TemplateResult => {
   return html`
-    <svg
-      xmlns="http://www.w3.org/2000/svg"
-      viewBox="0 0 36 36"
-    >
+    <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 36 36">
       <path
         d="M17.127 2.579.4 32.512A1 1 0 0 0 1.272 34h33.456a1 1 0 0 0 .872-1.488L18.873 2.579a1 1 0 0 0-1.746 0ZM20 29.5a.5.5 0 0 1-.5.5h-3a.5.5 0 0 1-.5-.5v-3a.5.5 0 0 1 .5-.5h3a.5.5 0 0 1 .5.5Zm0-6a.5.5 0 0 1-.5.5h-3a.5.5 0 0 1-.5-.5v-12a.5.5 0 0 1 .5-.5h3a.5.5 0 0 1 .5.5Z"
       />


1st-gen/tools/base/src/version.js
rubencarvalho, line 1: Can you verify why this file is showing?

2nd-gen/packages/swc/components/icon/stories/icon.stories.ts
```diff
@@ -0,0 +1,230 @@
...
+  },
+};
+
+/**
+ * Use the shared icon catalog to keep icon usage consistent across components.
+ *
+ * Example import:
+ *
+ * ```ts
+ * import { Chevron100Icon } from '@adobe/swc/icon/elements';
```
rubencarvalho:
```suggestion
 * import { Chevron100Icon } from '@adobe/spectrum-wc/icon/elements';
```

2nd-gen/packages/swc/components/icon/stories/icon.stories.ts
```diff
@@ -0,0 +1,230 @@
...
+};
+
+// ──────────────────────────
+//    OPTIONS STORIES
+// ──────────────────────────
+
+/**
+ * ### Shared templates
+ *
+ * Import reusable templates from `@adobe/swc/icon/elements` and slot them into `<swc-icon>`.
```
rubencarvalho:
```suggestion
 * Import reusable templates from `@adobe/spectrum-wc/icon/elements` and slot them into `<swc-icon>`.
```

2nd-gen/packages/swc/components/icon/stories/icon.internal.stories.ts
```diff
@@ -0,0 +1,230 @@
...
+//    METADATA
+// ────────────────
+
+const { args, argTypes, template } = getStorybookHelpers('swc-icon');
+
+/**
+ * **Internal-only component.**
+ *
+ * The `<swc-icon>` element renders icons from shared inline SVG templates.
+ * Use shared templates from `@adobe/swc/icon/elements` for consistent rendering and avoid duplicating SVG markup in each component.
```
rubencarvalho:
```suggestion
 * Use shared templates from `@adobe/spectrum-wc/icon/elements` for consistent rendering and avoid duplicating SVG markup in each component.
```

2nd-gen/packages/swc/components/icon/Icon.ts
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * @example
+ * <swc-icon label="Search">
+ *   <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16">
+ *     <path d="M14.5 13.09 11.41 10a6 6 0 1 0-1.41 1.41l3.09 3.09a1 1 0 0 0 1.41-1.41zM3 7a4 4 0 1 1 8 0 4 4 0 0 1-8 0z"/>
+ *   </svg>
+ * </swc-icon>
+ *
+ * @example
+ * import { Chevron100Icon } from '@adobe/swc/icon/elements';
```
rubencarvalho:
```suggestion
 * import { Chevron100Icon } from '@adobe/spectrum-wc/icon/elements';
```

2nd-gen/packages/swc/components/icon/test/icon.test.ts
```diff
@@ -0,0 +1,94 @@
...
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+import { html } from 'lit';
+import { expect } from '@storybook/test';
+import type { Meta, StoryObj as Story } from '@storybook/web-components';
+
+import { Icon } from '@adobe/swc/icon';
+
+import '@adobe/swc/icon';
```
rubencarvalho:
```suggestion
import '@adobe/spectrum-wc/icon';
```

2nd-gen/packages/swc/components/icon/test/icon.test.ts
```diff
@@ -0,0 +1,94 @@
...
+  play: async ({ canvasElement, step }) => {
+    const icon = await getComponent<Icon>(canvasElement, 'swc-icon');
+
+    await step('reflects size attribute on host', async () => {
+      expect(icon.getAttribute('size')).toBe('xl');
+    });
+  },
+};
+
+export const SlottedSvgAccessibilityTest: Story = {
```
rubencarvalho, line 78: Maybe we need the A11Y comment above this one

rubencarvalho: And we probably want to add a test for no label -> `aria-hidden="true"`

2nd-gen/packages/swc/.storybook/main.ts
```diff
@@ -18,11 +18,32 @@ import { mergeConfig } from 'vite';
 
 const __dirname = dirname(fileURLToPath(import.meta.url));
 const includeTestStories = process.env.NODE_ENV !== 'production';
+const isProductionBuild = process.env.NODE_ENV === 'production';
```
rubencarvalho, line 21: Not sure if this behaves as we expect it to. Currently I see the icons showing on the deployed Storybook. We may want to explore an alternative. Maybe we can lean into `internal.stories.ts` instead? Up to you, but please verify 😄

2nd-gen/packages/core/components/icon/Icon.base.ts
```diff
@@ -0,0 +1,92 @@
...
+ * @attribute {string} label - Accessible label for the icon.
+ * @attribute {string} size - T-shirt icon size.
+ */
+export abstract class IconBase extends SpectrumElement {
+  // ──────────────────
+  //     SHARED API
+  // ──────────────────
+
+  /**
+   * Accessible label for the icon.
```
rubencarvalho, line 34: nit: we may want to work on these descriptions - we're pending some docs around this - but I imagine we'll want to be more descriptive in the future.

2nd-gen/packages/core/components/icon/Icon.base.ts
```diff
@@ -0,0 +1,92 @@
...
+  // ──────────────────────
+  //     IMPLEMENTATION
+  // ──────────────────────
+
+  protected override updated(changedProperties: PropertyValues): void {
+    super.updated(changedProperties);
+    if (changedProperties.has('label')) {
+      this.updateSlottedIcon();
+    }
+    this.updateHostAccessibility();
```
rubencarvalho: Does this need to run on every `updated`? Or could this move up to inside the `label` changed guard above?

2nd-gen/packages/swc/components/icon/test/icon.test.ts
rubencarvalho: and potentially toggling the label, e.g. label: 'x' -> '' -> 'y'

2nd-gen/packages/core/components/icon/Icon.base.ts
TarunAdobe: No. Moved updateHostAccessibility() to only run when label changes

2nd-gen/packages/swc/.storybook/main.ts
TarunAdobe: Updated to now use .internal.stories.ts naming convention to exclude stories from production builds and added NODE_ENV var to build command

2nd-gen/packages/swc/components/icon/test/icon.test.ts
TarunAdobe: done

2nd-gen/packages/swc/package.json
```diff
@@ -49,7 +49,7 @@
     "dev": "vite build --watch",
     "dev:core": "yarn workspace @spectrum-web-components/core dev",
     "storybook": "yarn analyze && storybook dev -p 6006",
-    "storybook:build": "yarn analyze && storybook build",
+    "storybook:build": "yarn analyze && NODE_ENV=production storybook build",
```
rubencarvalho: I tested and you don't need this. NODE_ENV gets injected automatically 😄

2nd-gen/packages/swc/cem.config.js
```diff
@@ -14,6 +14,7 @@ export default {
   globs: [
     'components/**/*.ts',
     '../core/components/**/*.ts',
+    '../core/mixins/**/*.ts',
     '../core/shared/**/*.ts',
```
rubencarvalho: Can you update all paths to match all paths in the new folder structure, please?


```suggestion
    'components/**/*.ts',
    '../core/components/**/*.ts',
    '../core/controllers/**/*.ts',
    '../core/element/**/*.ts',
    '../core/mixins/**/*.ts',
    '../core/utils/**/*.ts',
```

2nd-gen/packages/swc/components/icon/stories/icon.internal.stories.ts
```diff
@@ -0,0 +1,244 @@
...
+  `,
+  tags: ['anatomy'],
+  parameters: {
+    flexLayout: true,
+  },
+};
+
+// ──────────────────────────
+//    OPTIONS STORIES
+// ──────────────────────────
```
rubencarvalho, line 140: According to the existing patterns for story documentation, all size options must be documented in a single `Sizes` story with proper labels and `flexLayout` for comparison. Could we do the same here?

2nd-gen/packages/core/components/icon/Icon.base.ts
```diff
@@ -0,0 +1,92 @@
...
+   * Accessible label for the icon.
+   */
+  @property()
+  public label = '';
+
+  /**
+   * Icon t-shirt size.
+   */
+  @property({ reflect: true })
+  public size: IconSize = 'm';
```
Rajdeepc: Other 2nd-gen components uses `SizedMixin`

2nd-gen/packages/core/components/icon/Icon.base.ts
```diff
@@ -0,0 +1,92 @@
...
+  private defaultSlot?: HTMLSlotElement;
+
+  // ──────────────────────
+  //     IMPLEMENTATION
+  // ──────────────────────
+
+  protected override updated(changedProperties: PropertyValues): void {
+    super.updated(changedProperties);
+    if (changedProperties.has('label')) {
+      this.updateSlottedIcon();
```
Rajdeepc, line 55: What about the initial render when a user provides
```ts
<swc-icon><svg>...</svg></swc-icon>
```

Can you override `firstUpdated()` to call `updateSlottedIcon()` and `updateHostAccessibility()` to ensure the initial state is correct regardless of property change timing. Also please think of using `@queryAssignedElements` instead of manually querying the slot.

2nd-gen/packages/swc/components/icon/elements/AlertIcon.ts
```diff
@@ -0,0 +1,25 @@
+/**
```
Rajdeepc, line 1: These are pure SVG template functions with no rendering layer dependency. Do you feel that we should move icon element templates to `2nd-gen/packages/core/components/icon/elements/` or a separate `core/icons/` directory?

2nd-gen/packages/swc/components/icon/elements/Chevron100Icon.ts
```diff
@@ -0,0 +1,40 @@
...
+ * governing permissions and limitations under the License.
+ */
+import { html, TemplateResult } from 'lit';
+
+export type Chevron100IconOptions = {
+  direction?: 'right' | 'left' | 'up' | 'down';
+};
+
+export const Chevron100Icon = ({
+  direction = 'right',
```
Rajdeepc, line 19: This is the only icon which is using an inline style transform. There is an inconsistency. We can either make direction a concern of the `swc-icon` host via a CSS approach or we can document why Chevron is special.

TarunAdobe: the rationale here is that we don't want another component to control the icon css or how icon should render.. this should be isolated in the icon template literal itself and so optionally some icons can have these attributes like direction in this case. it was an intentional choice

2nd-gen/packages/swc/components/icon/stories/icon.internal.stories.ts
TarunAdobe: Screenshot 2026-03-03 at 11 55 32 AM

Done

2nd-gen/packages/core/components/icon/Icon.base.ts
TarunAdobe: shoot. this was a big deal oops. got it right now

2nd-gen/packages/swc/components/icon/elements/AlertIcon.ts
TarunAdobe: I agree the icon templates are conceptually “core-ish”, but right now they return Lit TemplateResult (html\\``), so they still carry a rendering-layer dependency.

For this PR, I’d prefer to keep them under @adobe/spectrum-wc/icon/elements to avoid mixing concerns and expanding scope (move would also require export/typesVersions + import-path updates).

I’m happy to do a follow-up where we decide the long-term home (core/components/icon/elements vs core/icons) and, if needed, first make the templates rendering-agnostic so the move is clean.

1st-gen/tools/base/src/version.js
rubencarvalho: BLOCKING - can you verify, please?

2nd-gen/packages/swc/components/icon/stories/icon.internal.stories.ts
```diff
@@ -0,0 +1,282 @@
...
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+import { html } from 'lit';
+import { ifDefined } from 'lit/directives/if-defined.js';
+import type { Meta, StoryObj as Story } from '@storybook/web-components';
+import { getStorybookHelpers } from '@wc-toolkit/storybook-helpers';
+
+import { Chevron100Icon } from '@adobe/spectrum-wc/icon/elements';
+import * as iconElements from '@adobe/spectrum-wc/icon/elements';
```
Rajdeepc: [Here](https://github.com/adobe/spectrum-web-components/blob/ttomar/second-gen-icons/2nd-gen/packages/swc/package.json#L22) `*` only matches a single path segment so `@adobe/spectrum-wc/icon/elements` would not resolve through `./*`. This works in Storybook dev via the Vite `resolve.alias` but will fail for published customers.

2nd-gen/packages/swc/components/icon/stories/icon.internal.stories.ts
```diff
@@ -0,0 +1,282 @@
...
+
+import '@adobe/spectrum-wc/icon';
+
+import {
+  ICON_VALID_SIZES,
+  type IconSize,
+} from '../../../../core/components/icon/Icon.types.js';
```
Rajdeepc: You can use absolute import here since export was added to `core/package.json`

2nd-gen/packages/swc/components/icon/stories/icon.internal.stories.ts
```diff
@@ -0,0 +1,282 @@
...
+ * - **Extra-small (`xs`)**: Dense UIs and compact metadata rows
+ * - **Small (`s`)**: Supporting iconography in constrained spaces
+ * - **Medium (`m`)**: Default size for general component usage
+ * - **Large (`l`)**: Prominent icon usage in larger controls
+ * - **Extra-large (`xl`)**: High-emphasis icon presentation
+ *
+ * All sizes are shown below for comparison.
+ */
+export const Sizes: Story = {
+  render: (_args) => html`
```
Rajdeepc: Please pass `..args` here. You are ignoring it through prefixed with underscore. Storybook controls will not work.

2nd-gen/packages/swc/components/icon/elements/AlertIcon.ts
rubencarvalho: I'm OK with keeping it simpler for now! But interestingly enough, I'd actually argue these may be part of the "rendering layer" - they are theme-specific, e.g. if we were to have a `aether/components/icon/elements`, these SVGs would be different.

2nd-gen/packages/swc/components/icon/stories/icon.internal.stories.ts
rubencarvalho: Please do!

2nd-gen/packages/swc/components/icon/stories/icon.internal.stories.ts
```diff
@@ -0,0 +1,302 @@
...
+ * - **Large (`l`)**: Prominent icon usage in larger controls
+ * - **Extra-large (`xl`)**: High-emphasis icon presentation
+ *
+ * All sizes are shown below for comparison.
+ */
+export const Sizes: Story = {
+  render: (args) => html`
+    ${ICON_VALID_SIZES.map(
+      (size) => html`
+        <swc-icon
```
rubencarvalho: Can we use the template? 😄

2nd-gen/packages/swc/components/icon/stories/icon.internal.stories.ts
```diff
@@ -0,0 +1,302 @@
...
+
+/**
+ * ### Shared templates
+ *
+ * Import reusable templates from `../elements/index.js` and slot them into `<swc-icon>`.
+ * This keeps icon usage centralized and avoids per-component SVG duplication.
+ */
+export const Sources: Story = {
+  render: (args) => html`
+    <swc-icon
```
rubencarvalho: Template?

2nd-gen/packages/swc/components/icon/stories/icon.internal.stories.ts
```diff
@@ -0,0 +1,302 @@
...
+ *
+ * Example import:
+ *
+ * ```ts
+ * import { Chevron100Icon } from '../elements/index.js';
+ * ```
+ */
+export const SharedTemplates: Story = {
+  render: (args) => html`
+    <swc-icon
```
rubencarvalho: Same! Template wherever possible, please!

2nd-gen/packages/swc/components/icon/stories/icon.internal.stories.ts
```diff
@@ -0,0 +1,302 @@
...
+            style="
+                            display: inline-flex;
+                            flex-direction: column;
+                            align-items: center;
+                            gap: 8px;
+                            min-inline-size: 120px;
+                            padding: 8px;
+                        "
+          >
+            <swc-icon
```
rubencarvalho: (template)

2nd-gen/packages/swc/components/icon/stories/icon.internal.stories.ts
```diff
@@ -0,0 +1,302 @@
...
+ *
+ * ### Best practices
+ *
+ * - Always provide a descriptive `label` for informative icons
+ * - Use empty labels only for purely decorative icons
+ * - Keep labels short and specific (e.g., "Search" instead of "Icon")
+ */
+export const Accessibility: Story = {
+  render: (args) => html`
+    <swc-icon label=${ifDefined(args.label)} size=${ifDefined(args.size)}>
```
rubencarvalho: Last comment on the template! 😛

2nd-gen/packages/swc/components/icon/stories/icon.internal.stories.ts
```diff
@@ -0,0 +1,302 @@
...
+      .sort(([a], [b]) => a.localeCompare(b))
+      .map(([name, iconFactory]) => ({
+        name,
+        icon: (iconFactory as () => unknown)(),
+      }));
+    return html`
+      ${catalog.map(
+        (entry) => html`
+          <div
+            style="
```
rubencarvalho: This is breaking the formatting, could we use Lit's `styleMap` for this, please?

2nd-gen/packages/swc/components/icon/Icon.ts
```diff
@@ -0,0 +1,51 @@
...
+import { CSSResultArray, html, TemplateResult } from 'lit';
+
+import { IconBase } from '@spectrum-web-components/core/components/icon';
+
+import styles from './icon.css';
+
+/**
+ * Minimal icon renderer that accepts slotted SVG markup.
+ *
+ * @element swc-icon
```
rubencarvalho, line 21: I know it probably does not affect anything in our toolchain for now, but could we still mark this as `@internal` here?
```suggestion
 * @internal
 * @element swc-icon
```

2nd-gen/packages/swc/components/icon/stories/icon.internal.stories.ts
```diff
@@ -0,0 +1,276 @@
...
+const iconCardStyles = {
+  display: 'inline-flex',
+  'flex-direction': 'column',
+  'align-items': 'center',
+  gap: '8px',
+  'min-inline-size': '120px',
+  padding: '8px',
+} as const;
+
+const renderIcon = (
```
rubencarvalho: I think we may be talking past each other 😅  Maybe I'm missing what this does, but why do we need the `renderIcon` wrapper? Why can't we use `template` directly?
See the suggestions below

2nd-gen/packages/swc/components/icon/stories/icon.internal.stories.ts
```diff
@@ -0,0 +1,276 @@
...
+  },
+};
+
+// ────────────────────
+//    OVERVIEW STORY
+// ────────────────────
+
+export const Overview: Story = {
+  tags: ['overview'],
+  render: (args) => renderIcon(args),
```
rubencarvalho: Without the `renderIcon` this could be:
```suggestion
  render: (args) => template(args, iconSvg),
```

2nd-gen/packages/swc/components/icon/stories/icon.internal.stories.ts
```diff
@@ -0,0 +1,276 @@
...
+    icon
+  );
+
+// ────────────────────
+//    AUTODOCS STORY
+// ────────────────────
+
+export const Playground: Story = {
+  tags: ['autodocs', 'dev'],
+  render: (args) => renderIcon(args),
```
rubencarvalho: Without the `renderIcon` this could be:
```suggestion
  render: (args) => template(args, iconSvg),
```

2nd-gen/packages/swc/components/icon/stories/icon.internal.stories.ts
```diff
@@ -0,0 +1,276 @@
...
+ * An icon consists of:
+ *
+ * 1. **Rendered graphic** - A shared slotted SVG template
+ *
+ * ### Content
+ *
+ * - Default slot: Provide SVG markup to render.
+ */
+export const Anatomy: Story = {
+  render: (args) => renderIcon(args, { label: 'Chevron icon' }),
```
rubencarvalho: Without the `renderIcon` this could be:
```suggestion
  render: (args) => template({ ...args, label: 'Chevron icon' }, iconSvg),
```

2nd-gen/packages/swc/components/icon/stories/icon.internal.stories.ts
```diff
@@ -0,0 +1,276 @@
...
+          </div>
+        `
+      )}
+    `;
+  },
+  tags: ['options'],
+  parameters: {
+    flexLayout: true,
+    'section-order': 4,
+  },
```
rubencarvalho, line 243: This story bloats the documentation page by displaying a huge source code that is not very useful (it expands the SVGs). Can we hide it, in this case?

```suggestion
  parameters: {
    docs: {
      canvas: {
        sourceState: 'none',
      },
    },
    flexLayout: true,
    'section-order': 4,
  },
```

TarunAdobe: okayyy

rubencarvalho (approved):

PR #6021, opened 2026-02-12
feat(asset): exploring 2nd gen usage

aramos-adobe:
### Asset (2nd gen)

- Error state: Supports failed or unavailable asset content (e.g. broken images) with a clear error icon and optional visible label, improving feedback and accessibility.
- Aspect ratio: Provides ability to fit the media element on desired ratio ie. 16:9, 4:3, or 1:1. Portrait ratios are also supported
- Still maintains file and folder variants

### New `swc-image` component/tool (2nd-gen)

image-only component: a container that renders an `` from src and exposes the inner image for styling (e.g. width, height, border-radius) via the image part and `.spectrum-Image-image` class.

-    Research showed sp-asset is often used as a div wrapper around an image (or video), with a class like asset-image used to style size and border-radius. This adds a dedicated image-only component that:
     -  Always renders from `src` (no slot).
     -  Exposes the inner image via part="image" and .spectrum-Image-image so engineers can target it for width, height, border-radius, etc., similar to the existing asset-image pattern.

## Motivation and context

This PR leads the discussion on what to do with the asset component now that the design dictionary in Spectrum 2 has diverged from its usage while the component is still being used in applications such as Photoshop Web. 

## Related issue(s)

-   fixes [SWC-1463]

---

## Author's checklist

---

## Reviewer's checklist

### Manual review test cases

-   [ ] Asset error state

    1. Open Storybook → Asset → States → Error.
    2. Confirm the error icon and label “Failed to load image” (or current story label) are shown.
    3. Toggle the error control and confirm the asset switches between error and normal (e.g. file/folder/src/slot) as expected.

-   [ ] swc-image
    1. Open Storybook → Image → Playground.
    2. Open “Styling the image” and confirm Default, Rounded, and Circle examples render and the `::part(image)` styles apply (rounded corners and circle).
    3. Confirm ObjectFit examples (contain, cover, fill) behave as expected.

### Device review


2nd-gen/packages/core/components/image/Image.base.ts
```diff
@@ -0,0 +1,131 @@
...
+     * Height in pixels. Used for layout and aspect ratio.
+     */
+    @property({ type: Number })
+    public height?: number;
+
+    /**
+     * When true, shows an error state (e.g. failed to load).
+     */
+    @property({ type: Boolean, reflect: true })
+    public error = false;
```
Rajdeepc: `swc-image` has its own `ImageBase` with the exact same properties. This is a massive responsibility bloat on `AssetBase`. An `swc-asset` is conceptually a wrapperfor media. If you are making it also directly render  elements means it's trying to be both a container and an image component simultaneously.

Rajdeepc: How are you planning to create an automatic error detection when an image URL fails? This seems like a dead property. If consumer adds `<swc-image error>` nothing will happen. Kindly implement error state rendering in `Image.ts`.

2nd-gen/packages/swc/components/asset/test/asset.a11y.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+        expect(snapshot).toBeTruthy();
+        await expect(asset).toMatchAriaSnapshot();
+    });
+
+    test('should have correct accessibility tree for media wrapper', async ({
+        page,
+    }) => {
+        const asset = await gotoStory(
+            page,
+            'components-asset--media-wrapper',
```
Rajdeepc: Check this. This might fail.

2nd-gen/packages/core/components/asset/Asset.base.ts
```diff
@@ -17,42 +17,60 @@ import { SpectrumElement } from '@spectrum-web-components/core/shared/base/index
...
+    public rounded = false;
 
     protected override updated(changes: PropertyValues): void {
         super.updated(changes);
+
+        if (this.aspectRatio) {
+            this.style.aspectRatio = this.aspectRatio;
+        } else {
+            this.style.removeProperty('aspect-ratio');
+        }
```
Rajdeepc: This mutates the host element's inline style directly in `updated()`. This will run on every cycle not just when `aspectRatio` changes. It should guard with `if (changes.has('aspectRatio'))`. 
Think through this scenario too when a consumer sets `style="aspect-ratio:2` on the element this code will override the aspect-ratio portion but could also interact falsely if the consumer removes their inline style later.
```ts
if (changes.has('aspectRatio')) {
    if (this.aspectRatio) {
        this.style.aspectRatio = this.aspectRatio;
    } else {
        this.style.removeProperty('aspect-ratio');
    }
}
```

2nd-gen/packages/swc/components/image/Image.ts
```diff
@@ -0,0 +1,76 @@
...
+                    loading=${ifDefined(this.loading)}
+                    decoding=${ifDefined(this.decoding)}
+                    srcset=${ifDefined(this.srcset)}
+                    sizes=${ifDefined(this.sizes)}
+                    crossorigin=${ifDefined(this.crossorigin)}
+                    referrerpolicy=${ifDefined(this.referrerpolicy)}
+                    width=${ifDefined(this.width)}
+                    height=${ifDefined(this.height)}
+                    style=${styleMap(imageStyles)}
+                />
```
Rajdeepc: There's no way for consumers to know when the image finishes loading. At minimum you can add a error state here like `@error=${() => this.error = true} and dispatch a custom event like `swc-image-error`.

2nd-gen/packages/swc/components/image/image.css
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * The inner image element. Use this class or the `image` part to customize
+ * width, height, border-radius, object-fit, etc. from outside the component.
+ *
+ * Example: swc-image::part(image) { width: 80px; height: 80px; border-radius: 8px; }
+ */
+.spectrum-Image-image {
+    max-inline-size: 100%;
+    max-block-size: 100%;
+    object-fit: cover;
```
Rajdeepc: This conflicts with the `img` default and the JSDoc. You are only applying this via `StyleMap` when `this.objectFit` is explicitly set. The component API is not mentioning any default but here its defaulting to cover. Kindly verify.

2nd-gen/packages/swc/components/image/image.css
```diff
@@ -0,0 +1,42 @@
...
+    display: block;
+    block-size: 100%;
+    inline-size: 100%;
```
Rajdeepc: `block-size: 100% + inline-size: 100%` alongside `max-inline-size: 100% + max-block-size: 100%` makes the max-* properties redundant since a value of 100% can never exceed max of 100%. Check this too!

2nd-gen/packages/swc/components/asset/asset.css
```diff
@@ -27,16 +27,24 @@
...
+    border-radius: token('corner-radius-medium-default');
+    overflow: hidden;
+}
+
+/* Slotted media (e.g. swc-image, video, iframe) */
 ::slotted(*) {
     max-inline-size: 100%;
     max-block-size: 100%;
-    object-fit: contain;
+    object-fit: cover;
```
Rajdeepc: You are introducing a **breaking change** here.
Changing to cover means slotted images will now crop instead of letterbox. This is a breaking visual change for existing consumers who slot raw `` elements. Anyways this is also an anti-pattern no-op for `swc-image` as this is only meaningful on replaced elements and on a host custom element this does nothing.

2nd-gen/packages/swc/components/asset/stories/asset.stories.ts
```diff
@@ -58,9 +81,24 @@ export default meta;
...
+        'default-slot': `<swc-image src="https://picsum.photos/id/1015/400/400" alt="Landscape"></swc-image>`,
     },
+    render: (args) => html`
+        <div style="inline-size: 320px;">
+            <swc-asset
+                .aspectRatio=${args['aspect-ratio'] ?? undefined}
+                .label=${args.label ?? ''}
+                .variant=${args.variant}
+                .error=${args.error ?? false}
+                .rounded=${args.rounded ?? false}
```
Rajdeepc: If you bind manually instead of `template()` SB controls for other properties won't work. If new properties are added to the component this story wont pick them up.

2nd-gen/packages/swc/components/asset/stories/asset.stories.ts
```diff
@@ -58,9 +81,24 @@ export default meta;
...
+            >
+                ${args['default-slot']
+                    ? unsafeHTML(args['default-slot'])
```
Rajdeepc: This created an XSS vector. Please make sure this doesn't leak into production.

2nd-gen/packages/swc/components/asset/stories/asset.stories.ts
```diff
@@ -58,9 +81,24 @@ export default meta;
 
 export const Playground: Story = {
     args: {
-        label: 'Background',
-        'default-slot': `<img src="https://picsum.photos/id/56/80/80/?blur=2" alt="preview of background" />`,
+        'aspect-ratio': '1',
+        'default-slot': `<swc-image src="https://picsum.photos/id/1015/400/400" alt="Landscape"></swc-image>`,
     },
+    render: (args) => html`
+        <div style="inline-size: 320px;">
```
Rajdeepc: You can use Storybook layout parameters instead of added a fixed width.

2nd-gen/packages/swc/components/asset/stories/asset.stories.ts
```diff
@@ -132,43 +160,144 @@ export const Variants: Story = {
...
+    tags: ['options'],
+};
+
+/**
+ * Use the `aspect-ratio` attribute so the asset container keeps a consistent shape.
+ * Any valid CSS aspect-ratio value works (e.g. "1" for square, "16/9", "4/3").
+ * Useful for grids and cards where thumbnails should share the same ratio.
+ */
+export const AspectRatio: Story = {
+    render: () => html`
```
Rajdeepc: Always spread `...args `into template calls. This and Rounder stories are using `render: () => (no args)` and you are hardcoding everything which means Storybook controls are completely non-functional for these stories.

2nd-gen/packages/swc/components/asset/stories/asset.stories.ts
```diff
@@ -132,43 +160,144 @@ export const Variants: Story = {
...
+ * When `error` is set, the asset shows an error icon and optional label text.
+ * Use for failed or unavailable media.
+ */
+export const Error: Story = {
+    args: {
+        error: true,
+        label: 'Failed to load',
+    },
+    tags: ['states'],
+};
```
Rajdeepc, line 300: This is a very standalone error demonstration. can you be a more robust on this story and show 
1. Error with a variant set
2. Error without a label
3. Error with rounded corners

2nd-gen/packages/swc/components/image/Image.ts
```diff
@@ -0,0 +1,76 @@
...
+ * swc-image::part(image) { width: 80px; height: 80px; border-radius: 8px; }
+ */
+export class Image extends ImageBase {
+    public static override get styles(): CSSResultArray {
+        return [styles];
+    }
+
+    protected override render(): TemplateResult {
+        if (!this.src) {
+            return html`<div class="spectrum-Image"></div>`;
```
Rajdeepc: An empty div takes up space in layout (the CSS gives it `inline-size: 100% and block-size: 100%`). Please consider rendering `nothing` so the empty state is intentional.

2nd-gen/packages/swc/components/image/stories/image.stories.ts
```diff
@@ -0,0 +1,229 @@
...
+                ></swc-image>
+            </div>
+        </div>
+        <style>
+            .image-rounded::part(image) {
+                border-radius: 8px;
+            }
+            .image-circle::part(image) {
+                border-radius: 50%;
+            }
```
Rajdeepc: Don't use global `style` tags in storybook. It will leak unto other stries when SB reuses the DOM canvas. You can use a scoped decorator.

Rajdeepc (changes requested):
Good effort. Noticed a few structural gaps in this PR. 
The image-related types are duplicated in both `Asset.types.ts` and `Image.types.ts`, which is a drift — these should be consolidated into a shared source or imported from one place.

The new `swc-image` component is also introduced without any unit or accessibility tests (there’s no components/image/test/ coverage), which we shouldn’t ship for a new component. 

The accessibility documentation in `asset.stories.ts` is quite thin. It’s missing keyboard behavior (even stating non-interactive), visual accessibility considerations, ARIA implementation details, and clarification of error-state semantics. 

I’ve left more inline comments, but these should be addressed before moving forward.

PR #6028, opened 2026-02-16
fix(picker): prevent internal scroll event from crossing Shadow DOM boundary

rubencarvalho:
## Description

Changed `composed: true` to `composed: false` on the synthetic `scroll` event dispatched by the Picker's `onScroll` handler. This prevents the event from crossing shadow DOM boundaries, which could inadvertently trigger scroll-related close logic on ancestor scrollable containers (e.g., overlays or popovers higher in the DOM tree).

## Motivation and context

Under certain edge cases reported by PS Web, when a Picker was rendered inside another scrollable ancestor (such as a scrollable overlay or popover), the `composed: true` scroll event would leak through shadow roots and cause the ancestor to close unexpectedly. The scroll event has no reason to cross shadow DOM boundaries — it only needs to be observable within the Picker's own shadow root.

N/A — behavioral fix, no visual changes.

---

## Author's checklist

---

## Reviewer's checklist

- [ ] Picker scroll behavior still works as expected in isolation
    1. Open a Picker with enough items to require scrolling
    2. Scroll through the menu items
    3. Expect normal scroll behavior within the Picker menu

- [x] **Keyboard** (required — document steps below) — No keyboard behavior changes; the fix only affects event propagation, not interaction patterns.
    1. Open a Picker using <kbd>Enter</kbd> or <kbd>Space</kbd>
    2. Navigate menu items with <kbd>Arrow Up</kbd>/<kbd>Arrow Down</kbd>
    3. Expect focus management and keyboard navigation to work identically to before

- [x] **Screen reader** (required — document steps below) — No screen reader behavior changes; the fix only affects a synthetic scroll event's `composed` flag.
    1. Open a Picker with VoiceOver/NVDA enabled
    2. Navigate through menu items
    3. Expect all announcements (role, name, selection state) to remain unchanged


1st-gen/packages/picker/src/Picker.ts
```diff
@@ -808,7 +808,7 @@ export class PickerBase extends SizedMixin(SpectrumElement, {
         this.dispatchEvent(
             new Event('scroll', {
                 cancelable: true,
-                composed: true,
+                composed: false,
```
Rajdeepc: You need to update the relevant test also for this change as [here](https://github.com/adobe/spectrum-web-components/blob/ruben/avoid-picker-scroll-composed/1st-gen/packages/action-menu/test/index.ts#L937-L939)

In addition to this it would be great if you can add a test where it renders a Picker inside a scrollable overlay/popover --> Scrolls the Picker's menu --> Asserts the parent overlay does not close. 
Just making sure we are protecting the regression against this change.

1st-gen/packages/picker/src/Picker.ts
```diff
@@ -808,7 +808,7 @@ export class PickerBase extends SizedMixin(SpectrumElement, {
         this.dispatchEvent(
             new Event('scroll', {
                 cancelable: true,
```
Rajdeepc: I don't think native scroll events are cancelable. Users might assume "Oh I can cancel this event" but apparently they cannot. Slightly misleading.

rubencarvalho: updated

1st-gen/packages/picker/src/Picker.ts
rubencarvalho: done!

PR #6031, opened 2026-02-17
feat(typography): type variant classes

5t3ph:
## Description

This adds a generator to create the typography variant classes as a new stylesheet, and related Storybook files.

Adds classes for:
- heading
- title
- body
- detail
- code

Includes modifiers for `--serif` per variant, and "global" modifier for `.swc-Typography--emphasized` for italics that excludes CJK langs.

Also introduces modifiers for selectively applying margins to a variant, and a "prose" container class to apply margins and essential variant styles to foundational semantic HTML typography elements (these may be expanded later for "global styles"). Prose styles have design sign-off, and are opt-in and further customizable via associated custom properties.

### TBD outside of this PR:

_SWC-1545_
- where/how to export the stylesheet
- whether to export as JS in addition to the stylesheet
- whether to add to the base component class (possible "component" styles may be most appropriate)

_SWC-1544_
- generating styles based on composite type variants, including `component`

## Motivation and context

Typography classes are used for slotted/light DOM content for ease of applying groups of font style properties, including adjustments for CJK languages.

## Related issue(s)

SWC-1438

## Screenshots

image

## Author's checklist

- [X] I have read the **[CONTRIBUTING](<(https://github.com/adobe/spectrum-web-components/blob/main/CONTRIBUTING.md)>)** and **[PULL_REQUESTS](<(https://github.com/adobe/spectrum-web-components/blob/main/PULL_REQUESTS.md)>)** documents.
- [X] I have reviewed at the Accessibility Practices for this feature, see: [Aria Practices](https://www.w3.org/TR/wai-aria-practices/)
- [X] I have included updated documentation if my change required it.

---

## Reviewer's checklist

### Manual review test cases

- [x] Review the new Typography stories for clarity and completeness
- [ ] Review the generated stylesheet `typography.css` for any selector concerns or other potential issues
- [ ] Validate type variant styles against S2 token specs
    - likely easiest to review in generated stylesheet `typography.css` to compare tokens
    - CJK fonts are not yet loaded in our Storybook, but devtools can be used to see the property updates
    - "heavy" is only valid for Heading (for use by Express) so doesn't have a story but is toggleable in the playground
- [ ] Test the generator command by using `yarn tokens:update` at the project root and ensure no conflicts with token processing


2nd-gen/packages/swc/components/typography/stories/typography.stories.ts
```diff
@@ -0,0 +1,249 @@
...
+/**
+ * Variants default to Adobe Clean Spectrum VF, a sans-serif font family.
+ * Modifiers are available to use the serif font stack, and other sub-variants.
+ * CJK adjustments are applied based on the `:lang()` selector.
+ *
+ * Learn more about [loading the expected fonts](/?path=/docs/guides-customization-fonts--readme).
+ */
+const meta: Meta<TypographyTemplateProps> = {
+    title: 'Typography',
+    parameters: {
```
5t3ph: @caseyisonit if it's available, let me know how to maybe disable the "Getting Started" section for now? I have a follow-up task of how we want to export this stylesheet and then we can provide custom instructions in some fashion.

2nd-gen/packages/swc/vite.config.ts
```diff
@@ -50,6 +50,7 @@ export default defineConfig({
             'custom-properties': false, // Let lit-css handle this
             'light-dark-function': false,
             'logical-properties-and-values': false,
+            'is-pseudo-class': false,
```
5t3ph: `:is()` has been supported for over 5 years, so this keeps it intact vs. transpiling into individual selectors (used for attaching semantic element styles in the typography stylesheet)

2nd-gen/packages/tools/swc-tokens/src/typography.js
```diff
@@ -0,0 +1,758 @@
...
+    sizes: {
+      xl: ['h2'],
+      l: ['h3'],
+      m: ['h4'],
+    },
+    margins: [':is(h2, h3, h4)'],
+  },
+};
+
+function capitalize(str) {
```
5t3ph, line 148: Due to workspace related issues, I wasn't able to successfully import this from the core package, I think since this sits in the `/tools` directory.

2nd-gen/packages/swc/.storybook/preview.ts
```diff
@@ -17,13 +17,14 @@ import {
 } from '@wc-toolkit/storybook-helpers';
 
 import customElements from './custom-elements.json';
-import { withFlexLayout, withStaticColorsDemo } from './decorators';
-import { withContext } from './decorators/contexts';
-import { withStaticColorPlayground } from './decorators/static-color-playground';
+import { withContext } from './decorators/contexts.js';
+import { withFlexLayout, withStaticColorsDemo } from './decorators/index.js';
+import { withStaticColorPlayground } from './decorators/static-color-playground.js';
```
marissahuysentruyt, line 22: So previously, we had `import { withFlexLayout, withStaticColorsDemo } from './decorators';`, which makes it sound like we're basically importing these decorators from the decorators "module." However, the `withContext` and `withStaticColorPlayground` decorators are imported directly from their `decorators/<file>` instead. 

Just for my own understanding- is there a reason to do one or the other? Could we lump all 4 of these decorators into the `from './decorators.index.js'` import? Would we be losing or gaining anything by importing separately vs. from the entire decorator directory?

2nd-gen/packages/swc/components/typography/stories/typography.stories.ts
```diff
@@ -0,0 +1,251 @@
+/**
+ * Copyright 2025 Adobe. All rights reserved.
```
marissahuysentruyt: We could update the copyright! 📅 Although I think Casey mentioned the next linting PR might address this, too...

2nd-gen/packages/swc/components/typography/stories/typography.stories.ts
```diff
@@ -0,0 +1,251 @@
...
+  type TypographyTemplateProps,
+  VARIANTS,
+} from './typography.template.js';
+
+/**
+ * Variants default to Adobe Clean Spectrum VF, a sans-serif font family.
+ * Modifiers are available to use the serif font stack, and other sub-variants.
+ * CJK adjustments are applied based on the `:lang()` selector.
+ *
+ * Learn more about [loading the expected fonts](/?path=/docs/guides-customization-fonts--readme).
```
marissahuysentruyt:
```suggestion
 * Learn more about [loading the expected fonts](/docs/guides-customization-fonts--readme).
```

When i ran this locally, I was getting duplicate `?path=` in the URL. This seemed to work for me.

2nd-gen/packages/swc/.storybook/preview-head.html
```diff
@@ -116,35 +108,68 @@
...
-    padding: 0px 0px 4px;
-    cursor: text;
-    position: relative;
-    color: rgb(46, 52, 56);
-    border-bottom: 1px solid rgba(38, 85, 115, 0.15);
-    letter-spacing: unset;
-    line-height: unset;
+  .typography-samples {
+    #storybook-root & {
+      background-color: var(--swc-white);
```
marissahuysentruyt: Do you think we could opt for something like `--swc-gray-25` instead of white? 

It's interesting because on the docs page, the background color of the stories flips between dark and light modes with no issue (because I don't think there's a `#storybook-root` container): 

Image

But in the playground story, this is not quite working as I expected: 

Image

marissahuysentruyt: Here's the playground story with `gray-25`
Image

2nd-gen/packages/swc/components/typography/stories/typography.stories.ts
```diff
@@ -0,0 +1,251 @@
...
+ */
+const meta: Meta<TypographyTemplateProps> = {
+  title: 'Typography',
+  parameters: {
+    docs: {
```
marissahuysentruyt, line 34: I'd love to hear what Casey has to say on this, but this is Cursor's suggestion on how to remove the Getting Started section on the docs page 🤷‍♀️ : 

```
// in GettingStarted.tsx, add a flag
  if (resolvedOf.preparedMeta?.parameters?.docs?.hideGettingStarted) {
    return null;
  }

// in typography.stories.ts, use that flag in the parameters
parameters: {
    docs: {
      subtitle: `Typography variant styles are available via CSS classes.`,
      hideGettingStarted: true,
      canvas: {
        sourceState: 'hidden',
      },
    },
  },
```

2nd-gen/packages/swc/components/typography/stories/typography.stories.ts
```diff
@@ -0,0 +1,251 @@
...
+ * Applies block-direction margins to all type variants within a container with the `.swc-Typography--prose` class applied.
+ *
+ * This also applies foundational type properties based on heading, title and body styles to common semantic typography elements including:
+ * - `h1` - Heading, size M
+ * - `h2` - Title, size XL
+ * - `h3` - Title, size L
+ * - `h4` - Title, size M
+ * - `p, li` - Body, size M
+ */
+export const ProseContainer: Story = {
```
marissahuysentruyt, line 228: Ohhhhh ok. So this prose container will apply styles, without consumers needing to apply the typography classes themselves?? Neat!

2nd-gen/packages/swc/components/typography/stories/typography.stories.ts
```diff
@@ -0,0 +1,251 @@
...
+        sourceState: 'hidden',
+      },
+    },
+  },
+  argTypes: {
+    variant: { control: 'select', options: VARIANTS },
+    size: { control: 'select', options: SIZES },
+    serif: { control: 'boolean' },
+    heavy: { control: 'boolean' },
+    emphasized: { control: 'boolean' },
```
marissahuysentruyt, line 46: I don't know a ton about fonts, but I have heard that there are ways of not creating a "true italic." I don't quite understand what that means, honestly. But then thinking about the `<em>` tag, that does create a "true italic?" 

This doesn't really have anything to do with the work you've done here- just something that popped in my head while reviewing! I think I got here thinking about the emphasized modifier, wondering if it should have the `em` tag, but then maybe that's not really the semantic purpose of the emphasized modifier.... 😆 😵‍💫

2nd-gen/packages/swc/components/typography/stories/typography.stories.ts
```diff
@@ -0,0 +1,251 @@
...
+      canvas: {
+        sourceState: 'hidden',
+      },
+    },
+  },
+  argTypes: {
+    variant: { control: 'select', options: VARIANTS },
+    size: { control: 'select', options: SIZES },
+    serif: { control: 'boolean' },
+    heavy: { control: 'boolean' },
```
marissahuysentruyt, line 45: Would you like to add a story for the heavy modifier to the docs page at all? I think this is the only modifier without a docs story.

2nd-gen/packages/swc/package.json
```diff
@@ -43,7 +43,8 @@
     "dev:core": "yarn workspace @spectrum-web-components/core dev",
     "storybook": "yarn analyze && storybook dev -p 6006",
     "storybook:build": "yarn analyze && storybook build",
-    "stylesheets": "swc-tokens --outputType stylesheet --out ./stylesheets/tokens.css --prefix swc",
+    "stylesheet:tokens": "swc-tokens --outputType tokens --out ./stylesheets/tokens.css --prefix swc",
+    "stylesheet:typography": "swc-tokens --outputType typography --out ./stylesheets/typography.css --prefix swc",
```
marissahuysentruyt, line 53: Feel free to correct me if I'm wrong (or if I didn't read thoroughly enough), but when I run `yarn stylesheet:typography`, I get this message: 

Image

Is that expected? I only saw the `code-font-weight` token, but the other typography components all have a `<variant>-sans-serif-font-weight` token as well. It felt like an error, but I'm not sure.

2nd-gen/packages/tools/swc-tokens/src/typography.js
```diff
@@ -0,0 +1,758 @@
...
+    const mLineHeightTokenName = aliasedMFontSize
+      ? `line-height-${aliasedMFontSize}`
+      : null;
+
+    // use patch-aware lookup for variant tokens
+    const sansWeightRef = getVariantTokenRef(
+      tokens,
+      typeVar,
+      'sans-serif-font-weight'
+    );
```
marissahuysentruyt, line 540: I think this loop is related to that `code-sans-serif-font-weight` comment I left. I don't know enough about the requirements of the code type style- does it have "serif" and "sans-serif" differences? In my head, the code font just _is_ the code font, not sans or serifed.

2nd-gen/packages/swc/.storybook/preview.ts
5t3ph: honestly the only reason this got touched was due to merging in the first linting PR 😅 

maybe bring this question up in Slack?

2nd-gen/packages/swc/.storybook/preview-head.html
5t3ph: ahhh good catch, thanks!

2nd-gen/packages/swc/components/typography/stories/typography.stories.ts
5t3ph: yup, correct!

2nd-gen/packages/swc/components/typography/stories/typography.stories.ts
5t3ph: Good consideration, and I believe what you're referring to applies when the font family doesn't include an official italic and creates [faux italics](https://fonts.google.com/knowledge/introducing_type/introducing_weights_styles). In that case, `font-style: italic` forces the slant.

However, since we are now using a variable font, _if_ we instead were using [font metrics to modify the slant axis](https://fonts.google.com/knowledge/glossary/slant_axis), we would want to define that explicitly.

I have a question out to design to verify if we should set the axis or just use `font-style: italic` since it's not spelled out in the specs or other docs I can find. I'm also verifying if we should set generic bold text (`<strong>`) to a weight of `700` vs. the browser default `600` since body text is already `500`.

Either way we can address these in another pass if I don't get an answer before we're otherwise ready to merge 👍

2nd-gen/packages/swc/components/typography/stories/typography.stories.ts
5t3ph: I was on the fence on adding it, but I will. It's largely supposed to be reserved for Adobe Express per their specific brand guidelines, but we can include it for completeness.

2nd-gen/packages/swc/package.json
5t3ph: It is expected, but I will tuck it behind a "debug" flag to prevent confusion 👍

2nd-gen/packages/tools/swc-tokens/src/typography.js
5t3ph: Correct, the code font doesn't have these types of differences 👍

2nd-gen/packages/swc/components/typography/stories/typography.stories.ts
5t3ph: If @caseyisonit is good with this, I'll add it, thanks for investigating!

caseyisonit: this will hide it but not make it customizable, i think this is a solid temp solution so it doesnt add confusion.

2nd-gen/packages/swc/.storybook/preview-head.html
caseyisonit, line 1: thank you for this clean up!

2nd-gen/packages/swc/components/typography/stories/typography.stories.ts
caseyisonit: the other approach i would consider is using a tag of "utility" and if that tag is present the getting started is hidden, this leans in to storybook patterns versus a custom parameter we create

2nd-gen/packages/swc/components/typography/stories/typography.stories.ts
```diff
@@ -0,0 +1,268 @@
...
+  title: 'Typography',
+  parameters: {
+    docs: {
+      subtitle: `Typography variant styles are available via CSS classes.`,
+      canvas: {
+        sourceState: 'hidden',
+      },
+    },
+  },
+  argTypes: {
```
caseyisonit, line 41: can we use the argTypes from wc-toolkit? im not super keen on manually defining these if they can be implied from the source. this is different from a component so perhaps its not doable, still wrapping my head around the difference and what we can get from our tools.

2nd-gen/packages/swc/components/typography/stories/typography.template.ts
```diff
@@ -0,0 +1,290 @@
...
+  'XXXXL',
+] as const;
+export const VARIANTS = ['heading', 'title', 'body', 'detail', 'code'] as const;
+export const LANGS = ['en', 'zh', 'ja', 'ko'] as const;
+
+export const SIZES_BY_VARIANT: Record<
+  TypographyVariant,
+  readonly TypographySize[]
+> = {
+  heading: ['S', 'M', 'L', 'XL', 'XXL', 'XXXL', 'XXXXL'],
```
rise-erpelding: Comparing to the internal Typography fundamentals docs, it looks like heading exists in t-shirt sizes XS-XXXXL, is there a reason there's no XS included for heading here?

(The rest of these look great and match what I saw in those docs!)

2nd-gen/packages/swc/components/typography/stories/typography.stories.ts
5t3ph: updated to gate with the `utility` tag approach, and flexible to accept all tags for future other kinds of switches

2nd-gen/packages/swc/components/typography/stories/typography.stories.ts
5t3ph: I don't think we can since typography would not be included in CEM since there's no web component associated
image

2nd-gen/packages/swc/components/typography/stories/typography.stories.ts
marissahuysentruyt: Delightful- thanks for all of those resources!

2nd-gen/packages/tools/swc-tokens/README.md
```diff
@@ -232,13 +234,25 @@ All of those concerns are handled during the token generation phase.
...
-The main package provides a CLI to produce either the unified stylesheet `tokens.css` or the dump of processed data as `tokens.json`.
+The main package provides a CLI to produce either the unified stylesheet `tokens.css`, the typography stylesheet `typography.css`, or the dump of processed token data as `tokens.json`.
 
 > For purposes of the main package, those files are git ignored and intended for debugging purposes only.
 
-For the primary consuming package `@adobe/swc`, the following CLI command is used:
+For the primary consuming package `@adobe/swc`, the following CLI command is used for generating the token stylesheet:
 
 ```bash
 swc-tokens --outputType stylesheet --out ./stylesheets/tokens.css --prefix swc
```
rise-erpelding: Does this command need to be updated? To something like `--outputType tokens` instead of `--outputType stylesheet`?

2nd-gen/packages/tools/swc-tokens/README.md
rise-erpelding: This could be updated to mention that running `yarn tokens:update` generates the typography styles now too!

2nd-gen/packages/swc/.storybook/preview-head.html
```diff
@@ -116,35 +108,68 @@
...
+
+    &.typography-samples--grid {
+      display: grid;
+      gap: 1rem;
+    }
+
+    .typography-row {
+      display: grid;
+      gap: 0.5rem;
+    }
```
rise-erpelding, line 162: I can see these types of classes (along with typography classes) being useful in our component stories too, similar to how we used `.spectrum-Detail` when showing multiple component instances next to each other in the CSS repo:

Image

rise-erpelding (commented):
This is looking great, and I really like the prose option!

2nd-gen/packages/swc/.storybook/preview-head.html
5t3ph: I think a future pass on Storybook would be to work out some of these "story pattern" styles - there are some that are being inlined right now that could be moved to classes. I can create a ticket for the idea 👍

2nd-gen/packages/swc/components/typography/stories/typography.template.ts
5t3ph: good catch, just an oversight, thanks!

2nd-gen/packages/tools/swc-tokens/README.md
5t3ph: good call, updated!

2nd-gen/packages/tools/swc-tokens/README.md
5t3ph: yes, updated!

2nd-gen/packages/swc/.storybook/preview-head.html
5t3ph: added - SWC-1573

package.json
```diff
@@ -42,7 +42,7 @@
     "test:a11y:report": "playwright show-report 2nd-gen/test/playwright-a11y/report",
     "test:a11y:ui": "playwright test --config=playwright.a11y.config.ts --ui",
     "tokens:test": "yarn workspace @adobe/swc-tokens test && yarn workspace @adobe/postcss-token test && yarn workspace swc-vscode-token test",
```
marissahuysentruyt: This comment is based on our conversation during sync this morning, but it sounds like the package name is `@adobe/spectrum-wc`!

2nd-gen/packages/swc/components/typography/stories/typography.stories.ts
caseyisonit: ahhh gotcha! im good with this then

1st-gen/packages/breadcrumbs/src/Breadcrumbs.ts
```diff
@@ -30,7 +30,6 @@ import {
 } from '@spectrum-web-components/base/src/directives.js';
 import { ifDefined } from '@spectrum-web-components/base/src/directives.js';
 
-/* eslint-disable import/no-extraneous-dependencies */
```
caseyisonit, line 33: Global comment: these disables should not be deleted, i saw this as a side effect of the linting PR. if you add these back and it keeps happening please reach out!! i posted the clean up steps in slack but happy to assist if you need!

2nd-gen/packages/swc/.storybook/blocks/GettingStarted.tsx
caseyisonit, line 1: love this simplicity! and will make it extremely easy to customize as we figure out if there is a pattern to utility getting started instructions.

2nd-gen/packages/swc/.storybook/manager.ts
```diff
@@ -17,6 +17,7 @@ import { create } from '@storybook/theming';
 import logo from './assets/logo.svg';
 
 import '../stylesheets/swc.css';
+import '../stylesheets/typography.css';
 import './assets/manager.css';
```
caseyisonit, line 21: general comment: i have no idea if the manger and preview stylesheets are even working. i had to do so much styling in the head.html. no feedback regarding this, just wanted to bring your attention to this as maybe something we can look in to in the future.

2nd-gen/packages/swc/stylesheets/tokens.css
```diff
@@ -33,54 +36,167 @@
...
-  --swc-neutral-subdued-background-color-key-focus: light-dark(var(--swc-gray-800), var(--swc-gray-400));
-  --swc-neutral-subdued-content-color-selected: var(--swc-neutral-subdued-content-color-down);
-  --swc-accent-background-color-default: light-dark(var(--swc-accent-color-900), var(--swc-accent-color-800));
-  --swc-accent-background-color-hover: light-dark(var(--swc-accent-color-1000), var(--swc-accent-color-700));
-  --swc-accent-background-color-down: light-dark(var(--swc-accent-color-1000), var(--swc-accent-color-700));
-  --swc-accent-background-color-key-focus: light-dark(var(--swc-accent-color-1000), var(--swc-accent-color-700));
+  --swc-neutral-subdued-background-color-default: light-dark(
+    var(--swc-gray-700),
+    var(--swc-gray-500)
+  );
```
caseyisonit, line 42: if this is unintended, we can always increase the line length in stylelint....

1st-gen/packages/breadcrumbs/src/Breadcrumbs.ts
5t3ph: Reverted!

2nd-gen/packages/swc/.storybook/manager.ts
5t3ph: they seem to be (or at least preview?) since otherwise the styles wouldn't have another way of affecting the stories.

2nd-gen/packages/swc/package.json
```diff
@@ -49,7 +49,8 @@
     "dev:core": "yarn workspace @spectrum-web-components/core dev",
     "storybook": "yarn analyze && storybook dev -p 6006",
     "storybook:build": "yarn analyze && storybook build",
-    "stylesheets": "swc-tokens --outputType stylesheet --out ./dist/tokens.css --prefix swc",
+    "stylesheet:tokens": "swc-tokens --outputType tokens --out ./stylesheets/tokens.css --prefix swc",
+    "stylesheet:typography": "swc-tokens --outputType typography --out ./stylesheets/typography.css --prefix swc",
```
caseyisonit, line 53: Question/nit: On these commands, and perhaps this is captured in the docs below, just going in order of file, when would we need to run these commands and should we tie them in to our build command? or is that overkill? not sure how often we would need to run these

2nd-gen/packages/swc/stylesheets/tokens.css
5t3ph: I'm not worried about it for the auto-generated file, and either way, this is fine with me 👍

2nd-gen/packages/swc/.storybook/preview-head.html
caseyisonit, line 1: Question/nit: Should we make a ticket for using the code font family for the storybook code previews? i left it out in my original pass because the standard font was no bueno in it. Not sure if it would be a quick lift to add to this PR or if we should make a more dedicated storybook CSS cleanup/refine ticket.

2nd-gen/packages/swc/package.json
5t3ph: They are documented as part of swc-tokens/README where the `yarn tokens:update` command triggers them.

Build command would be too often as they only need updated when the source token data or generator logic updates, so it's intended to be a combo with that work as needed.

2nd-gen/packages/swc/.storybook/preview-head.html
5t3ph: I think dedicated ticket because I did forget to note for this that we are not currently loading the "correct" font via Adobe Fonts, so that would be part of it too in order to fully support it. I can make that ticket.

2nd-gen/packages/tools/swc-tokens/src/typography.js
```diff
@@ -0,0 +1,777 @@
...
+import { createRequire } from 'node:module';
+import path from 'path';
+
+const require = createRequire(import.meta.url);
+
+/**
+ * Default variant order + size keys.
+ */
+const DEFAULT_VARIANTS = ['heading', 'title', 'body', 'detail', 'code'];
+const SIZE_KEYS = ['xxs', 'xs', 's', 'm', 'l', 'xl', 'xxl', 'xxxl', 'xxxxl'];
```
caseyisonit, line 24: nit: are these a pattern from somewhere? could we use numbers versus x's?  `2xs` `2xl` `3xl` `4xl`? for me thats easier to read and reason quicker.

2nd-gen/packages/swc/.storybook/preview-head.html
5t3ph: created SWC-1574

2nd-gen/packages/tools/swc-tokens/src/typography.js
5t3ph: This is the current actual token naming pattern so we'd have to do secondary mapping if we changed them in order to retrieve the right token. And that would cause a disconnect with current token spec design docs if we emitted different token names.

2nd-gen/packages/swc/stylesheets/typography.css
```diff
@@ -0,0 +1,509 @@
...
+  color: var(--swc-heading-font-color, token("heading-color"));
+
+  &:lang(zh),
+  &:lang(ja),
+  &:lang(ko) {
+    line-height: var(--swc-heading-cjk-line-height, token("cjk-line-height-100"));
+    letter-spacing: var(--swc-heading-cjk-letter-spacing, token("cjk-letter-spacing"));
+
+    --swc-heading-font-size: token("font-size-400");
+  }
```
marissahuysentruyt, line 42: I think I may be experiencing a specificity issues with this code, but I'm not 100% sure. 

Either way, when I toggle `includeMultipleSizes` on, and set `lang="ja"` for example, I don't see multiple sizes. I believe this code with `--swc-heading-font-size: var(--swc-font-size-400)` takes over. In the screenshot, you can see that the correct font-size-1400 is defined and ready, but I think this selector with font-size-400 is more specific in some way (Cursor says it's because of `:is()`).

Image

2nd-gen/packages/tools/swc-tokens/src/typography.js
```diff
@@ -0,0 +1,777 @@
...
+
+  const cfg = selectorAliases?.[typeVar];
+  if (!cfg) {
+    return Array.from(out).join(', ');
+  }
+
+  const prose = proseSelector(prefix);
+  const key = ALIAS_KIND_KEYS[kind];
+
+  const add = (s) => out.add(`${prose} ${s}`);
```
marissahuysentruyt, line 249: Related back to my question about the `includeMultipleSizes` control + `lang`, Cursor suggested the following (using `:where()`): 

```
const add = (s) => out.add(`:where(${prose}) ${s}`);
``` 

...so that we're not adding extra specificity. There was a similar suggestion near line 625: 
```
const marginsSelector = `.${className}--margins, :where(${proseSelector(
      prefix
    )}) .${className}`;
```

Although the suggested Cursor fixes seem to solve the problem I was experiencing, the more I think about this, the more I'm concerned I found a storybook-specific bug. So feel free to disregard if you think this is a byproduct of storybook and the controls.

2nd-gen/packages/swc/stylesheets/tokens.css
```diff
@@ -1416,772 +1964,1848 @@
...
+  --swc-title-size-s: var(--swc-theme--sizeM, 14px)
+    var(--swc-theme--sizeL, 17px);
+  --swc-title-size-xl: var(--swc-theme--sizeM, 20px)
+    var(--swc-theme--sizeL, 24px);
+  --swc-title-size-xs: var(--swc-theme--sizeM, 12px)
+    var(--swc-theme--sizeL, 15px);
+  --swc-title-size-xxl: var(--swc-theme--sizeM, 22px)
+    var(--swc-theme--sizeL, 27px);
+  --swc-title-size-xxxl: var(--swc-theme--sizeM, 25px)
+    var(--swc-theme--sizeL, 31px);
```
marissahuysentruyt, line 2863: Just for my own clarification: title-size-l is set to font-size-300. Because we have `resolveAliases: true` in `tokens.js`, that's why the value here is populated with the end value, correct? It looks identical to the font-size-300 definition.

I was half expecting to see --swc-title-size-l: var(--swc-font-size-300), so I just wanted to make sure I understood why it _wasn't_ like that again.

2nd-gen/packages/swc/stylesheets/typography.css
5t3ph: Ah, I think I failed to check this in the more recent iterations - really important catch! 🙌 

The Cursor suggestions weren't "wrong" but the solution I've pushed up instead is to manage this with CJK-related custom property overrides, which gets around the specificity issues without creating new ones and also sets us up better long-term for ease of consumer CJK-specific overrides 👍 

If you're interested: the root of the issue is that use of nesting means that nested selectors take on the  [highest specificity of the parent selectors](https://kilianvalkhof.com/2023/css-html/the-gotchas-of-css-nesting/), meaning the prose selector was increasing specificity of all the nested lang selectors. I wanted to avoid `:where()` because the prose selector is intentionally increasing specificity (albeit slightly) to keep the intended application of styles on semantic elements. Adding `:where()` would decrease that to only an elemental selector which would be quite weak against other possible application styles.

2nd-gen/packages/swc/stylesheets/tokens.css
5t3ph: That's correct. This is an area where I hope to get consumer feedback and use cases before we fully put a bow on anything theming/customization-related 👍

PR #6038, opened 2026-02-19
feat(textfield, number-field): add tooltip for a11y on truncated input text

rise-erpelding:
## Description

Adds truncated-value tooltip behavior to single-line text inputs in `sp-textfield` and `sp-number-field`.

- Detects when the visible value is clipped and conditionally renders a tooltip with the full value.
- Applies behavior to both valid and invalid values (when truncated).
- Includes updates to tests, stories, docs, and changeset coverage.

**Updates (2/27/26):**
- Does not show tooltip for password and removes truncation for password fields
- Adds `describeTrigger` property to overlay to prevent overlay from setting `aria-describedby`, thus preventing screen reader double announcement
- Breaks tooltip feature out into a mixin which lazy-loads tooltip and overlay
- Although it worked in textfield, number-field was not reflecting real-time edits within the tooltip, so a fix was made to keep truncated tooltip in sync with edits
- In number-fields whose content went from untruncated to truncated, the tooltip was not showing up - this was fixed
- Tooltip is now maintained during focus (while typing), it doesn't suddenly disappear when content is small enough to not need truncation

## Motivation and context

Long values in narrow fields can be difficult to review, especially while correcting invalid input.

This change improves usability by exposing the full value on hover/focus only when truncation occurs, while keeping default behavior unchanged when content fits.

## Related issue(s)

- fixes SWC-1183

image

image

---

## Author's checklist

---

## Reviewer's checklist

### Manual review test cases

- [x] Textfield shows tooltip only when visually truncated
  1. Open `Textfield / truncatedValueTooltip` [story](https://swcpreviews.z13.web.core.windows.net/pr-6038/docs/first-gen-storybook/?path=/story/textfield--truncated-value-tooltip&globals=system:spectrum-two).
  2. Hover and keyboard-focus the truncated example.
  3. Verify full value is shown in tooltip.
  4. Verify non-truncated example does not show tooltip.

- [x] Invalid truncated textfield still exposes full value
  1. Use the invalid truncated textfield story/example.
  2. Hover/focus the field.
  3. Verify tooltip appears with the full value even while invalid.

- [x] Number-field truncation behavior works with and without stepper
  1. Open `Number Field / truncatedValueTooltip` [story](https://swcpreviews.z13.web.core.windows.net/pr-6038/docs/first-gen-storybook/?path=/story/number-field--truncated-value-tooltip&globals=system:spectrum-two).
  2. Verify tooltip appears for truncated examples (stepper visible and hidden).
  3. Verify formatted currency text appears in tooltip.

- [ ] Regression checks
  1. Verify multiline textfield behavior is unchanged (no new truncation tooltip behavior introduced there).
  2. Verify no regressions in validation icon/help-text rendering.

**Updates (2/27/26):**
- [ ] Check password field in [Textfield truncation stories](https://swcpreviews.z13.web.core.windows.net/pr-6038/docs/first-gen-storybook/?path=/story/textfield--truncated-value-tooltip&globals=system:spectrum-two), confirm that this does not have truncation ellipses and does not show the password in a tooltip
- [ ] Use screen-reader on [truncated textfield](https://swcpreviews.z13.web.core.windows.net/pr-6038/docs/first-gen-storybook/?path=/story/textfield--truncated-value-tooltip&globals=system:spectrum-two) to determine that the text input is not double announced
- [ ] Use screen-reader on [truncated numberfield](https://swcpreviews.z13.web.core.windows.net/pr-6038/docs/first-gen-storybook/?path=/story/number-field--truncated-value-tooltip&globals=system:spectrum-two) to determine that the number input is not double announced
- [ ] Check [truncated number-fields](https://swcpreviews.z13.web.core.windows.net/pr-6038/docs/first-gen-storybook/?path=/story/number-field--truncated-value-tooltip&globals=system:spectrum-two), when editing truncated content it should update the value in real time (_Note_: it doesn't re-place the commas in real-time... yet)
- [ ] Type in a number long enough (such as 12345678) to truncate in the [number-field story](https://swcpreviews.z13.web.core.windows.net/pr-6038/docs/first-gen-storybook/?path=/story/number-field--default&globals=system:spectrum-two) to confirm that the truncation tooltip appears in number-fields that are not initially truncated
- [ ] On that same [number-field story](https://swcpreviews.z13.web.core.windows.net/pr-6038/docs/first-gen-storybook/?path=/story/number-field--default&globals=system:spectrum-two) with a number that is long enough to truncate (such as 12345678), remove characters from the input so that it is short enough to not need truncation (for instance, change the input from 12345678 to 12345). Verify that the tooltip does not disappear until you mouse/key focus somewhere else, and when you focus on the number-field again (with the shorter number inputted), the tooltip does not reappear.
- [ ] Check [a textfield story](https://swcpreviews.z13.web.core.windows.net/pr-6038/docs/first-gen-storybook/?path=/story/textfield--truncated-value-tooltip&globals=system:spectrum-two) for the same. If text input truncates, you should see the tooltip while typing, and if the text becomes shortened enough to not need truncation, the tooltip still appears until the next time the input is focused.

### Device review

- [x] Did it pass in Desktop?

  1. Go to [truncated numberfield stories](https://swcpreviews.z13.web.core.windows.net/pr-6038/docs/first-gen-storybook/?path=/story/number-field--truncated-value-tooltip&globals=system:spectrum-two)
  2. Tab to the first number-field input
  4. When focused, the tooltip displaying the full value should be visible

  1. Go [truncated textfield stories](https://swcpreviews.z13.web.core.windows.net/pr-6038/docs/first-gen-storybook/?path=/story/textfield--truncated-value-tooltip&globals=system:spectrum-two)
  2. Turn on VoiceOver
  4. Expect truncated input values to be read - they actually are being read twice due to the overlay setting `aria-describedby`; addressing this felt a bit out-of-scope

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,7 +22,7 @@ parameters:
   # 3. Commit this change to the PR branch where the changes exist.
   current_golden_images_hash:
     type: string
-    default: 0c8451b052f0c19eeb3835b429e614461e461a48
+    default: bc3f05831171b8f420d00ee792b726e3b12da715
 
   wireit_cache_name:
     type: string


1st-gen/packages/textfield/src/Textfield.ts
```diff
@@ -369,13 +402,45 @@ export class TextfieldBase extends ManageHelpText(
...
   }
 
+  protected renderTruncatedValueTooltip(): TemplateResult | typeof nothing {
+    if (!this.isTruncated || this.disabled || !this.inputElement) {
+      return nothing;
+    }
+    return html`
+      <sp-overlay
+        id="truncated-value-tooltip"
+        aria-hidden="true"
```
rise-erpelding: I noticed that the truncated inputs were being read twice when using VoiceOver, I did apply `aria-hidden` which seems to remove it from the accessibility tree at least? My understanding is that this is due to the way that `sp-overlay` is written, touching overlay seems little out of scope 😬 

Image

1st-gen/packages/number-field/stories/number-field.stories.ts
```diff
@@ -453,7 +514,6 @@ export const validationIcons = (args: StoryArgs): TemplateResult => {
       Invalid Number Field without Stepper
     </sp-field-label>
     <sp-number-field
-      invalid
```
rise-erpelding, line 460: This was here twice

1st-gen/packages/number-field/test/number-field.test.ts
```diff
@@ -595,21 +728,21 @@ describe('NumberField', () => {
       el.value = 45;
       expect(el.value).to.equal(45);
       el.focus();
-      await sendKeys({ type: '7' }); // Visible text: EUR 45.007
```
rise-erpelding: Just linting fixes in this file below this point

1st-gen/packages/textfield/src/Textfield.ts
```diff
@@ -60,7 +60,7 @@ export class TextfieldBase extends ManageHelpText(
   @state()
   protected isTruncated = false;
 
-  private resizeObserver = new ResizeObserver(() => {
+  private truncationResizeObserver = new ResizeObserver(() => {
```
rise-erpelding: I guess `resizeObserver` will clash with combobox?

1st-gen/packages/textfield/src/Textfield.ts
```diff
@@ -369,13 +402,45 @@ export class TextfieldBase extends ManageHelpText(
     `;
   }
 
+  protected renderTruncatedValueTooltip(): TemplateResult | typeof nothing {
+    if (!this.isTruncated || this.disabled || !this.inputElement) {
+      return nothing;
+    }
```
Rajdeepc: This renders `this.displayValue` (which is` this.value.toString()`) into a tooltip. For `type="password"` fields, the browser masks characters in the input, but if the password is long enough to truncate, the tooltip will display the full password in plaintext.

1st-gen/packages/textfield/src/Textfield.ts
```diff
@@ -292,6 +309,22 @@ export class TextfieldBase extends ManageHelpText(
     return this.value.toString();
   }
 
+  protected get inputElementIsTruncated(): boolean {
+    if (!this.inputElement || this.multiline) {
```
Rajdeepc: This getter only checks `this.multiline`, not `this.type`. This needs a guard:

```suggestion
    if (!this.inputElement || this.multiline || this.type === 'password') {
      return false;
    }
```

1st-gen/packages/textfield/package.json
```diff
@@ -59,7 +59,9 @@
     "@spectrum-web-components/icon": "1.11.2",
     "@spectrum-web-components/icons-ui": "1.11.2",
     "@spectrum-web-components/icons-workflow": "1.11.2",
-    "@spectrum-web-components/shared": "1.11.2"
+    "@spectrum-web-components/overlay": "1.11.2",
+    "@spectrum-web-components/shared": "1.11.2",
+    "@spectrum-web-components/tooltip": "1.11.2"
```
Rajdeepc, line 64: Adding `@spectrum-web-components/overlay` and `@spectrum-web-components/tooltip` as runtime dependencies of textfield is a bit concerning architecturally as every consumer of these components now pay the cost of loading overlay and tooltip even if truncation never occurs.
It would be great if you can lazy load these dependencies when truncation is first detected.
```ts
private _tooltipDepsLoaded = false;

private async ensureTooltipDeps(): Promise<void> {
    if (!this._tooltipDepsLoaded) {
        await import('@spectrum-web-components/overlay/sp-overlay.js');
        await import('@spectrum-web-components/tooltip/sp-tooltip.js');
        this._tooltipDepsLoaded = true;
    }
}
```
You can create a mixin pattern also so that you are not baking this into the base class.

1st-gen/packages/textfield/src/Textfield.ts
```diff
@@ -55,6 +58,13 @@ export class TextfieldBase extends ManageHelpText(
   @state()
   protected appliedLabel?: string;
 
+  @state()
+  protected isTruncated = false;
+
+  private truncationResizeObserver = new ResizeObserver(() => {
```
Rajdeepc: This means every instance of Textfield like Textarea, password fields and all its subclasses will create a ResizeObserver at construction time even though those instances will never need truncation detection. You should create these lazily only when the component is single-line and connected to the DOM.

1st-gen/packages/textfield/src/Textfield.ts
```diff
@@ -386,6 +451,24 @@ export class TextfieldBase extends ManageHelpText(
...
+      changedProperties.has('quiet') ||
+      changedProperties.has('valid') ||
+      changedProperties.has('invalid') ||
+      changedProperties.has('multiline') ||
+      changedProperties.has('size')
+    ) {
+      this.updateComplete.then(() => {
+        this.refreshTruncationState();
+      });
+    }
```
Rajdeepc: Most of these `type, quiet, valid, invalid, multiline, size` cause visual layout changes that would already trigger the `ResizeObserver`. Only value changes the text content without necessarily changing the container size. This creates unnecessary layout thrashing as `refreshTruncationState()` reads `scrollWidth` and `clientWidth`, forcing the browser to synchronously compute layout.
Can you only trigger from `update()` on value changes. Let the `ResizeObserver` handle the rest.

1st-gen/packages/textfield/src/Textfield.ts
```diff
@@ -369,13 +402,45 @@ export class TextfieldBase extends ManageHelpText(
...
   }
 
+  protected override firstUpdated(changedProperties: PropertyValues): void {
+    super.firstUpdated(changedProperties);
+    this.truncationResizeObserver.observe(this);
+    if (this.inputElement) {
+      this.truncationResizeObserver.observe(this.inputElement);
+    }
+    this.refreshTruncationState();
+  }
```
Rajdeepc: In case of multiple textfields this will still attach the ResizeObserver and runs `refreshTruncationState()` but `unputElementIsTruncated` will return `false` for multiline. The observer will still fire on every resize. Can you guard this for multiline?
```ts
if(!this.multiline) {
    this.truncationResizeObserver.observe(this);
}

File: 1st-gen/packages/textfield/test/textfield.test.ts [hunk 11]
```diff
@@ -27,7 +27,7 @@ import {
   testForLitDevWarnings,
 } from '../../../test/testing-helpers.js';
 import { findDescribedNode } from '../../../test/testing-helpers-a11y.js';
-import { Textfield, TextfieldType } from '../';
+import { Textfield, TextfieldType } from '../src/index.js';
```
Rajdeepc: Is this intentional change here in this PR? Can we tackle this in a separate PR?

1st-gen/packages/textfield/src/Textfield.ts
```diff
@@ -149,6 +159,13 @@ export class TextfieldBase extends ManageHelpText(
   @property({ type: Boolean, reflect: true })
   public readonly = false;
 
+  /**
+   * Placement of the tooltip shown when the value is truncated (e.g. 'bottom', 'top').
+   * Defaults to 'bottom' per Spectrum design.
+   */
+  @property({ attribute: 'tooltip-placement', reflect: true })
```
Rajdeepc: If you add `reflect: true`, every textfield instance will have a `tooltip-placement="bottom"` HTML attribute, even when the default is fine and truncation never occurs. This adds DOM noise. This should stay as `false`.

Rajdeepc (changes requested):
Liked the approach here. Most of the direction looks sound to me except for a few areas which needs attention. One major blocker is the the password field security issue and the added dependencies on the textfield is a bundle size concern for me, we can discuss this with the team and the eager ResizeObserver allocation will hinder performance. Try to optimize and think of a more performant approach.
Most of the other things looks solid and thanks for handling this a11y compliance. Happy to work with you in resolving the blocker.

1st-gen/packages/textfield/src/Textfield.ts
nikkimk: My concern with this is that a non-sighted user may think the same text is entered twice and attempt to delete the extra text.

nikkimk (changes requested):
I think we might need to discuss our options with this one given the a11y concerns.

5t3ph (commented):
In addition to being concerned about the issue for screen readers, I noticed that for number field the tooltip does not reflect real-time edits to the value (deletions or insertions).

The value does update real-time for textfield, but still with a bit of oddness that the tooltip is removed if the value decreases below the threshold (which is a little jarring, possible cognitive issue ex "wait, what just happened? did I do something wrong?"). But, it doesn't re-appear when the value increases above the threshold. Not reappearing might be ok, since the user is actively in the field and their cursor position is still allowing them to see the relevant part that they're editing.

1st-gen/packages/textfield/src/TruncatedValueTooltipMixin.ts
```diff
@@ -103,7 +107,7 @@ export function TruncatedValueTooltipMixin<
       this.requestUpdate();
     }
 
-    renderTruncatedValueTooltip(): TemplateResult | typeof nothing {
+    protected renderTruncatedValueTooltip(): TemplateResult | typeof nothing {
```
rise-erpelding: I assumed we wanted this to be protected, but am happy to revert if not.

1st-gen/packages/number-field/src/NumberField.ts
```diff
@@ -546,6 +546,7 @@ export class NumberField extends TextfieldBase {
       this._trackingValue = value;
       this.inputElement.value = value;
       this.inputElement.setSelectionRange(selectionStart, selectionStart);
+      this.syncTruncatedValueTooltipText();
```
rise-erpelding: Number-field's tooltip wasn't updating the text in real-time if edited. This addresses that issue.

My understanding is that it's needed in number-field but not textfield because of the value formatting that number-field does.

1st-gen/packages/number-field/src/NumberField.ts
```diff
@@ -547,6 +547,7 @@ export class NumberField extends TextfieldBase {
       this.inputElement.value = value;
       this.inputElement.setSelectionRange(selectionStart, selectionStart);
       this.syncTruncatedValueTooltipText();
+      this.refreshTruncatedValueTooltipState();
```
rise-erpelding: Number-field's tooltip also wasn't appearing in some cases because it wasn't checking properly for truncation, so `refreshTruncatedValueTooltipState` addresses that.

1st-gen/packages/textfield/src/spectrum-textfield.css
```diff
@@ -275,6 +275,10 @@
   appearance: textfield;
 }
 
+:host([type="password"]) .input {
+  text-overflow: clip;
+}
+
```
rise-erpelding, line 281: Once I added a password field in Storybook, I realized how odd it looks when it truncates:

Image

This removes that truncation. This felt a little out of scope though so I'm happy to revert if we don't want this change in here!

1st-gen/packages/number-field/src/NumberField.ts
rise-erpelding: Noting also that the tooltip isn't parsing & formatting the values in real-time, so it's possible that when editing the number you'll see odd formatting in the tooltip like this:

Image

I could address it here, it could be addressed later, or something else. Any thoughts?

.changeset/some-horses-shine.md
```diff
@@ -0,0 +1,16 @@
...
+**Added**: Added truncated-value tooltip behavior for single-line text inputs.
+
+- `<sp-textfield>` now shows a tooltip with the full value when the displayed value is visually truncated.
+- Tooltip placement can be controlled with the `tooltip-placement` attribute on `<sp-textfield>` or `<sp-number-field>`
+- `<sp-number-field>` inherits the same truncation tooltip behavior from `TextfieldBase`, including formatted numeric/currency values.
+- `type="password"` textfields no longer use ellipsis truncation styling (`text-overflow: clip`) to avoid awkward visual truncation.
+
+**Changed**: `sp-overlay` (internal)
+
+- Added internal property `describeTrigger` (`'auto' | 'none'`, default `'auto'`). When set to `'none'`, the overlay does not set `aria-describedby` on the trigger when open, avoiding double announcement for screen readers when the overlay content duplicates the trigger (e.g. truncated-value tooltips). Textfield’s truncated-value tooltip uses this so the tooltip is visual-only for a11y.
```
marissahuysentruyt: I'm not sure if it's helpful to mention that we touched `HoverController` and `LongpressController`. I'm not actually sure what are preferences are on those types of changes- is it all sort of "understood" because we made updates to sp-overlay?

1st-gen/packages/number-field/README.md
```diff
@@ -341,6 +341,38 @@ The number field works with the following interactions:
 
 The input value incrementally increases or decreases by the value of the `step` attribute. The <kbd>shift</kbd> key can be used to apply steps at 10 times (or the value of the `step-modifier` attribute times) their normal rate.
 
+#### Truncated value tooltip
+
+Like textfield, number field will show a tooltip with the full formatted value when the single-line input content is visually truncated. This supports long formatted content (for example currency values) in constrained widths and supports accessibility when users adjust text spacing (WCAG 1.4.12).
```
marissahuysentruyt: Non-blocking: we could like to the textfield docs, as well as the WCAG criteria!

.changeset/some-horses-shine.md
```diff
@@ -0,0 +1,16 @@
+---
+'@spectrum-web-components/overlay': patch
+'@spectrum-web-components/textfield': patch
+'@spectrum-web-components/number-field': patch
+---
+
+**Added**: Added truncated-value tooltip behavior for single-line text inputs.
```
marissahuysentruyt, line 7: We could also mention the new `TruncatedValueTooltipMixin`

rise-erpelding: This `TruncatedValueTooltipMixin` and the updates to the `HoverController`/`LongpressController` are a little more internal, I'm not sure that we strictly need to include it given that existing documentation we have on changesets [says to focus on consumer impact](https://github.com/adobe/spectrum-web-components/blob/main/1st-gen/projects/documentation/content/guides/writing-changesets.md)... but I can also see how mentioning it could be helpful from our perspective if there's something we need to debug. I mentioned these but more parenthetically.

1st-gen/packages/textfield/src/TruncatedValueTooltipController.ts
```diff
@@ -0,0 +1,221 @@
...
+    }
+    this.refreshTruncationState();
+  }
+
+  hostDisconnected(): void {
+    if (this._resizeObserver) {
+      this._resizeObserver.disconnect();
+      this._resizeObserver = null;
+    }
+    this._observerInitialized = false;
```
rubencarvalho, line 239: If the host is disconnected and then reconnected, the observer won't be recreated until `hostUpdated()` runs. This seems fine for most cases, but only if the host actually triggers an update cycle after reconnection. If it doesn't (e.g. no property changes), the observer won't be re-attached.

1st-gen/packages/number-field/src/NumberField.ts
```diff
@@ -546,6 +546,18 @@ export class NumberField extends TextfieldBase {
...
       this.inputElement.value = value;
       this.inputElement.setSelectionRange(selectionStart, selectionStart);
+      this.truncatedValueTooltipController.syncTooltipText(
+        this.inputElement?.value ?? this.displayValue
+      );
+      const justBecameTruncated =
+        this.truncatedValueTooltipController.refresh();
+      if (justBecameTruncated) {
+        this.updateComplete.then(() => {
+          this.truncatedValueTooltipController.syncTooltipText(
```
rubencarvalho, line 556: nit: if it `justBecameTruncated`, we'd be syncing twice (it's already happening above on line 549).

1st-gen/packages/textfield/src/TruncatedValueTooltipController.ts
```diff
@@ -0,0 +1,221 @@
...
+    const host = this.host;
+    if (host.multiline || this._observerInitialized) {
+      return;
+    }
+    this._observerInitialized = true;
+    this._resizeObserver = new ResizeObserver(() => {
+      this.refreshTruncationState();
+    });
+    this._resizeObserver.observe(this.host as unknown as Element);
+    if (host.inputElement) {
```
rubencarvalho: Are we sure this is guaranteed to exist on hostUpdated? If `host.inputElement` is `null` on the first `hostUpdated()` call, only the host element gets observed. On subsequent renders, `_observerInitialized` is already `true` so the early return skips observing  inputElement even once it exists.

1st-gen/packages/overlay/src/HoverController.ts
```diff
@@ -131,6 +131,13 @@ export class HoverController extends InteractionController {
     if (!this.overlay || !this.overlay.elements.length) {
       return;
     }
+    // When describeTrigger is 'none', do not set aria-describedby on the trigger (visual hint only).
+    const overlayWithDescribe = this.overlay as {
+      describeTrigger?: 'auto' | 'none';
+    };
```
rubencarvalho: nit: couldn't we use the `Overlay` type directly?

1st-gen/packages/textfield/src/TruncatedValueTooltipController.ts
caseyisonit: the host is required have an inputElement based on the interface and null is not an accepted value so it should throw an error, i guess we could add a check with a console warning? but i dont see how it would ever be null since its only applied to textfield

1st-gen/packages/textfield/src/TruncatedValueTooltipController.ts
```diff
@@ -0,0 +1,221 @@
...
+ *
+ * The host must implement TruncatedValueTooltipHost (e.g. TextfieldBase). The host
+ * includes the controller's render output in its template and may call refresh() when
+ * value or focused changes; NumberField also calls syncTooltipText() from handleInput()
+ * so the tooltip stays in sync with the live input without re-renders.
+ */
+export class TruncatedValueTooltipController implements ReactiveController {
+  private host: HostElement;
+
+  private isTruncated = false;
```
caseyisonit: nit: should this be `_isTruncated` since its private property? would need to be updated in the logic below too

caseyisonit:
Edge case behvavior im observing.... 

- tab to input
- hit backspace (empty input)
- tab to another input
- tab back to cleared input
- notice no tooltip (this might be related to some of the lifecycle call out from ruben? not sure. could be related to displayValue being `null`?)

rise-erpelding:
> Edge case behvavior im observing....
> 
> * tab to input
> * hit backspace (empty input)
> * tab to another input
> * tab back to cleared input
> * notice no tooltip (this might be related to some of the lifecycle call out from ruben? not sure. could be related to displayValue being `null`?)

@caseyisonit This behavior is expected with how it's currently implemented. The idea is that the tooltip appears only when text in the input is truncated, and doesn't appear otherwise. However, it's jarring to the user if they are editing the truncated text by backspacing to make it shorter and the tooltip suddenly disappears because it's no longer needed. As a compromise, it's set up so that the tooltip persists while editing even if the truncation state changes, then the next time it's focused, it doesn't appear until the text is truncated again.

We didn't really have a lot of guidelines on how behavior should work for this feature aside from "add tooltip when truncated", I'm wondering if there's enough that's problematic here to add another ticket to address some of the usability/interaction issues? Or are any of them problematic enough that they need to be addressed here and now?

So far I'm tracking:
* Problems with number-field where the number formatting/comma placement gets really weird in the tooltip as you're editing the number.
* Unclear how we want the tooltip to behave when we go from truncated with a tooltip to not truncated and not needing a tooltip:
    * Do we definitely persist the tooltip once the text becomes short enough to not need truncation
    * Do we persist the tooltip even if the user deletes all of the input text (this would result in an empty tooltip)
    * Do we persist the tooltip if the text was originally truncated, edited to be shorter, then user focuses away from the textfield, then back to it?

1st-gen/packages/textfield/src/TruncatedValueTooltipController.ts
rise-erpelding: Ok, I added a `host.requestUpdate()` to `hostConnected()`, would that sufficiently handle a reconnection scenario?

1st-gen/packages/number-field/src/NumberField.ts
rise-erpelding: I moved the first one into the `else`, that should prevent it from syncing twice

1st-gen/packages/textfield/src/TruncatedValueTooltipController.ts
rise-erpelding: Added a `console.warn`

1st-gen/packages/textfield/src/TruncatedValueTooltipController.ts
```diff
@@ -0,0 +1,229 @@
...
+      return;
+    }
+    const tooltipTextNode =
+      Array.from(tooltip.childNodes).find(
+        (node) =>
+          node.nodeType === Node.TEXT_NODE &&
+          Boolean((node.textContent ?? '').trim().length)
+      ) ??
+      Array.from(tooltip.childNodes).find(
+        (node) => node.nodeType === Node.TEXT_NODE
```
Rajdeepc, line 201: You are manually walking the nodes and mutating the text node directly. This couples the controller to the internal DOM of tooltip. Here now if tooltip restructures its internal DOM like wrapping in a div or span this will break. 
At minimum, add a comment explaining the trade-off. Ideally, you can consider setting `tooltip.textContent` directly (which clears children and sets text) instead of walking child nodes, since the tooltip content here is always plain text. Or store a reference to the tooltip element after render and use a property/attribute if tooltip supports it.

1st-gen/packages/textfield/src/TruncatedValueTooltipController.ts
```diff
@@ -0,0 +1,229 @@
...
+    if (tooltipTextNode) {
+      tooltipTextNode.textContent = text;
+    }
+  }
+
+  hostConnected(): void {
+    // Defer ResizeObserver setup to hostUpdated so inputElement is in the DOM.
+    // On reconnection, request an update so hostUpdated runs and we re-attach the observer.
+    if (!this._observerInitialized) {
+      this.host.requestUpdate();
```
Rajdeepc, line 212: The hostConnected method calls `this.host.requestUpdate()` only when `!this._observerInitialized`. But` hostDisconnected` sets` _observerInitialized = false`, so on reconnection `requestUpdate()` will be called. However, between disconnection and reconnection,` _isTruncated` retains its previous value. If the component reconnects in a different layout context the stale` _isTruncated = true` will render a tooltip that might no longer be accurate until the ResizeObserver fires.
Can you please reset `_isTruncated = false` in `hostDisconnected()` alongside the observer teardown so the component starts clean on reconnection.

1st-gen/packages/number-field/test/number-field.test.ts
```diff
@@ -1003,7 +1263,7 @@ describe('NumberField', () => {
     });
     it('manages currency', async () => {
       const el = await getElFrom(currency({ value: 234.21 }));
-      expect(el.formattedValue).to.equal('EUR 234.21');
+      expect(el.formattedValue).to.equal('EUR\u00a0234.21');
```
Rajdeepc: This is unrelated to the truncation feature.

1st-gen/packages/textfield/src/TruncatedValueTooltipController.ts
Rajdeepc: @rise-erpelding Adding a console.warn wont solve the problem. You can try a couple of things here. You can defer `_observerInitialized = true` until inputElement is confirmed present or re-checking on the next `hostUpdated()` if inputElement was initially null

1st-gen/packages/textfield/src/Textfield.ts
```diff
@@ -374,20 +396,30 @@ export class TextfieldBase extends ManageHelpText(
...
-    ) {
+      (changedProperties.has('required') && this.required);
+
+    if (valueOrRequiredChanged) {
       this.updateComplete.then(() => {
         this.checkValidity();
+        if (changedProperties.has('value')) {
+          this.truncatedValueTooltipController.refresh();
+        }
       });
```
Rajdeepc, line 420: `this.updateComplete` resolves after the current render cycle completes, which means refresh() runs after the DOM is painted. But `refresh()` then calls `requestUpdate()` via `Promise.resolve().then(...)`, which means a truncation state change triggers a third microtask-chained render. 
This is correct but creates a 3-frame delay: `render → updateComplete → refresh → Promise.resolve → requestUpdate → render`. For most cases this is invisible, but it could cause a flash where the tooltip briefly doesn't appear.
If this is intentional please document this timing in a code comment so future maintainers understand the intentional multi-phase update.

Rajdeepc (changes requested):

1st-gen/packages/number-field/stories/number-field.stories.ts
```diff
@@ -523,29 +523,24 @@ export const validationIcons = (args: StoryArgs): TemplateResult => {
       Invalid Number Field without Stepper
     </sp-field-label>
     <sp-number-field
+      invalid
```
rise-erpelding, line 526: I'd previously removed `invalid`, the args were taking care of it making it unnecessary to have as an attribute. When looking at this again, I realized that `validationIcons.args` weren't applying to all the instances of `sp-number-field` here and made some changes... but given that this PR already has a number of changes and this isn't hurting anything, I'm reverting these updates for now.

1st-gen/packages/textfield/src/TruncatedValueTooltipController.ts
```diff
@@ -196,18 +196,16 @@ export class TruncatedValueTooltipController implements ReactiveController {
     if (host.multiline || this._observerInitialized) {
       return;
     }
+    // Defer full setup until inputElement is in the DOM so we can observe it for truncation.
+    if (!host.inputElement) {
+      return;
```
rise-erpelding, line 223: Would we want to request an update here? (Before returning)

I considered it but also didn't want to create potential for an infinite loop if there was some scenario where `!host.inputElement` was always true. I could add state here to ensure that we only retry once, but wanted to pause and ask for feedback on this before doing so.

1st-gen/packages/textfield/src/Textfield.ts
rise-erpelding: I added comments in textfield, number-field, and the controller in a separate commit (ad8157a63fd943d66a4948de95aa5eaf8396afec), let me know if that works with what you had in mind.

1st-gen/packages/textfield/src/TruncatedValueTooltipController.ts
Rajdeepc: The current approach is fine and no `requestUpdate()` needed here. Since inputElement is a `@query('.input:not(#sizer)')` on TextfieldBase, it's guaranteed to be in the shadow DOM by the time` hostUpdated()` runs. So the null case is effectively unreachable for both Textfield and NumberField.

The early return without setting `_observerInitialized` is a good defensive pattern if a future subclass ever delays rendering the input, this will naturally retry on the next property-driven update cycle without any risk of an infinite loop since no requestUpdate() is called from here. Adding a retry counter would be over-engineering for a case that can't happen with the current hosts.

Looks good as-is.

Rajdeepc (approved):
Thanks for addressing my queries. Loved the clean well-documented fixes. Great job!
I don't see any more blockers here. Please test this on your end and if it looks good ship it.

1st-gen/packages/textfield/src/TruncatedValueTooltipController.ts
rise-erpelding: @Rajdeepc While I initially tried the approach using `tooltip.textContent`, I realized that when changing a number in the number-field, while it would reflect the value correctly, it wouldn't ever reformat with the commas in the correct spot, even after focusing away from the component and refocusing:

image

The `textContent` approach also causes a runtime error:

image

After looking into it I think the better approach for now is to document the tradeoff to avoid both of these issues so I've reverted that commit and added some additional comments.

PR #6043, opened 2026-02-23
fix(number-field): avoid empty aria-label on internal input

blunteshwar:
## Summary

Fixes an accessibility regression in `sp-number-field` where the internal input could render with `aria-label=""`, resulting in no programmatic accessible name.

This change ensures the internal form control does not emit an empty `aria-label` and correctly picks up label text supplied via associated `sp-field-label`.

## Problem

When `sp-number-field` was rendered before label propagation completed, the inherited textfield template could output:

```html
<input class="input" aria-label="" ...>
```

An explicit empty `aria-label` can suppress meaningful naming behavior, causing screen readers to announce generic text like "edit blank" instead of the visible label context.

## Root Cause

`TextfieldBase` rendered `aria-label=${this.label || this.appliedLabel || this.placeholder}` directly. When all values were empty at render time, Lit produced `aria-label=""` instead of omitting the attribute.

`sp-number-field` inherits this rendering path, so it was affected.

## Fix

- Updated `TextfieldBase` input and textarea templates to conditionally set `aria-label` using `ifDefined(...)`.
- This omits the attribute when no label value exists, preventing `aria-label=""` from being emitted.
- Once `sp-field-label` applies label text via `applyFocusElementLabel`, the input receives a non-empty `aria-label`.

## Tests

Added regression coverage in number-field tests:

- `applies a non-empty input label from associated field label text`
  - Renders `sp-field-label` + `sp-number-field`
  - Waits for label propagation
  - Asserts internal `.input` has `aria-label` equal to visible label text and not empty

## Files Changed

- `1st-gen/packages/textfield/src/Textfield.ts`
- `1st-gen/packages/number-field/test/number-field.test.ts`

## Validation Notes

Local test execution is currently blocked by an existing repo/workspace lockfile issue:

- `@swc-react/accordion ... doesn't seem to be present in your lockfile; run "yarn install"`

So this PR includes code and regression tests, but full local test run could not be completed in the current environment.


1st-gen/packages/number-field/test/number-field.test.ts
```diff
@@ -1761,6 +1762,27 @@ describe('NumberField', () => {
...
+      const el = wrapper.querySelector('sp-number-field') as NumberField;
+      await waitUntil(
+        () =>
+          (el.shadowRoot?.querySelector('.input') as HTMLInputElement | null)
+            ?.ariaLabel === labelText
+      );
+
+      const input = el.shadowRoot?.querySelector('.input') as HTMLInputElement;
+      expect(input.getAttribute('aria-label')).to.equal(labelText);
+      expect(input.getAttribute('aria-label')).to.not.equal('');
```
rubencarvalho, line 1783: This `expect` is redundant 😃

blunteshwar: I know that this is logically redundant but If the fixture text is ever refactored (or becomes data-driven), the non-empty invariant still stands as the accessibility contract.

Issue #3596, opened 2023-08-25
[Bug]: VoiceOver does not read content of sp-tooltip on hover or click open


pwong09:
### Code of conduct

- [X] I agree to follow this project's code of conduct.

### Impacted component(s)

sp-tooltip, sp-action-button, sp-overlay-trigger

### Expected behavior

Expect VoiceOver to read the content of a sp-tooltip when user either hovers or keyboard focuses on the sp-action-button

### Actual behavior

VoiceOver tells me to use `control-option-space` to interact with the button, but no tooltip content is read out

### What browsers are you seeing the problem in?

Chrome

### How can we reproduce this issue?

See sample code

### Sample code that illustrates the problem

https://studio.webcomponents.dev/edit/J3BcJnTW4U45u7DTa9a0/src/index.ts?p=stories

Westbrook:
First, the demo as provided provides click content to the Button in the form of the Tooltip:
```html
         <overlay-trigger placement="bottom" type="modal">
            <sp-action-button slot="trigger" label="Information" quiet size="s">
              <sp-icon-info-outline slot="icon"></sp-icon-info-outline>
            </sp-action-button>
            <sp-tooltip slot="click-content" size="s" open>
              Style combinations defined in the font for quick access
            </sp-tooltip> 
        </overlay-trigger>
```
This means that when the Button is activated the screen reader cursor will be passed into the Tooltip and the content will be read.

When [moved to hover content](https://studio.webcomponents.dev/edit/UsSCkXGoBbGFmSqc4vSA/src/index.ts?p=stories), what I'm seeing right now looks like a weird timing reality:
```html
        <overlay-trigger placement="bottom" type="modal">
            <sp-action-button slot="trigger" label="Information" quiet size="s">
              <sp-icon-info-outline slot="icon"></sp-icon-info-outline>
            </sp-action-button>
            <sp-tooltip slot="hover-content" size="s" open>
              Style combinations defined in the font for quick access
            </sp-tooltip> 
        </overlay-trigger>
```
In this context, I get the Tooltip content read _BUT ONLY_ when the Tooltip is open first. So, if I move the screen reader cursor into the Button, wait for the Tooltip to open, move it _before_ the Button, then move it back onto the Button while the Tooltip is open, I get the text read, but in no other cases. We'll need to dig deeper on what that's saying, but wanted to get a first bit of usage and triage info into this issue.

Westbrook:
[...]

PR #6044, opened 2026-02-23
fix(overlay): add ARIA attribute management to overlay-trigger for screen reader accessibility

Rajdeepc:
## Description
The `<overlay-trigger>` component did not set any ARIA attributes on the trigger element for click/longpress overlays. This caused screen readers (e.g., NVDA on Chrome/Windows) to not announce the overlay relationship or read the overlay content when expanded, violating WCAG 1.3.2 (Meaningful Sequence, Level A).

## Motivation and context
A screen reader user navigating the [dialog behaviors documentation](https://opensource.adobe.com/spectrum-web-components/components/dialog/#behaviors) could not hear the overlay content after opening it via the trigger button. The root cause was that the trigger element lacked `aria-expanded`, `aria-controls`, and `aria-haspopup` attributes, so the screen reader had no programmatic way to discover or announce the controlled content.

This fix follows the same pattern used by other component libraries (Shoelace, Radix, Headless UI) where the component owns the ARIA contract for its trigger-content relationship.

This PR adds automatic ARIA attribute management to `<overlay-trigger>`:

- **`aria-expanded`** is set on the trigger element, toggling between `"false"` (closed) and `"true"` (open)
- **`aria-controls`** is set on the trigger, pointing to the overlay content element's `id` (auto-generated with `sp-overlay-content-` prefix if not provided)
- **`aria-haspopup`** is set to `"dialog"` by default (consumer-set values are never overwritten)
- ARIA attributes are managed only for click and longpress overlays; hover-only overlays (tooltips) are excluded since they use `aria-describedby` instead
- All managed ARIA attributes are cleaned up when:
  - The overlay-trigger is disconnected from the DOM
  - The trigger element is swapped (ARIA removed from old trigger, applied to new)
  - Click/longpress content is removed

## Related issue(s)

- fixes [#3596 ]

---

## Author's checklist

---

## Reviewer's checklist

#### ARIA attribute verification

1. Navigate to a page with overlay-trigger containing click content
2. Inspect the trigger element in DevTools:
   - Confirm aria-expanded=\"false\" is set initially
   - Confirm aria-haspopup=\"dialog\" is set
   - Confirm aria-controls points to the content element's id
3. Click the trigger to open the overlay:
   - Confirm aria-expanded updates to \"true\"
4. Close the overlay:
   - Confirm aria-expanded returns to \"false\"

#### Screen reader testing (NVDA + Chrome on Windows)

1. Turn on NVDA and navigate to the overlay-trigger button using Tab
2. Confirm NVDA announces the button with \"collapsed\" state
3. Press Enter/Space to open the overlay
4. Confirm NVDA announces \"expanded\" state
5. Navigate within the overlay content using arrow keys
6. Confirm the dialog heading and body text are announced
7. Press Escape to close
8. Confirm focus returns to the trigger and NVDA announces \"collapsed\"

### Device review

- [ ] Did it pass in Desktop?

  1. Go [here](url)
  2. Do this action
  3. Expect this result

  1. Go [here](url)
  2. Do this action
  3. Expect this result

### DIFF
--- a/CONTRIBUTOR-DOCS/01_contributor-guides/07_authoring-contributor-docs/verify-links.js
+++ b/CONTRIBUTOR-DOCS/01_contributor-guides/07_authoring-contributor-docs/verify-links.js
@@ -170,7 +170,7 @@ function extractLinksFromContent(filepath, content) {
  * Validate a single link.
  * Returns validation result with error details if invalid.
  */
-function validateLink(sourceFile, link, metadata) {
+function validateLink(sourceFile, link, metadata, docsRoot) {
     const { href } = link;
 
     // Parse link into file path and anchor
@@ -216,6 +216,16 @@ function validateLink(sourceFile, link, metadata) {
     }
 
     if (!targetFileMeta) {
+        // Target may be outside CONTRIBUTOR-DOCS (e.g. 2nd-gen component files)
+        if (docsRoot) {
+            const fullPath = path.isAbsolute(resolvedPath)
+                ? path.resolve(resolvedPath)
+                : path.resolve(path.dirname(path.join(docsRoot, sourceFile)), linkPath);
+            if (fs.existsSync(fullPath)) {
+                return { valid: true };
+            }
+        }
+
         // File not found - try to find similar files for suggestions
         const targetFilename = path.basename(linkPath, '.md');
         const suggestions = Object.keys(metadata.files)
--- a/CONTRIBUTOR-DOCS/01_contributor-guides/12_tools-vs-packages.md
+++ b/CONTRIBUTOR-DOCS/01_contributor-guides/12_tools-vs-packages.md
@@ -42,7 +42,7 @@ When deciding where something lives in 2nd-gen, ask: **does it depend on the ren
 
 The key distinction: **UI-less** code goes in `core/` (it works the same regardless of Spectrum version); **UI** code that depends on Spectrum styles goes in `swc/` (even if it's a utility, not a design component).
 
-For placement within `core/`, see [packages/core/MIGRATION.md](../../../2nd-gen/packages/core/MIGRATION.md).
+For placement within `core/`, see [packages/core/MIGRATION.md](../../2nd-gen/packages/core/MIGRATION.md).
 
 ## 2nd-gen layout
 
--- a/CONTRIBUTOR-DOCS/02_style-guide/01_css/01_component-css.md
+++ b/CONTRIBUTOR-DOCS/02_style-guide/01_css/01_component-css.md
@@ -11,15 +11,29 @@
 <details open>
 <summary><strong>In this doc</strong></summary>
 
-- [Contributor TL;DR](#contributor-tldr)
-- [Rule Order](#rule-order)
-- [Managing Specificity](#managing-specificity)
+- [Component CSS](#component-css)
+  - [Contributor TL;DR](#contributor-tldr)
+  - [Rule Order](#rule-order)
+  - [CSS property ordering](#css-property-ordering)
+  - [Class naming patterns](#class-naming-patterns)
+  - [Comment conventions](#comment-conventions)
+  - [Selector patterns](#selector-patterns)
+    - [When to use `:host`](#when-to-use-host)
+    - [When to use `:host([attribute])`](#when-to-use-hostattribute)
+    - [When to use classes vs attributes](#when-to-use-classes-vs-attributes)
+    - [Managing specificity with `:where()`](#managing-specificity-with-where)
+  - [Variant implementation patterns](#variant-implementation-patterns)
+  - [State implementation patterns](#state-implementation-patterns)
+  - [Size variant patterns](#size-variant-patterns)
+  - [Animation and transition patterns](#animation-and-transition-patterns)
+  - [Forced colors requirements](#forced-colors-requirements)
+  - [Managing Specificity](#managing-specificity)
     - [Shadow DOM Specificity and Custom Property Inheritance](#shadow-dom-specificity-and-custom-property-inheritance)
     - [Using Cascade Layers (`@layer`)](#using-cascade-layers-layer)
-- [Component Specs vs. Component Styles](#component-specs-vs-component-styles)
-- [Color Themes](#color-themes)
+  - [Component Specs vs. Component Styles](#component-specs-vs-component-styles)
+  - [Color Themes](#color-themes)
     - [Modifying Non-Color Properties](#modifying-non-color-properties)
-- [Closing Note for Contributors](#closing-note-for-contributors)
+  - [Closing Note for Contributors](#closing-note-for-contributors)
 
 </details>
 
@@ -29,7 +43,7 @@ The following are high-level guidelines for the CSS creation for components.
 
 ## Contributor TL;DR
 
-> For examples of all these rules in practice, review the [reference migration for Badge](04_spectrum-swc-migration.md#reference-migration-badge) as well as examples for avoiding issues in the [anti-patterns guide](05_anti-patterns.md).
+> For examples of all of these rules in practice, review [Badge](../../../2nd-gen/packages/swc/components/badge/badge.css) and [Status Light](../../../2nd-gen/packages/swc/components/status-light/status-light.css) as reference implementations, the [reference migration for Badge](04_spectrum-swc-migration.md#reference-migration-badge), and the [anti-patterns guide](05_anti-patterns.md).
 
 - `:host` is for defining how the container participates in the global layout, not the core component styles
 - Follow the prescribed rule order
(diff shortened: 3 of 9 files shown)


Rajdeepc:
## Design decisions for team review

This PR adds automatic ARIA attribute management to `<overlay-trigger>` for accessibility. Before merging, I'd like the team's input on a few design decisions:

### 1. Automatic ARIA management vs. opt-in

The component now **automatically** sets `aria-expanded`, `aria-controls`, and `aria-haspopup` on the trigger element for click/longpress overlays. This follows the pattern used by most component libraries (e.g., Shoelace, Radix, Headless UI) where the component owns the ARIA contract.

**Question**: Are we comfortable with OverlayTrigger taking ownership of these attributes, or should this be opt-in?

### 2. `aria-haspopup` default value

The implementation defaults `aria-haspopup` to `"dialog"` for modal/page overlay types, and `"true"` for auto/manual types. Per the ARIA spec, `"true"` maps to `"menu"`, which may not always be accurate.

**Question**: Is `"dialog"` the right default for modal overlays? Should the value be configurable via an attribute, or is the current type-based mapping sufficient?

### 3. Consumer override behavior

A `WeakSet` tracks which elements have had `aria-haspopup` set by the component. If a consumer has already set `aria-haspopup` before the component manages it, the consumer's value is preserved. However, once the component sets it, subsequent changes by the component will update it.

**Question**: Is this override-respecting approach correct, or should consumers always be able to override after initial render too?

### 4. Cleanup on disconnect and trigger swap

The PR adds cleanup in `disconnectedCallback` (removes all managed ARIA attributes) and handles trigger element swaps (removes ARIA from old trigger, applies to new one). This adds some complexity but prevents stale attributes.

**Question**: Any concerns about the cleanup scope or approach?

---

These are all non-breaking, additive changes following standard WAI-ARIA patterns. No RFC is needed per our contribution guidelines, but I wanted to get alignment on these decisions before requesting formal review. Feedback welcome async here or in Slack.

1st-gen/packages/overlay/src/OverlayTrigger.ts
```diff
@@ -177,6 +186,66 @@ export class OverlayTrigger extends SpectrumElement {
...
   }
 
+  private removeAriaFromTrigger(element: HTMLElement): void {
+    element.removeAttribute('aria-expanded');
+    element.removeAttribute('aria-controls');
+    if (this.ariaHaspopupManagedElements.has(element)) {
+      element.removeAttribute('aria-haspopup');
+      this.ariaHaspopupManagedElements.delete(element);
+    }
+  }
```
blunteshwar, line 224: `removeAriaFromTrigger()` always removes `aria-expanded` and `aria-controls`, and `manageAriaOnTrigger()` calls it whenever there is no click/longpress content. That means hover-only usage can strip consumer-authored ARIA state even though the docs say ARIA management is for click/longpress interactions.

1st-gen/packages/overlay/src/OverlayTrigger.ts
```diff
@@ -177,6 +186,66 @@ export class OverlayTrigger extends SpectrumElement {
...
+
+    if (
+      this.ariaHaspopupManagedElements.has(triggerElement) ||
+      !triggerElement.hasAttribute('aria-haspopup')
+    ) {
+      triggerElement.setAttribute('aria-haspopup', 'dialog');
+      this.ariaHaspopupManagedElements.add(triggerElement);
+    }
+
+    const content = this.clickContent[0] || this.longpressContent[0];
```
blunteshwar: when both `click + longpress` exist, aria-controls should switch to longpress content id when open='longpress' and back for click.

```suggestion
    const content =
  this.open === 'longpress'
    ? this.longpressContent[0]
    : this.open === 'click'
      ? this.clickContent[0]
      : this.clickContent[0] || this.longpressContent[0];

```

1st-gen/packages/overlay/src/OverlayTrigger.ts
```diff
@@ -177,6 +186,66 @@ export class OverlayTrigger extends SpectrumElement {
     }
   }
 
+  private removeAriaFromTrigger(element: HTMLElement): void {
+    element.removeAttribute('aria-expanded');
+    element.removeAttribute('aria-controls');
```
blunteshwar, line 221: Why are we not doing like this so that hover-only does not remove consumer-provided aria-expanded/aria-controls
```suggestion
    if (this.ariaExpandedManagedElements.has(element)) {
    element.removeAttribute('aria-expanded');
    this.ariaExpandedManagedElements.delete(element);
  }
  if (this.ariaControlsManagedElements.has(element)) {
    element.removeAttribute('aria-controls');
    this.ariaControlsManagedElements.delete(element);
  }
```

blunteshwar (changes requested):

1st-gen/packages/overlay/src/OverlayTrigger.ts
Rajdeepc: Good point. I would rather do an early return here if `ariaManageElements` has no element
```ts
if (!this.ariaManagedElements.has(element)) {
      return;
    }
```

nikkimk:
> The implementation defaults `aria-haspopup` to `"dialog"` for modal/page overlay types, and `"true"` for auto/manual types. Per the ARIA spec, `"true"` maps to `"menu"`, which may not always be accurate.
> 
> **Question**: Is `"dialog"` the right default for modal overlays? Should the value be configurable via an attribute, or is the current type-based mapping sufficient?

What about [listbox](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/listbox_role), [tree],(https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/tree_role), and [grid]? 

Pickers and comboboxes should be `listbox`.

And what about tooltip which is not interactive and should not have `aria-haspopup`?

1st-gen/packages/overlay/src/OverlayTrigger.ts
```diff
@@ -177,6 +186,74 @@ export class OverlayTrigger extends SpectrumElement {
...
+    }
+
+    const isExpanded = this.open === 'click' || this.open === 'longpress';
+    triggerElement.setAttribute('aria-expanded', String(isExpanded));
+
+    if (
+      this.ariaManagedElements.has(triggerElement) ||
+      !triggerElement.hasAttribute('aria-haspopup')
+    ) {
+      triggerElement.setAttribute('aria-haspopup', 'dialog');
```
nikkimk: See my comment here: https://github.com/adobe/spectrum-web-components/pull/6044#issuecomment-3954301861

Rajdeepc: Addressed -- `aria-haspopup` now auto-resolves from the content element's role instead of defaulting to `"dialog"`. See my reply to your issue-level comment for the full breakdown. Commit: [322006b](https://github.com/adobe/spectrum-web-components/commit/322006bfae).

Rajdeepc:
@nikkimk Great feedback -- both points addressed in [322006b](https://github.com/adobe/spectrum-web-components/commit/322006bfae):

**1. listbox, tree, grid, and menu support**

`aria-haspopup` is no longer hardcoded to `"dialog"`. It now auto-resolves from the content element's role using a new `resolveHaspopupValue()` method:

- Checks the content element's `role` attribute first (e.g., `<div slot="click-content" role="menu">`)
- Falls back to the first descendant with a `[role]` attribute (covers the common pattern of `<sp-popover>` wrapping an `<sp-dialog>` or `<sp-menu>`)
- Recognized values: `menu`, `listbox`, `tree`, `grid`, `dialog`
- Defaults to `"dialog"` only when no recognized role is found

So a picker/combobox using overlay-trigger with a `role="listbox"` content element will correctly get `aria-haspopup="listbox"`, and an action-menu will get `aria-haspopup="menu"`.

Consumer overrides are still respected -- if someone sets `aria-haspopup` directly on the trigger element before the component manages it, the component won't overwrite it.

**2. Tooltips**

Tooltips were already handled correctly -- the ARIA management only runs for click and longpress content. Hover-only overlays (tooltips) never get `aria-expanded`, `aria-controls`, or `aria-haspopup` set on the trigger. Added an explicit test to confirm this behavior is preserved.

5 new tests were added covering: `role="dialog"`, `role="menu"`, `role="listbox"`, nested role detection through a wrapper element, and fallback to `"dialog"` for unrecognized roles. Docs updated as well.

blunteshwar (approved):

1st-gen/packages/picker/test/picker-responsive.test.ts
```diff
@@ -273,10 +273,10 @@ describe('Picker, responsive', () => {
       // Ensure menu is not in scrolling state (which would prevent selection).
       el.optionsMenu.isScrolling = false;
 
-      // Click the menu item.
-      const closed = oneEvent(el, 'sp-closed');
+      // Click the menu item and wait for the change event.
+      const changed = oneEvent(el, 'change');
       menuItem.click();
-      await closed;
+      await changed;
```
TarunAdobe, line 279: good stuff!

PR #6047, opened 2026-02-24
fix(combobox): align slotted field label with standalone label

aramos-adobe:
## Description

Aligns the combobox slotted field label with standalone <sp-field-label> (typography, colors, sizes) and fixes picker button placement and disabled high-contrast styling. 

This is also aims to align grid layout of the combobox referenced in SWC-1316

## Motivation and context

The combobox uses FieldLabelMixin, which renders the label in shadow DOM. The slotted label’s CSS did not match the standalone `<sp-field-label>` component, and the picker button could misalign when a label was present. Disabled state in forced-colors also needed to use `GrayText` for the picker button.

## Related issue(s)

-   fixes [SWC-1320]

---

## Still left TO-DO

-   [x] Fix side alignment of the label and picker button
-   [ ] Remove border on `.spectrum-PickerButton--quiet` disabled state

## Author's checklist

---

## Reviewer's checklist

### Manual review test cases

-   [ ] **Slotted label vs standalone:** Combobox default story and "With standalone field label" story — slotted label (e.g. `<span slot="field-label">`) matches standalone `<sp-field-label>` in typography, spacing, and colors.

    1. Open Storybook → Combobox → Default and "With standalone field label".
    2. Compare label appearance; change size (s/m/l/xl), disabled, required.
    3. Expect slotted and standalone labels to look the same.

-   [ ] **Picker button alignment:** Combobox with a slotted label — picker button stays aligned with the input row.

    1. Open Combobox → Default (or any story with a field label).
    2. Check that the chevron/button is vertically aligned with the input, not shifted.
    3. Expect no misalignment.

-   [ ] **Forced-colors disabled:** Disabled combobox in high-contrast mode — label and picker button use GrayText.

    1. Enable forced-colors (e.g. DevTools → Rendering → "Emulate CSS media feature `forced-colors: active`" or Windows High Contrast).
    2. Open Combobox → With standalone field label, turn "disabled" on.
    3. Expect label and picker button (border/icon) to use GrayText.

-   [ ] **No regressions:** Existing combobox stories (sizes, quiet, invalid, readonly, etc.) still look and behave correctly.

    1. Run through Combobox stories (Default, disabled, invalid, quiet, readonly, etc.).
    2. Expect no visual or behavioral regressions.
    3. Test VRTs locally by running `yarn test:start --group vrt-combobox-single --update-visual-baseline`

### Device review

### DIFF
--- a/1st-gen/packages/combobox/src/spectrum-combobox.css
+++ b/1st-gen/packages/combobox/src/spectrum-combobox.css
@@ -147,7 +147,7 @@
   --spectrum-combobox-spacing-label-to-combobox: calc(var(--swc-scale-factor) * -6px);
 }
 
-/* Per-size overrides so quiet matches production when token resolution differs (e.g. medium 5.5px, small 2px from design). */
+/* Per-size overrides so quiet matches the standalone field label spacing when token resolution differs (e.g. medium 5.5px, small 2px from design). */
 :host([quiet][size="s"]) {
   --spectrum-combobox-spacing-label-to-combobox: calc(var(--swc-scale-factor) * -2px);
 }


1st-gen/packages/combobox/stories/combobox.stories.ts
```diff
@@ -20,12 +20,15 @@ import { defineElement } from '@spectrum-web-components/base/src/define-element.
...
 import '@spectrum-web-components/combobox/sp-combobox.js';
+import '@spectrum-web-components/field-label/sp-field-label.js';
 import '@spectrum-web-components/help-text/sp-help-text.js';
 import '@spectrum-web-components/tooltip/sp-tooltip.js';
 import '@spectrum-web-components/menu/sp-menu-item.js';
 
+import { spreadProps } from '../../../test/lit-helpers.js';
 import { Combobox, ComboboxOption } from '../src/Combobox.js';
 import { argTypes } from './args.js';
+import { argTypes } from './args.js';
```
nikkimk: This is a duplicate of the previous line.

```suggestion
```

1st-gen/packages/combobox/stories/combobox.stories.ts
```diff
@@ -130,24 +133,75 @@ export const noAutocomplete = (): TemplateResult => {
...
   `;
 };
 
+/**
+ * Standalone sp-field-label: two patterns.
+ * 1) Sibling – label outside the combobox; use combobox `label` for a11y.
+ * 2) Slotted – sp-field-label in the field-label slot (compare visually to slotted span).
+ * All story args pass through so you can use controls to see disabled, invalid, size, etc.
+ */
+export const withStandaloneFieldLabel = (args: StoryArgs): TemplateResult => {
```
nikkimk, line 156: Thanks for adding this story. Can we also do a story comparing `side-aligned="start"` in both components?

nikkimk (changes requested):
Looking good. I'm seeing some spacing differences between the label and the combobox we'll need to fix though.

Image
Image
Image

1st-gen/packages/combobox/stories/combobox.stories.ts
```diff
@@ -130,24 +132,130 @@ export const noAutocomplete = (): TemplateResult => {
...
+          style="min-width: 80px; width: 200px;"
+          ${spreadProps(args)}
+        >
+          <sp-field-label
+            slot="field-label"
+            size=${args.size ?? 'm'}
+            ?disabled=${args.disabled}
+          >
+            Where do you live?
+          </sp-field-label>
```
rise-erpelding: Is there a reason to slot `sp-field-label` into `sp-combobox` like this?

For instance, can it just be a `<span slot="field-label">Where do you live?</span>` inside of `<sp-combobox>` instead?

The `size` and `disabled` should inherit from the `:host`, I think?

1st-gen/packages/combobox/stories/combobox.stories.ts
```diff
@@ -130,24 +132,130 @@ export const noAutocomplete = (): TemplateResult => {
...
+ * 1) Sibling – label outside the combobox; use combobox `label` for a11y.
+ * 2) Slotted – sp-field-label in the field-label slot (compare visually to slotted span).
+ * All story args pass through so you can use controls to see disabled, invalid, size, etc.
+ */
+export const withStandaloneFieldLabel = (args: StoryArgs): TemplateResult => {
+  return html`
+    <div style="display: flex; flex-direction: column; gap: 2rem;">
+      <div>
+        <p
+          style="margin: 0 0 0.5rem 0; font-size: 12px; color: var(--spectrum-global-color-gray-600);"
```
rise-erpelding: Could we replace these `--spectrum-global` custom props with the non-"global" ones? The ones with "global" in their name are deprecated, see https://opensource.adobe.com/spectrum-web-components/tools/core-tokens/#color-tokens


```suggestion
          style="margin: 0 0 0.5rem 0; font-size: 12px; color: var(--spectrum-gray-600);"
```

1st-gen/packages/combobox/src/combobox.css
```diff
@@ -14,28 +14,69 @@
...
 
 :host([readonly]) sp-picker-button {
   visibility: hidden;
   pointer-events: none;
 }
 
-sp-field-label {
+sp-field-label,
+:host > label {
+  -webkit-font-smoothing: subpixel-antialiased;
```
rise-erpelding: The field label mixin does have some styles and custom properties set in [`field-label-mixin.css`](https://github.com/adobe/spectrum-web-components/blob/nikkimk/form-field-mixin/1st-gen/packages/field-label/src/field-label-mixin.css#L48-L65), could we leverage some of those? Some of the same styles being set here also exist there, so they're being set once in the mixin then again from combobox, for instance when I inspect the Default Combobox story:

Image

1st-gen/packages/combobox/src/spectrum-combobox.css
```diff
@@ -78,10 +78,14 @@
...
 :host,
-:host {
+:host([size="m"]) {
   --spectrum-combobox-block-size: var(--spectrum-component-height-100);
   --spectrum-combobox-icon-size: var(--spectrum-workflow-icon-size-100);
   --spectrum-combobox-font-size: var(--spectrum-font-size-100);
+
+  /* Align slotted label with standalone sp-field-label */
+  --spectrum-field-label-font-size: var(--spectrum-font-size-100);
+  --spectrum-field-label-min-height: var(--spectrum-component-height-100);
```
rise-erpelding: These are also defined in the mixin styles the same way you have them here, so it may not be necessary to define them again. It also keeps the `--spectrum-field-label-` custom property definitions out of combobox css.

Also, the font-sizes specified here don't align with what's seen in `sp-field-label`, is there a reason for that? It's causing some shifts in the VRTs, from what I see the fonts for field-label should be:
* Small: `--spectrum-font-size-75` (`12px` for medium/desktop)
* Medium: `--spectrum-font-size-75` (`12px` for medium/desktop)
* Large: `--spectrum-font-size-100` (`14px` for medium/desktop)
* XL: `--spectrum-font-size-200` (`16px` for medium/desktop)

1st-gen/packages/combobox/src/combobox.css
```diff
@@ -14,28 +14,69 @@
...
+  -moz-osx-font-smoothing: auto;
   display: block;
+  grid-area: 1 / 1;
   width: 100%;
+
+  /* Remove wrapper padding so slotted sp-field-label (which has its own padding) isn’t double-spaced */
+  padding-block: 0;
```
rise-erpelding: This `padding-block: 0` does make sense in the case that `sp-field-label` is put into the label slot, but it doesn't work for `:host > label`.

I had thought that the mixin made it so that a `<span slot="field-label">` could be used in place of `<sp-field-label>` and that was why it was used in the Default and other stories, and using `<sp-field-label>` is no longer necessary at all?

For cases where the `<span>` is used, it gets `padding-block: 0` and doesn't receive any of the `top-to-text` and `bottom-to-text` tokens that `sp-field-label` and the field label mixin use, which I think is part of what's causing some regression in the VRTs?

1st-gen/packages/combobox/stories/combobox.stories.ts
```diff
@@ -130,24 +132,130 @@ export const noAutocomplete = (): TemplateResult => {
...
+        >
+          Sibling: sp-field-label outside combobox
```
rise-erpelding, line 163: Do we want to note that this isn't the ideal usage and that the preference would be to use the mixin because it's more accessible? Or would it be better to not show this pattern at all?

rise-erpelding (changes requested):
I left comments inline as well, but some bigger things I noticed:

* I'm kind of confused about what the ideal usage of field label mixin should be, my understanding had been that the component uses the mixin (and mixin styles) and then it doesn't need a slotted `<sp-field-label>` within it, when I did textfield, the labels don't use `<sp-field-label>` at all, they're just the default slot and that was how I thought it was supposed to work. It looks like combobox usage might be a bit different (it seems like it needs the `span` with the `slot="field-label"`, because the default slot is for options?). Bottom line, if we need to style for `sp-field-label` within `sp-combobox` _**and**_ a `label` within `sp-combobox`, that makes the styles a bit more complex, because then we need selectors for both.
* There are still some VRTs not passing for Combobox stories that need addressing. My recommendation would be to compare these to the corresponding stories in [prod](https://opensource.adobe.com/spectrum-web-components/storybook/index.html?path=/story/combobox--default&globals=system:spectrum-two), you can set up a side-by-side view in the browser as you're developing or toggle between local and prod to get a feel for changes. Running VRTs locally can be helpful too. (Cursor can help with this.)
* Would we want a changeset for this? I know there's already a changeset for combobox in the base branch, but there are changes in this branch that could be explained. Before we merge the feature to `main` we could probably combine all the changesets from all of the separate PRs.

1st-gen/packages/combobox/stories/combobox.stories.ts
aramos-adobe: I'm gonna remove the usage of `sp-field-label` because `<label>` is more accessible.

1st-gen/packages/combobox/src/combobox.css
aramos-adobe: Since `sp-field-label` is not being used I'll remove this CSS property

aramos-adobe:
@rise-erpelding

After reading your comments, I gained contexted that `<label>` appears to be the preferred slotted element as I looked through the other stories. For simplicity, I removed the usage of `<sp-field-label>` in the test stories created.

I ran some VRTs locally and 23 tests passed 🎉

I also created a changeset as a patch

1st-gen/packages/combobox/stories/combobox.stories.ts
rise-erpelding: I think the `side-aligned` attribute needs to be added here, that will help with the visual regression for this story.

```suggestion
      side-aligned="start"
      style="min-width: 80px;--spectrum-textfield-m-min-width:0; width:160px;"
```

That won't resolve the regression completely though, it looks something like this on my end when I do that:

Image

I would maybe remove the `style="min-width: 80px;--spectrum-textfield-m-min-width:0; width:160px;"` above and try adjusting some of the other properties instead? There's a `--mod` for columns now, `--mod-textfield-grid-template-columns-side-label`, that might be all that needs adjusting to get it to pass?

(There's also nothing stopping you from just resetting `grid-template-columns` as a style here or on the `noAutocomplete` story, I just personally dislike that idea because the custom properties were intended to be used here to modify styles.)

1st-gen/packages/combobox/stories/combobox.stories.ts
```diff
@@ -130,24 +132,119 @@ export const noAutocomplete = (): TemplateResult => {
     <sp-combobox
       .options=${fruits}
       side-aligned="start"
-      style="min-width: 80px;--spectrum-textfield-m-min-width:0;width:100px;"
+      style="min-width: 80px;--spectrum-textfield-m-min-width:0;"
```
rise-erpelding: Wondering if `--spectrum-textfield-m-min-width:0;` was a typo that was added to these stories at some point? I'm not seeing it do anything.

But I'm seeing a visual regression on this story too. I think we could use `--mod-textfield-grid-template-columns-side-label` to try to adjust column sizing here.

1st-gen/packages/combobox/src/combobox.css
```diff
@@ -14,28 +14,58 @@
...
+  grid-template-rows: auto auto auto;
+  grid-template-columns: 1fr;
 }
 
-:host([label-position="inline-start"]) {
-  flex-wrap: nowrap;
+:host([side-aligned="start"]),
+:host([side-aligned="end"]) {
+  grid-template-rows: auto auto;
+  grid-template-columns: minmax(10ch, 1fr) minmax(16ch, 2fr);
```
rise-erpelding: Do side labels still work if we remove these lines? We have the same CSS added in textfield, so it looks like these styles are being applied twice:

Image

If we remove this one, we can redefine the custom prop (`--mod-textfield-grid-template-columns-side-label`) for the columns in stories to adjust some of the layout issues in individual stories like "No Autocomplete" and "Has Disabled Items" (these are the ones I also have comments on in the stories file).

(I feel gross recommending that you use a mod since we're getting rid of them in 2nd-gen, but it's also the way that things were supposed to work in 1st-gen, that's the line of thinking in my recommendation.)

1st-gen/packages/combobox/src/spectrum-combobox.css
```diff
@@ -165,9 +165,12 @@
   :host {
     --highcontrast-combobox-border-color-highlight: Highlight;
     --highcontrast-combobox-border-color-invalid: Highlight;
+    --mod-picker-button-border-color-disabled: GrayText;
+    --mod-picker-button-icon-color-disabled: GrayText;
```
rise-erpelding, line 168: This is a good fix to make the picker button disabled, looks like picker button doesn't have any `highcontrast` custom props for us to hook into. ✨

1st-gen/packages/combobox/src/spectrum-combobox.css
```diff
@@ -165,9 +165,12 @@
...
   }
 
   .button.spectrum-PickerButton--quiet .spectrum-PickerButton-fill {
+    border-color: transparent;
```
rise-erpelding: Do we need this `border-color: transparent` style? I might be missing something but I'm not 100% sure that the selector `.spectrum-PickerButton--quiet` exists in SWC, it might be an artifact left over from CSS packages that wasn't transformed properly?

1st-gen/packages/combobox/src/spectrum-combobox.css
```diff
@@ -275,7 +278,17 @@
...
 
 #textfield {
+  position: relative;
+  grid-area: 2 / 1;
   inline-size: 100%;
+  block-size: var(--mod-combobox-block-size, var(--spectrum-combobox-block-size));
+}
+
+.button,
+.progress-circle {
```
rise-erpelding, line 281: The button looks good but I'm not seeing the progress circle in the Pending story, it looks like it's missing in the VRTs as well.

1st-gen/packages/combobox/src/spectrum-combobox.css
rise-erpelding: I think this `--spectrum-combobox-spacing-label-to-combobox` is the custom prop that was responsible for setting the negative margin on the quiet combobox between the label and the combobox part of the component (the `margin-block-start` on the `sp-combobox` `:host`), and I think that's at least part of why there's a spacing [regression for the Quiet story](https://swcpreviews.z13.web.core.windows.net/pr-6047/spectrum-light-medium-ltr/review/#ComboboxStories/quiet.png). We might want this negative margin to be set on `#textfield` now that the label is slotted.

.changeset/true-tips-attend.md
```diff
@@ -0,0 +1,11 @@
+---
+'@spectrum-web-components/combobox': patch
+---
+
+**Fixed**: Align combobox slotted field label with standalone label and improve accessibility:
+
+- Slotted native `<label>` now receives the same typography and spacing as standalone `sp-field-label` via `::slotted(label)` styling (preferred, most accessible pattern)
+- Slotted `sp-field-label` and native label font sizes now match standalone; removed combobox overrides so the field-label mixin size scale applies consistently
+- Switched combobox layout from flex to CSS Grid so the picker button stays aligned with the input row when a label is present
+- Disabled combobox in forced-colors mode now uses `GrayText` for the label and picker button
```
rise-erpelding, line 10: The disabled combobox now also has a label with disabled coloring thanks to your changes, that's worth calling out here unless you did and I missed it.

Image

rise-erpelding (commented):
Thanks for doing some code cleanup! The changes are much easier to reason with now and I did a deeper dive than last time.

Leaving some notes mostly for my future self about the VRTs:

I can see there are some diffs in the VRTs we're going to have to accept here, like:
- All `ComboboxSizesStories` (since now we're sizing fonts/spacing correctly)
- `ComboboxStories` `disabled`
- `ComboboxStories` `invalid` (I can't really figure out why there's a diff)
- `ComboboxStories` `withFieldLabel` (only because the label text changed, this is fine!)

There are others that I would argue need addressing because they don't match, but we can "cheat" and just adjust inline styles within the story to prove that we can match what we had before.
- `ComboboxStories` `lightDOM`
- `ComboboxStories` `noAutocomplete`

And then some others that need some updates to the CSS files (more of a true regression):
- `ComboboxStories` `pending` (needs progress circle)
- `ComboboxStories` `quiet` (needs spacing adjustment)

1st-gen/packages/combobox/src/spectrum-combobox.css
aramos-adobe: Had to remove an unnecessary `position:relative` here

aramos-adobe:
@rise-erpelding, the `label` line-height in the grid layout visually adjusts the negative space between the combobox.

1st-gen/packages/combobox/src/spectrum-combobox.css
```diff
@@ -138,23 +138,22 @@
...
 
-:host([quiet][size="s"]) {
-  --spectrum-combobox-spacing-label-to-combobox: var(--spectrum-field-label-to-component-quiet-small);
+  /* Flush under label: negative of label's padding-block-end (field-label-bottom-to-text per size) so field sits tight under label. */
+  --spectrum-combobox-spacing-label-to-combobox: calc(var(--swc-scale-factor) * -6px);
 }
 
-:host([quiet]) {
-  --spectrum-combobox-spacing-label-to-combobox: var(--spectrum-field-label-to-component-quiet-medium);
+/* Per-size overrides so quiet matches production when token resolution differs (e.g. medium 5.5px, small 2px from design). */
```
rise-erpelding: Just a nit, but could we adjust this comment so it doesn't reference production? Since eventually this will be production.

Maybe something like this, but adjust as you like:

```suggestion
/* Per-size overrides so quiet matches the standalone field label spacing when token resolution differs (e.g. medium 5.5px, small 2px from design). */
```

PR #6048, opened 2026-02-24
fix(color-area): correct value setting for 'v' and optimize ResizeObserver callback

blunteshwar:
## Description

This PR fixes two issues in the `ColorArea` component:

1. **Incorrect brightness (`v`) value calculation**: The `v` channel was being set with `(1 - this.inputY.valueAsNumber) * 100`, which inverted the brightness. It is now correctly set as `this.inputY.valueAsNumber * 100`.

2. **ResizeObserver callback optimization**: Instead of iterating over all `entries` and assigning `entry.contentRect` as the bounding rect, the callback now calls `this.getBoundingClientRect()` directly. This ensures the most up-to-date bounding rect is used regardless of how many resize entries are reported.

## Motivation and context

The incorrect `v` value inversion caused the color area to set brightness in reverse — moving the handle down (lower Y) would increase brightness instead of decreasing it. The ResizeObserver change simplifies the logic and avoids stale rect data from the entries array.

## Related issue(s)

- fixes SWC-1547

---

## Author's checklist

---

## Reviewer's checklist

### Manual review test cases

Use the [StackBlitz playground](https://stackblitz.com/edit/vitejs-vite-ifeketdw?file=src%2Fmy-element.ts) to reproduce and verify both fixes interactively.

- [ ] Verify brightness control works correctly in the color area
  1. Open the [StackBlitz playground](https://stackblitz.com/edit/vitejs-vite-ifeketdw?file=src%2Fmy-element.ts)
  2. Drag the handle vertically — moving up should increase brightness, moving down should decrease it
  3. Confirm the selected color value updates correctly (check the displayed HEX/HSV output)

- [ ] Verify the color area resizes correctly
  1. Open the [StackBlitz playground](https://stackblitz.com/edit/vitejs-vite-ifeketdw?file=src%2Fmy-element.ts)
  2. Resize the preview pane to trigger a resize event on `<sp-color-area>`
  3. Drag the handle to select a color at the edges and expect the bounding rect to reflect the new size accurately

### Device review

- [ ] Did it pass in Desktop?

  1. Open the [StackBlitz playground](https://stackblitz.com/edit/vitejs-vite-ifeketdw?file=src%2Fmy-element.ts)
  2. Tab to focus the component
  3. Use arrow keys to navigate — expect brightness and saturation to update correctly

  1. Open the [StackBlitz playground](https://stackblitz.com/edit/vitejs-vite-ifeketdw?file=src%2Fmy-element.ts) with a screen reader active
  2. Focus the component and move the handle using arrow keys
  3. Expect correct color value announcements with no erroneous or duplicate readouts

### DIFF
--- a/.changeset/color-area-fix-value-resize.md
+++ b/.changeset/color-area-fix-value-resize.md
@@ -1,3 +1,5 @@
-## '@spectrum-web-components/color-area': patch
+---
+'@spectrum-web-components/color-area': patch
+---
 
 **Fixed**: Corrected an issue where the brightness axis in `ColorArea` was inverted on the Y-axis, causing the handle position to not match the expected brightness value. Dragging the handle now correctly maps to the HSV brightness scale.


blunteshwar:
### Why this was only visible on StackBlitz, not Storybook

**tl;dr**: The brightness inversion in `updated()` is guarded by a floating-point drift check. Storybook never exercises this path; StackBlitz does.

---

#### The guard in `updated()`

```ts
if (this.y !== this.inputY.valueAsNumber) {
    // before fix: (1 - this.inputY.valueAsNumber) * 100  ← inverts brightness
    this.colorController.color.set('v', this.inputY.valueAsNumber * 100);
}
```

This block is a drift-correction step. `this.y` reads back from colorjs as `color.hsv.v / 100`. If colorjs internally converts through multiple color spaces (e.g., stores as RGB and converts back to HSV), the round-trip can introduce sub-precision float differences — e.g. `this.y = 0.36999…` while `inputY.valueAsNumber = 0.37`. Only when they diverge does the guard fire.

---

#### Why Storybook was unaffected

Storybook stories control the component via `args` — clean values like `0`, `0.5`, `1.0` that round-trip through colorjs without any float drift. With `this.y === this.inputY.valueAsNumber` always true, the guard block never executes and the buggy formula is effectively dead code in that context.

#### Why StackBlitz exposed it

Real pointer interaction produces arbitrary decimal values (e.g., `0.37143…`) computed directly from raw `clientY` coordinates. These are more likely to cause float drift through colorjs's internal representation. When the drift condition fires, the old `(1 - value)` formula fully inverts the brightness — visible as the handle producing the wrong colour on drag.

---

Added two regression tests in `color-area.test.ts` (`caa00d99d4`) to cover both paths:

1. **Pointer drag direction** — simulates a drag from near-bottom to near-top and asserts `y` increases (brightness increases when moving up), catching any re-introduction of the coordinate inversion.
2. **Programmatic set across update cycles** — sets `el.y` to non-round floats (`0.37`, `0.72`), waits for multiple update cycles to let any drift-triggered correction settle, and asserts the value has not been flipped to `1 - y`.

.changeset/color-area-fix-value-resize.md
```diff
@@ -0,0 +1,5 @@
+---
+'@spectrum-web-components/color-area': patch
+---
+
+**Fixed**: Corrected the brightness (`v`) value calculation in `ColorArea` — removed the erroneous `(1 - value)` inversion so that dragging the handle now maps correctly to the HSV brightness axis. Also optimized the `ResizeObserver` callback to call `getBoundingClientRect()` directly instead of iterating over entries, ensuring the component always uses an accurate bounding rect after resize. [#6048](https://github.com/adobe/spectrum-web-components/pull/6048)
```
Rajdeepc:
```suggestion
**Fixed**: Corrected brightness (v) mapping in `ColorArea` so the drag handle now aligns properly with the HSV brightness axis. Also improved resize handling to ensure accurate layout calculations after component resize. (#6048) [#6048](https://github.com/adobe/spectrum-web-components/pull/6048)
```

1st-gen/packages/color-area/test/color-area.test.ts
```diff
@@ -534,6 +534,141 @@ describe('ColorArea', () => {
...
+        clientY: clientYBottom,
+        bubbles: true,
+        composed: true,
+        cancelable: true,
+      })
+    );
+    await elementUpdated(el);
+
+    const yNearBottom = el.y;
+
```
Rajdeepc, line 596: can you add an additional sanity assertion here. This is not required but just making sure it guards against a regression where both values are equal but increasing due to a math mistake.

```ts
expect(yNearBottom, 'y near bottom should be relatively low').to.be.lessThan(0.6);

```

1st-gen/packages/color-area/test/color-area.test.ts
```diff
@@ -534,6 +534,141 @@ describe('ColorArea', () => {
...
+        clientY: clientYTop,
+        bubbles: true,
+        composed: true,
+        cancelable: true,
+      })
+    );
+    await elementUpdated(el);
+
+    const yNearTop = el.y;
+
```
Rajdeepc, line 619: Same like above
```ts
expect(yNearTop, 'y near top should be relatively high').to.be.greaterThan(0.4);

```

Rajdeepc (approved):
Approving with some small nits!

blunteshwar:
### Why we need `getContentBoundingRect()` — explained from first principles

---

#### What is `getBoundingClientRect()`?

Imagine your browser window is a TV screen. Every element on the page has a position on that screen. `getBoundingClientRect()` says:

> "Give me a rectangle describing where this element sits **on the TV screen**, including its border."

So for a color area sitting 100px from the left edge of the screen, with a 1px border and 200px total width:

```
Screen:  |←─ 100px ─→|← border(1px) →|←── content (198px) ──→|← border(1px) →|
                      ↑
              rect.left = 100     rect.width = 200
```

`rect.left = 100` → left edge of the **border**
`rect.width = 200` → **includes** both 1px borders

---

#### What is `entry.contentRect`?

When the browser fires a `ResizeObserver` event, it also tells you the size of the element's **inner content area** (no borders, no padding). But — and this is the key — it gives you that size **relative to the element itself**, not to the screen.

Think of it like measuring a room from the inside. You'd say "the room is 198px wide and I'm 0px from the left wall." But you have no idea where the room is in the building.

```
contentRect.left  = 0    ← always 0 (no padding in this component)
contentRect.width = 198  ← content only, no borders
```

`contentRect` gives the correct **size** but its `left`/`top` are room-relative, not screen-relative.

---

#### Why this matters — the pointer math

When you drag your mouse to pick a colour, the browser gives you `event.clientX` — the mouse's position **on the TV screen**.

To find where inside the gradient the mouse is, the code does:

```
percentX = (mouseX - element's left edge on screen) / element's width
```

This needs **both values in the same coordinate system** — screen coordinates.

---

#### What broke with `entry.contentRect`

```
mouseX      = 250   (screen position)
rect.left   = 0     ← from contentRect (room-relative, NOT the screen position!)
rect.width  = 198

percentX = (250 - 0) / 198 = 1.26  → clamped to 1.0
```

The element is actually at `left = 100` on screen, so the correct answer is `(250 - 100) / 198 = 0.76`. Instead we get `1.0`. The handle snaps to the rightmost edge regardless of where you click. **Wrong.**

---

#### What the first fix did with `getBoundingClientRect()`

```
mouseX      = 250   (screen position)
rect.left   = 100   ← correct screen position ✓
rect.width  = 200   ← includes 1px borders on both sides ✗

percentX = (250 - 100) / 200 = 0.75
```

Position is now correct. But `width = 200` includes the borders. This means:

- The handle thinks the component spans pixel 100→300 (200px total)
- But the **visible gradient** only spans pixel 101→299 (198px)
- Dragging all the way to the far-right edge of the gradient gives `(299 - 100) / 200 = 0.995` — not exactly `1.0`
- The handle's centre never quite reaches the corner of the gradient

---

#### Why we need our own function

Neither built-in option gives us what we actually need:

| | `left` / `top` | `width` / `height` |
|---|---|---|
| `getBoundingClientRect()` | ✅ screen position | ❌ includes border |
| `entry.contentRect` | ❌ room-relative (always ~0) | ✅ excludes border |

We need to **combine** the two pieces of information:
- **Screen position** from `getBoundingClientRect()`
- **Border thickness** from `getComputedStyle()` to step past the border to the content edge
- **Content dimensions** (width/height minus borders) for accurate percentage calculations

```
Screen:  |←─ 100px ─→|← 1px →|←── 198px gradient ──→|← 1px →|
                      ↑        ↑
          getBCR().left=100    what we actually want: left=101, width=198
```

Now dragging to the far-right edge of the gradient (`mouseX = 299`):

```
percentX = (299 - 101) / 198 = 1.0  ✓  handle centre exactly at the corner
```

.changeset/color-area-fix-value-resize.md
```diff
@@ -0,0 +1,5 @@
+---
+'@spectrum-web-components/color-area': patch
+---
+
+**Fixed**: Corrected the brightness (`v`) value calculation in `ColorArea` — removed the erroneous `(1 - value)` inversion so dragging the handle maps correctly to the HSV brightness axis. Introduced `getContentBoundingRect()` to compute the content-box rect in viewport coordinates (excluding the element's border), used consistently in `handlePointerdown`, `firstUpdated`, and the `ResizeObserver` callback. This ensures pointer-to-handle mapping is geometrically accurate and the handle centre correctly superimposes with the gradient corners at the extremes. [#6048](https://github.com/adobe/spectrum-web-components/pull/6048)
```
rubencarvalho: nit: let’s remember that changesets surface as customer-facing changelog entries. Too much internal detail can distract from the core message about what changed and what users should care about.

The current description focuses more on the implementation than on the impact. I’d suggest framing it around the user-visible behavior instead.

Example:
```
'@spectrum-web-components/color-area': patch
---

**Fixed**: Corrected an issue where the brightness axis in `ColorArea` was inverted on the Y-axis, causing the handle position to not match the expected brightness value. Dragging the handle now correctly maps to the HSV brightness scale.```

PR #6067, opened 2026-03-06
docs: Add 2nd-gen TypeScript style guide and Storybook docs automation

caseyisonit:
## Description

Adds a comprehensive TypeScript style guide for 2nd-gen component development and automates publishing contributor docs to Storybook.

## Motivation and context

Contributors (human and AI) need clear, consistent guidance for writing 2nd-gen components. This PR provides detailed conventions for file organization, class structure, decorators, types, and composition patterns—all based on the Badge component as the reference implementation.

Additionally, contributor docs are now automatically converted and displayed within Storybook, making them easier to discover and browse alongside component documentation.

> NOTE: The generate contributor docs script is TEMPORARY and will be re-evaluated when we address the documentation information architecture. For the time being this will allow us to review and read contributor docs in storybook 

## Related issue(s)

-   fixes [SWC-1420]

## Screenshots (if appropriate)

_N/A - documentation changes_

## Author's checklist

-   [x] I have read the **[CONTRIBUTING](<(https://github.com/adobe/spectrum-web-components/blob/main/CONTRIBUTING.md)>)** and **[PULL_REQUESTS](<(https://github.com/adobe/spectrum-web-components/blob/main/PULL_REQUESTS.md)>)** documents.
-   [x] I have reviewed at the Accessibility Practices for this feature, see: [Aria Practices](https://www.w3.org/TR/wai-aria-practices/)
-   [x] I have added automated tests to cover my changes.
-   [x] I have included a well-written changeset if my change needs to be published.
-   [x] I have included updated documentation if my change required it.

## Reviewer's checklist

- [x] Includes a Github Issue with appropriate flag or Jira ticket number without a link
- [ ] Includes thoughtfully written changeset if changes suggested include `patch`, `minor`, or `major` features
- [ ] Automated tests cover all use cases and follow best practices for writing
- [ ] Validated on all supported browsers
- [ ] All VRTs are approved before the author can update Golden Hash

### Manual review test cases

- [x] Verify TypeScript guide renders correctly in Storybook

    1. Run `yarn storybook` in `2nd-gen/packages/swc`
    2. Navigate to the Contributor Docs section
    3. Confirm the TypeScript style guide pages load and display properly

- [ ] Verify navigation structure is accurate

    1. Open `CONTRIBUTOR-DOCS/02_style-guide/02_typescript/README.md`
    2. Confirm all 15 linked guides exist and have correct content

### Device review

- [ ] Did it pass in Desktop?
- [ ] Did it pass in (emulated) Mobile?
[...]

PR #6093, opened 2026-03-19
fix(test): skip flaky overlay-v1 context menu suite and fix picker responsive test timing

Rajdeepc:
## Summary

- **Overlay v1**: Skip the flaky `contextmenu` test suite (`describe.skip`) since it tests the legacy v1 overlay API that will be replaced during the Spectrum 2 migration
- **Picker responsive**: Fix flaky timing by replacing `elementUpdated()` with explicit `requestUpdate()` + `updateComplete`, adding timeouts to `waitUntil` calls, and waiting for menu item role registration before click interactions

## Justification

These tests have been intermittently failing in CI:

1. **Overlay contextmenu** (`overlay-v1.test.ts`): The test suite for handling multiple separate `contextmenu` events is flaky in CI. Since this is 1st-gen v1 overlay API code that will not be carried forward to Spectrum 2, skipping is the pragmatic choice.

2. **Picker responsive** (`picker-responsive.test.ts`): Tests were flaky because directly setting `isTouchDevice.matches = true` bypasses the `MatchMediaController`'s `onChange` callback, which normally triggers `requestUpdate()`. Without an explicit `requestUpdate()` call, the component never re-renders and the test races against unresolved state. Additionally, menu item registration (role + selectionRoot) was not awaited before clicking, causing intermittent failures.

## Related

Split from #6067

## Test plan

- [ ] CI passes with overlay contextmenu tests skipped
- [ ] Picker responsive tests pass reliably across multiple CI runs

### DIFF
--- a/1st-gen/packages/overlay/test/overlay-v1.test.ts
+++ b/1st-gen/packages/overlay/test/overlay-v1.test.ts
@@ -452,7 +452,8 @@ describe('Overlays, v1', () => {
   });
 });
 describe('Overlay - type="modal", v1', () => {
-  describe('handle multiple separate `contextmenu` events', async () => {
+  // @TODO: skipping this test suite because it's flaky in CI. Will review in the migration to Spectrum 2.
+  describe.skip('handle multiple separate `contextmenu` events', async () => {
     let width = 0;
     let height = 0;
     let firstMenu: Popover;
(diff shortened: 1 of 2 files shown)


Rajdeepc:
Closing this PR — the flaky test fixes are not needed at this time.

PR #6096, opened 2026-03-19
refactor(badge,statuslight,sized-mixin): reorganize types and remove S2 suffix from exports

Rajdeepc:
## Summary

- **Badge types**: Reorganize into SHARED/S1-ONLY/CANONICAL sections. Remove `S2` suffix from canonical exports (`BADGE_VARIANTS_S2` -> `BADGE_VARIANTS`, `BADGE_VARIANTS_COLOR_S2` -> `BADGE_VARIANTS_COLOR`, etc.) since 2nd-gen is the forward-looking default
- **Sized mixin**: Replace `ElementSizes` record with `ELEMENT_SIZES` and `DEFAULT_ELEMENT_SIZES` const arrays, enabling `as const satisfies` type-safe patterns. Old `ElementSizes` re-exported from `@spectrum-web-components/base` as deprecated alias
- **Code quality**: Asset uses `when` directive + `swc-` CSS class prefix; Divider uses styles getter; ProgressCircle uses consistent `styles` naming; StatusLight gets `@todo` annotations

## Motivation

The `S2` suffix in type exports (`BADGE_VARIANTS_S2`, `BadgeVariantS2`, etc.) creates a naming dependency that will require a breaking change to remove later. By making the canonical (unsuffixed) names the 2nd-gen default now, we avoid that future breaking change. S1-specific subsets remain with the `_S1` suffix and can be cleanly deleted when 1st-gen is retired.

The sized mixin refactoring from `Record<string, ElementSize>` to `as const` arrays enables better IDE autocomplete and compile-time validation throughout the codebase.

## Related

- Split from #6067

## Test plan

- [ ] Badge stories render correctly with renamed exports
- [ ] Badge tests pass with updated type references
- [ ] 1st-gen components still compile using deprecated `ElementSizes` alias
- [ ] Asset renders file/folder/slot variants correctly with `when` directive
- [ ] Divider and ProgressCircle render without regressions

### DIFF
--- a/.changeset/refactor-badge-statuslight-types.md
+++ b/.changeset/refactor-badge-statuslight-types.md
@@ -1,7 +1,7 @@
 ---
-"@spectrum-web-components/core": minor
-"@adobe/spectrum-wc": patch
-"@spectrum-web-components/base": patch
+'@spectrum-web-components/core': minor
+'@adobe/spectrum-wc': patch
+'@spectrum-web-components/base': patch
 ---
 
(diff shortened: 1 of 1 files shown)


caseyisonit:
closing this in favor of the original PR as feedback was not addressed before splitting

PR #6100, opened 2026-03-24
chore(docs): create custom API tables

rubencarvalho:
## Description

Replaces Storybook's built-in `<Controls />` / `<ArgTypes />` doc blocks with a custom `<ApiTable />` component that reads directly from the Custom Elements Manifest. The API section in the docs page now shows clean, read-only tables organized by category: **Properties**, **Slots**, **Events**, **CSS Custom Properties**, and **CSS Parts**.

### What changed

**Custom API table (`ApiTable.tsx`)**
- New block component that renders categorized tables from the CEM
- Properties table includes dedicated **Attribute** column with `(reflects)` annotation — no more confusing split between "attributes" and "properties" categories
- Type column expands type aliases (e.g., `BadgeVariant`) to their actual values (e.g., `'accent' | 'informative' | ...`) when `options` are set in argTypes
- Description column has `min-width: 300px` to prevent squishing
- Tables only render when the category has entries (no empty sections)

**Doc template (`DocumentTemplate.mdx`)**
- Replaced `<Controls />` with `<ApiTable />`
- Removed `<Primary />` from the API section (no longer needed without controls)

**Story cleanup**
- Updated badge, avatar, icon, progress-circle, asset, divider, and status-light stories

## Motivation and context

The previous API table had several issues:
- Properties and attributes were split into separate categories, which was misleading — every attribute is also a property
- Type aliases like `BadgeVariant` weren't expanded, giving consumers no visibility into valid values
- Controls were shown in the API table but weren't useful for a reference doc — the Playground already has interactive controls

The custom `ApiTable` reads from the CEM directly, giving full control over presentation without fighting Storybook's argTypes system.

## Screenshots
Screenshot 2026-04-14 at 14 31 14

---

## Author's checklist

---

### Manual review test cases

- [ ] Verify the API table shows categorized sections (Properties, Slots, Events, etc.)
  1. Navigate to any component docs (e.g., Badge, Avatar, Progress Circle)
  2. Scroll to the API section
  3. Expect: Properties table with Property, Attribute, Type, Default, Description columns
  4. Expect: Attribute column shows kebab-case name with `(reflects)` where applicable

- [ ] Verify type expansion works
  1. Open Badge docs → API → Properties table
  2. Check the `variant` row's Type column
  3. Expect: expanded values like `'accent' | 'informative' | ...` instead of `BadgeVariant`

- [ ] Verify Playground controls still work
  1. Navigate to any component's Playground story
  2. Expect: interactive controls panel is unchanged


2nd-gen/packages/swc/.storybook/helpers/get-story-helpers.ts
```diff
@@ -0,0 +1,103 @@
...
+ * Wraps `getStorybookHelpers` to merge the "attributes" category into
+ * "properties" and annotate each property's description with its
+ * reflected HTML attribute name when applicable.
+ */
+export function getStoryHelpers<T>(tagName: string) {
+  const helpers = getStorybookHelpers<T>(tagName);
+  const { argTypes } = helpers;
+
+  const cem: Cem | undefined = (window as unknown as Record<string, unknown>)
+    .__STORYBOOK_CUSTOM_ELEMENTS_MANIFEST__ as Cem | undefined;
```
Rajdeepc: Do `@wc-toolkit/storybook-helpers` exports a way to get the CEM, if so can we leverage that instead? This is a double casting. Would you rather declaring it via `declare global` instead of the double-cast?

scripts/lib/markdown.js
rubencarvalho, line 1: I needed to re-add this file because it seems missing from `main`. The 2nd-gen deployment preview was not working without it.

rubencarvalho: (it is not related to this PR, hopefully gets fixed in `main` and removed from here)

2nd-gen/packages/swc/.storybook/helpers/get-story-helpers.ts
```diff
@@ -0,0 +1,103 @@
...
+  members?: CemMember[];
+}
+
+interface CemModule {
+  declarations?: CemDeclaration[];
+}
+
+interface Cem {
+  modules?: CemModule[];
+}
```
Rajdeepc: Are these not available in `@custom-elements-manifest/analyzer`. This is not scalable, if we evolve CEM schema these local types won't catch the mismatch.

2nd-gen/packages/swc/.storybook/helpers/get-story-helpers.ts
```diff
@@ -0,0 +1,103 @@
...
+  tagName: string
+): CemDeclaration | undefined {
+  for (const mod of cem.modules || []) {
+    for (const decl of mod.declarations || []) {
+      if (decl.tagName === tagName) {
+        return decl;
+      }
+    }
+  }
+  return undefined;
```
Rajdeepc: For the time being there will be no performance issue here but a Map built once at module scope would have been cleaner if the CEM is reused across many components.

2nd-gen/packages/swc/.storybook/helpers/get-story-helpers.ts
```diff
@@ -0,0 +1,103 @@
...
+}
+
+/**
+ * Wraps `getStorybookHelpers` to merge the "attributes" category into
+ * "properties" and annotate each property's description with its
+ * reflected HTML attribute name when applicable.
+ */
+export function getStoryHelpers<T>(tagName: string) {
```
Rajdeepc: This function will be called by every component's stories a regression here would silently break the entire docs API table. Would love to have a test file written here. Not a blocker for now but it would be a good addition.

Rajdeepc (changes requested):
Thanks for doing this. Some minor nits
**Blocking**: Two contributor docs files still reference `getStorybookHelpers` directly: Kindly change this to `getStoryHelpers`
1. 07_add-stories-for-2nd-gen-component.md (line 15)
2. 01_washing-machine-workflow.md (line 455)

2nd-gen/packages/swc/.storybook/helpers/get-story-helpers.ts
rubencarvalho: Done!

2nd-gen/packages/swc/.storybook/helpers/get-story-helpers.ts
rubencarvalho: I agree - what about we defer this for the docs website smoke tests once we get those up and running? :)

caseyisonit (dismissed):
I do not agree with this direction; attributes and properties are different by nature and should be documented as such. The properties and methods are no longer populating the table; only attributes and slots are present. I don't think this should be a priority for us at this time. The API table we get from Storybook helpers is accurate and I dont see the value in this level of customization.

2nd-gen/packages/swc/.storybook/helpers/get-story-helpers.ts
Rajdeepc: Yes

rubencarvalho:
@caseyisonit Maybe we’re crossing wires a bit here 😛
JavaScript properties _are_ the canonical API of our components. Attributes are the declarative HTML interface for setting those properties.

> attributes and properties are different by nature and should be documented as such. 

They still are. We’re surfacing the attributes (for the properties that expose them) in the Description column so the mapping is visible. This also aligns better with our current documentation site (and others, e.g. shoelace). Ideally we would have a dedicated column for attribute ↔ property mapping, but Storybook currently doesn’t support that. Happy to explore alternative visual indicators if we think that would make this clearer.

> The properties and methods are no longer populating the table; only attributes and slots are present

Where do you see this? I can see everything populating the table, including props, attributes, methods, css properties,... LMK how I can reproduce!

Screenshot 2026-04-07 at 09 25 42

> I don't think this should be a priority for us at this time

Unfortunately this is a P0 😞 Whatever appears in the API table effectively becomes our public contract, so we need to ensure it accurately reflects what we support from day one. Any future removal/omission or change from that table is considered a breaking change. In this case we already know some internal properties/methods are leaking, so we must address that.

> The API table we get from Storybook helpers is accurate and I dont see the value in this level of customization.

Maybe accurate, but I find it currently very confusing for consumers. For example, before this change `static-color` only appeared under attributes, while consumers are currently interacting with it in JS as `staticColor` (valid and supported by us!). That mapping wasn’t visible anywhere, which makes the API harder to understand IMO.

Rajdeepc (approved):
Approving since this works and the table is well populated with our first class API mappings.
Follow-up: Track unit test coverage for getStoryHelpers.

caseyisonit:
@rubencarvalho let me know when you need me to re-review. 

We met last week and have reach alignment on the necessity and intention of this PR. My initial concerns are resolved with some minor content improvements based on our sync.

2nd-gen/packages/core/element/spectrum-element.ts
rubencarvalho, line 1: Marking these as `@internal` so they do not propagate to the API table.

2nd-gen/packages/swc/.storybook/blocks/ApiTable.tsx
rubencarvalho, line 1: This creates a Storybook-themed API table for each of our API sections (properties, slots, events, CSS props, CSS parts). We can replace this with our own `swc-table` once it is migrated.

2nd-gen/packages/swc/components/avatar/stories/avatar.stories.ts
rubencarvalho, line 1: All of these annotations were not needed.

2nd-gen/packages/swc/.storybook/preview-head.html
rubencarvalho, line 1: These styles were leaking into the Storybook controls in the playground.

TarunAdobe (approved):

PR #6129, opened 2026-04-01
feat(core): Add focus management primitives for 2nd-gen architecture

caseyisonit:
## Description

Implements the core focus management infrastructure for 2nd-gen Spectrum Web Components, as defined in the Focus Management Proposal. This PR delivers the foundational primitives that all component migrations will build on.

### Core Utilities

- **`get-active-element.ts`** — Shadow DOM-aware `activeElement` traversal that follows `shadowRoot.activeElement` chains to find the deepest focused element
- **`focusable-selectors.ts`** — Standard HTML `focusableSelector` and `tabbableSelector` strings (removes 1st-gen `[focusable]` attribute workaround in favor of native `delegatesFocus`)
- **`hasVisibleFocusInTree()` cleanup** — Refactored from ~30 lines to 3 in `spectrum-element.ts`, removing dead ancestor-chain traversal code and legacy `.focus-visible` polyfill fallback. Now delegates to the new `getActiveElement()` utility.

### Controller

- **`FocusgroupNavigationController`** — Implements the roving `tabindex` pattern from the WAI-ARIA APG and directional navigation aligned with the proposed `focusgroup` HTML attribute (Open UI), with a polyfill until browsers ship native support.

  Key capabilities:
  - Keeps exactly one item in the tab order (`tabindex="0"`) per composite widget
  - Supports `horizontal`, `vertical`, `both`, and `grid` direction modes
  - Optional wrap, last-focused memory, skip-disabled, and `pageStep` (Page Up/Down navigation)
  - Grid mode uses bounding-rect layout for row/column navigation with Ctrl+Home/Ctrl+End support
  - Printable character typeahead via `focusFirstItemByTextPrefix()`
  - Dispatches a bubbling, composed `swc-focusgroup-navigation-active-change` custom event on active item changes
  - Flat single-file architecture (no base class inheritance from 1st-gen)
  - Listener scope defaults to `host.renderRoot` for shadow DOM encapsulation

  Supersedes the 1st-gen `FocusGroupController` + `RovingTabindexController` split with a single, consolidated controller.

### Mixin

- **`DisabledMixin`** — Standalone mixin extracted from 1st-gen `Focusable` base class. Adds a reactive `disabled` property with `aria-disabled`, tabindex management, and blur-on-disable. Uses `aria-disabled` on the host so elements remain discoverable by assistive technology.

### Documentation & Testing

- **Contributor guide** (`CONTRIBUTOR-DOCS/01_contributor-guides/13_focus-management.md`) — Comprehensive focus management guide for contributors
- **Controller composition style guide** — Updated `14_controller-composition.md` with focus controller patterns
- **Storybook stories** — Interactive demos for all `FocusgroupNavigationController` modes (horizontal, vertical, both, grid, wrap, memory, skipDisabled, pageStep, typeahead)
- **Unit tests** — Test suite for `FocusgroupNavigationController`
- **Controller documentation** — `focus-group-navigation-controller.md` with API reference
- **Storybook config** — Added storybook configuration for 2nd-gen controller stories

### File Structure

```
2nd-gen/packages/core/
├── controllers/
│   ├── focus-group-navigation-controller.ts              NEW (re-export entry)
│   ├── focus-group-navigation-controller/
│   │   ├── src/focus-group-navigation-controller.ts      NEW (implementation)
│   │   ├── stories/focus-group-navigation-controller.stories.ts  NEW
│   │   ├── test/focus-group-navigation-controller.test.ts        NEW
│   │   └── focus-group-navigation-controller.md          NEW
│   └── index.ts                                          UPDATED
├── mixins/
│   ├── disabled-mixin.ts                                 NEW
│   └── index.ts                                          UPDATED
├── utils/
│   ├── get-active-element.ts                             NEW
│   ├── focusable-selectors.ts                            NEW
│   └── index.ts                                          UPDATED
├── element/
│   └── spectrum-element.ts                               UPDATED (hasVisibleFocusInTree cleanup)
├── overview.mdx                                          UPDATED
└── package.json                                          UPDATED

CONTRIBUTOR-DOCS/
├── 01_contributor-guides/13_focus-management.md           NEW
└── 02_style-guide/02_typescript/14_controller-composition.md  UPDATED
```

### Motivation and Context

The 1st-gen focus architecture uses a deep inheritance chain (`Focusable` base class) that forces all-or-nothing inheritance, carries dead polyfill code, and uses a fragile flag system. The 2nd-gen strategy replaces this with composable, opt-in primitives: native `delegatesFocus: true`, `FocusgroupNavigationController`, and `DisabledMixin`.

Centralizing composite widget keyboard navigation on a single controller aligned with the `focusgroup` proposal reduces duplicated roving-tabindex logic across components, improves consistency, and shrinks maintenance surface while preserving WAI-ARIA APG accessibility expectations.

### Related Issues

- fixes SWC-1676

## Manual Testing

Run Storybook from `2nd-gen/packages/swc`.

### 1. Horizontal Toolbar
- [ ] Open the **"Focus group navigation controller / Horizontal Toolbar"** story
- [ ] **Tab** into the toolbar — only one button receives focus (roving tabindex)
- [ ] **ArrowRight** / **ArrowLeft** move focus between Bold, Italic, Underline, Strikethrough
- [ ] Focus **wraps** from Strikethrough → Bold and vice versa
- [ ] **Home** jumps to Bold, **End** jumps to Strikethrough
- [ ] **Tab** away and back — the last-focused button is remembered

### 2. Both Axes Linear
- [ ] Open **"Both Axes Linear"** story
- [ ] **ArrowRight** and **ArrowDown** both advance forward through items
- [ ] **ArrowLeft** and **ArrowUp** both go backward
- [ ] Wrapping is enabled in both directions

### 3. Vertical Menu
- [ ] Open **"Vertical Menu"** story
- [ ] **ArrowDown** / **ArrowUp** move through Copy, Paste, Cut (unavailable), Select all
- [ ] "Cut (unavailable)" has `aria-disabled="true"` but is **still focusable** (`skipDisabled: false`)
- [ ] Clicking or pressing Enter/Space on "Cut (unavailable)" does nothing
- [ ] **Page Down** / **Page Up** skip 2 items at a time (`pageStep: 2`)

### 4. Skip Disabled Menu
- [ ] Open **"Skip Disabled Menu"** story
- [ ] **ArrowDown** skips "Save" (native `disabled`) and "Close" (`aria-disabled="true"`)
- [ ] Only New, Open, Print, Help are reachable via arrow keys
- [ ] Wrap is enabled — Help → New and New → Help

### 5. Grid (3×3)
- [ ] Open **"Grid"** story
- [ ] **ArrowRight** / **ArrowLeft** move within a row
- [ ] **ArrowDown** / **ArrowUp** move between rows (same column)
- [ ] **Home** / **End** jump to first/last cell in row-major order
- [ ] **Ctrl+Home** / **Ctrl+End** jump to cell 1 / cell 9
- [ ] **Page Down** / **Page Up** move 2 rows at a time (`pageStep: 2`)
- [ ] No wrapping — arrows stop at edges

### 6. Programmatic Focus
- [ ] Open **"Programmatic Focus"** story
- [ ] Click "Focus item C programmatically" — focus moves to Item C in the toolbar
- [ ] Arrow keys work from the programmatically focused item

### 7. Text Prefix (Typeahead)
- [ ] Open **"Text Prefix Focus"** story
- [ ] Click **'Focus match for "Pas"'** — focus jumps to **Paste** in the menu
- [ ] Click **'Focus match for "cu"'** — focus jumps to **Cut** (case-insensitive)

### 8. General Checks
- [ ] `:focus-visible` outline appears on keyboard navigation (not on click)
- [ ] `tabindex="0"` is on exactly one item in each group, all others are `tabindex="-1"`
- [ ] Test in RTL (`dir="rtl"`) — ArrowRight should go backward in horizontal mode


cdransf (approved):

2nd-gen/packages/core/package.json
```diff
@@ -152,6 +156,9 @@
       "controllers/index.js": [
         "dist/controllers/index.d.ts"
       ],
+      "controllers/ocus-group-navigation-controller/focus-group-navigation-controller.js": [
```
Rajdeepc:
```suggestion
      "controllers/focus-group-navigation-controller/focus-group-navigation-controller.js": [
```

Rajdeepc (changes requested):
I have two blocking concerns before technical review:

1. No consuming component validates composition. DisabledMixin, delegatesFocus, and FocusgroupNavigationController are designed to work together, but no 2nd-gen component uses them yet. Before merging infrastructure, I need to see at least one component (e.g., a simple toolbar or action-button) that exercises all three primitives together to prove they compose correctly.

2. DisabledMixin is not a replacement for Focusable. The 1st-gen Focusable class provides `focus()`, `blur()`, `click()` overrides, `autofocus`, `focusElement` delegation, `selfManageFocusElement`, and manipulating `Tabindex`. `DisabledMixin` only handles `aria-disabled`, `tabindex` save/restore, and blur-on-disable. The migration cost for components moving from Focusable is significant and the documentation does not surface this adequately.

**Additional concerns for discussion:**

`FocusgroupNavigationController` is missing APIs that 1st-gen composite widgets depend on: `elementEnterAction` (Tabs, Accordion), `stopKeyEventPropagation` (nested composites), `manage/unmanage` (Menu search mode), `focusOnItem`, `reset`. If these are intentionally deferred, please document what's deferred vs. what's deliberately dropped.

Holding off on line-by-line technical review until the above are addressed.

2nd-gen/packages/core/controllers/focusgroup-navigation-controller/src/focusgroup-navigation-controller.ts
```diff
@@ -0,0 +1,1109 @@
+/**
```
nikkimk, line 1: This is deliberately lightweight, as we can decide which features should be added to this controller, to a new controller, or to the component itself as components demonstrate need.

Note that many components that use focus group involve a11y changes and will be set to `DelegatesFocus: true` so that some of the old may not need what they used to from the old controller.

nikkimk:
> I have two blocking concerns before technical review:
> 
> 1. No consuming component validates composition. DisabledMixin, delegatesFocus, and FocusgroupNavigationController are designed to work together, but no 2nd-gen component uses them yet. Before merging infrastructure, I need to see at least one component (e.g., a simple toolbar or action-button) that exercises all three primitives together to prove they compose correctly.
> 2. DisabledMixin is not a replacement for Focusable. The 1st-gen Focusable class provides `focus()`, `blur()`, `click()` overrides, `autofocus`, `focusElement` delegation, `selfManageFocusElement`, and manipulating `Tabindex`. `DisabledMixin` only handles `aria-disabled`, `tabindex` save/restore, and blur-on-disable. The migration cost for components moving from Focusable is significant and the documentation does not surface this adequately.
> 
> **Additional concerns for discussion:**
> 
> `FocusgroupNavigationController` is missing APIs that 1st-gen composite widgets depend on: `elementEnterAction` (Tabs, Accordion), `stopKeyEventPropagation` (nested composites), `manage/unmanage` (Menu search mode), `focusOnItem`, `reset`. If these are intentionally deferred, please document what's deferred vs. what's deliberately dropped.
> 
> Holding off on line-by-line technical review until the above are addressed.

@Rajdeepc see my comments `FocusgroupNavigationController` here: https://github.com/adobe/spectrum-web-components/pull/6129#discussion_r3040240755

caseyisonit:
> No consuming component validates composition. DisabledMixin, delegatesFocus, and FocusgroupNavigationController are designed to work together, but no 2nd-gen component uses them yet. Before merging infrastructure, I need to see at least one component (e.g., a simple toolbar or action-button) that exercises all three primitives together to prove they compose correctly.
> 
> DisabledMixin is not a replacement for Focusable. The 1st-gen Focusable class provides focus(), blur(), click() overrides, autofocus, focusElement delegation, selfManageFocusElement, and manipulating Tabindex. DisabledMixin only handles aria-disabled, tabindex save/restore, and blur-on-disable. The migration cost for components moving from Focusable is significant and the documentation does not surface this adequately.

- delegatesFocus is an implementation detail and strategy approach and not something to be validated until a component comes over that would follow that guidance. 
- DisabledMixin is NOT a replacement of Focusable. it simply isolates an a11y property that should be opt in and not included in every instance of focus. Its removing the coupling and making it more flexible for usage.
- Please reference the Slack canvas as all the information about 1st-gen considerations are included and called out, this is a PR that supports the RFC proposal that has been documented and shared out. we cant't link internaly docs to PR.
- The documentation included is comprehensive of reasoning and refactor guidance, please review the whole PR (ive added even more details to be extra clear)

Rajdeepc:
> > No consuming component validates composition. DisabledMixin, delegatesFocus, and FocusgroupNavigationController are designed to work together, but no 2nd-gen component uses them yet. Before merging infrastructure, I need to see at least one component (e.g., a simple toolbar or action-button) that exercises all three primitives together to prove they compose correctly.
> > DisabledMixin is not a replacement for Focusable. The 1st-gen Focusable class provides focus(), blur(), click() overrides, autofocus, focusElement delegation, selfManageFocusElement, and manipulating Tabindex. DisabledMixin only handles aria-disabled, tabindex save/restore, and blur-on-disable. The migration cost for components moving from Focusable is significant and the documentation does not surface this adequately.
> 
> * delegatesFocus is an implementation detail and strategy approach and not something to be validated until a component comes over that would follow that guidance.
> * DisabledMixin is NOT a replacement of Focusable. it simply isolates an a11y property that should be opt in and not included in every instance of focus. Its removing the coupling and making it more flexible for usage.
> * Please reference the Slack canvas as all the information about 1st-gen considerations are included and called out, this is a PR that supports the RFC proposal that has been documented and shared out. we cant't link internaly docs to PR.
> * The documentation included is comprehensive of reasoning and refactor guidance, please review the whole PR (ive added even more details to be extra clear)

Thanks for the explanation but I do have some follow up thoughts here.

1. I’m still a bit curious about isn’t testing delegatesFocus in isolation, but how all three primitives behave together in a real component scenario. Right now, the Storybook demos are great for exercising the controller on its own, but we don’t yet have an example that combines `DisabledMixin + delegatesFocus: true + FocusgroupNavigationController` on the same element the way a migrated component would. Even a small proof-of-concept (like a simple action-group or toolbar prototype) will be helpful in validating that there aren’t any subtle interaction issues between tabindex management, roving tabindex, and browser focus delegation. 

2. I am aligned on the scope of DisabledMixin. One small thing I noticed: the `click()` guard (`if (this.disabled) return;`) is mentioned in the “Common mistakes” section, but not in the migration section for replacing Focusable. Since aria-disabled doesn’t block click events, it might be helpful to call that out explicitly there as well so it’s harder to miss during migration.

3. I want to make sure the PR feels self-contained for all the reviewers. Related to that, I noticed a discrepancy between the original proposal shared in Slack (with RovingTabindexController) and the RFC in this PR (which introduces `FocusgroupNavigationController`). Both approaches seem valid, but they are meaningfully different (API shape, callbacks, etc.). If this reflects an evolution in direction, should we call this out to make life simpler for other reviewers.

4. The documentation overall is strong, but one area that could add even more clarity is explaining **why certain 1st-gen APIs were intentionally not carried forward**. I’m assuming many of these were deliberate decisions, but capturing that reasoning would really help align the team especially for folks migrating components who might otherwise expect those APIs to still exist.

Curious what you think — happy to chat through any of this if helpful!

caseyisonit:
@Rajdeepc I have migrated the updated RFC to the PR in project planning so the slack canvas is now out of date and i will clean that up in my tomorrow. 

> I’m still a bit curious about isn’t testing delegatesFocus in isolation, but how all three primitives behave together in a real component scenario. Right now, the Storybook demos are great for exercising the controller on its own, but we don’t yet have an example that combines DisabledMixin + delegatesFocus: true + FocusgroupNavigationController on the same element the way a migrated component would. Even a small proof-of-concept (like a simple action-group or toolbar prototype) will be helpful in validating that there aren’t any subtle interaction issues between tabindex management, roving tabindex, and browser focus delegation.

I think to honor the scope of this RFC proposal, this work would be a follow up in the next sprint, im thinking Tabs since it would need all three. To incorporate an entire component migration with this is out of scope and we should approach this agilely and iterate as we migrate more components. Accepting this with the agnostic component setups in the documentation covers the testing concern to be accepted in to main. This will also unblock several other components that didnt need all the complexity of 1st-gen now that we have separated the concerns. 
TLDR We will migrate Tabs in the next round of components once this is merged in

I have addressed the three other concerns in the documentation and RFC so those should now be covered! Let me know if i missed anything else. :)

2nd-gen/packages/swc/.storybook/blocks/GettingStarted.tsx
caseyisonit, line 1: All of the storybook docs changes were to allow the controller docs to work correctly. this will be re-evaluated in the docs IA epic for scalability but can be accepted with little to no risk.

Rajdeepc:
> I think to honor the scope of this RFC proposal, this work would be a follow up in the next sprint, im thinking Tabs since it would need all three. To incorporate an entire component migration with this is out of scope and we should approach this agilely and iterate as we migrate more components. Accepting this with the agnostic component setups in the documentation covers the testing concern to be accepted in to main. This will also unblock several other components that didnt need all the complexity of 1st-gen now that we have separated the concerns. TLDR We will migrate Tabs in the next round of components once this is merged in
> 
> I have addressed the three other concerns in the documentation and RFC so those should now be covered! Let me know if i missed anything else. :)

Fair point. I think you have addressed the documentation gaps and would really appreciate the Tabs follow-up being tracked as  a ticket so it doesn't slip between sprints.

2nd-gen/packages/core/controllers/focusgroup-navigation-controller/src/focusgroup-navigation-controller.ts
```diff
@@ -0,0 +1,1109 @@
...
+    const safeActive = items.includes(active) ? active : items[0];
+    for (const el of items) {
+      if (el === safeActive) {
+        el.tabIndex = 0;
+      } else {
+        el.tabIndex = -1;
+      }
+    }
+    this.dispatchActiveChange(safeActive);
+    this.options.onActiveItemChange?.(safeActive);
```
Rajdeepc: This means every `refresh()` call dispatches the custom event and fires the callback even if the same element remains active. This can cause infinite loops if a consumer's `onActiveItemChange` handler triggers a re-render that calls `refresh()`. The best way to handle this is to track the previous active item and only dispatch when it changes
```ts
if (safeActive !== this.previousActive) {
  this.previousActive = safeActive;
  this.dispatchActiveChange(safeActive);
  this.options.onActiveItemChange?.(safeActive);
}
```

2nd-gen/packages/core/controllers/focusgroup-navigation-controller/src/focusgroup-navigation-controller.ts
```diff
@@ -0,0 +1,1109 @@
...
+   * @returns True if the element counts as navigable for this controller.
+   */
+  private isNavigableItem(el: HTMLElement): boolean {
+    if (!el.isConnected) {
+      return false;
+    }
+    if (el.hasAttribute('inert') || el.closest('[inert]')) {
+      return false;
+    }
+    const style = getComputedStyle(el);
```
Rajdeepc, line 546: `getEligibleItems()` is getting invoked in `refresh()`, `setActiveItem()`, `focusFirstItemByTextPrefix()`, `handleFocusin()`, and `handleKeydown()`. 
So for a 20-item list, that's 20 style recalculations per keystroke. On large grids or frequent DOM mutations, this will turn into a layout thrashing hot path.

Please cache eligible items per refresh() cycle and reuse the cache until the next refresh():
```ts
private cachedEligibleItems: HTMLElement[] | null = null;
private getEligibleItems(): HTMLElement[] {
    if (this.cachedEligibleItems) {
      return this.cachedEligibleItems;
    }
    this.cachedEligibleItems = this.getRawItems().filter((el) => this.isNavigableItem(el));
    return this.cachedEligibleItems;
  }
public refresh(): void {
    this.cachedEligibleItems = null;
    // ... rest of refresh
  }
// Invalidate on focusin/focusout too:
private handleFocusin(event: FocusEvent): void {
    this.cachedEligibleItems = null;
    // ...
  }
```

2nd-gen/packages/core/controllers/focusgroup-navigation-controller/src/focusgroup-navigation-controller.ts
```diff
@@ -0,0 +1,1109 @@
...
+    if (event.defaultPrevented || event.altKey) {
+      return;
+    }
+
+    const items = this.getEligibleItems();
+    const target = this.resolveManagedKeydownTarget(event, items);
+    if (!target) {
+      return;
+    }
+
```
Rajdeepc, line 803:
```ts
// Add these lines here to cache buildRows

const isGrid = this.options.direction === 'grid';
const rows = isGrid ? this.buildRows(items) : null;

```

2nd-gen/packages/core/controllers/focusgroup-navigation-controller/src/focusgroup-navigation-controller.ts
```diff
@@ -0,0 +1,1109 @@
...
+    if (
+      this.options.direction === 'grid' &&
+      event.ctrlKey &&
+      !event.metaKey &&
+      (event.key === 'Home' || event.key === 'End')
+    ) {
+      const grid = this.buildRows(items);
```
Rajdeepc: Same pattern here too you are calling this method multiple times per grid keydown. 
Please compute the rows once at the top of the `handleKeyDown` for grid mode and pass it through. You can add a `rows` parameter to navigation methods. 
Tried to create this as optimised as possible I copy pasted your existing implementation with some tweaks.

```ts
private handleKeydown(event: KeyboardEvent): void {
    if (event.defaultPrevented || event.altKey) {
      return;
    }

    const items = this.getEligibleItems();
    const target = this.resolveManagedKeydownTarget(event, items);
    if (!target) {
      return;
    }

    const isGrid = this.options.direction === 'grid';
    const rows = isGrid ? this.buildRows(items) : null;

    // Ctrl+Home/End for grid
    if (isGrid && event.ctrlKey && !event.metaKey &&
        (event.key === 'Home' || event.key === 'End')) {
      if (rows!.length > 0) {
        const firstRow = rows![0];
        const lastRow = rows![rows!.length - 1];
        const boundary = event.key === 'Home'
          ? (firstRow?.[0] ?? null)
          : (lastRow?.[lastRow.length - 1] ?? null);
        if (boundary && boundary !== target) {
          event.preventDefault();
          this.moveKeyNavigationFocusTo(boundary);
        }
      }
      return;
    }

    if (event.ctrlKey || event.metaKey) {
      return;
    }

    // Page keys
    const pageMagnitude = this.getEffectivePageMagnitude();
    if (pageMagnitude !== null &&
        (event.key === 'PageUp' || event.key === 'PageDown')) {
      const delta = event.key === 'PageDown' ? pageMagnitude : -pageMagnitude;
      const pageNext = isGrid
        ? this.navigatePageGridRows(items, target, delta, rows!)
        : this.navigatePageLinearItems(items, target, delta);
      if (pageNext && pageNext !== target) {
        event.preventDefault();
        this.moveKeyNavigationFocusTo(pageNext);
      }
      return;
    }

    const rtl = this.isRtl();
    let next: HTMLElement | null = null;

    switch (this.options.direction) {
      case 'horizontal':
        next = this.navigateLinear(items, target, event.key, 'horizontal', rtl);
        break;
      case 'vertical':
        next = this.navigateLinear(items, target, event.key, 'vertical', rtl);
        break;
      case 'both':
        next = this.navigateBothAxes(items, target, event.key, rtl);
        break;
      case 'grid':
        next = this.navigateGridWithRows(rows!, target, event.key, rtl);
        break;
    }

    if (next && next !== target) {
      event.preventDefault();
      this.moveKeyNavigationFocusTo(next);
      return;
    }

    if (event.key === 'Home' || event.key === 'End') {
      const ordered = isGrid ? rows!.flat() : items;
      if (ordered.length === 0) return;
      const boundary = event.key === 'Home' ? ordered[0] : ordered[ordered.length - 1];
      if (boundary && boundary !== target) {
        event.preventDefault();
        this.moveKeyNavigationFocusTo(boundary);
      }
    }
  }

// Then update navigateGrid to accept pre-built rows:
private navigateGridWithRows(
    grid: HTMLElement[][],
    current: HTMLElement,
    key: string,
    rtl: boolean
  ): HTMLElement | null {
    const pos = this.findGridIndex(grid, current);
    if (!pos) return null;
    // ... same logic, using the passed-in grid
  }

// And navigatePageGridRows to accept rows:
private navigatePageGridRows(
    items: HTMLElement[],
    current: HTMLElement,
    rowDelta: number,
    grid: HTMLElement[][]
  ): HTMLElement | null {
    if (grid.length === 0) return null;
    const pos = this.findGridIndex(grid, current);
    if (!pos) return null;
    // ... same logic, using the passed-in grid
  }

```

2nd-gen/packages/core/controllers/focusgroup-navigation-controller/src/focusgroup-navigation-controller.ts
```diff
@@ -0,0 +1,1109 @@
...
+        this.moveKeyNavigationFocusTo(boundary);
+      }
+    }
+  }
+
+  /**
+   * Applies roving tabindex to `item` and moves DOM focus; used for keyboard navigation only.
+   */
+  private moveKeyNavigationFocusTo(item: HTMLElement): void {
+    if (this.setActiveItem(item)) {
```
Rajdeepc: There is a double dispatch here. 
`setActiveItem` ---> `applyRovingTabindex` ---->  `item.focus()` triggers `handleFocusIn` which calls `applyRovingTabindex` again.
You can guard this with a flag.

```ts
private isNavigating = false;


private moveKeyNavigationFocusTo(item: HTMLElement): void {
    this.isNavigating = true;
    try {
      if (this.setActiveItem(item)) {
        item.focus();
      }
    } finally {
      this.isNavigating = false;
    }
  }

// in handleFocusin

private handleFocusin(event: FocusEvent): void {
    if (this.isNavigating) {
      return;
    }

....

}
```

2nd-gen/packages/core/mixins/disabled-mixin.ts
```diff
@@ -0,0 +1,85 @@
...
+     */
+    protected override update(changedProperties: PropertyValues): void {
+      if (changedProperties.has('disabled')) {
+        if (this.disabled) {
+          // Host gets aria-disabled for screen reader discoverability.
+          // Components wrapping native form controls should ALSO set
+          // disabled on the inner element in render().
+          this.setAttribute('aria-disabled', 'true');
+          if (this.hasAttribute('tabindex')) {
+            this.dataset.prevTabindex = this.getAttribute('tabindex')!;
```
Rajdeepc: This writes a `data-prev-tabindex` attribute on host element. Since you are using it in a core mixin my advice would be to use a WeakMap or a private class field.

```ts
const prevTabindexMap = new WeakMap<HTMLElement, string>();
```

2nd-gen/packages/core/controllers/focusgroup-navigation-controller/src/focusgroup-navigation-controller.ts
```diff
@@ -0,0 +1,1109 @@
...
+
+  /**
+   * Sets `tabindex="-1"` on ineligible raw items, then assigns `tabindex="0"` to
+   * `active` (or the first eligible item if `active` is not eligible) and `-1` to the rest.
+   * Dispatches the active-change event and {@link FocusgroupNavigationOptions.onActiveItemChange}.
+   *
+   * @param active - Preferred item to mark as the single tab stop when eligible.
+   */
+  private applyRovingTabindex(active: HTMLElement): void {
+    const items = this.getEligibleItems();
```
Rajdeepc, line 625: This calls getRawItems() internally but you can again calling `const raw = this.getRawItems();` below. Its redundant. Compute it once and derive the eligable from it.

2nd-gen/packages/core/controllers/focusgroup-navigation-controller/src/focusgroup-navigation-controller.ts
```diff
@@ -0,0 +1,1109 @@
...
+    const raw = this.getRawItems();
+    for (const el of raw) {
+      if (!items.includes(el)) {
```
Rajdeepc: How about you use a Set for O(1) lookps instead of `Array.includes`?

2nd-gen/packages/core/controllers/focusgroup-navigation-controller/src/focusgroup-navigation-controller.ts
```diff
@@ -0,0 +1,1109 @@
...
+
+  /**
+   * Lit `ReactiveController` hook: registers capture-phase listeners on `host` and runs
+   * an initial {@link refresh}.
+   */
+  public hostConnected(): void {
+    this.host.addEventListener('keydown', this.boundKeydown, true);
+    this.host.addEventListener('focusin', this.boundFocusin, true);
+    this.host.addEventListener('focusout', this.boundFocusout, true);
+    this.refresh();
```
Rajdeepc, line 428: You need to destroy lastFocused on reconnect.

```ts
// Add these
this.lastFocused = null;
this.previousActive = null;
this.cachedEligibleItems = null;
```

2nd-gen/packages/core/element/spectrum-element.ts
```diff
@@ -34,39 +35,10 @@ export function SpectrumMixin<T extends Constructor<ReactiveElement>>(
...
-      // selector will throw on the following test (Safari, older things).
-      // Some won't throw, but will be focusing item rather than the menu and
-      // will rely on the polyfill to know whether focus is "visible" or not.
-      return (
-        activeElement.matches(':focus-visible') ||
-        activeElement.matches('.focus-visible')
+      const active = getActiveElement(
+        this.getRootNode() as Document | ShadowRoot
       );
+      return active?.matches(':focus-visible') ?? false;
```
Rajdeepc, line 46: I will add a comment here
```suggestion
// .focus-visible polyfill class (used in 1st-gen for older Safari) is
// intentionally not checked; 2nd-gen targets browsers with native support.
      return active?.matches(':focus-visible') ?? false;
```

Rajdeepc (dismissed):
Nice effort. I have few suggestions in some areas where we can improve the cyclometric complexity as well as how we reduce DOM burnouts. I did an initial pass and please find my feedback. Open to discuss if you need any guidance or you feel it can be done in an alternate way.
Blocking for me would be
1. I see a conflict in DisabledMixin + FocusGroupNavigationController tabindex. Would love to see a new test and potentially a coordination mechanism. 
2. Please add a test for memory behaviour and RTL behaviour, dynamic item addition/removal, onActiveItemChange callback, setOptions(), for inert attribute handling, disconnection/reconnection
3. Add a migration path for `elementEnterAction`. We are now using `onActiveItemChange`

2nd-gen/packages/core/controllers/focusgroup-navigation-controller/src/focusgroup-navigation-controller.ts
```diff
@@ -0,0 +1,1109 @@
...
+// geometry-based **grid** navigation; **pageStep** (Page Up/Down magnitude); **skipDisabled** and
+// `isDisabledForSkip`; `isNodeWithinHostScope` / `getRawItems` if declarative scoping differs in
+// shadow DOM; `dispatchActiveChange`, `onActiveItemChange`, and the exported event name if
+// products still want a single composed integration hook; `focusFirstItemByTextPrefix` for
+// typeahead roving tabindex (callers focus `getActiveItem()` unless the platform adds an equivalent).
+// `isRtl()` may
+// duplicate or diverge from native axis mapping — revisit when testing RTL with native focusgroup.
+// ─────────────────────────────────────────────────────────────────────────────
+
+export class FocusgroupNavigationController implements ReactiveController {
```
miwha-adobe, line 223: **Suggestion (non-blocking):** This is a pretty large comment for the class. I wonder if it warrants a readme or some sort of supporting doc, instead of just JSDoc.

2nd-gen/packages/core/controllers/focusgroup-navigation-controller/src/focusgroup-navigation-controller.ts
caseyisonit:
```suggestion
      if (!items.has(el)) {
```

2nd-gen/packages/core/controllers/focusgroup-navigation-controller/src/focusgroup-navigation-controller.ts
```diff
@@ -0,0 +1,1136 @@
...
+  //
+  // Which parts may become redundant under native `focusgroup` is summarized in the
+  // “Native `focusgroup` (future)” comment block directly above the class declaration.
+
+  /**
+   * Resolves `dir` from the shadow host, nearest `dir` ancestor, or `document.documentElement`.
+   *
+   * @returns True when horizontal arrow directions should follow RTL semantics.
+   */
+  private isRtl(): boolean {
```
rubencarvalho, line 455: we're not applying `dir` anymore, so direction is inherited via CSS, so the attribute-based checks will prob fail!
I think we can try something like `getComputedStyle(this.host).direction === 'rtl'`

2nd-gen/packages/core/controllers/focusgroup-navigation-controller/src/focusgroup-navigation-controller.ts
```diff
@@ -0,0 +1,1136 @@
...
+    }
+
+    if (next && next !== target) {
+      event.preventDefault();
+      this.moveKeyNavigationFocusTo(next);
+      return;
+    }
+
+    if (event.key === 'Home' || event.key === 'End') {
+      const ordered = isGrid ? rows!.flat() : items;
```
rubencarvalho: should pressing Home/End in this grid mode scope to the current row rather than the entire grid? that seems the APG recommendation https://www.w3.org/WAI/ARIA/apg/patterns/grid/#keyboardinteraction

Rajdeepc: Good call — you're correct per the APG grid pattern spec. Home/End should scope to the current row, with Ctrl+Home/End covering the entire grid.

2nd-gen/packages/core/controllers/focusgroup-navigation-controller/src/focusgroup-navigation-controller.ts
Rajdeepc: Great catch

2nd-gen/packages/core/element/spectrum-element.ts
caseyisonit: made the comment in 1st gen so we dont have any tech debt comments in 2nd-gen

2nd-gen/packages/core/controllers/focusgroup-navigation-controller/src/focusgroup-navigation-controller.ts
caseyisonit: ill be looking in to the docs pattern for controllers and other types of content in the docs IA epic and will note this concern, i am in agreement, its not great.

5t3ph:
In the Playground story, I would expect to be able to arrow through items linearly, regardless of disabled status. I'm not sure if it should still give focus to the disabled items, or just skip over them. But if you try arrow nav, you will currently get blocked due to the disabled "Underline". It looks like at some point `tabindex="0"` _is_ trying to be set on the `disabled` button.
image

Edit: I now see the `skipDisabled` option, but maybe still worth some kind of change to the Playground, or maybe that should be default?

5t3ph (commented):
Suggestion: prevent `tabindex="0"` from being set on a `disabled` button and instead skip it since otherwise the other group items are modified to `tabindex="-1"` and the entire group becomes unnavigable via keyboard.

miwha-adobe:
Question: When inspecting the dom, it shows me that strikethrough has aria-disabled, and disabled has the "disabled" attribute. I'm curious why strikethrough visually appears disabled when hovered over but only has the aria-attribute. Should it also have the disabled attribute, not just aria? It gets skipped when `skipDisabled` is true, but when set to false, only 'disabled' button is skipped.

Screenshot 2026-04-13 at 12 54 55 PM

caseyisonit:
> Question: When inspecting the dom, it shows me that strikethrough has aria-disabled, and disabled has the "disabled" attribute. I'm curious why strikethrough visually appears disabled when hovered over but only has the aria-attribute. Should it also have the disabled attribute, not just aria? It gets skipped when skipDisabled is true, but when set to false, only 'disabled' button is skipped.

@miwha-adobe great question, so theres some nuance around `aria-disabled` and `disabled` attritbutes. Based on this article @5t3ph shared about the accessibility implications and having the ability to opt in to certain disability states, this story should be exemplifying the difference between the two. Aria-disable allows for tabbing to exist and for screen readers to inform a user of an action that is avaialble but disabled for some reason i.e. form isn't filled out entirely so submit button is aria disabled, the user should still know the button exists but isnt active. 

disabled attribute effectively makes it hidden and untabble/identifiable by screen readers and keyboard. I think Steph has highlighted an issue with the breaking of keyboard navigation and perhaps making skipDisabled the default so you can continue with keyboard and opt in to the block. e.g. walking a user through a stepper where steps need to be completed before you can reach the next step and dont want a user to skip it

Hope this helps, im still wrapping my brain around the nuance too but the article i think explains it much better than me! :D 

https://kittygiraudel.com/2024/03/29/on-disabled-and-aria-disabled-attributes/

nikkimk:
> In the Playground story, I would expect to be able to arrow through items linearly, regardless of disabled status. I'm not sure if it should still give focus to the disabled items, or just skip over them. But if you try arrow nav, you will currently get blocked due to the disabled "Underline". It looks like at some point `tabindex="0"` _is_ trying to be set on the `disabled` button. image
> 
> Edit: I now see the `skipDisabled` option, but maybe still worth some kind of change to the Playground, or maybe that should be default?

From the [APG](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/)

> By default, disabled HTML input elements are removed from the tab sequence. In most contexts, the normal expectation is that disabled interactive elements are not focusable. However, there are some contexts where it is common for disabled elements to be focusable, especially inside of composite widgets. For example, as demonstrated in the [menu and menubar pattern](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/), disabled items are focusable when navigating through a menu with the arrow keys.

Since componenents using focus group / RTI nav are considered "composite widgets", I leaned toward making disabled items focusable as the default.

Or perhaps we should skip `disabled` but not skip `aria-disabled` and recommend that composite widgets use `aria-disabled` and the we wouldn't need the `skipDisabled` parameter?

caseyisonit:
> Or perhaps we should skip disabled but not skip aria-disabled and recommend that composite widgets use aria-disabled and the we wouldn't need the skipDisabled parameter?

I think this is the correct mental model we should adopt. `disabled` is always skipped and `aria-disabled` behaves the way its intended. does this have any effect on how we setup disabledMixin?

2nd-gen/packages/core/controllers/focusgroup-navigation-controller/src/focusgroup-navigation-controller.ts
Rajdeepc: Great catch, Ruben. You're absolutely right — since 2nd-gen inherits direction via CSS rather than applying `dir` attributes directly, the attribute-based checks here (`getAttribute('dir')`, `closest('[dir]')`) would silently fail and fall through to `document.documentElement.dir`, which may also be unset.

Updated `isRtl()` to use `getComputedStyle(this.host).direction === 'rtl'` as you suggested. This correctly resolves direction whether it comes from a `dir` attribute, a CSS `direction` property, or CSS inheritance — all in a single line.

Also added a dedicated test (`CSSInheritedRTL`) that wraps the demo host in `<div style="direction: rtl">` with no `dir` attribute and verifies ArrowLeft/ArrowRight swap correctly. This test would have failed with the previous attribute-walking implementation.

2nd-gen/packages/core/controllers/focusgroup-navigation-controller/src/focusgroup-navigation-controller.ts
Rajdeepc: Good call — you're correct per the [APG Grid Pattern keyboard interaction spec](https://www.w3.org/WAI/ARIA/apg/patterns/grid/#keyboardinteraction). The spec defines:

- **Home**: moves focus to the first cell in the row that contains focus
- **End**: moves focus to the last cell in the row that contains focus
- **Ctrl+Home**: moves focus to the first cell in the first row
- **Ctrl+End**: moves focus to the last cell in the last row

Updated the implementation so Home/End in grid mode now uses `findGridIndex` to locate the current row and navigates to the first/last cell within that row only. Ctrl+Home/End behavior (full grid scope) was already correct and remains unchanged.

Updated the grid test to verify the new row-scoped behavior — for example, from cell 5 (middle row), Home now goes to cell 4 and End to cell 6, rather than cells 1 and 9.

Rajdeepc:
@5t3ph Great catch, Stephanie. This is a real accessibility issue — natively `disabled` elements can't receive focus regardless of their `tabindex` value, so if the controller assigns `tabindex="0"` to one and `tabindex="-1"` to everything else, the entire group becomes unreachable via the Tab key.

We should update `applyRovingTabindex` to detect natively disabled elements (via the `disabled` DOM property) and fall through to the next non-disabled eligible item for the roving tab stop. To be clear about the behavior:

- When `skipDisabled: false` (the default), disabled items remain in the arrow navigation order — this is intentional for `aria-disabled` patterns where the item should still be focusable and discoverable
- But only items that can actually receive browser focus are eligible to hold `tabindex="0"` (the tab entry point)

I have also added a dedicated test (`DisabledButtonNeverTabStop`) that natively disables the first button in a group and verifies the `tabindex="0"` tab stop moves to the next eligible item, preventing the keyboard trap.

Rajdeepc (approved):
Approving on the basis that the controller logic, mixin, utilities, tests, and documentation are in solid shape. 

One additional item surfaced during team discussion (non-blocking): skipDisabled currently treats natively disabled and aria-disabled identically, but these are fundamentally different states — natively disabled elements can't receive focus, while aria-disabled elements can and should remain discoverable per APG guidance. The Tabs migration should help clarify the right granularity here. Tracking as a follow-up.

2nd-gen/packages/core/controllers/focusgroup-navigation-controller/src/focusgroup-navigation-controller.ts
```diff
@@ -0,0 +1,1185 @@
...
+    }
+  }
+
+  /**
+   * Positive step count for {@link FocusgroupNavigationOptions.pageStep}, or null when page keys
+   * are disabled.
+   */
+  private getEffectivePageMagnitude(): number | null {
+    const raw = this.options.pageStep;
+    if (raw === undefined || raw === null) {
```
rubencarvalho: nit: `pageStep` is typed as `number | undefined` so it'll never be null.
```suggestion
    if (raw === undefined) {
```

2nd-gen/packages/core/controllers/focusgroup-navigation-controller/src/focusgroup-navigation-controller.ts
```diff
@@ -0,0 +1,1185 @@
...
+
+    this.cachedEligibleItems = null;
+    const items = this.getEligibleItems();
+    const target = this.resolveManagedKeydownTarget(event, items);
+    if (!target) {
+      return;
+    }
+
+    const isGrid = this.options.direction === 'grid';
+    const rows = isGrid ? this.buildRows(items) : null;
```
rubencarvalho: nit: I see we did cache eligible items above (`cachedEligibleItems`), curious why rows didn't get the same treatment?

2nd-gen/packages/core/mixins/disabled-mixin.ts
```diff
@@ -0,0 +1,87 @@
...
+          // Host gets aria-disabled for screen reader discoverability.
+          // Components wrapping native form controls should ALSO set
+          // disabled on the inner element in render().
+          this.setAttribute('aria-disabled', 'true');
+          if (this.hasAttribute('tabindex')) {
+            this.prevTabindex = this.getAttribute('tabindex')!;
+            this.setAttribute('tabindex', '-1');
+          }
+          if (this.matches(':focus-within')) {
+            (this.shadowRoot?.activeElement as HTMLElement)?.blur();
```
rubencarvalho, line 73: if the host element itself is focused (e.g. tabindex="0" on the host, focused directly with no shadow children), `shadowRoot?.activeElement` is `null` and I guess the blur never fires?? so the element will stay focused while disabled? maybe we'd also want to handle `this === document.activeElement` and call `this.blur()` in that case? could we cover this with a test?

2nd-gen/packages/core/controllers/focusgroup-navigation-controller/src/focusgroup-navigation-controller.ts
```diff
@@ -0,0 +1,1185 @@
...
+  memory?: boolean;
+
+  /**
+   * When true, items that are disabled for interaction are skipped for arrow
+   * navigation and are not chosen as the roving tab stop. When false, disabled
+   * items remain in sequence (useful for patterns such as menus where disabled
+   * items may still be focusable per APG guidance).
+   * Defaults to false.
+   */
+  skipDisabled?: boolean;
```
rubencarvalho, line 78: Can we call out that this skips both `disabled` and `aria-disabled` elements? IDK where we landed on our discussions, but are we sure we should be coupling these and by default skipping aria-disabled too? There's not a way to control that now but I'm not sure we'd want to also have `skipAriaDisabled` for more control. Not blocking, but just curious to learn about this 😄

2nd-gen/packages/core/controllers/focusgroup-navigation-controller/test/focusgroup-navigation-controller.test.ts
```diff
@@ -0,0 +1,1136 @@
...
+        (document.activeElement as HTMLElement)?.blur();
+
+        const tabbable = buttons.filter((b) => b.tabIndex === 0);
+        expect(tabbable.length).toBe(1);
+        expect(tabbable[0].textContent?.trim()).toBe('Underline');
+      }
+    );
+
+    await step(
+      'Tab re-entry to the group lands on the remembered item',
```
rubencarvalho, line 564: blocking: could we add a step to test the opposite?, when memory: false, Tab re-entry should reset to the first item, not the last focused one 😃

2nd-gen/packages/core/controllers/focusgroup-navigation-controller/test/focusgroup-navigation-controller.test.ts
```diff
@@ -0,0 +1,1136 @@
...
+      }
+    });
+
+    await step('Home and End stay within eligible items only', async () => {
+      buttonByLabel('Print').focus();
+      keydown(shadowActiveButton(host)!, 'Home');
+      expect(shadowActiveButton(host)?.textContent?.trim()).toBe('New');
+      keydown(shadowActiveButton(host)!, 'End');
+      expect(shadowActiveButton(host)?.textContent?.trim()).toBe('Help');
+    });
```
rubencarvalho, line 318: blocking: could we add a step here (or a separate story) for the case where skipDisabled: false and the first item in the group is natively disabled? there's a fallthrough in applyRovingTabindex that picks the first non-disabled item as the tab stop in that scenario, but I don't see it covered

2nd-gen/packages/core/controllers/focusgroup-navigation-controller/test/focusgroup-navigation-controller.test.ts
```diff
@@ -0,0 +1,1136 @@
...
+    );
+
+    await step('after removing inert, item is navigable again', async () => {
+      const buttons = Array.from(
+        root.querySelectorAll<HTMLButtonElement>('button')
+      );
+      buttons[0].focus();
+      keydown(buttons[0], 'ArrowRight');
+      expect(shadowActiveButton(host)?.textContent?.trim()).toBe('Beta');
+    });
```
rubencarvalho, line 934: blocking: could we also add a step here for `visibility: hidden` and `display: none`? `isNavigableItem` filters those the same way it handles `[inert]`, but they're not covered, feels fairly low effort to try and cover

rubencarvalho (changes requested):
left a couple of nits/questions, and a few requests to add a couple steps to the tests. other than that, looks very good!! 🚀

2nd-gen/packages/core/mixins/disabled-mixin.ts
caseyisonit: @rubencarvalho can we defer this until we have a component use case. This ensures that if a component has any actively focused children when the component is disabled, the children lose focus, so it's not trapped. I think this is a solid starting point and we can continue to iterate and evolve this.

2nd-gen/packages/core/controllers/focusgroup-navigation-controller/src/focusgroup-navigation-controller.ts
caseyisonit: i will be modifying this based on the comments above from rajdeep nikki and steph. this will be outdated soon :)

rubencarvalho (approved):
🚀 🚀

PR #4215, opened 2024-03-26
fix(coachmark): add "step-count" slot for custom/internationalized pagination content

Rajdeepc:
## Description

This PR adds the ability for users to customize the separator between the current step and the total steps in the YourComponent component. By default, the separator is 'of', but users can now specify a custom separator using the separator attribute.

### Changes Made
Added a new separator property to the YourComponent component to allow customization of the separator.
Updated the component's template to use the separator property for the separator between the current step and the total steps.

## Related issue(s)

- #4213 

## How has this been tested?

-   [x] _Test case 1_
    1. Go [here](https://feature-coachmark-localization--spectrum-web-components.netlify.app/storybook/?path=/story/coachmark--custom-separator)
    2. Use the separator attribute to specify a custom separator.
    3. Verify that the component renders with the custom separator between the current step and the total steps.
-   [ ] _Test case 2_
    1. Go here
    2. Do this

## Screenshots (if appropriate)
Screenshot 2024-03-26 at 4 23 54 PM

## Types of changes

-   [ ] Bug fix (non-breaking change which fixes an issue)
-   [x] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to change)
-   [ ] Chore (minor updates related to the tooling or maintenance of the repository, does not impact compiled assets)

## Checklist

-   [x] I have signed the [Adobe Open Source CLA](http://opensource.adobe.com/cla.html).
-   [x] My code follows the code style of this project.
-   [x] If my change required a change to the documentation, I have updated the documentation in this pull request.
-   [x] I have read the **[CONTRIBUTING](<(https://github.com/adobe/spectrum-web-components/blob/main/CONTRIBUTING.md)>)** document.
-   [x] I have added tests to cover my changes.
-   [x] All new and existing tests passed.
-   [x] I have reviewed at the Accessibility Practices for this feature, see: [Aria Practices](https://www.w3.org/TR/wai-aria-practices/)

## Best practices

This repository uses conventional commit syntax for each commit message; note that the GitHub UI does not use this by default so be cautious when accepting suggested changes. Avoid the "Update branch" button on the pull request and opt instead for rebasing your branch against `main`.

PR #6167, opened 2026-04-10
fix: make coachmark step count separator localizable

### LINKED ISSUE #4780: [Bug]: Coachmark step count is not localized
### Code of conduct

### Impacted component(s)

Coachmark

### Expected behavior

In the Coachmark component the step count, for example "1 of 3" has hardcoded the "of" string. The expected behavior is that this string is localized, or is at least configurable so the client can pass in the string to use in this case. 

### Actual behavior

The "of" word is hardcoded. 

### Screenshots

issue-2
issue-3

Chrome

Bug was discovered in a pre-release version so I can't share my own reproduction steps, as the feature has not been publicly released. But this surely reproduces anywhere Coachmark component is used. 
1. Change language to non-english
2. Enter multi-step coachmark tour 
Code where this is hard-coded is here https://github.com/adobe/spectrum-web-components/blob/a2b6bea7142930d8143d9ca887e2c56ffa71a669/packages/coachmark/src/Coachmark.ts#L278 

_No response_

### Severity

SEV 3

_No response_

mvanhorn:
## Summary

The Coachmark component has a hardcoded "of" string in the step count display (e.g., "1 of 3"). This adds a `step-count-of` property so the separator can be localized.

## Changes

`1st-gen/packages/coachmark/src/Coachmark.ts`:
- Added `stepCountOf` property with `step-count-of` attribute (defaults to `"of"`)
- Changed the step count template from hardcoded `${this.currentStep} of ${this.totalSteps}` to `${this.currentStep} ${this.stepCountOf} ${this.totalSteps}`

## Usage

```html
<!-- Default (English) -->
<sp-coachmark current-step="1" total-steps="3">...</sp-coachmark>
<!-- Outputs: "1 of 3" -->

<!-- Localized (Spanish) -->
<sp-coachmark current-step="1" total-steps="3" step-count-of="de">...</sp-coachmark>
<!-- Outputs: "1 de 3" -->
```

The existing `step-count` slot still works as a complete override for custom formatting.

Fixes #4780

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/1st-gen/packages/coachmark/src/Coachmark.ts
+++ b/1st-gen/packages/coachmark/src/Coachmark.ts
@@ -84,6 +84,9 @@ export class Coachmark extends Popover {
   @property({ type: Number, attribute: 'total-steps' })
   public totalSteps?: number;
 
+  @property({ type: String, attribute: 'step-count-of' })
+  public stepCountOf = 'of';
+
   @property({ type: String, attribute: 'primary-cta' })
   primaryCTA?: string;
 
@@ -268,7 +271,7 @@ export class Coachmark extends Popover {
       <div class="step" role="status">
         <span aria-live="polite">
           <slot name="step-count">
-            ${this.currentStep} of ${this.totalSteps}
+            ${this.currentStep} ${this.stepCountOf} ${this.totalSteps}
           </slot>
         </span>
       </div>


Rajdeepc (changes requested):
@mvanhorn The step-count slot was specifically added in [#4215](https://github.com/adobe/spectrum-web-components/issues/4215) to solve this exact problem. The [stories](https://github.com/adobe/spectrum-web-components/blob/rajdeepchandra/feat-pr-comment-stackblitz-links/1st-gen/packages/coachmark/stories/coachmark.stories.ts) already demonstrate localized usage via the slot:
```ts
<div slot="step-count">4 分の 1</div>
```

Adding step-count-of introduces a second, narrower mechanism for the same purpose. This creates API surface area that needs to be maintained, documented, and tested without providing meaningful value over the existing slot.
A step-count-of property can handle French and German but cannot handle Japanese or Chinese where the entire order and structure change. The existing step-count slot already handles all of these cases correctly because it lets the consumer provide the entire formatted string.

**Looking ahead:**
A proper solution would look more like using` LanguageResolutionController` to detect the locale, then formatting with a message catalog or template string system that supports argument reordering (e.g., ICU MessageFormat patterns like {current, number} of {total, number}).

Will close this PR once you validate the already existing approach above.

mvanhorn:
You're right - the `step-count` slot already handles this correctly with full flexibility for any locale structure. I missed that #4215 had already solved this. An attribute-based approach is too narrow for CJK and other languages where the number order/structure changes entirely.

Closing this PR. Thanks for the detailed explanation.

PR #6188, opened 2026-04-17
feat(docs): add RTL decorator

rubencarvalho:
## Description

Adds a dedicated **Direction** Storybook toolbar control (`auto` / `LTR` / `RTL`) alongside the existing **Language** control. Direction resolution:

- **Auto** — follows locale (RTL for Arabic, Hebrew, and Farsi when applicable; otherwise LTR).
- **LTR** / **RTL** — explicit override, independent of language (so typography/font kit can stay on one locale while layout direction is forced).

The `withLanguageWrapper` decorator applies the resolved direction on the preview root using both:

- `dir` on `document.documentElement` (semantic / bidi), and
- `direction` via inline CSS on the same element (explicit visual direction).

`context.globals.resolvedTextDirection` is set so stories or tooling can read the final `ltr` | `rtl` value. The `globalsUpdated` listener was updated so direction changes apply even when the decorator does not re-run (e.g. some docs flows).

## Motivation and context

Language alone implied RTL for some locales, which made it awkward to test **LTR vs RTL** without changing locale, especially when pairing a specific font/locale with a fixed reading direction. A separate Direction control fixes that.

This also **unblocks Chromatic VRT**: you can template permutations (e.g. theme × scale × **explicit LTR/RTL**) without relying on switching to an RTL locale only to get `dir`/layout. Stories or Chromatic configuration can key off `globals.textDirection` / `resolvedTextDirection`.

## Author's checklist

- [ ] I have read the **[CONTRIBUTING](https://github.com/adobe/spectrum-web-components/blob/main/CONTRIBUTING.md)** and **[PULL_REQUESTS](https://github.com/adobe/spectrum-web-components/blob/main/PULL_REQUESTS.md)** documents.
- [ ] I have reviewed the Accessibility Practices for this feature, see: [ARIA Practices](https://www.w3.org/TR/wai-aria-practices/) _(toolbar sets document `dir`; no new interactive component in the library.)_
- [ ] I have included updated documentation if my change required it _(consider a short note in the Fonts / customization guide about **Direction** vs **Language**)._

---

### Manual review test cases

- [ ] **Direction toolbar + document root**
  1. Open any 2nd-gen Storybook story.
  2. Set **Language** to English and **Direction** to **RTL**.
  3. Inspect `<html>`: expect `dir="rtl"` and inline `direction: rtl` (or equivalent).
  4. Set **Direction** to **LTR**; expect `dir="ltr"` and LTR CSS.

- [ ] **Auto follows locale**
  1. Set **Direction** to **Auto**.
  2. Switch **Language** between English and Arabic (or Hebrew).
  3. Expect RTL when locale is RTL and LTR for English, without manual Direction overrides fighting Auto.

### DIFF
--- a/2nd-gen/packages/swc/.storybook/decorators/language.ts
+++ b/2nd-gen/packages/swc/.storybook/decorators/language.ts
@@ -15,15 +15,30 @@ import { addons, makeDecorator, useEffect } from '@storybook/preview-api';
 import type {} from '../storybook-env';
 
 const DEFAULT_KIT_ID = 'obc6cux';
+const RTL_LANGS = new Set(['ar', 'fa', 'he']);
 
 let languageListenerAttached = false;
 
+function resolveTextDirection(
+  lang: string | false,
+  textDirection: string | undefined
+): 'ltr' | 'rtl' {
+  if (textDirection === 'ltr' || textDirection === 'rtl') {
+    return textDirection;
+  }
+
+  return RTL_LANGS.has(String(lang ?? 'en-US')) ? 'rtl' : 'ltr';
+}
+
 /**
- * Applies the selected language (lang/dir) to the document and loads the corresponding
+ * Applies the selected language/dir to the document and loads the corresponding
  * Adobe Fonts kit if needed. Safe to call from the decorator or from a toolbar listener
  * so that font loading works even when the decorator does not re-run (e.g. on some docs pages).
  */
-function applyLanguageAndFontKit(lang: string | false, isRTL: boolean): void {
+function applyLanguageAndFontKit(
+  lang: string | false,
+  textDirection: 'ltr' | 'rtl'
+): void {
   const langAttr = lang ? String(lang) : 'en-US';
   const root = document.documentElement;
 
@@ -33,7 +48,10 @@ function applyLanguageAndFontKit(lang: string | false, isRTL: boolean): void {
     root.setAttribute('lang', langAttr);
     hasChanged = true;
   }
-  root.setAttribute('dir', isRTL ? 'rtl' : 'ltr');
+
+  // Apply both semantic and CSS direction so docs/stories and bidi behavior stay aligned.
+  root.setAttribute('dir', textDirection);
+  root.style.direction = textDirection;
 
   // If the fonts are actively loading, do not re-trigger the load
   if (window.FontsLoading === true) {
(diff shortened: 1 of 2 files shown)


pfulton (approved):
Looks great. I clicked-through to the deployed Storybook instance and was able to follow the testing instructions. Everything worked as expected.

Thanks for taking care of this!

2nd-gen/packages/swc/.storybook/decorators/language.ts
```diff
@@ -15,15 +15,30 @@ import { addons, makeDecorator, useEffect } from '@storybook/preview-api';
...
 
+function resolveTextDirection(
+  lang: string | false,
+  textDirection: string | undefined
+): 'ltr' | 'rtl' {
+  if (textDirection === 'ltr' || textDirection === 'rtl') {
+    return textDirection;
+  }
+
+  return RTL_LANGS.has(String(lang ?? 'en-US')) ? 'rtl' : 'ltr';
```
Rajdeepc: `??` only short‑circuits on `null | undefined`. When `lang === false` this evaluates to `String(false) === 'false'`, which never matches RTL_LANGS. You get LTR for the right reason by accident. You can use `lang || 'en-US'` to mirror the convention used in applyLanguageAndFontKit (lang ? String(lang) : 'en-US')

2nd-gen/packages/swc/.storybook/decorators/language.ts
```diff
@@ -173,35 +191,41 @@ function attachLanguageChangeListener(): void {
...
+      viewMode,
+    } = context;
 
-    // Add a textDirection property to the globals for use in the stories
-    // fa/Farsi is currently not included in the storybook toolbar map
-    const isRTL = ['ar', 'fa', 'he'].includes(lang);
-    context.globals.textDirection = isRTL ? 'rtl' : 'ltr';
+    // Keep an explicit, resolved direction available for stories.
+    const resolvedTextDirection = resolveTextDirection(lang, textDirection);
+    context.globals.resolvedTextDirection = resolvedTextDirection;
```
Rajdeepc: This can be an anti-pattern of mutating `context.globals` during render. If you want to keep this would you like to add a comment explaining the contract. I just want to avoid the future contributors to propagate this pattern.

2nd-gen/packages/swc/.storybook/decorators/language.ts
```diff
@@ -33,7 +48,10 @@ function applyLanguageAndFontKit(lang: string | false, isRTL: boolean): void {
     root.setAttribute('lang', langAttr);
     hasChanged = true;
   }
-  root.setAttribute('dir', isRTL ? 'rtl' : 'ltr');
+
+  // Apply both semantic and CSS direction so docs/stories and bidi behavior stay aligned.
+  root.setAttribute('dir', textDirection);
+  root.style.direction = textDirection;
```
Rajdeepc: `dir="rtl"` on an element already maps to `direction: rtl` via the UA stylesheet. This will add an inline specificity which can fight overrides further down. Can you show me an example where the `dir` attribute alone is not propagating?

2nd-gen/packages/swc/.storybook/decorators/language.ts
rubencarvalho: You're right - no need for this now, so I removed it :)

2nd-gen/packages/swc/.storybook/decorators/language.ts
rubencarvalho: Updated!

2nd-gen/packages/swc/.storybook/decorators/language.ts
rubencarvalho: updated!

2nd-gen/packages/swc/.storybook/decorators/language.ts
```diff
@@ -15,15 +15,30 @@ import { addons, makeDecorator, useEffect } from '@storybook/preview-api';
...
+ * Applies the selected language/dir to the document and loads the corresponding
  * Adobe Fonts kit if needed. Safe to call from the decorator or from a toolbar listener
  * so that font loading works even when the decorator does not re-run (e.g. on some docs pages).
  */
-function applyLanguageAndFontKit(lang: string | false, isRTL: boolean): void {
+function applyLanguageAndFontKit(
+  lang: string | false,
+  textDirection: 'ltr' | 'rtl'
+): void {
   const langAttr = lang ? String(lang) : 'en-US';
```
miwha-adobe: **nit:** This could also be updated to `lang || 'en-US'`, since the truthy path is always going to be a string.  

Also what happens if you are passed a invalid lang string? I know this is just storybook, but we might want to consider some stronger type safety.

Rajdeepc (approved):
Looks good!

2nd-gen/packages/swc/.storybook/decorators/language.ts
Rajdeepc: If we pass an invalid string here right now it will silently LTR. The set check is `RTL_LANGS.has(String(lang || 'en-US'))`, so any string not in `{'ar','fa','he'} → 'ltr'`

PR #6194, opened 2026-04-22
fix(overlay): popover tooltip renders in ioswebkit

Rajdeepc:
## Description
On iOS Safari and WKWebView hosts (such as the Adobe Express iOS app), `sp-popover` opened via `Overlay.open()` or the `trigger` Lit directive lands `visualViewport.offsetTop` pixels below its trigger — typically 30–40 px — whenever the layout viewport diverges from the visual viewport (URL bar showing, pinch-zoom, virtual keyboard open, host-app bottom sheet).
This PR fixes the offset and keeps the overlay anchored as the viewports realign.

### What changed
- **`PlacementController.computePlacement`** — on WebKit, subtract `visualViewport.offsetLeft / offsetTop` from the computed translate before applying it. Floating UI calculates `(x, y)` from `getBoundingClientRect()` (layout viewport) but the overlay paints in the top layer relative to the visual viewport; this closes the gap.
- **`PlacementController.placeOverlay`** — on WebKit, subscribe to `window.visualViewport` `resize` and `scroll`, and recompute placement on each event, so an open overlay re-anchors when the URL bar shows/hides, the keyboard opens, etc. Floating UI's `autoUpdate` does not observe `visualViewport`. Listeners are torn down on close.
- Both branches are gated on `isWebKit()` so Chromium and Firefox are unaffected.
- New regression Storybook story under **Overlay → Bug Repros → iOS Popover Positioning** that mirrors the bug-report screenshot and prints live `visualViewport` telemetry for on-device verification.
- New WebKit-only unit test that asserts the `visualViewport` `resize` and `scroll` listeners are attached on open and removed on close.
- Changeset added (`patch` to `@spectrum-web-components/overlay`).

## Motivation and context
Reported by Adobe Express iOS engineering: popovers anchored to toolbar buttons render below their triggers in the iOS app. The bug reproduces in plain iOS Safari any time the URL bar is visible and disappears when the page is scrolled enough that the URL bar collapses. Root cause is a coordinate-system mismatch between Floating UI's layout-viewport input and the top layer's visual-viewport rendering on iOS WebKit.
## Related issue(s)
- fixes SWC-2031 <SWC-2031> 

| Before (iOS, URL bar showing) | After |
| --- | --- |
| _attach the original screenshot from the Slack/GitHub report_ | _attach a screenshot from the new `iOS Popover Positioning` story after the fix_ |

## Author's checklist

## Reviewer's checklist

### Manual review test cases
- [ ] _iOS Safari — popover anchors to trigger with URL bar visible_
  1. Open this [story](https://swcpreviews.z13.web.core.windows.net/pr-6194/docs/first-gen-storybook/?path=/docs/overlay-bug-repros-ios-popover-positioning--docs)
  2. Go to iOS device mode and Tap the "..." button on any row while the URL bar is showing (`visualViewport.offsetTop > 0` in the telemetry panel).
  3. Expect the popover to anchor flush to the button (no vertical gap).

### Device review
- [x] Did it pass in Desktop?

  1. Open **Overlay → Bug Repros → iOS Popover Positioning** in desktop Safari.
  2. Tab to a row's "..." button and press <kbd>Enter</kbd> / <kbd>Space</kbd> to open the popover; press <kbd>Escape</kbd> to dismiss.
  3. Expect focus to move into the popover on open and return to the trigger on close — no change from current behavior; this fix is positional only.
  1. With VoiceOver on, navigate to a row's "..." button on iOS Safari and activate it.
  2. Expect VoiceOver to announce the popover's role/name as it does on `main`; re-anchoring fires only `sp-update-overlays` and does not announce.
  3. Confirm there are no duplicate or unexpected announcements when the URL bar shows/hides while the popover is open.

### DIFF
--- a/1st-gen/packages/overlay/src/PlacementController.ts
+++ b/1st-gen/packages/overlay/src/PlacementController.ts
@@ -210,15 +210,35 @@ export class PlacementController implements ReactiveController {
     );
 
     // Recompute placement when the visual viewport diverges from the layout
-    // viewport (iOS URL bar showing/hiding, pinch-zoom, virtual keyboard,
-    // host-app bottom sheets, etc.). Floating UI's `autoUpdate` does not
-    // observe `window.visualViewport`, so without this an open overlay can
-    // drift away from its trigger on iOS WebKit until the next resize.
-    // See `computePlacement` for the corresponding offset compensation.
+    // viewport (URL bar showing/hiding, pinch-zoom, virtual keyboard, host-app
+    // bottom sheets, etc.). Floating UI's `autoUpdate` does not observe
+    // `window.visualViewport`, so without this an open overlay can drift away
+    // from its trigger on WebKit (iOS Safari, WKWebView, and desktop Safari
+    // when pinch-zoomed) until the next resize. See `computePlacement` for the
+    // corresponding offset compensation.
+    //
+    // Listeners are passive (we never `preventDefault` in `updatePlacement`)
+    // and rAF-coalesced so a burst of `resize`/`scroll` events during a URL
+    // bar collapse or pinch gesture compresses to one Floating UI compute per
+    // frame.
     const visualViewport = window.visualViewport;
+    let visualViewportRafId = 0;
+    const onVisualViewportChange = (): void => {
+      if (visualViewportRafId) {
+        return;
+      }
+      visualViewportRafId = requestAnimationFrame(() => {
+        visualViewportRafId = 0;
+        this.updatePlacement();
+      });
+    };
     if (visualViewport && isWebKit()) {
-      visualViewport.addEventListener('resize', this.updatePlacement);
-      visualViewport.addEventListener('scroll', this.updatePlacement);
+      visualViewport.addEventListener('resize', onVisualViewportChange, {
+        passive: true,
+      });
+      visualViewport.addEventListener('scroll', onVisualViewportChange, {
+        passive: true,
+      });
     }
 
     // Define the cleanup function to remove event listeners and reset placements.
@@ -241,8 +261,11 @@ export class PlacementController implements ReactiveController {
       cleanupAncestorResize();
       cleanupElementResize();
       if (visualViewport && isWebKit()) {
-        visualViewport.removeEventListener('resize', this.updatePlacement);
-        visualViewport.removeEventListener('scroll', this.updatePlacement);
+        visualViewport.removeEventListener('resize', onVisualViewportChange);
+        visualViewport.removeEventListener('scroll', onVisualViewportChange);
+        if (visualViewportRafId) {
+          cancelAnimationFrame(visualViewportRafId);
+        }
       }
     };
   }
--- a/1st-gen/packages/overlay/src/PlacementController.ts
+++ b/1st-gen/packages/overlay/src/PlacementController.ts
@@ -209,6 +209,18 @@ export class PlacementController implements ReactiveController {
       }
     );
 
+    // Recompute placement when the visual viewport diverges from the layout
+    // viewport (iOS URL bar showing/hiding, pinch-zoom, virtual keyboard,
+    // host-app bottom sheets, etc.). Floating UI's `autoUpdate` does not
+    // observe `window.visualViewport`, so without this an open overlay can
+    // drift away from its trigger on iOS WebKit until the next resize.
+    // See `computePlacement` for the corresponding offset compensation.
+    const visualViewport = window.visualViewport;
+    if (visualViewport && isWebKit()) {
+      visualViewport.addEventListener('resize', this.updatePlacement);
+      visualViewport.addEventListener('scroll', this.updatePlacement);
+    }
+
     // Define the cleanup function to remove event listeners and reset placements.
     this.cleanup = () => {
       this.host.elements?.forEach((element) => {
@@ -228,6 +240,10 @@ export class PlacementController implements ReactiveController {
       });
       cleanupAncestorResize();
       cleanupElementResize();
+      if (visualViewport && isWebKit()) {
+        visualViewport.removeEventListener('resize', this.updatePlacement);
+        visualViewport.removeEventListener('scroll', this.updatePlacement);
+      }
     };
   }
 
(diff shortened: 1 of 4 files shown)


1st-gen/packages/overlay/src/PlacementController.ts
```diff
@@ -209,6 +209,18 @@ export class PlacementController implements ReactiveController {
...
     );
 
+    // Recompute placement when the visual viewport diverges from the layout
+    // viewport (iOS URL bar showing/hiding, pinch-zoom, virtual keyboard,
+    // host-app bottom sheets, etc.). Floating UI's `autoUpdate` does not
+    // observe `window.visualViewport`, so without this an open overlay can
+    // drift away from its trigger on iOS WebKit until the next resize.
+    // See `computePlacement` for the corresponding offset compensation.
+    const visualViewport = window.visualViewport;
+    if (visualViewport && isWebKit()) {
```
rubencarvalho, line 256: Doesn't this match all webkit? i.e. Safari Desktop too?

1st-gen/packages/overlay/src/PlacementController.ts
```diff
@@ -209,6 +209,18 @@ export class PlacementController implements ReactiveController {
...
+      visualViewport.addEventListener('resize', this.updatePlacement);
```
rubencarvalho: Should we be a bit more conservative here? Can you test and see how it feels with or without this being passive + maybe wrapping the handler in a raf?

rubencarvalho (commented):
Looks good! I have two nits, which prob are not blocking! 😄

1st-gen/packages/overlay/src/PlacementController.ts
Rajdeepc: `isWebKit()` is intentionally broad here because the top-layer / popover-API rendering paints relative to visualViewport on all WebKit, so desktop Safari has the same bug under pinch-zoom. I've updated the inline comment to drop the "iOS" framing so the gate's intent is clearer.

1st-gen/packages/overlay/src/PlacementController.ts
Rajdeepc: Both are good calls. Listeners are now { passive: true } (we never preventDefault in updatePlacement) and the handler is rAF-coalesced so a burst of resize/scroll during a URL-bar collapse or pinch gesture compresses to one Floating UI compute per frame. Felt smoother in the iOS Simulator on the repro story; no behavior change on desktop.

rubencarvalho:
WhatsApp Image 2026-04-27 at 09 05 40
I don't see the tip when opening the popovers. When I scroll they briefly show and then immediately close. Is this expected? cc: @Rajdeepc

Rajdeepc:
> I don't see the tip when opening the popovers. When I scroll they briefly show and then immediately close

I confirmed its a pre-existing bug from main which I will tackle it as a separate ticket. There seems to be a timing issue with the computePlacement when multiple tooltips are present.

rubencarvalho (approved):

1st-gen/packages/overlay/src/PlacementController.ts
```diff
@@ -209,6 +209,38 @@ export class PlacementController implements ReactiveController {
...
+    // from its trigger on WebKit (iOS Safari, WKWebView, and desktop Safari
+    // when pinch-zoomed) until the next resize. See `computePlacement` for the
+    // corresponding offset compensation.
+    //
+    // Listeners are passive (we never `preventDefault` in `updatePlacement`)
+    // and rAF-coalesced so a burst of `resize`/`scroll` events during a URL
+    // bar collapse or pinch gesture compresses to one Floating UI compute per
+    // frame.
+    const visualViewport = window.visualViewport;
+    let visualViewportRafId = 0;
```
blunteshwar, line 242: Async `computePlacement` can write styles after cleanup

The new `onVisualViewportChange` schedules `updatePlacement()` → `computePlacement()` on a frame. `computePlacement` then awaits `document.fonts.ready` and a WebKit `requestAnimationFrame` before its final `Object.assign(target.style, …)`. If the overlay closes during those awaits, cleanup runs (removing listeners and cancelling the outer rAF), but the in-flight `computePlacement` has already passed those guards and will still write `translate` to a torn-down target.

`cleanupAncestorResize` / `cleanupElementResize` use a `cancelled` flag for exactly this reason. Suggest the same pattern here:

```ts
let cancelled = false;
const visualViewport = window.visualViewport;
let visualViewportRafId = 0;

const onVisualViewportChange = (): void => {
    if (cancelled || visualViewportRafId) return;
    visualViewportRafId = requestAnimationFrame(() => {
        visualViewportRafId = 0;
        if (cancelled) return;
        this.updatePlacement();
    });
};

// existing add-listener block unchanged

this.cleanup = () => {
    cancelled = true;
    cleanupAncestorResize();
    cleanupElementResize();
    if (visualViewport && isWebKit()) {
        visualViewport.removeEventListener('resize', onVisualViewportChange);
        visualViewport.removeEventListener('scroll', onVisualViewportChange);
        if (visualViewportRafId) {
            cancelAnimationFrame(visualViewportRafId);
            visualViewportRafId = 0;
        }
    }
};

File: 1st-gen/packages/overlay/src/PlacementController.ts [hunk 4]
```diff
@@ -357,11 +396,30 @@ export class PlacementController implements ReactiveController {
...
+    let translateX = x;
+    let translateY = y;
+    const visualViewport = window.visualViewport;
+    if (visualViewport && isWebKit()) {
+      translateX -= visualViewport.offsetLeft;
+      translateY -= visualViewport.offsetTop;
+    }
+
     // Update the overlay's style with the computed position.
     Object.assign(target.style, {
```
blunteshwar, line 464: For belt-and-suspenders, also guard the final style write
```ts
if (!this.host.open || !this.target) return;
Object.assign(target.style, { /* translate, etc. */ });

1st-gen/packages/overlay/src/PlacementController.ts
blunteshwar, line 189: placeOverlay reassigns this.cleanup without first invoking the previous one. If placeOverlay runs twice before hostUpdated fires the previous cleanup (rapid open/open, programmatic toggles, lifecycle reorder), the first set of listeners — now including the new visualViewport resize/scroll plus a possibly-pending rAF — leaks.

This pattern pre-dates the PR (same is true for cleanupAncestorResize / cleanupElementResize), but the new rAF is more expensive than what was there before, so worth tightening here. Two-line defensive fix at the top of placeOverlay:
```ts
public placeOverlay(/* args */): void {
    this.cleanup?.();
    this.cleanup = undefined;
    // ...rest unchanged
}

blunteshwar (changes requested):

1st-gen/packages/overlay/src/PlacementController.ts
Rajdeepc: Yup good catch. This is a pre-existing double.

1st-gen/packages/overlay/src/PlacementController.ts
Rajdeepc: I have short-circuited the behaviour before the final translate write. Thanks

blunteshwar (approved):
LGTM

PR #6215, opened 2026-04-28
docs(meter): 2nd-gen migration plan

blunteshwar:
## Description

Adds the Phase 1 (Preparation) migration plan for the `<sp-meter>` → `<swc-meter>` 2nd-gen migration at `CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md`, along with companion analysis documents (`accessibility-migration-analysis.md`, `rendering-and-styling-migration-analysis.md`). This is a planning deliverable only — no code, CSS, or component changes are included. The contributor-docs nav script regenerates breadcrumbs and TOC.

The plan codifies the public API, breaking changes, additive items, architecture (mixin / core / SWC / shared-CSS split), migration sequencing, and per-phase checklists for the remaining tickets in the Meter migration epic.

Key decisions documented:

- Align `<swc-meter>` with the [React Spectrum S2 Meter API](https://react-spectrum.adobe.com/Meter.html) and the supplied Figma `S2 / Web (Desktop scale)` Meter frame.
- Rename `progress` → `value`; add `minValue`/`maxValue` (defaults `0`/`100`); replace `side-label` boolean with `label-position` enum (`'top'` default, `'side'`); expose `value-label` and `formatOptions` (`Intl.NumberFormatOptions`, JS property only); normalize the variant set to `{informative (default), positive, notice, negative}`.
- Introduce a thin **`LinearProgressMixin`** in `2nd-gen/packages/core/mixins/` shared with the future `<swc-progress-bar>`. The mixin owns all shared API (value clamping, fill-fraction, locale formatting, slot tracking, DEBUG warning) but is intentionally silent on ARIA role and animation. `MeterBase` extends the mixin and adds only `variant`. Shared bar/track/fill and label-layout CSS lives in `linear-progress-base.css`; meter-specific rules (variant fill colors, description spacing) live in `meter.css`.
- Render `<span class="swc-Meter-label">` / `<span class="swc-Meter-value">` in the shadow root — `<span>`, not `<label>`, because `role="meter"` is not pair-able with native `<label>` semantics.
- Fix the invalid combined `role="meter progressbar"` to a single `role="meter"` placed on the shadow `.swc-Meter` element (not the host); add missing `aria-valuemin`, `aria-valuemax`, and `aria-valuetext`.
- Accessible-name model: visible label via a named **`label`** slot (`aria-labelledby` references the slot container's shadow-internal ID); `accessibleLabel` property (`accessible-label` attribute) as a rare-case fallback for contexts with no visible label (e.g. a data grid); named **`description`** slot for additional text below the meter (`aria-describedby`). "Help text" terminology is not used — meter is not a form field.
- Ship `static-color="black"` and the `description` named slot as net-new from S2; defer `static-color="auto"` (no S2 CSS support yet).

## Motivation and context

This is the first deliverable in the Meter migration epic. The plan is the shared baseline that every later phase (`SWC-2008` setup → `SWC-2016` review) consumes for naming, public API, behavioral semantics, and architecture decisions, per `.claude/skills/migration-prep/references/migration-plan-contract.md`. Aligning with React Spectrum and Figma now avoids a second, more disruptive migration later; introducing `LinearProgressMixin` during this migration means the progress-bar team inherits a tested, reviewed contract without additional breaking changes to consumers; and fixing the 1st-gen ARIA defects (combined role string, missing value attributes, host-level role placement) before consumers depend on the 2nd-gen API avoids a silent a11y regression.

## Related issue(s)

- SWC-2007 (Meter migration planning, child of epic SWC-2005)

## Author's checklist

- [x] I have reviewed at the Accessibility Practices for this feature, see: [Aria Practices](https://www.w3.org/TR/wai-aria-practices/) — the plan addresses the WAI-ARIA `meter` pattern explicitly.
- [ ] I have added automated tests to cover my changes. _(N/A — planning document only; tests ship in `SWC-2013`.)_
- [ ] I have included a well-written changeset if my change needs to be published. _(N/A — internal contributor docs; no published package change.)_

## Manual review test cases

- [ ] _Plan content review_
  1. Open `CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md` on the PR branch.
  2. Read the `TL;DR`, `Changes overview`, `2nd-gen API decisions`, and `Architecture: core vs SWC split` sections.
  3. Confirm the API matches the [React Spectrum S2 Meter](https://react-spectrum.adobe.com/Meter.html) reference and the Figma `S2 / Web (Desktop scale)` Meter frame supplied during planning. Flag any divergence.

- [ ] _Architecture call_
  1. Review the `Migration sequencing and prerequisites` and `Architecture: core vs SWC split` sections.
  2. Confirm the `LinearProgressMixin` design (shared with `<swc-progress-bar>`, silent on ARIA role and animation) is acceptable as the foundation for the progress-bar migration on its own epic.
  3. Flag any objection before `SWC-2008` (Setup) starts.

This PR ships a planning document, not a rendered component. The full accessibility testing checklist applies to `SWC-2010` (Implement accessibility features) and `SWC-2013` (Review and complete test suites). Steps below verify that the documented ARIA recommendations are coherent.

- [ ] **Screen reader** — verify the documented ARIA contract reads cleanly
  1. Read the `2nd-gen API decisions → Accessibility semantics notes` section. Confirm the documented `role="meter"` on the shadow `.swc-Meter` element, the `aria-valuemin`/`aria-valuemax`/`aria-valuenow`/`aria-valuetext` bindings, the `aria-labelledby`/`aria-label` mutual-exclusion model, and the `aria-describedby` wiring match the [WAI-ARIA 1.2 meter](https://www.w3.org/TR/wai-aria-1.2/#meter) spec and the [APG meter pattern](https://www.w3.org/WAI/ARIA/apg/patterns/meter/).
  2. Confirm the documented dev-mode accessible-name warning (fires when neither the `label` slot has assigned nodes nor `accessibleLabel` is set) is coherent with the DEBUG warning behavior described in `Architecture: core vs SWC split`.


CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
```diff
@@ -0,0 +1,526 @@
...
+
+### Dependency-aware recommendation
+
+`<swc-meter>` is migrated **independently** of `<swc-progress-bar>`. No shared base is extracted in core.
+
+Rationale:
+
+- 1st-gen `<sp-meter>` extended progress-bar visuals via CSS `@import` of `spectrum-progress-bar.css` and `progress-bar-overrides.css`. In 2nd-gen, this is replaced with a self-contained `meter.css` that copies the relevant bar/track/fill rules from `spectrum-css` `spectrum-two` ([`components/progressbar/index.css`](https://github.com/adobe/spectrum-css/blob/spectrum-two/components/progressbar/index.css)) plus the meter-specific rules from [`components/meter/index.css`](https://github.com/adobe/spectrum-css/blob/spectrum-two/components/meter/index.css).
+- The 1st-gen progress-bar package is not migrated yet; making meter wait on or share with it would couple two epics unnecessarily and slow SWC-2005.
+- The components have different ARIA semantics (`role="meter"` vs `role="progressbar"`), different default behaviors (meter is determinate-only; progress-bar can be indeterminate), and different value semantics. A shared base would have to gate most of its API per subclass anyway.
```
5t3ph: I would challenge this as a reason to not have a shared base in core. The ARIA role would be on the rendering layer and not shared in core anyway. Styles would be fine to remain separate, at least for now.

Many of the exposed consumer attributes are the same, and the ones that are different can be reserved and added via the rendering layer. This is how we are approaching Button as well, for reference (you could have your agent look at that migration plan for comparison).

The lingering issue may be deciding how to name it in core - maybe `progress-meter`?

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
```diff
@@ -0,0 +1,526 @@
...
+| **B6** | `--mod-*` passthroughs removed (`--mod-progressbar-*`, `--mod-meter-*`). _(Source: [CSS style guide — custom properties](../../../../CONTRIBUTOR-DOCS/02_style-guide/01_css/02_custom-properties.md#component-custom-property-exposure).)_ | Consumers customize via `--mod-*`. | Customize via a small reviewed set of `--swc-meter-*` tokens. | Replace `--mod-progressbar-fill-color`, `--mod-progressbar-thickness`, `--mod-meter-min-width`, `--mod-meter-max-width`, `--mod-meter-help-text-to-progress-bar` with the corresponding `--swc-meter-*` properties (final list locked in implementation). |
+| **B7** | CSS class wrapper. _(Source: [CSS style guide — selector conventions](../../../../CONTRIBUTOR-DOCS/02_style-guide/01_css/02_custom-properties.md#selector-conventions); Badge precedent.)_ | Styling applied off shadow root and host. | Internal `<div class="swc-Meter">` wrapper inside the shadow root; styling does not target `:host`. | Consumers piercing shadow DOM update selectors to `.swc-Meter` and its descendants (rare; shadow DOM is not a stable styling boundary). |
```
5t3ph: I've been having folks remove this as a "breaking change" because as it already notes our internal DOM is _not_ a consumer concern.

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
```diff
@@ -0,0 +1,526 @@
...
+| `variant`       | `'informative' \| 'positive' \| 'notice' \| 'negative'`       | `'informative'` | `variant`       | **Confirmed.** Reflected. Default is the named `informative` value (1st-gen used the empty string for this case). |
+| `labelPosition` | `'top' \| 'side'`                                             | `'top'`       | `label-position`  | **Confirmed.** Reflected. Replaces 1st-gen `side-label` boolean. |
+| `staticColor`   | `'white' \| 'black' \| undefined`                             | `undefined`   | `static-color`    | **Confirmed.** Reflected. `'auto'` from React Spectrum is **deferred** — `spectrum-css` `spectrum-two` does not include a `staticAuto` modifier. |
+| `size`          | `'s' \| 'm' \| 'l' \| 'xl'`                                   | `'m'`         | `size`            | **Confirmed.** Default `m` per Figma `S2 / Web (Desktop scale)` Properties panel and React Spectrum default. |
+
+#### Visual matrix (2nd-gen)
+
+Confirmed against the Figma `S2 / Web (Desktop scale)` Meter primary frame supplied with this plan.
+
+| Variant / style | Top label | Side label | 0% | 25% | 50% | 75% | 100% | Help text |
```
5t3ph: Since there is not variance among the visual variants (they all support the other visual features) you could safely drop this table.

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
```diff
@@ -0,0 +1,526 @@
...
+Sizes: `s`, `m` (default), `l`, `xl` — confirmed across all variants and static-color modes per Figma.
+
+`label-position="side"` is exposed via `spectrum-css` `spectrum-two` `progressbar/index.css` `.spectrum-ProgressBar--sideLabel` (line 144), even though the Figma Desktop frame focuses on top-label. Side-label coverage in implementation/tests is required for parity with React Spectrum.
+
+#### Slots (2nd-gen)
+
+| Slot          | Content              | Notes |
+| ------------- | -------------------- | ----- |
+| (default)     | Visible meter label  | **Confirmed.** Slot text hoists into `label` property via `slotchange` + `getLabelFromSlot`. Single naming path. |
+| `value-label` | Custom value content | **Confirmed.** Optional. Renders inside the percent label region. Used for richer value displays (e.g. `<strong>1</strong> of 4`). When the slot is empty and `value-label` attribute is unset, the auto-formatted value is rendered. |
```
5t3ph: This would probably be safer as an attribute. I know you show an example of passing formatting like use of `<strong>` but at least in terms of matching React, we just need to support a string.

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
```diff
@@ -0,0 +1,526 @@
...
+- **Locale-aware formatting.** When `value-label` is unset, format the visible value and `aria-valuetext` with `new Intl.NumberFormat(language, formatOptions ?? { style: 'percent' })`. For percent formatting, feed the normalized fraction `(value − minValue) / (maxValue − minValue)`. Re-format on language-resolver updates and on any of `value`, `minValue`, `maxValue`, `valueLabel`, `formatOptions` changing.
+- **`value-label` precedence.** Non-empty `value-label` attribute or non-empty `slot="value-label"` overrides auto-formatted text in both the visible value cell and `aria-valuetext`. Slot wins over attribute when both are present.
+- **Variant validation.** `variant` is a typed enum; unknown values fall back to `'informative'`. Tracked in `Meter.types.ts` rather than via the 1st-gen string-coercing setter.
+- **No custom events.** No `dispatchEvent` calls; behavior parity with 1st-gen.
+
+### Accessibility semantics notes (2nd-gen)
+
+Sourced from [`accessibility-migration-analysis.md`](./accessibility-migration-analysis.md) and the React Spectrum S2 Meter API:
+
+- `role="meter"` set in `firstUpdated` if not already set. Fixed; not author-overridable.
```
5t3ph: This can be simplified as it should be part of the render template permanently.

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
```diff
@@ -0,0 +1,526 @@
...
+- `aria-valuemin=<minValue>`, `aria-valuemax=<maxValue>`, `aria-valuenow=<value>` (clamped), `aria-valuetext=<formatted value>` set on host. Update on every relevant property change and on locale change.
+- Accessible name from any of: `label` attribute, default slot text, `aria-label`, `aria-labelledby`. `aria-describedby` and `aria-details` are author-managed pass-throughs (per React Spectrum). When `slot="help-text"` is non-empty, the meter wires `aria-describedby` to the slotted node's id (or generates an internal id).
```
5t3ph: @nikkimk how do you feel about the idea of these aria passthroughs? Should this simplify and support `accessible-label` and then take the default slot content as the primary "label" instead of an attribute?

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
```diff
@@ -0,0 +1,526 @@
...
+  <div class="spectrum-ProgressBar-helptext">
+    <slot name="help-text"></slot>
+  </div>
+</div>
+```
+
+Notes:
+
+- `<fillPercent>` is `((value − minValue) / (maxValue − minValue)) * 100`, clamped to `[0, 100]`.
+- The `helptext` wrapper renders unconditionally; CSS hides it when no slotted content is assigned (use `:has(slot[name="help-text"]:not(:empty))` or a slot-content observer — locked in implementation).
```
5t3ph: `:has()` will not successfully work here. Instead of the div wrapper around the slot, you can style the content by way of `::slotted([slot="help-text"])` which will affect the content when it exists. When the slot is not filled, the default browser style of `display: contents` on the `<slot>` will keep it hidden/prevent it affecting layout.

5t3ph: ^ which doesn't need explicitly noted in the migration plan, but just for your reference!

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
```diff
@@ -0,0 +1,526 @@
...
+- The percent `<label>` is hidden by CSS when `static-color` is set (matching Figma `Static white` / `Static black` panels which show no value text).
```
5t3ph: I think this was to simplify the Figma and isn't meant to be taken as a valid state. All features for labeling should be supported by all visual styles, just like is implemented in React and available in Spectrum CSS.

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
```diff
@@ -0,0 +1,526 @@
...
+- Class names match `spectrum-css` `spectrum-two` HTML output exactly so the imported S2 CSS rules apply without rewriting selectors.
```
5t3ph: This is incorrect - they should update to use the `swc` prefix and also follow the [contributor docs guidelines for selectors](https://github.com/adobe/spectrum-web-components/blob/main/CONTRIBUTOR-DOCS/02_style-guide/01_css/01_component-css.md#selector-patterns).

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
```diff
@@ -0,0 +1,526 @@
...
+- [ ] `Meter.base.ts`: when `slot="help-text"` is non-empty, set `aria-describedby` on the host pointing at the slotted node (or an internal id assigned to it)
+- [ ] `Meter.base.ts`: emit DEBUG-mode warning when no accessible name is provable, mirroring `ProgressCircleBase`
+
+#### Alignment checks
+
+- [ ] `variant` set matches React Spectrum S2 (`informative`, `positive`, `notice`, `negative`) and Figma `S2 / Web (Desktop scale)` Variant panel
+- [ ] `staticColor` set matches `spectrum-css` `spectrum-two` (`staticWhite`, `staticBlack`) and Figma `Static white` / `Static black` panels; React's `'auto'` is deferred (no S2 CSS support)
+- [ ] `labelPosition` matches React Spectrum (`top` default, `side`); side-label coverage backed by `spectrum-css` `spectrum-two` `progressbar/index.css` `.spectrum-ProgressBar--sideLabel`
+- [ ] `value-label` slot + attribute matches React Spectrum `valueLabel`
+- [ ] `format-options` matches React Spectrum `formatOptions` (property only — no string serialization)
```
5t3ph: The "unit" option from React technically accepts multiple items as an object:

```js
{style: "decimal"}
{style: "percent"}
{style: "unit", unit: "inch"}
{style: "currency", currency: "USD"}
```

To my awareness, setting this via an object would require the consumer to use JS to set the attribute rather than pass an object-styled string to the DOM attribute. Do we want to force that, or should we have a series of attributes like `format-style`, `format-unit`, `format-currency`?

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
```diff
@@ -0,0 +1,526 @@
...
+
+- [ ] Render the internal wrapper as `<div class="swc-Meter">`; do not target `:host` for component visuals
+- [ ] Copy S2 source from `spectrum-css` `spectrum-two` branch — `components/meter/index.css` plus the relevant rules from `components/progressbar/index.css` (the meter file is mostly progress-bar passthroughs and contains no bar visuals on its own) — into `meter.css` as the baseline
+- [ ] Strip all `--mod-*` properties; replace with the reviewed `--swc-meter-*` set defined in [2nd-gen API decisions](#2nd-gen-api-decisions)
+- [ ] Variant fill colors via tokens (`positive-visual-color`, `notice-visual-color`, `negative-visual-color`; `accent-content-color-default` for the `informative` default)
+- [ ] Static-color rules for both `staticWhite` and `staticBlack` modifiers
+- [ ] `label-position="side"` layout rule (`.swc-Meter--sideLabel`) mirroring `spectrum-css` `spectrum-two` `.spectrum-ProgressBar--sideLabel`
+- [ ] Help-text spacing rule (`--swc-meter-help-text-spacing`) and visibility gating when `slot="help-text"` is empty
+- [ ] Hide the percent label cell when `static-color` is set (matches Figma `Static white` / `Static black` panels)
```
5t3ph: Reminder not to do this, remove this item.

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
```diff
@@ -0,0 +1,526 @@
...
+
+- [ ] Verify i18n size modifiers (`:lang(ja)`, `:lang(ko)`, `:lang(zh)`) if present in S2 source
+- [ ] Pass stylelint (property order, `no-descending-specificity`, token validation)
+- [ ] Forced-colors / high-contrast block at the bottom of the file
+
+### Accessibility
+
+#### Naming and semantics
+
+- [ ] Single `role="meter"` on host. No combined role string.
```
5t3ph:
```suggestion
- [ ] Single `role="meter"` on `.swc-Meter` container. No combined role string.
```

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
```diff
@@ -0,0 +1,526 @@
...
+- [ ] `aria-valuemin=<minValue>`, `aria-valuemax=<maxValue>`, `aria-valuenow=<clamped value>`, `aria-valuetext=<formatted value>` on host
```
5t3ph:
```suggestion
- [ ] `aria-valuemin=<minValue>`, `aria-valuemax=<maxValue>`, `aria-valuenow=<clamped value>`, `aria-valuetext=<formatted value>` on `.swc-Meter` container
```

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
```diff
@@ -0,0 +1,526 @@
...
+- [ ] Accessible name from `label`, default slot text, `aria-label`, or `aria-labelledby`
```
5t3ph: Update pending @nikkimk feedback on earlier comment

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
```diff
@@ -0,0 +1,526 @@
...
+- [ ] PR created with description referencing Epic SWC-2005
+- [ ] Peer engineer sign-off
+
+---
+
+## Blockers and open questions
+
+All drafting-time questions are resolved. Resolutions:
+
+- **Q1 (architecture)** — Closed. Independent `<swc-meter>` with no shared base. Bar styles live in `meter.css`. Reflected in [Architecture: core vs SWC split](#architecture-core-vs-swc-split) and [Migration sequencing and prerequisites](#migration-sequencing-and-prerequisites).
```
5t3ph: Noting that the shared base idea should be re-opened, see earlier comment.

5t3ph (commented):
Let me know if you need clarification on any of these!

The most impactful is to re-consider having a shared base between meter and progress bar.

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
nikkimk: @5t3ph @Piyu34 At [RedHat](https://patternflyelements.org/components/progress/) we'd use custom properties and then set them as [`ElementInternals`](https://developer.mozilla.org/en-US/docs/Web/API/ElementInternals)

We did this because not all ARIA states properties and attributes may live on the host (even though in this case, the host owns the role and so they do). It provided consistency across all components. So the API could be consistent by setting the same custom properties and letting the component to decide where and how to apply them. 

It might be worth bringing this up at a sync to get consensus.

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
nikkimk: Could this be done as a mixin or controller?

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
nikkimk: > This can be simplified as it should be part of the render template permanently.

@5t3ph If the role is on an element in the shadow DOM we will not be able to use IDREFs like `aria-labelledby`, although eventually we will have a corss-root ARIA solution.

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
nikkimk: > The "unit" option from React technically accepts multiple items as an object:
> 
> ```js
> {style: "decimal"}
> {style: "percent"}
> {style: "unit", unit: "inch"}
> {style: "currency", currency: "USD"}
> ```
> 
> To my awareness, setting this via an object would require the consumer to use JS to set the attribute rather than pass an object-styled string to the DOM attribute. Do we want to force that, or should we have a series of attributes like `format-style`, `format-unit`, `format-currency`?

Could we flatten them? `format`: `decimal`| `percent` |`unit-inch`| `currency-USD` ?

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
5t3ph: disregard for now, pending aria discussion

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
5t3ph: disregard for now, pending aria discussion

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
blunteshwar: Thanks for pushing on reuse — we’re aligned that role and styling living in SWC don’t rule out pulling out small shared pieces later (localized formatting, normalized fill fraction, label-from-slot) via helpers, a mixin, or a controller if it stays cheap to maintain.

I'd still gently suggest treating meter and progress-bar as separate semantic cores, not one shared base type, mainly because the parity numbers don’t look like Button vs action-button:

Properties (1st gen): 5 of 8 line up on both (progress, label, side-label, static-color, size). 3 of 8 are only on one (variant on meter; indeterminate and over-background on progress-bar), so cross-component duplication is already partial.
Slots: Same default label slot/hoisting pattern (1/1) in gen 1 — good candidate for helpers, less so for inheritance.
Role / ARIA: Only naming via label → aria-label lines up cleanly; role, aria-valuemin/max, aria-valuetext, and aria-valuenow when indeterminate don’t align with meter’s shape today — that mismatch grows once meter picks up value/min/max and progress-bar locks its own gen‑2 API.
Templates: Visibility rules for the label chrome and the percent/value row differ materially; shared bar markup is mechanical, not a shared semantic contract like Button’s family.
Button/action-button deliberately share one interactive primitive and a roadmap of sibling components on the same semantic contract. Here we have different roles/use cases, uneven property overlap, and different timelines — so I'd suggest keeping MeterBase meter-scoped, and optional shared utilities once progress-bar’s gen‑2 design is settled, rather than coupling this epic on a hypothetical *-base. Happy to revise if we later see a firmer convergence on API and semantics.

blunteshwar: In 1st gen they already don’t share a TypeScript base class — they’re separate components; only styling is shared

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
blunteshwar: Done!

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
blunteshwar: Noted!

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
blunteshwar: Makes Sense!

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
blunteshwar: Going with role + aria-value* on host (not on the shadow `.swc-Meter` wrapper). This keeps consumer IDREFs (`aria-labelledby`, `aria-describedby`) working — light-DOM IDs can't resolve into shadow DOM today. Cross-root ARIA (Reference Target) is not yet shipped in stable browsers, so host placement is the only safe option right now.

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
blunteshwar: Going with @5t3ph's simplified API for now: drop the four raw aria-* passthroughs (`aria-label`, `aria-labelledby`, `aria-describedby`, `aria-details`) and the `label` attribute. Keep two inputs only — default slot as the primary visible label, and `accessible-label` as the a11y-only fallback when there's no visible text. `help-text` slot continues to wire `aria-describedby` for description.
Treating this as interim direction until the team sync lands on @nikkimk's `ElementInternals` + shared a11y mixin proposal — that decision affects every SWC component, not just meter, so it makes sense to align there before locking. Plan updated; will revisit `MeterBase` accordingly once the sync resolves.

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
blunteshwar: Good catch on the `:has()` not working here — switching to `::slotted([slot="help-text"])` for content styling, and relying on the `<slot>` default `display: contents` to keep an empty slot inert without needing extra JS. Plan updated.

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
blunteshwar: You're right — the Figma omission is for visual simplicity, not a spec constraint. React Spectrum and Spectrum CSS both support full labeling under static-color. Removed the "hide percent label when static-color is set" rule from the architecture render notes and the styling checklist. Static-color variants now ship with full label + value + help-text parity.

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
blunteshwar: Yeah this one's on me — Claude went full autopilot on the render template and just kept the `spectrum-ProgressBar-*` classes from the S2 source like we were editing Spectrum CSS directly. Pointed it at the selector patterns guide and reworked the whole template to use the SWC namespace. CSS still gets ported from `spectrum-css` `spectrum-two`, just with selectors rewritten and tokens preserved. Plan updated.

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
blunteshwar: Going with @nikkimk's flat shorthand. Single `format` string attribute, default `"percent"`. Syntax: `<style>` or `<style>-<arg>`:
- `format="decimal"`
- `format="percent"` (default)
- `format="currency-USD"` → `{ style: 'currency', currency: 'USD' }`
- `format="unit-inch"` → `{ style: 'unit', unit: 'inch' }`
- `format="unit-kilometer-per-hour"` → `{ style: 'unit', unit: 'kilometer-per-hour' }`
`MeterBase` parses on first `-`, splits style + arg, builds the `Intl.NumberFormatOptions` internally, feeds `Intl.NumberFormat(language, opts)`. Works in plain HTML / Lit / Vue / SSR — no JS object juggling. Storybook controls become a simple text/dropdown input.
Tradeoff: doesn't cover the full `Intl.NumberFormatOptions` surface (e.g. `notation: 'compact'`, `roundingMode`, `maximumFractionDigits`). Logging that as a deferred follow-up — if real consumer demand shows up.

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
```diff
@@ -0,0 +1,526 @@
...
+
+### Related components and ordering notes
+
+- **Progress bar** ([`1st-gen/packages/progress-bar`](../../../../1st-gen/packages/progress-bar/)) — independent migration on its own epic. No coupling to this work.
+- **Progress circle** ([`2nd-gen/packages/swc/components/progress-circle`](../../../../2nd-gen/packages/swc/components/progress-circle/)) — already migrated. Used as the structural precedent for `aria-value*` plumbing, slot-as-label hoisting, locale-aware formatting, and the dev-mode accessible-name warning. Not extended.
+- **Field label** — internal render dependency; not migrated. 2nd-gen `<swc-meter>` renders plain `<label class="spectrum-FieldLabel ...">` matching `spectrum-css` `spectrum-two` HTML output. No dependency on `<swc-field-label>`.
+- **Help text** — 1st-gen `<sp-help-text>` package is not migrated. The `help-text` slot in `<swc-meter>` accepts any DOM, so consumers can place 1st-gen `<sp-help-text>` or any composed help-text element until `<swc-help-text>` exists.
```
rubencarvalho: Should this be a string attribute instead? I'm not sure we need a slot here...

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
```diff
@@ -0,0 +1,526 @@
...
+
+`label-position="side"` is exposed via `spectrum-css` `spectrum-two` `progressbar/index.css` `.spectrum-ProgressBar--sideLabel` (line 144), even though the Figma Desktop frame focuses on top-label. Side-label coverage in implementation/tests is required for parity with React Spectrum.
+
+#### Slots (2nd-gen)
+
+| Slot          | Content              | Notes |
+| ------------- | -------------------- | ----- |
+| (default)     | Visible meter label  | **Confirmed.** Slot text hoists into `label` property via `slotchange` + `getLabelFromSlot`. Single naming path. |
+| `value-label` | Custom value content | **Confirmed.** Optional. Renders inside the percent label region. Used for richer value displays (e.g. `<strong>1</strong> of 4`). When the slot is empty and `value-label` attribute is unset, the auto-formatted value is rendered. |
+| `help-text`   | Optional help text   | **Confirmed.** Renders inside the meter's S2 `helptext` region (per `spectrum-css` `spectrum-two` HTML output). Consumers compose `<sp-help-text>` (1st-gen, until `<swc-help-text>` exists) or any DOM. Wire to `aria-describedby` if the slot is non-empty. |
```
rubencarvalho: same comment as above - I think we can simplify this with a property.

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
rubencarvalho: React forwards these options to [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#options). If we want this parity we probably shouldn’t try to flatten or remap everything the native API supports today (and may support in the future).
I think the cleanest option is to expose this as a JavaScript property only, I don't see a very scalable path trying to model it as attributes.

Also, we'll need to hook this up to our `LanguageResolutionController` (which now that I think about it, should actually be named `LocaleResolutionController` 😛

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
rubencarvalho: Interested in this outcome too! My first impression is 2 bases with the shared mixin sounds like a reasonable approach.

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
rubencarvalho: Is there any difference in the APIs?

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
5t3ph: I _could_ see folks trying to add like a link, or an icon/tooltip. But, maybe we're comfortable with those being a reason they might have to extend the component.

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
blunteshwar: Makes sense

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
blunteshwar: On it

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
blunteshwar: Okay

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
5t3ph: This is valid again since you are taking aria off the host, per other comment discussion.

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
5t3ph: This is valid again since you are taking aria off the host, per other comment discussion.

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
```diff
@@ -0,0 +1,516 @@
...
+- The components have different ARIA semantics (`role="meter"` vs `role="progressbar"`), different default behaviors (meter is determinate-only; progress-bar can be indeterminate), and different value semantics. A shared base would have to gate most of its API per subclass anyway.
+
+### Related components and ordering notes
+
+- **Progress bar** ([`1st-gen/packages/progress-bar`](../../../../1st-gen/packages/progress-bar/)) — independent migration on its own epic. No coupling to this work.
+- **Progress circle** ([`2nd-gen/packages/swc/components/progress-circle`](../../../../2nd-gen/packages/swc/components/progress-circle/)) — already migrated. Used as the structural precedent for `aria-value*` plumbing, slot-as-label hoisting, locale-aware formatting, and the dev-mode accessible-name warning. Not extended.
+- **Field label** — internal render dependency; not migrated. 2nd-gen `<swc-meter>` renders plain `<label class="swc-Meter-label">` / `<label class="swc-Meter-value">` (SWC-namespaced selectors per the [contributor docs selector patterns](../../../../CONTRIBUTOR-DOCS/02_style-guide/01_css/01_component-css.md#selector-patterns)). No dependency on `<swc-field-label>`.
```
5t3ph: Just validated that use of `role="meter"` does _not_ make it eligible to be attached to `<label>` - it would only work if we were going to [use native meter](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/meter).

Image

Since you will be creating the meter with aria attributes to support custom styling, it is semantically meaningless to use `<label>` and instead a span could be used given the plan is to use `aria-labelledby` to associate it to the meter.

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
```diff
@@ -0,0 +1,516 @@
...
+- **Help text** — exposed as a `help-text` string attribute on `<swc-meter>` (not a slot). Renders as plain text inside the internal `swc-Meter-helptext` region. Consumers needing rich/interactive help-text content compose externally and wire `aria-describedby` themselves.
```
5t3ph: In the current plan to move ARIA internally, consumers will _not_ be able to successfully use `aria-describedby` for external help text due to cross-root ARIA.

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
```diff
@@ -0,0 +1,516 @@
...
+- **Variant validation.** `variant` is a typed enum; unknown values fall back to `'informative'`. Tracked in `Meter.types.ts` rather than via the 1st-gen string-coercing setter.
+- **No custom events.** No `dispatchEvent` calls; behavior parity with 1st-gen.
+
+### Accessibility semantics notes (2nd-gen)
+
+Sourced from [`accessibility-migration-analysis.md`](./accessibility-migration-analysis.md) and the React Spectrum S2 Meter API:
+
+- `role="meter"` set on the host in `firstUpdated`. Always set (no "if not already set" conditional). Fixed; not author-overridable.
+- `aria-valuemin=<minValue>`, `aria-valuemax=<maxValue>`, `aria-valuenow=<value>` (clamped), `aria-valuetext=<formatted value>` set on host. Update on every relevant property change and on locale change.
+- **Placement rationale.** `role` + `aria-value*` live on the host (not on the shadow `.swc-Meter` wrapper) so consumer-supplied IDREFs bind correctly. Cross-root ARIA (Reference Target / cross-root ARIA spec) is not yet shipped in stable browsers.
```
5t3ph: These need updated as well with the decision to move aria internal. Essentially, nothing will be set on the host after all.

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
```diff
@@ -0,0 +1,516 @@
...
+
+All variants (`informative` default, `positive`, `notice`, `negative`) support all sizes (`s`, `m` default, `l`, `xl`), both `label-position` values (`top` default, `side`), both `static-color` values (`white`, `black`), the full `value` range (0–100, including 0%/25%/50%/75%/100%), and the `help-text` attribute. There are no variant-restricted features.
+
+`label-position="side"` is exposed via `spectrum-css` `spectrum-two` `progressbar/index.css` `.spectrum-ProgressBar--sideLabel` (line 144), even though the Figma Desktop frame focuses on top-label. Side-label coverage in implementation/tests is required for parity with React Spectrum.
+
+#### Slots (2nd-gen)
+
+| Slot          | Content              | Notes |
+| ------------- | -------------------- | ----- |
+| (default)     | Primary visible meter label | **Confirmed.** Default slot is the **primary** visible label and primary a11y name source. Slot text hoists into `accessibleLabel` via `slotchange` + `getLabelFromSlot` only when `accessibleLabel` is unset. |
```
5t3ph: This reads to me as if you will be forcing the default slot text into `aria-label` which wouldn't be correct since it would be intended to be a visible label. Am I misinterpreting this?

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
```diff
@@ -0,0 +1,516 @@
...
+```html
+<div
+  class="swc-Meter swc-Meter--size<S> swc-Meter--<variant> [swc-Meter--sideLabel] [swc-Meter--staticWhite|swc-Meter--staticBlack]"
+>
+  <label class="swc-Meter-label">
+    <slot @slotchange=...>{accessibleLabel}</slot>
+  </label>
+  <label class="swc-Meter-value">
+    {valueLabel ?? auto-formatted value}
+  </label>
```
5t3ph: Explained in earlier comment, but these should be simple `<span>` elements

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
```diff
@@ -0,0 +1,516 @@
...
+  <div class="swc-Meter-track">
+    <div class="swc-Meter-fill" style="inline-size: <fillPercent>%;"></div>
+  </div>
+  <div class="swc-Meter-helptext" id="<internal-id>" hidden=${!helpText}>
+    {helpText}
+  </div>
```
5t3ph: Rather than toggling the `hidden` attribute, you can conditionally prevent this entire block from showing if there isn't help text. True for the other labeling parts as well.

5t3ph (commented):
I would suggest making these updates, then asking the agent to run a consistency pass based on your staged changes before committing as I didn't mark every instance of where these changes impact the plan. You should see changes throughout the checklists as well.

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
rise-erpelding: Would love to chat about this as I'm currently assigned to progress bar!

nikkimk (changes requested):
Based on our initiative leads discussion:
- Use a `label` slot for visible label
- Use `accessibleLabel` property to provide a label for screen readers in rare cases where there is no visible label, for example a data grid that has meters.
- Use a `description` slot for additional text below the meter. Help text implies meter is a form field but it is not.
- The `meter` role should be in the shadow DOM not the host itself.
   - if an `accessibleLabel` is provided, the element with this role will set this as its `aria-label`
- The `label` slot should be have an id reference on its container in the shadow DOM, so that the element with the `meter` role can ID reference it for its `aria-labelledby`
- The `description` slot should be have an id reference on its container in the shadow DOM, so that the element with the `meter` role can ID reference it for its `aria-describedby`

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
blunteshwar: Confirmed — switched all label containers in the render shape (label slot, value, description) from `<label>` to `<span>`. The `meter` role element associates via `aria-labelledby` to the label `<span>` container's internal id rather than relying on `<label>` pairing.

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
blunteshwar: Noted

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
blunteshwar: You read it right and it was wrong — the slot table previously implied hoisting default-slot text into `aria-label`. That's gone. Visible label and accessible name are now distinct paths: the `label` slot is the visible label (no copy into `aria-label`), and `aria-labelledby` on the role element references its container. `accessibleLabel` only applies when the `label` slot has no content.

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
blunteshwar: Noted

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
blunteshwar: Switched. The description container block is now conditionally rendered (Lit `?` directive) — the `<span class="swc-Meter-description">` is only present in the DOM when the `description` slot has assigned nodes. Same approach extended to the `aria-labelledby` / `aria-label` / `aria-describedby` attributes themselves (only emitted when their source resolves). No `hidden` attribute toggling anywhere.

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
5t3ph: Change needed: it was resolved in a sync meeting to go ahead and use a thin mixin for shared Meter and Progress Bar functionality

5t3ph: Plan updated to reflect use of a `LinearProgress` mixin - closing this comment thread in favor of more scoped comments if needed.

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/migration-plan.md
5t3ph: See plan updates to reflect creation of `LinearProgressMixin` and shared styles

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/accessibility-migration-analysis.md
```diff
@@ -81,13 +82,13 @@ This doc explains how **`swc-meter`** should work for **accessibility**. It supp
...
-| [Non-text content (WCAG 1.1.1)](https://www.w3.org/WAI/WCAG22/Understanding/non-text-content.html) | **Track** and **fill** are in **shadow DOM**; the **host** should expose **role** and **ARIA** values. Inner nodes stay **graphical** unless a separate spec says otherwise. |
+| [Name, role, value (WCAG 4.1.2)](https://www.w3.org/WAI/WCAG22/#name-role-value) | **Name** from the **label slot** via **`aria-labelledby`**, or from **`accessibleLabel`** via **`aria-label`** when no visible label is present. **Role** must be **`meter`**, not **`progressbar`**, for this component. **Value** = min, max, now, and (when used) valuetext. |
+| [Non-text content (WCAG 1.1.1)](https://www.w3.org/WAI/WCAG22/Understanding/non-text-content.html) | **Track** and **fill** are in **shadow DOM**. The element with **`role="meter"`** is also in the **shadow DOM** with internal ID references for labeling; the **host** carries no ARIA **role**. Inner nodes stay **graphical** unless a separate spec says otherwise. |
```
nikkimk, line 86: <3

CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/accessibility-migration-analysis.md
```diff
@@ -105,28 +106,60 @@ This doc explains how **`swc-meter`** should work for **accessibility**. It supp
...
+  aria-labelledby="label"
+  aria-describedby="description"
+  aria-valuemin="0"
+  aria-valuemax="100"
+  aria-valuenow="[progress]"
+  aria-valuetext="[localized percent]"
+>
+  <div class="track"><div class="fill"></div></div>
+</div>
+<div id="description"><slot name="description"></slot></div>
```
nikkimk, line 142: Thanks for adding this!

nikkimk (approved):
Thanks for updating this!

rise-erpelding (approved):
Looks great!

PR #6226, opened 2026-04-30
feat(button): full s2 visual fidelity

5t3ph:
## Description

> ⚠️ **Scope note**: Storybook and documentation refinement are part of a later phase. This includes just enough stories to validate the styles.

Phase 4 (styling) for the 2nd-gen `<swc-button>` migration. Implements Spectrum 2 CSS, finalizes the base class API, and adds a Phase 4 Storybook scaffold.

**`Button.base.ts`**
- Fixed `delegatesFocus` via `shadowRootOptions` instead of `createRenderRoot()` override (the override silently bypassed Lit's `adoptStyles()`, dropping all CSS)
- Added `ObserveSlotText` and `ObserveSlotPresence` mixins to derive `hasLabel` / `hasIcon` from slot content without a consumer-settable attribute
- Added `accessible-label` property (replaces the previous `aria-label` forwarding pattern) with a debug warning when an icon-only button omits it
- Added `_pendingActive` state with a 1-second delay timer, so transient operations don't flash the unavailable appearance

**`button.css` / `button-base.css`**
- Full S2 token implementation across all 4 sizes (s / m / l / xl), 4 semantic variants (primary / secondary / accent / negative), outline fill-style, and static-color (white / black)
- Pending state: cursor changes immediately on `[pending]`; disabled colors and spinner defer to `.swc-Button--pendingActive` after 1 s
- Icon-only and truncate layout via derived class modifiers
- `prefers-reduced-motion` override at the bottom of the file
- Subcomponent classes corrected to single-hyphen convention (`swc-Button-label`, `swc-Button-pendingSpinner`) consistent with all other 2nd-gen components

**`Button.ts`** — variant/fillStyle/staticColor/truncate properties; debug validation for invalid combinations; complete `@cssprop` JSDoc for all 26 exposed custom properties

## Related issue(s)

- fixes SWC-1879

> **Note**: Global button element reconciliation is split to a follow-up branch for reviewer ease.

image

---

## Manual review test cases

- [x] **Sizes**
    - Confirm s / m / l / xl all render at visually distinct sizes with correct pill border-radius and padding.
- [x] **Variants**
    - Confirm primary (neutral), secondary, accent (blue), and negative (red) each render with correct fill, content color, and hover/focus/active states.
- [x] **Outline**
    - Confirm transparent background and visible border for both primary and secondary
    - Verify hover adds a subtle fill.
    - Confirm accent and negative do not expose an outline option.
- [x] **Static color**
    - Confirm `static-color="white"` renders legibly on the dark half of the demo background; `static-color="black"` on the light half.
    - Check the secondary and outline combinations within each static-color.
- [x] **Icon-only**
    - Confirm the button renders as a circle (pill radius) with centered icon.
- [x] **Pending state**
    - Set `pending` on a button and wait ~1 second.
    - Confirm: cursor changes immediately, then after the delay the label is hidden and the spinner appears, while the button dimensions stay the same.
    - Remove `pending` and confirm the button returns to its normal size and appearance instantly.
- [x] **Truncate**
    - Confirm the label clips with an ellipsis rather than wrapping.
- [x] **Disabled**
    - Confirm the button uses disabled token colors and the cursor is `default`.
    - Confirm outline and static-color variants each get their correct disabled appearance.
- [x] **Focus ring**
    - Tab to the button.
    - Confirm a `focus-visible` outline appears (blue by default, white on `static-color="black"`, black on `static-color="white"`).
- [x] **Down state**
    - Confirm buttons scale down while clicked/pressed (in `:active` state)
- [x] **Themes and Scales**
    - Confirm switching to the "Dark" theme causes all color tokens to adapt.
    - Confirm "Dark" token values do not create contrast issues.
    - Confirm switching to "Large" scale increases the size of all variants.

### Device review

- [ ] Did it pass in Desktop?

- [ ] **Keyboard**
  1. Open the Playground story in Storybook
  2. Tab to `<swc-button>` — confirm focus lands on the button (delegatesFocus)
  3. Tab to a disabled button — confirm it is not reachable
  4. Tab to a pending button — confirm it is reachable but `aria-disabled="true"` is announced

- [ ] **Screen reader**
  1. Navigate to an icon-only button — confirm `aria-label` provides the accessible name
  2. Via Playground: Set `pending` and wait for the spinner — confirm the pending accessible name is announced (not the spinner SVG)

- [ ] **`prefers-reduced-motion`** — In DevTools, enable the reduced-motion emulation. Confirm the pending spinner still appears but animates at a very slow, linear pace rather than the default easing.


2nd-gen/packages/swc/components/button/button.css
```diff
@@ -10,8 +10,432 @@
...
+  color: var(--swc-button-content-color-focus, token("gray-25"));
+  background-color: var(--swc-button-background-color-focus, token("neutral-background-color-key-focus"));
+  border-color: var(--swc-button-border-color-focus, transparent);
+}
+
+.swc-Button:active {
+  color: var(--swc-button-content-color-down, token("gray-25"));
+  background-color: var(--swc-button-background-color-down, token("neutral-background-color-down"));
+  border-color: var(--swc-button-border-color-down, transparent);
+  scale: 0.95;
```
aramos-adobe: Following RSP's animation style here. 

```css
  transform: perspective(--_swc-button-min-block-size) translate3d(0, 0, -2px);
  transition-timing-function: token("animation-ease-in-out");
  transition-duration: token("animation-duration-200");
  will-change: transform;
```

2nd-gen/packages/swc/components/button/button.css
aramos-adobe, line 1: In the new documentation, there's a [justified property](https://s2.spectrum.corp.adobe.com/page/button/#justified) that allows the button fill its parent's container width. It's also show in the Figma component matrix (S2/Web Desktop) Should we add `--justified` variant and show an example in Storybook?

2nd-gen/packages/swc/components/button/button.css
5t3ph: Updated! I was initially stuck on not wanting to measure the button size, but your suggestion to use the `min-block-size` appears to achieve the desired effect, thanks!

2nd-gen/packages/swc/components/button/button.css
5t3ph: Good call - while technically additive, I went ahead and added it! Require a little finangling to get the story to render it, but that also represented a real-world constraint that I put a TODO to document in that phase 👍

2nd-gen/packages/swc/components/button/button.css
```diff
@@ -10,8 +10,445 @@
...
+  color: var(--swc-button-content-color-focus, token("gray-25"));
+  background-color: var(--swc-button-background-color-focus, token("neutral-background-color-key-focus"));
+  border-color: var(--swc-button-border-color-focus, transparent);
+}
+
+.swc-Button:active {
+  color: var(--swc-button-content-color-down, token("gray-25"));
+  background-color: var(--swc-button-background-color-down, token("neutral-background-color-down"));
+  border-color: var(--swc-button-border-color-down, transparent);
+  transform: perspective(var(--_swc-button-min-block-size)) translate3d(0, 0, -2px);
```
rise-erpelding: I know we discussed this a little earlier, but to restate: I'm cool with this tradeoff to avoid needing to calculate height/width! I do notice with the really wide buttons that they "move" more when pressed compared to the spectrum-css [version](https://spectrumcss.z13.web.core.windows.net/pr-2352/index.html?path=/story/components-button--default&globals=testingPreview:!true) but that feels more edge-case-y. Might be worth surfacing in a team discussion or somewhere if it's not already planned.

Also, there should be a token for the `-2px`: `component-size-difference-down`, not sure if there was a reason that was left out!

2nd-gen/packages/swc/components/button/button.css
```diff
@@ -10,8 +10,445 @@
...
+  flex-grow: 1;
+  justify-self: stretch;
+  inline-size: 100%;
+
+  .swc-Button {
+    inline-size: 100%;
+  }
+}
+
+@media (prefers-reduced-motion: reduce) {
```
rise-erpelding, line 441: This is awesome and probably we should use some version of this in progress-circle.

2nd-gen/packages/swc/components/button/button.css
```diff
@@ -10,8 +10,445 @@
...
+
+:host([truncate]) .swc-Button-label {
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+}
+
+/* ── Justified ────────────────────────────────────── */
+
+:host([justified]) {
```
rise-erpelding, line 431: It’s my first time seeing the justified variant, looks like maybe it's new to the designs?

Looks like the way button behaves compared to S1/S2 in spectrum-CSS is that the CSS set a min-inline-size on the default button to enforce a minimum width vs. now button width grows/shrinks based on content (which is consistent with React). It's probably worth noting in migration guides that any consumer who wants that same minimum-width behavior would need to use this variant instead. That's probably already on your list when we get to that step of the migration, but wanted to note at least for my memory!

rise-erpelding (approved):
I pointed out a few nits but this is a pretty clean, well-organized migration so far!

One thing I wanted to ask about though is: Was leaving out `-webkit-font-smoothing: antialiased;` a deliberate choice? It does make a difference in how bold the text appears (Spectrum-CSS on top, SWC on the bottom).

Image
Image

A couple other things here that could be surfaced for team discussion but don't need any changes now:
* How we're handling reduced motion (I think what you have here is fantastic)
* Whether we're going to establish an infield-progress-circle component for use in components like button, or just use some version of progress-circle, or something else.

Also (forgot to mention this earlier): I did test this with WHCM and didn't see any issues. I did note that we used to differentiate in WHCM between accent and every other variant, whereas now all our buttons look the same in WHCM regardless of variant. Also, only the border color changes on hover. That all seems like a good tradeoff for simplicity (since no forced-colors media queries were added), but wanted to note that I did test!

2nd-gen/packages/swc/components/button/button.css
```diff
@@ -10,8 +10,445 @@
...
+.swc-Button--pendingActive .swc-Button-label,
+.swc-Button--pendingActive slot[name="icon"]::slotted(*) {
+  display: none;
+}
+
+/* Positioning and sizing shared with icon */
+.swc-Button-pendingSpinner {
+  --_swc-pending-spinner-track-border-color: token("track-color");
+  --_swc-pending-spinner-fill-border-color: token("accent-content-color-default");
+  --_swc-pending-spinner-thickness: var(--swc-progress-circle-thickness, 2px);
```
rise-erpelding: Also forgot to say! There's a token for this `2px` value that might fit here: `progress-circle-thickness-small`, it scales to `3px` for mobile if that's something that we want.

aramos-adobe (approved):

2nd-gen/packages/swc/components/button/button.css
5t3ph: I missed adding back the token when I updated to transform, thanks!

Really wide buttons would definitely be edge-case, but I wouldn't say the movement is detrimental.

TBH, if I click the button normally (mouse or keyboard) I usually don't even linger long enough to trigger this state, which is another reason I'd vote for keeping this as low-level as possible.

2nd-gen/packages/swc/components/button/button.css
5t3ph: Thanks! It felt that particularly in the case of "I pressed a button, is somehting happening?" just presenting a paused progress circle would be more harmful. This version should pass the associated motion guidelines, and we could look into bringing it back to the progress-circle component too 👍 I will add a work item.

2nd-gen/packages/swc/components/button/button.css
5t3ph: Yes, I went with React consistency to remove the min-inline-size.

I'm a little torn on documenting that this is what to use for the min-width use case, since `justified` is a different intent, and it would require some wrapper-level management that would really be defining a _max_-width limitation. I'm thinking like a flex container that has multiple `justified` buttons, which is really saying "fill the available space, which is shared with your siblings" resulting in perhaps two equal width buttons, depending on available inline space.

I think I'm inclined to reserve that as a recommendation, and instead we'll lean on the customization strategies we'll eventually implement (like parts) that would allow a consumer to assign a true `min-inline-size` if needed.

2nd-gen/packages/swc/components/button/button.css
5t3ph: I was leaning into React's version, which doesn't change thickness for sizes, and is static at `2px`.

Also, since we are ultimately dropping scales, we don't need to use a token just for that concern anymore 👍

5t3ph:
> One thing I wanted to ask about though is: Was leaving out `-webkit-font-smoothing: antialiased;` a deliberate choice? It does make a difference in how bold the text appears (Spectrum-CSS on top, SWC on the bottom).

React is not using that property, and our past investigation (I can share in Slack if you'd like) was _tenatively_ recommending that we evaluate it per-component on whether it was creating a legibility issue. I'm inclined to leave it off Button for now.

> * How we're handling reduced motion (I think what you have here is fantastic)

Ticket added - SWC-2064

> * Whether we're going to establish an infield-progress-circle component for use in components like button, or just use some version of progress-circle, or something else.

Ticket added - SWC-2065

> Also (forgot to mention this earlier): I did test this with WHCM and didn't see any issues. I did note that we used to differentiate in WHCM between accent and every other variant, whereas now all our buttons look the same in WHCM regardless of variant. Also, only the border color changes on hover. That all seems like a good tradeoff for simplicity (since no forced-colors media queries were added), but wanted to note that I did test!

Since we are using the semantic `<button>` I removed WHCM styles to rely on browser native adjustments. React's only adjustment is to keep the "fill" vs "outline" distinction, but doesn't modify for like `accent` or even the `premium` or `genai` versions. I'd like to lean into default adjustments for now.

PR #6227, opened 2026-04-30
feat(button|tools): vite plugin for global element styles transform / Button global styles

5t3ph:
## Description

Implements the global button element stylesheet (`global-button.css`) and introduces a Vite plugin (`vite-global-elements-css`) that auto-generates it from the component CSS, keeping the two in sync without duplication.

**`vite-global-elements-css` (new tool)**  
Vite plugin that reads a component's `.css` file and outputs a `@layer swc-global-elements`-wrapped stylesheet suitable for use on native HTML elements (no Shadow DOM). Sections that require a JS runtime or are not globally applicable — `@keyframes`, `:host`, and JS-driven state classes like `.swc-Button--pendingActive` — are excluded using `/* @global-exclude */` / `/* @global-exclude-end */` annotations placed in the source CSS.

**`button.css` / `button-base.css`**  
Added `@global-exclude` annotations around: the pending spinner `@keyframes`, the `:host` display/alignment block, the full pending state (`.swc-Button--pendingActive`) and its reduced-motion override. Also exposed `max-inline-size` as `--swc-button-max-inline-size` (replaces a hardcoded `inherit`) to allow global element truncation via CSS.

**`global-button.css`**  
Fully regenerated by the plugin from `button.css`. The hand-authored placeholder is replaced with the auto-generated output. The file header marks it as generated and points to the source.

**`global-elements.mdx`**  
Expanded the customization guide with live examples for: variant modifier classes (secondary, accent, outline), static-color, disabled, label wrapping, truncation via `--swc-button-max-inline-size`, icons with `swc-Button-icon`, and icon-only with `aria-label`.

## Motivation and context

Resolves the global button reconciliation deferred from the Phase 4 PR.

Plugin allows for a single source of truth: edit `button.css`, the global stylesheet regenerates automatically.

## Related issue(s)

- SWC-1873 (follow-up — global styles split)

---

## Reviewer items

Verify new global classes and documentation in Storybook: `Guides > Customization > Global Element Styling`

**The Vite plugin (`vite-global-elements-css`)**  
This is new infrastructure. Key things to verify:

- The `@global-exclude` / `@global-exclude-end` annotation syntax is simple and doesn't introduce fragility — confirm the plugin correctly strips paired blocks and rejects unclosed ones
- The plugin has its own test suite (`index.test.js`) — check that it covers the main exclusion cases, selector rewriting (`:host([attr])` → `.swc-Button.swc-Button--attr`), and the `@layer` wrapper
- The generated output is committed to the repo (`global-button.css`) — reviewers should confirm the committed file matches what the build actually produces (i.e., run a build and diff)

**`global-button.css` correctness**  
Verify the visual output is equivalent where it matters (sizes, variants, outline, static-color, disabled) between the web component and global element styles.

**`@global-exclude` placement in component CSS**  
The annotations are part of the component's source CSS. Check that they don't accidentally affect how the Shadow DOM stylesheet is parsed or applied — they should be treated as normal CSS comments by the browser and Lit's CSS pipeline.


2nd-gen/packages/tools/vite-global-elements-css/index.js
```diff
@@ -0,0 +1,601 @@
...
+
+// ── PostCSS transformation ──────────────────────────────────────────────────
+
+/**
+ * Removes nodes enclosed by @global-exclude / @global-exclude-end comment fences.
+ * Works at any nesting depth: fences inside a media query strip only rules within
+ * that block. After stripping, at-rules left empty are removed.
+ *
+ * @param {import('postcss').Container} container - PostCSS container to strip fences from.
+ */
```
cdransf, line 249: Let me know if this makes sense or not 😇 The idea is to refactor `stripExcludedBlocks` to return a boolean value indicating whether the operation was successful. If not, it'll throw out of `deriveCSS` to prevent the plugin from writing malformed output.

```suggestion
 * @returns {boolean} `true` if the operation succeeded (all fences were closed).
 */
```

2nd-gen/packages/tools/vite-global-elements-css/index.js
```diff
@@ -0,0 +1,601 @@
...
+        excluding = false;
+        toRemove.push(node);
+        return;
+      }
+    }
+    if (excluding) {
+      toRemove.push(node);
+    } else if (node.type === 'atrule' && node.nodes) {
+      // Recurse so fences inside @media, @supports, etc. work within their scope
+      stripExcludedBlocks(node);
```
cdransf:
```suggestion
      if (!stripExcludedBlocks(node)) {
        excluding = true;
      }
```

2nd-gen/packages/tools/vite-global-elements-css/index.js
```diff
@@ -0,0 +1,601 @@
...
+    node.remove();
+  }
+
+  // Remove at-rules left empty after fence removal (e.g. @media with only fenced content)
+  container.each((node) => {
+    if (node.type === 'atrule' && node.nodes && node.nodes.length === 0) {
+      node.remove();
+    }
+  });
+}
```
cdransf, line 290:
```suggestion
  return !excluding;
}
```

2nd-gen/packages/tools/vite-global-elements-css/index.test.js
```diff
@@ -0,0 +1,378 @@
...
+      @media (prefers-reduced-motion: reduce) {
+        /* @global-exclude: all content is pending-only */
+        :host([pending]) .swc-Button { animation: none; }
+        /* @global-exclude-end */
+      }
+    `;
+    const result = deriveCSS(css, 'swc-Button');
+    expect(result).not.toContain('@media');
+  });
+
```
cdransf, line 228: Test whether we throw and prevent malformed output if we don't have a closing exclude fence. ✨ 

```suggestion
  it('throws on an unclosed @global-exclude fence', () => {
    const css = `
      /* @global-exclude */
      :host([pending]) { color: blue; }
      :host { color: red; }
    `;
    expect(() => deriveCSS(css, 'swc-Button')).toThrow(
      'Unclosed @global-exclude fence'
    );
  });
```

2nd-gen/packages/tools/vite-global-elements-css/index.js
```diff
@@ -0,0 +1,601 @@
...
+ *   4. mergeRules           — combine duplicate selectors, deduplicate declarations
+ *   5. wrapInLayer          — wrap in @layer and add revert-layer escape hatch
+ *
+ * @param {string} sourceCss - Concatenated raw CSS (base + component, token() calls intact).
+ * @param {string} block - BEM block class name, e.g. 'swc-Button'.
+ * @returns {string}
+ */
+export function deriveCSS(sourceCss, block) {
+  const root = postcss.parse(sourceCss);
+  stripExcludedBlocks(root);
```
cdransf:
```suggestion
  if (!stripExcludedBlocks(root)) {
    throw new Error(
      '[vite-global-elements-css] Unclosed @global-exclude fence — missing @global-exclude-end');
  }
```

cdransf (approved):
Looks great! I added a chain of suggestions to the plugin index. ✨

2nd-gen/packages/tools/vite-global-elements-css/index.js
5t3ph: cool, thanks - updated!

PR #6228, opened 2026-05-01
feat(ai|icon): update storybook parameter guidance, fix for icon

5t3ph:
## Description

Resolves recommendation for Storybook `parameters` for layout of multi-item stories.

## Motivation and context

It was found that `flexLayout: true` doesn't actually change the story layout, and instead `flexLayout: 'row-wrap'` is most often the intended result of aligning items, adding gap, and allowing them to wrap.

**Before** icon's use of just `true` had no actual effect
image

**After** improved layout with correct `row-wrap` value
image

## Manual review test cases

- [ ] All instances of `flexLayout: true` within component stories and AI rules and skills is updated to `row-wrap` or removed if example isn't for multi-item stories


cdransf (approved):

nikkimk (approved):
lgtm

PR #6243, opened 2026-05-07
fix(dropzone): resolve drop event not firing on Windows Chrome

Rajdeepc:
## Description

Fixes `sp-dropzone-drop` event not firing on Windows Chrome. The root cause was that `event.preventDefault()` in the `onDragOver` handler was gated behind a `dataTransfer` null check and the `shouldAccept` guard. On Windows Chrome, this meant the browser never saw `dragover` as canceled and refused to fire the native `drop` event — silently breaking all dropzone interactions.

## Motivation and context

**SEV 2** bug report: `sp-dropzone-drop` does not fire on Windows Chrome (latest), preventing all drag-and-drop functionality in downstream consumers (sceneline, quick actions, etc.). The same code works on macOS Chrome because macOS Chrome is lenient about the `dragover` cancellation requirement, while Windows Chrome strictly follows the HTML5 DnD spec.

Reproduction: https://stackblitz.com/edit/vitejs-vite-qrxb5dua — works on macOS Chrome, fails on Windows Chrome.

## Related Issue
 Fixes - https://github.com/adobe/spectrum-web-components/issues/6251

### Bug fix: `onDragOver` — unconditional `preventDefault()`

Moved `event.preventDefault()` to the top of `onDragOver` so it runs before any guards. The HTML5 DnD spec requires `dragover` to be canceled for the browser to allow a `drop`. The `shouldAccept` rejection path now sets `dropEffect = 'none'` (visual feedback) while still keeping `preventDefault()` so the browser maintains the drop gesture.

### Improvement: `onDragLeave` — `relatedTarget` guard

Added a check for `event.relatedTarget` to suppress internal `dragleave` events that fire when the cursor moves between child elements within the dropzone. Previously, moving from the SVG to a label would trigger a spurious `sp-dropzone-dragleave`.

### Cleanup: `disconnectedCallback`

Added `clearDebouncedDragLeave()` call to prevent stale timeouts from firing after the element is removed from the DOM.

### Story: `FileDrop`

Added a `FileDrop` story that accepts image drops with `console.log` on `sp-dropzone-drop` for manual cross-platform verification.

## Author's checklist

- [x] I have read the **[CONTRIBUTING]((https://github.com/adobe/spectrum-web-components/blob/main/CONTRIBUTING.md))** and **[PULL_REQUESTS]((https://github.com/adobe/spectrum-web-components/blob/main/PULL_REQUESTS.md))** documents.

## Reviewer's checklist

### Manual review test cases

- [ ] Verify drop fires on Windows Chrome
  1. Open the [Dropzone > event Logger story](https://swcpreviews.z13.web.core.windows.net/pr-6243/docs/first-gen-storybook/?path=/story/dropzone--event-logger) in Storybook on **Windows Chrome**
  2. Drag an image file from the desktop onto the dropzone
  3. Expect `sp-dropzone-drop` logged in the browser console

- [ ] Verify drop still fires on macOS Chrome
  1. Open the [Dropzone > Event Logger](https://swcpreviews.z13.web.core.windows.net/pr-6243/docs/first-gen-storybook/?path=/story/dropzone--event-logger) in Storybook on **macOS Chrome**
  2. Drag an image file from Finder onto the dropzone
  3. Expect `sp-dropzone-drop` logged in the browser console

### Device review

- [x] Did it pass in Desktop?

- [ ] **Keyboard**
  1. Open the [Dropzone > Default story](https://swcpreviews.z13.web.core.windows.net/pr-6243/docs/first-gen-storybook/?path=/story/dropzone--default)
  2. Press <kbd>Tab</kbd> to focus the dropzone (it has `tabindex="0"`)
  3. Confirm visible focus ring appears (solid border, accent color)
  4. Confirm no keyboard regressions — dropzone is non-interactive beyond focus

- [ ] **Screen reader**
  1. Open the [Dropzone > Default story](https://swcpreviews.z13.web.core.windows.net/pr-6243/docs/first-gen-storybook/?path=/story/dropzone--default) with VoiceOver (macOS) or NVDA (Windows)
  2. Navigate to the dropzone
  3. Confirm the heading "Drag and Drop Your File" and link "Select a File" are announced
  4. Confirm no unexpected or duplicate announcements from the drag state changes


1st-gen/packages/dropzone/src/Dropzone.ts
```diff
@@ -87,66 +89,96 @@ export class Dropzone extends SpectrumElement {
...
     }, 100);
   }
 
   public onDrop(event: DragEvent): void {
+    /**
+     * Prevent browser default behavior (opening files in browser).
+     */
     event.preventDefault();
 
     this.clearDebouncedDragLeave();
```
cdransf, line 176: Should we add a guard like this? If `sp-dropzone-should-accept` rejects the drag the handler would return early with `dropEffect = 'none'` and `isDragged` stays false so `sp-dropzone-dragover` is never dispatched.

Since `preventDefault()` ran the browser fires the `drop` event anyways, `onDrop` dispatches `sp-dropzone-drop` and isn't aware of the rejection.

Should we test that `sp-dropzone-drop` doesn't fire when `shouldAccept` is cancelled? ✨ 

```suggestion
    if (!this.isDragged) {
      return;
    }
```

1st-gen/packages/dropzone/stories/dropzone.stories.ts
```diff
@@ -206,3 +206,55 @@ export const Controlled = (): TemplateResult => {
...
+          Search Adobe Stock
+        </sp-link>
+      </div>
+    </sp-dropzone>
+  `;
+};
+
+EventLogger.parameters = {
+  tags: ['!dev'],
+};
```
cdransf, line 260:
```suggestion
EventLogger.tags = ['!dev'];
```

1st-gen/packages/dropzone/test/dropzone.test.ts
```diff
@@ -162,8 +229,36 @@ describe('Dropzone', () => {
...
+    };
+    const el = await fixture<Dropzone>(html`
+      <sp-dropzone id="dropzone" @sp-dropzone-dragleave=${onDragLeave}>
+        <div id="child-a">A</div>
+        <div id="child-b">B</div>
+      </sp-dropzone>
+    `);
+
+    await elementUpdated(el);
+
```
cdransf, line 245: `isDragged` starts as false, so we need to set it to true here... ✨ 

```suggestion

el.isDragged = true;

```

1st-gen/packages/dropzone/test/dropzone.test.ts
```diff
@@ -162,8 +229,36 @@ describe('Dropzone', () => {
...
+    el.dispatchEvent(internalLeave);
+
+    await new Promise((resolve) => setTimeout(resolve, 150));
 
-  it('manages `dragleave` events', async () => {
+    expect(
+      dragLeftCount,
+      'dragleave is suppressed when relatedTarget is an internal child'
+    ).to.equal(0);
+    expect(el.isDragged).to.be.false;
```
cdransf: ...and verify that it remains true here so we can verify that an internal `dragleave` does not clear the drag state. ✨ 

```suggestion
    expect(el.isDragged).to.be.true;
```

cdransf (dismissed):
I highlighted a few things that jumped out and added suggestions. ✨

1st-gen/packages/dropzone/stories/dropzone.stories.ts
Rajdeepc: Yes tags is a top level property. We would like to clean this up but not in this PR.

1st-gen/packages/dropzone/src/Dropzone.ts
Rajdeepc: Good catch this is a real semantic gap I missed.

1st-gen/packages/dropzone/test/dropzone.test.ts
Rajdeepc: Yup this is correct, without setting `isDragged = true` first, the test just proves `false` stays `false`, which would pass regardless of whether the `relatedTarget` guard works.

1st-gen/packages/dropzone/test/dropzone.test.ts
Rajdeepc: Agreed the whole point is to verify the drag state survives an internal child transition.

5t3ph (approved):
Validated with BrowserStack, Windows Chrome

rubencarvalho (approved):
Nice! 🚀

PR #6287, opened 2026-05-13
refactor(upload-artifact): remove redundant data-preview-only mechanism

rubencarvalho:
## Description

While reviewing the `UploadArtifact` component I noticed a `_previewOnly` private property that didn't make sense to me, so I dug into it. It turned out to be entirely redundant and removing it had no visible effect.

The mechanism worked like this:

- `_previewOnly` was a `private` field decorated with `@property` (which is for public API), always set to `this.type === 'media'` via a `_syncMode` helper called from `firstUpdated` and `updated`
- It reflected to a `data-preview-only` attribute on the host solely for CSS targeting
- Every CSS rule that used `[data-preview-only]` also required `[type="media"]`, so the combined selector was functionally identical to `[type="media"]` alone

Removed `_previewOnly`, `_syncMode`, and the two lifecycle hooks. Collapsed the four duplicate CSS selectors into the existing `[type="media"]` block and updated the test assertions that were checking the now-gone attribute.

## Motivation and context

The `@property` decorator on a `private` field is not idiomatic Lit — `@property` is for public reactive properties. The attribute it reflected was never set externally, was 100% derivable from `type`, and added a layer of indirection with no benefit.

## Related issue(s)

N/A

## Author's checklist

No interactive behavior changed. The dismiss button and slot structure are untouched.

- [x] **Keyboard**: No focusable parts were added or removed; no regressions expected.
- [x] **Screen reader**: No ARIA roles, labels, or state announcements changed.

### DIFF
--- a/.stylelintignore
+++ b/.stylelintignore
@@ -34,4 +34,3 @@
 1st-gen/packages/iconset
 1st-gen/packages/icons-ui
 1st-gen/packages/icons-workflow
-2nd-gen/packages/swc/stylesheets/global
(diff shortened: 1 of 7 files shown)


rubencarvalho:
Closing in favour of a clean branch — the original branch had unrelated commits on it.

PR #6292, opened 2026-05-13
fix(color-wheel): refresh layout cache before ring hit-test

rubencarvalho:
## Description

The ring hit-test in `handleGradientPointerdown` relied on a cached `boundingClientRect` that was only refreshed in `firstUpdated`, on the handle's pointerdown, and from the `ResizeObserver`. Because `getBoundingClientRect` is viewport-relative, any scroll or ancestor layout shift between those updates left the cache stale, causing valid clicks on the ring to be silently rejected by `isPointerInRing`.

This change calls `cacheLayoutData()` at the start of `handleGradientPointerdown` (after the early `button` / `innerCircle` checks, before the ring hit-test) so the rect used for the comparison is always in sync with the wheel's current on-screen position.

## Motivation and context

`<sp-color-wheel>` is typically used inside scrollable panels or popovers where the wheel can move within the viewport between pointer events. In those cases the ring became intermittently non-clickable until the next resize or handle interaction reseeded the cache. Refreshing on pointerdown is cheap (one synchronous layout read on user input) and removes the entire class of stale-rect bugs without changing the public API.

You can probably trigger the bug yourself by opening the [Color Wheel Storybook page](https://opensource.adobe.com/spectrum-web-components/storybook/?path=/story/color-wheel--default) and clicking around on some of the examples. Sometimes it will not work 🤷 

## Author's checklist

## Reviewer's checklist

### Manual review test cases

- [ ] _Ring click works after page scroll_
  1. Open the [Color Wheel Storybook page](https://swcpreviews.z13.web.core.windows.net/pr-6292/docs/second-gen-storybook/?path=/docs/color-wheel--docs) and ensure the wheel is in a scrollable container (or zoom the browser so the page scrolls).
  2. Scroll the page so the wheel moves to a different viewport position than it had on initial render.
  3. Click anywhere on the colored ring (not the inner cutout, not the handle).
  4. Expect the handle to jump to the clicked position and the `change`/`input` events to fire, every time. Before this fix, clicks were silently ignored until the next resize.

- [ ] _Handle drag and inner-circle behavior unchanged_
  1. From the same story, drag the ring handle around the wheel.
  2. Click inside the inner circle (the cutout in the middle).
  3. Expect handle dragging to work as before, and clicks on the inner circle to be ignored as before (no regression).


1st-gen/packages/color-wheel/test/color-wheel.test.ts
```diff
@@ -696,6 +696,65 @@ describe('ColorWheel', () => {
...
+        clientY,
+        bubbles: true,
+        composed: true,
+        cancelable: true,
+      })
+    );
+
+    await elementUpdated(el);
+
+    expect(el.value).to.not.equal(initialValue);
```
cdransf: Could we tighten up this assertion? ✨ 

```suggestion
    expect(el.value).to.be.closeTo(270, 10);
```

cdransf (approved):
Looks good! Just one small question ✨

caseyisonit:
Is there a ticket or issue associated with this? or where did the fix bubble up from?

caseyisonit (approved):

rubencarvalho:
> Is there a ticket or issue associated with this? or where did the fix bubble up from?

No ticket - I found it when manual verifying from my side on our docs website 😓

PR #6302, opened 2026-05-14
feat(ai): add rules for migration plan and code conformance adherence

5t3ph:
## Description

Adds two new AI agent rules that address drift and quality gaps during multi-phase 2nd-gen component migrations:

- **`migration-phase-awareness.md`**: Keeps migration phase obligations in context throughout a session, even when work drifts to unrelated tasks. Validates completion against both the phase skill's quality gate and the migration plan before declaring a phase done. Emits a structured Migration Checkpoint block at phase completion to serve as a cross-session handoff artifact (no shared file needed, avoids git conflicts across concurrent migrations). Appends an actionable resume prompt when the session drifts.

- **`consistency-pass.md`**: Defines a two-part self-audit — code conformance (delegating to the existing `code-conformance` rule) and migration plan validity. The plan validity check treats the plan as authoritative, flags implementation divergence rather than silently updating the plan, and performs an intra-plan cascading consistency scan when any confirmed change is made (e.g., renamed attribute → update all plan sections that reference the old name).

## Motivation and context

During multi-phase migrations, agents completing one phase and moving on to other work frequently lost track of outstanding obligations like the conformance sub-task and the review phase. Engineers were also seeing plan inconsistencies accumulate as implementation decisions were made — a renamed attribute updated in one section but not others, a removed dependency still referenced in behavior notes.

These rules make those checks automatic and keep them visible throughout the session, reducing reliance on engineer memory to ask for them.

## Manual review test cases

> **NOTE**: You probably don't actually want to run a migration phase ad-hoc since they require a large context window, and shouldn't be run out of order even for testing since they are co-dependent. These will be things to keep an eye on for the next few valid phase skill runs.

- [ ] _Migration checkpoint is emitted at phase completion_
  1. Start a migration session using any `migration-*` skill
  2. Complete the phase workflow
  3. Expect a Migration Checkpoint block at the end of the response with `Phase completed`, `Next phase`, `Consistency pass`, and `Phase blockers` fields

- [ ] _Drift reminder fires when work leaves the migration context_
  1. Start a migration session using any `migration-*` skill
  2. Ask the agent to fix an unrelated bug or make an unrelated change
  3. Expect the agent to complete the request and append a "Migration paused: to continue [Component], invoke `migration-[skill]`" line

- [ ] _Consistency pass covers both code conformance and plan validity_
  1. Ask the agent to run a "consistency pass" on a component mid-migration
  2. Expect output with two labeled sections: Code conformance and Plan validity
  3. Plan validity section should flag any implementation divergence from the plan, and note any cascading effects of confirmed in-session changes

### DIFF
--- a/.ai/rules/consistency-pass.md
+++ b/.ai/rules/consistency-pass.md
@@ -57,6 +57,8 @@ Read the plan sections covering the current phase and verify the implementation 
 
 Flag any divergence. If a change was confirmed in-session, update the plan and proceed to Check 2.
 
+Once implementation and plan are aligned, update the plan's checklists for the current phase to reflect the work completed. Mark off completed items, fill in any "to be determined" fields that were resolved during this phase, and note any decisions made. The plan checklists are the record that this phase was done and what was produced — keep them current.
+
 ### Check 2: Intra-plan consistency (cascading updates)
 
 When any part of the plan changes — whether from a session-confirmed decision or a correction — scan the entire plan for other sections affected by that change. Plans are multi-section documents where a single decision can appear in several places.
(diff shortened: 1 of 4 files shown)


cdransf (approved):
Looks great! ✨

caseyisonit (approved):

PR #6309, opened 2026-05-15
feat(tooltip) migration: files, API, a11y

5t3ph:
## Description

Implements Phases 2–4 of the `swc-tooltip` 2nd-gen component migration: file structure setup, public API, behavioral layer, and accessibility semantics. This PR does not include styling (Phase 5), tests (Phase 6), or stories/documentation (Phase 7) — those follow in subsequent PRs against this branch.

**2nd-gen files added (`2nd-gen/packages/`):**

- `core/components/tooltip/Tooltip.base.ts` — abstract base class; declares all public properties (`variant`, `placement`, `open`, `for`, `triggerElement`, `delayed`, `disabled`, `manual`, `offset`, `labeling`), `VARIANTS`/`PLACEMENTS` statics, and sets `role="tooltip"` in `connectedCallback`
- `core/components/tooltip/Tooltip.types.ts` — `TooltipVariant`, `TooltipPlacement`, and related const arrays
- `core/components/tooltip/index.ts` — core re-export
- `swc/components/tooltip/Tooltip.ts` — concrete SWC class; native `popover="auto"` integration, `beforetoggle`/`toggle`/`transitionend` event listeners, `swc-open`/`swc-close`/`swc-after-open`/`swc-after-close` event dispatch, trigger resolution via `for` (ID lookup) or `triggerElement` (element reference), `ariaDescribedByElements` wiring on `open` change
- `swc/components/tooltip/swc-tooltip.ts` — side-effect registration entry point
- `swc/components/tooltip/index.ts` — class re-export
- `swc/components/tooltip/tooltip.css` — empty stub (Phase 5)

**1st-gen deprecation notices (`1st-gen/packages/tooltip/src/Tooltip.ts`):**

- `@deprecated` JSDoc on `selfManaged` and `tipPadding`
- Runtime `window.__swc.warn()` calls for `variant="info"` (renamed to `"informative"`), `variant="positive"` (removed in S2), and the `self-managed` attribute

## Motivation and context

Part of the 1st-gen → 2nd-gen component migration workstream. Tracking issue for this component: Epic SWC-2017.

Key architectural decisions made in this PR (see migration plan for full rationale):

- **Native `popover="auto"`** replaces `sp-overlay type="hint"` — the browser handles stacking, `Escape` dismissal, and AT visibility automatically
- **Cycle prevention** for the `open` property setter: used `updated()` + `:popover-open` CSS pseudo-class check instead of a getter/setter override, avoiding a TS2611 incompatibility with Lit's `@property()` class fields
- **`ariaDescribedByElements`** (element reference API, Baseline April 2025) replaces string-based `aria-describedby` wiring — no auto-generated tooltip `id` required; consumers who need manual wiring provide their own `id`
- **Inner-button resolution**: `trigger.shadowRoot?.querySelector('button') ?? trigger` to reach the AT-facing element inside SWC components
- **`swc-after-open`/`swc-after-close`** dispatched from `transitionend`, with an immediate fallback when `transitionDuration === '0s'`

## Related issue(s)

- Fixes SWC-1558 (`role="tooltip"` missing from 1st-gen)
- SWC-2020/21/22
- Epic SWC-2017

## Manual review test cases

**Architecture and core/SWC split:**

- [ ] Properties on `Tooltip.base.ts` match those listed in the migration plan; none are accidentally placed in the SWC class instead
- [ ] `role="tooltip"` is set via `connectedCallback` in `TooltipBase`, not in the SWC class
- [ ] `VARIANTS` and `PLACEMENTS` statics are `readonly string[]` typed against the const arrays in `Tooltip.types.ts`

**Behavioral layer (`swc/components/tooltip/Tooltip.ts`):**

- [ ] `updated()` + `this.matches(':popover-open')` correctly prevents an open/close API cycle — `showPopover()`/`hidePopover()` are only called when `this.open` and the browser's popover state disagree
- [ ] `syncAriaRelationship()` resolves `trigger.shadowRoot?.querySelector('button') ?? trigger`; adds the tooltip on open, removes it on close, no orphan references
- [ ] `swc-after-open`/`swc-after-close` dispatches immediately when `transitionDuration === '0s'`; otherwise waits for `transitionend`
- [ ] Event listeners added in `connectedCallback` are removed in `disconnectedCallback` using the same bound references

**1st-gen deprecation notices (`1st-gen/packages/tooltip/src/Tooltip.ts`):**

- [ ] `@deprecated` JSDoc is present on `selfManaged` and `tipPadding` with correct wording (no 2nd-gen API references in warn messages)
- [ ] `window.__swc.warn()` calls use `{ level: 'deprecation' }` and are inside existing code paths only — no new backing fields, no structural refactors
- [ ] `variant="info"` warn message includes `Use "informative" instead.`; `variant="positive"` warn has no "Use" clause (no direct replacement)

Accessibility testing requires a rendered component and will be documented in the Phase 6/7 PR. Reviewers can verify accessibility semantics from code:

- [ ] `role="tooltip"` is set on the host element in `TooltipBase.connectedCallback()` — this fixes the gap tracked in SWC-1558
- [ ] `ariaDescribedByElements` is set on the trigger's inner interactive element when `open` changes to `true`; removed when `open` changes to `false`
- [ ] Native `popover="auto"` ensures the closed tooltip is not in the accessibility tree and that `Escape` dismissal is handled without focus trapping — no additional code required


2nd-gen/packages/swc/components/tooltip/Tooltip.ts
```diff
@@ -0,0 +1,153 @@
...
+    }
+    return null;
+  }
+
+  private syncAriaRelationship(): void {
+    const trigger = this.resolveTrigger();
+    if (!trigger) {
+      return;
+    }
+    const target = (trigger.shadowRoot?.querySelector('button') ??
```
cdransf: Is this a potential issue if there's more than one button in the shadow root and we only select the first? ✨

2nd-gen/packages/swc/components/tooltip/Tooltip.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  private readonly handleToggle = (event: Event): void => {
+    const { newState } = event as ToggleEvent;
+    const isOpen = newState === 'open';
+    if (isOpen === this.open) {
+      return;
+    }
+    this.open = isOpen;
+    // When no CSS transition is active, dispatch after-* immediately since transitionend will not fire.
+    if (getComputedStyle(this).transitionDuration === '0s') {
```
cdransf: Is this an issue if we get a comma separated list when multiple properties are transitioning? ✨

2nd-gen/packages/core/package.json
```diff
@@ -59,6 +59,10 @@
       "types": "./dist/components/tabs/index.d.ts",
       "import": "./dist/components/tabs/index.js"
     },
+    "./components/tooltip": {
```
cdransf, line 62: Do we need a corresponding change in `swc/package.json`? ✨

cdransf (changes requested):
Just a few questions, looks awesome overall ✨

2nd-gen/packages/swc/components/tooltip/Tooltip.ts
5t3ph: Good question - I would say we make it explicit in the Storybook docs that there is an expectation that the element `for` is attached to contains exactly one button. We _may_ extend this selector in the future, such as to support `input`, but that will be additive pending the forms strategies.

2nd-gen/packages/swc/components/tooltip/Tooltip.ts
5t3ph: `getComputedStyle()` looks at the final value as used by the browser to style the element, and this is only looking at the `transitionDuration` which shouldn't be variable for different properties for tooltip's purposes. Use of `0s` is informed by [the initial value](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/transition-duration#formal_definition) of the `transition-duration` property. If we change strategy during the styling phase, or find conflicts during reduced motion testing, we'll re-asses 👍

2nd-gen/packages/core/package.json
5t3ph: `swc/package.json` has components covered using wildcards vs. explicit exports 👍

cdransf (approved):

2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
```diff
@@ -0,0 +1,266 @@
...
+    if (getComputedStyle(this).transitionDuration === '0s') {
+      this.dispatchAfterEvent(isOpen);
+    }
+  };
+
+  private readonly handleTransitionEnd = (event: TransitionEvent): void => {
+    if (event.target !== this) {
+      return;
+    }
+    this.dispatchAfterEvent(this.open);
```
rubencarvalho, line 234: Interesting... Will this fire on per property transition end?

5t3ph: this issue is resolved in the next batch of work for styling where I noticed it was an issue 👍

PR #6316, opened 2026-05-18
feat(meter): scaffold 2nd-gen meter (setup, API, a11y, styling) — SWC-2008..2011

blunteshwar:
## Description

Scaffolds the 2nd-gen `<swc-meter>` against the approved migration plan, covering four sequential phase tickets (Setup, API, Accessibility, Styling). Targets the long-lived `meter-migration` integration branch — not `main`. Subsequent phase tickets (conformance, testing, documentation, consumer migration guide, final review) ship as follow-up PRs into the same base.

**Phases delivered:**

- **SWC-2008 — Setup.** Core / SWC file structure, `package.json` exports for `@spectrum-web-components/core/components/meter`, `swc/shared/linear-progress-base.css` stub.
- **SWC-2009 — API.** `LinearProgressMixin` owns value clamping, fill-fraction, `Intl.NumberFormat` locale formatting via `LanguageResolutionController`, `label` / `description` slot tracking, and the DEBUG-mode accessible-name warning. `MeterBase` adds `variant` plus DEBUG variant validation. `formatOptions` is a JS-only `Intl.NumberFormatOptions` pass-through.
- **SWC-2010 — Accessibility.** `role="meter"` on the shadow `.swc-Meter` wrapper (host carries no ARIA); `aria-valuemin` / `aria-valuemax` / `aria-valuenow` / `aria-valuetext` on the role element; `label` slot → `aria-labelledby`, `accessibleLabel` → `aria-label` fallback when no label slot, `description` slot → `aria-describedby`; non-focusable.
- **SWC-2011 — Styling.** Ports `spectrum-css` `spectrum-two` `progressbar/index.css` + `meter/index.css` to the `swc-Meter` namespace, replaces all `--mod-*` with `--swc-meter-*` tokens, covers variants (`informative` / `positive` / `notice` / `negative`), sizes (`s` / `m` / `l` / `xl`), `label-position="side"`, `static-color="white"` / `"black"`, and a forced-colors block. Stylelint, ESLint, and Prettier all clean; build green.

## Motivation and context

Phase-1 plan is merged via [#6215](https://github.com/adobe/spectrum-web-components/pull/6215). This PR is the first implementation pass against that plan. ARIA contract follows the binding decisions in `CONTRIBUTOR-DOCS/03_project-planning/03_components/meter/accessibility-migration-analysis.md` (role on shadow, label slot + `accessibleLabel` fallback, description slot, no host ARIA).

## Related issue(s)

- SWC-2008 — Setup
- SWC-2009 — Migrate API and TypeScript code
- SWC-2010 — Implement accessibility features
- SWC-2011 — Full S2 visual fidelity

## Known deferrals

- **Shared linear-progress CSS** stays inlined in `meter.css` for this PR; the lift-out to `swc/shared/linear-progress-base.css` happens when `<swc-progress-bar>` migrates (vite's postcss pipeline does not include `postcss-import`, and no other 2nd-gen component uses `@import`).
- **Description container** is always emitted with an `is-empty` class instead of being fully omitted via `nothing`, so the `<slot>` stays mounted and `slotchange` fires reliably when content is added at runtime. ARIA-describedby is still gated on the tracked flag, and the empty container is hidden via CSS.
- **Stories and tests** are scaffolded under separate tickets (SWC-2013, SWC-2014). This PR does not include `meter.stories.ts` or `meter.test.ts`.

## Author's checklist

- [x] I have reviewed the Accessibility Practices for this feature, see: [Aria Practices](https://www.w3.org/WAI/ARIA/apg/patterns/meter/)
- [ ] I have added automated tests to cover my changes. _(Deferred — tests ship under SWC-2013.)_
- [ ] I have included a well-written changeset if my change needs to be published. _(N/A — meter-migration branch; no published package version bump until the integration branch lands on `main`.)_
- [ ] I have included updated documentation if my change required it. _(Deferred — documentation ships under SWC-2014.)_

## Reviewer's checklist

- [ ] Includes a GitHub issue with appropriate flag or Jira ticket number without a link

### Manual review test cases

- [ ] _Default meter renders with role="meter" on the shadow `.swc-Meter`._
  1. Build `2nd-gen` (`yarn build:2nd-gen`).
  2. Import `@adobe/spectrum-wc/components/meter/swc-meter.js` in a sandbox page.
  3. Render `<swc-meter value="40"><span slot="label">Storage used</span></swc-meter>`.
  4. Inspect the shadow root. Expect `role="meter"` only on the inner `.swc-Meter` element; the host element has no `role` attribute. `aria-valuemin="0"`, `aria-valuemax="100"`, `aria-valuenow="40"`, `aria-valuetext` matches the rendered percent.

- [ ] _Accessible name resolves from the label slot._
  1. Render the same meter as above.
  2. Confirm the role element has `aria-labelledby` pointing at the shadow id of the label container.
  3. Remove the label slot and add `accessible-label="Storage usage"`.
  4. Confirm the role element switches to `aria-label="Storage usage"` and `aria-labelledby` is absent.

- [ ] _Variants and sizes apply S2 tokens._
  1. Render meters with `variant` values `informative`, `positive`, `notice`, `negative` across sizes `s`, `m`, `l`, `xl`.
  2. Confirm fill colors match the `*-visual-color` tokens (and `accent-content-color-default` for `informative`) and bar thickness matches the size token.

- [ ] _Static color and side label._
  1. Render `<swc-meter value="50" static-color="white" label-position="side">...</swc-meter>` against a dark background.
  2. Confirm the bar, label, and value use the `static-white-*` tokens and the label sits inline with the bar.

### Device review

- [ ] Did it pass in Desktop?

  1. Tab through a page containing `<swc-meter>` instances at varying values.
  2. Expect Tab focus to skip the meter; no focus ring should appear on the host or any shadow descendant.
  3. Confirm no focus traps and no surprising focus stops on adjacent components.

  1. Open a story with `<swc-meter value="40" variant="positive"><span slot="label">Storage used</span><span slot="description">2 GB of 5 GB used</span></swc-meter>` in NVDA / JAWS / VoiceOver.
  2. Switch to browse mode (NVDA / JAWS) or VO arrow navigation.
  3. Encounter the meter and confirm the announcement includes: name from the label slot, role of "meter" (or platform equivalent), the value text (`40%` in `en-US`, Arabic-Indic digits + `٪` in `ar-sa`), and the description text.
  4. Replace the label slot with `accessible-label="Storage usage"` and confirm the announcement now uses that name instead, with `aria-labelledby` no longer referenced.


2nd-gen/packages/core/mixins/linear-progress-mixin.ts
```diff
@@ -0,0 +1,259 @@
...
+  readonly hasLabelSlotContent: boolean;
+  readonly hasDescriptionSlotContent: boolean;
+  readonly labelContainerId: string;
+  readonly descriptionContainerId: string;
+}
+
+let nextLinearProgressId = 0;
+
+/**
+ * Thin mixin shared by linear progress components (`<swc-meter>`, future
```
rubencarvalho: Can we not have these "future" references in the code? They may make sense today, but will inevitably become outdated.

2nd-gen/packages/core/mixins/linear-progress-mixin.ts
```diff
@@ -0,0 +1,259 @@
...
+     * stays internally consistent.
+     */
+    public get sanitizedMin(): number {
+      const min = Number.isFinite(this.minValue) ? this.minValue : 0;
+      const max = Number.isFinite(this.maxValue) ? this.maxValue : 100;
+      return Math.min(min, max);
+    }
+
+    /**
+     * Sanitized upper bound. See {@link sanitizedMin}.
```
rubencarvalho: Do these `@link` notations work as expected? I'm not sure it is a pattern we have.

2nd-gen/packages/core/mixins/linear-progress-mixin.ts
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    /**
+     * Sanitized lower bound: falls back to 0 when `minValue` is non-finite,
+     * and to the smaller of (min, max) when the two are reversed. ARIA
+     * `aria-valuemin`, `aria-valuenow`, `aria-valuetext`, and the rendered
+     * fill all read from the same sanitized range so screen-reader output
+     * stays internally consistent.
+     */
+    public get sanitizedMin(): number {
```
rubencarvalho, line 185: Are we intentionally making all these properties public?

2nd-gen/packages/swc/shared/linear-progress-base.css
rubencarvalho, line 1: could we get @5t3ph input on how do we imagine css sharing to happen? no strong opinion, but we probably want to standardize it

2nd-gen/packages/swc/components/meter/Meter.ts
```diff
@@ -0,0 +1,133 @@
...
+      !hasLabel && this.accessibleLabel ? this.accessibleLabel : undefined;
+    const ariaDescribedBy = hasDescription
+      ? this.descriptionContainerId
+      : undefined;
+
+    return html`
+      <div
+        class=${classMap({
+          ['swc-Meter']: true,
+          [`swc-Meter--${this.variant}`]: typeof this.variant !== 'undefined',
```
rubencarvalho: isn't variant always defined?

2nd-gen/packages/swc/components/meter/Meter.ts
```diff
@@ -0,0 +1,133 @@
...
+          [`swc-Meter--sideLabel`]: this.labelPosition === 'side',
+          [`swc-Meter--staticWhite`]: this.staticColor === 'white',
+          [`swc-Meter--staticBlack`]: this.staticColor === 'black',
+        })}
+        role="meter"
+        aria-valuemin=${this.sanitizedMin}
+        aria-valuemax=${this.sanitizedMax}
+        aria-valuenow=${this.clampedValue}
+        aria-valuetext=${this.formattedValue}
```
rubencarvalho: we should be catching this formattedValue, rn we're calling this function twice. here and below on the value. that's actually creating a new object each time.

2nd-gen/packages/core/mixins/linear-progress-mixin.ts
```diff
@@ -0,0 +1,259 @@
...
+      }
+      return formatter.format(this.clampedValue);
+    }
+
+    protected override updated(changes: PropertyValues): void {
+      super.updated(changes);
+
+      if (changes.has(languageResolverUpdatedSymbol)) {
+        // Locale shifted; re-render so formatted value + aria-valuetext refresh.
+        this.requestUpdate();
```
rubencarvalho: by calling request update, aren't we firing 2 rerenders here?

rubencarvalho: can you look into https://lit.dev/docs/components/lifecycle/#willupdate so the first render already has the correct locale value?

2nd-gen/packages/core/mixins/linear-progress-mixin.ts
```diff
@@ -0,0 +1,259 @@
...
+      }
+
+      if (window.__swc?.DEBUG) {
+        this.warnMissingAccessibleName();
```
rubencarvalho, line 273: isn't this running on each updated? sounds like it can fire a lot of times? maybe put a guard on this?

rubencarvalho (dismissed):

2nd-gen/packages/swc/shared/linear-progress-base.css
blunteshwar: @5t3ph What do you suggest here?

2nd-gen/packages/core/mixins/linear-progress-mixin.ts
blunteshwar: Good catch. They need to stay accessible from subclass render templates (Meter.render() reads clampedValue, fillPercent, formattedValue, sanitizedMin/sanitizedMax, labelContainerId, descriptionContainerId, and the hasLabelSlotContent / hasDescriptionSlotContent flags), so I kept them public rather than protected for ergonomics. Marked each with @internal JSDoc — same pattern as slotContentIsPresent on ObserveSlotPresence — to signal they are not part of the public custom-element API.

2nd-gen/packages/swc/components/meter/Meter.ts
blunteshwar: Yes, it is — variant declares a default of 'informative' and is never undefined. Guard was vestigial from a copy of the Badge classMap (where fixed is genuinely optional). Dropped the ```typeof !== 'undefined'```

2nd-gen/packages/swc/components/meter/Meter.ts
blunteshwar: Agreed, and the same applies to `sanitizedMin/sanitizedMax/clampedValue/fillPercent` — each getter either allocates an Intl.NumberFormat or rescans the slot presence map. Cached all of them into locals at the top of `render()`  so each one is evaluated once per paint.

2nd-gen/packages/core/mixins/linear-progress-mixin.ts
blunteshwar: You are right. LanguageResolutionController already calls `host.requestUpdate(languageResolverUpdatedSymbol, ...)`, which schedules an update; by the time updated() runs, the current render has already happened with the new locale (the template reads this.languageResolver.language lazily via formattedValue). The explicit requestUpdate() I added was scheduling a redundant second cycle. Locale changes now drive a single render via the controller's own request.
Verified — first render is already correct without a willUpdate hook. LanguageResolutionController is instantiated in the constructor and resolves the language synchronously during host setup; formattedValue reads this.languageResolver.language lazily inside render(), so the very first render sees the resolved locale. No precomputed state to cache in willUpdate. If we ever move the formatter to a memoized instance, I'll revisit and hoist it into willUpdate. Added an inline comment in updated() documenting the rationale so the next reader doesn't add requestUpdate() back.

2nd-gen/packages/core/mixins/linear-progress-mixin.ts
blunteshwar: Gated the call on `changes.has('accessibleLabel')` plus a `_hasWarnedNoAccessibleName` first-run flag, mirroring how `ProgressCircleBase` scopes its missing-name warning under `changes.has('label')`. Warning now fires on the first render with no resolvable name (covering initial mount) and again only if accessibleLabel toggles — not on every unrelated property update. The flag resets when a name appears so a later regression can still surface.

2nd-gen/packages/swc/shared/linear-progress-base.css
5t3ph: We have a couple other needs for shared styles, one is Button which I had just left in the main component as `button-base` but the use case/need is a little different since all other buttons do extend from it.

In this case, there are styles and a mixin, which is named and organized independently.

I woud say three main changes:
- move the stylesheet to `2nd-gen/packages/swc/stylesheets/shared` (where `shared` will be a new directory)
- use a base CSS class that matches the mixin name, so update this stylesheet to use `.swc-LinearProgress` instead of "meter"
    - keep individual component class names based on the component for component-specific styles, that way we can tell by class name where styles originate
- adjust the CSS custom property exposure to only properties that change per size or for things like static color, and otherwise use the expected token value in the shared styles without having to do so many property assignments in the individual components

2nd-gen/packages/swc/components/meter/meter.css
```diff
@@ -0,0 +1,54 @@
...
+
+/*
+ * Meter-specific styles. Shared bar/track/fill structure, sizes,
+ * static-color rules, and forced-colors live in
+ * `swc/shared/linear-progress-base.css`. This file owns the default
+ * `--swc-meter-*` token defaults and the variant fill-color overrides.
+ */
+
+:host {
+  --swc-meter-thickness: token("progress-bar-thickness-medium");
```
5t3ph: Noted in the shared styles as well, but these should be assigned in there since they are the same for both meter and progress bar.

_If_ any are unique to meter, move the assignment off `:host` and to `.swc-Meter` instead to follow our guidelines.

2nd-gen/packages/swc/shared/linear-progress-base.css
```diff
@@ -0,0 +1,162 @@
...
+  line-height: var(--swc-meter-line-height);
+  color: var(--swc-meter-text-color);
+}
+
+/* Side label */
+.swc-Meter--sideLabel {
+  grid-template-areas:
+    "label track value"
+    "description description description";
+  grid-template-columns: auto 1fr auto;
```
5t3ph: Let's update this to ensure the meter isn't allowed to shrink out of existence:

```suggestion
  grid-template-columns: auto minmax(30%, 1fr) auto;
```

2nd-gen/packages/swc/components/meter/meter.css
```diff
@@ -0,0 +1,54 @@
...
+  --swc-meter-line-height: token("line-height-100");
+  --swc-meter-line-height-cjk: token("cjk-line-height-100");
+  --swc-meter-top-to-text: token("component-top-to-text-100");
+  --swc-meter-label-to-track: token("spacing-75");
+  --swc-meter-label-to-value-spacing: token("spacing-200");
+  --swc-meter-description-spacing: token("spacing-75");
+}
+
+/* Variants */
+.swc-Meter--positive {
```
5t3ph: Move to use the `:host` selector, such as `:host([variant="positive"])`, for each.

2nd-gen/packages/swc/components/meter/Meter.ts
```diff
@@ -0,0 +1,140 @@
...
+      !hasLabel && this.accessibleLabel ? this.accessibleLabel : undefined;
+    const ariaDescribedBy = hasDescription
+      ? this.descriptionContainerId
+      : undefined;
+
+    return html`
+      <div
+        class=${classMap({
+          ['swc-Meter']: true,
+          [`swc-Meter--${this.variant}`]: true,
```
5t3ph: Remove, as noted in the CSS the `:host()` should be used instead of classes for the variants.
```suggestion
```

2nd-gen/packages/swc/components/meter/stories/meter.stories.ts
```diff
@@ -0,0 +1,316 @@
...
+// Custom min / max range. Plan B3 surfaces arbitrary numeric ranges; the
+// rendered fill + aria-valuetext use the sanitized range.
+export const CustomRange: Story = {
+  render: (args) => html`
+    ${template({
+      ...args,
+      'min-value': 0,
+      'max-value': 10,
+      value: 3,
+      'label-slot': '3 of 10 inputs filled',
```
5t3ph:
```suggestion
      'label-slot': 'Inputs filled',
```

2nd-gen/packages/swc/components/meter/stories/meter.stories.ts
```diff
@@ -0,0 +1,316 @@
...
+
+// `value-label` overrides the auto-formatted percentage in both the
+// rendered value cell and aria-valuetext.
+export const ValueLabel: Story = {
+  render: (args) => html`
+    ${template({
+      ...args,
+      value: 40,
+      'label-slot': 'Storage used',
+      'value-label': '2 GB / 5 GB',
```
5t3ph: Good opportunity to test/show decimal values:

```suggestion
      value: 2.4,
      maxValue: 5,
      'label-slot': 'Storage used',
      'value-label': '2.4 GB / 5 GB',
```

5t3ph (commented):
A few things mostly around styles!

2nd-gen/packages/swc/components/meter/stories/meter.stories.ts
blunteshwar: Done!

2nd-gen/packages/swc/components/meter/stories/meter.stories.ts
blunteshwar: Done!

2nd-gen/packages/swc/components/meter/Meter.ts
blunteshwar: Done!

2nd-gen/packages/swc/components/meter/meter.css
blunteshwar: Done!

2nd-gen/packages/swc/shared/linear-progress-base.css
blunteshwar: Done!

2nd-gen/packages/swc/components/meter/meter.css
blunteshwar: Done!

blunteshwar:
Mixin vs Abstract Class

I think a mixin is the better fit here because this behavior isn't inherently "Meter" behavior — it's shared linear-progress behavior (range sanitization, value formatting, label/description slot tracking, accessible-name validation, etc.) that can be reused by other linear progress-based components.

Using a mixin keeps this functionality composable and avoids forcing components into a specific inheritance hierarchy. With an abstract base class we'd consume our single inheritance slot immediately, making it harder to combine with other base classes in the future. This also aligns with the existing SWC architecture, which already favors composition through mixins such as `SizedMixin`, `DisabledMixin`, and `ObserveSlotPresence`.

Even though Meter is the first consumer, the extracted logic is generic enough that it can be shared by future linear-progress-style components without duplicating implementation.

.codex/hooks.json
```diff
@@ -0,0 +1,15 @@
+{
```
5t3ph: should this be committed?

2nd-gen/packages/swc/components/meter/meter.css
```diff
@@ -11,44 +11,20 @@
...
-  --swc-meter-top-to-text: token("component-top-to-text-100");
-  --swc-meter-label-to-track: token("spacing-75");
-  --swc-meter-label-to-value-spacing: token("spacing-200");
-  --swc-meter-description-spacing: token("spacing-75");
-}
-
-/* Variants */
-.swc-Meter--positive {
-  --swc-meter-fill-color: token("positive-visual-color");
+:host([variant="positive"]) .swc-LinearProgress-fill {
```
5t3ph: Bit of a misunderstanding on my feedback, here's what I expect for the output across these variant rules:

```css
:host([variant="positive"]) {
  --swc-linear-progress-fill-color: token("positive-visual-color");
}
```

2nd-gen/packages/swc/stylesheets/shared/linear-progress-base.css
```diff
@@ -0,0 +1,208 @@
...
+ * future `<swc-progress-bar>`). Selectors use the `.swc-LinearProgress`
+ * namespace so each component's own stylesheet can layer component-
+ * specific rules on top via `.swc-Meter`, `.swc-ProgressBar`, etc.
+ *
+ * Tokens are read directly via `token("…")`; the component stylesheet
+ * only needs to expose a custom property when consumers genuinely need
+ * to override it. Size, variant, label-position, and static-color
+ * modifiers are driven by attribute selectors on `:host()`.
+ */
+
```
5t3ph, line 20: Misunderstanding on my custom property feedback throughout this file as well. Continue to expose custom properties when they are modified for variants. The main change I expected to see was inlining `token()` values here rather than delaying assignment to the component for values that are not component-specific, but not a total removal of exposed custom properties.

Have the agent refer to these docs to adjust the custom property exposure in this stylesheet: `CONTRIBUTOR-DOCS/02_style-guide/01_css/02_custom-properties.md#component-custom-property-exposure`

5t3ph (commented):
Some adjustment needed on the custom property strategy, sorry if my feedback before was confusing!

2nd-gen/packages/swc/stylesheets/shared/linear-progress-base.css
blunteshwar: Got it!

5t3ph (approved):
LGTM! :shipit:

2nd-gen/packages/core/mixins/linear-progress-mixin.ts
```diff
@@ -0,0 +1,301 @@
...
+     */
+    @property({ type: Number, reflect: true, attribute: 'max-value' })
+    public maxValue = 100;
+
+    /**
+     * Rare-case accessible-name fallback used when there is no visible
+     * `label` slot content (for example, a data grid of meters).
+     * Renders into `aria-label` on the role element when set.
+     */
+    @property({ type: String, reflect: true, attribute: 'accessible-label' })
```
nikkimk: Does this need to be reflected?

nikkimk (approved):
Just one nit/question

2nd-gen/packages/core/mixins/linear-progress-mixin.ts
blunteshwar: Yeah, you are right, as per our style guide we should reflect only when a property has a visual effect when present or updated. Thus removing the `reflect: true`

2nd-gen/packages/swc/stylesheets/shared/linear-progress-base.css
```diff
@@ -0,0 +1,194 @@
...
+  border-radius: var(--_swc-linear-progress-corner-radius);
+  overflow: hidden;
+}
+
+.swc-LinearProgress-fill {
+  block-size: var(--swc-linear-progress-thickness, token("progress-bar-thickness-medium"));
+  background: var(--swc-linear-progress-fill-color, token("accent-content-color-default"));
+  border: none;
+  border-radius: var(--_swc-linear-progress-corner-radius);
+  transition: inline-size 1s;
```
rubencarvalho, line 92: Should we reset this in `@prefers-reduced-motion`?

PR #6317, opened 2026-05-18
feat(storybook): testing grid WIP

aramos-adobe:
## Description

Adds a **Storybook testing grid** for 2nd-gen SWC, with Button as the reference implementation. The grid powers local VRT preview and Chromatic captures, separate from docs-friendly single-instance stories.

### Storybook infrastructure

- **`testing-grid/`** — Modular Lit helpers for VRT matrices (`Variants`, `Container`, `States`, `ArgGrid`, `Sizes`, `vrtCase`, …). Public API re-exported from `helpers/testing-grid.ts`.
- **`testing-preview.ts` decorator** + **Testing preview** toolbar global (beaker icon) — maps to `parameters.showTestingGrid` for local grid preview.
- **`pseudo-states-helpers.ts`** + **`decorators/pseudo-states.ts`** — Shadow-root pseudo-state augmentation and `data-vrt-*` cell patching for forced hover/focus/active captures.
- **`main.ts`** — Play-function tests stay under `**/test/*.test.ts`; VRT case lists live in `**/test/*.vrt.ts` (imported by `*.stories.ts`, not indexed as separate CSF files).
- **`preview.ts`** — Global `chromatic.disableSnapshot: true` with per-grid opt-in; `autoplay: isChromatic()` so `play()` runs before Chromatic snapshots.
- **Docs** in `.storybook/helpers/README.md` for the VRT file pattern and module layout.
- **Unit tests** — `.storybook/helpers/test/testing-grid.unit.test.ts` (`yarn test:unit`, project `swc-unit`).

### Button reference implementation

- **`button.template.ts`** — Pure Lit template for grid cells (avoids `@wc-toolkit/storybook-helpers` `template()`, which calls `useArgs()` per cell and triggers React hooks errors in large grids).
- **`button.vrt.ts`** — VRT case list (`testData`, `stateData`) imported by stories; not a standalone Storybook entry.
- **`button.stories.ts`** — **VRT Grid** story wired to `ButtonVRTRender`.
- **Grid layout** — Grouped **by variant** (Primary → Secondary → Accent → Negative); within each variant, columns for **Default / Disabled / Hovered / Focused / Active / Pending**, then fill-style × content (label, icon+label, icon-only). Static white/black sections only show **primary** and **secondary** (static-color API limit).
- **VRT cell contract** — Cells declare `data-vrt-host`, `data-vrt-control`, optional `data-vrt-state`, and `data-vrt-layout-classes` (icon-only layout). Forced states use class-based pseudo rules (`.is-hover`, `.is-focus-visible`, `.is-active`), not component attributes.

---

## Testing grid

The testing grid renders a full variant × state × size matrix for Chromatic VRT. In local Storybook it stays hidden until **Testing preview** is enabled (or Chromatic runs). Docs stories continue to show a single instance.

**Per-component setup** (see `helpers/README.md`):

1. `components/<name>/stories/<name>.template.ts` — pure Lit `Template()` for grid cells
2. `components/<name>/test/<name>.vrt.ts` — `Variants()` config (`testData`, `stateData`, nested `ArgGrid` / `Container` builders)
3. `<name>.stories.ts` — `VRTGrid` story with `TESTING_GRID_STORY_NAME` + `TESTING_GRID_STORY_PARAMETERS`

### Module layout (`.storybook/helpers/testing-grid/`)

| Module | What it does |
|--------|----------------|
| **`testing-grid.ts`** (barrel) | Re-exports the public API. Import from `helpers/index.js` or `helpers/testing-grid.js`. |
| **`constants.ts`** | Shared VRT constants: `TESTING_GRID_STORY_NAME` (`'VRT Grid'`), `TESTING_GRID_STORY_PARAMETERS` (Chromatic opt-in, pending delay), grid border token, human-readable `SIZE_LABELS`. |
| **`types.ts`** | TypeScript contracts for grid templates and configs — `GridTemplateFn`, `VariantsConfig`, `ContainerProps`, `StateItem`, `TestCaseItem`, builder prop types. |
| **`internal.ts`** | Internal utilities: `readStaticColor`, `wrapWithStaticColorDemo` (gradient panel for static white/black sections), `isDarkTheme`, `getRandomId`. |
| **`primitives.ts`** | Layout building blocks — `Heading` (SWC typography labels, `chromatic-ignore`), `Container` (nested bordered flex sections; `level` controls scale/spacing), `renderContent` (recursive nested content). |
| **`builders.ts`** | Composable matrix builders — `States` (interaction-state columns), `ArgGrid` (sweep one arg across options), `Sizes` (`size` preset), `vrtCase` (one-off rows like line wrap / truncation with `withStates: false`). |
| **`variants.ts`** | **`Variants()`** — main entry point. Returns a Storybook `render` function: single `data-html-preview` in docs/dev, full `data-testing-preview` matrix when Testing preview is on or under Chromatic. Orchestrates `testData` sections, `stateData` matrices, optional `Sizes` row, and static-color wrapping. |

**Related helpers** (outside `testing-grid/`):

| Module | What it does |
|--------|----------------|
| **`pseudo-states-helpers.ts`** | Applies forced interaction classes via `data-vrt-host` / `data-vrt-control` / `data-vrt-state`; `applyTestingGridPseudoStates` for VRT story `play` functions. |
| **`pseudo-states.ts`** | Augments shadow-root stylesheets so `:hover` / `:focus-visible` / etc. map to `.is-*` classes. |
| **`decorators/pseudo-states.ts`** | Global decorator — re-applies pseudo-state patches after each story render. |
| **`decorators/testing-preview.ts`** | Maps toolbar **Testing preview** to `parameters.showTestingGrid`. |

### VRT cell wrapper attributes (`<name>.template.ts`)

Each grid cell wraps the component in a `div.vrt-cell` with `data-vrt-*` attributes. These are **Storybook/VRT only** — not part of the public component API. `applyTestingGridPseudoStates` (VRT story `play` function) reads them and patches the correct shadow-DOM node after render.

| Attribute | Required | Purpose |
| --------- | -------- | ------- |
| `data-vrt-host` | Yes | Custom-element tag inside the wrapper (e.g. `swc-button`). Used to `querySelector` the host and to `customElements.whenDefined` before patching. |
| `data-vrt-control` | Yes | CSS selector for the **internal** element that receives forced classes — typically the root class inside the host's shadow root (e.g. `.swc-Button`). Omit only if the host element itself is the control. |
| `data-vrt-state` | No | Interaction state to force: `hover`, `focus`, or `active`. Omit for default and disabled cells. Maps to `.is-hover`, `.is-focus-visible`, and `.is-active` on the control. Disabled state uses the component's `disabled` attribute instead. |
| `data-vrt-layout-classes` | No | Space-separated classes applied to the control when layout depends on slot detection or other async work (e.g. `swc-Button--hasIcon swc-Button--iconOnly` for icon-only cells). Ensures the correct layout is captured before Chromatic snapshots. |

Example (Button):

```html
<div
  class="vrt-cell"
  data-vrt-host="swc-button"
  data-vrt-control=".swc-Button"
  data-vrt-state="hover"
  data-vrt-layout-classes="swc-Button--hasIcon swc-Button--iconOnly"
>
  <swc-button>...</swc-button>
</div>
```

---

## Motivation and context

2nd-gen SWC had hand-rolled option stories (`Sizes`, `Variants`, etc.) but no shared VRT grid or Chromatic-oriented layout. This PR establishes the shared helper layer and proves the pattern on Button. SWC is the long-term home for this workflow as Spectrum CSS sunsets.

---

### Local Storybook

```bash
cd 2nd-gen/packages/swc
yarn storybook
```

1. Open **Button → VRT Grid** (dev-only story).
2. In the toolbar, open the **beaker** menu → **Show testing preview**.
3. Confirm the full matrix appears (`data-testing-preview` wrapper with bordered sections).
4. Toggle back to **Default mode** — single preview in `data-html-preview` (docs-style).
5. Spot-check:
   - **Primary / Secondary / Accent / Negative** sections, each with state columns.
   - **Static white** / **Static black** — only Primary + Secondary.
   - **Sizing** row at the bottom.
   - Hovered / Focused / Active rows show forced visuals (not real pointer/focus).

### Unit tests

```bash
cd 2nd-gen/packages/swc
yarn test:unit
```

### Chromatic / PR preview

- [2nd-gen Storybook preview](https://swcpreviews.z13.web.core.windows.net/pr-6317/docs/second-gen-storybook/)
- Label **`run_vrt`** is on the PR for 2nd-gen Chromatic when ready.

### Regression spot-checks

- **Button** docs stories (Playground, Sizes, Variants, etc.) — unchanged behavior.
- **Color loupe → Parent driven visibility** — still renders trigger button.

---

## Screenshots

_Add VRT Grid + Testing preview on/off screenshots before review._

---

## Open questions for the team

1. **Default story for Chromatic** — Should **VRT Grid** become the Chromatic default for Button, or stay a separate dev story while Playground remains docs default?
2. **Rollout order** — Which 2nd-gen components should adopt grids next? (Badge, Status light, and other variant-heavy components are likely candidates.)
3. **Changeset scope** — Patch `@adobe/spectrum-wc` for Storybook helpers only, or also document the VRT pattern in contributor docs?

---

## Known gaps / follow-ups

| Gap | Notes |
|-----|--------|
| **Single reference component** | Only Button has `*.vrt.ts` + grid; pattern not yet applied elsewhere. |
| **Accent / Negative + outline** | Grid still renders outline column for all variants; component warns and falls back — filter fill-style per variant? |
| **Pending row timing** | `delay: 1100` in grid Chromatic params; real 1s pending animation not fully exercised in all cells. |
| **Button-specific pseudo-state patching** | `data-vrt-host` / `data-vrt-control` are generic, but Button is the first consumer — validate pattern on the next component. |
| **No changeset yet** | changeset-bot flagged missing changeset — add before merge if we publish Storybook/helper changes. |

---

## Author's checklist

- [x] I have reviewed Accessibility Practices for this feature (N/A for Storybook infra).
- [x] I have added automated tests to cover my changes (`testing-grid.unit.test.ts`).
- [x] I have included updated documentation if my change required it (`helpers/README.md`).

---

## Reviewer's checklist

- [ ] Focus review on `.storybook/helpers/testing-grid/`, `decorators/testing-preview.ts`, `pseudo-states-helpers.ts`, and `components/button/test/button.vrt.ts` / `button.template.ts`.
- [ ] Validate VRT Grid in PR Storybook preview with **Testing preview** enabled.
- [ ] Approve 2nd-gen Chromatic run when `run_vrt` label is applied.

### Manual review test cases

- [ ] **VRT Grid — Button**
  1. Open [PR 2nd-gen Storybook](https://swcpreviews.z13.web.core.windows.net/pr-6317/docs/second-gen-storybook/) → **Button → VRT Grid**
  2. Enable toolbar **Testing preview → Show testing preview**
  3. Expect variant-grouped matrix with state columns and Sizing section; static sections show Primary/Secondary only.

- [ ] **Testing preview toggle**
  1. Same story, toggle **Default mode**
  2. Expect single button preview (no matrix).

### Device review

- [ ] Desktop (primary target for Storybook review)
- [ ] Mobile / iPad (optional — grid is desktop-oriented VRT)

N/A for Storybook infrastructure. VRT forced states only affect visual presentation in Storybook and do not change runtime a11y behavior for consumers.


2nd-gen/packages/swc/.storybook/decorators/testing-preview.ts
```diff
@@ -0,0 +1,44 @@
...
+
+import { makeDecorator } from '@storybook/preview-api';
+import type { DecoratorFunction } from '@storybook/types';
+
+import { isChromatic } from '../helpers/testing-grid.js';
+
+/**
+ * Maps the "Testing preview" toolbar global to `parameters.showTestingGrid`, matching
+ * spectrum-css `.storybook/decorators/testing-preview.js`.
+ */
```
caseyisonit, line 24: Can we also copy over the comments and description for this from that file that way we can remove this reference and its fully migrated?

https://github.com/adobe/spectrum-css/blob/main/.storybook/decorators/testing-preview.js

2nd-gen/packages/swc/components/button/stories/button.test.ts
caseyisonit, line 1: not loving the idea of another test file with the same file name in the stories directory, could we name it something like button.vrt.ts and keep it in the tests folder?

2nd-gen/packages/swc/components/button/stories/button.template.ts
caseyisonit, line 1: im curious why we need this file if we can get a configured template from wc-toolkit? Is this that default story thing you mentioned in slack?

2nd-gen/packages/swc/components/button/stories/button.stories.ts
caseyisonit, line 1: should we set `{ disableSnapshot: true }` as the default in the global config and you just have to opt in to `{ disableSnapshot: false }` once in the stories?

2nd-gen/packages/swc/components/button/stories/button.stories.ts
```diff
@@ -392,5 +423,43 @@ export const Accessibility: Story = {
...
+ * - **Form submission** (`type="submit"` / `type="reset"`): Native form participation.
+ *   Until then, use a native `<button type="submit">` with
+ *   [global button styles](/docs/guides-customization-global-element-styling--readme).
+ * - **`genai` variant**: For generative AI actions.
+ * - **`premium` variant**: For premium or upgrade flows.
+ */
+export const UpcomingFeatures: Story = {
+  tags: ['upcoming'],
 };
+UpcomingFeatures.storyName = 'Upcoming features';
```
caseyisonit: i think this might be legacy code from an old rebase. this content should be represented in the MDX file now. if it is can we remove this?

2nd-gen/packages/swc/components/button/stories/button.stories.ts
```diff
@@ -392,5 +423,43 @@ export const Accessibility: Story = {
...
+export const ButtonGroups: Story = {
+  render: buttonGroupsRender as Story['render'],
+  parameters: {
+    layout: 'centered',
+    flexLayout: false,
+    ...TESTING_GRID_STORY_PARAMETERS,
+  },
+  tags: ['!autodocs', 'dev'],
+};
+ButtonGroups.storyName = TESTING_GRID_STORY_NAME;
```
caseyisonit: This is confusing since we have a Button Group component. Could we name this story TestingGridStoryName and update the decorator to use that instead of having to alias it?

Im also curious if this could live in the button.vrt.ts test file so the context is all together? or does chromatic look for only stories.ts files? is this configurable?

2nd-gen/packages/swc/components/button/stories/button.test.ts
caseyisonit, line 1: Can any of our document stories be used in the vrt grid so we are also testing what we are rendering in production? it should potentially reduce the complexity of this file since those are already set up with args

2nd-gen/packages/swc/components/button/stories/button.test.ts
```diff
@@ -0,0 +1,336 @@
...
+      level: 1,
+      ...args,
+    },
+    context
+  );
+
+/**
+ * Full variant × treatment × content × state matrix for Chromatic / VRT.
+ */
+export const ButtonGroups = withVariantsGrid({
```
caseyisonit: Recommend changing this name to `ButtonVRTRender`, same confusion as mentioned about with Button Group the component

2nd-gen/packages/swc/components/button/Button.ts
```diff
@@ -119,6 +122,27 @@ export class Button extends ButtonBase {
...
+   */
+  @property({ type: Boolean, reflect: true, attribute: 'vrt-pending-active' })
+  public vrtPendingActive: boolean = false;
+
+  /**
+   * Forces icon-only visuals in VRT grids when light-DOM slot detection lags.
+   * Not supported for production use.
+   */
+  @property({ type: Boolean, reflect: true, attribute: 'vrt-icon-only' })
+  public vrtIconOnly: boolean = false;
```
Rajdeepc: Blocking: You can't do this. 
These are reflected and they will appear in custom-elements.json and public element API surface despite the JSDoc disclaimer. Every components that adds testing grid will need to repeat these attributes. This is not scalable.
You should move these force state logic to Storybook only CSS stylesheets or a storybook decorator not in component class.
I would love to talk on this coz if these are needed we need to properly mark them and gate it.

2nd-gen/packages/swc/chromatic.config.json
```diff
@@ -1,21 +1,8 @@
 {
   "buildScriptName": "storybook:build",
   "exitOnceUploaded": true,
-  "onlyChanged": true,
+  "onlyStoryNames": ["**/Default"],
```
Rajdeepc: Are you only snapshoting stories named Default? What about existing stories. This is a regression risk.
This would be my recommendation:

1. Keep onlyChanged/traceChanged as-is for now
2. Add onlyStoryNames only when >=3 components have grids (establish the pattern first, then cut over)
3. Or use a dual approach: disableSnapshot: true globally (as done here) with per-story opt-in, without removing onlyChanged (they're compatible)

2nd-gen/packages/swc/components/button/stories/button.template.ts
```diff
@@ -0,0 +1,97 @@
...
+      pending-label=${ifDefined(pendingLabel)}
+      ?truncate=${truncate}
+      ?justified=${justified}
+      accessible-label=${ifDefined(accessibleLabel)}
+      vrt-state=${ifDefined(vrtState)}
+      ?vrt-pending-active=${vrtPendingActive}
+      ?vrt-icon-only=${iconOnly}
+      style=${ifDefined(style)}
+    >
+      ${iconSlot ? unsafeHTML(iconSlot) : nothing}${labelContent}
```
Rajdeepc: This can execute arbitrary HTML and if this pattern propagates to other templates (or if iconSlot ever comes from user-controlled args in Storybook), it's an XSS vector in the Storybook preview.
You can use a pre-rendered TemplateResult for the icon instead of raw string + unsafeHTML.

2nd-gen/packages/swc/components/button/stories/button.stories.ts
Rajdeepc: Yes we should remove per-story `disabledSnapshot` The global already handles it. I feel only the testing grid story needs `disableSnapshot: false` to opt-in?

2nd-gen/packages/swc/.storybook/decorators/testing-preview.ts
```diff
@@ -0,0 +1,44 @@
...
+    } = context;
+
+    if (isChromatic()) {
+      context.parameters.showTestingGrid = true;
+    } else if (viewMode === 'docs') {
+      context.parameters.showTestingGrid = false;
+    } else if (typeof showTestingGrid === 'undefined') {
+      context.parameters.showTestingGrid = testingPreview;
+    } else if (showTestingGrid !== testingPreview) {
+      context.parameters.showTestingGrid = testingPreview;
```
Rajdeepc: This is redundant. The preceding `else if (typeof showTestingGrid === 'undefined')` already handles the unset case, and if it's defined and differs, the next branch assigns.

2nd-gen/packages/swc/stylesheets/global/global-button.css
```diff
@@ -52,35 +52,42 @@
...
+.swc-Button:disabled:is(*, :hover) {
+  color: var(--swc-button-content-color-disabled, token("disabled-content-color"));
+  background-color: var(--swc-button-background-color-disabled, token("disabled-background-color"));
+  border-color: var(--swc-button-border-color-disabled, transparent);
+  transform: none;
+}
+
+.swc-Button--disabled .swc-Button-icon {
+  pointer-events: none;
+}
```
Rajdeepc, line 72: Is this generated by `classMap`? Maybe dead code?

Rajdeepc (changes requested):
Good going!
I am expecting to see at minimum tests for `testing-grid.ts` as this piece of infrastructure will be depended upon by every component. 
Can you start with unit tests for `isChromatic()` , `renderContent()`, `heading()` , `Container()` and a smoke test that renders `Variants()` should be good.

If you want to discuss on the blocking vectors we can chat!

2nd-gen/packages/swc/stylesheets/global/global-button.css
aramos-adobe: Oh yes this is dead code and unrelated to this PR

2nd-gen/packages/swc/components/button/button.css
```diff
@@ -57,19 +57,22 @@
   transition-property: outline, border-color, color, background-color, transform;
 }
 
-.swc-Button:hover {
+.swc-Button:hover,
+.swc-Button.is-hover {
```
5t3ph: I'm sharing a possible storybook decorator solution with @aramos-adobe to add these in on the fly through Storybook instead of bake-in to component stylesheets.

2nd-gen/packages/swc/.storybook/helpers/testing-grid.ts
```diff
@@ -0,0 +1,748 @@
...
+    weight,
+    withMargin = false,
+  }: HeadingProps = {},
+  context: GridStoryContext = {} as GridStoryContext
+): TemplateResult | typeof nothing {
+  if (!content) {
+    return nothing;
+  }
+
+  const headingStyles: Record<string, string> = {
```
5t3ph: Typography styles should be loaded, so can you use existing classes instead? The dark/light mode handling might be excessive, too, since that color should already be inherited.

2nd-gen/packages/swc/chromatic.config.json
aramos-adobe: I like the dual approach here. Going to try it out

2nd-gen/packages/swc/components/button/stories/button.template.ts
aramos-adobe: Okay! That's a good call

2nd-gen/packages/swc/components/button/stories/button.template.ts
aramos-adobe: @caseyisonit Since there a lot of items per cell in the grid, using wc-toolkit in each cell would invoke `template()` → `useArgs()` again and again, which violates React’s rules of hooks and produces the “Rendered more hooks than during the previous render” error we hit during development.

This template file is a lightweight Lit template file is VRT only where you have the creative freedom to create a sample or group of component artifacts to display in the grid effectively.

2nd-gen/packages/swc/.storybook/decorators/static-colors-demo.ts
```diff
@@ -15,12 +15,16 @@ import { styleMap } from 'lit/directives/style-map.js';
...
-  black: 'linear-gradient(45deg, rgb(255 241 246), rgb(238 245 255))',
+/** Gradient backgrounds for static-color demos — matching spectrum-css. */
+export const STATIC_COLOR_DEMO_BACKGROUNDS = {
   white: 'linear-gradient(45deg, rgb(64 0 22), rgb(14 24 67))',
+  black: 'linear-gradient(45deg, rgb(255 241 246), rgb(238 245 255))',
+} as const;
+
+/** Foreground text color on each static-color demo background. */
+export const STATIC_COLOR_DEMO_FOREGROUNDS = {
+  white: 'white',
```
caseyisonit, line 26: :chef-kiss:

2nd-gen/packages/swc/.storybook/helpers/README.md
```diff
@@ -48,6 +48,26 @@ const componentName = formatComponentName(meta.title);
 3. Replaces spaces with hyphens for kebab-case
 4. Optionally converts to PascalCase by capitalizing each word and removing hyphens
 
+## Testing grid (`test/*.vrt.ts`)
+
+Spectrum CSS keeps VRT case lists in a sibling module (e.g. `button.test.js`) that is **imported** by `button.stories.ts`, not indexed as its own Storybook file.
```
caseyisonit: Is this a true fact for SWC? If so can we update the language to not reference Spectrum CSS?

2nd-gen/packages/swc/.storybook/decorators/pseudo-states.ts
caseyisonit, line 1: should live in decorator directory

2nd-gen/packages/swc/.storybook/pseudo-states-helpers.ts
caseyisonit, line 1: should live in helpers directory

2nd-gen/packages/swc/.storybook/pseudo-states.ts
caseyisonit, line 1: since this looks to only be used in helpers (correct me if im wrong) we should just roll this in to that file or rename this to a semantically correct helper and store it in the helpers directory along side the file using it

2nd-gen/packages/swc/.vitest-attachments/a860f0712c1bc95442886205fb311f45e2db75cf.png
caseyisonit, line 1: not sure what this file is lol

caseyisonit: might need to add to gitignore?

2nd-gen/packages/swc/components/button/stories/button.template.ts
caseyisonit: thank you for investigating!

2nd-gen/packages/swc/.storybook/helpers/testing-grid.ts
```diff
@@ -0,0 +1,812 @@
...
+    disableSnapshot: false,
+    // Pending cells use `pending: true`; wait for pendingActive (1s) before capture.
+    delay: 1100,
+    prefersReducedMotion: 'no-preference',
+    pauseAnimationAtEnd: true,
+  },
+} as const;
+
+/** Same detection logic as `chromatic/isChromatic` (browser Storybook + CI). */
+export function isChromatic(windowArgument?: Window): boolean {
```
5t3ph: Why not import the existing one?

`import isChromatic from 'chromatic/isChromatic';`

2nd-gen/packages/swc/.storybook/helpers/testing-grid.ts
```diff
@@ -0,0 +1,812 @@
...
+  const w =
+    windowArgument ?? (typeof window !== 'undefined' ? window : undefined);
+  return !!(
+    w &&
+    (/Chromatic/.test(w.navigator.userAgent) ||
+      /chromatic=true/.test(w.location.href))
+  );
+}
+
+function capitalize(str: string): string {
```
5t3ph: Import from `2nd-gen/packages/core/utils/capitalize.ts`

2nd-gen/packages/swc/.storybook/helpers/testing-grid.ts
```diff
@@ -0,0 +1,812 @@
...
+  if (!str) {
+    return str;
+  }
+  return str.charAt(0).toUpperCase() + str.slice(1);
+}
```
caseyisonit: 1. we have a capitalize util in core we can reuse
2. the if statement is confusing me, im reading this as if str is false meaning doesnt exist, wasnt passed in.... the there would be nothing for the return to send. Im not seeing a need for this check or the return needs to be empty OR it should throw a warning that str is required?

2nd-gen/packages/swc/.storybook/preview.ts
```diff
@@ -121,27 +125,46 @@ const preview = {
...
     withFlexLayout,
   ],
   parameters: {
     layout: 'centered',
-    // The Storybook patch (.yarn/patches/storybook-npm-10.4.1.patch) gates play
-    // functions on initial render behind `parameters.autoplay` so they do not run
-    // automatically while browsing the dev UI. Chromatic renders each story via that
-    // same initial path, so without this it would snapshot the pre-play state. Enable
-    // autoplay only under Chromatic to restore correct visual regression snapshots.
-    autoplay: isChromatic(),
```
5t3ph, line 145: This needs added back - it's required for Tooltip and any other story that relies on `play()` for correct snapshots.

2nd-gen/packages/swc/.storybook/pseudo-states-helpers.ts
```diff
@@ -0,0 +1,147 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import { augmentTree } from './pseudo-states.js';
+
+export type PseudoState =
```
5t3ph: Do you need `disabled` and `focus-visible` here? It looks like you're relying on component attributes for `disabled` and normalizing to just `focus`.

2nd-gen/packages/swc/.storybook/pseudo-states-helpers.ts
```diff
@@ -0,0 +1,147 @@
...
+  canvasElement: ParentNode,
+  hostSelector: string,
+  innerSelector: string,
+  state: PseudoState
+): void {
+  const host = canvasElement.querySelector(hostSelector);
+  host?.shadowRoot?.querySelector(innerSelector)?.classList.add(`is-${state}`);
+}
+
+function getInternalButton(host: Element): HTMLButtonElement | null {
```
5t3ph: I assume this is an outstanding question, but this will need generalized somehow - maybe consuming component will need to pass the right element, or maybe it can somehow pickup on the component name to find `.swc-${componentName}`? Same for other "button" specific portions.

2nd-gen/packages/swc/.storybook/decorators/static-colors-demo.ts
```diff
@@ -54,13 +58,13 @@ export const withStaticColorsDemo: DecoratorFunction = makeDecorator({
         }
 
         .spectrum-examples-static-colors > *:first-child {
-          color: white;
-          background: ${staticColorSettings.white};
+          color: ${STATIC_COLOR_DEMO_FOREGROUNDS.white};
```
5t3ph, line 61: I think this is maybe somehow getting applied outside of the demo? Or maybe from another location? Highlighting the subtitle span that shouldn't be changing color:

Image

2nd-gen/packages/swc/components/button/button-base.css
```diff
@@ -42,7 +42,8 @@
   user-select: none;
 }
 
-.swc-Button:focus-visible {
+.swc-Button:focus-visible,
+.swc-Button.is-focus-visible {
```
5t3ph: This can be removed now, right?

```suggestion
.swc-Button:focus-visible {
```

5t3ph (commented):
Really good work to get it this far! 👏

2nd-gen/packages/swc/components/button/button-base.css
aramos-adobe: @5t3ph That is correct

2nd-gen/packages/swc/.vitest-attachments/a860f0712c1bc95442886205fb311f45e2db75cf.png
aramos-adobe: @caseyisonit Don't need this. It was generated after vite unit test was run for helpers in test grid

2nd-gen/packages/swc/.storybook/helpers/README.md
aramos-adobe: Removed all mention of Spectrum CSS and making sure we're handling this as SWC legacy

2nd-gen/packages/swc/.storybook/pseudo-states.ts
```diff
@@ -0,0 +1,127 @@
+/**
+ * Copyright 2026 Adobe. All rights reserved.
```
Rajdeepc, line 2: This is the most complex file in the PR and I am expecting some unit tests here. augmentTree, buildAugmentedSheet, and walkRules constitute the heart of the pseudo-state system. An error here can produce wrong VRT output with no signal. 

Expecting some minimum test coverage here like walkRules with each supported at-rule type, buildAugmentedSheet with a multi-rule sheet, error handling for cross-origin (CORS) stylesheets, and @starting-style falling back via constructor.name check.

2nd-gen/packages/swc/.storybook/helpers/pseudo-states-helpers.ts
```diff
@@ -0,0 +1,179 @@
...
+    augmentTree(scope);
+    applyVrtGridPseudoStates(scope);
+    applyVrtLayoutClasses(scope);
+  };
+
+  run();
+  requestAnimationFrame(() => {
+    run();
+    requestAnimationFrame(run);
+  });
```
Rajdeepc: This is a not a solution. This can be a quick fix. The correct fix should be to await `element.updateComplete` after each play step. Spraying timers makes it impossible to reason about when the DOM is stable, and the next component that renders slower than 500ms (e.g., a component doing async work in firstUpdated) will silently produce stale screenshots.

2nd-gen/packages/swc/.storybook/decorators/pseudo-states.ts
```diff
@@ -0,0 +1,27 @@
...
+
+import { scheduleTestingGridPseudoStates } from '../helpers/pseudo-states-helpers.js';
+
+/**
+ * Injects class-based pseudo-state rules into open shadow roots after each story
+ * renders, then patches testing-grid cells (`data-vrt-state`, `data-vrt-layout-classes`). Runs on
+ * every render so toggling **Testing preview** re-applies classes after the grid mounts.
+ */
+export const pseudoStatesDecorator: DecoratorFunction = (storyFn, context) => {
+  scheduleTestingGridPseudoStates();
```
Rajdeepc: This first call queues timers that will fire on whatever DOM the previous story has left behind. This might be working out of coincidence but its incorrect reasoning. Verify removing this call.

2nd-gen/packages/swc/.storybook/pseudo-states.ts
```diff
@@ -0,0 +1,127 @@
...
+      console.warn('[pseudo-states] Failed to insert rule:', r, e);
+    }
+  });
+  return sheet;
+}
+
+function augmentShadowRoot(root: ShadowRoot): void {
+  if (augmented.has(root)) {
+    return;
+  }
```
Rajdeepc, line 121: This won't effect Button today but its a silent correctness hole for any components that adds style in response to an attribute change. You can document this constraint for now but think of a resetAugmented API also if you can expose.

2nd-gen/packages/swc/.storybook/helpers/testing-grid/internal.ts
```diff
@@ -0,0 +1,81 @@
...
+  `;
+}
+
+/** True when the Storybook theme global is dark or adaptive. */
+export function isDarkTheme(context: StoryContext): boolean {
+  const theme = context.globals?.theme as string | undefined;
+  return theme === 'dark' || theme === 'adaptive';
+}
+
+/** Stable random id for controls that need unique `name` / `id` values across grid cells. */
```
Rajdeepc: nit: in your JSDoc you are saying `getRandomID` is stable which is not . Every call will produce a new API. Best if you want to rename this to `generateUniqueId` or `uniqueId`

2nd-gen/packages/swc/.storybook/helpers/testing-grid/primitives.ts
```diff
@@ -0,0 +1,248 @@
...
+  }
+
+  return html`
+    ${list.map((c) => {
+      if (typeof c === 'undefined') {
+        return nothing;
+      }
+
+      const isLit =
+        typeof c === 'object' && c !== null && '_$litType$' in (c as object);
```
Rajdeepc: `_$litType$` is a mangled internal property, not part of the public Lit API. It could disappear or change in a patch release. Replace this with `c instanceOf TemplateResult`

2nd-gen/packages/swc/.storybook/helpers/testing-grid/builders.ts
```diff
@@ -0,0 +1,233 @@
...
+ * governing permissions and limitations under the License.
+ */
+
+/**
+ * Composable grid builders: state matrices, arg sweeps, and one-off VRT rows.
+ */
+
+import { nothing, type TemplateResult } from 'lit';
+
+import { capitalize } from '@spectrum-web-components/core/utils/capitalize.js';
```
Rajdeepc: A Storybook helper should not depend on component package internals. If this util is needed, copy the one-liner into testing-grid/internal.ts or into a Storybook utils.ts.

Rajdeepc (changes requested):
Architecture looks promising but there are still few blocking issues that needs to be addressed before the final pass. Let me know if you have any questions around these.

2nd-gen/packages/swc/.storybook/helpers/README.md
```diff
@@ -48,6 +48,68 @@ const componentName = formatComponentName(meta.title);
 3. Replaces spaces with hyphens for kebab-case
 4. Optionally converts to PascalCase by capitalizing each word and removing hyphens
 
+## Testing grid (`test/*.vrt.ts`)
```
5t3ph, line 51: Can you also update the `migration-prep` skill to be able to determine if this is needed, then the `migration-styling` skill to be responsible for adding this since it also scaffolds the initial stories, and maybe also a double-check in the `migration-documentation` skill that this exists/is updated?

2nd-gen/packages/swc/.storybook/helpers/README.md
```diff
@@ -48,6 +48,68 @@ const componentName = formatComponentName(meta.title);
...
+| `data-vrt-state`          | No       | Interaction state to force: `hover`, `focus`, or `active`. Omit for default and disabled cells. Maps to `.is-hover`, `.is-focus-visible`, and `.is-active` on the control. Disabled state uses the component's `disabled` attribute instead.       |
+| `data-vrt-layout-classes` | No       | Space-separated classes applied to the control when layout depends on slot detection or other async work (e.g. `swc-Button--hasIcon swc-Button--iconOnly` for icon-only cells). Ensures the correct layout is captured before Chromatic snapshots. |
+
+Example (Button):
+
+```html
+<div
+  class="vrt-cell"
+  data-vrt-host="swc-button"
+  data-vrt-control=".swc-Button"
```
5t3ph, line 84: Question: for something like a combobox where we might want to test states on multiple elements (the initial box, then the separate dropdown trigger, then items within the listbox), could we pass multiple things to test at once, or would it require separate tests? Separate tests might be best, just making sure that is supported without issue.

2nd-gen/packages/swc/.storybook/helpers/README.md
```diff
@@ -48,6 +48,68 @@ const componentName = formatComponentName(meta.title);
...
+| `data-vrt-host`           | Yes      | Custom-element tag inside the wrapper (e.g. `swc-button`). Used to `querySelector` the host and to `customElements.whenDefined` before patching.                                                                                                   |
+| `data-vrt-control`        | Yes      | CSS selector for the **internal** element that receives forced classes — typically the root class inside the host's shadow root (e.g. `.swc-Button`). Omit only if the host element itself is the control.                                         |
+| `data-vrt-state`          | No       | Interaction state to force: `hover`, `focus`, or `active`. Omit for default and disabled cells. Maps to `.is-hover`, `.is-focus-visible`, and `.is-active` on the control. Disabled state uses the component's `disabled` attribute instead.       |
+| `data-vrt-layout-classes` | No       | Space-separated classes applied to the control when layout depends on slot detection or other async work (e.g. `swc-Button--hasIcon swc-Button--iconOnly` for icon-only cells). Ensures the correct layout is captured before Chromatic snapshots. |
```
5t3ph, line 76: Can you explain this a bit more? These classes should be applied due to attribute changes, so I'm missing the importance. Does this mean it just is double-checking that those classes exist on the element before snapshots? If so, is there another maybe Lit provided method to make sure the component is considered updated? (Maybe not, since you'd be looking at children)

2nd-gen/packages/swc/.storybook/preview.ts
```diff
@@ -122,27 +127,50 @@ const preview = {
...
+    // Chromatic captures testing-grid stories only; each grid opts in with
+    // `chromatic: { disableSnapshot: false }` (see testing-grid helpers).
+    chromatic: {
+      disableSnapshot: true,
+      forcedColors: 'none',
```
5t3ph, line 165: Can we make a follow-up ticket to enable forced-colors for the test grid as well? I assume that would have to be like a second pass over the grid, but we do want to capture states for that.

5t3ph (commented):
Will the VRT story page show up in the nav for the prod deploy, or just local and PR?

Image

2nd-gen/packages/swc/.storybook/preview.ts
aramos-adobe: Yes that's a great idea

aramos-adobe:
> prod deploy,

Just changed this to only shown in dev mode. @caseyisonit also brought this up.

PR #6331, opened 2026-05-20
feat(tooltip): s2 fidelity

5t3ph:
## Description

Implements the 2nd-gen `<swc-tooltip>` **Phase 5** — Styling: full S2 token stylesheet, RTL-aware placement margins and open-state animations, directional tip geometry (physical `right`/`left`, logical `start`/`end`), forced-colors block, scaffold Storybook stories

### Notable API change from the feature branch plan

_This is not style impacting or demonstrated yet in this work, but was discovered during this phase_

The `delayed: boolean` (default `false`, opt-in) property has been replaced with `delay: number` (default `1500`, warm-up inherent). Setting `delay="0"` opts out of warm-up delay. This aligns with Spectrum's intent that warm-up is the default behavior for tooltips, not an add-on. The 1st-gen behavioral difference is called out in the migration plan.

### Infrastructure and Docs/Skills Changes

_These will be moved to a new PR to merge into main ASAP_

- **`stylelint.config.js`** — All CSS system color keywords (`CanvasText`, `ButtonFace`, etc.) added to the `scale-unlimited/declaration-strict-value` allowlist so forced-colors overrides can assign system colors without triggering the design token lint rule.
- **`linters/stylelint-property-order.js`** — Missing transition property added to order definition.

## Motivation and context

Continues the Spectrum 2 component migration workstream. This branch delivers the full implementation through styling.

## Related issue(s)

- SWC-2023

## Manual review test cases

> Pixel-accurate anchoring requires `PlacementController` (additive phase) and the `HoverController` for events; these stories temporarily use Floating UI directly to verify visual appearance across placements and clicks for toggling.

- [ ] **Variant colors are correct**
  1. In the `Variants` story, validate that:
      - "Save" uses the neutral default color
      - "Upload" uses the blue informative color
      - "Delete" uses the red negative color

- [ ] **Handles text wrapping, matching tooltip content size**
  1. In the `Anatomy` story, validate:
      - Action" tooltip shrinks to fit the content size
      - "Another action" tooltip content wraps after reaching the max allowed inline size

- [ ] **Placement margins and tip direction are correct for all six placements**
  1. Inspect each placement (`top`, `bottom`, `left`, `right`, `start`, `end`)
  3. Expect the tip arrow to point toward the trigger, and the body margin to space correctly

- [ ] **RTL placement is correct for logical placements**
  1. In the `Placements` story, set "RTL" from the Storybook toolbar
  3. Inspect `start` and `end` placements
  4. Expect tip direction and animation direction to flip correctly for RTL

- [ ] **Forced-colors (high contrast mode) renders correctly**
  1. Enable high contrast mode in OS or DevTools forced-colors emulation
  3. Toggle open the tooltip
  4. Expect no loss of shape or tooltip visibility

### Device review

- [ ] Did it pass in Desktop?

- [ ] **Keyboard** (required)
  1. Press <kbd>Tab</kbd> to focus the trigger button
  2. Press <kbd>Enter</kbd> to toggle the tooltip (_temporary until HoverController available_)

  - Fully correct connection is pending the `HoverController`


2nd-gen/packages/swc/components/tooltip/stories/tooltip.stories.ts
```diff
@@ -0,0 +1,351 @@
...
+  tags: ['options'],
+  parameters: { flexLayout: 'row-wrap', 'section-order': 1 },
+};
+
+/**
+ * The `placement` attribute sets the preferred position of the tooltip relative to its trigger.
+ * Pixel-accurate anchoring requires `PlacementController` (additive phase); these stories
+ * temporarily use Floating UI directly to verify visual appearance across placements.
+ */
+export const Placements: Story = {
```
aramos-adobe, line 285: Love this display here ✨

aramos-adobe (approved):

2nd-gen/packages/swc/components/tooltip/tooltip.css
```diff
@@ -10,6 +10,287 @@
...
+   Space between the tip arrow and the trigger; based on tip-height token.
+   These margins create gap when PlacementController computes position.
+   NOTE: top / default margin lives in .swc-Tooltip above to avoid
+   re-declaring .swc-Tooltip after the nested lang rules.
+   ───────────────────────────────────────────────────────────────────────────── */
+
+/* Bottom: margin at block-start */
+:host([placement="bottom"]) .swc-Tooltip,
+:host([placement="bottom-start"]) .swc-Tooltip,
+:host([placement="bottom-end"]) .swc-Tooltip {
```
rise-erpelding: Let's talk about these compound placement variants (bottom-start, bottom-end, right-top, etc. etc.)! Will they stay or go in 2nd-gen?

It feels reasonable to keep the placements simple (top, bottom, left, right, start, end only) as you have demonstrated currently, and that's what matches the design and React. This would definitely be something we'd want to communicate to consumers at the appropriate phase in the migration.

I also see that [React](https://react-spectrum.adobe.com/Tooltip#api) has a cool `crossOffset` property that allows you to adjust tip placement, which could be something we think about eventually? Seems like that would be different than our current `offset`.

If we have no plans to implement compound placement variants later, can we cut out the selectors?

CONTRIBUTOR-DOCS/03_project-planning/03_components/tooltip/migration-plan.md
```diff
@@ -603,21 +600,23 @@ The impact is most acute in the additive phase, when `HoverController` will call
...
-- [ ] Add all six logical placement classes, consolidated with their non-logical equivalents: `start`, `start-top`, `start-bottom`, `end`, `end-top`, `end-bottom`
-- [ ] Verify CJK language modifiers (`:lang(ja)`, `:lang(ko)`, `:lang(zh)`)
-- [ ] Verify visibility in WHCM
-- [ ] Add `@cssprop` JSDoc tag for any exposed `--swc-*` property
-- [ ] Pass stylelint (property order, `no-descending-specificity`, token validation)
+- [x] Add `.swc-Tooltip` to the internal semantic container in `render()`; keep styling off `:host`
+- [x] Copy S2 source from `spectrum-css/components/tooltip/index.css` (`spectrum-two` branch, not `/dist`) into `tooltip.css` as baseline
+- [x] Map Spectrum CSS selectors to SWC equivalents following CSS selector guidance in CONTRIBUTOR_DOCS
+- [x] Remove `.spectrum-Tooltip-typeIcon` styles (no icon in S2)
+- [x] Add all six logical placement classes, consolidated with their non-logical equivalents: `start`, `start-top`, `start-bottom`, `end`, `end-top`, `end-bottom`
```
rise-erpelding: I asked about this in a different comment, but are we implementing the combined/compound placements like `start-top`, `start-bottom`, etc.?

2nd-gen/packages/swc/components/tooltip/tooltip.css
```diff
@@ -10,6 +10,287 @@
...
+   positioning.
+   ───────────────────────────────────────────────────────────────────────────── */
 
-/* Source: spectrum-css/components/tooltip/index.css (spectrum-two branch). */
+:host {
+  --_swc-tooltip-animation-distance: token("spacing-75");
+
+  position: absolute;
+  inset: auto;
+  max-inline-size: min(100%, token("tooltip-maximum-width"));
```
rise-erpelding, line 25: The reason we add these extra styles to the `:host` here is because we need them on the popover, is that right? Could be worth documenting the exception in the style guide docs for future work.

CONTRIBUTOR-DOCS/02_style-guide/01_css/02_custom-properties.md
```diff
@@ -105,6 +105,21 @@ When a private property captures a token value, reference the private property i
...
+  --_swc-tooltip-border-radius: token("corner-radius-400");
+  --_swc-tooltip-tip-corner-offset: var(--_swc-tooltip-border-radius);
+
+  border-radius: var(--_swc-tooltip-border-radius);
+}
+
+:host([placement="top-start"]) .swc-Tooltip-tip {
+  inset-inline-start: var(--_swc-tooltip-tip-corner-offset);
+}
+```
```
rise-erpelding: I can see this example becoming confusing for the agent, since we aren't really doing this in `tooltip.css`, is there another example in a shipped component we could use to illustrate this instead?

2nd-gen/packages/swc/components/tooltip/tooltip.css
```diff
@@ -10,6 +10,287 @@
  * governing permissions and limitations under the License.
  */
 
-/* Placeholder — CSS implementation is Phase 5. */
+/* ─────────────────────────────────────────────────────────────────────────────
+   HOST
+   Reset browser UA popover styles. All visual styles live on .swc-Tooltip.
+   PlacementController (additive phase) will apply inset/left/top for
+   positioning.
+   ───────────────────────────────────────────────────────────────────────────── */
```
rise-erpelding, line 18: I like these banner comments, could be a place to expand documentation on [comment conventions and section headers](https://github.com/adobe/spectrum-web-components/blob/main/CONTRIBUTOR-DOCS/02_style-guide/01_css/01_component-css.md#comment-conventions)

2nd-gen/packages/swc/components/tooltip/tooltip.css
```diff
@@ -10,6 +10,287 @@
...
+:host(:not(:popover-open)) {
+  display: none;
+}
+
+* {
+  box-sizing: border-box;
+}
+
+/* ─────────────────────────────────────────────────────────────────────────────
+   BASE TOOLTIP
```
rise-erpelding, line 49: A nit, but this could mention these are styles for top placement, I know you mention it above styles for placement margins but initially I didn't see that one.

2nd-gen/packages/swc/components/tooltip/tooltip.css
```diff
@@ -10,6 +10,287 @@
...
+  color: inherit !important;
+}
+
+/* ─────────────────────────────────────────────────────────────────────────────
+   FORCED COLORS (HIGH CONTRAST MODE)
+   ───────────────────────────────────────────────────────────────────────────── */
+
+@media (forced-colors: active) {
+  .swc-Tooltip {
+    border: 1px solid CanvasText;
```
rise-erpelding, line 282: Checking AssistivLabs the tip is sometimes slightly oddly positioned:

Image

I'm wondering if throwing the border in here (which is definitely needed, no arguments there) is throwing off the tip math and if we should try to adjust for that?

2nd-gen/packages/swc/components/tooltip/tooltip.css
5t3ph: Good catch, an oversight to leave these selectors behind.

In talking to Ruben re: the placement controller, I think `crossOffset` is something we'll evaluate in a future pass.

CONTRIBUTOR-DOCS/03_project-planning/03_components/tooltip/migration-plan.md
5t3ph: Removed, matching React

2nd-gen/packages/swc/components/tooltip/tooltip.css
5t3ph: Good idea, updated!

CONTRIBUTOR-DOCS/02_style-guide/01_css/02_custom-properties.md
5t3ph: The property this refers to ended up getting removed due to some logical placements being removed.

I think the guideline is low likelihood, so I've removed it for now as a more generic use case isn't coming to mind.

2nd-gen/packages/swc/components/tooltip/tooltip.css
5t3ph: Added to docs!

2nd-gen/packages/swc/components/tooltip/tooltip.css
5t3ph: It seems the extra +/- `0.5px` isn't needed for regular appearance, and removal resolved forced-colors

2nd-gen/packages/swc/components/tooltip/tooltip.css
```diff
@@ -263,7 +254,7 @@
    by the browser (right wins over left for RTL absolutely-positioned elements). */
 :host([placement="start"]) .swc-Tooltip-tip {
   inset-block-start: calc(50% - (0.5 * var(--_swc-tooltip-tip-square-size)));
-  inset-inline-start: calc(100% - (0.5 * var(--_swc-tooltip-tip-square-size)) - 0.5px);
+  inset-inline-start: calc(100% - (0.5 * var(--_swc-tooltip-tip-square-size)));
```
rise-erpelding, line 258: Ok, so a funny thing happened to me while I was looking at this placements story in WHCM. I noticed this was still happening:

Image

Specifically only for the `start` placement, it's just a little bit off, and only the first time the page is loaded/reloaded. This screen shot is from forced-colors in Chrome but I also saw this in WHCM via Assistiv Labs.

I found if I added half a pixel that it looked correct, but the solution that I ended up with grabbed the idea of an antialiasing-inset custom prop from the CSS work and was more like this (with the other styles omitted in this snippet). Maybe there's another happy medium solution, or maybe we don't solve for this--I can't pretend that I fully understand why it happens just for the start placement.

```css
.swc-Tooltip-tip {
  --_swc-tooltip-tip-edge-inset: calc(
    (0.5 * var(--_swc-tooltip-tip-square-size)) + var(--_swc-tooltip-tip-antialiasing-inset)
  );

  top: calc(100% - var(--_swc-tooltip-tip-edge-inset));
}

:host([placement="top"]) .swc-Tooltip-tip {
  top: calc(100% - var(--_swc-tooltip-tip-edge-inset));
}

:host([placement="bottom"]) .swc-Tooltip-tip {
  inset-block: auto calc(100% - var(--_swc-tooltip-tip-edge-inset));
}

:host([placement="end"]) .swc-Tooltip-tip {
  inset-inline: auto calc(100% - var(--_swc-tooltip-tip-edge-inset));
}

:host([placement="start"]) .swc-Tooltip-tip {
  inset-inline: calc(100% - var(--_swc-tooltip-tip-edge-inset)) auto;
}
```

rise-erpelding (approved):
This looks good, left one comment about the `start` placement still looking just a tiny bit off, if that's something you think needs addressing!

2nd-gen/packages/swc/components/tooltip/tooltip.css
5t3ph: Ahh, ok - when the agent copied in the styles it seemed like just a random half-pixel, but I bet it's a subpixel rendering issue. It looks like the lowest-lift "fix" is to use `z-index: -1` to tuck it behind the content visually rather than trying to calc around it. As long as it's visible as a pointer, even if it's reduced, that's the main thing! Otherwise we could try a different solution for craeting the tip shape that ensures a whole pixel size if we need to, we can keep an eye on it.

PR #6337, opened 2026-05-26
feat: add PlacementController to Gen2

rubencarvalho:
## Description

Adds `PlacementController` — a Lit reactive controller that positions a floating element relative to a trigger using [Floating UI](https://floating-ui.com/)'s `computePosition` + `autoUpdate`. It's the 2nd-gen replacement for the 1st-gen `PlacementController` baked into `<sp-overlay>`, but extracted from the overlay lifecycle so popover, picker, menu, tooltip, coachmark, etc. can each compose it without inheriting overlay machinery.

The controller is the first 2nd-gen primitive that downstream popover-shaped components will use for positioning.

### Public API

```typescript
import { PlacementController } from '@spectrum-web-components/core/controllers/placement-controller';

class MyPopover extends LitElement {
  private placement = new PlacementController(this);

  open(): void {
    this.placement.start(this.trigger, this.panel, {
      placement: 'bottom-start',
      offset: 8,
      shouldFlip: true,
      onPlacementChange: (p) => {
        this.actualPlacement = p;
      },
    });
  }

  close(): void {
    this.placement.stop();
  }
}
```

| Member | Description |
|---|---|
| `start(trigger, floating, options?)` | Begin positioning; tear down any prior session and subscribe to `autoUpdate`. |
| `stop()` | Tear down positioning, clear inline `max-height` / `max-width` written by `size` and the tip styles written by `arrow`, reset state. Called automatically from `hostDisconnected`. |
| `recompute()` | Force one `computePosition` pass outside `autoUpdate`. |
| `actualPlacement` | Readonly. Computed placement after `flip` (hyphenated), or `null` when stopped. Refreshed on every compute. |
| `isConstrained` | Readonly. `true` while `size` middleware is applying `max-height` because content would otherwise overflow. |

Options: `placement` (22 hyphenated values including logical `start` / `end`), `offset`, `crossOffset`, `containerPadding`, `shouldFlip`, `tipElement`, `tipPadding`, `onPlacementChange`.

`VirtualTrigger` is supported — any `{ getBoundingClientRect(); contextElement? }` shape works as the anchor.

### Behavioural parity with 1st-gen

Where the controller's behaviour is observable to consumers, it matches 1st-gen:

- **Middleware order** is the same: `offset → shift → flip → size → arrow`. `shift` runs before `flip` so the panel slides along the current side before any decision to flip; `size` runs after the final placement is known; `arrow` runs last (only when a `tipElement` is provided) so it positions the tip after every other middleware has settled.
- **`size` middleware always installed.** Writes `max-width` on every compute. Writes `max-height` only when content would otherwise overflow — the controller tracks an `initialHeight` baseline from the first un-constrained compute (1st-gen's pattern) to detect when content is currently being clamped, and exposes that as `isConstrained`.
- **`arrow` middleware** (opt-in via `tipElement`) positions a caller-provided tip so it points at the trigger's center, inset from the panel corners by `tipPadding`. After every compute it pins the tip to the relevant floating edge (`top` / `left`) and writes inline `translate` to slide it along that edge; the tip styles are cleared on `stop()`.
- **iOS WebKit `visualViewport` listeners + offset compensation** for URL-bar collapse / pinch-zoom / virtual-keyboard scenarios.
- **DPR rounding** (`roundByDPR`) so translated coordinates stay sharp on high-DPR displays.
- **`document.fonts.ready` await + WebKit rAF wait** before measuring.
- **Fallback-placement table** for virtual triggers.

### Where the controller intentionally diverges from 1st-gen

These are scope changes, not behaviour changes. They were called out and confirmed in the popover migration plan:

| Area | 1st-gen | 2nd-gen |
|---|---|---|
| Scope | Overlay lifecycle + ARIA + positioning | Geometry only |
| `autoUpdate` channels | Two channels (one closed the overlay on ancestor scroll) | One channel — just repositions. Caller decides whether ancestor scroll should close the surface. |
| DOM mutation on trigger | Wrote `placement` / `actual-placement` attributes on host elements | None. Computed placement surfaces via the `actualPlacement` property and `onPlacementChange` callback. |
| Host coupling | Required `host.elements: OpenableElement[]` | Plain `ReactiveControllerHost` |
| `sp-update-overlays` global event | Listened to it for reflow triggers | Removed. The caller invokes `recompute()` directly when needed. |
| `placeOverlay(target, options)` | Single async method | Split into `start(trigger, floating, options)` / `stop()` / `recompute()` with separated lifecycle. |
| Options shape | `OverlayOptionsV1` (abortPromise, delayed, notImmediatelyClosable, receivesFocus, root, type, …) | `PlacementOptions` — positioning fields only (including `tipElement` / `tipPadding` for arrow support). Trigger is a separate first argument. |
| Offset shape | `number \| [number, number]` (array for multi-axis) | Separate `offset` and `crossOffset` fields. |

### Public API additions

- `Placement` union (22 values including logical `start` / `end` and physical sub-alignment variants), exported with `ALL_PLACEMENTS` list.
- `VirtualTrigger` interface.
- `PlacementOptions` interface, including `tipElement` / `tipPadding` for opt-in `arrow` middleware.
- Pure conversion helpers `toFloatingPlacement` / `fromFloatingPlacement` — exported and unit-tested. The SWC `Placement` union doesn't map 1:1 to Floating UI's 12 values; these bridge the two:
  - Logical sides (`start`, `end`) → physical sides (`left`, `right`)
  - Logical sub-alignments (`start-top`, `end-bottom`) → Floating UI's `start` / `end` suffixes on the resolved physical side
  - Physical alignments (`bottom-left`, `left-top`) → Floating UI's `start` / `end` suffixes
  - Reverse: Floating UI's `left-start` / `left-end` / `right-start` / `right-end` (not in the SWC union) map back to `left-top` / `left-bottom` / `right-top` / `right-bottom`

### Storybook

Docs page at `Controllers / Placement controller` with the playground, behavioral demos (requested placement, offset / cross-offset, container padding, should-flip, size-always-clamps, on-placement-change, virtual trigger, arrow), and an inline API table. All 14 stories tagged appropriately for the new core-controller documentation template.

## Motivation and context

This is the first geometry primitive shared across the 2nd-gen popover-shaped components. Splitting it out of the 1st-gen overlay model lets popover, picker, menu, tooltip, coachmark, etc. each own their open/close lifecycle and ARIA wiring, while positioning math is centralised here. Aligns with the popover migration plan ([`CONTRIBUTOR-DOCS/03_project-planning/03_components/popover/migration-plan.md`](../tree/main/CONTRIBUTOR-DOCS/03_project-planning/03_components/popover/migration-plan.md)).

## Related issue(s)

- SWC-1996

## Author's checklist

- [ ] I have included a well-written changeset if my change needs to be published. _(internal-only controller for now; no changeset)_

## Reviewer's checklist

### Manual review test cases

- [ ] **Storybook docs page renders cleanly**
  1. Open Storybook → `Controllers / Placement controller`
  2. Expect a docs page with Overview, Usage, Behaviors (Requested placement, Offset, Container padding, Should flip, Size always clamps, On placement change, Virtual trigger, Arrow), API, Accessibility, and Appendix sections
  3. No "Component not found in manifest" warnings should surface for controller stories

- [ ] **Playground reacts to every option**
  1. Open the Playground story
  2. Toggle each option (placement, offset, cross-offset, container padding, should-flip) and confirm the floating panel responds immediately

- [ ] **Offset demo shows the difference**
  1. Open Behaviors → Offset
  2. Compare the default (8 px) and 48 px variants; the larger-offset panel sits visibly further from the trigger
  3. Same check for `crossOffset` in the lower row

- [ ] **Size always clamps the list**
  1. Open Behaviors → Size always clamps
  2. The 24-item list is visibly clamped; `host.floatingEl.style.maxHeight` reads as a numeric px value
  3. `host.isConstrained` is `true`

- [ ] **Arrow tip points at the trigger**
  1. Open Behaviors → Arrow
  2. The triangular tip on the floating panel points at the trigger's center
  3. Resize the surface so the panel shifts to stay inside the viewport; the tip continues to point at the trigger's center (inset from the panel corners by `tipPadding`)

- [ ] **Virtual trigger follows clicks**
  1. Open Behaviors → Virtual trigger
  2. Click anywhere inside the 320×320 surface
  3. The floating panel repositions near the click point with a 12 px offset; coordinates in the panel update

- [ ] **29 placement-controller unit tests pass**
  1. From `2nd-gen/packages/swc`, run `yarn test "core/controllers/placement-controller"`
  2. All 29 tests should pass — covers conversion functions, all options, flip behaviour both ways, size middleware lifecycle, arrow tip positioning, rapid `start()` replacement, and disconnect cleanup

### Device review

- [ ] Did it pass in Desktop?
- [ ] Did it pass in (emulated) Mobile? — iOS WebKit `visualViewport` compensation is included; verify popovers don't drift when toggling the simulated URL bar

The controller handles **geometry only** — it does not manage focus, ARIA, or keyboard dismissal. Consumer components remain responsible for accessible open/close behaviour. There is no UI to test against directly.

- [x] **Keyboard** — N/A. No focusable parts in the controller itself; consuming components wire their own keyboard handling.
- [x] **Screen reader** — N/A. The controller writes `top` / `left` / `translate` inline styles only (on the floating element and, when provided, the tip element); no roles, names, or live regions. Consumers attach `aria-controls`, `aria-expanded`, and labelling as appropriate.

Accessibility guidance for consumers is documented in the Storybook **Accessibility** story (under `Controllers / Placement controller`).


2nd-gen/packages/core/controllers/placement-controller/src/placement-controller.ts
```diff
@@ -0,0 +1,430 @@
...
+          // Track the natural (un-constrained) height on the first compute
+          // so subsequent computes can detect whether content is currently
+          // being clamped.
+          this.initialHeight = !this.isConstrained
+            ? actualHeight
+            : (this.initialHeight ?? actualHeight);
+          this.isConstrained =
+            actualHeight < this.initialHeight || maxHeight <= actualHeight;
+          Object.assign(floating.style, {
+            maxWidth: `${Math.floor(availableWidth)}px`,
```
5t3ph: An inline style is stronger than component's style, and this means that for something like Tooltip where the max-width is much smaller in comparison to the "available width" that it will not successfully wrap unless it truly reaches the width of the viewport or imposed container.

I think we should pass this off to components to implement, see for example [how this is handled for Tooltip](https://github.com/adobe/spectrum-web-components/blob/swc-2017/tooltip/2nd-gen/packages/swc/components/tooltip/tooltip.css#L25) where use of CSS `min()` is used to ensure it doesn't expand beyond the viewport/container but otherwise respects the intended max-width value.

2nd-gen/packages/core/controllers/placement-controller/src/placement-controller.ts
```diff
@@ -0,0 +1,430 @@
...
+            maxHeight: this.isConstrained ? `${maxHeight}px` : '',
```
5t3ph: This will need to ensure `box-sizing: border-box` to apply as intended (could be a requirement of consuming component, which is already expected to be the standard per our CSS guidelines).

Current - overflows:
Image

With `box-sizing: border-box`:
Image

2nd-gen/packages/core/controllers/placement-controller/src/placement-conversion.ts
```diff
@@ -0,0 +1,118 @@
...
+    'left-start': 'left-top',
+    'left-end': 'left-bottom',
+    'right-start': 'right-top',
+    'right-end': 'right-bottom',
+  };
+
+/**
+ * Normalize a hyphenated `Placement` for Floating UI's `computePosition`.
+ *
+ * - Logical **sides** (`start`, `end`) map to `left` / `right`.
```
5t3ph: This _must_ handle flipping start/end based on RTL like was [done temporarily for tooltip](https://github.com/adobe/spectrum-web-components/blob/swc-2017/tooltip/2nd-gen/packages/swc/components/tooltip/stories/tooltip.stories.ts#L83). Yes, CSS picks up to do things like reposition the arrow, but the pixel positioning has to come from here to get it to the correct side.

Using `getComputedStyle` _may_ be more accurate in case there is a more scoped direction mode change than the document.

5t3ph (changes requested):
Checked against Toolitp expectations, and overall looks pretty good beyond addressing the RTL issue as noted.

Also, I wish the flip behavior didn't necessarily flip _back_ to the initial position just because it re-gains the space. I know that comes from FloatingUI, and looking at their docs I don't think there's a way to adjust it. That's one thing that at first seemed odd from Anchor Positioning but now I think is the better call which is once flipped it won't flip back unless lack of space necessitates it again. The hyper-active flip can be jarring, especially with larger, scrolling elements. But, I think you'd have to create our own middleware to change the behavior, so I'm not suggesting a change at this time but flagging to keep an eye on it while we build out things like menu and picker.
https://github.com/user-attachments/assets/bf5d3b5b-a0b2-4680-836f-708248b672bf

2nd-gen/packages/core/controllers/placement-controller/src/placement-controller.ts
rubencarvalho: OK, great call! Can you check if you like my solution? sooo, I stopped the controller from writing max-width/max-height inline. If I understood right your tooltip example, I made it so components opt in by combining the intended max-width in CSS min().
```css
  .floating {
    max-inline-size: min(<intended>, var(--swc-placement-available-width));
    max-block-size:  min(<intended>, var(--swc-placement-available-height));
    overflow: auto; /* scroll when block space is constrained */
  }
```

I think this also solves the box-sizing comment, since we're not writing an inline max-height, "clamping" is applied by the consuming component with its own border-box. LMK if you had something else in mind!

2nd-gen/packages/core/controllers/placement-controller/src/placement-conversion.ts
rubencarvalho: Yes!! I am addressing this now. Apparently, FloatingUI already does flip all but start/end, so we're doing that ourselves. I added some tests to cover this too 😄

2nd-gen/packages/core/controllers/placement-controller/src/placement-controller.ts
```diff
@@ -84,8 +84,39 @@ type ActiveSession = {
...
+ *
+ * Only components positioned by this controller receive these properties.
+ * Such a component opts into viewport-bounded sizing by combining them with
+ * its own intended size via `min()`, with a fallback for when the controller
+ * is not active:
+ *
+ * ```css
+ * .floating {
+ *   max-inline-size: min(var(--intended-width), var(--swc-placement-available-width, 100vw));
+ *   max-block-size: min(var(--intended-height), var(--swc-placement-available-height, 100vh));
```
5t3ph: Slight adjustment for logical properties and also make sure that there's a little buffer around the height to not scroll the trigger out of view (decent default anyway).

```js
 *   max-inline-size: min(var(--intended-width), var(--swc-placement-available-width, 100vi));
 *   max-block-size: min(var(--intended-height), var(--swc-placement-available-height, 90vb));
```

2nd-gen/packages/core/controllers/placement-controller/src/placement-controller.ts
```diff
@@ -305,7 +339,17 @@ export class PlacementController implements ReactiveController {
...
 
     const requestedPlacement = options.placement ?? DEFAULT_PLACEMENT;
-    const floatingPlacement = toFloatingPlacement(requestedPlacement);
+    // Resolve logical sides (`start` / `end`) against the trigger's writing
+    // direction so RTL positions the panel on the correct physical side. Read
+    // from the trigger (it lives in the consumer's possibly scoped-RTL subtree);
+    // fall back to the floating element for a `VirtualTrigger`.
+    const directionSource = trigger instanceof HTMLElement ? trigger : floating;
+    const direction =
+      getComputedStyle(directionSource).direction === 'rtl' ? 'rtl' : 'ltr';
```
5t3ph, line 387: This works if the page is set to RTL and then I update the position to `start`, but if I have it in the `start` position and change to RTL _or_ I load the Storybook playground (`?path=/story/core-controllers-placement-controller--playground&args=placement:start&globals=textDirection:rtl`) with it preset to RTL and `start` it doesn't adjust correctly.

2nd-gen/packages/core/controllers/placement-controller/stories/demo-hosts.ts
```diff
@@ -755,6 +755,10 @@ export class DemoPlacementConstrainSize extends LitElement {
       }
 
       .floating {
+        /* Consume the controller's available-space props so the list clamps
+         * and scrolls instead of overflowing. */
+        max-inline-size: var(--swc-placement-available-width);
+        max-block-size: var(--swc-placement-available-height);
```
5t3ph, line 881: See note in the src comment

5t3ph (commented):
Possible bug, slight adjustment to suggested dimension styles

2nd-gen/packages/core/controllers/placement-controller/src/placement-controller.ts
5t3ph: Comment left in context, but `box-sizing: border-box` still matters due to added padding (as seen on your demo even with this update).

2nd-gen/packages/core/controllers/placement-controller/src/placement-controller.ts
```diff
@@ -0,0 +1,488 @@
...
+          placement.startsWith('bottom') || placement.startsWith('top')
+            ? '0px'
+            : '',
+        translate: `${roundByDPR(arrowX ?? 0)}px ${roundByDPR(arrowY ?? 0)}px`,
+      });
+    }
+
+    const nextPlacement = fromFloatingPlacement(placement);
+    this.actualPlacement = nextPlacement;
+    options.onPlacementChange?.(nextPlacement);
```
Rajdeepc: The callback fires on every autoUpdate tick even when placement hasn't changed. The PR description documents this behavior, but it creates unnecessary re-renders in consuming components. You can add a equality check before invoking the callback which would be more efficient:
```ts
const nextPlacement = fromFloatingPlacement(placement);
if (nextPlacement !== this.actualPlacement) {
    this.actualPlacement = nextPlacement;
    options.onPlacementChange?.(nextPlacement);
}
```

Rajdeepc (changes requested):
Blocking for few specific things apart from @5t3ph 'scomment and it would be great if you can cover the below cases in your tests too.
1. Can you add a test to test `recompute()` in isolation? The VirtualTrigger test exercises it indirectly but there's no dedicated test verifying recompute() actually triggers a new compute and updates position
2. There are no test verifying two PlacementController instances on the same host don't interfere with each other.

1. I added a note on placementChange which was firing on every tick. Is that intentional? If so, what's the use case that requires it over a change-only callback?
2. I see you have added @floating-ui/dom as a dependency to core. Does this require a changeset? Leaving this upto you. No strong opinions since no one is consuming anything.

2nd-gen/packages/core/controllers/placement-controller/src/placement-controller.ts
rubencarvalho: Great catch!! I added a MutationObserver for this.

2nd-gen/packages/core/controllers/placement-controller/stories/demo-hosts.ts
rubencarvalho: Fixed! It does look better now! 😄

2nd-gen/packages/core/controllers/placement-controller/src/placement-controller.ts
rubencarvalho: 1. I added a recompute() test!
2. Added two controllers in one host! (demo-placement-multi-controller) fixture and the TwoControllersDoNotInterfere test.
3. I addressed the onPlacementChange every tick, I was tryung to keep parity with Gen1 but there it worked a bit differently because we reflected the actual-placement attribute instead of firing the callback. Claude fixed my implementation with a sentinel so it fires at the first call too.
4. I added a minor for @spectrum-web-components/core.

Rajdeepc (approved):
Thanks for working through the feedback.

5t3ph (approved):
Let's goooooo :shipit:

PR #6344, opened 2026-05-27
feat(tooltip): tests and code conformance

5t3ph:
## Description

Adds Phase 6 (migration testing) and the post-Phase-6 code conformance pass for `swc-tooltip`.

**Testing (SWC-2024)**

Adds two test files:

- `tooltip.test.ts` — Storybook play-function tests using Vitest and `@storybook/test`. Covers defaults, property/attribute mutation, open/close state, lifecycle events (`swc-open`, `swc-close`, `swc-after-open`, `swc-after-close`), ARIA `ariaDescribedByElements` wiring for native triggers, `swc-button` shadow triggers, `triggerElement` overrides, standalone (no trigger), and manual mode. Also covers Escape key dismissal, all variants, all placements, and the dev-mode warning for an unresolved `for` attribute.
- `tooltip.a11y.spec.ts` — Playwright ARIA snapshot tests. Validates that a closed tooltip is hidden from the accessibility tree, that an open tooltip exposes `role="tooltip"`, that it is removed from the tree when closed, that Escape closes it, and that variant and placement trigger buttons are reachable.

**Conformance pass (SWC-2025)**

- `Tooltip.ts` — Added `@fires` JSDoc tags for all four lifecycle events.
- `tooltip.css` — Added explanatory comment for `!important` on `::slotted()` resets.
- `tooltip.stories.ts` — Section separator renamed from "METADATA SETUP" to "METADATA"
- `tooltip.test.ts` — Restructured from ad-hoc section headers to the 5 prescribed standard sections; step labels updated to start with verbs; `EscapeClosesTest` moved to the Variants / States section; section renamed from "Debug warnings" to "Dev mode warnings".

## Motivation and context

Phase 6 of the 2nd-gen tooltip migration. Tests validate the full public API surface defined in Phase 3 and the accessibility wiring implemented in Phase 4. The conformance pass ensures all files align with project style guides before Phase 7 (documentation) begins.

## Related issue(s)

- SWC-2024
- SWC-2025

## Manual review test cases

- [ ] Play-function tests pass in the Storybook test runner
  1. Run `yarn test:storybook` (or equivalent) scoped to the tooltip component
  2. Confirm all stories under `Tooltip/Tests` show green in the test results panel
  3. Confirm no regressions in any other component's test stories

- [ ] Playwright a11y tests pass
  1. Run `yarn test:a11y` scoped to tooltip
  2. Confirm all six tests in `tooltip.a11y.spec.ts` pass
  3. Confirm ARIA snapshots match expected output

- [ ] Conformance-only changes produce no behavioral regressions
  1. Open `Tooltip` → `Overview` in Storybook
  2. Click the trigger button; confirm the tooltip opens and closes as expected
  3. Open `Tooltip` → `Variants` and `Tooltip` → `Placements`; confirm all variants and placements render correctly

### DIFF
--- a/2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
+++ b/2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
@@ -251,7 +251,10 @@ export abstract class TooltipBase extends SpectrumElement {
       this.open = isOpen;
     }
     // When no CSS transition is active, dispatch after-* immediately since transitionend will not fire.
-    if (getComputedStyle(this).transitionDuration === '0s') {
+    // transitionDuration is comma-separated when multiple properties transition ("0s, 0s, …"),
+    // so check that every value in the list is zero rather than comparing the full string.
+    const durations = getComputedStyle(this).transitionDuration.split(',');
+    if (durations.every((d) => d.trim() === '0s')) {
       this.afterEventPending = false;
       this.dispatchAfterEvent(isOpen);
     }
--- a/2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
+++ b/2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
@@ -147,9 +147,9 @@ export abstract class TooltipBase extends SpectrumElement {
 
   /**
    * When set, wires `ariaLabelledByElements` instead of `ariaDescribedByElements` on the trigger's
-   * inner interactive element. For icon-only triggers where the tooltip text is the sole accessible name.
-   *
-   * Additive/deferred: active in the additive phase.
+   * inner interactive element. Use for icon-only triggers where the tooltip text is the sole accessible
+   * name and adding an accessible label directly to the trigger host is not possible.
+   * Prefer `accessible-label` on the trigger when feasible — it works without this attribute.
    *
    * @default false
    */
@@ -195,11 +195,28 @@ export abstract class TooltipBase extends SpectrumElement {
     const target = (trigger.shadowRoot?.querySelector('button') ??
       trigger) as Element & {
       ariaDescribedByElements: Element[] | null;
+      ariaLabelledByElements: Element[] | null;
     };
-    const current = target.ariaDescribedByElements ?? [];
-    target.ariaDescribedByElements = this.open
-      ? [...current.filter((el) => el !== this), this]
-      : current.filter((el) => el !== this);
+
+    if (this.labeling) {
+      // Remove any stale describedby reference (e.g. if labeling changed while open).
+      const described = target.ariaDescribedByElements ?? [];
+      target.ariaDescribedByElements = described.filter((el) => el !== this);
+
+      const labelled = target.ariaLabelledByElements ?? [];
+      target.ariaLabelledByElements = this.open
+        ? [...labelled.filter((el) => el !== this), this]
+        : labelled.filter((el) => el !== this);
+    } else {
+      // Remove any stale labelledby reference (e.g. if labeling changed while open).
+      const labelled = target.ariaLabelledByElements ?? [];
+      target.ariaLabelledByElements = labelled.filter((el) => el !== this);
+
+      const described = target.ariaDescribedByElements ?? [];
+      target.ariaDescribedByElements = this.open
+        ? [...described.filter((el) => el !== this), this]
+        : described.filter((el) => el !== this);
+    }
   }
 
   private dispatchAfterEvent(isOpen: boolean): void {
@@ -265,6 +282,8 @@ export abstract class TooltipBase extends SpectrumElement {
           this.hidePopover();
         }
       }
+    }
+    if (changedProperties.has('open') || changedProperties.has('labeling')) {
       this.syncAriaRelationship();
     }
   }
--- a/2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
+++ b/2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
@@ -248,6 +248,13 @@ export abstract class TooltipBase extends SpectrumElement {
     this.dispatchAfterEvent(this.open);
   };
 
+  // Allows Escape behavior to be testable, does not interfere with native popover dismissal
+  private readonly handleKeyDown = (event: KeyboardEvent): void => {
+    if (event.key === 'Escape' && this.open) {
+      this.open = false;
+    }
+  };
+
   protected override updated(changedProperties: PropertyValues): void {
     super.updated(changedProperties);
     if (changedProperties.has('open')) {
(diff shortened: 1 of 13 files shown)


2nd-gen/packages/swc/components/tooltip/test/tooltip.test.ts
```diff
@@ -0,0 +1,556 @@
...
+} as Meta;
+
+// Type alias for elements with ariaDescribedByElements (available in Chrome 135+, Firefox 136+, Safari 16.4+).
+type AriaRelatable = Element & { ariaDescribedByElements: Element[] | null };
+
+// Awaits a DOM event dispatched on the given element, resolving with the event object.
+const waitForEvent = <T extends Event>(
+  el: EventTarget,
+  eventName: string
+): Promise<T> =>
```
cdransf, line 53: Should we add a timeout here in case the event doesn't fire? ✨ 

```suggestion
  eventName: string,
  timeoutMs: number = 3000
): Promise<T> =>
```

cdransf (approved):
Looks good! Just one nit ✨

2nd-gen/packages/swc/components/tooltip/test/tooltip.test.ts
5t3ph: TIL (thanks Claude):

> Every promise in the file is either:
> 
> - A `waitForEvent` backed by a synchronous browser guarantee (`beforetoggle` → `swc-open/swc-close`), or
> - A `waitForEvent` for `swc-after-open/swc-after-close` that fires either synchronously (no-transition path) or via > `transitionend` (bounded by the CSS), or
> - A `tooltip.updateComplete` / `swcButton.updateComplete` which is Lit's own update microtask — resolves promptly by design.
>
> None of them have a code path where the event could silently never arrive. A broken implementation causes a hang, and the global Storybook timeout turns that into a test failure with a clear "test timed out" message.

2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
rubencarvalho, line 195: I've been thinking about this (also because of the popover), we had these "focusable-selectors" in Gen1 (`1st-gen/tools/shared/src/focusable-selectors.ts`), I'm wondering if we'll ever want to extend this.

rubencarvalho: what's exactly the plan for this property? This is going to surface on the API table but it does not seem like it has effect.

rubencarvalho: I think this will not work for all cases... Check here: https://codepen.io/rubencarvalho/pen/yyVpGez
when multiple transition-property values are declared, we get a comma-separated duration list (`"0s, 0s, …"`), which would never `=== '0s'`

2nd-gen/packages/swc/components/tooltip/test/tooltip.test.ts
```diff
@@ -0,0 +1,559 @@
...
+  title: 'Tooltip/Tests',
+  parameters: {
+    ...meta.parameters,
+    docs: { disable: true, page: null },
+  },
+  tags: ['!autodocs', 'dev'],
+} as Meta;
+
+// Type alias for elements with ariaDescribedByElements (available in Chrome 135+, Firefox 136+, Safari 16.4+).
+type AriaRelatable = Element & { ariaDescribedByElements: Element[] | null };
```
rubencarvalho: nit: could we tighten these casts? maybe it even makes sense to augment the globals in `in 2nd-gen/packages/core/global.d.ts` if we need.

Something along the lines of:
```
  /**
   * ARIA element-reflection properties (the IDREF-free associations). Not yet in
   * the TypeScript DOM lib. Available in Chrome 135+, Firefox 136+, Safari 16.4+.
   */
  interface ARIAMixin {
    ariaDescribedByElements: Element[] | null;
    ariaLabelledByElements: Element[] | null;
  }
```
I think [ARIAMixin](https://w3c.github.io/aria/#ARIAMixin) is the interface that Element/HTMLElement/HTMLButtonElement all implement, so this lands the property on every element type without more casting 😄

2nd-gen/packages/swc/components/tooltip/test/tooltip.test.ts
```diff
@@ -0,0 +1,559 @@
...
+    <swc-button id="tt-swc-trigger">Save</swc-button>
+    <swc-tooltip for="tt-swc-trigger" placement="top">
+      Changes will be saved
+    </swc-tooltip>
+  `,
+  play: async ({ canvasElement, step }) => {
+    const swcButton = canvasElement.querySelector(
+      '#tt-swc-trigger'
+    ) as HTMLElement;
+    await (swcButton as HTMLElement & { updateComplete: Promise<boolean> })
```
rubencarvalho: We can cast this as an actuall `Button` clas which will bring `updateComplete` out of the box 😄

2nd-gen/packages/swc/stylesheets/global/global-button.css
rubencarvalho, line 1: Is this supposed to be included? I've been getting this file regenerating and getting stale every time I run Storybook locally, prob worth looking into this.

rubencarvalho (changes requested):
Left some out of scope comments, but I stumbled upon them when checking the tests 😛

2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
5t3ph: We may, but I was waiting to see how we landed with for example inputs in terms of setting up to attach to the "right" thing.

2nd-gen/packages/swc/components/tooltip/test/tooltip.test.ts
5t3ph: Suggestion moved to the globals 👍

2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
5t3ph: Good catch this was getting exposed. There wasn't a great reason to keep this as "additive" as the migration plan had noted, so I have updated to add this support.

2nd-gen/packages/swc/components/tooltip/test/tooltip.test.ts
5t3ph: updated

2nd-gen/packages/swc/stylesheets/global/global-button.css
5t3ph: it seems that if AI runs tests that this gets generated but without the linting step so it has "changes" that shoudl just be ignored. i think the checked in version has been corrected in main.

2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
5t3ph: The use case here is either `transition-duration` property isn't set at all for the component or it has been overridden for purposes of `prefers-reduced-motion` in which case we would wholesale adjust it to `transition-duration: 0s` meaning there would only be one final computed value.

But - updated on the off-chance we/consumers allow some properties to transition.

rubencarvalho (approved):

PR #6327, opened 2026-05-19
feat(action-button): adds migration plan for the S2 action button component

cdransf:
## Description

Adds the Phase 1 migration plan for the `swc-action-button` 1st-gen → 2nd-gen migration at `CONTRIBUTOR-DOCS/03_project-planning/03_components/action-button/migration-plan.md`.

The plan covers:

- Full 1st-gen API surface (`@spectrum-web-components/action-button@1.12.1`) with all properties, events, slots, CSS custom properties, and shadow DOM structure
- React Spectrum S2 API comparison, used to validate naming, behavior, and scope decisions
- 10 breaking changes (B1–B10) with consumer migration paths
- 6 additive features (A1–A6), including the new `pending` state
- Complete 2nd-gen API decisions table with Confirmed / Inferred / Deferred labels for every 1st-gen property, including `autofocus`, `tabIndex`, `name`, and `active`
- Architecture decision: `swc-action-button` extends `ButtonBase` directly; no separate `ActionButton.base.ts` in core is needed
- Migration sequencing and prerequisites, including dependency ordering with `swc-action-group`
- Full migration checklist covering Setup, API, Styling, Accessibility, Testing, Documentation, and Review phases
- Q#-format blockers table (Design, Architecture and behavior, Scope and prerequisites)
- "Breaking changes to verify" QA table mapping each B# to a manual verification step
- Deferred follow-up ticket table and Resolved decisions subsection

Key decisions:

- `toggles`, `selected`, `emphasized`, and consumer-controlled `role` are removed; toggle UX moves to `swc-toggle-button` / `swc-toggle-button-group`
- `hold-affordance` / `longpress` are deferred with documented consumer options
- `pending` is a new additive feature matching the `swc-button` contract in `ButtonBase`
- `accessible-label` (attribute) replaces `label`
- `size` includes `xs`, requiring `ACTION_BUTTON_VALID_SIZES` in a new `ActionButton.types.ts`
- `href` and the full link API are removed

## Motivation and context

[...]

PR #6346, opened 2026-05-27
chore(action-button): migrate api and typescript code

cdransf:
## Description

Phase 3 of the `swc-action-button` 1st-gen → 2nd-gen washing machine migration. This PR migrates the public API and TypeScript to the 2nd-gen `ActionButton` class, adds runtime deprecation warnings to 1st-gen, and cleans up the structural artifact left by Phase 2 (removing the intermediate `ActionButtonBase` class from core so the package exports types only, as the migration plan specifies).

**2nd-gen `ActionButton.ts`**

- `value` promoted to a getter/setter pair. When the stored value is empty, `textContent.trim()` is used as the effective value, matching the 1st-gen identity contract for `swc-action-group` membership.
- `aria-haspopup` and `aria-expanded` forwarding: private `@property` fields observe those attributes on the host and forward them into the inner `<button>`, enabling popup disclosure patterns without a `MutationObserver`.
- `@cssprop` entries updated to use logical property names (`--swc-action-button-min-block-size`, `--swc-action-button-border-radius`); two incorrect entries (`--swc-action-button-edge-to-text`, `--swc-action-button-edge-to-visual`) removed.
- `size` override (added in structural fix) moves the full `xs`–`xl` range onto the SWC class via `VALID_SIZES` static, a getter/setter, and a `_size` backing field. An `update()` override counteracts `SizedMixin`'s auto-reflect of `size="m"` for cases where no size was explicitly set.

**1st-gen `ActionButton.ts`**

- `@deprecated` JSDoc added to `emphasized`, `hold-affordance`, `selected`, and `toggles`.
- Runtime `window.__swc.warn({ level: 'deprecation' })` calls added inside existing code paths in `updated()` for each deprecated property.

**Structural fix (squashed from Phase 2 branch)**

- Deleted `ActionButton.base.ts` from `core/components/action-button/`; the package now exports types only, matching the migration plan architecture table.
- `ActionButton.ts` (SWC) extends `ButtonBase` directly.

## Motivation and context

This is Phase 3 of the action button migration as tracked in the 2nd-gen component migration workstream. The migration plan was merged in #6327 and the structural foundation was established in the Phase 2 PR.

The `value` getter/setter fallback and the `aria-haspopup`/`aria-expanded` forwarding are required for `swc-action-group` integration and popup disclosure patterns respectively; both are specified in the migration plan.

SWC-2043

## Related issue(s)

- Depends on Phase 2 structural PR (branch `cdransf/s2-action-button-migration-structure`)

N/A — TypeScript and API changes only; no visual output in this phase.

## Author's checklist

## Reviewer's checklist

### Manual review test cases

- [ ] _`value` fallback behavior_
  1. Render a `<swc-action-button>` with no `value` attribute set and text content `"Edit"`
  2. Read `element.value` in the console
  3. Expect `"Edit"` (falls back to `textContent.trim()`)

- [ ] _`value` explicit set_
  1. Render a `<swc-action-button value="edit-action">Edit</swc-action-button>`
  2. Read `element.value` in the console
  3. Expect `"edit-action"` (explicit value takes precedence)

- [ ] _`aria-haspopup` / `aria-expanded` forwarding_
  1. Render `<swc-action-button aria-haspopup="menu" aria-expanded="false">Open menu</swc-action-button>`
  2. Inspect the inner `<button>` in DevTools
  3. Expect `aria-haspopup="menu"` and `aria-expanded="false"` present on the inner `<button>`

- [ ] _Size attribute absent by default_
  1. Render `<swc-action-button>Edit</swc-action-button>` with no `size` attribute
  2. Inspect the host element
  3. Expect no `size` attribute on the host (the SizedMixin workaround suppresses auto-reflect of `size="m"`)

- [ ] _1st-gen deprecation warnings_
  1. Enable debug mode (`window.__swc = { DEBUG: true }`)
  2. Set `emphasized`, `selected`, `toggles`, or `hold-affordance` on a `<sp-action-button>`
  3. Expect a deprecation-level `window.__swc.warn` console message for each property

### Device review

- [ ] Did it pass in Desktop?

  1. Go to the `swc-action-button` Storybook stories (Playground story)
  2. Press <kbd>Tab</kbd> to move focus to the button; confirm focus ring is visible
  3. Press <kbd>Enter</kbd> or <kbd>Space</kbd>; confirm the button activates (click handler fires)
  4. Set `disabled` on the button; confirm it is no longer reachable via <kbd>Tab</kbd>
  5. Set `aria-haspopup="menu"` and `aria-expanded="false"` on the host; confirm focus and activation still work normally

  1. Open the Playground story with VoiceOver (macOS) or NVDA (Windows)
  2. Tab to the button; confirm it is announced as "button" with its visible label
  3. With `disabled` set, confirm the button is announced as dimmed/unavailable
  4. With `aria-haspopup="menu"` and `aria-expanded="false"` set on the host, confirm the inner button is announced with "has popup, menu" and "collapsed"
  5. With `pending` set, confirm `aria-disabled="true"` and the pending accessible name are announced


2nd-gen/packages/swc/components/action-button/ActionButton.ts
```diff
@@ -61,9 +61,18 @@ export class ActionButton extends ButtonBase {
   }
 
   public override set size(value: ActionButtonSize) {
+    if (!value) {
```
5t3ph: Missed removal of `value` handling since that is at least deferred, and probably deprecated (noted in previous phase).

2nd-gen/packages/swc/components/action-button/ActionButton.ts
```diff
@@ -85,17 +94,28 @@ export class ActionButton extends ButtonBase {
   // ───────────────────
 
   /**
-   * Applies the quiet (low-emphasis) visual treatment.
+   * Applies the quiet (low-emphasis) visual treatment. The button renders
```
5t3ph: This type of comment expansion kind of blows up the API documentation when the intent of this is better served in the Storybook docs alongside the example.

2nd-gen/packages/swc/components/action-button/ActionButton.ts
```diff
@@ -104,6 +124,14 @@ export class ActionButton extends ButtonBase {
     return [styles];
   }
 
+  protected override update(changes: PropertyValues): void {
```
5t3ph: Why is this needed? In today's sync, we made a purposeful decision to reflect the size.

CONTRIBUTOR-DOCS/03_project-planning/03_components/action-button/migration-plan.md
```diff
@@ -626,6 +626,8 @@ What `swc-action-button` adds on top of `ButtonBase`:
...
 | TBD | Avatar slot and accessible name composition (A7) | Avatar accessible name and its relationship to the button's composite accessible name requires a11y review (@nikkimk) before shipping. | [A7](#additive--ships-when-ready-zero-breakage-for-consumers-already-on-2nd-gen) |
 | TBD (under SWC-2039) | Cross-root ARIA mapping | Shared with `swc-button` dependency on `ElementInternals` / tooling path. | [Deferred semantics note](#deferred-semantics-note-2nd-gen) |
+| Phase 4 | `aria-haspopup` / `aria-expanded` host-attribute retention | Phase 3 forwards these to the inner `<button>` via `@property` observation but does not remove the originals from the host element. ATs in browse mode may read them from the host. Proper fix (guard flag + `attributeChangedCallback` override, or `MutationObserver`) should be addressed in Phase 4 alongside a11y review. | [2nd-gen API decisions — passthrough host attributes](#properties--attributes) |
```
5t3ph: [nit] My understanding is that AT would skip any ARIA that isn't compatible with the role. So since the host won't have the button role, these ARIA attributes also won't be doing any harm. What might be interesting is if having or not having them would cause issues with us testing relationships such as with Menu.

CONTRIBUTOR-DOCS/03_project-planning/03_components/action-button/migration-plan.md
```diff
@@ -626,6 +626,8 @@ What `swc-action-button` adds on top of `ButtonBase`:
...
+| Phase 5 | Pending width-lock CSS property name | `ButtonBase.update()` sets `--_swc-button-pending-inline-size` on the inner `<button>` when the pending delay fires. Phase 5 action-button CSS must reference this same property name for the width-lock to activate, not `--_swc-action-button-pending-inline-size`. If a component-scoped name is preferred, `ActionButton` must override the pending timer logic to write under the action-button name instead. | [Pending state](#pending-state-new-in-2nd-gen) |
```
5t3ph, line 630: This is a good callout. In the styling phase, you might look into creating `button-pending.css` and moving all shareable styles for the pending state there in addition to the stylesheet that is already intended to be shared of `button-base.css`.

5t3ph (changes requested):

2nd-gen/packages/swc/components/action-button/ActionButton.ts
cdransf: Ah! I missed the sync, but removed this to align with that decision. ✨

2nd-gen/packages/swc/components/action-button/ActionButton.ts
cdransf: [Removed this as part of the size removal](https://github.com/adobe/spectrum-web-components/pull/6346#discussion_r3320627726). ✨

2nd-gen/packages/swc/components/action-button/ActionButton.ts
cdransf: Good call! Went ahead and trimmed this back ✨

CONTRIBUTOR-DOCS/03_project-planning/03_components/action-button/migration-plan.md
cdransf: Ah yeah, it will ignore those — [stripping those attributes](https://github.com/adobe/spectrum-web-components/pull/6346/changes#diff-9c0b1ce9e65a887b459b95117e4ae334f75b2c80ed47aa85dc31b5a773ca88f7R115) keeps the host clean for tooling and avoids any ambiguity if the host is given a role in the future. ✨

CONTRIBUTOR-DOCS/03_project-planning/03_components/action-button/migration-plan.md
cdransf: Noted this for phase 5 ✨

5t3ph (approved):

2nd-gen/packages/swc/components/action-button/ActionButton.ts
```diff
@@ -26,6 +26,8 @@ import styles from './action-button.css';
 
 /**
  * A compact action button for toolbars, action groups, and icon-first chrome.
+ * Supports sizes `xs`–`xl`; `xs` is an action-button-specific addition not
+ * available on `swc-button`.
```
rise-erpelding, line 30: Should applying `xs` as a size to the action button work? When I try to change the attribute to `xs` either in the inspector or in Storybook it bounces back to `m` because it doesn't seem to recognize it.

aramos-adobe (approved):

2nd-gen/packages/swc/components/action-button/ActionButton.ts
```diff
@@ -51,51 +53,38 @@ export class ActionButton extends ButtonBase {
...
+
+  /**
+   * @internal
+   * Forwarded to the inner `<button>` for menu-trigger patterns. After the
+   * attribute is read, it is stripped from the host so assistive technologies
+   * in browse mode do not encounter duplicate ARIA state on both the host
+   * element and the inner button.
+   */
+  @property({ type: String, attribute: 'aria-expanded' })
+  protected ariaExpanded?: string;
```
rise-erpelding: I see some issues locally, but maybe it's something on my end? It's complaining about `ariaExpanded` and `ariaHasPopup` not using `override` since they already exist on the `HTMLElement` (and the types are slightly different too, `string | undefined` vs `string | null`), plus it's marked as a public property even though it isn't publicly available (would `@state` with a backing field work better)?

rise-erpelding (commented):
Just a couple of questions but otherwise looks good!

2nd-gen/packages/swc/components/action-button/ActionButton.ts
cdransf: Huh! That was an interesting one. The `size` setter in `SizedMixin` validates against the `validSizes` array captured when the `ButtonBase` calls the mixin, not against `this.constructor.VALID_SIZES`. `ACTION_BUTTON_VALID_SIZES` contained `xs` but was always rejected by the setter. ✨

2nd-gen/packages/swc/components/action-button/ActionButton.ts
cdransf: Ah! Sooo `ariaHasPopup` and `ariaExpanded` both exist on `HTMLElement` via `ARIAMixin` as `string | null`. `@property` conflicted with the native type, required `override` and surfaced those fields in the CEM.

I swapped them out with `@state` private fields and a manual static override that adds the attribute names. `attributeChangedCallback` should capture them in `@state`, strip the attributes from the host and return without calling `super`. _Finally_, the `@state` decorator triggers the re-render that attaches the value to the interior button. ✨ 🫠

2nd-gen/packages/swc/components/action-button/ActionButton.ts
```diff
@@ -135,8 +137,8 @@ export class ActionButton extends ButtonBase {
         aria-label=${ifDefined(
           this.pending ? this.getPendingAccessibleName() : this.accessibleLabel
         )}
-        aria-haspopup=${ifDefined(this.ariaHasPopup)}
-        aria-expanded=${ifDefined(this.ariaExpanded)}
+        aria-haspopup=${ifDefined(this._ariaHasPopup)}
+        aria-expanded=${ifDefined(this._ariaExpanded)}
```
rise-erpelding, line 141: I do get a lit-plugin warning in the editor the the types are incompatible with whatever is already set for those attributes, for instance for `aria-haspopup`:

`Type 'string' is not assignable to '"false" | "true" | "menu" | "listbox" | "tree" | "grid" | "dialog"'lit-plugin(no-incompatible-type-binding)(2304)`

Pretty small potatoes though, I don't think it triggers linting or throws a runtime error.

rise-erpelding (approved):
Looks good, one small thing that could optionally be addressed:

2nd-gen/packages/swc/components/action-button/ActionButton.ts
cdransf: Ah, yeah! We're just passing through what the consumer sets and the lit plugin enforces it, so it's a constraint on the call site I think rather than us forwarding it. ✨

PR #6350, opened 2026-05-28
docs(css,ai): update custom property rules and migration skill coverage

5t3ph:
## Description

Three areas of work combined on this branch:

**CSS stylelint and linting updates:**
- Adds `transition-behavior` to the stylelint property order so it sorts correctly within the `transition-*` group
- Adds the full set of CSS system color keywords (`ButtonFace`, `CanvasText`, `GrayText`, etc.) as allowed values in stylelint so forced-colors overrides don't trip the design token enforcement rule

**CSS style guide additions (`CONTRIBUTOR-DOCS/02_style-guide/01_css/`):**
- `01_component-css.md`: Documents section banner format, expands the `:host` exception with code examples covering UA resets, entry/exit transitions, and positioning surfaces; adds `inherit` guidance for sub-elements that track a variant value; clarifies native pseudo-class preference over attribute selectors for states
- `02_custom-properties.md`: Adds explicit rules for private property placement (`:host` does not encapsulate `--_swc-*`), exposure criteria (expose only when the component itself overrides, with nested component relationship exception), size variant pattern (single property overridden per size selector), state-specific custom property pattern for native browser states on internal elements, and full property name convention
- `03_component-css-pr-checklist.md`: Adds a **Custom properties** section covering all five new rules as checkable items
- `05_anti-patterns.md`: Updates anti-pattern # 6 with the nested component relationship exception; adds new anti-pattern # 10 covering size-specific custom properties

**AI migration skill and rule updates (`.ai/`):**
- `migration-phase-awareness.md`: Adds "Migration plan checklist" as an explicit phase-completion step and surfaces it in the Migration Checkpoint format
- `migration-styling` SKILL.md: Adds Step 5 to write `@cssprop` JSDoc tags on the SWC class for every exposed property after writing CSS
- `migration-documentation` SKILL.md: Adds a `@cssprop` verification step to Phase 7 cross-check
- `consumer-migration-guide` SKILL.md and prompt: Scopes guide content to what actually shipped in the source, not migration plan categories (Additive/Deferred reflect intent, not final state)
- `code-conformance.md`: Makes the CSS PR checklist the first and explicit verification step rather than a passive reference

## Motivation and context

Custom property exposure decisions were underdocumented, leading to inconsistent implementations across components — in particular: private properties declared on `:host` (which exposes them), size-specific custom properties that bloat the API surface, and missing `@cssprop` tags that break Storybook API docs. The PR feedback on Button that prompted these additions also identified that state-specific custom properties (`:hover`, `:focus-visible`, `:active`) on internal elements are correct and necessary because static-color compound overrides require them.


cdransf (approved):
This looks awesome! ✨

caseyisonit (approved):

PR #6354, opened 2026-06-01
feat(popover): scaffold + API contract (phases 2+3)

rubencarvalho:
## Description

**Foundation for the Popover migration (Epic SWC-1993): Phase 2 (Setup) + Phase 3 (API contract).** No runtime behavior — the lifecycle, positioning, ARIA wiring, and styling land in #6357.

### Phase 2 — Setup

- **core:** `PopoverBase` (property surface), `Popover.types`, and the `resolve-trigger` / `dismissible-stack` util stubs wired into `utils/index` + `package.json`.
- **swc:** `Popover` render stub (branches `<div popover="auto">` vs `<dialog>` on `modal`), `popover.css` stub, `swc-popover` registration, and empty-but-valid stories + tests.

### Phase 3 — API contract (aligned with React Spectrum S2 `Popover`)

Public surface: `placement`, `size` (`s`/`m`/`l`), `offset` (default 8), `cross-offset`, `should-flip`, `hide-arrow` (arrow shown by default), plus the web-component-specific `open`, `modal`, `for`, `trigger-element`, `manual`, and the readonly `actualPlacement`.

- **types:** `Placement` (re-exported from the PlacementController), `POPOVER_VALID_PLACEMENTS`, `POPOVER_VALID_SIZES`, `PopoverSize`, `PopoverCloseSource`, `PopoverCloseEventDetail`.
- **base:** full property surface; `container-padding` / `tip-padding` marked `@internal` (positioning details, excluded from the public API table); dev-only `placement` validation via the `VALID_PLACEMENTS` static.
- **swc:** `@fires` for `swc-open` / `swc-after-open` / `swc-close` / `swc-after-close` (surfaced in the CEM; `@internal` props excluded).

## Related issue(s)

- Epic SWC-1993

## Notes

- Targets the `ruben/popover-migration` integration branch. Behavior is in #6357; the changeset is added once at the integration→`main` boundary (#6356), not per phase.

- [ ] **Keyboard** — no behavior in these phases (scaffolding + contract); keyboard handling is implemented in #6357 / verified in Phase 6.
- [ ] **Screen reader** — event and type contract defined here; ARIA wiring is implemented in #6357 and verified in Phase 6.


2nd-gen/packages/core/components/popover/Popover.base.ts
```diff
@@ -0,0 +1,192 @@
...
+   */
+  @property({ type: String, reflect: true })
+  public size?: PopoverSize;
+
+  /**
+   * The computed placement after the `flip` middleware reorients the popover.
+   * Readonly; not reflected as an attribute.
+   */
+  @property({ attribute: false })
+  public actualPlacement: Placement | null = null;
```
Rajdeepc: The migration plan says this is **readonly** updated by the controller via `onPlacementChange` callback, but `@property` + `public` means consumers can freely write to it:

```ts
popover.actualPlacement = 'top'; // silently desyncs from controller
```
Would it be ok to add a `@readonly` JSDoc so that the CEM communicates the intent.

2nd-gen/packages/core/components/popover/Popover.base.ts
```diff
@@ -0,0 +1,192 @@
...
+
+  /**
+   * Hide the popover's arrow (tip). The arrow is shown by default.
+   *
+   * @default false
+   */
+  @property({ type: Boolean, reflect: true, attribute: 'hide-arrow' })
+  public hideArrow = false;
```
Rajdeepc, line 101: The migration plan API table still documents `tip` as a `boolean` (opt-in, default `false`). Your code implements `hideArrow` with attribute `hide-arrow`. You need to document this as breaking change and update the API table.

2nd-gen/packages/core/components/popover/Popover.base.ts
```diff
@@ -0,0 +1,192 @@
...
+
+  /**
+   * Main-axis offset in pixels from the trigger.
+   *
+   * @default 8
+   */
+  @property({ type: Number })
+  public offset = 8;
```
Rajdeepc, line 109: The migration plan API table says:
> `offset | number | 0 | ... we default to 0 to make the controller-host contract neutral`

But the code defaults to `8`. If 8 is correct kindly document it.

2nd-gen/packages/core/components/popover/Popover.base.ts
```diff
@@ -0,0 +1,192 @@
...
+  public containerPadding = 8;
+
+  /**
+   * Allow the popover to flip to the opposite side when constrained. When
+   * `false`, the popover stays in the requested placement.
+   *
+   * @default true
+   */
+  @property({ type: Boolean, reflect: true, attribute: 'should-flip' })
+  public shouldFlip = true;
```
Rajdeepc, line 138: The migration plan marks `should-flip` as **Internal (`@internal`)**  but the JSDoc here has no `@internal` tag, unlike `containerPadding` and `tipPadding` which do.
Without the tag, the CEM will surface `shouldFlip` in the public API table.

2nd-gen/packages/core/components/popover/Popover.base.ts
```diff
@@ -0,0 +1,192 @@
...
+   */
+  @property({ type: String, reflect: true })
+  public placement: Placement = 'bottom';
+
+  /**
+   * Optional fixed size. When set, the popover uses a fixed inline size
+   * (`s` → 336px, `m` → 416px, `l` → 576px); when unset, it fits its contents.
+   */
+  @property({ type: String, reflect: true })
+  public size?: PopoverSize;
```
Rajdeepc, line 84: Kindly add `size` as is not documented in the migration plan API table

2nd-gen/packages/core/utils/resolve-trigger.ts
```diff
@@ -0,0 +1,48 @@
...
+export interface ResolvedTrigger {
+  trigger: HTMLElement | null;
+  interactiveElement: HTMLElement | null;
+}
+
+/**
+ * Resolve the trigger for a popover-like host and discover the AT-facing inner
+ * button across shadow boundaries.
+ *
+ * @todo Phase 3 (API): implement `for=` resolution via
```
Rajdeepc: > `@todo Phase 3 (API): implement for= resolution...`

This PR is Phase 3. The todo should reference Phase 4/5 (lifecycle implementation).

Also: the function returns `{ trigger: null, interactiveElement: null }` unconditionally. Any code importing and calling it will get a null trigger with no warning. Since the types and interface are published to `core/utils/index.ts`, this could be consumed by a parallel work.

2nd-gen/packages/swc/components/popover/Popover.ts
```diff
@@ -0,0 +1,71 @@
...
+      ${this.hideArrow
+        ? nothing
+        : html`
+            <span class="swc-Popover-tip"></span>
+          `}
+    `;
+
+    // The render shape branches on `modal`: a `<div popover="auto">` in the
+    // default (non-modal) mode, a `<dialog>`(.showModal()) in modal mode.
+    return this.modal
```
Rajdeepc, line 65: If `modal` changes at runtime while the popover is open, Lit will swap the entire internal element (`<div>` ↔ `<dialog>`) and will destroy the existing top-layer session (`showPopover()` / `showModal()` state
and all attached event listeners (`beforetoggle`, `cancel`, etc.)
Phase 4/5 should guard against this: if `open` is `true` when `modal` changes, you should close the current element first, let Lit re-render, then re-open with the new lifecycle. A dev-mode warning would also help.

Rajdeepc (commented):
**Some of my findings:**
1. **`tip` → `hide-arrow` rename** is undocumented in the migration plan API table (B7 breaking change needs updating)
2. **`offset` default** is 8 in code but 0 in the plan. This needs reconciliation
3. **`size` and `manual`** are new properties not yet in the plan API table
4. **Stub utilities** (`dismissible-stack`, `resolve-trigger`) are exported but non-functional
5. **`actualPlacement`** is writable by consumers despite being documented as readonly

2nd-gen/packages/core/components/popover/Popover.base.ts
5t3ph: For consideration: Tooltip was tentatively planning a `no-tip` property (additive, not yet implemented). Either way, it would be good to have consistent naming.

2nd-gen/packages/core/components/popover/Popover.base.ts
5t3ph: I recommend keeping this public - the flipping behavior is rather aggressive and may be unwanted in cases such as long menus or pickers (unless we will be blocking page scroll while open).

2nd-gen/packages/core/components/popover/Popover.base.ts
rubencarvalho: I have an active thread with Nadeen on this, because design calls it "tip" and React an arrow. Trying to figure out which way to go. We called "tip" on Gen1 too.

2nd-gen/packages/core/components/popover/Popover.base.ts
rubencarvalho: Yes, 1) I tried aligning with React here and 2) Steph's comment regarding aggressive flipping. I also have an active thread with Nadeen regarding this "aggressive flipping" concern.

Rajdeepc (approved):
Approving since most of the concerns are addressed. 
Few nits
1. You can add unit tests for dismissible-stack.ts (register/unregister/isTop)
2. It would be great if you can document closest('[dir]') shadow boundary limitation in PlacementController JSDoc.

2nd-gen/packages/core/components/popover/Popover.base.ts
5t3ph: For consideration - tooltip ended up setting this via regular attribute methods and not exposing here. Maybe `@readonly` accomplishes the same thing, but it felt like it should be more protected. I may have had the wrong train of thought, though!

2nd-gen/packages/core/components/popover/Popover.base.ts
5t3ph: You will likely have to take a cue from Tooltip as far as styling is concerned where I adjusted it to use the offset via a custom property in the CSS with a token fallback, that way the distance stays in sync and the CSS didn't add on top.

2nd-gen/packages/swc/components/popover/stories/popover.stories.ts
```diff
@@ -0,0 +1,73 @@
...
+
+import '@adobe/spectrum-wc/components/popover/swc-popover.js';
+
+// ────────────────
+//    METADATA
+// ────────────────
+
+const { events, args, argTypes, template } = getStorybookHelpers('swc-popover');
+
+/**
```
5t3ph, line 24: [nit at this stage] reminder to switch to MDX for final docs

2nd-gen/packages/swc/components/popover/popover.css
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+/* @todo Phase 5 (styling): port S2 source from spectrum-css and add placement modifiers. */
+
+* {
```
5t3ph, line 15: [FYI] You'll probably want to make the agent aware of tooltip, and copy most if not all of the `:host` styles onto `.swc-Popover` to correctly place and transition the element, plus take its cues to resolve some of the other positioning styles.

CONTRIBUTOR-DOCS/03_project-planning/03_components/popover/migration-plan.md
```diff
@@ -69,12 +69,12 @@
```
5t3ph, line 69: [nit] consider adding a decision log to capture the update to change from a `div[popover]` for _both_ modal and non-modal to instead the swap to `dialog`

5t3ph (approved):
Looks good! Just some suggestions to consider for the later phases.

PR #6322, opened 2026-05-19
chore(close-button): migration guide

TarunAdobe:
## Description

- Adds the close-button **Phase 1 planning artifacts** under `CONTRIBUTOR-DOCS`:
  - `CONTRIBUTOR-DOCS/03_project-planning/03_components/close-button/migration-plan.md`
  - `CONTRIBUTOR-DOCS/03_project-planning/03_components/close-button/rendering-and-styling-migration-analysis.md`
- Removes the previously added **consumer-facing** guide:
  - `2nd-gen/packages/swc/components/close-button/migration-guide.mdx`
- Updates `CONTRIBUTOR-DOCS/03_project-planning/03_components/README.md` to index the new close-button plan + roadmap docs.
- Captures confirmed design inputs in the plan/docs:
  - S2 Web Desktop Figma node for close button
  - S2 close-button anatomy doc
  - `spectrum-css` `spectrum-two` closebutton source

## Motivation and context

This work corrects scope and artifact type for the close-button epic.  
The prior MDX was consumer migration documentation, but this stage requires **internal migration planning docs** (washing-machine Phase 1 style) before implementation.  

This PR establishes the review surface for:
- 1st-gen API inventory
- must-ship vs additive migration scope
- core vs SWC architecture direction
- sequencing/checklist for setup → API → styling → a11y → testing → docs → review

## Related issue(s)

- Epic SWC-2087 — Close button migration
- SWC-2088 — Accessibility recommendations (already documented)
- SWC-2089 — Analyze component and create migration plan (this PR)

## Screenshots (if appropriate)

- N/A (documentation-only change)

---

## Author's checklist

- [x] I have read the **[CONTRIBUTING](https://github.com/adobe/spectrum-web-components/blob/main/CONTRIBUTING.md)** and **[PULL_REQUESTS](https://github.com/adobe/spectrum-web-components/blob/main/PULL_REQUESTS.md)** documents.
- [x] I have reviewed at the Accessibility Practices for this feature, see: [Aria Practices](https://www.w3.org/TR/wai-aria-practices/)
- [ ] I have added automated tests to cover my changes.
- [ ] I have included a well-written changeset if my change needs to be published.
- [x] I have included updated documentation if my change required it.

---

## Reviewer's checklist

- [x] Includes a Github Issue with appropriate flag or Jira ticket number without a link
- [ ] Includes thoughtfully written changeset if changes suggested include `patch`, `minor`, or `major` features
- [ ] Automated tests cover all use cases and follow best practices for writing
[...]

PR #6355, opened 2026-06-01
chore(close-button): add structure for s2 migration

TarunAdobe:
## Description
  Phase 2 (Setup) of the `swc-close-button` 1st-gen → 2nd-gen washing machine migration. This PR scaffolds the component so it builds, registers, and appears in Storybook before API polish and S2 styling land in follow-up PRs.
    
  **Core (`2nd-gen/packages/core/components/close-button/`)**
  - `CloseButton.types.ts` — `CLOSE_BUTTON_VALID_SIZES`, static-color / variant types
  - `CloseButton.base.ts` — extends ButtonBase with close-button VALID_SIZES, static-color, and legacy variant 
    (maps to static-color when set; full deprecation warnings deferred to Phase 3). Size uses shared SizedMixin 
    behavior from ButtonBase (default m, reflected on the host).
  - Package export wired in `core/package.json`
  **SWC (`2nd-gen/packages/swc/components/close-button/`)**
  - `CloseButton.ts` — inner `<button type="button">`, size-mapped cross icon, visually hidden default slot, `aria-label` from resolved accessible name
  - `close-button.css` — stub layout and visually hidden label styles (not S2 visual parity)
  - `swc-close-button.ts` — element registration
  - Storybook Playground + Overview stories
  - Placeholder unit and a11y spec tests (expanded in Phase 6)
  
  **Docs**
  - Migration status table: Close Button **Setup** marked complete
  
  ## Motivation and context
  Close button migration is tracked under the 2nd-gen component workstream. Setup must land before Phase 3 (API / TypeScript) and Phase 5 (S2 CSS). This PR intentionally ships a **minimal, buildable skeleton** — not visual parity with Spectrum 2 closebutton CSS.
  
  ## Related issue(s)
  - Epic SWC-2087 — Close button migration
  - Planning (separate PR): SWC-2089 — migration plan 
  ([#6322](https://github.com/adobe/spectrum-web-components/pull/6322))
  
  N/A — stub styling only; no meaningful visual regression vs Spectrum 2 
  yet.
  ---
  ## Author's checklist
  - [x] I have read the **[CONTRIBUTING](https://github.com/adobe/spectru
  m-web-components/blob/main/CONTRIBUTING.md)** and 
  **[PULL_REQUESTS](https://github.com/adobe/spectrum-web-components/blob
  /main/PULL_REQUESTS.md)** documents.
  - [x] I have reviewed the Accessibility Practices for this feature, 
  see: [Aria Practices](https://www.w3.org/TR/wai-aria-practices/)
  - [x] I have added automated tests to cover my changes (placeholder 
  Storybook + a11y snapshot; full coverage in Phase 6).
  - [ ] I have included a well-written changeset if my change needs to be
   published.
  - [x] I have included updated documentation if my change required it 
  (status table only).
  ---
  ## Reviewer's checklist
  - [ ] Includes a Github Issue with appropriate flag or Jira ticket 
  number without a link
  - [ ] Includes thoughtfully written changeset if changes suggested 
  include `patch`, `minor`, or `major` features
  - [ ] Automated tests cover all use cases and follow best practices for
   writing
  ### Manual review test cases
  - [ ] _Verify close-button scaffold builds and registers_
    1. Run `yarn build` from repo root (or `2nd-gen` workspace build for 
  `core` + `swc`).
    2. Confirm build completes without errors.
  - [ ] _Verify element registration and Storybook_
    1. Open 2nd-gen Storybook → **Close Button** → **Playground** (or 
  branch preview Storybook if available).
    2. Confirm `<swc-close-button accessible-label="Close">` renders with
   a cross icon.
    3. In DevTools, confirm shadow DOM contains a real inner `<button 
  type="button">` with `aria-label="Close"`.
  - [ ] _Verify size and static-color attributes (stub visuals)_
    1. In Playground, set `size` to `s`, `m`, `l`, and `xl`; confirm the 
  host updates and the icon mapping changes
    2. Set `static-color` to `white` or `black`; confirm classes 
  `swc-CloseButton--staticWhite` / `swc-CloseButton--staticBlack` appear 
  on the inner button (styling is stub-only).
  - [ ] _Verify default size is reflected on the host_  
  1. Render <swc-close-button accessible-label="Close"> with no author-set size.  
  2. Inspect the host; expect size="m" on the host (property default m via SizedMixin).
  ### Device review
  - [ ] Did it pass in Desktop?
    1. Open 2nd-gen Storybook → **Close Button** → **Playground**.
    2. Press <kbd>Tab</kbd> until focus reaches the close button; confirm
   a visible focus indicator on the inner button (delegated focus).
    3. Press <kbd>Enter</kbd> or <kbd>Space</kbd>; confirm the button 
  activates (click handler / default button behavior).
    4. Set `disabled` on the control; confirm the button is not reachable
   via <kbd>Tab</kbd> and does not activate.
    1. Open Playground with VoiceOver (macOS) or NVDA (Windows).
    2. Tab to the button; confirm it is announced as a **button** named 
  **Close** (from `accessible-label` or slotted text).
    3. With `disabled` set, confirm the control is announced as 
  unavailable/disabled.

### DIFF
--- a/2nd-gen/packages/core/components/button/Button.base.ts
+++ b/2nd-gen/packages/core/components/button/Button.base.ts
@@ -20,25 +20,26 @@ import {
   SizedMixin,
 } from '@spectrum-web-components/core/mixins/index.js';
 
-import { BUTTON_VALID_SIZES } from './Button.types.js';
+import {
+  BUTTON_STATIC_COLORS,
+  BUTTON_VALID_SIZES,
+  type ButtonStaticColor,
+} from './Button.types.js';
 
 /**
- * Abstract base class for all button-like components. Owns shared semantic
- * concerns: interaction state, sizing, slot-derived icon/label state,
- * accessible-name resolution, and host-to-internal-button attribute forwarding.
+ * Abstract base for button-like components that share sizing, slots, disabled
+ * state, accessible-name resolution, pending state, and static-color treatment.
  *
- * Visual API specific to `sp-button` (`variant`, `fill-style`, `static-color`)
- * is intentionally absent so that ActionButton, ClearButton, CloseButton,
+ * Visual API specific to `swc-button` (`variant`, `fill-style`) is
+ * intentionally absent so that ActionButton, ClearButton, CloseButton,
  * PickerButton, and InfieldButton can extend this base without inheriting
- * the `swc-button` visual surface.
+ * the full `swc-button` visual surface.
  *
  * @slot - Visible button label.
  * @slot icon - Optional leading icon.
  *
  * @attribute {ElementSize} size - The size of the button.
- *
- * @todo We currently have 3 levels of mixins on this class, but the mixin
- * composition guide recommends a maximum of 2. Explore reducing after milestone 2.
+ * @attribute {'white' | 'black'} static-color - Static color treatment for display over colored or image backgrounds.
  */
 export abstract class ButtonBase extends SizedMixin(
   ObserveSlotText(ObserveSlotPresence(SpectrumElement, '[slot="icon"]'), ''),
@@ -83,6 +84,12 @@ export abstract class ButtonBase extends SizedMixin(
   public pendingLabel?: string;
 
   /**
+   * Static color treatment for display over colored or image backgrounds.
+   */
+  @property({ type: String, reflect: true, attribute: 'static-color' })
+  public staticColor?: ButtonStaticColor;
+
+  /**
    * Tracks whether the pending visual (disabled colors + spinner) is currently
    * active. Set to `true` after a 1-second delay once `pending` becomes true,
    * so the button does not immediately flash to its unavailable appearance.
@@ -112,7 +119,7 @@ export abstract class ButtonBase extends SizedMixin(
    *
    * @internal
    */
-  protected getResolvedAccessibleName(): string | null {
+  public getResolvedAccessibleName(): string | null {
     return this.accessibleLabel ?? (this.textContent?.trim() || null);
   }
 
@@ -151,11 +158,11 @@ export abstract class ButtonBase extends SizedMixin(
     super.connectedCallback();
     // Capture phase so slotted light-DOM clicks are suppressed before host
     // listeners (e.g. Storybook actions) run.
-    this.addEventListener('click', this.handleClick, true);
+    this.addEventListener('click', this.handleActivationClick, true);
   }
 
   public override disconnectedCallback(): void {
-    this.removeEventListener('click', this.handleClick, true);
+    this.removeEventListener('click', this.handleActivationClick, true);
     if (this._pendingTimer !== null) {
       clearTimeout(this._pendingTimer);
       this._pendingTimer = null;
@@ -165,17 +172,27 @@ export abstract class ButtonBase extends SizedMixin(
   }
 
   /**
-   * Suppresses click activation while the button is `disabled` or `pending`.
+   * Suppresses click activation while the button is disabled or pending.
    *
    * Slotted icon content lives in the light DOM, so pointer clicks on icons
    * bypass the disabled inner `<button>` and bubble on the host. The host
    * listener (capture) and inner `@click` binding both call this handler.
    */
-  protected readonly handleClick = (event: Event): void => {
-    if (this.disabled || this.pending) {
+  protected handleActivationClick(event: Event): void {
+    if (this.disabled) {
       event.preventDefault();
       event.stopImmediatePropagation();
+    } else if (this.pending) {
+      event.stopImmediatePropagation();
     }
+  }
+
+  /**
+   * Suppresses click activation while the button is disabled or pending.
+   * Subclasses' templates wire this onto the rendered `<button>` via `@click`.
+   */
+  protected readonly handleClick = (event: Event): void => {
+    this.handleActivationClick(event);
   };
 
   protected override update(changedProperties: PropertyValues): void {
@@ -213,6 +230,17 @@ export abstract class ButtonBase extends SizedMixin(
           `<${this.localName}> should not set both "pending" and "disabled" simultaneously. Use "pending" to keep the button focusable while unavailable, or "disabled" to fully remove it from the tab order.`,
           'https://opensource.adobe.com/spectrum-web-components/components/button/#pending',
           { issues: ['pending + disabled'] }
+        );
+      }
+      if (
+        typeof this.staticColor !== 'undefined' &&
+        !BUTTON_STATIC_COLORS.includes(this.staticColor)
+      ) {
+        window.__swc.warn(
+          this,
+          `<${this.localName}> element expects the "static-color" attribute to be one of the following:`,
+          'https://opensource.adobe.com/spectrum-web-components/components/button/#static-color',
+          { issues: [...BUTTON_STATIC_COLORS] }
         );
       }
       if (this.hasIcon && !this.hasLabel && !this.accessibleLabel) {
--- a/2nd-gen/packages/swc/components/button/Button.ts
+++ b/2nd-gen/packages/swc/components/button/Button.ts
@@ -26,7 +26,6 @@ import {
   BUTTON_VARIANTS,
   ButtonBase,
   type ButtonFillStyle,
-  type ButtonStaticColor,
   type ButtonVariant,
 } from '@spectrum-web-components/core/components/button';
 
@@ -97,13 +96,6 @@ export class Button extends ButtonBase {
    */
   @property({ type: String, reflect: true, attribute: 'fill-style' })
   public fillStyle: ButtonFillStyle = 'fill';
-
-  /**
-   * Static color treatment for display over colored or image backgrounds.
-   * Only supported with `primary` and `secondary` variants.
-   */
-  @property({ type: String, reflect: true, attribute: 'static-color' })
-  public staticColor?: ButtonStaticColor;
 
   /**
    * Whether overflowing text is truncated with an ellipsis rather than
--- a/2nd-gen/packages/swc/components/close-button/CloseButton.ts
+++ b/2nd-gen/packages/swc/components/close-button/CloseButton.ts
@@ -1,0 +1,85 @@
+/**
+ * Copyright 2026 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import { CSSResultArray, html, TemplateResult } from 'lit';
+import { classMap } from 'lit/directives/class-map.js';
+import { ifDefined } from 'lit/directives/if-defined.js';
+
+import {
+  ButtonBase,
+  type ButtonSize,
+} from '@spectrum-web-components/core/components/button';
+
+import {
+  Cross200Icon,
+  Cross300Icon,
+  Cross400Icon,
+  Cross500Icon,
+} from '../icon/elements/index.js';
+
+import styles from './close-button.css';
+
+const crossIconBySize: Record<ButtonSize, () => TemplateResult> = {
+  s: Cross200Icon,
+  m: Cross300Icon,
+  l: Cross400Icon,
+  xl: Cross500Icon,
+};
+
+/**
+ * A compact dismiss control for dialogs, banners, toasts, and similar chrome.
+ *
+ * @element swc-close-button
+ * @since 0.0.1
+ *
+ * @slot - Accessible text label rendered visually hidden next to the cross icon.
+ *
+ * @example
+ * <swc-close-button accessible-label="Close"></swc-close-button>
+ *
+ * @example
+ * <swc-close-button>Close</swc-close-button>
+ */
+export class CloseButton extends ButtonBase {
+  // ──────────────────────────────
+  //     RENDERING & STYLING
+  // ──────────────────────────────
+
+  public static override get styles(): CSSResultArray {
+    return [styles];
+  }
+
+  protected override render(): TemplateResult {
+    const resolvedName = this.getResolvedAccessibleName();
+
+    return html`
+      <button
+        class=${classMap({
+          'swc-CloseButton': true,
+          'swc-CloseButton--staticWhite': this.staticColor === 'white',
+          'swc-CloseButton--staticBlack': this.staticColor === 'black',
+        })}
+        type="button"
+        @click=${this.handleActivationClick}
+        ?disabled=${this.disabled}
+        aria-label=${ifDefined(resolvedName ?? undefined)}
+      >
+        <span class="swc-CloseButton-icon" aria-hidden="true">
+          ${crossIconBySize[this.size]()}
+        </span>
+        <span class="swc-CloseButton-label">
+          <slot></slot>
+        </span>
+      </button>
+    `;
+  }
+}
--- a/2nd-gen/packages/core/AGENTS.md
+++ b/2nd-gen/packages/core/AGENTS.md
@@ -8,7 +8,7 @@
 core/
 ├── element/       # SpectrumElement base class, defineElement, version tracking
 ├── mixins/        # SizedMixin, ObserveSlotPresence, ObserveSlotText
-├── controllers/   # LanguageResolutionController
+├── controllers/   # LanguageResolutionController, PlacementController, …
 ├── utils/         # capitalize, getLabelFromSlot
 └── components/    # One folder per component
     └── badge/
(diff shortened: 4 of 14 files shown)


TarunAdobe:
I implemented this close-button scaffold migration with the Cursor agent, then self-reviewed the diff once before marking it ready for review.

A few things worth calling out for reviewers:

  • Cross400 / Cross500 icons — added so l and xl close-button sizes map to the right cross scale (not just
    reusing the medium icon).
  • Storybook docs — narrowed the autodocs API table so inherited ButtonBase members like pending and
    pending-label don’t show up as close-button API (they aren’t part of this component’s public surface). A
    post-inheritance CEM plugin strips @internal overrides after inheritance applies.
  • 2nd-gen API — swc-close-button does not ship the 1st-gen variant="white|black" alias; use static-color only. Per 2nd-gen convention we drop legacy APIs here rather than deprecating them; variant deprecation stays a 1st-gen (sp-close-button) follow-up.
  • Migration plan — updated to match: close-button extends ButtonBase for shared semantics (naming, disabled, sizing), not for swc-button-style pending behavior;

  Please flag anything that looks off in the scaffold or wiring

2nd-gen/packages/swc/components/close-button/CloseButton.ts
```diff
@@ -0,0 +1,98 @@
...
+ * <swc-close-button accessible-label="Close"></swc-close-button>
+ *
+ * @example
+ * <swc-close-button>Close</swc-close-button>
+ */
+export class CloseButton extends CloseButtonBase {
+  /**
+   * @internal
+   */
+  @property({ type: Boolean, reflect: true })
```
rubencarvalho: no need to redeclare `@property` probably

2nd-gen/packages/core/components/close-button/CloseButton.base.ts
```diff
@@ -0,0 +1,65 @@
...
+   * @internal
+   * Not part of the close-button public API.
+   */
+  public override pending = false;
+
+  /**
+   * @internal
+   * Not part of the close-button public API.
+   */
+  public override pendingLabel?: string;
```
Rajdeepc: Does this work? CEM will scan the Buttonbase properties first. Kindly verify this. Keeping it blocking for now!

Rajdeepc (changes requested):

2nd-gen/packages/core/components/close-button/CloseButton.base.ts
TarunAdobe: works with the plugin that i added. and here's the ticket for broader button base refractor work: https://jira.corp.adobe.com/browse/SWC-2250

Rajdeepc: You can abstract the CloseButton required properties in an abstract class in `button.base.ts` which ButtonBase and CloseButtonBase should extend. This would keep the API surface cleaner even if future components extends ButtonBase.
```ts

// button.base.ts

export abstract class InteractiveButtonbase extends SizedMixin(
  ObserveSlotText(ObserveSlotPresence(SpectrumElement, '[slot="icon"]'), ''),
  { validSizes: BUTTON_VALID_SIZES }
) {
 static override shadowRootOptions: ShadowRootInit = {
    ...SpectrumElement.shadowRootOptions,
    delegatesFocus: true,
  };
  /**
   * Whether the button is disabled. Removes focusability and prevents
   * interaction.
   */
  public disabled: boolean = false;
  /**
   * Accessible label forwarded to the internal `<button>` element as
   * `aria-label`. Required for icon-only buttons, which have no visible text.
   */
  @property({ type: String, attribute: 'accessible-label' })
  public accessibleLabel?: string;
  protected get hasIcon(): boolean {
    return this.slotContentIsPresent;
  }
 protected get hasLabel(): boolean {
    return this.slotHasContent;
  }
  /**
   * Resolves the accessible name for the button from `accessibleLabel` or
   * visible text content. Returns `null` when no accessible name is
   * determinable.
   *
   * @internal
   */
  protected getResolvedAccessibleName(): string | null {
    return this.accessibleLabel ?? (this.textContent?.trim() || null);
  }
  /**
   * Returns the set of attributes that should be forwarded to the internal
   * semantic `<button>` element, if not otherwise directly managed.
   *
   * @internal
   */
  protected getForwardedButtonAttributes(): Record<
    string,
    string | boolean | undefined
  > {
    return {
      disabled: this.disabled,
    };
  }
  protected override update(changedProperties: PropertyValues): void {
    if (window.__swc?.DEBUG) {
      if (this.hasIcon && !this.hasLabel && !this.accessibleLabel) {
        window.__swc.warn(
          this,
          `<${this.localName}> with an icon and no label must have an "accessible-label" attribute to be accessible.`,
          'https://opensource.adobe.com/spectrum-web-components/components/button/#icon-only',
          { issues: ['accessible-label'] }
        );
      }
    }
  }
  
 export abstract class ButtonBase extends InteractiveButtonbase {
 
   // pending attributes of button stays here
 
  }
  


// CloseButton.base.ts
  
 export abstract class CloseButtonBase extends InteractiveButtonbase { }
  
  ```

Comment (TarunAdobe) [reply 2nd-gen/packages/core/components/close-button/CloseButton.base.ts]: Done

**Comment by TarunAdobe** [comment]
CloseButtonBase initially extended ButtonBase, which owns full button semantics including pending /
  pending-label. That leaked into:

  • CEM / Storybook controls (pending showed up on swc-close-button)
  • Docs MDX inherited from the button family

  Close-button is not a pending control (unlike swc-button), so that surface was wrong.

  First fix: @internal overrides on CloseButtonBase + a custom CEM internal-override-plugin to strip inherited
  members after analyze. That worked but was brittle (plugin had to reason about inheritance order;
  badge/status-light variant interactions were a footgun).

  Final fix (latest commit): structural split instead of doc/CEM patching:

  • SharedButtonBase — sizing, slots, disabled, accessible-label, accessible-name helpers (no pending)
  • PendingMixin — existing pending behavior moved here unchanged (pendingActive, 1s delay, click suppression,
    aria-disabled while pending, etc.)
  • ButtonBase — thin layer: PendingMixin(SharedButtonBaseCore) + aria-disabled forwarding

  Close-button no longer needs @internal pending overrides or the CEM plugin → plugin removed.

**Review by Rajdeepc — CHANGES_REQUESTED** [review]
Blocking this! You added a hybrid approach with `ButtonBaseCore + SharedButtonbase and PendingMixin`, This is creating 2 new classes why not go with pure abstract class hierarchy which will create 1 new class only. Mixin will have type safety and `instanceof` issues as i have noted.

**Comment by TarunAdobe** [comment]
note: ci failures are coming from main not from any changes in this branch

File: 2nd-gen/packages/core/components/close-button/CloseButton.base.ts (hunk 2 above)
Comment (caseyisonit) [reply 2nd-gen/packages/core/components/close-button/CloseButton.base.ts]: if we are making a base class for close button it should follow the other base class patterns, i prefer that we do not define multiple base classes in a single file. each one should be their own including the interactive button base

File: 2nd-gen/packages/core/components/button/Button.base.ts
Comment (caseyisonit) [inline 2nd-gen/packages/core/components/button/Button.base.ts:1]: I would like to advocate that button base be the shared button base. the pending mixin should be applied at the swc button level. if theres disagreement then i will request that shared button base (i would want to discuss the naming of this as well) should be treated like every other base class and have its own directory for discoverability. or we need to discuss architecture of these kinds of components and how we organize them, i can see this coming up with cards too.

**Review by caseyisonit — CHANGES_REQUESTED** [review]
i would like a discussion or a mini RFC about the shared button base and architecture

File: 2nd-gen/packages/core/components/button/Button.base.ts
Comment (caseyisonit) [reply 2nd-gen/packages/core/components/button/Button.base.ts:1]: the concepts of base classes is that they store core functionality. i would almost argue that we dont need a core base class per component. i would use one button base for all buttons and then in SWC add the differing functionality that makes them unique components. i thought that was the pattern we were aiming for.

Comment (TarunAdobe) [reply 2nd-gen/packages/core/components/button/Button.base.ts:1]: As per our discussion yesterday I have removed the mixin, created a controller and removed the sharedButtonbase. 
So buttonbase is the base that other buttons can use. 
One thing though right now that is still there is a closebuttonbase which extends the button base and overrides close button specific size and color properties. Let me know if this is okay or do we want to not hav ea close button base at all.

File: 2nd-gen/packages/swc/components/button/Button.ts [hunk 3]
```diff
@@ -127,6 +148,25 @@ export class Button extends ButtonBase {
...
+  }
+
+  protected override handleActivationClick(event: Event): void {
+    if (this.disabled || this.pending) {
+      event.preventDefault();
+      event.stopImmediatePropagation();
+    }
+  }
+
+  protected get handleClick(): (event: Event) => void {
```
Rajdeepc: This is creating a new function reference on every getter call. You can make this a bound arrow property. very small performance issue just fyi!

Rajdeepc (approved):
Thanks for doing this and being so patient. Kindly check my above comment here too around scalability.

2nd-gen/packages/core/components/button/Button.base.ts
Rajdeepc: @caseyisonit @TarunAdobe I like this pattern except for the fact that it has scalability issue. In future  if 3+ more components uses `PendingController` you will have 3-4 hosts each copy pasting the same property block. In that case you need to revisit this and extract a `PendingButtonBase` class or a helper. For now this is fine.

2nd-gen/packages/swc/components/button/Button.ts
```diff
@@ -29,6 +29,7 @@ import {
   type ButtonStaticColor,
   type ButtonVariant,
 } from '@spectrum-web-components/core/components/button';
+import { PendingController } from '@spectrum-web-components/core/controllers/pending-controller/index.js';
```
blunteshwar: `Button.ts` imports `PendingController` from `@spectrum-web-components/core/controllers/pending-controller/index.js`, but that subpath is not exported by `@spectrum-web-components/core/package.json`. Consumers/builds that respect package `exports` will fail to resolve this import. Since `PendingController` is already re-exported from `controllers/index.ts`, can this import use `@spectrum-web-components/core/controllers/index.js` instead, or can the new subpath be added to `package.json` exports?

2nd-gen/packages/swc/components/close-button/CloseButton.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  protected override render(): TemplateResult {
+    const resolvedName = this.getResolvedAccessibleName();
+
+    return html`
+      <button
+        class=${classMap({
+          'swc-CloseButton': true,
+          'swc-CloseButton--staticWhite': this.staticColor === 'white',
+          'swc-CloseButton--staticBlack': this.staticColor === 'black',
```
blunteshwar, line 69: These static-color classes are applied, but `close-button.css` does not define `.swc-CloseButton--staticWhite` or `.swc-CloseButton--staticBlack`, so `static-color="white"` / `"black"` currently has no visual effect. Can we add the corresponding CSS color rules, plus coverage that verifies the attribute changes the rendered color?

blunteshwar (changes requested):

2nd-gen/packages/swc/components/close-button/CloseButton.ts
TarunAdobe: we will do these in the css work that will be in the follow up pr not this one!!

2nd-gen/packages/swc/components/button/Button.ts
TarunAdobe: Good catch I have switched to @spectrum-web-components/core/controllers/index.js since PendingController is already re-exported there and that path is in core package exports.

blunteshwar (approved):

2nd-gen/packages/core/components/button/Button.base.ts
caseyisonit: in the example you provided, thats the intended pattern of the pendingController. a pendingButtonBase would only ever exist if there are 3 unique `role=button` type components that need pending. based on gen1 that use case doesnt exist. but pending controller can be used in search-field, combobox, etc.

2nd-gen/packages/core/components/button/Button.base.ts
```diff
@@ -36,9 +35,6 @@ import { BUTTON_VALID_SIZES } from './Button.types.js';
  * @slot icon - Optional leading icon.
  *
  * @attribute {ElementSize} size - The size of the button.
- *
- * @todo We currently have 3 levels of mixins on this class, but the mixin
- * composition guide recommends a maximum of 2. Explore reducing after milestone 2.
```
caseyisonit, line 41: should not be removed. this todo is not resolved

2nd-gen/packages/core/components/button/Button.base.ts
```diff
@@ -23,11 +23,10 @@ import {
 import { BUTTON_VALID_SIZES } from './Button.types.js';
 
 /**
- * Abstract base class for all button-like components. Owns shared semantic
- * concerns: interaction state, sizing, slot-derived icon/label state,
- * accessible-name resolution, and host-to-internal-button attribute forwarding.
+ * Abstract base for button-like components that share sizing, slots, disabled
+ * state, and accessible-name resolution.
```
caseyisonit: nit: prefer the original description of this

2nd-gen/packages/core/components/button/Button.base.ts
```diff
@@ -23,11 +23,10 @@ import {
...
  *
- * Visual API specific to `sp-button` (`variant`, `fill-style`, `static-color`)
+ * Visual API specific to `swc-button` (`variant`, `fill-style`, `static-color`)
```
caseyisonit: Static color is a shared property across all button type components and it should be defined here.

2nd-gen/packages/core/components/close-button/CloseButton.base.ts
caseyisonit, line 1: we should not need this base class or types once static-color is added to button base. the sizes are the same so you shouldnt need to override them again. the only thing that needs to move to swc-close-button is the override for hasIcon but i dont know why that is needed either.

2nd-gen/packages/core/controllers/pending-controller/src/pending-controller.ts
```diff
@@ -0,0 +1,146 @@
...
+        );
+      }
+    }
+  }
+
+  private handlePendingChange(): void {
+    if (this.host.pending) {
+      this._pendingTimer = setTimeout(() => {
+        if (this.host.pending) {
+          const internalButton = this.host.renderRoot.querySelector('button');
```
caseyisonit: is this locking down the controller to only render inside of a button? Thats too tight of a contract, this could be used in inputs for instance. let me know if im reasoning this wrong. :)

2nd-gen/packages/core/components/button/Button.base.ts
```diff
@@ -36,9 +35,6 @@ import { BUTTON_VALID_SIZES } from './Button.types.js';
  * @slot icon - Optional leading icon.
  *
  * @attribute {ElementSize} size - The size of the button.
- *
- * @todo We currently have 3 levels of mixins on this class, but the mixin
- * composition guide recommends a maximum of 2. Explore reducing after milestone 2.
  */
 export abstract class ButtonBase extends SizedMixin(
```
caseyisonit, line 44: This should be using the DisabledMixin

2nd-gen/packages/core/controllers/pending-controller/src/pending-controller.ts
```diff
@@ -0,0 +1,146 @@
...
+  /**
+   * Suppresses click activation while the button is in a `pending` state.
+   * Wire onto the rendered `<button>` via `@click`.
+   */
+  public readonly handleClick = (event: Event): void => {
+    if (this.host.pending) {
+      event.preventDefault();
+      event.stopImmediatePropagation();
+    }
+  };
```
caseyisonit: This feels specific to buttons, i cant see this being necessary for other components.

caseyisonit (changes requested):
provided more feedback

- the pending controller needs to be agnostic and its currently very intertwined with button logic
- close button shouldnt need its own base class now that pending is removed from baseButton
- static-colors needs to be added to button base since thats a shared public api across all buttons

2nd-gen/packages/core/controllers/pending-controller/src/pending-controller.ts
caseyisonit: a lot of this lifecycle method logic should be moved to swc-button where the controller is implemented.

2nd-gen/packages/core/controllers/pending-controller/src/pending-controller.ts
caseyisonit, line 1: The scope of this can be dramatically reduced. The pending-controller should essentially migrate the PendingState reactive controller in gen1 but instead of rendering a sp-progress-circle it should just render an SVG with the pending animation.

caseyisonit: quick reference to what im talking about:
docs - https://github.com/adobe/spectrum-web-components/blob/main/1st-gen/tools/reactive-controllers/pending-state.md
src to be migrated - https://github.com/adobe/spectrum-web-components/blob/main/1st-gen/tools/reactive-controllers/src/PendingState.ts

TarunAdobe:
removed all the pendingcontroller work from here so this is purely just 2nd gen setup for close button :)

PR #6357, opened 2026-06-01
feat(popover): lifecycle, positioning, a11y + styling (phase 4+5)

### LINKED ISSUE #5731: [Bug]: Overlay in focus region dismissed when click into its content
### Library version

1.7.0, {**}<ins>but persists in latest as well</ins>{**}.
### Expected behavior

When using sp-overlay (type=auto) with a button and the user clicks on any item in the overlay content - the overlay should stay open.
### Actual behavior

When that button and sp-overlay are wrapped into a focusable div (with tabindex=0) clicking into the overlay content auto-closes it.
### Screenshots

*No response*

*No response*
1. Go to <https://stackblitz.com/edit/vitejs-vite-tqg9vjko?file=src%2Fmy-element.ts>
1. Click on first or second button
1. See overlay content with popover
1. Click on overlay content
1. See overlay closes
1. Click on 3rd button (that is not wrapped into focusable div) - and overlay doesn't close when clicking into its content

rubencarvalho:
## Description

Phase 4 (a11y) + Phase 5 (styling) of the Popover migration, plus the open/show **lifecycle** deferred from Phase 3 (without it the component was invisible). Hardened through review. Scoped to Popover; later-phase items tracked below.

You can play with it here: https://swcpreviews.z13.web.core.windows.net/pr-6357/docs/second-gen-storybook/?path=/story/components-popover--playground&args=open:!true;actual-placement:bottom

## What's in it

- **Lifecycle** — `open` drives `showPopover()` (default `<div popover="auto">`) / `showModal()` (modal `<dialog>`); `swc-open` / `-after-open` / `-close` / `-after-close` events; `swc-close.detail.source` is `escape` / `outside` / `programmatic`.
- **Positioning** — `PlacementController` anchored to the resolved trigger, tip via the `arrow` middleware; surface is `position: absolute` to match the controller (fixes scroll drift). Flip-resolved side reflected as an internal `actual-placement` attribute (CSS-only, not a public property).
- **a11y** — `aria-controls` via the element-reference IDL, `aria-expanded`, modal `aria-haspopup="dialog"` + native focus trap; Escape coordinated through a dismissible stack; focus returns to the trigger on close; modal page scroll-lock.
- **CSS** — S2 surface + tip, modal `::backdrop` (no dimming), forced-colors.
- **Shared `core/utils`** — `resolve-trigger`, `dismissible-stack`, `transition`, `deep-contains` (checked: no existing equivalent to reuse).
- **Fixes #5731** (overlay in a focus region dismissed on content click) via native light-dismiss, plus broad tests (lifecycle, modal, focus, scroll-lock, nested layering, placement).

## Later-phase work

- Changeset (publish step); Phase 6 test depth + utils unit tests + screen-reader passes; Phase 7 docs MDX.
- Tooltip `popover="manual"` follow-up so a tooltip doesn't light-dismiss an open popover (Tooltip is reverted here to keep this PR Popover-only).
- Tip pixel/forced-colors polish; VRT.

## Related

- Epic SWC-1993; fixes #5731.

## Notes

- Targets the `ruben/popover-migration` integration branch; the #6354 API work appears in this diff until that merges.

- [ ] **Keyboard** — default: Escape + click-outside dismiss, no trap. Modal: native focus trap + Escape (`cancel`), backdrop-click closes, focus returns to the trigger. _Full manual pass in Phase 6._
- [ ] **Screen reader** — trigger announces the controls relationship + expanded state; modal announces `role="dialog"`. _Pending Phase 6._


2nd-gen/packages/core/components/popover/Popover.base.ts
```diff
@@ -13,23 +13,65 @@
...
+import {
+  deepContains,
+  getActiveElement,
+  isTopDismissible,
+  physicalSide,
+  registerDismissible,
+  resolveTrigger,
+  runAfterTransition,
+  unregisterDismissible,
+} from '@spectrum-web-components/core/utils/index.js';
```
rubencarvalho, line 31: I created these utils in preparation for the Tooltip to use them too.
As a follow-up, we'll refactor it to use the same source as Popover.

2nd-gen/packages/swc/components/popover/popover.css
```diff
@@ -10,18 +10,192 @@
...
+ * (`margin: auto`, `inset: 0`) and let the controller anchor it. */
 .swc-Popover {
+  /* Arrow clearance: the space the tip occupies between the surface and the
+   * trigger when the arrow is shown. The component reads this token-derived
+   * value and adds it to the PlacementController's main-axis offset (a CSS
+   * margin can't shift the absolutely-positioned surface on every side), so the
+   * gap stays in sync with the tip dimensions. */
+  --_swc-popover-tip-height: token("popover-tip-height");
+
+  position: absolute;
```
rubencarvalho, line 37: `position: absolute` (not `fixed`) is deliberate: it matches the PlacementController's Floating UI `strategy: 'absolute'`. With `fixed` + absolute coordinates the popover drifted on page scroll.

2nd-gen/packages/swc/components/popover/popover.css
```diff
@@ -10,18 +10,192 @@
...
+
+/* Shown state: open and anchored. Positioning is async (the controller awaits
+ * `document.fonts.ready`), so the fade is gated on `actual-placement`, which the
+ * component sets only once the first placement lands; otherwise the surface would
+ * fade in at its reset 0,0 origin and jump to the trigger mid-transition. The
+ * attribute is added after the element is rendered, so the entry is an ordinary
+ * attribute-driven transition (no `@starting-style`). `:popover-open` is for the
+ * default `<div>`, `[open]` for the modal `<dialog>`. Ordered after the
+ * lower-specificity `.swc-Popover` rules above for `no-descending-specificity`. */
+:host([actual-placement]) .swc-Popover:popover-open,
```
rubencarvalho, line 133: Fade/visibility is gated on `actual-placement` (set by the controller's first compute), not `:popover-open` alone. Positioning is async (awaits `document.fonts.ready`), so revealing earlier paints the surface at its reset 0,0 origin and then jumps to the trigger. claude helped me with this so this now does 2 things: orientation + "anchored, safe to show".

2nd-gen/packages/swc/components/popover/popover.css
```diff
@@ -10,18 +10,192 @@
...
-  display: block;
+  display: flex;
+  flex-direction: column;
+  padding: token("popover-edge-to-content-area");
+}
+
+/* Modal mode renders no dimming backdrop: keep the `<dialog>`'s backdrop
+ * transparent (overriding any UA default). The backdrop pseudo-element still
+ * exists, so backdrop-click light-dismiss continues to work. */
+.swc-Popover::backdrop {
```
rubencarvalho, line 158: Modal backdrop is intentionally transparent, to match the React approach

2nd-gen/packages/core/components/popover/Popover.base.ts
```diff
@@ -191,4 +341,519 @@ export abstract class PopoverBase extends SpectrumElement {
...
+  // button's inner control) is detected; the content is slotted, so its DOM
+  // ancestors lead back to this host.
+  private _isFocusWithin(): boolean {
+    return deepContains(this, getActiveElement());
+  }
+
+  // Lock page scroll behind a modal popover. The component's shadow stylesheet
+  // cannot reach `html`, so this is done in JS: set `overflow: hidden` on the
+  // document element, saving the prior inline value to restore on unlock.
+  private _lockPageScroll(): void {
```
rubencarvalho: Need to lock the scroll for the modal, to align with React.

2nd-gen/packages/core/components/popover/Popover.base.ts
```diff
@@ -191,4 +341,519 @@ export abstract class PopoverBase extends SpectrumElement {
...
+  }
+
+  // ──────────────────
+  //     OPEN/CLOSE
+  // ──────────────────
+
+  // `reShow` is true when re-engaging the show API for a popover that is already
+  // open (a `modal` toggle swaps the internal element). It suppresses a second
+  // `swc-open` for a popover that never closed.
+  private _show(reShow = false): void {
```
rubencarvalho, line 611: `reShow` handles toggling `modal` while open: Lit swaps the internal element (`<div popover>` <-> `<dialog>`), so we re-run the show API on the new one. It suppresses a second `swc-open` since the popover never actually closed.
I don't think this will happen on actual products, but it was annoying enough on our Storybook playground, if you are testing modal vs non-modal

2nd-gen/packages/swc/components/popover/test/popover.test.ts
rubencarvalho, line 1: I know tests are for the next phase, but I could not validate some things I was doing without them :/

rubencarvalho (commented):
Left some context in the comments

2nd-gen/packages/core/components/popover/Popover.base.ts
```diff
@@ -191,4 +341,519 @@ export abstract class PopoverBase extends SpectrumElement {
...
+    this._prevDocumentOverflow = html.style.overflow;
+    html.style.overflow = 'hidden';
+    this._pageScrollLocked = true;
+  }
+
+  private _unlockPageScroll(): void {
+    if (!this._pageScrollLocked) {
+      return;
+    }
+    document.documentElement.style.overflow = this._prevDocumentOverflow;
```
Rajdeepc: there might be a race condition here. If two modal popovers opened simultaneously the second instances `_prevDocumentOverflow` will be `hidden` from first instance's lock. This might lock the page scroll permanently until the first model is also dismissed. 
How about you use a module level scope here maybe a reference counter.  Check `dismissible-stack.ts`

2nd-gen/packages/core/utils/resolve-trigger.ts
```diff
@@ -36,13 +36,33 @@ export interface ResolvedTrigger {
...
+  if (!trigger) {
+    return { trigger: null, interactiveElement: null };
+  }
+
+  // Inner-button discovery: for a 2nd-gen component trigger with an open shadow
+  // root, wire ARIA on the AT-facing inner <button>. Closed-shadow or native
+  // triggers fall back to the trigger host itself.
+  const innerButton = trigger.shadowRoot?.querySelector('button') ?? null;
+
+  return { trigger, interactiveElement: innerButton ?? trigger };
```
Rajdeepc, line 67: if this is for inner button discovery the querySelector has no scoping. `shadowRoot?.querySelector('button')` is a depth first DOM traversal and will return the deepest nested first button if any component renders nested buttons.

Can we define a standard for 2nd gen component to expose their interactive element via a known interface. This will make the contract explicit. We should document this requirement in the contributor guide too before downstream migrations run.

2nd-gen/packages/core/components/popover/Popover.base.ts
```diff
@@ -13,23 +13,65 @@
...
+interface ARIAControlsElements {
+  ariaControlsElements?: readonly Element[] | null;
+}
+
+/**
+ * Span of a single click gesture (pointerdown through click). A trigger click
+ * within this window after a native light-dismiss is read as the close rather
+ * than a reopen. See `_onTriggerClick`.
+ */
+const LIGHT_DISMISS_REOPEN_WINDOW_MS = 200;
```
Rajdeepc: Is this a environment assumption or battle tested timeout?
Why dont we track the last close source triggered by a pointerdown event on the trigger element specifically instead and eliminate this fragile dependency?

2nd-gen/packages/swc/components/popover/Popover.ts
```diff
@@ -31,21 +35,35 @@ import styles from './popover.css';
...
+    const surface = this.internalElement;
+    if (!surface) {
+      return 0;
+    }
+    return (
+      parseFloat(
+        getComputedStyle(surface).getPropertyValue('--_swc-popover-tip-height')
+      ) || 0
+    );
+  }
```
Rajdeepc, line 84: The custom property becomes stable after the first render and will not change at runtime. Why don't we cache this value e.g. reading it once on first open to avoid repeated reflow triggers on every positioning update.

2nd-gen/packages/core/utils/resolve-trigger.ts
```diff
@@ -36,13 +36,33 @@ export interface ResolvedTrigger {
...
- * @todo Phase 4/5: implement `for=` resolution via
- * `host.getRootNode().getElementById()`, the `triggerElement` override, and
- * inner-button discovery (`trigger.shadowRoot?.querySelector('button')`).
+ * Resolution order: a direct `triggerElement` reference wins; otherwise `for`
+ * is resolved by ID within the host's tree root (`getRootNode().getElementById`,
+ * strictly same-root). When the resolved trigger exposes an open shadow root,
+ * its inner `<button>` becomes the `interactiveElement` that receives ARIA;
+ * otherwise ARIA is wired on the trigger itself.
  */
 export function resolveTrigger(
```
Rajdeepc, line 45: Would love to have a standalone test of this utility like open shadow root and then the inner button discovery path.

Rajdeepc (changes requested):
First pass: Some observations/feedback. 
I see that your plan still has unchecked boxes for `trigger-disconnect` degradation and stable instance IDs. Are those blocking for a merge into main, or deferred?

2nd-gen/packages/swc/components/popover/stories/popover.stories.ts
```diff
@@ -48,16 +81,76 @@ const meta: Meta = {
...
 export default meta;
 
+// ────────────────────
+//    HELPERS
+// ────────────────────
+
+// A compact account card composed from existing components (avatar, divider,
+// button) and the global typography classes. Passed as the popover's slotted
+// content to show a realistic, polished example rather than a bare sentence.
+const accountCard = `
```
nikkimk, line 99: I could not fully test the a11y without the stories themselves being accessible.

nikkimk (changes requested):
I know docs isn't in scope of this PR, but I could not fully test the a11y without the stories themselves being accessible.

2nd-gen/packages/swc/components/popover/stories/popover.stories.ts
rubencarvalho: I brought those in!

2nd-gen/packages/core/components/popover/Popover.base.ts
rubencarvalho: Ooops, good call - I extracted this to a controller, that should address your concerns!

2nd-gen/packages/core/utils/resolve-trigger.ts
rubencarvalho: It was covered on the tests phase, but I pulled my test from there here.

2nd-gen/packages/core/utils/resolve-trigger.ts
rubencarvalho: Ye... this is a good call, I've commented on the tooltip PR with something similar... this is correct for every second-generation trigger we ship today (button, action-button) - and one of the reasons was we wanted to have some A11Y control at first. I’d prefer not to this exercise in this PR while this assumption holds (or we get a request for something different). WDYT?

2nd-gen/packages/core/components/popover/Popover.base.ts
rubencarvalho: good point, I was also on the fence about this and discussed with AI back and forth. I did apply your suggestion now.

2nd-gen/packages/swc/components/popover/Popover.ts
rubencarvalho: cached!

2nd-gen/packages/swc/components/popover/stories/popover.stories.ts
```diff
@@ -40,24 +80,100 @@ const meta: Meta = {
...
+      <div style="display: flex; flex-direction: column; gap: 2px;">
+        <span class="swc-Title swc-Title--sizeS">Jess Lee</span>
+        <span class="swc-Detail swc-Detail--sizeS">Product designer</span>
+      </div>
+    </div>
+    <swc-divider size="s"></swc-divider>
+    <p class="swc-Body swc-Body--sizeS" style="margin: 0;">
+      Manage your profile, notification preferences, and connected apps.
+    </p>
+    <div style="display: flex; gap: 8px; justify-content: flex-end;">
```
5t3ph: Since you're using other swc components, might as well have this use Button Group :) 

Same for other similar instances

2nd-gen/packages/swc/components/popover/stories/popover.stories.ts
```diff
@@ -40,24 +80,100 @@ const meta: Meta = {
...
+        alt=""
+        size="300"
+      ></swc-avatar>
+      <div style="display: flex; flex-direction: column; gap: 2px;">
+        <span class="swc-Title swc-Title--sizeS">Jess Lee</span>
+        <span class="swc-Detail swc-Detail--sizeS">Product designer</span>
+      </div>
+    </div>
+    <swc-divider size="s"></swc-divider>
+    <p class="swc-Body swc-Body--sizeS" style="margin: 0;">
```
5t3ph: since you are using our typography classes, `margin: 0` is not needed as the class uses a `0` default (same for other instances)

2nd-gen/packages/swc/components/popover/popover.css
```diff
@@ -10,18 +10,192 @@
...
+   * gap stays in sync with the tip dimensions. */
+  --_swc-popover-tip-height: token("popover-tip-height");
+
+  position: absolute;
+  inset: auto;
+  inline-size: max-content;
+
+  /* Never exceed the viewport minus 2 × container padding. */
+  max-inline-size: calc(100vw - 24px);
+  max-block-size: none;
```
5t3ph: Adjust per the example in tooltip so that you also take advantage of the provided custom properties from the PlacementController, which especially matters for right/left placements.

2nd-gen/packages/swc/components/popover/popover.css
```diff
@@ -10,18 +10,192 @@
...
+}
+
+/* The padded content wrapper. Padding lives here (not on `.swc-Popover`) so a
+ * backdrop-click on the dialog's surface is not absorbed by padding in modal
+ * mode. */
 .swc-Popover-content {
-  display: block;
+  display: flex;
+  flex-direction: column;
+  padding: token("popover-edge-to-content-area");
```
5t3ph: This is imposing a bit too much - slotted content should control their `display` related behavior.

And this is probably also a good case to "break" our custom property exposure rules and go ahead and expose padding since the intent of popover is a blank canvas container.

```suggestion
  padding: var(--swc-popover-content-padding, token("popover-edge-to-content-area"));
```

2nd-gen/packages/swc/components/popover/popover.css
5t3ph: Interesting - this probably would solve the `@starting-style` related bug I was experiencing with tooltip 🤔 I wasn't considering that `actual-placement` provided the visibility hint/delay.

2nd-gen/packages/swc/components/popover/popover.css
```diff
@@ -10,18 +10,192 @@
...
+  /* Never exceed the viewport minus 2 × container padding. */
+  max-inline-size: calc(100vw - 24px);
+  max-block-size: none;
   padding: 0;
   margin: 0;
-  border: none;
+  color: inherit;
+  background-color: token("background-layer-2-color");
+  border: token("border-width-100") solid token("popover-border-color");
+  border-radius: token("corner-radius-100");
```
5t3ph: React shows a `10px` value which is this token:

```suggestion
  border-radius: token("corner-radius-700");
```

2nd-gen/packages/swc/components/popover/popover.css
```diff
@@ -10,18 +10,192 @@
...
 * {
   box-sizing: border-box;
 }
 
+/* The host stays in the light DOM and owns no layout; the internal element
+ * (a `<div popover="auto">` or a `<dialog>`) enters the top layer when opened. */
+:host {
+  display: contents;
+}
+
```
5t3ph, line 22: React does do a slight animate out from the trigger, just like tooltip. Do you want to tackle that here (using tooltip as reference) or in a separate pass?

2nd-gen/packages/swc/components/popover/stories/popover.stories.ts
```diff
@@ -40,24 +80,100 @@ const meta: Meta = {
...
 
+// ────────────────────
+//    HELPERS
+// ────────────────────
+
+// A compact account card composed from existing components (avatar, divider,
+// button) and the global typography classes. Passed as the popover's slotted
+// content to show a realistic, polished example rather than a bare sentence.
+const accountCard = `
+  <div style="display: flex; flex-direction: column; gap: 16px; inline-size: 240px;">
```
5t3ph: Remove this inline size or it makes it odd for the Playground when adjusting sizes.
```suggestion
  <div style="display: flex; flex-direction: column; gap: 16px;">
```

2nd-gen/packages/swc/components/popover/stories/popover.stories.ts
```diff
@@ -40,24 +80,100 @@ const meta: Meta = {
...
+      <swc-button variant="accent" size="s">Open settings</swc-button>
+    </div>
+  </div>
+`;
+
+// Playground content with several focusable controls so the modal focus trap is
+// demonstrable: enable `modal` in the controls, open the popover, then press Tab
+// to see focus cycle within these buttons instead of escaping to the page.
+const playgroundContent = `
+  <div style="display: flex; flex-direction: column; gap: 12px; inline-size: 260px;">
```
5t3ph: Remove this inline size or it makes it odd for the Playground when adjusting sizes. Maybe then show this example starting at size small (like React), or reduce this default content.


```suggestion
  <div style="display: flex; flex-direction: column; gap: 12px;">
```

2nd-gen/packages/swc/components/popover/stories/popover.stories.ts
```diff
@@ -40,24 +80,100 @@ const meta: Meta = {
...
+    <swc-divider size="s"></swc-divider>
+    <p class="swc-Body swc-Body--sizeS" style="margin: 0;">
+      Manage your profile, notification preferences, and connected apps.
+    </p>
+    <div style="display: flex; gap: 8px; justify-content: flex-end;">
+      <swc-button
+        variant="secondary"
+        fill-style="outline"
+        size="s"
+        onclick="this.closest('swc-popover').open = false"
```
5t3ph, line 121: For discussion/can table for later: do we maybe want to look into providing a declarative method for close?

2nd-gen/packages/swc/components/popover/popover.css
```diff
@@ -10,18 +10,192 @@
...
+  border: token("border-width-100") solid token("popover-border-color");
+  rotate: 45deg;
+}
+
+/* Oriented onto an edge (any side): safe to show. */
+:host([actual-placement]) .swc-Popover-tip {
+  visibility: visible;
+}
+
+/* Tip on the top edge: popover sits below the trigger. */
```
5t3ph, line 197: You're currently exposing all placement values, which leads to incorrect arrow placement for now:

Image

Do you want to resolve now, or pull back on what's exposed and resolve in a follow-up?

2nd-gen/packages/swc/components/popover/popover.css
```diff
@@ -10,18 +10,192 @@
...
+  border-bottom: none;
+  border-left: none;
+}
+
+/* Tip on the right edge: popover sits to the left of the trigger. */
+:host([actual-placement="left"]) .swc-Popover-tip {
+  right: calc(var(--_swc-popover-tip-size) / -2);
+  border-top: none;
+  border-right: none;
+}
```
5t3ph: Need to flip values per testing for forced-colors:

Image


```suggestion
:host([actual-placement="left"]) .swc-Popover-tip {
  left: calc(var(--_swc-popover-tip-size) / -2);
  border-bottom: none;
  border-left: none;
}

/* Tip on the right edge: popover sits to the left of the trigger. */
:host([actual-placement="right"]) .swc-Popover-tip {
  right: calc(var(--_swc-popover-tip-size) / -2);
  border-top: none;
  border-right: none;
}
```

5t3ph (commented):
Mostly comments on styles :)

2nd-gen/packages/swc/components/popover/popover.css
```diff
@@ -33,19 +33,25 @@
...
   inline-size: max-content;
 
-  /* Never exceed the viewport minus 2 × container padding. */
-  max-inline-size: calc(100vw - 24px);
+  /* Cap to the placement-aware usable width the PlacementController exposes
+   * (matters for left/right placements, where the available inline space is the
+   * gap beside the trigger); falls back to the viewport minus container padding
+   * before the first position is computed. */
+  max-inline-size: var(--swc-placement-available-width, calc(100vw - 24px));
   max-block-size: none;
```
5t3ph: Use `min()` to have the browser consider these as dynamic options ("smallest" wins to retain some visual margin).

```css
max-inline-size: min(var(--swc-placement-available-width), calc(100vi - 24px));
max-block-size: min(var(--swc-placement-available-height), 90vb);
```

2nd-gen/packages/swc/components/popover/popover.css
```diff
@@ -142,8 +180,11 @@
   position: absolute;
   inline-size: var(--_swc-popover-tip-size);
   block-size: var(--_swc-popover-tip-size);
-  background-color: token("background-layer-2-color");
-  border: token("border-width-100") solid token("popover-border-color");
+
+  /* Reuse the surface custom properties so theming the surface also themes the
+   * arrow. */
+  background-color: var(--swc-popover-background-color, token("background-layer-2-color"));
+  border: token("border-width-100") solid var(--swc-popover-border-color, token("popover-border-color"));
```
5t3ph: These can simply use inheritance since these props are initially applied on the wrapper:
```css
background-color: inherit;
border: inherit;
```

2nd-gen/packages/core/components/popover/Popover.types.ts
```diff
@@ -21,18 +21,31 @@ import {
...
- * The full set of placement values accepted by `<swc-popover>`. Downstream
- * first-party components narrow this set per the proxy pattern.
+ * The placement values `<swc-popover>` supports: each physical side plus its two
+ * alignments, matching the Spectrum design API. The controller can resolve more
+ * (the logical `start`/`end` variants), but the popover narrows to this set per
+ * the proxy pattern.
  */
-export const POPOVER_VALID_PLACEMENTS =
-  ALL_PLACEMENTS satisfies readonly Placement[];
+export const POPOVER_VALID_PLACEMENTS = [
```
5t3ph, line 37: Maybe I'm not interpreting this correctly, but this seems to be excluding logical placements which doesn't address the real issue of the tip alignment for something like `bottom-left` (arrow aligned to the left corner) vs `bottom` (arrow aligned in the center)

5t3ph (commented):
I'll check the transforms once the build is working again.

nikkimk (changes requested):
As we discussed in Slack, let me know when you want me to re-test for a11y.

2nd-gen/packages/core/components/popover/Popover.types.ts
rubencarvalho: That was a bad commit - it is fixed now. WDYM about th arrow alignment? From what I can see, it is working as expected in the latest commit :)

2nd-gen/packages/swc/components/popover/popover.css
rubencarvalho: Done!

2nd-gen/packages/swc/components/popover/popover.css
rubencarvalho: Done!

2nd-gen/packages/swc/components/popover/popover.css
rubencarvalho: Fixed!

rubencarvalho: Only 12 are exposed now (3 for each side)

2nd-gen/packages/swc/components/popover/stories/popover.stories.ts
rubencarvalho: Ah, I forgot this from when I was testing the sizes. Removed now!

2nd-gen/packages/swc/components/popover/popover.css
rubencarvalho: I hadn't notice that, actually. I needed to slow it down to notice! 😄 I updated to do that too.

2nd-gen/packages/swc/components/popover/popover.css
rubencarvalho: done!

2nd-gen/packages/swc/components/popover/popover.css
rubencarvalho: yes! I also documented it in the API table

2nd-gen/packages/core/components/popover/Popover.types.ts
5t3ph: Looking at React's a bit more, their behavior matches. I was expecting the arrow to also shift, so that for example in a "bottom start" position the arrow would mimic with "top start", accounting for any cross offset as well. That is also [technically what design expects](https://www.figma.com/design/Mngz9H7WZLbrCvGQf3GnsY/S2---Web--Desktop-scale-?node-id=125644-16619&t=8y1C4klA4cLhF4ek-0), but we can follow-up on it in a different ticket, it doesn't need to be blocking.

2nd-gen/packages/core/components/popover/Popover.base.ts
```diff
@@ -635,23 +630,19 @@ export abstract class PopoverBase extends SpectrumElement {
...
-        // anchored position a tick later.
-        dialog.style.top = '0px';
-        dialog.style.left = '0px';
-        dialog.style.translate = `${window.scrollX}px ${window.scrollY}px`;
+        // `showModal()` synchronously moves focus into the dialog, scrolling the
+        // page to it, and has no `preventScroll` option. Positioning is async, so
+        // the surface is still at its 0,0 origin when focus lands. Capture and
+        // restore the scroll position around the call; it is synchronous, so the
+        // restore is seamless (no paint between).
+        const { scrollX, scrollY } = window;
```
5t3ph, line 638: Commenting here since scroll related - edge case of opening a modal when the trigger and possibly part or all of the modal is out of view. May seem odd but I can see this happening to keyboard users or in multi-pane apps where scroll is already modified.

Higher risk potential: you can also get it to happen when a trigger opens a modal on an anchor somewhere else.

Image

Scroll blocking prevents moving it into view. Should we do any sort of edge collision detection and scroll into view only in that case?

5t3ph (approved):
Approving, this item can be a ticket to investigate later if you prefer.

2nd-gen/packages/core/components/popover/Popover.types.ts
rubencarvalho: I think that makes sense, but we'll need to jump into a call because I'm not sure I'm looking at the right thing 😛

PR #6358, opened 2026-06-01
feat(core): add HoverController

5t3ph:
## Description

Adds `HoverController`, a Lit `ReactiveController` that centralises hover and keyboard-focus event wiring for 2nd-gen components that use the native Popover API (Tooltip, and any future popover-adjacent component).

### What it provides

- **Warm-up / cooldown timing** — the popover opens only after the pointer has rested on the trigger for `delay` ms; leaving starts a matching cooldown. `delay="0"` opens and closes synchronously.
- **Warm-state acceleration** — once any instance of a component type warms up, subsequent hovers on the same type open immediately without re-timing.
- **WCAG 1.4.13 pointer bridge** — moving the pointer from the trigger into the popover cancels the cooldown, satisfying the "hoverable" requirement so users can interact with popover content.
- **Keyboard focus priority** — `Tab` focus opens immediately and suppresses all pointer-driven timers until `focusout`.
- **Pointer-click exclusion** — a pointer click does not open the popover. `popover="auto"` fires a light dismiss on `pointerdown` (the trigger is outside the popover surface); opening on the subsequent `focusin` would cause a visible flash. The controller detects this event sequence and skips the open.
- **`disabled` / `manual` guards** — re-evaluated live on every host update, no page reload required.

### Basic usage

```typescript
import { LitElement, type PropertyValues } from 'lit';
import {
  HoverController,
  type HoverControllerHost,
} from '@spectrum-web-components/core/controllers/hover-controller.js';

class SwcTooltip extends LitElement implements HoverControllerHost {
  @property({ type: Number }) delay = 1500;
  @property({ type: Boolean }) manual = false;
  @property({ type: Boolean }) disabled = false;

  private triggerElement: HTMLElement | null = null;

  private readonly hoverController = new HoverController(this, {
    warmStateKey: 'swc-tooltip',
  });

    super.updated(changes);
    if (changes.has('triggerElement')) {
      this.hoverController.setTarget(this.triggerElement ?? null);
    }
  }
}
```

**Method**

| Member | Description |
|---|---|
| `setTarget(trigger: HTMLElement \| null)` | Sets the element that receives pointer and focus listeners. Call from `updated()` whenever the resolved trigger changes. Passing `null` detaches all listeners. |

**Constructor options**

| Option | Type | Description |
|---|---|---|
| `warmStateKey` | `string` | Per-component-type key for shared warm state on `document`. Use the element tag name (e.g. `'swc-tooltip'`). Must be static. |

**`HoverControllerHost` interface**

| Member | Type | Description |
|---|---|---|
| `delay` | `number` | Warm-up and cooldown duration in ms. `0` = immediate. |
| `manual` | `boolean` | When `true`, controller skips all event wiring. |
| `disabled` | `boolean` | When `true`, controller skips all event wiring. |
| `showPopover()` | method | Called by the controller to open the popover. |
| `hidePopover()` | method | Called by the controller to close the popover. |

The controller does **not** resolve trigger elements, set ARIA attributes, or dispatch lifecycle events — those remain the consuming component's responsibility.

## Motivation and context

Tooltip, and any future component that uses `popover="auto"`, requires identical hover/focus logic. Extracting it into a shared controller keeps that logic in one place and tested independently before any consuming component is built.

## Related issue(s)

- fixes SWC-2211

## Author's checklist

- [x] I have included a well-written changeset if my change needs to be published. (internal-only controller for now; no changeset)

## Reviewer's checklist

### Manual review test cases

- [ ] _Warm-up and cooldown_
  1. Open the Hover controller Storybook story
  2. Hover "Trigger A" and hold until the tooltip appears (250 ms demo delay)
  3. Move immediately to "Trigger B"; expect it to open without delay (warm state)
  4. Leave trigger B and wait ~300 ms; expect tooltip to close
  5. Hover "Trigger B" again; expect it to require the full 250 ms warm-up (state has cooled)

- [ ] _WCAG 1.4.13 pointer bridge_
  1. Open the "WCAG 1.4.13 pointer bridge" story
  2. Hover the trigger until the tooltip appears
  3. Move the pointer from the trigger into the tooltip bubble without pausing outside
  4. Expect the tooltip to stay open indefinitely

- [ ] _Keyboard focus_
  1. Open the "Keyboard focus opens immediately" story
  2. Tab to the button; expect the tooltip to open at once (no delay)
  3. While focused, move the pointer off the button; expect tooltip to remain open
  4. Tab away; expect tooltip to close immediately

- [ ] _Pointer-click focus does not open_
  1. Open the Playground story with `delay=1500`
  2. Click the trigger button without hovering first
  3. Expect no tooltip to appear (click does not trigger focus-open)

### Device review

- [ ] Did it pass in Desktop?

  1. Open the "Keyboard focus opens immediately" story
  2. Press <kbd>Tab</kbd> to move focus to the trigger button
  3. Expect tooltip to open immediately without warm-up delay
  4. Move pointer away from trigger while focused; expect tooltip to remain open
  5. Press <kbd>Tab</kbd> again to move focus away; expect tooltip to close immediately
  6. Confirm focus indicator is visible on the trigger throughout

  1. `HoverController` does not set ARIA attributes directly; consuming components (e.g. Tooltip) are responsible for wiring `aria-describedby` and roles


2nd-gen/packages/swc/.storybook/DocumentTemplate.mdx
```diff
@@ -53,12 +64,50 @@ export const ConditionalSection = ({ tag, title, hideTitle = false }) => {
         </div>
     );
 
+};
+export const ConditionalApi = () => {
```
5t3ph: These changes copied from #6337 to enable showing the custom API story since it's not constructed from CEM.

2nd-gen/packages/swc/.storybook/DocumentTemplate.mdx
```diff
@@ -53,6 +53,39 @@ export const ConditionalSection = ({ tag, title, hideTitle = false }) => {
...
+    return (
+        <>
+            <HeaderMdx as="h2" id="api">API</HeaderMdx>
+            <ApiTable />
+        </>
+    );
+
+};
+
+export const ConditionalGettingStarted = () => {
```
Rajdeepc: This is dead code? Didnt see this in render

2nd-gen/packages/core/controllers/hover-controller/src/hover-controller.ts
```diff
@@ -0,0 +1,352 @@
...
+    if (this.host.delay > 0) {
+      warmState.cooldownTimer = setTimeout(() => {
+        warmState.cooldownTimer = null;
+        warmState.isWarm = false;
+        this.callHidePopover();
+      }, this.host.delay);
+    } else {
+      warmState.isWarm = false;
+      this.callHidePopover();
+    }
```
Rajdeepc: When `delay=0`, `startCooldown()` synchronously hides the popover. The pointer physically cannot move from the trigger to the popover host between pointerleave on the trigger and this synchronous close. The WCAG bridge is effectively dead for delay=0. If this is by design can you document this for delay=0 if WCAG compliance is required at all delay values.

2nd-gen/packages/core/controllers/hover-controller/src/hover-controller.ts
```diff
@@ -0,0 +1,352 @@
...
+
+  /** Re-evaluates `disabled` and `manual` guards whenever the host updates. */
+  public hostUpdated(): void {
+    if (this.host.disabled || this.host.manual) {
+      this.clearWarmupTimer();
+      this.hasFocusOpen = false;
+      this.hadPointerdown = false;
+    }
+    this.unwireTarget();
+    this.wireTarget();
```
Rajdeepc: When disabled transitions to true while the popover is open `unwireBridge` and `callHidePopover` are not called so they remain on host. When guards are active, Please call `this.unwireBridge()` and `this.callHidePopover()`.

2nd-gen/packages/core/controllers/hover-controller/src/hover-controller.ts
```diff
@@ -0,0 +1,352 @@
...
+    this.clearCooldownTimer();
+
+    if (this.host.delay === 0 || warmState.isWarm) {
+      this.showWithBridge();
+    } else {
+      this.warmupTimer = setTimeout(() => {
+        this.warmupTimer = null;
+        warmState.isWarm = true;
+        this.showWithBridge();
+      }, this.host.delay);
```
Rajdeepc, line 342: If `host.disabled` or `host.manual` becomes true during the warmup timer, `showWithBridge()` still fires. The `hostUpdated()` does clear the timer when guards activate, but there's a race condition here, if `clearWarmupTimer()` in hostUpdated runs after the timer callback is already dequeued but not yet executed, the popover opens when it shouldn't.

You can add a guard at the top of the timer callback:
```ts
this.warmupTimer = setTimeout(() => {
  this.warmupTimer = null;
  if (this.host.disabled || this.host.manual) return;
  warmState.isWarm = true;
  this.showWithBridge();
}, this.host.delay);
```

2nd-gen/packages/core/controllers/hover-controller/src/hover-controller.ts
Rajdeepc: This unconditionally removes and re-adds 5 event listeners on every Lit update cycle, regardless of whether disabled or manual actually changed. So every reactive property change on the host triggers a full listener teardown/setup. This can have performance concerns.
Can you track whether the effective should be wired state changed or rewire when it differs. Something like this
```ts
public hostUpdated(): void {
  const shouldWire = !this.host.disabled && !this.host.manual && !!this.target;
  const isWired = /* track a boolean for current wired state */;
  
  if (shouldWire === isWired) return;
  
  if (!shouldWire) {
    this.unwireTarget();
    this.clearWarmupTimer();
    this.hasFocusOpen = false;
    this.hadPointerdown = false;
  } else {
    this.wireTarget();
  }
}
```

Rajdeepc (changes requested):
I like how you have created the pointer bridge and keyboard focus. Some feedbacks on optimizations. Overall looks excellent.

2nd-gen/packages/swc/.storybook/DocumentTemplate.mdx
5t3ph: there was a conflict with recent updates on main, now resolved in the correct location

2nd-gen/packages/core/controllers/hover-controller/src/hover-controller.ts
5t3ph: updated

2nd-gen/packages/core/controllers/hover-controller/src/hover-controller.ts
5t3ph: updated

2nd-gen/packages/core/controllers/hover-controller/src/hover-controller.ts
5t3ph: updated

2nd-gen/packages/core/controllers/hover-controller/src/hover-controller.ts
5t3ph: This uncovered an important distinction I missed earlier in looking at React which is that the `delay` is really the _opening_ delay whereas the cooldown is an independent value that ensures time for the bridge regardless of that value.

Added an optional `closeDelay` that defaults to `300ms` (matching React's value) to keep the bridge intact.

Rajdeepc (approved):
Nice! A couple of questions/feedback you can do post merge too.
1. non-default closeDelay value don't have a test case. Can you please add a test with an explicit non-default closeDelay to prove the property flows through end-to-end.
2. Is closeDelay intended to be exposed as an attribute on consuming components (e.g. <swc-tooltip close-delay="200">) or remain an internal implementation detail?

blunteshwar (approved):
nit: Do you feel we need a changeset here?

PR #6359, opened 2026-06-01
chore(action-button): styling implementation

cdransf:
## Description

Implements Phase 5 (Render & Style) of the `swc-action-button` S2 migration. Migrates `action-button.css` from a stub to a full S2 implementation using Spectrum 2 design tokens, and updates the render template and stories to match.

**What changed:**

`action-button.css` — full S2 implementation:
- All five sizes (`xs`, `s`, `m`, `l`, `xl`) with correct S2 tokens for height, font size, gap, padding, icon size, and corner radius per size
- `quiet` treatment via `:host([quiet])` custom property overrides (transparent default, transparent disabled)
- `static-color="white"` and `static-color="black"` with full state color overrides including quiet combinations
- Interactive states: hover, focus-visible (outline/outline-offset), active, disabled (inflated `:disabled:is(*, :hover)` specificity pattern from `swc-button`)
- Pending state: animated inline SVG spinner with 1-second delay (width-locked via `--_swc-button-pending-inline-size` from `ButtonBase`), static-color spinner track/fill color overrides, reduced-motion treatment, Windows High Contrast forced-colors handling
- Icon slot and icon-only layout (`swc-ActionButton--iconOnly` modifier), same margin-negative edge-to-visual pattern as `swc-button`
- `@global-exclude` fences around all JS-runtime-only blocks so the generated global stylesheet stays static-only

`ActionButton.ts`:
- `nothing` import added; pending placeholder `<span class="swc-ActionButton-pendingIcon">` replaced with a conditional inline SVG spinner matching `swc-button` (only rendered when `this.pending`)
- Full `@cssprop` JSDoc documentation for all exposed `--swc-action-button-*` custom properties

`action-button.stories.ts` — expanded from Playground/Overview-only to a full Phase 5 scaffold:
- Anatomy (label-only, icon + label)
- Options: Sizes, Quiet, Static colors (with `staticColorsDemo` decorator and `!test` tag)
- States: default, disabled, pending
- Behaviors: Icon only (using raw `html` template with `accessible-label`)
- Accessibility story deferred to Phase 7 (`// TODO` comment)

Migration plan checklist updated; Phase 5 (Render & Style) marked complete in the workstream status table.

## Motivation and context

Phase 5 of the `swc-action-button` S2 migration (Epic SWC-2039). Phases 1–4 (prep, setup, API, accessibility) are complete on upstream branches; this phase adds the S2 visual layer on top of the finalized API surface.

## Related issue(s)

- SWC-2045 (Part of SWC-2039)

## Author's checklist

## Reviewer's checklist

### Manual review test cases

- [ ] _Verify default action button renders with correct S2 visual appearance_
  1. Go to Storybook → Action Button → Overview
  2. Inspect the button: expect `border-radius` from `corner-radius-medium-size-medium` token (not pill), background from `gray-100`, no visible border
  3. Hover: background shifts to `gray-200`
  4. Focus (keyboard): outline ring appears at `focus-indicator-gap` offset

- [ ] _Verify all five sizes render at correct heights_
  1. Go to Action Button → Sizes
  2. Confirm `xs` is visibly smaller than `s`, up through `xl`
  3. Confirm border-radius decreases proportionally with size (XS is tightest curve, XL is largest)

- [ ] _Verify quiet treatment_
  1. Go to Action Button → Quiet
  2. Default quiet button: no background fill at rest; background appears on hover/focus
  3. Quiet disabled: no background fill (transparent)

- [ ] _Verify static colors with decorator background_
  1. Go to Action Button → Static colors
  2. `static-color="white"` renders correctly against dark background
  3. `static-color="black"` renders correctly against light background
  4. Both quiet combinations also render correctly

- [ ] _Verify pending state_
  1. Go to Action Button → States
  2. Pending button: no spinner visible immediately; after ~1 second, spinner appears and label disappears
  3. Button width does not shift when spinner appears (width-locked)
  4. Pending button is focusable and shows focus ring

- [ ] _Verify icon-only layout_
  1. Go to Action Button → Icon only
  2. Button is square (equal inline and block size)
  3. Icon is centered

### Device review

- [ ] Did it pass in Desktop?

  1. Go to Storybook → Action Button → States
  2. Tab to each button in turn (Default, Disabled, Pending)
  3. Expect: focus ring visible on Default and Pending buttons; Disabled button is not reachable via Tab
  4. Press Space/Enter on Default: expect click fires; on Pending: expect click suppressed (no action)
  5. Go to Action Button → Icon only; Tab to the icon-only button; expect `accessible-label` announced as button name

  1. Go to Storybook → Action Button → Overview with VoiceOver (macOS) or NVDA (Windows)
  2. Navigate to the button: expect announced as "Edit, button" (role and name from slot text)
  3. Go to States → Pending button: expect announced as "Edit, busy, button" (pending suffix from `getPendingAccessibleName()`)
  4. Go to Icon only story: expect announced as "Edit, button" (name from `accessible-label`)
  5. Confirm no "button" announcement from the host element — semantics land only on the inner `<button>`

### DIFF
--- a/1st-gen/packages/action-button/src/ActionButton.ts
+++ b/1st-gen/packages/action-button/src/ActionButton.ts
@@ -82,6 +82,11 @@ export class ActionButton extends SizedMixin(ButtonBase, {
   }
   private _emphasized = false;
 
+  /**
+   * Hold affordance is not included in the initial 2nd-gen `swc-action-button`
+   * release. It is planned for a future phase. See the action-button migration
+   * plan for consumer options in the interim.
+   */
   @property({ type: Boolean, reflect: true, attribute: 'hold-affordance' })
   public holdAffordance = false;
 
@@ -303,6 +308,14 @@ export class ActionButton extends SizedMixin(ButtonBase, {
     }
     if (changes.has('holdAffordance')) {
       if (this.holdAffordance) {
+        if (window.__swc?.DEBUG) {
+          window.__swc.warn(
+            this,
+            `The "hold-affordance" attribute on <${this.localName}> is not included in the initial 2nd-gen release and will be addressed in a future phase.`,
+            'https://opensource.adobe.com/spectrum-web-components/components/action-button/',
+            { level: 'deprecation' }
+          );
+        }
         this.addEventListener(
           'pointerdown',
           this.handlePointerdownHoldAffordance
--- a/1st-gen/packages/action-button/src/ActionButton.ts
+++ b/1st-gen/packages/action-button/src/ActionButton.ts
@@ -308,14 +308,6 @@ export class ActionButton extends SizedMixin(ButtonBase, {
     }
     if (changes.has('holdAffordance')) {
       if (this.holdAffordance) {
-        if (window.__swc?.DEBUG) {
-          window.__swc.warn(
-            this,
-            `The "hold-affordance" attribute on <${this.localName}> is not included in the initial 2nd-gen release and will be addressed in a future phase.`,
-            'https://opensource.adobe.com/spectrum-web-components/components/action-button/',
-            { level: 'deprecation' }
-          );
-        }
         this.addEventListener(
           'pointerdown',
           this.handlePointerdownHoldAffordance
--- a/2nd-gen/packages/swc/.storybook/guides/customization/global-elements.mdx
+++ b/2nd-gen/packages/swc/.storybook/guides/customization/global-elements.mdx
@@ -171,8 +171,7 @@ Icons are supported with the class `swc-ActionButton-icon`. Add `swc-ActionButto
 
 {/* prettier-ignore */}
 <button type="button" class="swc-ActionButton swc-ActionButton--hasIcon">
-  <svg class="swc-ActionButton-icon" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 36 36" aria-hidden="true" focusable="false"><path d="M33.567 8.2 27.8 2.432a1.215 1.215 0 0 0-.866-.353H26.9a1.371 1.371 0 0 0-.927.406L5.084 23.372a.99.99 0 0 0-.251.422L2.055 33.1c-.114.377.459.851.783.851a.251.251 0 0 0 .062-.007c.276-.063 7.866-2.344 9.311-2.778a.972.972 0 0 0 .414-.249l20.888-20.889a1.372 1.372 0 0 0 .4-.883 1.221 1.221 0 0 0-.346-.945ZM11.4 29.316c-2.161.649-4.862 1.465-6.729 2.022l2.009-6.73Z"/></svg>
-  Edit
+  <svg class="swc-ActionButton-icon" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 36 36" aria-hidden="true" focusable="false"><path d="M33.567 8.2 27.8 2.432a1.215 1.215 0 0 0-.866-.353H26.9a1.371 1.371 0 0 0-.927.406L5.084 23.372a.99.99 0 0 0-.251.422L2.055 33.1c-.114.377.459.851.783.851a.251.251 0 0 0 .062-.007c.276-.063 7.866-2.344 9.311-2.778a.972.972 0 0 0 .414-.249l20.888-20.889a1.372 1.372 0 0 0 .4-.883 1.221 1.221 0 0 0-.346-.945ZM11.4 29.316c-2.161.649-4.862 1.465-6.729 2.022l2.009-6.73Z"/></svg> Edit
 </button>
 {/* prettier-ignore */}
 <button type="button" class="swc-ActionButton swc-ActionButton--iconOnly" aria-label="Edit">
--- a/2nd-gen/packages/swc/.storybook/guides/customization/global-elements.mdx
+++ b/2nd-gen/packages/swc/.storybook/guides/customization/global-elements.mdx
@@ -171,7 +171,8 @@ Icons are supported with the class `swc-ActionButton-icon`. Add `swc-ActionButto
 
 {/* prettier-ignore */}
 <button type="button" class="swc-ActionButton swc-ActionButton--hasIcon">
-  <svg class="swc-ActionButton-icon" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 36 36" aria-hidden="true" focusable="false"><path d="M33.567 8.2 27.8 2.432a1.215 1.215 0 0 0-.866-.353H26.9a1.371 1.371 0 0 0-.927.406L5.084 23.372a.99.99 0 0 0-.251.422L2.055 33.1c-.114.377.459.851.783.851a.251.251 0 0 0 .062-.007c.276-.063 7.866-2.344 9.311-2.778a.972.972 0 0 0 .414-.249l20.888-20.889a1.372 1.372 0 0 0 .4-.883 1.221 1.221 0 0 0-.346-.945ZM11.4 29.316c-2.161.649-4.862 1.465-6.729 2.022l2.009-6.73Z"/></svg> Edit
+  <svg class="swc-ActionButton-icon" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 36 36" aria-hidden="true" focusable="false"><path d="M33.567 8.2 27.8 2.432a1.215 1.215 0 0 0-.866-.353H26.9a1.371 1.371 0 0 0-.927.406L5.084 23.372a.99.99 0 0 0-.251.422L2.055 33.1c-.114.377.459.851.783.851a.251.251 0 0 0 .062-.007c.276-.063 7.866-2.344 9.311-2.778a.972.972 0 0 0 .414-.249l20.888-20.889a1.372 1.372 0 0 0 .4-.883 1.221 1.221 0 0 0-.346-.945ZM11.4 29.316c-2.161.649-4.862 1.465-6.729 2.022l2.009-6.73Z"/></svg>
+  Edit
 </button>
 {/* prettier-ignore */}
 <button type="button" class="swc-ActionButton swc-ActionButton--iconOnly" aria-label="Edit">
--- a/2nd-gen/packages/swc/.storybook/guides/customization/global-elements.mdx
+++ b/2nd-gen/packages/swc/.storybook/guides/customization/global-elements.mdx
@@ -171,7 +171,7 @@ Icons are supported with the class `swc-ActionButton-icon`. Add `swc-ActionButto
 
 {/* prettier-ignore */}
 <button type="button" class="swc-ActionButton swc-ActionButton--hasIcon">
-  <svg class="swc-ActionButton-icon" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 36 36" aria-hidden="true" focusable="false"><path d="M33.567 8.2 27.8 2.432a1.215 1.215 0 0 0-.866-.353H26.9a1.371 1.371 0 0 0-.927.406L5.084 23.372a.99.99 0 0 0-.251.422L2.055 33.1c-.114.377.459.851.783.851a.251.251 0 0 0 .062-.007c.276-.063 7.866-2.344 9.311-2.778a.972.972 0 0 0 .414-.249l20.888-20.889a1.372 1.372 0 0 0 .4-.883 1.221 1.221 0 0 0-.346-.945ZM11.4 29.316c-2.161.649-4.862 1.465-6.729 2.022l2.009-6.73Z"/></svg> Edit
+  <svg class="swc-ActionButton-icon" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 36 36" aria-hidden="true" focusable="false"><path d="M33.567 8.2 27.8 2.432a1.215 1.215 0 0 0-.866-.353H26.9a1.371 1.371 0 0 0-.927.406L5.084 23.372a.99.99 0 0 0-.251.422L2.055 33.1c-.114.377.459.851.783.851a.251.251 0 0 0 .062-.007c.276-.063 7.866-2.344 9.311-2.778a.972.972 0 0 0 .414-.249l20.888-20.889a1.372 1.372 0 0 0 .4-.883 1.221 1.221 0 0 0-.346-.945ZM11.4 29.316c-2.161.649-4.862 1.465-6.729 2.022l2.009-6.73Z"/></svg><span class="swc-ActionButton-label">Edit</span>
 </button>
 {/* prettier-ignore */}
 <button type="button" class="swc-ActionButton swc-ActionButton--iconOnly" aria-label="Edit">
@@ -183,7 +183,7 @@ Icons are supported with the class `swc-ActionButton-icon`. Add `swc-ActionButto
   <svg class="swc-ActionButton-icon" aria-hidden="true" focusable="false">
     <!-- ... -->
   </svg>
-  Edit
+  <span class="swc-ActionButton-label">Edit</span>
 </button>
 
 <button
--- a/2nd-gen/packages/swc/components/action-button/action-button.css
+++ b/2nd-gen/packages/swc/components/action-button/action-button.css
@@ -49,7 +49,7 @@
   overflow: visible;
   cursor: default;
   user-select: none;
-  outline: 0 solid transparent;
+  outline: token("focus-indicator-thickness") solid transparent;
   outline-offset: token("focus-indicator-gap");
   transition-timing-function: token("animation-ease-in-out");
   transition-duration: token("animation-duration-100");
@@ -68,7 +68,6 @@
   color: var(--swc-action-button-content-color-focus, token("neutral-content-color-key-focus"));
   background-color: var(--swc-action-button-background-color-focus, token("gray-200"));
   border-color: var(--swc-action-button-border-color-focus, transparent);
-  outline-width: token("focus-indicator-thickness");
   outline-color: var(--swc-action-button-focus-indicator-color, token("focus-indicator-color"));
 }
 
@@ -229,10 +228,10 @@ slot[name="icon"]::slotted(*) {
 
 /* @global-exclude: pending state requires JS runtime (ButtonBase.pendingActive) */
 
-/* Disabled colors, label/icon fade, and spinner appearance are deferred to
-   .swc-ActionButton--pendingActive, which is added by ButtonBase after a 1-second
-   delay so the button does not flash to its unavailable appearance for operations
-   that complete quickly. */
+/* Cursor changes immediately on [pending]. Disabled colors, label/icon fade,
+   and spinner appearance are deferred to .swc-ActionButton--pendingActive, which is
+   added by ButtonBase after a 1-second delay so the button does not flash to
+   its unavailable appearance for operations that complete quickly. */
 .swc-ActionButton--pendingActive {
   --swc-action-button-down-state-transform: none;
 
@@ -272,6 +271,12 @@ slot[name="icon"]::slotted(*) {
 }
 
 /* ── Forced colors ────────────────────────────────── */
+
+@media (forced-colors: active) {
+  .swc-ActionButton:focus-visible {
+    outline-color: ButtonText;
+  }
+}
 
 /* @global-exclude: pending state requires JS runtime (ButtonBase.pendingActive) */
 @media (forced-colors: active) {
--- a/2nd-gen/packages/swc/components/action-button/action-button.css
+++ b/2nd-gen/packages/swc/components/action-button/action-button.css
@@ -100,9 +100,10 @@
 }
 
 slot[name="icon"]::slotted(*),
-.swc-ActionButton-pendingSpinner {
+.swc-PendingSpinner {
   --_swc-action-button-icon-size: var(--swc-action-button-icon-size, token("workflow-icon-size-100"));
   --_swc-action-button-icon-block-size: var(--swc-action-button-icon-block-size, var(--_swc-action-button-icon-size));
+  --_swc-pending-spinner-size: var(--_swc-action-button-icon-size);
 
   flex-shrink: 0;
   align-self: center;
@@ -225,30 +226,6 @@ slot[name="icon"]::slotted(*) {
 
 /* ── Pending ──────────────────────────────────────── */
 
-/* @global-exclude: pending spinner animations require JS runtime */
-@keyframes swc-pending-spinner-rotate {
-  0% {
-    transform: rotate(var(--swc-pending-spinner-rotate-start, -90deg));
-  }
-
-  100% {
-    transform: rotate(var(--swc-pending-spinner-rotate-end, 270deg));
-  }
-}
-
-@keyframes swc-pending-spinner-dashoffset {
-  0%,
-  100% {
-    stroke-dashoffset: 75px;
-  }
-
-  30% {
-    stroke-dashoffset: var(--swc-pending-spinner-dashoffset-30, 20px);
-  }
-}
-
-/* @global-exclude-end */
-
 /* @global-exclude: pending state requires JS runtime (ButtonBase.pendingActive) */
 
 /* Cursor changes immediately on [pending]. Disabled colors, label/icon fade,
@@ -287,53 +264,6 @@ slot[name="icon"]::slotted(*) {
   display: none;
 }
 
-.swc-ActionButton-pendingSpinner {
-  --_swc-pending-spinner-track-border-color: token("track-color");
-  --_swc-pending-spinner-fill-border-color: token("accent-content-color-default");
-  --_swc-pending-spinner-thickness: 2px;
-
-  display: none;
-}
-
-.swc-ActionButton--pendingActive .swc-ActionButton-pendingSpinner {
-  display: inline-block;
-}
-
-:host([static-color="white"]) .swc-ActionButton-pendingSpinner {
-  --_swc-pending-spinner-track-border-color: token("static-white-track-color");
-  --_swc-pending-spinner-fill-border-color: token("static-white-track-indicator-color");
-}
-
-:host([static-color="black"]) .swc-ActionButton-pendingSpinner {
-  --_swc-pending-spinner-track-border-color: token("static-black-track-color");
-  --_swc-pending-spinner-fill-border-color: token("static-black-track-indicator-color");
-}
-
-.swc-ActionButton-pendingSpinner-track,
-.swc-ActionButton-pendingSpinner-fill {
-  inline-size: var(--_swc-action-button-icon-size);
-  block-size: var(--_swc-action-button-icon-size);
-}
-
-.swc-ActionButton-pendingSpinner-track {
-  stroke: var(--_swc-pending-spinner-track-border-color);
-  stroke-width: var(--_swc-pending-spinner-thickness);
-}
-
-.swc-ActionButton-pendingSpinner-fill {
-  stroke: var(--_swc-pending-spinner-fill-border-color);
-  stroke-width: var(--_swc-pending-spinner-thickness);
-  transform: rotate(-90deg);
-  transform-origin: center;
-}
-
-.swc-ActionButton--pendingActive .swc-ActionButton-pendingSpinner-fill {
-  animation:
-    swc-pending-spinner-rotate 1s cubic-bezier(0.6, 0.1, 0.3, 0.9) infinite,
-    swc-pending-spinner-dashoffset 1s cubic-bezier(0.25, 0.1, 0.25, 1.3) infinite;
-  will-change: transform;
-}
-
 /* @global-exclude-end */
 
 /* ── Reduced motion ───────────────────────────────── */
@@ -342,18 +272,6 @@ slot[name="icon"]::slotted(*) {
   .swc-ActionButton {
     transition-duration: 0ms;
   }
-
-  /* @global-exclude: pending spinner reduced-motion override requires JS runtime */
-  .swc-ActionButton--pendingActive .swc-ActionButton-pendingSpinner-fill {
-    --swc-pending-spinner-dashoffset-30: 0;
-    --swc-pending-spinner-rotate-start: 0deg;
-    --swc-pending-spinner-rotate-end: 360deg;
-
-    animation-duration: 15s;
-    animation-timing-function: linear, linear;
-  }
-
-  /* @global-exclude-end */
 }
 
 /* ── Forced colors ────────────────────────────────── */
@@ -379,21 +297,6 @@ slot[name="icon"]::slotted(*) {
     --swc-action-button-content-color-hover: GrayText;
     --swc-action-button-content-color-down: GrayText;
     --swc-action-button-content-color-focus: GrayText;
-  }
-
-  /* SVG stroke values are not auto-remapped by forced-colors. */
-  .swc-ActionButton-pendingSpinner-track {
-    @media (prefers-color-scheme: dark) {
-      --_swc-pending-spinner-track-border-color: token("static-white-track-color");
-    }
-
-    @media (prefers-color-scheme: light) {
-      --_swc-pending-spinner-track-border-color: token("static-black-track-color");
-    }
-  }
-
-  .swc-ActionButton-pendingSpinner-fill {
-    --_swc-pending-spinner-fill-border-color: Highlight;
   }
 }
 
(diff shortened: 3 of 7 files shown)


2nd-gen/packages/swc/components/action-button/action-button.css
```diff
@@ -0,0 +1,390 @@
...
+  font-weight: token("bold-font-weight");
+  line-height: round(down, calc(token("line-height-100") * 1em), 1px);
+  color: var(--swc-action-button-content-color-default, token("neutral-content-color-default"));
+  text-transform: none;
+  text-decoration: none;
+  background-color: var(--swc-action-button-background-color-default, token("gray-100"));
+  border: var(--_swc-action-button-border-width) solid var(--swc-action-button-border-color-default, transparent);
+  border-radius: var(--swc-action-button-border-radius, token("corner-radius-medium-size-medium"));
+  overflow: visible;
+  cursor: pointer;
```
aramos-adobe: Should be `cursor: default` similar to button. Links explicitly use `pointer`

2nd-gen/packages/swc/components/action-button/action-button.css
```diff
@@ -0,0 +1,390 @@
...
+
+.swc-ActionButton:focus-visible {
+  color: var(--swc-action-button-content-color-focus, token("neutral-content-color-key-focus"));
+  background-color: var(--swc-action-button-background-color-focus, token("gray-200"));
+  border-color: var(--swc-action-button-border-color-focus, transparent);
+  outline: token("focus-indicator-thickness") solid var(--swc-action-button-focus-indicator-color, token("focus-indicator-color"));
+  outline-offset: token("focus-indicator-gap");
+}
+
+.swc-ActionButton:active {
```
aramos-adobe, line 74: Needs perspective change. Refer to SWC 2nd gen button and RSP action button

2nd-gen/packages/swc/components/action-button/action-button.css
```diff
@@ -0,0 +1,390 @@
...
+
+.swc-ActionButton--hasIcon .swc-ActionButton-label {
+  text-align: start;
+}
+
+slot[name="icon"]::slotted(*),
+.swc-ActionButton-pendingSpinner {
+  --_swc-action-button-icon-size: var(--swc-action-button-icon-size, token("workflow-icon-size-100"));
+  --_swc-action-button-icon-block-size: var(--swc-action-button-icon-block-size, var(--_swc-action-button-icon-size));
+
```
aramos-adobe, line 107:
```css
    --_swc-action-button-visual-top-spacing: token("component-top-text-100");
```

aramos-adobe: XS, S,L,XL need [their own](https://www.figma.com/design/eoZHKJH9a3LJkHYCGt60Vb/S2-Token-specs?node-id=12062-17741&t=SPmqTtqGlZuUlzGW-4)

2nd-gen/packages/swc/components/action-button/action-button.css
```diff
@@ -0,0 +1,390 @@
...
+   added by ButtonBase after a 1-second delay so the button does not flash to
+   its unavailable appearance for operations that complete quickly. */
+:host([pending]) .swc-ActionButton {
+  cursor: default;
+}
+
+.swc-ActionButton--pendingActive {
+  justify-content: center;
+  inline-size: var(--_swc-button-pending-inline-size);
+
```
aramos-adobe, line 241: `align-conent: center`
`flex-wrap: nowrap`

Wondering if it makes sense to also ensure vertical alignment here for the progress circle.

2nd-gen/packages/swc/components/action-button/stories/action-button.stories.ts
```diff
@@ -0,0 +1,207 @@
...
+// ──────────────────────────
+
+export const Sizes: Story = {
+  render: (args) => html`
+    ${ACTION_BUTTON_VALID_SIZES.map((size) =>
+      template({ ...args, size, 'default-slot': sizeLabels[size] })
+    )}
+  `,
+  tags: ['options'],
+  parameters: { flexLayout: 'row-wrap', 'section-order': 1 },
```
rise-erpelding: We may have retired `section-order` now with MDX, this can definitely be addressed later when we do the storybook phase though

2nd-gen/packages/swc/components/action-button/stories/action-button.stories.ts
```diff
@@ -0,0 +1,207 @@
...
+
+const staticColorLabels = {
+  white: 'Static white',
+  black: 'Static black',
+} as const satisfies Record<ActionButtonStaticColor, string>;
+
+const editIconSvg = `<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 36 36" aria-hidden="true" focusable="false"><path d="M33.567 8.2L27.8 2.432a1.215 1.215 0 0 0-1.717 0L23 5.516 30.485 13l3.082-3.083a1.215 1.215 0 0 0 0-1.717zM21.586 7l-3.805 3.805 7.414 7.415 3.805-3.805zM3 29.788V37h7.212L23.414 23.8l-7.414-7.415zM6 32v-1.591l9.914-9.914 1.591 1.591L7.591 32z"/></svg>`;
+
+// ────────────────────
+//    AUTODOCS STORY
```
rise-erpelding: Should the separator match (ie, should it be `PLAYGROUND STORY`)? Looking into this, because I realize my WIP for accordion is also like this, so something must be informing it.

Update: found it, the issue might be with [the skill](https://github.com/adobe/spectrum-web-components/blob/main/.ai/skills/migration-styling/assets/stories-template.md?plain=1#L138) and maybe we should update it.

2nd-gen/packages/swc/components/action-button/stories/action-button.stories.ts
```diff
@@ -0,0 +1,207 @@
...
+  l: 'Large',
+  xl: 'Extra-large',
+} as const satisfies Record<ActionButtonSize, string>;
+
+const staticColorLabels = {
+  white: 'Static white',
+  black: 'Static black',
+} as const satisfies Record<ActionButtonStaticColor, string>;
+
+const editIconSvg = `<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 36 36" aria-hidden="true" focusable="false"><path d="M33.567 8.2L27.8 2.432a1.215 1.215 0 0 0-1.717 0L23 5.516 30.485 13l3.082-3.083a1.215 1.215 0 0 0 0-1.717zM21.586 7l-3.805 3.805 7.414 7.415 3.805-3.805zM3 29.788V37h7.212L23.414 23.8l-7.414-7.415zM6 32v-1.591l9.914-9.914 1.591 1.591L7.591 32z"/></svg>`;
```
rise-erpelding: I don't know what the rules are for icons in gen2 exactly (since I think we're not planning to ship them, feels a bit like the wild west 🤠), I think hardcoding a random svg is fine but this one looks a little broken to me.

Image

2nd-gen/packages/swc/components/action-button/stories/action-button.stories.ts
```diff
@@ -0,0 +1,207 @@
...
+  render: (args) => html`
+    ${template({ ...args, quiet: false, 'default-slot': 'Default' })}
+    ${template({ ...args, quiet: true, 'default-slot': 'Quiet' })}
+  `,
+  tags: ['options'],
+  parameters: { flexLayout: 'row-wrap', 'section-order': 2 },
+};
+
+export const StaticColors: Story = {
+  render: (args) => html`
```
rise-erpelding, line 157: Could be nice to show quiet static colors here (thinking about how button shows primary and secondary variants for static) - another thought that doesn't necessarily need to happen until the storybook PR

2nd-gen/packages/swc/components/action-button/action-button.css
```diff
@@ -0,0 +1,390 @@
...
+  position: relative;
+  gap: var(--swc-action-button-gap, token("text-to-visual-100"));
+  align-items: center;
+  justify-content: center;
+  min-block-size: var(--_swc-action-button-min-block-size);
+  padding-inline: calc(var(--swc-action-button-edge-to-text, token("component-edge-to-text-100")) - var(--_swc-action-button-border-width));
+  margin: 0;
+  font-family: token("sans-serif-font");
+  font-size: var(--swc-action-button-font-size, token("font-size-100"));
+  font-weight: token("bold-font-weight");
```
rise-erpelding: I think this needs a different font weight:


```suggestion
  font-weight: token("medium-font-weight");
```

2nd-gen/packages/swc/components/action-button/stories/action-button.stories.ts
```diff
@@ -0,0 +1,207 @@
...
+        slot="icon"
+        xmlns="http://www.w3.org/2000/svg"
+        viewBox="0 0 36 36"
+        aria-hidden="true"
+        focusable="false"
+      >
+        <path
+          d="M33.567 8.2L27.8 2.432a1.215 1.215 0 0 0-1.717 0L23 5.516 30.485 13l3.082-3.083a1.215 1.215 0 0 0 0-1.717zM21.586 7l-3.805 3.805 7.414 7.415 3.805-3.805zM3 29.788V37h7.212L23.414 23.8l-7.414-7.415zM6 32v-1.591l9.914-9.914 1.591 1.591L7.591 32z"
+        />
+      </svg>
```
rise-erpelding, line 253: Could this use `editIconSvg`? Or could `editIconSvg` be refactored to work with both?

2nd-gen/packages/swc/components/action-button/action-button.css
```diff
@@ -0,0 +1,390 @@
...
+slot[name="icon"]::slotted(*),
+.swc-ActionButton-pendingSpinner {
+  --_swc-action-button-icon-size: var(--swc-action-button-icon-size, token("workflow-icon-size-100"));
+  --_swc-action-button-icon-block-size: var(--swc-action-button-icon-block-size, var(--_swc-action-button-icon-size));
+
+  flex-shrink: 0;
+  align-self: flex-start;
+  inline-size: var(--swc-action-button-icon-inline-size, var(--_swc-action-button-icon-size));
+  block-size: var(--_swc-action-button-icon-block-size);
+  margin-block: calc((var(--_swc-action-button-icon-block-size) - 1lh) / 2 * -1);
```
rise-erpelding: Something here might need an adjustment? Spinner and icon both are misaligned a bit.

Image

Image

rise-erpelding (commented):
Did a first pass but will take a look further tomorrow, most likely!

2nd-gen/packages/swc/components/action-button/stories/action-button.stories.ts
cdransf: We did! I'll drop it proactively so I don't miss it later. ✨

2nd-gen/packages/swc/components/action-button/stories/action-button.stories.ts
cdransf: Updated across the board ✨

2nd-gen/packages/swc/components/action-button/stories/action-button.stories.ts
cdransf: Oops! Tweaked this ✨

2nd-gen/packages/swc/components/action-button/stories/action-button.stories.ts
cdransf: Makes sense! Took a pass at this ✨

2nd-gen/packages/swc/components/action-button/stories/action-button.stories.ts
cdransf: Yep! Updated ✨

2nd-gen/packages/swc/components/action-button/action-button.css
cdransf: Fixed this too! ✨

2nd-gen/packages/swc/components/action-button/action-button.css
cdransf: Right on! Updated ✨

2nd-gen/packages/swc/components/action-button/action-button.css
cdransf: Updated! ✨

2nd-gen/packages/swc/components/action-button/action-button.css
cdransf: Hmm, so the CSS repo has block height + line height specified for S2. Is this a measure in Figma resulting from that rather than a spec? 🤔✨

2nd-gen/packages/swc/components/action-button/action-button.css
cdransf: Updated! I don't think it's strictly necessary but it makes the intent clearer without having to trace it. ✨

2nd-gen/packages/swc/components/action-button/action-button.css
rise-erpelding, line 1: Leaving a reminder here in case I forget to mention somewhere else, we'll want this PR to go into the feature branch also!

2nd-gen/packages/swc/components/action-button/stories/action-button.stories.ts
```diff
@@ -0,0 +1,238 @@
...
+  parameters: { flexLayout: 'row-wrap' },
+};
+
+// ──────────────────────────────
+//    BEHAVIORS STORIES
+// ──────────────────────────────
+
+export const IconOnly: Story = {
+  render: (args) =>
+    template({ ...args, 'accessible-label': 'Edit', 'icon-slot': editIconSvg }),
```
rise-erpelding, line 230: Hmmm, something about switching to `template()` and using `editIconSvg` is keeping the `--iconOnly` class from being applied properly and now there are some layout issues that will need addressing - we can revert this back to the hardcoded svg if that's the easiest way.

Image

2nd-gen/packages/swc/components/action-button/action-button.css
```diff
@@ -0,0 +1,392 @@
...
+  --swc-action-button-focus-indicator-color: token("static-white-focus-indicator-color");
+  --swc-action-button-background-color-default: token("transparent-white-100");
+  --swc-action-button-background-color-hover: token("transparent-white-200");
+  --swc-action-button-background-color-down: token("transparent-white-200");
+  --swc-action-button-background-color-focus: token("transparent-white-200");
+  --swc-action-button-background-color-disabled: token("disabled-static-white-background-color");
+  --swc-action-button-content-color-default: token("white");
+  --swc-action-button-content-color-hover: token("white");
+  --swc-action-button-content-color-down: token("white");
+  --swc-action-button-content-color-focus: token("white");
```
rise-erpelding: For `token("white")` and `token("black")` instances in the static color variant selectors, I'm seeing `transparent-white/800` and `transparent-black/800` in Figma - is it possible that `token("white")` and `token("black")` are legacy carryover, or is this a deliberate decision we made to not use the transparent colors?

rise-erpelding (commented):
Looks really close, just minor feedback! ✨

2nd-gen/packages/swc/components/action-button/stories/action-button.stories.ts
cdransf: So, as best I can tell, this is happening because Lit inserts a comment marker in the light DOM as a placeholder for the dynamic part. The comment has `textContent` so the mixin counts it as label content and sets `slotHasContent = true` which makes `hasLabel = true` and blocks `--iconOnly`.

The most durable fix is a one-liner in `observe-slot-text.ts`: 

```diff
108 -          return node.textContent ? node.textContent.trim() : false;
108 +          return node.nodeType === Node.TEXT_NODE && node.textContent
109 +            ? node.textContent.trim()
110 +            : false;
```

(I can also revert this in favor of using `unsafeStatic` scoped to the action button, but this should address similar issues going forward). ✨

2nd-gen/packages/swc/components/action-button/action-button.css
cdransf: Ah! Looks like they are. Just updated these ✨

2nd-gen/packages/swc/components/action-button/action-button.css
cdransf: Oops! Fixed this one and the others ✨

2nd-gen/packages/swc/components/action-button/stories/action-button.stories.ts
```diff
@@ -61,3 +101,143 @@ export const Overview: Story = {
...
+export const States: Story = {
+  render: (args) => html`
+    ${template({
+      ...args,
+      'default-slot': 'Default',
+      'icon-slot': editIconSvg,
+    })}
+    ${template({
+      ...args,
+      disabled: true,
```
aramos-adobe, line 209: Disabled state is clickable here. We could just remove pointer-events in CSS for now until button's base form disable clickable events like pending state does. I [created a PR](https://github.com/adobe/spectrum-web-components/pull/6374) to hopefully fix this.

cdransf: Right on! I'll wait for that to merge and rebase it in ✨

2nd-gen/packages/swc/components/action-button/action-button.css
```diff
@@ -10,8 +10,383 @@
...
+  margin: 0;
+  font-family: token("sans-serif-font");
+  font-size: var(--swc-action-button-font-size, token("font-size-100"));
+  font-weight: token("medium-font-weight");
+  line-height: round(down, calc(token("line-height-100") * 1em), 1px);
+  color: var(--swc-action-button-content-color-default, token("neutral-content-color-default"));
+  text-transform: none;
+  text-decoration: none;
+  background-color: var(--swc-action-button-background-color-default, token("gray-100"));
+  border: var(--_swc-action-button-border-width) solid var(--swc-action-button-border-color-default, transparent);
```
rise-erpelding, line 47: Looking through Storybook in Assistiv Labs, I notice that the Default and Quiet action buttons are styled the same in WHCM:

Image

This differs from Spectrum-CSS, which is ok, because we're trying to avoid adding styles for WHCM wherever possible. I want to call it out though because I'm torn on whether quiet border vs. quiet transparent border would be better. I see some pros & cons to this change:

* **Pro**: seems ideal for accessibility to have the WHCM border visible
* **Con**: if we use a quiet action button in another component (like in the accordion actions, in breadcrumbs, calendar, search, tag group), in WHCM it will have a border - although I guess we could override that

This isn't a blocker, and I'm not sure which way is better, but regardless, it's really small and we can come back to it later, just wanted to mention it as I thought of it and get your thoughts 🙂

2nd-gen/packages/swc/components/action-button/action-button.css
```diff
@@ -10,8 +10,383 @@
...
+  transform: var(--swc-action-button-down-state-transform, perspective(var(--_swc-action-button-min-block-size)) translate3d(0, 0, token("component-size-difference-down")));
+  will-change: transform;
+}
+
+/* Inflated specificity to prevent variant override when disabled */
+.swc-ActionButton:disabled:is(*, :hover) {
+  color: var(--swc-action-button-content-color-disabled, token("disabled-content-color"));
+  background-color: var(--swc-action-button-background-color-disabled, token("disabled-background-color"));
+  border-color: var(--swc-action-button-border-color-disabled, transparent);
+  cursor: default;
```
rise-erpelding, line 87: Maybe a `transform: none;` here to prevent disabled from having down state?

rise-erpelding (commented):
I have a few more thoughts now that I've played with this more!

* CSS versions of action button would truncate with ellipses if container inline-size was restricted... this seems like a behavior that might be nice to have here too!
* Left a comment about transform showing up for disabled - we want to prevent that.
* Left a comment about quiet styling for WHCM - this one's pretty trivial and I don't have a strong opinion either way but would love to know your thoughts from an a11y perspective especially!

1st-gen/packages/action-button/src/ActionButton.ts
```diff
@@ -287,7 +287,7 @@ export class ActionButton extends SizedMixin(ButtonBase, {
         if (window.__swc?.DEBUG) {
           window.__swc.warn(
             this,
-            `The "hold-affordance" attribute on <${this.localName}> is deprecated and will be removed in a future release.`,
+            `The "hold-affordance" attribute on <${this.localName}> is not included in the initial 2nd-gen release and will be addressed in a future phase.`,
```
5t3ph: [nit] I'm not sure this is worth broadcasting at this time since we're still considered in Beta so we don't expect consumers to pick this up right away anyway

2nd-gen/packages/swc/components/action-button/action-button.css
cdransf: Hmm, I'm inclined to leave it as is. Quiet's a visual distinction that gets flattened in WHCM. The border helps indicate that it's interactive and hiding that would hide a visual affordance. I noted it in the migration plan though and I figure consuming/enclosing components could override if they really need to? ✨

2nd-gen/packages/swc/components/action-button/action-button.css
```diff
@@ -10,8 +10,383 @@
...
+  background-color: var(--swc-action-button-background-color-hover, token("gray-200"));
+  border-color: var(--swc-action-button-border-color-hover, transparent);
+}
+
+.swc-ActionButton:focus-visible {
+  color: var(--swc-action-button-content-color-focus, token("neutral-content-color-key-focus"));
+  background-color: var(--swc-action-button-background-color-focus, token("gray-200"));
+  border-color: var(--swc-action-button-border-color-focus, transparent);
+  outline: token("focus-indicator-thickness") solid var(--swc-action-button-focus-indicator-color, token("focus-indicator-color"));
+  outline-offset: token("focus-indicator-gap");
```
5t3ph: Something funny is happening where the outline transition is not smooth and jars from a black outline to the blue

2nd-gen/packages/swc/components/action-button/action-button.css
cdransf: Added on line 85 since that should beat the `:active` rule across the board. ✨

2nd-gen/packages/swc/components/action-button/action-button.css
```diff
@@ -10,8 +10,383 @@
...
+  padding-inline: calc(var(--swc-action-button-edge-to-visual-only, token("component-edge-to-visual-only-100")) - var(--_swc-action-button-border-width));
+}
+
+.swc-ActionButton--iconOnly slot[name="icon"]::slotted(*) {
+  --swc-action-button-edge-to-visual: 0;
+}
+
+/* ── Pending ──────────────────────────────────────── */
+
+/* @global-exclude: pending spinner animations require JS runtime */
```
5t3ph: Can you please add action button to the `globalElementCSS` config in vite.config.ts to have that file generated? Or do you want to do that in a separate pass? It will also need examples added to the [global elements docs](https://spectrum-web-components.adobe.com/?path=/docs/guides-customization-global-element-styling--docs), and cross-referenced for the link use case once that is documented like Button.

2nd-gen/packages/swc/components/action-button/action-button.css
```diff
@@ -10,8 +10,383 @@
...
+  --swc-action-button-content-color-down: var(--swc-action-button-content-color-disabled, token("disabled-content-color"));
+  --swc-action-button-content-color-focus: var(--swc-action-button-content-color-disabled, token("disabled-content-color"));
+}
+
+.swc-ActionButton--pendingActive .swc-ActionButton-label,
+.swc-ActionButton--pendingActive slot[name="icon"]::slotted(*) {
+  display: none;
+}
+
+.swc-ActionButton-pendingSpinner {
```
5t3ph: For all the styles related to the `pendingSpinner` I was hoping we could have a shared stylesheet which could be attached to like `swc-PendingSpinner` and imported for Button and ActionButton. The custom properties are already isolated/shared. It should be all inclusive with static colors and the prefers-reduced-motion update - anything targeting the spinner and its parts.

The Meter work isn't merged yet but is going to be setting up shared stylesheets, so you could borrow that pattern to place it in `2nd-gen/packages/swc/stylesheets/shared/pending-spinner.css` and then you'll need [these small vite config updates](https://github.com/adobe/spectrum-web-components/pull/6316/changes#diff-5a989c6d17a29c9b31232f307e6082fc1edd21a82ef80279fc762667dbc37e04) until it gets to main.

5t3ph (commented):
Let me know if you have questions, or want to do any of these updates in a different pass.

1st-gen/packages/action-button/src/ActionButton.ts
cdransf: Good call! I'll go ahead and remove it. ✨

2nd-gen/packages/swc/components/action-button/action-button.css
cdransf: As best I can tell this was happening when transitioning from no outline to an outline. I updated it to transition from a transparent outline and transition the outline property color. ✨

2nd-gen/packages/swc/components/action-button/action-button.css
cdransf: Ah! Updated to account for this. ✨

2nd-gen/packages/swc/components/action-button/action-button.css
cdransf: Got it! I went ahead and took a pass at this when you have the chance to take a look. ✨

2nd-gen/packages/swc/components/action-button/action-button.css
rise-erpelding: For sure! I think noting in the migration plan is an excellent idea.

2nd-gen/packages/swc/.storybook/guides/customization/global-elements.mdx
```diff
@@ -136,3 +136,64 @@ Icon-only should include the class `swc-Button--iconOnly` on the base element, a
...
+</button>
+```
+
+Icons are supported with the class `swc-ActionButton-icon`. Add `swc-ActionButton--hasIcon` when combining an icon with a label, or `swc-ActionButton--iconOnly` with `aria-label` for icon-only variants.
+
+{/* prettier-ignore */}
+<button type="button" class="swc-ActionButton swc-ActionButton--hasIcon">
+  <svg class="swc-ActionButton-icon" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 36 36" aria-hidden="true" focusable="false"><path d="M33.567 8.2 27.8 2.432a1.215 1.215 0 0 0-.866-.353H26.9a1.371 1.371 0 0 0-.927.406L5.084 23.372a.99.99 0 0 0-.251.422L2.055 33.1c-.114.377.459.851.783.851a.251.251 0 0 0 .062-.007c.276-.063 7.866-2.344 9.311-2.778a.972.972 0 0 0 .414-.249l20.888-20.889a1.372 1.372 0 0 0 .4-.883 1.221 1.221 0 0 0-.346-.945ZM11.4 29.316c-2.161.649-4.862 1.465-6.729 2.022l2.009-6.73Z"/></svg>
+  Edit
+</button>
```
5t3ph, line 175: Resolve markdown forcing in paragraph and breaking the appearance:

```suggestion
<button type="button" class="swc-ActionButton swc-ActionButton--hasIcon">
  <svg class="swc-ActionButton-icon" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 36 36" aria-hidden="true" focusable="false"><path d="M33.567 8.2 27.8 2.432a1.215 1.215 0 0 0-.866-.353H26.9a1.371 1.371 0 0 0-.927.406L5.084 23.372a.99.99 0 0 0-.251.422L2.055 33.1c-.114.377.459.851.783.851a.251.251 0 0 0 .062-.007c.276-.063 7.866-2.344 9.311-2.778a.972.972 0 0 0 .414-.249l20.888-20.889a1.372 1.372 0 0 0 .4-.883 1.221 1.221 0 0 0-.346-.945ZM11.4 29.316c-2.161.649-4.862 1.465-6.729 2.022l2.009-6.73Z"/></svg> Edit
</button>
```

2nd-gen/packages/swc/components/action-button/action-button.css
```diff
@@ -10,8 +10,294 @@
...
+
+/* @global-exclude: pending state requires JS runtime (ButtonBase.pendingActive) */
+
+/* Cursor changes immediately on [pending]. Disabled colors, label/icon fade,
+   and spinner appearance are deferred to .swc-ActionButton--pendingActive, which is
+   added by ButtonBase after a 1-second delay so the button does not flash to
+   its unavailable appearance for operations that complete quickly. */
+:host([pending]) .swc-ActionButton {
+  cursor: default;
+}
```
5t3ph: Not necessary since the base cursor is also default, right?

2nd-gen/packages/swc/components/action-button/action-button.css
```diff
@@ -10,8 +10,294 @@
...
+
+@media (prefers-reduced-motion: reduce) {
+  .swc-ActionButton {
+    transition-duration: 0ms;
+  }
+}
+
+/* ── Forced colors ────────────────────────────────── */
+
+@media (forced-colors: active) {
```
5t3ph, line 275: Is this override necessary?

2nd-gen/packages/swc/components/action-button/action-button.css
```diff
@@ -10,8 +10,294 @@
...
+  color: var(--swc-action-button-content-color-default, token("neutral-content-color-default"));
+  text-transform: none;
+  text-decoration: none;
+  background-color: var(--swc-action-button-background-color-default, token("gray-100"));
+  border: var(--_swc-action-button-border-width) solid var(--swc-action-button-border-color-default, transparent);
+  border-radius: var(--swc-action-button-border-radius, token("corner-radius-medium-size-medium"));
+  overflow: visible;
+  cursor: default;
+  user-select: none;
+  outline: token("focus-indicator-thickness") solid transparent;
```
5t3ph, line 52: I understand how this might be fixing this for the transition, but it's also created a permanent double-border appearance in WHCM (true for regular Button also):

Image

[Testing in a quick CodePen](https://codepen.io/spectrum-css/pen/jEybvPW), looks like this type of division will resolve both issues (transition color, appearance in WHCM):

```css
button {
	transition: outline-color 300ms ease-in;
	outline-offset: 2px;
	outline: 0px solid transparent;
}

button:focus-visible {
	outline-width: 2px;
	outline-color: blue;
}
```

5t3ph (commented):
Pending and global stylesheets look good!

2nd-gen/packages/swc/components/action-button/action-button.css
cdransf: Oops! Removed ✨

2nd-gen/packages/swc/components/action-button/action-button.css
cdransf: I trimmed this down but kept the pending state override that's consistent with button. ✨

2nd-gen/packages/swc/components/action-button/action-button.css
cdransf: Ah, thank you! ✨

2nd-gen/packages/swc/.storybook/guides/customization/global-elements.mdx
5t3ph: I still see the issue:

Image

I think the reason is it still needs the span around the label like regular Button, right? And that would apply to each Action Button demo/code snippet.

5t3ph (commented):
Realized there is a missing span for the global demos (if the stylesheet more or less matches Button patterns, which it ideally does)

rise-erpelding (approved):
This looks good to me! The only thing I've noticed is that we don't do anything to handle longer label text, I'm not sure if we'd want to leave an option for wrapping or truncation now, or whether it would be better to defer that?

cdransf:
> This looks good to me! The only thing I've noticed is that we don't do anything to handle longer label text, I'm not sure if we'd want to leave an option for wrapping or truncation now, or whether it would be better to defer that?

I think we can defer this to a follow up since the examples use concise labels and 1st gen doesn't address it. We can always follow up and use the same pattern button does. ✨

5t3ph (approved):
Messaging you about one thing that can be separate, but otherwise looks good! 🙌

PR #6352, opened 2026-05-28
docs(swc): retire story-level JSDoc, description-only tag, and section-order parameter

caseyisonit:
## Description

Final cleanup in the JSDoc-to-MDX conversion chain. Now that components, patterns, and controllers all author their docs page in per-unit MDX (PRs [#6349](https://github.com/adobe/spectrum-web-components/pull/6349) and [#6351](https://github.com/adobe/spectrum-web-components/pull/6351)), retire the workarounds that JSDoc-on-stories required, and rewrite the authoring rules and migration skills so future contributors learn the new pattern.

This branch is based on `caseyisonit/docs-mdx-patterns` (SWC-2230). Merge that PR first; this PR depends on it.

Scope:

- **Runtime simplification**: `SpectrumStories.tsx` drops the `description-only` branch and the `section-order` sort. `validate-story-tags.js` removes `description-only` from the allowed-tags allowlist.
- **Stories-file guidance** (`.ai/rules/stories-format.md`, contributor-docs Phase 4 step, `migration-styling` skill + template): the stories file is definitions-only. No story-level JSDoc above story exports. `Playground` uses `tags: ['dev']` (no `'autodocs'`) when a per-unit MDX exists. Genre coverage extended to patterns and controllers.
- **Per-unit MDX authoring guide** (`.ai/rules/stories-documentation.md`): reframed from "what to put in JSDoc above stories" to "what to author in the per-unit MDX". New sections cover the per-unit MDX file template, canonical section order, genre-specific table (MDX path, blocks import path, ApiTable branch for `DocsFooter`, production-build inclusion), per-story title rules (PascalCase → Title Case, `storyName` override), single-story-section guidance, and untagged-stories-do-not-render warning.
- **Code conformance** (`.ai/rules/code-conformance.md`): storybook checks now cover both surfaces. Stories file: no story-level JSDoc, no `section-order`, no `description-only`, correct Playground tags. Per-unit MDX: file exists, `<Meta of={Stories} />` declared once, `<DocsHeader />`/`<DocsFooter />` placement, canonical section order, every section-tagged story surfaced via `<Canvas>`, controller-API hand-authoring branch.
- **Migration documentation skill** (`migration-documentation/SKILL.md`): Phase 7 now authors the per-component MDX docs page + public-API JSDoc on `Component.ts`. The Upcoming features pattern moves to MDX prose-only sections.
[...]

PR #6351, opened 2026-05-28
docs(swc): convert 2nd-gen pattern and controller story JSDoc to per-unit MDX

caseyisonit:
## Description

Follow-up to **PR #6349** (components). Move long-form documentation that previously lived as JSDoc comments above story exports in `.stories.ts` files into per-unit MDX files at each unit's root, for the remaining 2nd-gen genres: conversational AI patterns and the focusgroup navigation controller.

Scope (this PR): **11 patterns + 1 controller.**

- **Patterns** (`2nd-gen/packages/swc/patterns/conversational-ai/<name>/<name>.mdx`):
  - system-message, suggestion-item, conversation-turn, message-sources, upload-artifact, user-message, suggestion-group, message-feedback, response-status, prompt-field, conversation-thread
- **Controller** (`2nd-gen/packages/core/controllers/focusgroup-navigation-controller/focusgroup-navigation-controller.mdx`):
  - Validates the controller branch of `DocsFooter`, which auto-omits `<ApiTable />` (controllers expose a TypeScript class, not a custom element manifest) and renders a hand-authored `## API` section instead.

This branch is based on `caseyisonit/docs-to-mdx`. Merge that PR first; this PR contains pattern + controller changes only.

## Motivation and context

Same as PR #6349: story-level JSDoc was the only mechanism for long-form prose attached to each story, which forced documentation-only stories (`tags: ['description-only']`) and `section-order` parameters to influence rendering order. Per-unit MDX provides:

- A single, hand-authored source of truth per pattern/controller, with explicit `<Canvas of={Story} />` references.
- Prose-only sections without fake stories: description-only stories are deleted; their prose lives in MDX as Canvas-less sections (e.g. the controller's `Usage`, `API`, `Accessibility`, `Appendix` prose).
- Exact reuse of the `DocsHeader` / `DocsFooter` blocks shipped in PR #6349 — no new infrastructure.

## Related issue(s)

- Follows PR #6349 (SWC-2228); tracked internally under the same umbrella effort.
- SWC-2230

## Screenshots (if appropriate)

Visual parity verified by spot-checking `system-message`, `conversation-thread`, and the `focusgroup-navigation-controller` Docs pages against the template-driven rendering.

## Author's checklist

[...]

PR #6349, opened 2026-05-28
docs(swc): convert 2nd-gen component story JSDoc to per-component MDX

caseyisonit:
## Description

Move long-form documentation that previously lived as JSDoc comments above story exports in `.stories.ts` files into per-component MDX files at the component root. Each MDX file uses Storybook doc blocks (`<Meta of={...} />`, `<Canvas of={Story} />`) to render stories explicitly, and becomes the authoritative Docs page for that component, replacing the template-driven page generated by `DocumentTemplate.mdx` on a per-component basis.

Scope (this PR): **9 components + 2 internal stories + shared infrastructure.**

- **Components**: `badge`, `avatar`, `color-loupe`, `divider`, `progress-circle`, `status-light`, `typography`, `illustrated-message`, `button`, `tabs`
- **Internal components**: `asset` (`.internal.mdx`), `icon` (`.internal.mdx`)
- **Shared blocks**: `DocsHeader` and `DocsFooter` (in `2nd-gen/packages/swc/.storybook/blocks/`) own all the standard chrome at the top and bottom of every docs page, so the template-driven page and per-component MDX share the exact same surface with no duplicate boilerplate.
- **Template rewrite**: `DocumentTemplate.mdx` now composes itself from the new blocks; components without a per-component MDX continue to render through it unchanged.
- **Sidebar refresh**: Auto-generated `preview.ts` registrations for several new migration-analysis contributor docs (action-button, button-group, close-button, grid, infield-button, link, popover, changelog strategy) — picked up from `update-nav.js`.

> **Patterns and controllers are deferred** to a follow-up PR (saved on `caseyisonit/docs-mdx-patterns`). This PR is scoped to components only.

## Motivation and context

Story-level JSDoc was the only mechanism for long-form prose attached to each story, which forced documentation-only stories (`tags: ['description-only']`) and `section-order` parameters to influence rendering order. Moving to MDX:

- Eliminates the `description-only` anti-pattern: prose-only sections live in MDX without needing a fake story to hang JSDoc on.
- Removes `section-order` workarounds: ordering is explicit in MDX.
- Enables a single, hand-authored source of truth per component, with explicit `<Canvas of={Story} />` references that survive lint and refactors.
- Sets up patterns and controllers to use the same `DocsHeader` / `DocsFooter` shell (out of scope for this PR but enabled by it).

[...]

PR #6363, opened 2026-06-02
feat(ai): add yarn lint:docs-pages structural lint for per-unit MDX

caseyisonit:
## Description

Adds a structural lint that enforces the per-unit MDX authoring standards from [`.ai/rules/stories-documentation.md`](https://github.com/adobe/spectrum-web-components/blob/main/.ai/rules/stories-documentation.md) across 2nd-gen components, internal components, patterns, and controllers.

For every in-scope `<unit>.mdx` file the lint verifies the four structural invariants spelled out in SWC-2231:

1. **Exactly one** `<Meta of={Stories} />` declaration linking the docs page to its stories module.
2. **Canonical `##` section order**, genre-aware:
   - **Component / pattern / internal**: `Anatomy → Upcoming features → Usage → Options → States → Behaviors → Accessibility → Full pattern → API → Appendix → Feedback`
   - **Controller**: `What it does → Basic usage → Usage → Upcoming features → Options → States → Behaviors → Accessibility → API → Appendix → Feedback` (no `Anatomy` — controllers have no DOM)
   - Unknown section names are flagged with the allowed list for that genre.
3. **Every section-tagged story is surfaced** — for each story tagged `anatomy`, `options`, `states`, `behaviors`, or `a11y`, the MDX must include at least one matching `<Canvas of={Stories.ExportName} />`.
4. **No broken Canvas references** — every `<Canvas of={...} />` points to an export that exists in the stories module.

## Motivation and context

The components / patterns / controller MDX rollout shipped in [#6349](https://github.com/adobe/spectrum-web-components/pull/6349), [#6351](https://github.com/adobe/spectrum-web-components/pull/6351), and [#6352](https://github.com/adobe/spectrum-web-components/pull/6352). Without an enforcement gate the docs surface will silently drift the next time someone adds a component, renames a story, or reorders sections. This lint catches all four problem classes that show up most often during authoring.

The check is path-aware and only inspects per-unit MDX files (`<unit>.mdx`, `<unit>.internal.mdx`). Migration guides, pattern overviews, and other supporting docs are out of scope. Scope is encoded in the file's path under `2nd-gen/packages/swc/components/`, `2nd-gen/packages/swc/patterns/`, or `2nd-gen/packages/core/controllers/`.

Wiring:

- New script: `.ai/scripts/validate-docs-pages.js`
- Standalone command: `yarn lint:docs-pages`
- Bundled into `yarn lint:ai` (and therefore `yarn lint`) so it runs in CI alongside the existing story-tag, AGENTS.md path, config schema, and symlink validators

Docs updates:

- `.ai/README.md` documents both `yarn lint:docs-pages` and the expanded `yarn lint:ai`
- `.ai/rules/stories-documentation.md` adds a "Self-check with `yarn lint:docs-pages`" section so authors know how to validate their MDX during authoring

## Related issue(s)

Resolves SWC-2231. Final ticket in the SWC-2229 epic.

N/A — CI-only change.

## Author's checklist

## Reviewer's checklist

### Manual review test cases

- [ ] **Lint passes against current main**
  1. From the repo root, run `yarn lint:docs-pages`
  2. Expect: `Checked 24 per-unit MDX file(s).` and `All per-unit MDX docs pages pass.`
  3. Also run `yarn lint:ai` and confirm the new `Per-unit MDX docs pages` section reports `✔ passed`.

- [ ] **Lint catches a missing `<Canvas>` for a section-tagged story**
  1. Temporarily comment out `<Canvas of={BadgeStories.Anatomy} />` in `2nd-gen/packages/swc/components/badge/badge.mdx`
  2. Run `yarn lint:docs-pages`
  3. Expect: a clear error pointing at `badge.mdx` and naming the `Anatomy` story
  4. Restore the file and confirm the lint passes again.

- [ ] **Lint catches an unknown section heading**
  1. Temporarily rename `## Anatomy` to `## Bogus` in a converted component MDX
  2. Run `yarn lint:docs-pages`
  3. Expect: `unknown ## section "Bogus" for component (expected one of: ...)`

- [ ] **Lint catches an out-of-order section**
  1. Temporarily swap the order of `## Behaviors` and `## Accessibility` in a converted MDX
  2. Run `yarn lint:docs-pages`
  3. Expect: a clear ordering violation message

- [ ] **Lint catches a broken Canvas reference**
  1. Temporarily change one `<Canvas of={Stories.Anatomy} />` to `<Canvas of={Stories.NotARealStory} />`
  2. Run `yarn lint:docs-pages`
  3. Expect: a clear "references a story that does not exist" error

- [ ] **Genre-aware section list**
  1. Temporarily add `## Anatomy` to the focusgroup-navigation-controller MDX
  2. Run `yarn lint:docs-pages`
  3. Expect: `unknown ## section "Anatomy" for controller (expected one of: ...)` — confirming controllers do not allow `Anatomy`.

### Device review

- [ ] Did it pass in Desktop?

  1. This is a CI-only Node script with no UI. No focusable parts; confirm no regressions in surrounding examples by running `yarn lint` and observing all existing lint targets still pass.

  1. No new user-facing surface. The lint's output is plaintext terminal-only; confirm errors are readable as plain text without ANSI color (e.g. piped to a file).


scripts/validate-docs-pages.js
```diff
@@ -0,0 +1,516 @@
...
+  'Options',
+  'States',
+  'Behaviors',
+  'Accessibility',
+  'API',
+  'Appendix',
+  'Feedback',
+];
+
+const SECTION_TAGS = ['anatomy', 'options', 'states', 'behaviors', 'a11y'];
```
caseyisonit, line 92: **Why two canonical lists?** Controllers don't have a DOM, so `## Anatomy` doesn't apply. They open with `## What it does` and `## Basic usage` instead (per the controller MDX shipped in #6351). Keeping the two lists side by side rather than a single config makes it easy to see what's allowed where without cross-referencing.

**Interpretation of the ticket**: SWC-2231 says "required sections subset of {Anatomy, Options, States, Behaviors, Accessibility}". I'm reading this as "the *allowed* sections are this set" (treating "required" as "permitted to be required"), not as "these sections must all be present" — a docs page can have any subset and unknowns are flagged. Happy to add a "must have at least one" check if that was the intent.

scripts/validate-docs-pages.js
```diff
@@ -0,0 +1,516 @@
...
+  ) {
+    return {
+      genre: 'controller',
+      requireMeta: true,
+      requireCanvasForTaggedStories: true,
+      canonicalSections: CONTROLLER_SECTIONS,
+    };
+  }
+
+  return {
```
caseyisonit, line 167: **Two gaps from SWC-2231 I deliberately left out** — calling out here so reviewers can decide whether either should land before merge:

1. **"Patterns: pattern groups also require a `<group>/pattern-overview.mdx`"** — not checked. `conversational-ai` ships with one; a future group could skip it and this lint would let it through.
2. **"Components: `<ApiTable />` and `<Controls />` expected"** — not checked directly. Both render automatically via `<DocsFooter />`. Cheapest equivalent would be to verify the MDX includes a `<DocsFooter />` reference.

Neither is wired up because the four explicit "For every MDX page, the script verifies:" invariants in the ticket cover the high-value failure modes. Easy to add if you want either.

scripts/validate-docs-pages.js
```diff
@@ -0,0 +1,516 @@
...
+/**
+ * Parse a stories.ts file and return a map of exported story names to their
+ * tag arrays. Only `export const X: Story = ...` and `export const X = ...`
+ * exports with a `tags: [...]` field are considered.
+ *
+ * The parser is regex-based and intentionally tolerant — it does not try to
+ * cover every TypeScript edge case, just the conventions documented in
+ * `.ai/rules/stories-format.md`.
+ */
+function parseStoryTags(storiesPath) {
```
caseyisonit, line 324: **Regex over AST**: considered pulling in `@babel/parser` for the stories file (and `remark` for MDX) but landed on regex for both. Reasoning:

1. We only check four structural invariants — not full schema validation.
2. The conventions in `.ai/rules/stories-format.md` are tight enough that regex covers them without false positives in the current corpus (24 in-scope files; all pass).
3. No extra deps; cold start is fast enough for the pre-commit hook.

If the rules expand (e.g. we want to validate nested MDX components or `<Story>` blocks), revisit with a real parser. Until then this is intentionally minimal.

scripts/validate-docs-pages.js
```diff
@@ -0,0 +1,516 @@
...
+    // Skip the meta object — it uses `const meta: Meta = { ... }` (lowercase).
+    if (name === 'meta') {
+      continue;
+    }
+
+    // Find the matching closing brace by counting depth from match.index.
+    const openIdx = content.indexOf('{', match.index);
+    if (openIdx === -1) {
+      continue;
+    }
```
caseyisonit, line 340: **Manual brace counting**: walks `{ ... }` depth from the matched `export const X: Story = {` to find the matching `}`. Avoids pulling in a TypeScript parser for one structural check. Tolerant by design — if a story body uses an unusual layout that breaks the regex match earlier, we skip that export rather than fail the whole file.

scripts/validate-docs-pages.js
```diff
@@ -0,0 +1,516 @@
...
+  // Check 4: No broken story references
+  for (const ref of canvasRefs) {
+    if (!(ref in stories)) {
+      errors.push(
+        `${relPath}: <Canvas of={${importInfo.binding}.${ref}} /> references a story that does not exist in ${path.relative(repoRoot, storiesPath)}`
+      );
+    }
+  }
+
+  // Check 3: Every section-tagged story is surfaced via <Canvas>
```
caseyisonit, line 468: This is the most useful check in practice — it's what catches the "author added a new story but forgot to reference it in MDX" case. Note the genre exclusion: `internal` MDX skips this check because internal docs are dev-only and intentionally free-form (per ticket: "no section is required"). The Meta + Canvas-reference resolution still runs for internal pages so genuinely broken references are still caught.

caseyisonit (commented):
Self-review for reviewer context. Most of the script is self-documenting; flagging the four spots where intent or trade-offs aren't obvious from the code alone, plus two items from SWC-2231 I deliberately left out.

scripts/validate-docs-pages.js
```diff
@@ -0,0 +1,516 @@
...
+ */
+function countMetaOf(content) {
+  const matches = content.match(/<Meta\s+of=\{[^}]+\}/g);
+  return matches ? matches.length : 0;
+}
+
+/**
+ * Extract all top-level `##` headings (not `###` or deeper) in document order.
+ */
+function extractH2Headings(content) {
```
cdransf, line 255: Will this/should this hit headings in fenced code blocks? ✨

scripts/validate-docs-pages.js
```diff
@@ -0,0 +1,516 @@
+#!/usr/bin/env node
```
cdransf, line 1: Should this live in `.ai/scripts`? ✨

scripts/validate-docs-pages.js
```diff
@@ -0,0 +1,516 @@
...
+}
+
+/**
+ * Extract all top-level `##` headings (not `###` or deeper) in document order.
+ */
+function extractH2Headings(content) {
+  const headings = [];
+  for (const line of content.split('\n')) {
+    const m = line.match(/^##\s+(.+?)\s*$/);
+    if (m && !line.startsWith('###')) {
```
cdransf, line 259: The regex will catch the 2nd condition in the guard ✨ 

```suggestion
    if (m) {
```

scripts/validate-docs-pages.js
```diff
@@ -0,0 +1,516 @@
...
+  // body, then look for a `tags: [ ... ]` inside it.
+  const exportRe =
+    /export\s+const\s+([A-Z][A-Za-z0-9_]*)\s*(?::\s*[A-Za-z<>]+)?\s*=\s*\{/g;
+  let match;
+  while ((match = exportRe.exec(content)) !== null) {
+    const name = match[1];
+    // Skip the meta object — it uses `const meta: Meta = { ... }` (lowercase).
+    if (name === 'meta') {
+      continue;
+    }
```
cdransf:
```suggestion
```

The regex on line 319 only allows uppercase starting identifiers so we can remove this guard. ✨

cdransf (approved):
Looks rad! Just a few comments ✨

scripts/validate-docs-pages.js
caseyisonit: i would imagine if we are using a code block we would use an html heading instead of markdown but let me know if you know of a use case where this may be an issue. I can review and see if i can harden this check or we can address it when/if it becomes an issue in the future.

scripts/validate-docs-pages.js
caseyisonit: no because we have this also running in the pre-commit hook too and its represented in the package.json commands. all .ai/scripts are/should be ran in AI skills and arent a part of our checks. now this is something we could discuss, if we want to leverage those scripts in CI or husky checks.

scripts/validate-docs-pages.js
caseyisonit: so the 2nd condition is needed because `### ` will still match `## ` because thats how it ends, this ensures its an h2 not an h3 in markdown

scripts/validate-docs-pages.js
```diff
@@ -0,0 +1,512 @@
...
+  'API',
+  'Appendix',
+  'Feedback',
+];
+
+const CONTROLLER_SECTIONS = [
+  'What it does',
+  'Basic usage',
+  'Usage',
+  'Upcoming features',
```
miwha-adobe: In `COMPONENT_LIKE_SECTIONS`, 'Upcoming features" comes before `Usage`. Should that be the same in this `CONTROLLER_SECTION`?

miwha-adobe: Additionally, if you want to make the intent obvious, you could do something like:

```
const SHARED_SECTIONS = ['Upcoming features', 'Usage', 'Options', 'States', 'Behaviors', 'Accessibility'];
const SHARED_FOOTER = ['API', 'Appendix', 'Feedback'];

const COMPONENT_LIKE_SECTIONS = ['Anatomy', ...SHARED_SECTIONS 'Full pattern', ...SHARED_FOOTER];
const CONTROLLER_SECTIONS = ['What it does', 'Basic usage', ...SHARED_SECTIONS, ...SHARED_FOOTER];
```

scripts/validate-docs-pages.js
miwha-adobe: Responding to Q1: The script validates some things that feel like they could be enforced upstream by TypeScript,  for example, valid story tag values (tags: `SectionTag[]` with a union type would catch typos like "anatomny" at compile time). 

Are any of these already type-enforced in the stories files? If not, is the script the right place for them, or would tightening the types be cleaner and leave the script to handle only what TypeScript genuinely can't reach (heading order, <Canvas> coverage, <Meta> count)?

Q2: I think its sufficient if it's rendered in `<DocsFooter />`

scripts/validate-docs-pages.js
miwha-adobe: Mild correctness edge case in `parseStoryTags`: the brace-depth counter doesn't account for `{ / }` inside template literals or strings. eg:

```render: (args) => html`<div style="${'{'}">`) ```

If unmatched braces appear before the tags field, the body slice is truncated and the story gets recorded with no tags and check 3 silently passes for that story instead of flagging a missing `<Canvas> `reference. Unlikely to hit in practice but worth noting as a gap.

File: scripts/validate-docs-pages.js (hunk 7 above)
Comment (blunteshwar) [reply scripts/validate-docs-pages.js:1]: I agree

File: scripts/validate-docs-pages.js [hunk 11]
```diff
@@ -0,0 +1,512 @@
...
+    }
+  }
+  return headings;
+}
+
+/**
+ * Extract all `<Canvas of={Stories.ExportName} />` references and return the
+ * referenced story export names.
+ */
+function extractCanvasReferences(content, binding) {
```
blunteshwar, line 275: It will only recognizes exact one-line `<Canvas of={Stories.X} /> `tags. Valid Storybook MDX with extra props or multiline props, already used in `/private/tmp/swc-pr-6363/2nd-gen/packages/swc/patterns/conversational-ai/pattern-overview.mdx` (line 76), would be treated as missing if used in an in-scope per-unit page. This can create false lint failures for normal <Canvas> usage such as sourceState or additionalActions.

blunteshwar (commented):
The Pr looks great. I have left a couple of comments and after that it is good to go

scripts/validate-docs-pages.js
caseyisonit: Deferring this one — the convention so far is to use HTML headings inside code samples (not Markdown), so this shouldn't fire in practice. If we hit a case where Markdown headings inside a fenced block trip the check, hardening to skip lines inside backtick fences is straightforward.

scripts/validate-docs-pages.js
caseyisonit: Good catch — fixed in 2c0f0adcf1. Reordered so `Upcoming features` sits immediately before the API/Appendix/Feedback footer block in both lists. Reads better as a forward-looking note after the current API/behavior, not as a preamble before Options. Existing component MDX files (badge, avatar, button, illustrated-message, tabs) all moved to match.

caseyisonit: Applied — see 2c0f0adcf1. `SHARED_SECTIONS` and `SHARED_FOOTER` constants now compose into both lists. Cleaner than repeating the middle/footer rows verbatim.

scripts/validate-docs-pages.js
caseyisonit: **Q1 (TypeScript for tag values):** `.ai/scripts/validate-story-tags.js` already does this check today — it validates every tag in 2nd-gen `.stories.ts` against a known allowlist. Tightening to a TS union type on the story `tags` field would be cleaner but feels like a separate effort (touches the story type definitions, not just lint). Leaving the validator as-is for now and open to a follow-up if we want to migrate.

**Q2 (DocsFooter sufficient):** Added a `<DocsFooter />` presence check in 2c0f0adcf1. Component, pattern, and controller MDX must include the block; internal MDX stays exempt (dev-only, free-form).

scripts/validate-docs-pages.js
caseyisonit: Real edge case — noted. Two reasons I'm leaving the parser as-is for now: (1) none of the 24 in-scope stories files trigger it today, and (2) if we hit it, the fix is local to `parseStoryTags` rather than a structural change. If it shows up in the wild we can pull in a real TS parser for this one function. Filing as a deferred consideration rather than a blocker.

scripts/validate-docs-pages.js
caseyisonit: Fixed in 2c0f0adcf1. Relaxed the regex to allow attributes between `}` and `/>` and switched to the `s` flag for multi-line matches. Smoke-tested with `<Canvas of={Stories.X} sourceState=\"shown\" />` — now matched.

blunteshwar (approved):
LGTM

miwha-adobe (approved):
Great job!

PR #6376, opened 2026-06-06
feat(tooltip): integrate HoverController and PlacementController

5t3ph:
## Description

Integrates `HoverController` and `PlacementController` into `<swc-tooltip>`, completing the additive phase of the Spectrum 2 tooltip migration (Epic SWC-2017). All seven previously deferred attributes (`delay`, `disabled`, `manual`, `offset`, `cross-offset`, `container-padding`, `should-flip`) are now active API.

### HoverController

- Hover/focus open/close with warm-up delay (default `1500 ms`; `delay="0"` for immediate)
- Keyboard focus always opens immediately regardless of `delay`
- WCAG 1.4.13 pointer bridge: pointer can move from trigger into the tooltip bubble without closing
- `disabled` suppresses hover/focus; `manual` suppresses all automatic wiring
- `_requestedPlacement` / `_placementFromController` pattern preserves the consumer's original requested placement so viewport-driven flips can revert when space is restored

### PlacementController

- Viewport-aware pixel positioning via Floating UI (`translate: Xpx Ypx`)
- Placement flip: `placement` reflects the computed side; existing `[placement]` CSS selectors handle tip direction automatically without new rules
- `tipElement` passed to `PlacementController` so the arrow middleware keeps the tip aligned with the trigger center when `cross-offset` or viewport `shift` displaces the bubble
- `--_swc-tooltip-animation-distance` is now driven by `offset` so the entry animation travel distance matches the gap

### CSS animation

- `@starting-style` direction corrected: bubble animates **from** the trigger outward to its resting position (was opposite)
- Final open state has no CSS `transform` — `PlacementController`'s `translate` owns the resting position
- Exit positioning (`translate`/`top`/`left`) cleared after the exit transition completes, not before, preventing a mid-fade flash at displaced position

### HoverController bug fix

`hostDisconnected()` was cancelling the shared `WarmState` cooldown timer, which could leave `isWarm` stuck at `true` indefinitely when an unrelated instance disconnected during another instance's cooldown. Regression test added.

### Stories

- Removed Floating UI inline-style workaround (`positionTooltip`, `@swc-open` wrapper)
- Added `Disabled`, `Manual`, and `TriggerElement` state/behavior stories
- `TriggerElement` uses `lit/directives/ref` so wiring is active in the Docs page canvas (not just in play mode)

### Tests

- `HoverOpensTest`, `FocusOpensTest`, `DisabledPreventsHoverTest`, `ManualPreventsHoverTest`
- `TriggerElementHoverTest`, `TriggerElementOverridesForHoverTest`
- `PlacementControllerTest`, `ShadowRootScopeTest`
- `DisconnectDoesNotCancelSharedCooldown` (HoverController regression)

## Motivation and context

Closes the additive phase documented in the tooltip migration plan. All automated-testable behavior items in the checklist are now `[x]`. Remaining open items are manual AT verification and process steps (PR, follow-on tickets, sign-off).

## Related issue(s)

- SWC-2210

## Screenshots

https://github.com/user-attachments/assets/7056b81f-1723-4b79-ac4d-ff839669f91e

### Manual review test cases

- [x] Hover over a trigger with a `swc-tooltip` attached via `for` — tooltip opens after ~1.5 s, tip points at trigger
- [x] Move pointer quickly between two adjacent tooltips — second opens immediately (warm state)
- [x] Move pointer from trigger into the tooltip bubble — tooltip stays open (WCAG 1.4.13 bridge)
- [ ] Tab to a trigger — tooltip opens immediately; blur closes immediately
- [x] Set `disabled` on a tooltip — hover and focus no longer open it
- [x] Set `manual` on a tooltip — only `tooltip.open = true` opens it; hover does nothing
- [ ] Resize viewport so the requested placement has no room — tooltip flips to opposite side, tip direction updates, animation direction correct; enlarge again — tooltip returns to requested side
- [ ] Set `cross-offset` — bubble shifts along the trigger edge; tip stays aligned with trigger center

### Device review

- [ ] Did it pass in Desktop?

- [ ] **Keyboard**
  1. Open the Tooltip Storybook story
  2. Press <kbd>Tab</kbd> to focus the trigger button — tooltip should open immediately
  3. Press <kbd>Escape</kbd> — tooltip should close; focus should remain on the trigger
  4. Press <kbd>Tab</kbd> again to blur — tooltip should close
  5. Verify tooltip text is never in the tab order

- [ ] **Screen reader** (NVDA on Windows / VoiceOver on macOS)
  1. Navigate to the trigger button with a screen reader
  2. Verify the trigger's accessible name is announced, followed by the tooltip description text
  3. Navigate to the icon-only `Labeling` story — verify the tooltip text is announced as the button's accessible name (not description)
  4. Verify the tooltip itself is not focusable and has no interactive content

### DIFF
--- a/2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
+++ b/2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
@@ -321,6 +321,8 @@ export abstract class TooltipBase
       this.style.removeProperty('translate');
       this.style.removeProperty('top');
       this.style.removeProperty('left');
+      this.style.removeProperty('--swc-placement-available-width');
+      this.style.removeProperty('--swc-placement-available-height');
 
       if (this.tipElement) {
         this.tipElement.style.removeProperty('translate');
--- a/2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
+++ b/2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
@@ -321,6 +321,12 @@ export abstract class TooltipBase
       this.style.removeProperty('translate');
       this.style.removeProperty('top');
       this.style.removeProperty('left');
+
+      if (this.tipElement) {
+        this.tipElement.style.removeProperty('translate');
+        this.tipElement.style.removeProperty('top');
+        this.tipElement.style.removeProperty('left');
+      }
     }
   }
 
--- a/2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
+++ b/2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
@@ -15,6 +15,10 @@ import { property } from 'lit/decorators.js';
 import { SpectrumElement } from '@spectrum-web-components/core/element/index.js';
 
 import {
+  HoverController,
+  type HoverControllerHost,
+} from '../../controllers/hover-controller/index.js';
+import {
   TOOLTIP_PLACEMENTS,
   TOOLTIP_VARIANTS,
   type TooltipPlacement,
@@ -23,18 +27,15 @@ import {
 
 /**
  * Abstract base class for the Tooltip component.
- *
- * Declares all public properties, sets `role="tooltip"` and `popover="auto"` on
- * the host element, wires the popover lifecycle (`beforetoggle`, `toggle`,
- * `transitionend`), dispatches `swc-open`, `swc-close`, `swc-after-open`, and
- * `swc-after-close` events, resolves the trigger element via `for` or
- * `triggerElement`, and syncs the ARIA relationship (`ariaDescribedByElements`
- * or `ariaLabelledByElements` when `labeling` is set) on `open` and `labeling`
- * changes. No rendering logic, including placement.
+ * Handles all non-rendering logic: property declarations, popover lifecycle,
+ * event dispatch, trigger resolution, ARIA wiring, and `HoverController` integration.
  *
  * @slot - Text label displayed in the tooltip.
  */
-export abstract class TooltipBase extends SpectrumElement {
+export abstract class TooltipBase
+  extends SpectrumElement
+  implements HoverControllerHost
+{
   // ──────────────────
   //     SHARED API
   // ──────────────────
@@ -64,7 +65,7 @@ export abstract class TooltipBase extends SpectrumElement {
 
   /**
    * The preferred placement of the tooltip relative to its trigger.
-   * Controls the tip direction via the reflected `placement` attribute; pixel positioning requires `PlacementController` (additive phase).
+   * Determines the tip arrow direction. Pixel positioning requires `PlacementController`.
    *
    * @default 'top'
    */
@@ -81,16 +82,13 @@ export abstract class TooltipBase extends SpectrumElement {
 
   /**
    * The `id` of the trigger element in the same document tree root.
-   * Resolved via `getRootNode().getElementById(this.for)`.
-   * Drives ARIA relationship wiring on `open` change.
    */
   @property({ attribute: 'for', type: String })
   public for: string | undefined;
 
   /**
-   * Explicit trigger element reference. Overrides `for` when set.
-   * Use for cross-shadow-root triggers or programmatic insertion where `getElementById` is scoped to the wrong root.
-   * Setter only; no HTML attribute.
+   * Explicit trigger element reference; overrides `for` when set.
+   * Use when `getElementById` cannot reach the trigger, such as across a shadow boundary.
    *
    * @default null
    */
@@ -98,14 +96,8 @@ export abstract class TooltipBase extends SpectrumElement {
   public triggerElement: HTMLElement | null = null;
 
   /**
-   * Duration in milliseconds of the warm-up delay before the tooltip shows on pointer hover.
-   * Set to `0` to show immediately on hover. Keyboard focus (`focusin` when `:focus-visible`)
-   * always shows the tooltip immediately regardless of this value. The cooldown duration after the
-   * pointer leaves the trigger matches this value. Warm-up/cooldown state is shared across all tooltips in the same document,
-   * so moving quickly between adjacent triggers (e.g. a toolbar) shows each subsequent tooltip
-   * immediately after the first warm-up elapses.
-   *
-   * Additive/deferred: active when `HoverController` is integrated.
+   * Warm-up delay in milliseconds before the tooltip opens on pointer hover.
+   * Set to `0` to open immediately. Keyboard focus always opens immediately.
    *
    * @default 1500
    */
@@ -113,10 +105,7 @@ export abstract class TooltipBase extends SpectrumElement {
   public delay: number = 1500;
 
   /**
-   * When set, prevents automatic trigger wiring from responding to hover and focus events.
-   * No-op when `manual` is also set.
-   *
-   * Additive/deferred: active when `HoverController` is integrated.
+   * When set, the tooltip does not respond to hover or focus events.
    *
    * @default false
    */
@@ -124,11 +113,8 @@ export abstract class TooltipBase extends SpectrumElement {
   public disabled: boolean = false;
 
   /**
-   * Suppresses controller wiring for automatic hover and focus open/close.
-   * The consumer manages visibility via the `open` property or the popover API directly.
-   * ARIA relationship wiring still fires on `open` change when `for` or `triggerElement` is set.
-   *
-   * Additive/deferred: effective when controllers are integrated.
+   * Suppresses automatic hover and focus wiring.
+   * The consumer manages visibility via the `open` property or the popover API.
    *
    * @default false
    */
@@ -137,9 +123,7 @@ export abstract class TooltipBase extends SpectrumElement {
 
   /**
    * Pixel offset between the tooltip and its trigger.
-   * Passed to `PlacementController` offset middleware.
-   *
-   * Additive/deferred: active when `PlacementController` is integrated.
+   * Requires `PlacementController`; currently inactive.
    *
    * @default 0
    */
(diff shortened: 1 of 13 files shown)


Rajdeepc (commented):
I went through your PR in depth and want to share some thoughts and questions on the patterns you flagged mostly on the generalisation part

1. You have a deferred inline style cleanup to prevent flash. I totally understand why you did but there is an edge case and iwan tot know your views here. If the close in cancalled mid fade that means if the user re-hovers a popover thats fading out start() gets called but the stale translate and ip left from the previous sessions will still be on the element. They will be incorrect if the trigger moved during the fade animation.  Once alternative I am thinking you can do is instead of removing the inline styles post transition you can toggle a CSS class that controls whether the translate is visually applied. That way styles say on the DOM but are deactivated via specificity?
2. For the position absolute should we make strategy a PlacementOptions field? Coz since this is directly baked into the shared controller any future consumer on a non popover element inside a positioned ancestor would get wrong coordinates.
3. Can you elaborate on what you are seeing in the delay for flip position using starting-style? Is this a flash or more noticeable?

5t3ph:
@Rajdeepc 

> 1. You have a deferred inline style cleanup to prevent flash. I totally understand why you did but there is an edge case and iwan tot know your views here. If the close in cancalled mid fade that means if the user re-hovers a popover thats fading out start() gets called but the stale translate and ip left from the previous sessions will still be on the element. They will be incorrect if the trigger moved during the fade animation.  Once alternative I am thinking you can do is instead of removing the inline styles post transition you can toggle a CSS class that controls whether the translate is visually applied. That way styles say on the DOM but are deactivated via specificity?

I'm not sure if you actually encountered this condition in practice, but with only 300ms to fade I could not actually move quick enough for this condition to exist 😅 Plus with the the FloatingUI absolute positioning strategy, the translate value doesn't actually change upon (typical) scroll, just resize, which would be even harder to actually activate and get back to the trigger under 300ms. A CSS class solution wouldn't quite work since the styles are written inline which means they would have higher specificity.

> 2. For the position absolute should we make strategy a PlacementOptions field? Coz since this is directly baked into the shared controller any future consumer on a non popover element inside a positioned ancestor would get wrong coordinates.

I was wondering that, however it still works for the non-Popover elements in the PlacementController demo stories, so I'm not yet sure it needs to be an option?

> 3. Can you elaborate on what you are seeing in the delay for flip position using starting-style? Is this a flash or more noticeable?

It is very barely noticable for Tooltip specifically simply because it is generally a physically small in overall dimensions and the transition isn't very long, so I only really noticed it because I knew the "correct" appearance it should have had. I wanted to call it out for Popover because for something physically larger, like a select listbox, having say a bottom positioned one flip to top might make the fact it is "settling down from the top" rather than correctly "extending up from the top" post-flip might be more noticeable/jarring. It does resolve on the next interaction if the user moves away and hovers/opens the popover again.

This video demonstrates correct transition direction, then the flip with initial incorrect direction, followed by the correction on next interaction. I do not think it is severely detrimental for Tooltip, and not worth the tradeoff of juggling classnames vs. use of `@starting-style` to handle these transitions.

https://github.com/user-attachments/assets/c7833f2a-4e9c-4644-98ef-55be28a13cce

CONTRIBUTOR-DOCS/03_project-planning/03_components/tooltip/migration-plan.md
```diff
@@ -474,7 +476,7 @@ Follow the [Badge migration reference](../../02_workstreams/02_2nd-gen-component
 | Layer | Path | Contains |
 | ----- | ---- | -------- |
 | **Core** | `2nd-gen/packages/core/components/tooltip/` | `Tooltip.base.ts`, `Tooltip.types.ts`: property definitions (including `triggerElement` declaration), type validation, state management, accessible-name rules. Sets `role="tooltip"` and `popover="auto"` on the host element via `connectedCallback`. Wires `beforetoggle`/`toggle`/`transitionend` listeners for state sync and `swc-open`/`swc-after-open`/`swc-close`/`swc-after-close` dispatch. Resolves trigger via `for` attribute or `triggerElement` property. Maintains `Element.ariaDescribedByElements` on the trigger's inner interactive element on `open` change. No rendering. No Floating UI. |
```
rise-erpelding: [nit] Does this "No Floating UI" in the migration plan need an update too?

2nd-gen/packages/swc/components/tooltip/tooltip.mdx
```diff
@@ -42,7 +38,9 @@ The `placement` attribute sets the preferred position of the tooltip relative to
 
 The `start` and `end` values are RTL-aware: `start` maps to the left side in LTR and to the right in RTL; `end` maps to the right in LTR and to the left in RTL.
 
-Pixel-accurate positioning requires `PlacementController`, which is in a future release. The `placement` attribute currently applies only the tip direction.
+`PlacementController` handles pixel positioning automatically. When viewport space is insufficient on the requested side, the tooltip flips to the opposite side and the `placement` attribute reflects the computed side. The original requested value is preserved internally, so if the viewport changes the tooltip can return to the originally requested position.
```
rise-erpelding: Not absolutely necessary, but it would be nice to have a story that shows this, or to mention that you can see this behavior on the Placements story below by changing the viewport (and changing it back).

2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
```diff
@@ -290,6 +425,38 @@ export abstract class TooltipBase extends SpectrumElement {
...
+    // when positioning options or the trigger change while already open.
+    const openChanged = changedProperties.has('open');
+    if (openChanged) {
+      if (this.open) {
+        this.startPlacement();
+      } else {
+        // Stop autoUpdate immediately so the position stops tracking the trigger
+        // during the exit transition. The translate/top/left are cleared by
+        // dispatchAfterEvent once the exit transition completes.
+        this.placementController.stop();
```
rise-erpelding, line 453: I'm not sure if this would be a fix in this area, but I notice when the tooltip closes, when I apply a `cross-offset` (in this case, 20px, with tooltip top positioned), I see the tip snap back to its CSS-centered position mid-fade, do we need to handle that?

Here is a video if you are able to see it!

https://github.com/user-attachments/assets/e1571c7f-cb50-43e4-988b-01ec0bd89f54

rise-erpelding (commented):
Left some comments, but also wanted to note: I see the `should-flip` functionality working in the `Placement` story so I do have some faith that it's working, but I'm seeing it behave a little oddly in the Storybook playground, it will open on the bottom here even if it doesn't have enough space. I took a quick video if you're able to see it:

https://github.com/user-attachments/assets/c284dfa4-c8b6-43f5-8986-8a20ebc0caf6

Rajdeepc:
> @Rajdeepc
> 
> > 1. You have a deferred inline style cleanup to prevent flash. I totally understand why you did but there is an edge case and iwan tot know your views here. If the close in cancalled mid fade that means if the user re-hovers a popover thats fading out start() gets called but the stale translate and ip left from the previous sessions will still be on the element. They will be incorrect if the trigger moved during the fade animation.  Once alternative I am thinking you can do is instead of removing the inline styles post transition you can toggle a CSS class that controls whether the translate is visually applied. That way styles say on the DOM but are deactivated via specificity?
> 
> I'm not sure if you actually encountered this condition in practice, but with only 300ms to fade I could not actually move quick enough for this condition to exist 😅 Plus with the the FloatingUI absolute positioning strategy, the translate value doesn't actually change upon (typical) scroll, just resize, which would be even harder to actually activate and get back to the trigger under 300ms. A CSS class solution wouldn't quite work since the styles are written inline which means they would have higher specificity.
> 
> > 2. For the position absolute should we make strategy a PlacementOptions field? Coz since this is directly baked into the shared controller any future consumer on a non popover element inside a positioned ancestor would get wrong coordinates.
> 
> I was wondering that, however it still works for the non-Popover elements in the PlacementController demo stories, so I'm not yet sure it needs to be an option?
> 
> > 3. Can you elaborate on what you are seeing in the delay for flip position using starting-style? Is this a flash or more noticeable?
> 
> It is very barely noticable for Tooltip specifically simply because it is generally a physically small in overall dimensions and the transition isn't very long, so I only really noticed it because I knew the "correct" appearance it should have had. I wanted to call it out for Popover because for something physically larger, like a select listbox, having say a bottom positioned one flip to top might make the fact it is "settling down from the top" rather than correctly "extending up from the top" post-flip might be more noticeable/jarring. It does resolve on the next interaction if the user moves away and hovers/opens the popover again.
> 
> This video demonstrates correct transition direction, then the flip with initial incorrect direction, followed by the correction on next interaction. I do not think it is severely detrimental for Tooltip, and not worth the tradeoff of juggling classnames vs. use of `@starting-style` to handle these transitions.
> 
>  tooltip-flip-transition.mov
@5t3ph Thanks for the clarifications.
1. Acceptable for tooltip but the real fix for Popover would be that on re-open we should immediately call startPlacement() which calls stop() then re-wrires and this would overwrite any stale inline styles with fresh coordinates on the same frame.
2. I am fine now as is but we should file a follow-up to make strategy configurable if Popover needs it.
3. I see this is  a fundamental limitation of @starting-style with async placement computation. Can we document this for Popover consumers if you haven't.

CONTRIBUTOR-DOCS/03_project-planning/03_components/tooltip/migration-plan.md
5t3ph: Updates made to match migration plan to implementation 👍

2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
5t3ph: Good catch! Resolved - the cleanup was happening too early via PlacementController, so moved to be a responsibility of the render to handle post close transition.

2nd-gen/packages/swc/components/tooltip/tooltip.mdx
5t3ph: re: this as well as your comment on `should-flip` - added clarification in the docs to emphasize that this can _only_ be viewport-relative due to popover's top-layer placement, which means it can't look to an ancestor scroll container. Hence why you aren't seeing it change within Playground (not close enough to "viewport" edge).

I was validating it from the Docs page on the Overview example which is able to get close enough to the viewport to trigger the flip behavior (as in my video response in an earlier comment).

PR #6387, opened 2026-06-10
feat(tooltip): Migrate tooltip to Gen2

5t3ph:
## Description

Adds the 2nd-gen `<swc-tooltip>` component, completing the Spectrum 1 → Spectrum 2 migration.

The implementation was built and reviewed incrementally through several PRs into this feature branch — see the commit history for the individual review artifacts. This PR merges the completed feature branch into `main`.

Full API decisions, behavioral rationale, and architectural notes are in the migration plan (`CONTRIBUTOR-DOCS/03_project-planning/03_components/tooltip/migration-plan.md`).

## Motivation and context

Migrates `<sp-tooltip>` to the 2nd-gen architecture: native Popover API replaces `sp-overlay`; `HoverController` and `PlacementController` provide hover/focus wiring and viewport-aware positioning; `role="tooltip"` and `ariaDescribedByElements` ARIA wiring are added as critical a11y fixes.

## Related issue(s)

- Epic: SWC-2017

## Screenshots

image

## Author's checklist

## Reviewer's checklist

- [ ] All VRTs are approved

### Manual review test cases

- [ ] `<swc-tooltip for="id">` opens on hover and focus of the referenced trigger; closes on pointer leave, focus loss, and Escape
  1. Go to the Tooltip Storybook
  2. Hover the trigger button; verify tooltip opens after warm-up delay
  3. Focus the trigger via Tab; verify tooltip opens immediately
  4. Press Escape; verify tooltip closes and focus stays on trigger
  5. Move pointer from trigger into the tooltip bubble; verify tooltip stays open

- [ ] Breaking changes from `<sp-tooltip>` behave as documented in the consumer migration guide
  1. Go to Tooltip > Migration guide in Storybook
  2. Confirm `slot="icon"`, `variant="positive"`, `variant="info"`, and `self-managed` are absent from the 2nd-gen API
  3. Confirm event names (`swc-open`, `swc-after-open`, `swc-close`, `swc-after-close`) are documented

### Device review

- [ ] Did it pass in Desktop?

  1. Go to Tooltip Storybook
  2. Tab to the trigger button; verify tooltip opens and `role="tooltip"` is announced by screen reader
  3. Press Escape; verify tooltip closes and focus stays on the trigger with no trap
  4. Tab away; verify tooltip closes

  1. Go to Tooltip Storybook > Accessibility story
  2. With NVDA (Windows) or VoiceOver (macOS), focus the trigger; verify the trigger name is announced followed by the tooltip text as a description
  3. For the `labeling` example: verify the tooltip text is announced as the trigger's accessible name, not a description
  4. Verify closed tooltips are not reachable or announced by the screen reader


2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
```diff
@@ -0,0 +1,491 @@
...
+      }
+    }
+    if (changedProperties.has('open') || changedProperties.has('labeling')) {
+      this.syncAriaRelationship();
+    }
+    if (
+      changedProperties.has('for') ||
+      changedProperties.has('triggerElement')
+    ) {
+      this.hoverController.setTarget(this.resolveTrigger());
```
rise-erpelding, line 489: If `open` or `labeling` change, `syncAriaRelationship` updates `aria-describedby` and `aria-labelledby`, but if `for` or `triggerElement` change, those aria attributes aren't updated -- could that potentially cause an issue?

2nd-gen/packages/swc/components/tooltip/tooltip.mdx
```diff
@@ -0,0 +1,159 @@
...
+### Events
+
+`<swc-tooltip>` dispatches four lifecycle events as visibility changes:
+
+- **`swc-open`**: Dispatched when the tooltip begins to open, before the enter transition plays.
+- **`swc-after-open`**: Dispatched after the enter transition completes.
+- **`swc-close`**: Dispatched when the tooltip begins to close, before the exit transition plays.
+- **`swc-after-close`**: Dispatched after the exit transition completes.
+
+When no CSS transition is active (for example, when `prefers-reduced-motion` removes it), `swc-after-open` and `swc-after-close` dispatch synchronously on the same tick as `swc-open` and `swc-close`.
```
rise-erpelding, line 87: I don't think we have `prefers-reduced-motion` in CSS, the a11y plan does mention it (the last line in the "[Guidelines that apply](https://github.com/adobe/spectrum-web-components/blob/swc-2017/tooltip/CONTRIBUTOR-DOCS/03_project-planning/03_components/tooltip/accessibility-migration-analysis.md#guidelines-that-apply)" table/[L160](https://github.com/adobe/spectrum-web-components/blob/swc-2017/tooltip/CONTRIBUTOR-DOCS/03_project-planning/03_components/tooltip/accessibility-migration-analysis.md?plain=1#L160)) - was it deliberately left out?

2nd-gen/packages/swc/components/tooltip/stories/tooltip.stories.ts
```diff
@@ -0,0 +1,509 @@
...
+  `,
+  args: {
+    variant: 'neutral',
+    placement: 'top',
+    'default-slot': 'Save your changes',
+  },
+  tags: ['behaviors'],
+};
+
+export const TriggerElement: Story = {
```
rise-erpelding, line 469: Does this need a `storyName`? Per the [stories-format rule](https://github.com/adobe/spectrum-web-components/blob/swc-2017/tooltip/.ai/rules/stories-format.md#story-naming):

> When the camelCase export name does not produce a readable display name — for example, TextWrapping for a story that should appear as "Text wrapping" — set the display name via storyName assigned after the export

2nd-gen/packages/swc/components/tooltip/test/tooltip.a11y.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+      - button "Open"
+    `);
+  });
+
+  test('open tooltip exposes role="tooltip" in accessibility tree', async ({
+    page,
+  }) => {
+    await gotoStory(page, 'components-tooltip--overview', 'swc-button');
+
+    // Open the tooltip programmatically (HoverController not yet wired).
```
rise-erpelding: Is this comment stale?

2nd-gen/packages/swc/components/tooltip/test/tooltip.test.ts
```diff
@@ -0,0 +1,1161 @@
...
+      // cause an unbounded hang if we waitForEvent('swc-after-close').
+      await waitFor(
+        () =>
+          expect(
+            tooltip.style.translate,
+            'bubble translate is cleared after close'
+          ).toBeFalsy(),
+        { timeout: 2000 }
+      );
+      // Tip translate is cleared synchronously by placementController.stop().
```
rise-erpelding: Did this change? I think we're handling this in `dispatchAfterEvent` now so the comment might be a bit stale.

CONTRIBUTOR-DOCS/03_project-planning/03_components/tooltip/migration-plan.md
```diff
@@ -599,130 +603,133 @@ The impact is most acute in the additive phase, when `HoverController` will call
...
+- [x] Automatic mode: `trigger-element` wires correctly for cross-shadow-root trigger relationships — implementation complete (`triggerElement` bypasses `getElementById` scoping); not explicitly tested with a cross-root fixture
+- [x] Automatic mode: `delay="0"` shows immediately on hover (`HoverOpensTest`); full 1500ms warm-up/cooldown timing covered by hover-controller test suite
+- [x] Automatic mode: `for` resolves correctly when trigger and tooltip share a shadow root (`ShadowRootScopeTest` — creates trigger + tooltip inside `attachShadow({ mode: 'open' })`; verifies `getRootNode().getElementById` scopes to the shadow root and ARIA wiring reaches the native trigger)
+- [x] `manual` attribute: controller wiring is skipped; consumer-driven `open` changes still work; ARIA wiring still fires (`ManualPreventsHoverTest`, `AriaWiringManualModeTest`)
+- [x] `labeling` attribute: `ariaLabelledByElements` is set on the inner interactive element instead of `ariaDescribedByElements`; stale references in the opposite property are cleaned up; re-syncs when `labeling` changes while open (`LabelingAriaWiringTest`)
```
rise-erpelding: [nit] this is implemented as `AriaWiringLabelingTest`, the name is a tiny bit off

rise-erpelding (commented):
Doing a quick pass and will come back tomorrow!

nikkimk (approved):
lgtm

2nd-gen/packages/swc/components/tooltip/stories/tooltip.stories.ts
5t3ph: It seems like it is getting displayed correctly as "Trigger element"

2nd-gen/packages/swc/components/tooltip/test/tooltip.a11y.spec.ts
5t3ph: updated

2nd-gen/packages/swc/components/tooltip/test/tooltip.test.ts
5t3ph: updated

2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
5t3ph: A pretty low likelihood of changing the trigger when the tooltip is already open but went ahead and added the patch for this just in case!

2nd-gen/packages/swc/components/tooltip/tooltip.mdx
5t3ph: Good catch - added!

2nd-gen/packages/swc/components/tooltip/stories/tooltip.stories.ts
```diff
@@ -0,0 +1,513 @@
...
+} from '@spectrum-web-components/core/components/tooltip';
+
+import '@adobe/spectrum-wc/components/button/swc-button.js';
+import '@adobe/spectrum-wc/components/tooltip/swc-tooltip.js';
+
+// ────────────────
+//    METADATA
+// ────────────────
+
+const { args, argTypes, template } = getStorybookHelpers('swc-tooltip');
```
rise-erpelding, line 33: I tested out the Storybook controls, here are some thoughts I have:

* The `for` control doesn't have any effect with the way that we have our Storybook implementation set up, we might consider removing the control there?
* Ditto `labeling`, maybe? I didn't see a way to see this change in Storybook, but it's possible I missed something.
* I can't remember if I mentioned this one before, but the `container-padding` and `should-flip` seem to work fine, but I'm seeing Storybook having an issue if I change `placement` to `bottom` and I shrink the viewport so that there's no space for the tooltip on the bottom, it still opens on the bottom (see screenshot below). This is super minor, I think it's probably a Storybook limitation, so it's probably not a huge concern but could at some point cause confusion.

Image

2nd-gen/packages/swc/components/tooltip/tooltip.mdx
```diff
@@ -0,0 +1,159 @@
...
+
+#### Keyboard behavior
+
+- <kbd>Escape</kbd>: closes the tooltip; focus remains on the trigger. Handled
+  by native `popover="auto"`.
+- Tooltip text is never in the tab order; focus always stays on the trigger.
+
+#### High contrast mode
+
+A `1px solid transparent` border in base styles is automatically filled with `CanvasText` in forced-colors mode, ensuring the tooltip bubble remains visible in Windows High Contrast.
```
rise-erpelding: The border is only added for forced-colors, it doesn't exist as a transparent border outside of forced-colors mode. It might be worthwhile to make a small update here.

rise-erpelding (approved):
I have only a couple of nits that I don't consider to be blocking!

2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
```diff
@@ -0,0 +1,522 @@
...
+  }
+
+  set placement(value: TooltipPlacement) {
+    const old = this._placement;
+    this._placement = value;
+    if (!this._placementFromController) {
+      this._requestedPlacement = value;
+    }
+    this.requestUpdate('placement', old);
+  }
```
rubencarvalho: (Sorry for the long wall of text, couldn't really make it shorter to get my point across)

Sooooo, I was checking and I don't think we can/should do this... the placement stops representing what the consumer actually set.

They might write `placement="start"`, but after the first compute, `fromFloatingPlacement()` only returns a physical side. The attribute then gets rewritten to left or right, depending on direction. If the consumer reads .placement after opening, they get a value they never provided.

That’s why we currently need the _requestedPlacement / _placementFromController "dance": we have to preserve the original request because the public property no longer does.

I’d prefer placement to remain the consumer’s requested value and never be updated internally. The resolved side could live in a separate reflected `actual-placement` attribute (mb owned by onPlacementChange?).

The main benefit for me is the CSS. If I understand correctly, we currently need both logical selectors (start/end, including the :dir(rtl) flips) and physical selectors (left/right) because the same attribute is logical before the first compute and physical afterwards.

If actual-placement is always be physical (since the controller already resolves logical placement against the current direction). If the rendering is based on that attribute, we can remove the start/end selectors and the RTL-specific flips. Removing the flag is mostly a side effect of that simplification.

BUT there is still one thing we’d need to account for: the opening animation.
 he first placement computation is deferred because `computePlacement()` waits for `document.fonts.ready`, and WebKit also gets an additional `requestAnimationFrame`. That means `onPlacementChange` may not run until a frame or two after the popover is painted. rn, the animation starts from the correct side immediately because the logical placement is already reflected on the host. If the CSS instead relies on actual-placement and only the async callback sets it, the first frames could/would use a stale value.
We could avoid that by setting actual-placement synchronously when opening: resolve the requested placement to a physical side for the current direction, using `toFloatingPlacement()`, immediately before `showPopover()`. onPlacementChange would then only need to update it if the controller flips the placement.

I hope this makes somewhat sense. LMK if you want to talk!

2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
```diff
@@ -0,0 +1,522 @@
...
+    }
+    if (changedProperties.has('open')) {
+      if (this.open !== this.isPopoverOpen) {
+        if (this.open) {
+          this.showPopover();
+        } else {
+          this.hidePopover();
+        }
+      }
+    }
```
rubencarvalho: Seeing this all put together now, I have some thoughts... If I understand correctly, the `HoverController` calls `host.showPopover()`/`hidePopover()` directly, and this block here *also* drives the popover from the `open` property. They seem to reconcile through the native `toggle` event plus the `isPopoverOpen` / `isOpen !== this.open` guards... Feels like a lot is happening + there's multiple "sources of truth"

I know tooltip is the only `HoverController` consumer today, but have we considered making `open` the single source of truth for now?: e.g., changing the `HoverController` contract from `showPopover()`/`hidePopover()` to `requestOpen()`/`requestClose()` (tooltip implements them as `this.open = true/false`), leaving this `updated()` block the only caller of the popover API. `handleToggle` then only needs to resync on native light-dismiss.

These are mostly thoughts, I'm happy to discuss this further!

2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
```diff
@@ -0,0 +1,522 @@
...
+      this.style.removeProperty('left');
+      this.style.removeProperty('--swc-placement-available-width');
+      this.style.removeProperty('--swc-placement-available-height');
+
+      if (this.tipElement) {
+        this.tipElement.style.removeProperty('translate');
+        this.tipElement.style.removeProperty('top');
+        this.tipElement.style.removeProperty('left');
+      }
+    }
```
rubencarvalho: nit: this method is called "dispatchAfterEvent" but we're also cleaning up the styles. Could we separate these concerns?

2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
```diff
@@ -0,0 +1,522 @@
...
+        stale.ariaLabelledByElements ?? []
+      ).filter((el) => el !== this);
+      this._lastWiredTrigger = null;
+    }
+
+    const trigger = this.resolveTrigger();
+    if (!trigger) {
+      return;
+    }
+    const target = (trigger.shadowRoot?.querySelector('button') ??
```
rubencarvalho, line 321: (No action required!), but I keep coming back to this thinking we may need to extend this selector to all focusable elements. Kinda of re-sharing this thought so we keep it in mind!

2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
```diff
@@ -0,0 +1,522 @@
...
+    if (event.target !== this || !this.afterEventPending) {
+      return;
+    }
+    if (this.afterEventFallbackTimer !== null) {
+      clearTimeout(this.afterEventFallbackTimer);
+      this.afterEventFallbackTimer = null;
+    }
+    this.afterEventPending = false;
+    this.dispatchAfterEvent(this.open);
+  };
```
rubencarvalho, line 442: I think this beforetoggle/toggle/after-event dispatch, the transitionend coordination, and especially the Firefox fallback will apply to *any* transitioning native-popover. Popover/Menu/Picker overlays will prob need the same logic, including the same Firefox quirk and the same flash-avoidance ordering.

Because tooltip is the first popover-based Gen2 component, this is great, but I may consider making this shareable or part of a controller (thinking of my Popover here).

2nd-gen/packages/swc/components/tooltip/Tooltip.ts
```diff
@@ -0,0 +1,61 @@
...
+  public static override get styles(): CSSResultArray {
+    return [styles];
+  }
+
+  protected override get tipElement(): HTMLElement | null {
+    return (
+      (this.renderRoot?.querySelector('.swc-Tooltip-tip') as HTMLElement) ??
+      null
+    );
+  }
```
rubencarvalho, line 61: I think we would be safe caching the tip element, this now re-runs `renderRoot.querySelector('.swc-Tooltip-tip')` on every access, and it's read from both `startPlacement()` and `dispatchAfterEvent()` on each open/close cycle. I'm assuming the tip will always be static in the template, so it can be resolved once after `firstUpdated()` and stored (mb with a null reset on disconnect?).

rubencarvalho (commented):
Looking great!! This will help me a lot with the Popover stuff too. I do have a couple of comments I _think_ we should try to address, though. Happy to sync on any of these 😄

2nd-gen/packages/swc/components/tooltip/stories/tooltip.stories.ts
5t3ph: re flip-related behavior - [I tried to explain that here](https://github.com/adobe/spectrum-web-components/pull/6376#discussion_r3389564176) but TL;DR it is _viewport_ relative due to the top-layer behavior of popover.

`labeling` is flipping the aria method so it doesn't have a visual change, just a behavior one

good call on removing `for` from controls, updated 👍

2nd-gen/packages/swc/components/tooltip/tooltip.mdx
5t3ph: true, updated!

2nd-gen/packages/swc/components/tooltip/tooltip.mdx
```diff
@@ -0,0 +1,159 @@
...
+#### ARIA role and relationship
+
+1. **`role="tooltip"`**: set on the host element via `connectedCallback`.
+2. **`ariaDescribedByElements`**: when the tooltip opens, the SWC layer resolves the trigger via `for` or `triggerElement` and sets `Element.ariaDescribedByElements = [tooltipHost]` on the trigger's interactive surface. For 2nd-gen SWC components with an open shadow root, the inner `<button>` is targeted; for native elements, the trigger host is used directly. The association is removed when the tooltip closes.
+3. **`labeling` attribute**: when set, the SWC layer wires `ariaLabelledByElements` instead of `ariaDescribedByElements` on the trigger's inner interactive element. For icon-only triggers where the tooltip text is the sole accessible name and a name cannot be added to the trigger directly.
+
+#### Keyboard behavior
+
+- <kbd>Escape</kbd>: closes the tooltip; focus remains on the trigger. Handled
+  by native `popover="auto"`.
```
Rajdeepc, line 132: in Tooltip.base.ts, your implementation is using document level keydown listener as [here](https://github.com/adobe/spectrum-web-components/blob/48538806773d3e1041cde0d21f52f593852f69e2/2nd-gen/packages/core/components/tooltip/Tooltip.base.ts#L425-L429)

2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
```diff
@@ -0,0 +1,522 @@
...
+
+  public override disconnectedCallback(): void {
+    super.disconnectedCallback();
+    this.removeEventListener('beforetoggle', this.handleBeforeToggle);
+    this.removeEventListener('toggle', this.handleToggle);
+    this.removeEventListener('transitionend', this.handleTransitionEnd);
+    document.removeEventListener('keydown', this.handleKeyDown);
+    if (this.afterEventFallbackTimer !== null) {
+      clearTimeout(this.afterEventFallbackTimer);
+      this.afterEventFallbackTimer = null;
```
Rajdeepc, line 558: Is there any reason you are not calling `syncAriaRelationship()` to remove the tooltip from `ariaDescribedByElements` on the trigger. If a tooltip is removed from DOM while open the trigger can retain a reference to a disconnected node.
Maybe you can do this in disconnectedCallback - clear `_lastWiredTrigger` references like `set open = false` or run cleanup directly.

2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
```diff
@@ -0,0 +1,522 @@
...
+    this.afterEventPending = false;
+    this.dispatchAfterEvent(this.open);
+  };
+
+  // Allows Escape behavior to be testable, does not interfere with native popover dismissal
+  private readonly handleKeyDown = (event: KeyboardEvent): void => {
+    if (event.key === 'Escape' && this.open) {
+      this.open = false;
+    }
+  };
```
Rajdeepc, line 454: Every connected tooltip instance registers on document. Would advise you to use host level listener or a single delegated listener.

2nd-gen/packages/swc/components/tooltip/tooltip.mdx
```diff
@@ -0,0 +1,159 @@
...
+
+### Best practices
+
+- **Use focusable triggers only.** Place `<swc-tooltip>` only on elements that can receive keyboard focus. Non-interactive elements require a focusable wrapper or a different pattern such as contextual help.
+- **Use plain text.** Do not include interactive elements or rich formatting inside a tooltip. `role="tooltip"` is not focusable; use `<swc-popover>` or contextual help for disclosures that require interactive content.
+- **Supplement, do not duplicate.** Tooltip text supplements the trigger's accessible name. Do not repeat what `aria-label` or `aria-labelledby` already conveys.
+- **Icon-only triggers.** Prefer adding an accessible name directly to the trigger: `aria-label` on native elements; `accessible-label` on 2nd-gen SWC components. Use the `labeling` attribute when the trigger host cannot be modified.
+- **No auto-dismiss timer.** The tooltip must remain visible until the user dismisses it or the triggering state ends (WCAG 1.4.13).
+- **Touch devices.** Hover-triggered tooltips are not accessible on touch-only devices. Use `<swc-popover>` or contextual help for touch disclosure.
```
Rajdeepc: Maybe product teams should also know `delay="0"` doesn't fix touch?

2nd-gen/packages/core/controllers/placement-controller/src/placement-controller.ts
```diff
@@ -292,25 +294,13 @@ export class PlacementController implements ReactiveController {
   }
 
   /**
-   * Stop positioning: tear down `autoUpdate`, clear `actualPlacement`,
-   * reset `isConstrained`, and remove inline styles the controller wrote
-   * (the `size` middleware's `--swc-placement-available-width` /
-   * `--swc-placement-available-height` custom properties on the floating
-   * element, plus the `arrow` middleware's `translate` / `top` / `left`
-   * on the tip element). Safe to call multiple times.
+   * Stop positioning: tear down `autoUpdate`, clear `actualPlacement` and
```
Rajdeepc, line 297: nit: If you can add a test around a tooltip with long string to test how placementController `autoUpdate` handle this, that would be nice.

Rajdeepc (commented):
Some nice to havee:
1. ARIA cleanup on disconnect while open.
2. Align Escape docs with implementation; consider scoping keydown listener.

2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
5t3ph: this update has been made! I'm not "reflecting" it, just setting and removing the attribute so that it doesn't become part of the actual API.

The opening animation is still slightly buggy post-flip, [as noted previously.](https://github.com/adobe/spectrum-web-components/pull/6376#issuecomment-4664105302) Your suggestion led to a _different_ inconsistency which was more noticeable of the physical placement being correct visually but the `actual-placement` not fully updating in all cases, leading to an incorrectly pointing arrow. I would say the slightly incorrect post-flip transition direction is not a blocker for now, but we could create a ticket to investigate further.

2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
5t3ph: updated

2nd-gen/packages/swc/components/tooltip/tooltip.mdx
5t3ph: As noted in the comment for that line, it's only for testability. Native light-dismiss on Escape is unreliable in headless/test environments, so the explicit handler guarantees `open` reflects the dismissal and keeps `open` authoritative.

2nd-gen/packages/core/controllers/placement-controller/src/placement-controller.ts
5t3ph: Added `LongContentPlacementTest` to cover this

2nd-gen/packages/swc/components/tooltip/tooltip.mdx
5t3ph: updated the wording to clarify the "why" here

2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
5t3ph: You're right that registering one document listener per connected instance doesn't scale. I've scoped it to the `open` state — it's now added in `updated()` when the tooltip opens and removed on close (and defensively on disconnect). Since `popover="auto"` only allows one open tooltip at a time, that means at most one listener is ever active, which gets us the single-delegated-listener behavior you're after.

I kept it on `document` rather than the host, though: the tooltip is non-interactive and never receives keyboard focus (focus stays on the trigger), so a host-level `keydown` would never fire. `document` is the only place that reliably catches `Escape` here.

2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
5t3ph: Good catch — confirmed and fixed. The relationship was only torn down on `open = false`, so removing the tooltip while open left the trigger holding a reference to the detached node. I pulled the removal logic into a `clearAriaRelationship()` helper and now call it from `disconnectedCallback()` as well as before each re-wire.

2nd-gen/packages/swc/components/tooltip/Tooltip.ts
5t3ph: updated

2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
5t3ph: Changed the `HoverControllerHost` contract from `showPopover()`/`hidePopover()` to `requestOpen()`/`requestClose()`. The host now owns its visibility state and `updated()` is the only caller of the Popover API; the controller asks the host to change `open` rather than driving the popover directly. This removes the dual-source-of-truth reconciliation between the controller's direct API calls and the `open` property.

2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
```diff
@@ -0,0 +1,521 @@
...
+   *
+   * @default false
+   */
+  @property({ type: Boolean, reflect: true })
+  public open: boolean = false;
+
+  /**
+   * The `id` of the trigger element in the same document tree root.
+   */
+  @property({ attribute: 'for', type: String })
```
rubencarvalho, line 94: nit: I don't think we need to specify the attribute when it matches the property

2nd-gen/packages/swc/components/tooltip/migration-guide.mdx
```diff
@@ -0,0 +1,273 @@
...
+| Trigger wiring attribute        | `self-managed`                                   | `for="[id]"` wires the ARIA relationship and drives automatic hover/focus open/close. |
+
+### Added in Spectrum 2
+
+| Addition                                    | Notes                                                                                                                                        |
+| ------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------- |
+| `for` attribute                             | Declares the trigger relationship by ID. Required for ARIA wiring and automatic hover/focus open/close.                                      |
+| `manual` boolean attribute                  | Opt out of automatic hover/focus wiring. Consumer manages open/close via `open`. ARIA wiring still fires.                                    |
+| `disabled` boolean attribute                | Suppresses automatic hover/focus response. Trigger remains interactive; `open` still works directly.                                         |
```
Rajdeepc: nit: This is now stale. Do you mind changing this statement?

Rajdeepc (approved):
Code design looks solid. Once CI is green and VRT is accepted go ahead and merge.

rubencarvalho (approved):
AWESOME work here!! Let's get it in!

2nd-gen/packages/swc/components/tooltip/migration-guide.mdx
5t3ph: updated!

PR #6395, opened 2026-06-11
feat(button-group): S2 migration

Rajdeepc:
## Description

Complete 2nd-gen migration of `<swc-button-group>` from 1st-gen `<sp-button-group>`. This PR lands the full feature branch (`button-group-migration`) into `main`, encompassing:

- **Core base class** (`ButtonGroupBase`) with `orientation`, `size`, `disabled`, `align` properties, size/disabled propagation to slotted children, and `role="group"` semantics
- **SWC component** with S2 CSS (`:host([attribute])` selectors, `token()` spacing, exposed `--swc-button-group-*` custom properties)
- **Storybook documentation** with Playground, Overview, Anatomy, Sizes, Orientations, Alignment, Disabled, and Accessibility stories
- **Tests**: Play-function unit tests and Playwright a11y tests (ARIA snapshots, keyboard interactions, aXe validation)
- **Consumer migration guide** (`migration-guide.mdx`) with What Changed tables, numbered update steps, and a11y/styling callouts
- **1st-gen deprecation** of the `vertical` property with `@deprecated` JSDoc and runtime `window.__swc.warn()`
- **Migration status table** and **migration plan checklist** updated

### Breaking changes from 1st-gen

| Area | `sp-button-group` | `swc-button-group` |
|------|--------------------|--------------------|
| Orientation | `vertical` boolean | `orientation="horizontal\|vertical"` |
| Default size | None (`noDefaultSize`) | `'m'` (propagated to children) |
| CSS overrides | `--mod-buttongroup-*` | `--swc-button-group-*` |

### New in 2nd-gen

- `align` property (`start`, `center`, `end`)
- `disabled` propagation from group to all children
- `role="group"` automatically set on host
- XL size support

## Motivation and context

Button group is the first component to complete the full multi-phase migration workflow (prep → setup → API → styling → a11y → testing → documentation → review). It serves as a reference implementation for subsequent component migrations.

## Related issue(s)

- Epic: SWC-2071
- SWC-2073 (migration plan)
- SWC-2074 (file structure, API, a11y)
- SWC-2075 (styling)
- SWC-2076 (testing)
- SWC-2077 (storybook docs)
- SWC-2078 (consumer migration guide)

---

## Author's checklist

---

## Reviewer's checklist

### Manual review test cases

- [ ] Verify horizontal layout with default settings
  1. Go to [Storybook → Button Group → Playground](http://localhost:6007/?path=/story/button-group--playground)
  2. Confirm buttons render in a horizontal row with 12px gap
  3. Expect no visual overflow or wrapping with 3 buttons

- [ ] Verify vertical orientation
  1. Set `orientation="vertical"` in Storybook controls
  2. Confirm buttons stack in a column
  3. Expect same gap spacing between items

- [x] Verify size propagation
  1. Change `size` control to `s`, `l`, `xl`
  2. Confirm all child buttons adopt the group's size
  3. Expect consistent sizing across all children

- [x] Verify disabled propagation
  1. Toggle `disabled` in controls
  2. Confirm all child buttons become disabled
  3. Expect no button is clickable when group is disabled

- [x] Verify alignment
  1. Set `align="end"` with sufficient container width
  2. Confirm buttons align to the inline-end
  3. Expect `center` centers the group

### Device review

- [ ] Did it pass in Desktop?

- [x] **Keyboard** (required)
  1. Go to [Storybook → Button Group → Playground](http://localhost:6007/?path=/story/button-group--playground)
  2. Press <kbd>Tab</kbd> — focus moves to the first button in the group
  3. Press <kbd>Tab</kbd> again — focus moves sequentially to each subsequent button (no roving tabindex)
  4. Press <kbd>Shift+Tab</kbd> — focus moves backwards through buttons
  5. Press <kbd>Enter</kbd> or <kbd>Space</kbd> on a focused button — button activates
  6. Confirm the group host element itself is NOT focusable
  7. When `disabled` is set on group, confirm no button receives focus via Tab

- [x] **Screen reader** (required)
  1. Navigate to button group with VoiceOver/NVDA
  2. Expect announcement: "group" role with the `aria-label` value (e.g. "Document actions, group")
  3. Navigate into the group — each button is announced individually with its label
  4. Confirm `aria-orientation` is NOT announced (intentionally omitted since no arrow-key navigation)
  5. When `disabled` is set, confirm each button announces as "dimmed" / "unavailable"


2nd-gen/packages/core/components/button-group/ButtonGroup.base.ts
```diff
@@ -0,0 +1,153 @@
...
+   * children receive the current size and disabled state.
+   */
+  protected handleSlotchange(): void {
+    this.propagateToChildren();
+  }
+
+  private propagateToChildren(): void {
+    for (const button of this.buttons) {
+      button.size = this.size;
+      button.disabled = this.disabled;
```
blunteshwar: It looks like `propagateToChildren()` always writes the group’s `disabled` value onto every slotted button ,which means an individually disabled child gets re-enabled whenever the group itself is enabled, including on initial slotchange:

blunteshwar (changes requested):

2nd-gen/packages/core/components/button-group/ButtonGroup.base.ts
Rajdeepc: If a consumer needs one button disabled while the group is enabled, they should set disabled on the individual button imperatively after render (or use a reactive binding), rather than as a static attribute at slot-time. This matches the group's ownership model. It's authoritative, not additive.
We could optionally add a disabled propagation mode that only forces true (not false), but that would be an additive enhancement and introduce asymmetry with size propagation.

blunteshwar: I get the “group owns propagated state” model for `size`, but I’m less convinced it should apply the same way to `disabled`.

A statically disabled child is normal declarative HTML/component usage, and clearing that state on slotchange is surprising because `disabled` affects interaction semantics, not just presentation. It also conflicts with the migration guide text that says individual buttons can still be disabled independently when the group itself is not disabled.

If the intended contract is that `swc-button-group` is authoritative for `disabled` and individual child disabled states are unsupported, then I think we should update the docs to say that directly. Otherwise, I’d still prefer preserving child disabled state when the group is enabled, and only forcing `true` while the group-level `disabled` state is active.

blunteshwar (approved):
nit:
Approving the PR, but I am still inclined towards adding a disabled propagation mode that only forces true (not false)

2nd-gen/packages/swc/components/button-group/button-group.css
```diff
@@ -0,0 +1,46 @@
...
+:host {
+  display: flex;
+}
+
+* {
+  box-sizing: border-box;
+}
+
+.swc-ButtonGroup {
+  --_swc-button-group-gap: var(--swc-button-group-gap, token("spacing-300"));
```
nikkimk: <blockquote>
[ ] Verify horizontal layout with default settings

i. Go to [Storybook → Button Group → Playground](http://localhost:6007/?path=/story/button-group--playground)
ii. Confirm buttons render in a horizontal row with 12px gap
iii. Expect no visual overflow or wrapping with 3 buttons

[ ] Verify vertical orientation

i. Set orientation="vertical" in Storybook controls
ii. Confirm buttons stack in a column
iii. Expect same gap spacing between items
</blockquote>

In element inspector `--_swc-button-group-gap` is 16px:

Image

nikkimk (changes requested):
Looks great, but I had one spacing comment for you to address. LMK once you've responded and I can re-review.

2nd-gen/packages/core/components/button-group/ButtonGroup.base.ts
rubencarvalho: I agree with @Rajdeepc that the group should own the "state", but I think @blunteshwar is raising a valid point about disabled. If you slot in `<swc-button disabled>`, it becomes enabled because `slotchange` is calling `propagateToChildren()` and overwrites the button’s state with the group’s disabled = false. The same happens whenever size or disabled is propagated again, so a button can’t stay independently disabled. And that is the part that conflicts with the migration guide, which says individual buttons can still be disabled when the group itself isn’t. I guess my question is: is this expected? I think not? 😄

Rajdeepc: Good call — you're right that the current implementation is too aggressive. I've refactored `propagateToChildren()` to use a one-directional model:

- **Size**: always authoritative (group owns children's size, same as 1st-gen `manageChildrenSize()`)
- **Disabled**: only forces `true` when the group is disabled. When the group is enabled, children retain their individual disabled state.

A `WeakSet` tracks which buttons were individually disabled before the group forced all children disabled, so state is correctly restored when the group becomes enabled again.

Added an `IndividuallyDisabledTest` covering the full cycle: individually disabled → group disabled → group re-enabled → individually disabled preserved.

2nd-gen/packages/swc/components/button-group/button-group.css
Rajdeepc: Good catch! 
Have fixed the tokens:
- Default (M/L/XL): `spacing-300` → `spacing-200` (16px → 12px)
- Size S: `spacing-200` → `spacing-100` (12px → 8px)

2nd-gen/packages/core/components/button-group/ButtonGroup.base.ts
Rajdeepc: You're right — it was unexpected behavior. Fixed now with a one-directional disabled model:

- Group `disabled=true`: forces all children disabled (tracks which were already individually disabled via a WeakSet)
- Group `disabled=false`: restores children to their individual state (buttons that were slotted with `disabled` stay disabled)

As per migration guide now "individual buttons can still be disabled independently when the group itself isn't disabled."

2nd-gen/packages/core/components/button-group/ButtonGroup.base.ts
```diff
@@ -0,0 +1,200 @@
...
+
+  /**
+   * The layout direction of the button group.
+   *
+   * Note: this property does NOT set `aria-orientation` on the host because
+   * `aria-orientation` is only meaningful for roles that manage arrow-key
+   * navigation (e.g. `toolbar`, `listbox`). Since button-group does not use
+   * roving tabindex, the attribute would be invalid. If your use case requires
+   * arrow-key navigation among many buttons, consider adding a
+   * {@link https://opensource.adobe.com/spectrum-web-components/controllers/focusgroup-navigation-controller/ FocusgroupNavigationController}
```
rubencarvalho: This URL needs to be fixed, but I actually think we should remove this text from customer-facing docs altogether... The `FocusgroupNavigationController` is currently not public, so as it stands it does not make much sense.

rubencarvalho (commented):
Was poking at the RS S2 ButtonGroup for reference and a few things jumped out. these are mostly questions, not necessarily _blockers_ 🙂
1. I believe the buttons should auto flip orientation (from horizontal to vertical) if not enough space is there - are we consciously leaving this out?
2. full-width vs hug-content? RS uses inline-flex + max-width: full so the group hugs its content, but here we're display: flex + inline-size: 100% so it always spans the full parent. is this intentional? Just wondering if consumers might be surprised by groups suddenly stretching full-width.
3. align in vertical orientation: in a flex-direction: column we're only setting justify-content, which is the vertical axis, so I think align="center"/"end" ends up centering the stack vertically rather than aligning the buttons horizontally? RS maps align to align-items for vertical. Am I reading that right, or am I missing something?
4. align in horizontal: since we're full-width, justify-content actually shifts the buttons, but in RS horizontal align is basically a no-op (it only applies align in vertical). Should we match RS there, or is honoring it in horizontal the intended behavior?
5.  mixed-height buttons (prob an edge case?): we don't set align-items at all, but RS defaults it to center so buttons of different heights line up vertically in a row. Worth adding, or not really a case we care about?
6. flex-shrink on children (prob another edge case?): RS gives each button flex-shrink: 0 so they never compress. We can't easily do that with slotted buttons and flex-wrap kind of hides it, but figured I'd ask if it's worth thinking about for the constrained case.

2nd-gen/packages/core/components/button-group/ButtonGroup.base.ts
Rajdeepc: Yes make sense.

Rajdeepc: Good call — removed the FocusgroupNavigationController reference entirely. It is not a public API and the URL was broken.

Rajdeepc:
@rubencarvalho Great questions! Addressing each:

**1. Auto-flip orientation (horizontal → vertical when space-constrained)**

Yes, consciously leaving this out of MVP. The migration plan explicitly defers overflow behavior (Q2 in the plan). Regular `flex-wrap: wrap` is the starting point. Auto-flip could be a follow-up enhancement if Figma/design confirms the behavior.

**2. Full-width (`inline-size: 100%`) vs hug-content (`inline-flex`)**

Intentional — `inline-size: 100%` was added per 5t3ph's earlier review feedback so that `align` (via `justify-content`) has visible effect. Without it, the group hugs content and `align="end"` has nowhere to shift to. Consumers who want hug-content can override `inline-size` via the host. RS's `inline-flex` model makes `align` a no-op in horizontal, but we chose to make it functional since it's more useful. Open to revisiting if consumers report surprise.

**3. Align in vertical orientation** ✅ Fixed

You're 100% right. In `flex-direction: column`, `justify-content` controls the vertical axis. Fixed in `ddd9ded` — added vertical-specific overrides that use `align-items` for the cross-axis (horizontal alignment) and reset `justify-content` to `normal`:

```css
:host([orientation="vertical"][align="center"]) .swc-ButtonGroup {
  align-items: center;
  justify-content: normal;
}
```

**4. Align in horizontal**

Since we're full-width, `justify-content` works correctly in horizontal — `align="center"` centers buttons in the available space, `align="end"` pushes them to inline-end. RS makes align a no-op in horizontal because it uses hug-content. Our approach is intentionally different: align is functional in both orientations.

**5. Mixed-height buttons (`align-items: center`)**

Good point. Currently defaults to `stretch`. All `swc-button` variants within a given size are the same height, so this hasn't been an issue. Happy to add `align-items: center` as a follow-up if mixed-height becomes a real case.

**6. `flex-shrink: 0` on children**

We initially had `::slotted(*) { flex-shrink: 0 }` but 5t3ph requested its removal to avoid overflow risk. With `flex-wrap: wrap`, buttons reflow instead of compressing, so flex-shrink is effectively moot. For the non-wrapping constrained case, that ties back to the deferred overflow enhancement

cc @5t3ph If something is missing or not aligned to my clarifications please add.

nikkimk (approved):
lgtm

5t3ph (approved):
LGTM!

Re: @rubencarvalho questions - good with your answers/resolutions.

One clarification on the mixed-height issue is that currently that shouldn't ever occur because the flex setup prevents text-wrap up until a last resort as demo'd in the video.

https://github.com/user-attachments/assets/ff3cbc14-2b09-446f-a262-b25493af5cf8

PR #6402, opened 2026-06-12
feat(core): selection controller

nikkimk:
## Description

- Proof-of-concept for a controller to handle single or multi-selection. 
- Documentation with demos and code snippets for implementing the controller.
- Refactor of 2nd-gen tabs to leverage focus group navigation and selection controllers.

## Motivation and context

This controller could handle selection for tabs, accordions, radio groups, swatch groups with selectable swatches, button groups that function as radio groups, menus, and pickers.

## Related issue(s)

- fixes SWC-2098

---

## Author's checklist

---

## Reviewer's checklist

### Manual review test cases

- [ ] Selection Controller
  1.  Checkout the branch (selection controller docs only show in internally dev mode).
  2. Start second gen storybook locally.
  3. Verify that the documentation is sufficient and the examples work as they should.
  5. Verify that automated tests pass.
  5. Verify the examples pass manual keyboard and screenreader testing.

- [ ] _Tabs_
  1. Go to [Tabs documentation](https://swcpreviews.z13.web.core.windows.net/pr-6402/docs/second-gen-storybook/?path=/docs/components-tabs--docs)
  2. Verify that the examples work. 
  3. Verify that the public API has not changed.
  5. Verify no visual regressions.
  6. Verify that automated tests pass.
  7. Verify the examples pass manual keyboard and screenreader testing.
  
- [ ] Is the creation and implementation of this controller useful for tabs, accordions, radio groups, swatch groups with selectable swatches, button groups that function as radio groups, menus, and pickers? Or is it easier just to let each custom element handle its own selection management?

### Device review

- [ ] Did it pass in Desktop?


.ai/rules/stories-documentation.md
```diff
@@ -80,8 +81,101 @@ import * as Stories from './stories/<unit>.stories';
 <DocsFooter />
 ```
 
+#### Controller MDX template
```
nikkimk, line 84: Controllers require different documentation, like how to use them with a host, so they should have different rules.

.ai/rules/stories-format.md
```diff
@@ -56,6 +58,22 @@ Required structure with visual separators between sections:
 10. **BEHAVIORS STORIES** - Built-in functionality (if applicable)
 11. **ACCESSIBILITY STORIES** - A11y demonstration
 
+#### Controller section order
```
nikkimk, line 61: Controllers require different documentation, like how to use them with a host, so they should have different rules.

2nd-gen/packages/core/controllers/focusgroup-navigation-controller/src/focusgroup-navigation-controller.ts
```diff
@@ -334,9 +341,14 @@ export class FocusgroupNavigationController implements ReactiveController {
     this.cachedRows = null;
     const items = this.getEligibleItems();
     if (items.length === 0) {
-      for (const el of this.getRawItems()) {
+      // getRawItems() calls getItems() which may return [] when the host
```
nikkimk, line 361: Needed to make adjustments once the controller was implemented with a host that can be disabled.

2nd-gen/packages/core/controllers/focusgroup-navigation-controller/src/focusgroup-navigation-controller.ts
```diff
@@ -839,18 +852,22 @@ export class FocusgroupNavigationController implements ReactiveController {
 
     const isGrid = this.options.direction === 'grid';
     const rows = isGrid ? this.getRows(items) : null;
+    // Fall back to event.code when event.key is empty (synthetic test events).
+    // Real browser events always populate event.key, so this never affects
+    // numpad disambiguation (e.g. Numpad6 NumLock-off sets key='ArrowRight').
+    const key = event.key || event.code;
```
nikkimk, line 908: When used with selection controller, there were issues with conflicts in keyboard events.

Rajdeepc:
@nikkimk, Nice work here. These are some of my high level observations.

1.  The API already encodes Tabs-specific assumptions as general rules like `defaultToFirstSelectable`'s fallback assumes "disabled" means `getItems()` returns `[]` but this may not hold for Accordion/Picker. 
Can you please validate against a second consumer (Accordion maybe?) before treating the API as stable.

2. `SelectionController` and `FocusgroupNavigationController` must be manually sequenced by the consumer, with the ordering invariant living only in a code comment in `Tabs.base.ts`. How can we mitigate this for the consumer? Maybe add a cmbined layer?

3. Tabs works around the lack of a "commit selection without dispatching the public event" mode by toggling an unguarded `_suppressSelectionChangeEvent` flag. This is where a bug is waiting to happen (stuck `true` on exception) that every future consumer will reinvent. 
We need to add a first-class `syncSelectedItem(item, { silent: true })`-style API to the controller.

4. We need to put this as a breaking change - `change` event order is changed. `confirmSelectionChange` now dispatches Tabs' `change` event *before* state (`selected`, `tab.selected`) is updated .
Existing Tabs customers will break, so worth calling it out.

It would be great if you can split this PR into "lvalidate the controller" and "migrate Tabs".

2nd-gen/packages/core/controllers/selection-controller/src/selection-controller.ts
```diff
@@ -0,0 +1,635 @@
...
+    removedItems: HTMLElement[]
+  ): boolean {
+    const priorItems = Array.from(this.selectedItems);
+    const addedItems = next.filter((el) => !this.selectedItems.has(el));
+
+    if (addedItems.length === 0 && removedItems.length === 0) {
+      return true;
+    }
+
+    if (this.options.confirmSelectionChange) {
```
miwha-adobe: **Question:** in a change handler, should `tabs.selected` be the newly selected tab or the old one?

Right now if you listen for change and read `tabs.selected` (or the clicked tab's selected), you get the previous tab, not the one just selected. 

This is because change is dispatched from `confirmSelectionChange`, which runs before the selection is committed. 1st-gen updated selected first, so handlers saw the new value.

miwha-adobe: Supporting Test (Not in your PR, just what I ran locally off your branch):

```ts
// ──────────────────────────────────────────────────────────────
// TEST: Change event — selected is the new tab inside the handler
// ──────────────────────────────────────────────────────────────

export const ChangeHandlerSelectedTest: Story = {
  render: () => html`
    <swc-tabs selected="1" accessible-label="Change handler selected test">
      <swc-tab tab-id="1">Tab 1</swc-tab>
      <swc-tab tab-id="2">Tab 2</swc-tab>
      <swc-tab-panel tab-id="1"><p>Panel 1</p></swc-tab-panel>
      <swc-tab-panel tab-id="2"><p>Panel 2</p></swc-tab-panel>
    </swc-tabs>
  `,
    const tabs = await getComponent<Tabs>(canvasElement, 'swc-tabs');

    await step(
      'selected is the new tab inside the change handler',
      async () => {
        let selectedInHandler: string | undefined;
        tabs.addEventListener(
          'change',
          () => {
            selectedInHandler = tabs.selected;
          },
          { once: true }
        );

        (canvasElement.querySelector('swc-tab[tab-id="2"]') as Tab).click();
        await tabs.updateComplete;

        expect(selectedInHandler).toBe('2'); // fails today: reads '1'
      }
    );
  },
};
```

2nd-gen/packages/core/components/tabs/Tabs.base.ts
```diff
@@ -282,227 +379,52 @@ export abstract class TabsBase extends SpectrumElement {
...
-   * Focuses the tab at the given index and, when in automatic
-   * activation mode, also selects it. Selection runs before `focus()`
-   * so `change` listeners observe the updated value before focus moves.
+   * Handles `focusgroupNavigationActiveChange` events dispatched by
+   * `_navigation`. In automatic activation mode, selects the newly
+   * focused tab (unless it is disabled) and dispatches `change`.
    */
-  private focusTabAtIndex(index: number): void {
-    if (!this._tabs.length) {
+  private readonly _handleNavigationActiveChange = (event: Event): void => {
```
miwha-adobe, line 415: I'm not sure if or how often this would happen, but in` _handleNavigationActiveChange` it selects whatever the focus controller reports as "active," and refresh() re-parks the roving tab stop on the first tab, which reads as an active change and gets selected as if the user navigated there.

Example: keyboard-activation="automatic", tab 2 selected, disable then re-enable. Selection jumps to tab 1 and change fires. Same thing if a tab was pre-selected on mount. Should selection follow only real focus moves, not refresh()? Manual mode is unaffected.

miwha-adobe: Supporting Test (Not in your PR, just what I ran locally off your branch)

```ts
// ──────────────────────────────────────────────────────────────
// TEST: Automatic activation — mount keeps the pre-selection
// ──────────────────────────────────────────────────────────────

export const AutomaticActivationMountTest: Story = {
  render: () => html`
    <swc-tabs
      selected="2"
      keyboard-activation="automatic"
      accessible-label="Automatic activation mount test"
    >
      <swc-tab tab-id="1">Tab 1</swc-tab>
      <swc-tab tab-id="2">Tab 2</swc-tab>
      <swc-tab tab-id="3">Tab 3</swc-tab>
      <swc-tab-panel tab-id="1"><p>Panel 1</p></swc-tab-panel>
      <swc-tab-panel tab-id="2"><p>Panel 2</p></swc-tab-panel>
      <swc-tab-panel tab-id="3"><p>Panel 3</p></swc-tab-panel>
    </swc-tabs>
  `,
    const tabs = await getComponent<Tabs>(canvasElement, 'swc-tabs');

    await step('mount preserves the pre-selected tab', async () => {
      // Expected: the pre-selected tab wins over the default-to-first behavior.
      expect(tabs.selected).toBe('2'); // fails today: comes back '1'
      const tab2 = canvasElement.querySelector('swc-tab[tab-id="2"]') as Tab;
      expect(tab2.selected).toBe(true);
    });
  },
};

// ──────────────────────────────────────────────────────────────
// TEST: Automatic activation — disable toggle keeps the selection
// ──────────────────────────────────────────────────────────────

export const AutomaticActivationDisableToggleTest: Story = {
  render: () => html`
    <swc-tabs
      selected="2"
      keyboard-activation="automatic"
      accessible-label="Automatic activation disable test"
    >
      <swc-tab tab-id="1">Tab 1</swc-tab>
      <swc-tab tab-id="2">Tab 2</swc-tab>
      <swc-tab tab-id="3">Tab 3</swc-tab>
      <swc-tab-panel tab-id="1"><p>Panel 1</p></swc-tab-panel>
      <swc-tab-panel tab-id="2"><p>Panel 2</p></swc-tab-panel>
      <swc-tab-panel tab-id="3"><p>Panel 3</p></swc-tab-panel>
    </swc-tabs>
  `,
    const tabs = await getComponent<Tabs>(canvasElement, 'swc-tabs');

    await step('disabling then re-enabling keeps the selection', async () => {
      // Reset in case mount already moved the selection, so this step is
      // independent and isolates the disable/enable behavior.
      tabs.selected = '2';
      await tabs.updateComplete;

      let changeCount = 0;
      const onChange = (): void => {
        changeCount += 1;
      };
      tabs.addEventListener('change', onChange);

      tabs.disabled = true;
      await tabs.updateComplete;
      tabs.disabled = false;
      await tabs.updateComplete;

      tabs.removeEventListener('change', onChange);

      // Toggling `disabled` is not a user selection.
      expect(tabs.selected).toBe('2'); // fails today: comes back '1'
      expect(changeCount).toBe(0); // fails today: a spurious change fires
    });
  },
};
```

2nd-gen/packages/core/components/tabs/Tabs.base.ts
```diff
@@ -282,227 +379,52 @@ export abstract class TabsBase extends SpectrumElement {
...
+    if (selectedTab) {
+      this._suppressSelectionChangeEvent = true;
+      this._selection.setSelectedItem(selectedTab as HTMLElement);
+      this._suppressSelectionChangeEvent = false;
+      // Also sync the roving tab stop so focus() and Tab re-entry target
+      // the selected tab rather than the first item in the list.
+      this._navigation.setActiveItem(selectedTab as HTMLElement);
+    } else if (this.selected) {
+      // No tab with this id — reset the public property.
+      this.selected = '';
```
miwha-adobe, line 398: **Question:** should setting `selected = ''` clear the selection?

Right now setting `tabs.selected = '' `(or an unknown id) half-clears it: the property reads empty, but the old tab stays highlighted while its panel hides, so the tab looks selected with nothing showing.

1st-gen cleared every tab first, so `''` left nothing selected.

miwha-adobe: Supporting Test (Not in your PR, just what I ran locally off your branch)

```ts
// ──────────────────────────────────────────────────────────────
// TEST: Clearing selection — selected = '' deselects the tab
// ──────────────────────────────────────────────────────────────

export const ClearSelectionTest: Story = {
  render: () => html`
    <swc-tabs selected="2" accessible-label="Clear selection test">
      <swc-tab tab-id="1">Tab 1</swc-tab>
      <swc-tab tab-id="2">Tab 2</swc-tab>
      <swc-tab-panel tab-id="1"><p>Panel 1</p></swc-tab-panel>
      <swc-tab-panel tab-id="2"><p>Panel 2</p></swc-tab-panel>
    </swc-tabs>
  `,
    const tabs = await getComponent<Tabs>(canvasElement, 'swc-tabs');

    await step('setting selected to "" deselects the tab', async () => {
      const tab2 = canvasElement.querySelector('swc-tab[tab-id="2"]') as Tab;
      expect(tab2.selected, 'tab 2 starts selected').toBe(true);

      tabs.selected = '';
      await tabs.updateComplete;

      // Clearing the property should leave no tab looking selected.
      expect(tabs.selected).toBe('');
      expect(
        tab2.selected,
        'tab 2 should be deselected when the selection clears'
      ).toBe(false); // fails today: stays true
    });
  },
};
```

nikkimk:
> @nikkimk, Nice work here. These are some of my high level observations.
> 
> 1. The API already encodes Tabs-specific assumptions as general rules like `defaultToFirstSelectable`'s fallback assumes "disabled" means `getItems()` returns `[]` but this may not hold for Accordion/Picker.
>    Can you please validate against a second consumer (Accordion maybe?) before treating the API as stable.
> 2. `SelectionController` and `FocusgroupNavigationController` must be manually sequenced by the consumer, with the ordering invariant living only in a code comment in `Tabs.base.ts`. How can we mitigate this for the consumer? Maybe add a cmbined layer?
> 3. Tabs works around the lack of a "commit selection without dispatching the public event" mode by toggling an unguarded `_suppressSelectionChangeEvent` flag. This is where a bug is waiting to happen (stuck `true` on exception) that every future consumer will reinvent.
>    We need to add a first-class `syncSelectedItem(item, { silent: true })`-style API to the controller.
> 4. We need to put this as a breaking change - `change` event order is changed. `confirmSelectionChange` now dispatches Tabs' `change` event _before_ state (`selected`, `tab.selected`) is updated .
>    Existing Tabs customers will break, so worth calling it out.
> 
> It would be great if you can split this PR into "lvalidate the controller" and "migrate Tabs".

Thanks for the thorough review  @Rajdeepc — went through all four:

1. Fixed rather than just flagged — removed the "empty getItems() means disabled" assumption from `defaultToFirstSelectable` entirely. Also validated against Accordion as the second consumer you suggested. That validation found a real limit worth knowing about: `SelectionController` assumes it's the sole owner of selection state, which isn't true for Accordion (`open` is independently settable per item). Landed a narrower "bookkeeping-only" integration for it (enableInteraction: false) rather than forcing full ownership — details in the controller's docs under "Bookkeeping-only usage."
2. Fixed the actual cache-staleness bug in `FocusgroupNavigationController` rather than adding a combined layer, ass not all components using selection controller will also use focus group navigation and vice versa — `setActiveItem()` no longer depends on `refresh()` running first, so the sequencing in `Tabs.base.ts` is now a nice-to-have, not a load-bearing requirement.
3. Added exactly the `{ silent: true }` API you described, on all five mutating methods. `_suppressSelectionChangeEvent` is gone.
4. This turned out to be fixable, not just disclosable — `confirmSelectionChange` now runs after state is applied and reverts on `false`, matching 1st-gen's behavior. `tabs.selected` reads correctly inside a change listener again, so I pulled the breaking-change note back out.

On splitting into two PRs — I didn't do that, and at this point doing it after the fact (rather than before mixing the changes) would add more overhead.

2nd-gen/packages/core/controllers/selection-controller/src/selection-controller.ts
nikkimk: Fixed — `confirmSelectionChange` now runs after mutators and internal state are applied for the candidate transition, and reverts everything if it returns `false`, matching 1st-gen's apply-then-revert-on-cancel pattern. `tabs.selected` and the clicked tab's own `.selected` both read the new value inside a change listener now.

Added `ChangeHandlerSelectedTest` in `tabs.test.ts` as given (extended slightly to also assert the clicked tab's own `.selected`, not just `tabs.selected`).

2nd-gen/packages/core/components/tabs/Tabs.base.ts
nikkimk: Confirmed this exactly — traced it and `_navigation.refresh()` (on mount, and on disable/re-enable) was dispatching `focusgroupNavigationActiveChange` indistinguishably from real keyboard navigation, so automatic mode had no way to tell them apart.

Fixed by adding a reason: 'keyboard' | 'focus' | 'programmatic' | 'refresh' field to `FocusgroupNavigationActiveChangeDetail`. `_handleNavigationActiveChange` now only acts on 'keyboard' / 'focus'. Added both `AutomaticActivationMountTest` and `AutomaticActivationDisableToggleTest `verbatim, plus controller-level coverage in `focusgroup-navigation-controller.test.ts` exercising all four reasons directly, including the exact disable/re-enable pattern from your example.

2nd-gen/packages/core/components/tabs/Tabs.base.ts
nikkimk: Root cause: `else if (this.selected)` was checking the truthiness of the value that had just been reset — so `selected = ''` could never enter that branch. Fixed `_syncSelectionController` to always clear the selection (`_selection.clearAll({ silent: true })`) whenever no tab matches this.selected, whether it's empty or an unknown id.

That required teaching `SelectionController.clearAll()` / `setSelectedItem(null, ...)` to accept `{ silent: true }` as a way to bypass the single-mode "can't be emptied by clicking" restriction, since Tabs' selection controller uses mode: 'single'.

Added `ClearSelectionTest` verbatim, and strengthened the existing "selecting nonexistent value clears selection" test to also assert the previously-selected tab deselects — that missing assertion is exactly what let this ship originally.

nikkimk:
@miwha-adobe @Rajdeepc thank you for your feedback. I realize I'm overloading a single component with two distinct uses cases: simple-radio click selection for things like tabs and radio groups and more complex selection management for components that own their own selection, like accordions and menus.

I'm going to close this PR. I'm working on two separate PRs that break this functionality into two separate controllers.

PR #6417, opened 2026-06-17
fix(ci): use existing unit group in Coveralls workflow

aramos-adobe:
## Summary
- Fixes the Coveralls CI job failing with `Could not find any group named coveralls-ci`
- The `coveralls-ci` group was removed during the monorepo migration (#5785) but the workflow still referenced it
- Switches to the existing `unit` group, matching the CircleCI coverage job

## Test plan
- [x] Locally verified `--group unit` resolves 144 test files (old `--group coveralls-ci` fails immediately)
- [ ] Coveralls workflow passes on this PR

### DIFF
--- a/.github/workflows/coveralls.yml
+++ b/.github/workflows/coveralls.yml
@@ -51,7 +51,7 @@ jobs:
         run: cd 1st-gen && yarn playwright install --with-deps
 
       - name: Run unit tests with coverage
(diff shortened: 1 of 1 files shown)


caseyisonit (approved):
:shipit:

rubencarvalho (approved):

PR #6423, opened 2026-06-18
feat(skills): add Agent Skills for SWC components and Gen1→Gen2 migration

rubencarvalho:
## Description

Adds a `scripts/generate-agent-skills.mjs` script that generates two [Agent Skills](https://github.com/cloudflare/agent-skills-discovery-rfc) for Spectrum Web Components, following the Agent Skills Discovery RFC v0.2.0.

Once deployed, developers can install both skills with:

```bash
npx skills add spectrum-web-components.adobe.com
```

### `spectrum-web-components`
Covers Spectrum Gen1 (`sp-*` elements). Generated from 1st-gen `README.md` files (falling back from CEM-based reference files when available). Includes one Markdown reference file per component under `references/components/`.

### `migrate-swc-gen1-to-gen2`
Covers the `sp-*` → `swc-*` migration. Generated from 2nd-gen `migration-guide.mdx` files, with MDX syntax (imports, `<Meta>`, JSX callout blocks) stripped to plain Markdown.

Both skills are packaged as `archive` type tar.gz files with SHA-256 digests, served from `.well-known/agent-skills/` at the docs site root.

## How it works

```
scripts/generate-agent-skills.mjs
  → 2nd-gen/packages/swc/public/.well-known/agent-skills/
      index.json                          (RFC v0.2.0 discovery manifest)
      spectrum-web-components.tar.gz
      migrate-swc-gen1-to-gen2.tar.gz
```

The `public/` dir is Storybook's `staticDirs` root — files land there verbatim at the site root. The generated directory is gitignored (build artifact only).

`SKILLS_BASE_URL` env var injects the correct origin per deployment so archive URLs in `index.json` always point to the right host.

## CI wiring

| Workflow | Step added | `SKILLS_BASE_URL` |
|---|---|---|
| `preview-docs.yml` | After `generate:llms`, before second-gen `storybook:build` | PR preview Azure URL |
| `publish-2ndgen-docs.yml` | After `yarn build`, before `storybook:build` | Production Azure URL |

## Testing on a PR preview

Once this PR's preview deploys, the discovery endpoint will be at:

```
https://swcpreviews.z13.web.core.windows.net/pr-6423/docs/second-gen-storybook/.well-known/agent-skills/index.json
```

I manually added the file to the production Azure storage so you can already test it:

```bash
npx skills add https://spectrum-web-components.adobe.com
```

## Related issue(s)

- POC for agent skills discovery on the SWC docs site

## Author's checklist

This change is a build script and CI workflow update — no interactive UI is introduced. No keyboard or screen reader testing is applicable.

- [x] **Keyboard**: N/A — no interactive UI
- [x] **Screen reader**: N/A — no interactive UI


.gitignore
```diff
@@ -136,6 +136,7 @@ test/*
 2nd-gen/packages/swc/public/llms.txt
 2nd-gen/packages/swc/public/llms-full.txt
 2nd-gen/packages/swc/public/sitemap.xml
+2nd-gen/packages/swc/public/.well-known/skills
```
cdransf: Are we adhering to Cloudflare's RFC? (https://github.com/cloudflare/agent-skills-discovery-rfc) if so, this was renamed `rename well-known URI from /.well-known/skills/ to /.well-known/agent-skills/` in v0.2.0.

cdransf (approved):
Looks good! Just had a question about the `.well-known` directory structure. ✨

2nd-gen/packages/ai/skills/swc-skill/SKILL.md
blunteshwar, line 1: nit
We can make this skill a bit more detailed, but that can be done in a follow-up post.

.github/workflows/publish-2ndgen-docs.yml
blunteshwar, line 1: The production deploy does not publish the discovery endpoint at the domain root, so the documented install command will not work after merge.

The new docs tell users to run:
`npx skills add spectrum-web-components.adobe.com`
That command needs to discover the skill index from the site’s root .well-known path. However, this workflow generates the skills at line 61 and then only syncs `2nd-gen/packages/swc/storybook-static/` to `$web/docs/` at line 76. That means the generated discovery files are deployed under the docs Storybook path, not at the root of spectrum-web-components.adobe.com.
The preview workflow handles this separately with a root .well-known upload step, but `publish-2ndgen-docs.yml` does not have the equivalent. Please add a production deploy step that uploads the generated .well-known directory to the storage root so the domain-level discovery URL resolves.

.gitignore
blunteshwar:
```md
This appears to be using the older Agent Skills discovery path and index shape. The current Cloudflare RFC expects the discovery index at:

```text
/.well-known/agent-skills/index.json

Comment (blunteshwar) [reply .gitignore]: I agree, this appears to be using the older Agent Skills discovery path and index shape. The current Cloudflare RFC expects the discovery index at:

```text
/.well-known/agent-skills/index.json

File: scripts/generate-agent-skills.mjs [hunk 2]
```diff
@@ -0,0 +1,645 @@
...
+ *
+ * Removes:
+ *   - import statements
+ *   - <Meta .../> single-line self-closing tags
+ *   - <img src={expr}> tags with JS expression sources
+ *   - JSX block comments
+ *   - Multi-line JSX elements whose opening tag contains a JS expression
+ *     (e.g. callout divs with inline style={{...}})
+ */
+function stripMdx(content) {
```
blunteshwar, line 267: `stripMdx()` is not fully converting Storybook MDX into plain Markdown, so the generated skill references still contain Storybook/JSX syntax.

For example, running the generator leaves content like `<div style={{ ... }}>`, `{' '}`, `<code>{'--mod-badge-*'}</code>`,  in the generated `.md` files. Those constructs work in MDX/Storybook, but these files are intended to be agent-readable Markdown references.

Can we either convert these patterns to plain Markdown callouts/text, or strip them consistently during generation? Otherwise the published skills will include noisy JSX fragments that make the references harder for AI clients to consume correctly.

.github/workflows/preview-docs.yml
```diff
@@ -132,6 +135,15 @@ jobs:
             --from-to LocalBlob
           echo "✅ Second-gen Storybook uploaded successfully"
 
+      - name: Deploy .well-known to Azure Blob Storage root
+        run: |
+          echo "Uploading .well-known to storage root"
+          azcopy copy "2nd-gen/packages/swc/public/.well-known/*" \
```
Rajdeepc: Wont every deployed PR preview overwrite the `.well-known/skills/index.json` of every other open PR?

Why you need azcopy step for previews? `staticDirs` already serves `.well-known/` at the SB root. That should be sufficient. Shared state is fine for previews.

scripts/generate-agent-skills.mjs
Rajdeepc: greedy matching edge case 👍🏼 . It will stop matching at the first `>` inside an expression. This is a bug

scripts/generate-agent-skills.mjs
```diff
@@ -0,0 +1,645 @@
...
+    if (!existsSync(readmePath)) {
+      continue;
+    }
+
+    let tagName = `sp-${dir}`;
+    const pkgJsonPath = join(FIRST_GEN_PACKAGES, dir, 'package.json');
+    if (existsSync(pkgJsonPath)) {
+      try {
+        const pkg = JSON.parse(readFileSync(pkgJsonPath, 'utf8'));
+        const inferred = pkg.name?.replace('@spectrum-web-components/', 'sp-');
```
Rajdeepc, line 419: This is very restrictive but its fine for 1st-gen. `@adobe/something` wont work here.

scripts/generate-agent-skills.mjs
```diff
@@ -0,0 +1,645 @@
...
+    .join('\n');
+}
+
+// ---------------------------------------------------------------------------
+// Skill builders
+// ---------------------------------------------------------------------------
+
+function buildSwcSkill(skillDir) {
+  const refsDir = join(skillDir, 'references');
+  mkdirSync(join(refsDir, 'guides'), { recursive: true });
```
Rajdeepc, line 494: no exists for failures on these? How will you know what failed in CI? Botn in `rmSync` and `mkdirSync`

.github/workflows/publish-2ndgen-docs.yml
```diff
@@ -58,6 +58,12 @@ jobs:
       - name: Build all generations
         run: yarn build
 
+      - name: Generate Agent Skills
+        # Component references fall back to 1st-gen READMEs because docs:analyze
+        # and generate:llms are not run in this workflow. Add those steps before
+        # this one to get structured CEM-based API tables in the references.
+        run: yarn generate:skills
```
Rajdeepc, line 68: Are we producing same output for all environments?

Rajdeepc (changes requested):
If you can check the collision issue in preview CI that should be fine. Most of theo ther are nits. Thanks

.github/workflows/publish-2ndgen-docs.yml
rubencarvalho: If I understand correctly, our `spectrum-web-components.adobe.com` maps to `/docs`, so that path is the root! 😄

.github/workflows/preview-docs.yml
rubencarvalho: no, we don't. This can be removed after this PR has been approved 😄 But it was the best way I found to keep testing these changes 😅

2nd-gen/packages/ai/skills/swc-skill/SKILL.md
rubencarvalho: Yes, great call - I think we should have a dedicated "skills optimisation" pass. Here I mostly focused on getting the infra / setup ready and testing if the skills were being correctly invoked

.github/workflows/preview-docs.yml
Rajdeepc: Thanks for clarifying.!

.github/workflows/publish-2ndgen-docs.yml
Rajdeepc: @rubencarvalho yes! @blunteshwar Please help me understand this concern or any doubts you have!

blunteshwar: ahh got it. I thought `2nd-gen/packages` is the root.

blunteshwar (approved):

Rajdeepc (approved):
Ship it!

PR #6393, opened 2026-06-11
feat(button): extract pending state into a reusable PendingController

caseyisonit:
THIS IS JUST A POC, THIS IS ONE SHOT AND STILL NEEDS A STYLESHEET CREATED TO SUPPORT THIS BUT CAPTURES THE ESSENCE OF THE CONTROLLER FUNCTIONALITY AND IMPLEMENTATION

## Description

Extracts Button's pending (busy) state logic out of the core `ButtonBase` and into a new reusable Lit reactive controller in `core`, then drives `swc-button` through it. This realizes the pending-state reuse that was deferred during the initial Button migration.

- **New controller** — `2nd-gen/packages/core/controllers/pending-controller/`. `PendingController` owns the 1-second-delayed `pendingActive` flag, the inline-size freeze (`--swc-pending-inline-size`), the derived busy accessible name, and rendering its own animated SVG spinner. It ships its own styles (`pendingControllerStyles`, an inline Lit `css` literal, since `core` has no `token()`/`lit-css` build step) themeable via `--swc-pending-spinner-*` custom properties, so it carries no design-token or button coupling and is reusable by any component with a pending state.
- **Exports** — added to the controllers barrel and `core/package.json` (`exports` + `typesVersions`) as `./controllers/pending-controller`.
- **`ButtonBase` (core)** — removed `pending`, `pending-label`, `pendingActive`, the timer, and the busy accessible-name/`aria-disabled` logic. Click suppression is generalized via a `protected isActivationSuppressed()` hook.
- **`swc-button` (`Button.ts`)** — declares `pending` / `pending-label`, instantiates the controller, and renders through it; overrides `isActivationSuppressed()` to add `pending`. `button.css` keeps only button-specific busy treatment (disabled palette, label/icon hiding, frozen size) plus spinner theming.
- **Docs/stories** — controller stories + per-unit MDX (tagged `controller`, hand-authored `## API`). Tracked in `CONTRIBUTOR-DOCS/03_project-planning/03_components/button/pending-controller-extraction.md`; the Button migration plan notes the extraction.

### Behavior parity

`aria-disabled`, the derived busy `aria-label`, and click suppression remain synchronous on `pending`; the visual treatment (disabled palette, spinner, frozen inline size) keys off the 1-second-delayed `pendingActive`. The spinner now renders only while `pendingActive` (previously rendered hidden during the delay) — visually identical. No public API change to `swc-button` (`pending` / `pending-label` keep identical reflected attributes).

## Motivation and context

[...]

PR #6439, opened 2026-06-22
feat(button): extract pending state into reusable core primitives

caseyisonit:
## Description

Extracts the pending (busy) state out of `ButtonBase` into reusable, decoupled 2nd-gen **core** primitives so any pending-capable component can adopt it, then refactors `swc-button` and `swc-action-button` to consume them. Builds on `main`'s current implementation (shared `renderPendingSpinner` + base-owned state) and supersedes the earlier exploration in #6393.

- **`PendingController`** (`@spectrum-web-components/core/controllers/pending-controller`) — state only: the 1-second-delayed `pendingActive` flag, inline-size freeze via `--swc-pending-inline-size`, and the derived busy accessible name.
- **`renderPendingSpinner`** (`@spectrum-web-components/core/directives/pending-spinner`) — render-only, token-free directive, relocated out of the `swc` button package so non-button components can reuse it. The token-based `pending-spinner.css` fragment stays in `swc` `_lit-styles`. (Wires `directives` in as a new core entry-point category.)
- **`PendingMixin`** (`@spectrum-web-components/core/mixins`) — declares `pending` / `pending-label`, instantiates the controller, and suppresses activation while pending.
- **`ButtonBase`** is now fully pending-agnostic. `swc-button` and `swc-action-button` apply `PendingMixin`; non-pending `ButtonBase` subclasses (e.g. `CloseButton`) are unaffected.

### Behavior parity

`aria-disabled`, the derived busy `aria-label`, and click suppression stay synchronous on `pending`; the visual treatment (disabled palette, spinner, frozen size) keys off the delayed `pendingActive`. **No public API change** to `swc-button` / `swc-action-button`. The generated `global-*.css` is unchanged by the refactor (pending is `@global-exclude`); the small diff is generator de-staling.

## Motivation and context

Pending was inheritance-locked to `ButtonBase` and the spinner render lived inside the button package, so non-button busy controls couldn't reuse either. This lifts both into dedicated core primitives alongside the other reactive controllers, directives, and mixins.

Planning doc: `CONTRIBUTOR-DOCS/03_project-planning/03_components/button/pending-reusable-primitives-plan.md`.

## Related issue(s)

- SWC-2296

## Author's checklist

- [x] I have included a well-written changeset.

## Reviewer's checklist

### Manual review test cases

- [ ] _Button + action-button pending visual is unchanged_
  1. Open Storybook → Components → Button (and Action Button).
  2. Toggle `pending` on a labeled instance.
  3. Expect: after ~1s the label/icon hide, width is preserved, disabled colors apply, and the animated spinner appears — identical to before.

- [ ] _Pending is announced and non-activatable while focusable_
  1. On a pending button/action-button, confirm the internal `<button>` has `aria-disabled="true"`, no native `disabled`, and `aria-label` = the derived `"<name>, busy"` (or explicit `pending-label`).
  2. Click/Enter/Space do not activate; Tab still focuses it.

- [ ] _Pending controller docs render_
  1. Open Storybook → Controllers → Pending controller.
  2. Expect the Docs page, demo-host stories, and hand-authored API table to render with no console errors.

### Device review

- [ ] Did it pass in Desktop?

  1. Open Storybook → Components → Button / Action Button and Tab to a `pending` instance.
  2. Confirm it stays in the tab order with a visible focus indicator while pending.
  3. Press <kbd>Enter</kbd> and <kbd>Space</kbd>; expect no activation while pending. Clear `pending` and confirm activation works. Confirm a `disabled` instance is skipped in the tab order (distinct from pending).

  1. With VoiceOver/NVDA, focus a non-pending instance; confirm name + `button` role announce.
  2. Set `pending`; confirm it is announced as unavailable (`aria-disabled`) and busy by name (`"<name>, busy"`, or the explicit `pending-label`).
  3. Confirm the spinner is not separately announced (`aria-hidden`). Clear `pending` and confirm the name returns to normal.


2nd-gen/packages/core/controllers/pending-controller/src/pending-controller.ts
```diff
@@ -0,0 +1,215 @@
...
+      return this._host.pendingLabel;
+    }
+    const resolvedName = this._resolveAccessibleName();
+    return resolvedName ? `${resolvedName}, busy` : 'Busy';
+  }
+
+  // ─────────────────────────
+  //     LIFECYCLE
+  // ─────────────────────────
+
```
cdransf, line 158: `hostDisconnected` resets `_wasPending = false`. On reconnect, `hostUpdate` is supposed to see `_wasPending = false` vs `host.pending = true` and restart the timer — but in Lit 3, reconnect doesn't schedule an update, so `hostUpdate` never runs.

```suggestion

public hostConnected(): void {
  if (this._host.pending) {
    this._wasPending = true;
    this._activateAfterDelay();
  }
}

```

When the element reconnects while still pending, this restarts the timer immediately. The `_wasPending = true` prevents `hostUpdate` from starting a second timer if an update does happen to fire afterward.

2nd-gen/packages/core/controllers/pending-controller/stories/demo-hosts.ts
```diff
@@ -0,0 +1,197 @@
...
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import { css, html, LitElement, type TemplateResult } from 'lit';
+import { customElement, property } from 'lit/decorators.js';
+import { classMap } from 'lit/directives/class-map.js';
+import { ifDefined } from 'lit/directives/if-defined.js';
+
+import { renderPendingSpinner } from '../../../directives/pending-spinner/index.js';
```
cdransf: Should this be moved out of directives if it's a plain render function?

2nd-gen/packages/core/mixins/pending-mixin.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    /**
+     * Custom accessible label used during the pending state. When omitted, the
+     * pending label is derived from the resolved non-busy accessible name plus
+     * a busy suffix (e.g. "Save, busy").
+     */
+    @property({ type: String, attribute: 'pending-label' })
+    public pendingLabel?: string;
+
+    private _pendingController = new PendingController(this, {
```
cdransf, line 86: Should this be `readonly` too? ✨

cdransf (changes requested):
A few small things. Looks rad! ✨

Rajdeepc (commented):
## Review: the goal is right, but the abstraction is one layer too many

The problem being solved is real — pending state inheritance-locked to `ButtonBase` is a genuine coupling issue. Relocating `renderPendingSpinner` to `core/directives` is clearly correct. The behavior parity, focus-retention contract, test coverage, and documentation are all solid.

My concern is structural: **`PendingController` is a phantom abstraction.** `PendingMixin` is a thin façade that delegates every substantive behavior to it:

```typescript
// PendingMixin — pure delegation, no value added
public get pendingActive(): boolean {
    return this._pendingController.pendingActive;
}
public getPendingAccessibleName(): string {
    return this._pendingController.getPendingAccessibleName();
}
```

The controller is the real implementation; the mixin is just a wrapper around it. That indirection costs complexity without buying anything.

### The justification for a separate controller breaks down under scrutiny

The PR describes the controller as an escape hatch: "components that cannot use the mixin can use the controller and directive directly." But `PendingControllerHost` — the interface the controller requires — declares `pending` and `pendingLabel`. Those are exactly what the mixin adds. Any component using the controller directly still needs to implement that same interface, so it saves nothing over applying the mixin.

### What the single-mixin version looks like

Everything currently in `PendingController` moves directly into `PendingMixin`:

```typescript
export function PendingMixin<T extends Constructor<ReactiveElement>>(Base: T) {
  class PendingElement extends Base {
    @property({ type: Boolean, reflect: true }) pending = false;
    @property({ type: String, attribute: 'pending-label' }) pendingLabel?: string;
    @state() protected pendingActive = false;  // @state() handles re-renders automatically

    private _pendingTimer: ReturnType<typeof setTimeout> | null = null;

    protected resolvePendingAccessibleName(): string | null {
      return this.textContent?.trim() || null;
    }

    getPendingAccessibleName(): string { /* ... */ }

    override update(changedProperties: PropertyValues) {
      if (changedProperties.has('pending')) { /* timer + freeze */ }
    }
  }
  return PendingElement;
}
```

This is structurally what `ButtonBase` already had — just as an opt-in mixin instead of an inherited base class, which is exactly the right minimal change.

A secondary consequence of the controller: it uses `hostUpdate()` + a `_wasPending` tracking field to detect `pending` transitions, because controllers don't receive `changedProperties`. A mixin uses `update(changedProperties)` with `changedProperties.has('pending')` — the standard idiomatic Lit pattern, no extra state needed.

### Smaller issues

**Changeset severity is undersized.** Adding three new public entry points to `@spectrum-web-components/core` with named exports and package.json export maps is additive API — that's `minor`, not `patch`.

**`pendingActive` regressed from `@state()` to manual `requestUpdate()`.** `ButtonBase` used `@state() protected pendingActive`, which gives automatic re-render scheduling, change detection, and dev-mode logging. The controller drops to a plain private field + `host.requestUpdate()`. This works, but it's a step back from idiomatic Lit. A mixin can use `@state()` directly.

**The `disabled` cast in `PendingMixin.update()` is fragile.**
```typescript
(this as { disabled?: boolean }).disabled
```
Works today, but silently becomes a false negative when applied to a host without `disabled`. Either constrain the mixin's type parameter to include `disabled?`, or guard with `'disabled' in this`.

**Debug warning URL is already stale.**
The `pending + disabled` warning links to `/components/button/#pending`. When this mixin is applied to a non-button component in the future, that link is misleading.

### Recommended resolution

1. Drop `PendingController` and `PendingControllerHost` — collapse their logic into `PendingMixin` directly.
2. Keep `renderPendingSpinner` in `core/directives` as-is (correct decision).
3. Update the changeset to `minor`.
4. Restore `@state() protected pendingActive` in the mixin.

This halves the public API surface and removes the indirection layer, while delivering the same decoupling from `ButtonBase` that motivates the PR.

Rajdeepc:
@caseyisonit I think the direction here is right. Decoupling pending state behavior from `ButtonBase` and moving `renderPendingSpinner` into `core/directives` both make sense. The behavior parity, focus-retention handling, tests, and docs all look solid.

But my main concern is that `PendingController` feels like an extra abstraction layer without providing much additional value. Today, `PendingMixin` mostly delegates directly to the controller, making the controller the real implementation and the mixin a thin wrapper around it.

The rationale for exposing the controller as an escape hatch also seems a bit weak. Components using the controller still need to satisfy `PendingControllerHost` (`pending`, `pendingLabel`, etc.), which are the same pieces the mixin provides. In practice, it doesn't appear to reduce much integration work compared to simply applying the mixin.

I wonder if a simpler approach would be to move the controller logic directly into `PendingMixin`. I created one for you here: https://github.com/adobe/spectrum-web-components/pull/6440

A few smaller observations in your PR:

* The changeset feels more like a `minor` than a `patch` since we're adding new public exports.
* `pendingActive` moving from `@state()` to manual `requestUpdate()` feels like a step away from standard Lit patterns.
* The `disabled` cast in `PendingMixin.update()` is a little fragile; constraining the type or guarding with `'disabled' in this` would be safer.
* The warning URL is button-specific and may become misleading if the mixin is adopted elsewhere.

Overall, I like the goal and most of the implementation. My suggestion would be to collapse `PendingController` into `PendingMixin`, keep `renderPendingSpinner` where it is, and simplify the architecture while preserving the same behavior.

rubencarvalho:
Overall, I agree this is the most idiomatic approach, but I also agree with Rajdeep that we could simplify it a bit 😄 RN, I can’t really see a strong reason to introduce a controller here, especially since we know this exact behaviour is only needed by two components. LMK what you think!

2nd-gen/packages/core/controllers/pending-controller/stories/demo-hosts.ts
caseyisonit: Could you say more about this question?

cdransf: Ah, yeah! It's a plain function that returns HTML and doesn't use the Lit `directive` func or `Directive` class. I think it can stay here but the semantics don't necessarily line up with Lit directives.

caseyisonit:
After several discussion we are going to keep the controller and directive and its now wired up through the mixin and the controller owns the directive

caseyisonit:
thanks for the thorough review @Rajdeepc 🙏 talked it over and we're going to keep the controller + mixin + directive split rather than collapse into one mixin, so passing on #6440 — but genuinely appreciate you writing it out, it made the tradeoff way easier to reason about.

the way im thinking about it, each piece does one job and stands on its own:

- controller = the pending behavior/state (delay, inline-size freeze, busy name, host updates)
- directive = the render
- mixin = the element wiring (props + click suppression) that composes the two for the common button case

on the "extra abstraction" point — youre right the mixin mostly delegates today, but the controller earning its keep isnt really the escape-hatch framing. its that its instance-based, so a component can run more than one pending region each tied to its own prop, instead of the single `pending` property a mixin forces onto the host. thats the agnostic-controller pattern (kinda like multiple useEffects) and we lose it if we fold everything into one mixin. keeping the render in a standalone directive also stays reusable for stateless cases.

on `@state()` vs `requestUpdate()` — since `pendingActive` lives on the controller and not the element, this is just following Lit's documented controller pattern: a controller holds its own state and calls `this.host.requestUpdate()` to trigger the render (their `ClockController` example does exactly that), and `@state()` is scoped to `LitElement` internal state so a controller cant use it anyway. lines up with Lit's [composition guidance](https://lit.dev/docs/composition/overview/) — controllers for reusable stateful behavior, mixins for the property/API surface.

the smaller notes are all in the latest push: changeset bumped to minor, the `disabled` read is guarded with `'disabled' in this` instead of a cast, the pending+disabled warning url is no longer button-specific, and i added a `hostConnected()` to re-arm the delay timer on reconnect (good catch @cdransf).

cdransf (approved):

2nd-gen/packages/core/controllers/pending-controller/stories/demo-hosts.ts
```diff
@@ -0,0 +1,196 @@
...
+ * `pending-spinner.css` fragment lives in the `swc` package and is not reachable
+ * from `core`.
+ */
+@customElement('demo-pending-host')
+export class DemoPendingHost
+  extends LitElement
+  implements PendingControllerHost
+{
+  static override styles = css`
+    button {
```
5t3ph, line 52: [nit] can this use the shared pending spinner styles?

2nd-gen/packages/swc/components/button/Button.ts
```diff
@@ -124,6 +123,11 @@ export class Button extends ButtonBase {
     return [baseStyles, pendingSpinnerStyles, styles];
   }
 
+  /** The busy accessible name derives from the button's resolved name. */
+  protected override resolvePendingAccessibleName(): string | null {
```
5t3ph: This doesn't seem to be used, `aria-label` is still calling `this.getPendingAccessibleName()` (same in ActionButton). I might not be understanding the hops, though.

2nd-gen/packages/core/components/button/Button.base.ts
```diff
@@ -156,65 +114,26 @@ export abstract class ButtonBase extends SizedMixin(
...
-      clearTimeout(this._pendingTimer);
-      this._pendingTimer = null;
-    }
-    this.pendingActive = false;
     super.disconnectedCallback();
   }
 
   /**
-   * Suppresses click activation while the button is `disabled` or `pending`.
+   * Suppresses click activation while the button is `disabled`.
```
5t3ph, line 107: We need to bring back click suppression while pending

2nd-gen/packages/core/components/button/Button.base.ts
```diff
@@ -156,65 +114,26 @@ export abstract class ButtonBase extends SizedMixin(
...
-        }
-        this.renderRoot
-          .querySelector('button')
-          ?.style.removeProperty('--_swc-button-pending-inline-size');
-        this.pendingActive = false;
-      }
-    }
     super.update(changedProperties);
     if (window.__swc?.DEBUG) {
-      if (this.pending && this.disabled) {
```
5t3ph, line 207: I think this warning still has merit since it's about Button behavior expectations?

2nd-gen/packages/core/components/button/Button.base.ts
caseyisonit: so this is in the base, i will make sure in the component its overridden

2nd-gen/packages/core/components/button/Button.base.ts
caseyisonit: this is present in the mixin, please check if that satisifies your concern

CONTRIBUTOR-DOCS/03_project-planning/03_components/button/pending-reusable-primitives-plan.md
```diff
@@ -0,0 +1,108 @@
...
+2. **`renderPendingSpinner` directive** (`core/directives/pending-spinner`) — the render out of `swc/components/button/` as a render-only core directive (no token dependency). It is the implementation behind `PendingController.renderPendingState()`; consume it directly only for a stateless spinner without a controller. The token-based spinner CSS stays as the `swc` `_lit-styles/pending-spinner.css` fragment.
+3. **`PendingMixin`** (`core/mixins`) — a thin mixin that declares the `pending` / `pending-label` reactive properties, instantiates the controller, and wires click suppression, so a host opts in with one mixin.
+4. **`ButtonBase` stays pending-agnostic** — it gains no pending logic. `swc-button` and `swc-action-button` each apply `PendingMixin` to opt in. Other `ButtonBase` subclasses that do not need pending (e.g. `CloseButton`) are unaffected.
+
+## Behavior parity (must hold)
+
+- ARIA (`aria-disabled`, busy `aria-label`) stays synchronous on `pending`; the visual treatment keys off the delayed `pendingActive`.
```
5t3ph: @nikkimk should ARIA changes wait the 1 second delay, or go ahead and change? They are currently changing immediately. Design's logic for the visual delay was in case of very short processes the delay prevents the blink of the spinner swap.

2nd-gen/packages/core/controllers/pending-controller/stories/demo-hosts.ts
```diff
@@ -0,0 +1,196 @@
...
+      block-size: 18px;
+      stroke-width: 2px;
+    }
+
+    .swc-PendingSpinner-track {
+      stroke: rgb(255 255 255 / 40%);
+    }
+
+    .swc-PendingSpinner-fill {
+      stroke: #fff;
```
5t3ph: [nit] I know this is just a demo, but can we use a more visible color?
```suggestion
      stroke: var(--swc-blue-900);
```

2nd-gen/packages/core/components/button/Button.base.ts
5t3ph: validated, works as expected

2nd-gen/packages/core/components/button/Button.base.ts
5t3ph: lgtm, thanks!

5t3ph (approved):
Functionality checks out, thanks for thoughtfully working through this!

One nit, and one last min a11y check.

blunteshwar (approved):

CONTRIBUTOR-DOCS/03_project-planning/03_components/button/pending-reusable-primitives-plan.md
nikkimk: @5t3ph @caseyisonit Good question. The visual delay was mainly for the blink. On the off chance someone tried to access the component in that second it's still pending, let's not have them delay.

caseyisonit: ill be deleting this file before merge so what would you like me to do @5t3ph? i believe the aria attributes are updated immediately so if im understanding correctly we may not have to do anything new??

5t3ph: yes, all good!

2nd-gen/packages/core/controllers/pending-controller/stories/demo-hosts.ts
caseyisonit: so we are unable to reuse those here because we dont have css processing established in core. heres what claude gave me as the issue and paths forward but i think we should get this in and return to this after we discuss css in core as a concept and what that might look like

>  Short answer: not from where it lives. demo-hosts.ts is in core, and the shared fragment is a swc, token-based stylesheet — using it would invert the package dependency and pull in a build step core doesn't have.
> 
> Two concrete blockers:
> 
> Dependency direction. swc depends on core, never the reverse — and core's package.json has no @adobe/spectrum-wc dependency. Importing swc/stylesheets/_lit-styles/pending-spinner.css from a core file (even via a deep relative path) makes core reach up into swc. It might happen to resolve inside the swc Storybook that hosts these stories, but it would break core's standalone build/tests and is the wrong layering.
> Token build. The shared fragment is authored with token("track-color"), token("accent-content-color-default"), token("static-white-track-color"), etc. Those token(...) calls only become real values through swc's lit-css/token build, and they assume swc's tokens.css is loaded and a themed :host([static-color]). Core has no token/lit-css build step (that's the documented reason the spinner CSS stayed in swc), so the raw fragment wouldn't resolve in a core context, and the built artifact would again be a core→swc import.
> That's exactly why the demo inlines its own hardcoded, token-free styles (its own #fff/#2680eb, its own keyframes) — it's a deliberately standalone, non-themed element showing the controller + directive wired up without the swc theming layer. The comment already says as much.
> 
> So:
> 
> Keep it inlined — it's the correct trade-off for a core-local demo, and there's no clean core-side way to share the swc fragment. The only "dedupe" available in core would be another hand-written css literal, i.e. what's already there.
> If visual fidelity to the real spinner mattered, the move would be to relocate the demo host into the swc package (where tokens and the shared fragment are reachable) — but that trades away having the controller's demo co-located with the core controller it documents.

PR #6415, opened 2026-06-17
feat(icon): make swc-icon public with BYO SVG docs and private elements catalog

TarunAdobe:
## Description

  Make `<swc-icon>` a public 2nd-gen component: public Storybook docs, consumer migration guide, and npm export
  boundary, while keeping the internal `elements/` SVG catalog monorepo-only.

  ### Included in this PR

  - Contributor [icon migration plan](CONTRIBUTOR-DOCS/03_project-planning/03_components/icon/migration-plan.md)
  - Remove `@status internal` from `swc-icon` JSDoc and keep it aligned with other public 2nd-gen components
  (`@since` only)
  - Export `Icon` only from `components/icon/index.ts` so `@adobe/spectrum-wc/icon` stops leaking internal SVG
  factories
  - Public `icon.stories.ts`, `icon.mdx`, and `migration-guide.mdx` for a BYO SVG contract with no public
  `elements/` import path
  - Trim `icon.internal.*` down to a maintainer-only catalog for dev Storybook
  - Fix conversational-ai pattern imports to use local relative `elements/` paths instead of a public package
  path
  - Update `swc-icon` JSDoc examples to inline SVG only
  - Add a changeset for `@adobe/spectrum-wc`
  - Update 2nd-gen button docs and stories to prefer `<swc-icon slot="icon">...</swc-icon>` in icon-slot
  examples, matching the intended public icon path

  ## Motivation and context

  `<swc-icon>` already exists in 2nd-gen and is used internally by components and patterns, but it is still
  treated like an internal implementation detail. Consumers migrating from `<sp-icon>` need a documented public
  wrapper whose contract is simple: slot your own SVG.

  This PR makes that public contract explicit.

  It does **not** turn the internal `elements/` SVG catalog into a public icon registry. Those factories remain a
  maintainer convenience for SWC development only.

  It also aligns button docs with that direction so 2nd-gen documentation points consumers toward `swc-icon`
  instead of encouraging raw slotted SVG as the primary path.

  ## Related issue(s)

  - N/A

  ## Screenshots

  N/A. This is primarily a docs, Storybook, and package export-boundary change.

  ## Author's checklist

  - [x] I have read the **[CONTRIBUTING](https://github.com/adobe/spectrum-web-components/blob/main/
  CONTRIBUTING.md)** and **[PULL_REQUESTS](https://github.com/adobe/spectrum-web-components/blob/main/
  PULL_REQUESTS.md)** documents.
[...]

PR #6410, opened 2026-06-16
feat(close-button): migrate to 2nd-gen

TarunAdobe:
## Description
  Completes the **2nd-gen migration of Close Button** (Epic SWC-2087) by landing the full stacked workstream
   on `main`:
  | Merged PR | Phase |
  |-----------|-------|
  | #6355 | Scaffold — component structure, `ButtonBase` reuse, cross icons, exports |
  | #6360 | API + accessibility — public surface, Storybook contract tests |
  | #6385 | S2 styling + Storybook docs + consumer migration guide |
  | #6405 | Playwright a11y snapshots + keyboard coverage |
  ### SWC (`2nd-gen/packages/swc/components/close-button/`)
  - **`CloseButton`** — extends `ButtonBase`; native inner `<button type="button">` with delegated focus; 
  cross icon per `size` (`s` \| `m` \| `l` \| `xl`)
  - **Public API:** `size`, `static-color`, `accessible-label`, `disabled`
  - **`close-button.css`** — Spectrum 2 close-button visuals (size/static-color states, border + outline 
  focus, press `perspective` transform, aligned with 2nd-gen button interaction patterns)
  - **Storybook** — Playground, Overview, Anatomy, Sizes, Static colors, States, Accessibility + `Close 
  Button/Tests` interaction suite
  - **Docs** — `close-button.mdx`, `migration-guide.mdx`
  - **Tests** — Storybook interaction tests + `close-button.a11y.spec.ts` (ARIA snapshots + tab-focus)
  ### Core / shared
  - **`ButtonBase`** — `static-color` on shared button base; DEBUG validation for invalid `static-color` and
   missing `accessible-label` on icon-only controls (inherited by close-button)
  - **Cross icons** — `Cross400Icon`, `Cross500Icon` for `l` / `xl` sizes
  ### Breaking changes (1st-gen → 2nd-gen)
  | Change | Migration |
  |--------|-----------|
  | Tag | `<sp-close-button>` → `<swc-close-button>` |
  | Accessible name | `label` → `accessible-label` (or default slot text) |
  | Static color | `variant="white\|black"` **not supported** — use `static-color` only |
  | CSS overrides | `--mod-closebutton-*` → reviewed `--swc-close-button-*` |
  `sp-close-button` in 1st-gen is **unchanged** in this PR.
  ## Motivation and context
[...]

PR #6441, opened 2026-06-23
docs(infield-button): add Phase 1 migration plan

Rajdeepc:
## Description

Adds the Phase 1 (prep) migration plan for `swc-infield-button` under `CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/`, completing the planning artifact set alongside the existing rendering/styling and accessibility migration analyses.

The plan documents the 1st-gen API surface, breaking changes, 2nd-gen API decisions, the core/SWC architecture split, and the full Setup → API → Styling → Accessibility → Testing → Documentation → Review checklist. It follows the `migration-prep` template and is structurally aligned with the `button` and `action-button` plans.

### Highlights

- **API**: `quiet` + four sizes (`s`/`m`/`l`/`xl`, `noDefaultSize: true`), extends 2nd-gen `ButtonBase`; link API, `block`/`inline`, `pending`, and `--mod-*` removed; `label` → `accessible-label`; icon moves to the named `icon` slot.
- **Design/CSS inputs confirmed**: validated against the Figma S2 In-field button spec (node `126176-34080`) and the `spectrum-two` `components/infieldbutton/index.css` baseline (sizes, states, quiet, forced-colors, no stacked/`:lang()` classes).
- **Styling**: token mapping to `token()` + `--swc-infield-button-*`, slotted-icon styling via `::slotted`, parent-field-owned focus model and inline-group wrapper.
- Updates the components README navigation.

## Open questions (tracked in the plan)

These are team decisions and intentionally remain open before implementation begins:

- **Q3**: confirm sequencing (infield-button before number-field)
- **Q4**: keep `swc-infield-button` standalone vs. merge with clear/close-button
- **Q5**: ownership of the "clear field" affordance

## Type of change

- Documentation (planning artifact only; no production code or component behavior changes)

## Related issue(s)

SWC-2147 (Epic: SWC-2107)

## Test plan
- [ ] Plan reviewed by at least one other engineer


CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
```diff
@@ -0,0 +1,570 @@
...
+> - **Accessibility is non-negotiable** — all a11y requirements ship as part of this migration.
+> - **Breaking changes** are assessed on merit.
+
+### Must ship — breaking or a11y-required
+
+#### API and naming
+
+| # | What changes | 1st-gen behavior | 2nd-gen behavior | Consumer migration path |
+| --- | ------------ | ---------------- | ---------------- | ----------------------- |
+| **B1** | Remove `block` attribute | `block="start"/"end"` adjusted corner radius and border for stacked stepper pairs | Removed. S2 uses consistent corner radius; parent field host owns stacked layout via DOM order and CSS | Remove `block` attribute from all `sp-infield-button` usages. Restructure parent field layout to manage corner radius and borders on the containing field element. |
```
rise-erpelding: We could clean this up a bit, it's a bit misleading. S2 won't have a stacked layout as the buttons will lay side by side, I believe. And because the buttons aren't flush against the parent field anymore, we don't need to manage corner radius and borders as before.

```suggestion
| **B1** | Remove `block` attribute | `block="start"/"end"` adjusted corner radius and border for stacked stepper pairs | Removed. S2 uses consistent corner radius. | Remove `block` attribute from all `sp-infield-button` usages. |
```

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
```diff
@@ -0,0 +1,570 @@
...
+| **B2** | Remove `inline` attribute | `inline="start"/"end"` adjusted corner radius and edge attachment for horizontal inline groups | Removed. Same reason as B1. The parent field host owns inline-group layout (it renders the `.swc-InfieldButton-inline` wrapper). | Remove `inline` attribute. Ensure the parent field host wraps the in-field buttons in the `.swc-InfieldButton-inline` container (or equivalent CSS context). |
+| **B3** | Rename `label` → `accessible-label` | `label` attribute maps to `aria-label` on inner `<button>` | `accessible-label` attribute maps to `aria-label` on inner `<button>`, via `ButtonBase`'s `accessibleLabel` property | Replace `label="…"` with `accessible-label="…"` on each `sp-infield-button`. **Required for icon-only; dev warning fires if omitted.** |
```
rise-erpelding: "Required for icon-only" is a bit confusing here given that the component is always icon-only.

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
```diff
@@ -0,0 +1,570 @@
...
+| **B4** | Remove link API (`href`, `target`, `download`, `rel`, `referrerpolicy`) | Inherited from 1st-gen `ButtonBase` via `LikeAnchor` mixin | Not implemented. `swc-infield-button` is always `type="button"`. 2nd-gen `ButtonBase` does not expose link attributes. | Remove any `href`-based usage of `sp-infield-button`. Use `<a>` or `swc-button` with an `href` when navigation is needed in a field context (uncommon). |
+| **B5** | Remove `--mod-infield-button-*` CSS custom properties | 17 modifier properties supported | None of the `--mod-infield-button-*` properties are carried forward. `--swc-infield-button-*` properties may be introduced selectively (see A1). | Consumers overriding infield-button styling via `--mod-infield-button-*` must switch to `--swc-infield-button-*` equivalents when available, or use CSS custom property targeting directly. |
```
rise-erpelding: I'm not totally sure what "use CSS custom property targeting directly" means, but overall our styling in Gen 2 is meant to be less reliant on custom properties and would use the tokens directly when appropriate.

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
```diff
@@ -0,0 +1,570 @@
...
+| **B6** | Icons move from default slot → `icon` slot | `<sp-infield-button label="…"><sp-icon-add></sp-icon-add></sp-infield-button>` (icon in default slot) | Icon must be in the named `icon` slot to trigger `ButtonBase`'s `hasIcon` detection and the `accessible-label` dev warning | Add `slot="icon"` to slotted icon elements: `<swc-infield-button accessible-label="…"><sp-icon-add slot="icon"></sp-icon-add></swc-infield-button>` |
```
rise-erpelding: This `<swc-infield-button...>` code snippet includes `<sp-icon>`, should it be `<swc-icon>` instead?

I think if #6415 goes through we'd also be slotting the `<svg>` element into `<swc-icon>` rather than using something like `<swc-icon-add>`

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
```diff
@@ -0,0 +1,570 @@
...
+Additional presentation modes **not supported in 2nd-gen:**
+- Stacked position variant (`block="start"/"end"`) — removed; consistent corner radius in S2
+- Inline position variant (`inline="start"/"end"`) — removed; the parent field host owns inline-group layout (it renders the `.swc-InfieldButton-inline` wrapper around the slotted buttons), not `swc-infield-button` itself
+- Static color (`static-color="white"/"black"`) — **not present** in Figma S2 spec; confirmed absent (not in 1st-gen; no field-chrome use case for static color)
+
+#### Slots (2nd-gen)
+
+| Slot | Content | Notes |
+| ---- | ------- | ----- |
+| `icon` | Icon element | **Confirmed.** Named `icon` slot (from `ButtonBase`). `ObserveSlotPresence` in `ButtonBase` monitors `[slot="icon"]` for icon presence. Icon must be decorative (`aria-hidden="true"` on the icon element when `accessible-label` provides the name). **Breaking change from 1st-gen default slot (B6).** |
```
rise-erpelding: I read this part about the `aria-hidden` and interpret it as being something that needs to be added manually, I think the icon component takes care of putting `aria-hidden="true"` on the icon element so it's not necessary to add it, but worth double-checking.

That might just be me misreading what this is trying to say, definitely feel free to disregard this comment if you don't think it needs a change.

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
```diff
@@ -0,0 +1,570 @@
...
+  - Quiet: `transparent` background for default, hover, active, and disabled (not the gray disabled color)
+- [ ] Focus model: the inner `<button>` sets `outline: none` (matches S2 source). In **composed** field contexts the parent field (number-field, textfield, picker) owns the focus-visible ring while the inner `<button>` stays keyboard-reachable via `delegatesFocus`; verify with a field-level story. In **isolated** standalone usage there is no visible focus indicator by default; if one is required it must be added deliberately and meet WCAG 2.4.7. Do not add a competing `:focus-visible` outline that would double-ring inside a field
+- [ ] Forced colors (Windows High Contrast): default variant on `.swc-InfieldButton` → background `ButtonText`, icon `ButtonFace`, `forced-color-adjust: none` on the `<button>`; disabled → background `GrayText`; hover/active/focus-visible → background `Highlight`. Quiet variant → background `Canvas`, icon `ButtonText`; disabled → `Canvas` / `GrayText`; hover/active/focus-visible → `Canvas` / `Highlight`. Sort the `@media (forced-colors: active)` block to the bottom of the file per style guide
```
rise-erpelding: I have some doubts about this item. Our style guide explicitly says:

> Forced colors styles are only included when browser defaults are not sufficient

So I would hesitate to just bring over styles from spectrum-css before checking them manually.

However, after consulting with a couple of agents, they both pushed back on my doubts and said it was reasonable to port the styles over from spectrum-css... wonder if we need an adjustment in a skill or doc somewhere? Tagging @5t3ph if she has an opinion on this.

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
```diff
@@ -0,0 +1,570 @@
...
+
+| # | Item | Blocking? | Status | Owner |
+| --- | ---- | --------- | ------ | ----- |
+| **Q1** | **Figma spec reviewed.** The `S2 / Web (Desktop scale)` In-field button spec (node [`126176-34080`](https://www.figma.com/design/Mngz9H7WZLbrCvGQf3GnsY/S2---Web--Desktop-scale-?node-id=126176-34080)) was reviewed. Confirmed: 4 sizes (S, M, L, XL), quiet variant, standard interaction states (default, hover, active, disabled), no static-color variant, no stacked position variants, and four consumer-supplied icon affordances (disclosure, clear, dash, add). Visual matrix updated. | Resolved | **Resolved** | — |
+
+### Architecture and behavior
+
+| # | Item | Blocking? | Status | Owner |
+| --- | ---- | --------- | ------ | ----- |
+| **Q3** | **Migration ordering: infield-button before number-field.** Recommended based on the number-field rendering analysis explicitly composing `swc-infield-button`. If the team has a different sequencing preference, both this plan and the number-field analysis need updating. | Yes — affects planning | Open (recommendation made) | Architecture reviewer. Next action: confirm or override the sequencing recommendation. |
```
rise-erpelding: We might be able to remove this item or at least close it out - the sequencing seems to be confirmed, especially given that we're working on this migration plan first and holding number field for later.

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
```diff
@@ -0,0 +1,570 @@
...
+| **Q5** | **Scope intersection with `swc-clear-button`: who owns "clear field" affordances?** In 1st-gen, `sp-infield-button` is used for generic field actions (stepper +/−, disclosure). `sp-clear-button` is a separate component. In 2nd-gen, if `swc-infield-button` (quiet variant) and `swc-clear-button` have the same visual appearance, there is an argument for a single component. **Recommendation: keep separate, but resolve the composition contract before `swc-number-field` migration.** Confirm: does `swc-textfield` / `swc-search` use `swc-infield-button` or `swc-clear-button` for their clear actions? The answer affects the scope boundary and the consumer migration guide for both components. The Figma In-field button spec ([`126176-34080`](https://www.figma.com/design/Mngz9H7WZLbrCvGQf3GnsY/S2---Web--Desktop-scale-?node-id=126176-34080)) documents a **clear** icon affordance as one of the in-field button icon types, which is direct evidence for the overlap. Next action: review that frame and the textfield/search rendering analyses to confirm which button type each field uses for clear. | Partial — blocks consumer guide scope | Open | Architecture + design. |
```
rise-erpelding: > Next action: review that frame and the textfield/search rendering analyses to confirm which button type each field uses for clear.

These exist, no? Feels like we could probably have an agent knock this out quickly and close this out? Also I believe textfield/search and also tag use the clear-button rather than the infield button.

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
```diff
@@ -0,0 +1,569 @@
...
+
+#### Styling and visuals
+
+| # | What changes | 1st-gen behavior | 2nd-gen behavior | Consumer migration path |
+| --- | ------------ | ---------------- | ---------------- | ----------------------- |
+| **S1** | CSS custom properties (see B5 in API table above) | `--mod-infield-button-*` | `--swc-infield-button-*` (limited set; see A1) | Same as B5 in API table |
+| **S2** | Stacked-border tokens removed | `--mod-infield-button-stacked-*` tokens handled inner border radius resets for stacked pairs | Removed entirely; S2 uses a consistent corner radius on each button | No consumer action needed if `block`/`inline` are removed per B1/B2 |
+| **S3** | Focus state ownership shifts to parent | 1st-gen: each `sp-infield-button` manages its own `:focus-visible` ring | 2nd-gen: the inner `<button>` suppresses its own outline (`outline: none`, matching S2 source); the parent field shows the focus ring in composed contexts while the inner `<button>` stays keyboard-reachable | Verify with field-level stories that focus visibility is preserved and meets WCAG 2.4.7 |
```
5t3ph: If the button is independently focusable, then it needs to retain a focus ring for keyboard a11y.

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
```diff
@@ -0,0 +1,569 @@
...
+
+#### Properties / attributes (2nd-gen)
+
+| Property | Type | Default | Attribute | Confidence | Notes |
+| -------- | ---- | ------- | --------- | ---------- | ----- |
+| `quiet` | `boolean` | `false` | `quiet` | **Confirmed** | Visual variant only; no role or name change. From 1st-gen; preserved in S2 spec. |
+| `size` | `'s' \| 'm' \| 'l' \| 'xl'` | none | `size` | **Confirmed** | `noDefaultSize: true` preserved; parent field host must provide size. |
+| `disabled` | `boolean` | `false` | `disabled` | **Confirmed** | From `ButtonBase`. Both self-disabled and parent-driven `disabled` must work. |
+| `accessibleLabel` | `string \| undefined` | undefined | `accessible-label` | **Confirmed** | From 2nd-gen `ButtonBase`. Required for icon-only. Dev warning fires when `hasIcon && !hasLabel && !accessibleLabel`. |
+| `active` | `boolean` | `false` | `active` | **Inferred** | From `ButtonBase`. Preserves pressed visual state. |
```
5t3ph: This is not available in 2nd-gen ButtonBase. Do we actually need this state? What is the use case for a "toggle" button in-field?

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
```diff
@@ -0,0 +1,569 @@
...
+| `type` | `'button' \| 'submit' \| 'reset'` | `'button'` | `type` | **Confirmed** | From `ButtonBase`. Always `"button"` for in-field adornments in practice; `submit`/`reset` deferred (A3). |
```
5t3ph: Also not in 2nd-gen ButtonBase, so perhaps consider it dropped from the API or at least deferred?

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
```diff
@@ -0,0 +1,569 @@
...
+| ---------- | ----------- | ----- |
+| Disclosure | Picker / combobox trigger | Default Figma reference icon; uses `field-edge-to-disclosure-icon-*` padding |
+| Clear | Text field / search clear | Overlaps `swc-clear-button` scope — see Q5 |
+| Dash (−) | Number-field stepper decrement | Used in the inline stepper pair composed by the field host |
+| Add (+) | Number-field stepper increment | Used in the inline stepper pair composed by the field host |
+
+Additional presentation modes **not supported in 2nd-gen:**
+- Stacked position variant (`block="start"/"end"`) — removed; consistent corner radius in S2
+- Inline position variant (`inline="start"/"end"`) — removed; the parent field host owns inline-group layout (it renders the `.swc-InfieldButton-inline` wrapper around the slotted buttons), not `swc-infield-button` itself
+- Static color (`static-color="white"/"black"`) — **not present** in Figma S2 spec; confirmed absent (not in 1st-gen; no field-chrome use case for static color)
```
5t3ph: ButtonBase is currently slated to get `staticColor` [exposed via this PR](https://github.com/adobe/spectrum-web-components/pull/6410/changes#diff-6f15e534a3662f3733ff49cb2327064865d31e5e9802f9324960afef2fe57267R89), so if that happens you'll need to exclude it.

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
```diff
@@ -0,0 +1,569 @@
...
+
+#### Slots (2nd-gen)
+
+| Slot | Content | Notes |
+| ---- | ------- | ----- |
+| `icon` | Icon element | **Confirmed.** Named `icon` slot (from `ButtonBase`). `ObserveSlotPresence` in `ButtonBase` monitors `[slot="icon"]` for icon presence. The `swc-icon` component sets `aria-hidden="true"` on itself; no manual attribute needed on the slotted element. **Breaking change from 1st-gen default slot (B6).** |
+| (default) | Optional visible label text | **Inferred.** Unlikely to contain content in typical use — `swc-infield-button` is always icon-only in practice. Present because `ButtonBase` exposes it; `ObserveSlotText` monitors it for `hasLabel`. |
```
5t3ph: These are not in 2nd-gen ButtonBase since that doesn't manage render.

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
```diff
@@ -0,0 +1,569 @@
...
+- Core owns: `quiet` property declaration, size validation (`['s', 'm', 'l', 'xl']` with `noDefaultSize: true`), `accessible-label` wiring inherited from `ButtonBase`, `disabled` forwarding
+- SWC renders the inner `<button>` with size and quiet modifier classes applied via `classMap` (matching the S2 selectors), the slotted icon styled with `slot[name="icon"]::slotted(*)`:
+
+```ts
+html`
+  <button
+    class=${classMap({
+      'swc-InfieldButton': true,
+      [`swc-InfieldButton--size${size?.toUpperCase()}`]: !!size,
+      'swc-InfieldButton--quiet': this.quiet,
```
5t3ph: [nit] attribute selectors will be used for these states

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
```diff
@@ -0,0 +1,569 @@
...
+    })}
+  >
+    <div class="swc-InfieldButton-fill">
```
5t3ph: [nit] this is kind of odd naming, what is "fill" trying to communicate? Also I wonder if this will be able to be removed in favor of just `::slotted()` styles.

5t3ph (changes requested):
It looks like the agent was mixing up ButtonBase from 1st-gen vs 2nd-gen, and I probably missed some spots where that impacts the recommendations here.

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
```diff
@@ -0,0 +1,573 @@
...
+- [ ] Map visual states to `token()`:
+  - Default: background `token("gray-100")`, icon `token("neutral-content-color-default")`
+  - Hover: background `token("gray-200")`, icon `token("neutral-content-color-hover")`
+  - Active/down: background `token("gray-200")`, icon `token("neutral-content-color-down")`; `transform: perspective(...) translateZ(...)` on the `<button>` using the S2 downstate tokens
+  - Disabled: background `token("disabled-background-color")`, icon `token("disabled-content-color")`
+  - Quiet: `transparent` background for default, hover, active, and disabled (not the gray disabled color)
+- [ ] Focus model: the inner `<button>` sets `outline: none` (matches S2 source). In **composed** field contexts the parent field (number-field, textfield, picker) owns the focus-visible ring while the inner `<button>` stays keyboard-reachable via `delegatesFocus`; verify with a field-level story. In **isolated** standalone usage there is no visible focus indicator by default; if one is required it must be added deliberately and meet WCAG 2.4.7. Do not add a competing `:focus-visible` outline that would double-ring inside a field
```
rise-erpelding: Further up, the plan says:

> When used standalone (outside a parent field), the button must retain its own `:focus-visible` ring to satisfy WCAG 2.4.7.

Here it says:

> In **isolated** standalone usage there is no visible focus indicator by default; if one is required it must be added deliberately and meet WCAG 2.4.7.

We might change this part to remove the "if one is required"

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
```diff
@@ -0,0 +1,573 @@
...
+- [ ] Wire exports in both `package.json` files
+- [ ] For Phase 5, check out `spectrum-css` at the `spectrum-two` branch as a sibling directory for local copy-from-source (baseline already confirmed via GitHub API — Q2 resolved)
+
+### API
+
+#### Naming and public surface
+
+- [ ] `InfieldButton.types.ts`: define `InfieldButtonSize` (alias or reuse from `BUTTON_VALID_SIZES`), `InfieldButtonQuiet` boolean
+- [ ] `InfieldButton.base.ts`: extend 2nd-gen `ButtonBase` with `SizedMixin(ButtonBase, { noDefaultSize: true })`; `validSizes` does not need overriding because `BUTTON_VALID_SIZES` already equals `['s', 'm', 'l', 'xl']` (see Q6); add `quiet: boolean` property (reflect: true)
+- [ ] `InfieldButton.ts`: extend `InfieldButton.base.ts`; register as `swc-infield-button`; render inner `<button>` (with size and `quiet` modifier classes via `classMap`) containing a `.swc-InfieldButton-fill` div that wraps the `icon` slot
```
rise-erpelding: We're not using `classMap` for size/`quiet` modifiers, we'd use the host attribute selectors - this appears in a couple places:

Here:
> (with size and `quiet` modifier classes via `classMap`)

Around L429:
> Apply `.swc-InfieldButton` (with size/quiet modifier classes via `classMap`)

rise-erpelding: Also, an additional note that we're saying earlier that the `fill` `<div>` might not be needed, it could be helpful for the agent to ensure that's mentioned again here

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
```diff
@@ -0,0 +1,573 @@
...
+
+- [ ] Copy S2 source from `spectrum-css` `spectrum-two` branch `components/infieldbutton/index.css` (not `/dist`) into `infield-button.css` as the functionally-equivalent baseline
+- [ ] Apply `.swc-InfieldButton` (with size/quiet modifier classes via `classMap`) to the inner `<button>`; keep styling off `:host`
+- [ ] Update class and custom property prefixes from `.spectrum-InfieldButton` to `.swc-InfieldButton`; **remove all `--mod-infield-button-*` and `--spectrum-infield-button-*` fallback chains**, collapsing each into a single intentional `--swc-infield-button-*` property with a `token(...)` default per the [custom properties style guide](../../../../CONTRIBUTOR-DOCS/02_style-guide/01_css/02_custom-properties.md)
+- [ ] Style the slotted icon with `slot[name="icon"]::slotted(*)` (color + size-specific padding), as `swc-button` does; do **not** put a `.swc-InfieldButton-icon` class on the consumer-slotted node (the S2 `.spectrum-InfieldButton-icon` rule targets an inline SVG in the CSS-only template, which does not apply to the slotted web-component case)
+- [ ] `.swc-InfieldButton-fill` (inner `<div>`) owns `background-color`, `border-radius`, and the centering flex
```
rise-erpelding: Noting we said earlier that `.swc-InfieldButton-fill` _might_ not be necessary, I mentioned this as well in my comment on L412

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
```diff
@@ -0,0 +1,573 @@
...
+
+**Focus delegation and keyboard:**
+`delegatesFocus: true` is set in `ButtonBase.shadowRootOptions`. The inner `<button>` receives focus; `Enter` / `Space` activate. The component does not implement `FocusgroupNavigationController`. The parent field host owns navigation among input, stepper pairs, and other field siblings.
+
+**No link API:**
+`swc-infield-button` has no `href` or related anchor attributes. It is always `type="button"` in practice.
+
+**`noDefaultSize: true`:**
+The component does not pick a default size. The parent field host must set `size` (either by attribute or inherited context). This behavior is preserved from 1st-gen. 2nd-gen `ButtonBase` already applies `SizedMixin` with `validSizes: BUTTON_VALID_SIZES` (= `['s', 'm', 'l', 'xl']`), so `InfieldButton.base.ts` only needs to re-apply `SizedMixin(..., { noDefaultSize: true })`; the `validSizes` set already matches and does not need to be overridden (see Q6). Note that M is the Figma **reference** size, not a component default.
```
rise-erpelding: Claude's informed me that we might copy the action button's approach on this?

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
```diff
@@ -0,0 +1,573 @@
...
+| ---------- | ----------- | ----- |
+| Disclosure | Picker / combobox trigger | Default Figma reference icon; uses `field-edge-to-disclosure-icon-*` padding |
+| Clear | Text field / search clear | Overlaps `swc-clear-button` scope — see Q5 |
+| Dash (−) | Number-field stepper decrement | Used in the inline stepper pair composed by the field host |
+| Add (+) | Number-field stepper increment | Used in the inline stepper pair composed by the field host |
+
+Additional presentation modes **not supported in 2nd-gen:**
+- Stacked position variant (`block="start"/"end"`) — removed; consistent corner radius in S2
+- Inline position variant (`inline="start"/"end"`) — removed; the parent field host owns inline-group layout (it renders the `.swc-InfieldButton-inline` wrapper around the slotted buttons), not `swc-infield-button` itself
+- Static color (`static-color="white"/"black"`) — **not present** in Figma S2 spec; confirmed absent (not in 1st-gen; no field-chrome use case for static color). **Note:** if `ButtonBase` gains a `staticColor` property (tracked in PR #6410), `swc-infield-button` must explicitly exclude or override it to prevent consumers from setting an unsupported value.
```
rise-erpelding: Do we also need to do this for pending? The infield button doesn't have pending, but the button base does, so we would need to ensure it doesn't unintentionally set pending styles without any pending action?

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
```diff
@@ -285,14 +287,14 @@ Use lightweight confidence labels: **Confirmed**, **Inferred**, **Open question*
 Additional presentation modes **not supported in 2nd-gen:**
 - Stacked position variant (`block="start"/"end"`) — removed; consistent corner radius in S2
 - Inline position variant (`inline="start"/"end"`) — removed; the parent field host owns inline-group layout (it renders the `.swc-InfieldButton-inline` wrapper around the slotted buttons), not `swc-infield-button` itself
-- Static color (`static-color="white"/"black"`) — **not present** in Figma S2 spec; confirmed absent (not in 1st-gen; no field-chrome use case for static color)
+- Static color (`static-color="white"/"black"`) — **not present** in Figma S2 spec; confirmed absent (not in 1st-gen; no field-chrome use case for static color). **Note:** if `ButtonBase` gains a `staticColor` property (tracked in PR #6410), `swc-infield-button` must explicitly exclude or override it to prevent consumers from setting an unsupported value.
```
5t3ph: Update that the close button PR ended up reverting this decision since at least infield and clear do not use static colors

5t3ph (commented):
Apologies for causing confusion on the focus behavior. I have chatted with Nikki and gotten clarification on the expected behavior.

**Infield button focus/keyboard behavior corrections**

- Infield buttons are **never standalone** — remove that scenario and its `:focus-visible` ring requirement.
- They are **clickable but not focusable**. Keyboard behavior equivalent to clicking (e.g., arrow keys for increment/decrement on number field) is owned by the parent field, not the button.
- The parent field's focus ring is the only focus indicator needed — no separate WCAG 2.4.7 concern for the button itself.

Please update the plan to reflect this with language that makes the non-focusable, parent-delegated keyboard model explicit.

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
Rajdeepc: Fixed — simplified the B1 row per the suggestion. Removed the misleading stacked-layout language; the row now reads: "Removed. S2 uses consistent corner radius."

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
Rajdeepc: Fixed — dropped "for icon-only" from the B3 note. Since the component is always icon-only, the qualifier was redundant.

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
Rajdeepc: Fixed — replaced the vague phrase with an accurate note: 2nd-gen styling relies on design tokens directly rather than the `--mod-*` override surface.

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
Rajdeepc: Fixed — updated the code example to use `<swc-icon slot="icon"><svg>…</svg></swc-icon>` per the PR #6415 slotted-SVG pattern.

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
Rajdeepc: Fixed — clarified that `swc-icon` sets `aria-hidden="true"` on itself; no manual attribute is needed on the slotted element.

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
Rajdeepc: Fixed — the checklist item now requires checking each state against browser defaults first before adding `@media (forced-colors: active)` overrides, per the style guide. We should not port spectrum-css forced-colors rules wholesale.

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
Rajdeepc: Fixed — Q3 is now marked as **Resolved**. Sequencing is confirmed in practice: infield-button migration plan is proceeding first; number-field follows.

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
Rajdeepc: Fixed — Q5 is now marked as **Resolved**. Confirmed that textfield, search, and tag use `sp-clear-button`, not `sp-infield-button`, for clear actions. The two components remain separate.

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
Rajdeepc: Initially we added a standalone ring requirement here, but this has been corrected following Nikki's clarification (see subsequent comments). Infield buttons are never standalone and not independently focusable — the parent field's focus ring is the only indicator, and WCAG 2.4.7 is satisfied at the field level, not the button level. The plan now reflects the non-focusable, parent-delegated keyboard model throughout.

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
Rajdeepc: Fixed — `active` has been dropped from the public API table. The `:active` CSS pseudo-class handles the pressed visual state; there is no programmatic toggle use case for an in-field button.

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
Rajdeepc: Fixed — `type` has been dropped from the public API. The concrete template hardcodes `type="button"` on the inner `<button>`; `submit`/`reset` values remain deferred as A3 (form-associated).

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
Rajdeepc: Updated in light of your follow-up comment — the plan now reflects that the close-button PR reverted this decision. `ButtonBase` will not gain `staticColor` for infield/clear use cases, so no exclusion is needed.

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
Rajdeepc: Fixed — the default slot note now correctly says the slot is defined in the concrete component's template, not provided by `ButtonBase` (which has no render method). `ObserveSlotText` in `ButtonBase` monitors the slot text to set `hasLabel`.

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
Rajdeepc: Fixed — the render template and both Phase 3/5 checklist items now use `:host([quiet])` / `:host([size="s"])` host attribute selectors instead of `classMap` modifiers for `quiet` and `size`.

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
Rajdeepc: Fixed — removed the `fill` wrapper from the render template example. Added a Phase 5 verification note: confirm whether `.swc-InfieldButton-fill` is required by S2 CSS selectors or can be dropped in favour of `::slotted()` alone.

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
Rajdeepc: Corrected — and then superseded by Nikki's clarification. The plan now reflects the correct model throughout: infield buttons are never standalone, not independently focusable, and not in the tab order. The parent field's focus ring satisfies WCAG 2.4.7; the button itself has no focus ring. There is no standalone scenario anywhere in the plan.

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
Rajdeepc: Fixed — both instances ('with size and `quiet` modifier classes via `classMap`' at the API checklist and 'with size/quiet modifier classes via `classMap`' at the styling checklist) have been updated to use `:host([quiet])` / `:host([size="s"])` attribute selectors.

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
Rajdeepc: Fixed — the Phase 5 styling checklist item is now framed as a verification task: confirm whether `.swc-InfieldButton-fill` is required by S2 selectors or can be removed in favour of `::slotted()` alone.

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
Rajdeepc: Updated — the `noDefaultSize` explanation now follows the action-button plan's pattern: `getAttribute('size')` returns `null` until the parent field sets it; the JS property has no useful fallback default.

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
Rajdeepc: Good catch — added a note to the Behavioral semantics section: `infield-button.css` must not include pending visual rules; an inherited `pending` attribute on the host should have no visual effect on this component.

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
Rajdeepc: Updated — the plan now reflects this: the close-button PR confirmed that `ButtonBase` will not gain `staticColor` for infield/clear use cases. No exclusion is needed; the original warning has been removed.

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
```diff
@@ -0,0 +1,573 @@
...
+| **B5** | Remove `--mod-infield-button-*` CSS custom properties | 17 modifier properties supported | None of the `--mod-infield-button-*` properties are carried forward. `--swc-infield-button-*` properties may be introduced selectively (see A1). | Consumers overriding infield-button styling via `--mod-infield-button-*` must switch to `--swc-infield-button-*` equivalents when available. 2nd-gen styling relies on design tokens directly rather than the `--mod-*` override surface. |
+| **B6** | Icons move from default slot → `icon` slot | `<sp-infield-button label="…"><sp-icon-add></sp-icon-add></sp-infield-button>` (icon in default slot) | Icon must be in the named `icon` slot to trigger `ButtonBase`'s `hasIcon` detection and the `accessible-label` dev warning | Slot the icon element with `slot="icon"`: `<swc-infield-button accessible-label="…"><swc-icon slot="icon"><svg>…</svg></swc-icon></swc-infield-button>`. See PR [#6415](https://github.com/adobe/spectrum-web-components/pull/6415) for the `<swc-icon>` SVG-slot pattern. |
```
rise-erpelding, line 209: What do we need to consider being that #6415 is currently paused? Until it moves forward, will we just continue to use svg elements with `slot="icon"`?

rise-erpelding (approved):
This looks good to me. I would be aware that the icons changes mentioned here haven't landed yet and probably won't land in time for the first iteration of the component, so the outlined approach with `<swc-icon>` might not work as planned.

I know @5t3ph had some thoughts on tokens and I don't have the full context on which tokens would be potentially best to use, but that's a phase 5 concern that we could touch on a bit later.

5t3ph (approved):

PR #6453, opened 2026-06-25
ci: notify slack when checks fail after merging to main

pfulton:
## Description

Sets up a new GitHub Action for sending messages to Slack if checks fail after branches are merged into `main`.

Right now, this is set up so that I get a Slack DM if failures occur. Once I know that this is working and isn't absolutely obnoxious with the message(s) it's delivering, I'll land a follow-up PR to switch this from a DM to me to reporting out to our team channel in Slack.

## Motivation and context
Right now, no one is notified of failing checks after branches are merged into `main`. We wanted _some_ way of being alerted instead of manually checking GitHub periodically. An alert to the team's channel should be much more effective in eliciting a response.

## Author's checklist

- [X] I have read the **[CONTRIBUTING](<(https://github.com/adobe/spectrum-web-components/blob/main/CONTRIBUTING.md)>)** and **[PULL_REQUESTS](<(https://github.com/adobe/spectrum-web-components/blob/main/PULL_REQUESTS.md)>)** documents.

---

### Manual review test cases

- [ ] _Descriptive Test Statement_
  1. Merge this PR into `main`
  2. Wait for checks failures to occur
  3. Expect that @pfulton gets a DM from the Slack bot alerting him to the failure.


cdransf (approved):

caseyisonit (approved):
one tiny nit, can we add a comment in this file that this is sending the message to patrick via whatever the bots name is so we can track it and know what all the slack stuff is doing?

PR #6461, opened 2026-06-30
fix(global-css): make generated global element CSS deterministic

rubencarvalho:
## Description

`yarn build` reliably left `global-action-button.css` and `global-button.css` modified in the working tree, even on a clean checkout.

**Root cause:** the `vite-global-elements-css` plugin ran a best-effort `stylelint --fix` on each generated file purely to reorder properties. That subprocess **ignored its exit code and swallowed all errors**, so whenever it didn't apply — binary not found, or the process killed under CPU contention during parallel workspace builds — the raw generation order was written instead of the stylelint-ordered version that was checked in. The result was a phantom diff after every build.

**Fix:** remove the reorder step. Output is now deterministic (generation order), and the two affected files are regenerated to match.

## Key decisions

- **Deleted the reorder rather than making it reliable.** The `2nd-gen/packages/swc/stylesheets/global` directory is already in `.stylelintignore`, so property order on these `DO NOT EDIT`, machine-generated files is enforced nowhere. It is also functionally irrelevant: the cascade is unaffected and duplicate declarations are already deduped during generation (`mergeRules`). The reorder was cosmetic and was the sole source of nondeterminism.
- **Chose this over generating in canonical order.** Reproducing stylelint's property order in-process would mean owning a second ordering implementation coupled to `stylelint.config.js` — real maintenance to satisfy a convention nothing enforces on this directory. The order difference originates only from concatenating the base file's declarations with the component's during merge; both inputs are individually canonical.

## Motivation and context

Removes a recurring source of dirty working trees and accidental file churn in unrelated PRs.

## Related issue(s)

- fixes [N/A — internal build-determinism bug]

## Author's checklist

- [x] I have reviewed at the Accessibility Practices for this feature.

### Manual review test cases

- [ ] _Build no longer dirties generated global CSS_
  1. From a clean checkout on this branch, run `yarn build`
  2. Run `git status`
  3. Expect no changes to `2nd-gen/packages/swc/stylesheets/global/*.css`

- [ ] **Keyboard** (required)
  - No focusable parts; this changes only build output (CSS property ordering in generated files). No runtime or DOM behavior changes; confirm no regressions in surrounding component examples.

  - No role, name, structure, or announcement changes — generated CSS is byte-equivalent in behavior; property order does not affect rendering or the cascade.


.changeset/global-css-deterministic-build.md
```diff
@@ -0,0 +1,7 @@
+---
+'@adobe/spectrum-wc': patch
```
Rajdeepc: You dont need changeset for this

Rajdeepc (changes requested):

.changeset/global-css-deterministic-build.md
rubencarvalho: updated!

cdransf (approved):

Rajdeepc (approved):

PR #6462, opened 2026-06-30
feat(response-status): adding agent steps to response status

rubencarvalho:
## Description
  Refactors `swc-response-status` toward the alternate ConvAI response status API design, centered on a single `status` property and explicit slotted content. This branch keeps the agentic response-status direction from the original implementation, but simplifies the public contract so consumers use:
  - `status`: `pending` | `active` | `complete` | `stopped`
  - `open`: controls the expandable step timeline
  - `accessible-label`: names the step timeline panel
  - `slot="label"`: visible header label
  - `slot="summary"`: optional fallback summary text
  - `swc-response-status-step`: light-DOM step descriptor with `status`, `slot="label"`, and `slot="description"`
  
### What's included
  - Adds `swc-response-status-step` as the step descriptor for execution timelines.
  - Removes the step-level `pending` state so provided steps always render.
  - Removes component-owned label dwell/queue timing; rapid update cadence is consumer-managed.
  - Keeps rolling header label animation for visible label changes.
  - Updates response-status docs, stories, and neighboring Conversational AI examples to use the `status` + `slot="label"` API.
  - Adds regression coverage for active-step header fallback, invalid step-status coercion, slotted label updates, and 
  composed toggle events.
  
## Motivation and context
  ConvAI needs a compact status surface that can show the current response lifecycle, roll through active execution labels, and optionally expand into an execution timeline. This PR narrows the API around the model we want consumers to author directly, instead of supporting multiple overlapping state models like `loading`, `phase`, `duration`, and plain-text reasoning modes. The goal is to make the response-status contract easier to document, test, and use before the component lands.
  
## Related issue(s)
  - addresses SWC-2134
 
## Test plan
  - `yarn workspace @adobe/spectrum-wc test response-status.test.ts`
  - `yarn test:a11y:2nd`
  - `yarn lint:ai`
  - `yarn build:2nd-gen`
  - Scoped ESLint, Stylelint, and Prettier checks on changed files
  - Pre-commit hooks
  
## Author's checklist
  - [x] I have included a changeset.
  - [x] I have included updated documentation.
  
## Reviewer's checklist
  - [ ] API shape is approved: `status`, `open`, `accessible-label`, label/summary slots, and step children
  - [ ] Automated tests cover the expected status, step, label, and toggle behavior
  - [ ] Accessibility behavior is approved for collapsed and expanded timelines
  - [ ] Design parity is acceptable against the Figma reference
  - [ ] VRTs are reviewed/approved if visual changes are expected
  
## Manual review test cases
  - [ ] **Active with expanded steps**
    1. Open Storybook → **Conversational AI → Response status**
    2. Use `status="active"`, `open`, and multiple `swc-response-status-step` children
    3. Expect animated three dots, active step label in the header, and timeline rows with connectors
  - [ ] **Complete collapsed**
    1. Set `status="complete"` with completed steps
    2. Expect the complete label/checkmark treatment and a disclosure control when steps exist
  - [ ] **Consumer-managed label cadence**
    1. Open the `LabelRoll` story
    2. Trigger the queued-label demo
    3. Expect the app/demo layer to pace label updates while the component only animates visible label changes
  
  - [ ] **Keyboard** — Tab to the disclosure button when steps exist; Enter/Space toggles `open`
  - [ ] **ARIA state** — `aria-expanded` and `aria-controls` reflect disclosure state
  - [ ] **Timeline semantics** — expanded step timeline exposes a labelled group and list/listitem structure
  - [ ] **No timeline case** — status without steps does not expose a dead disclosure control
  
## Notes
  - This branch intentionally builds on the existing response-status PR branch so reviewers can compare the alternate API design against the original implementation.
  - Unrelated local `.gitignore` / global stylesheet edits were left unstaged and are not included in this PR.


.changeset/soft-ravens-think.md
```diff
@@ -0,0 +1,9 @@
+---
+'@spectrum-web-components/2nd-gen': patch
```
rubencarvalho: no need to bump this

2nd-gen/packages/swc/patterns/conversational-ai/response-status/response-status-step/ResponseStatusStep.ts
```diff
@@ -0,0 +1,66 @@
...
+  /** Timeline state for connector icons. */
+  @property({ type: String, reflect: true })
+  public status: 'active' | 'complete' | 'stopped' = 'active';
+
+  public static override get styles(): CSSResultArray {
+    return [styles];
+  }
+
+  /** @internal */
+  private _handleSlotChange = (): void => {
```
rubencarvalho: Is it really slot change that we're handling here? maybe handleStatusChange?

2nd-gen/packages/swc/patterns/conversational-ai/response-status/response-status.css
```diff
@@ -85,27 +171,226 @@
...
+.swc-ResponseStatus--agentic {
+  --swc-response-status-roll-duration: 650ms;
+
+  gap: token("spacing-100");
+  inline-size: 100%;
+  max-inline-size: 540px;
+}
+
+.swc-ResponseStatus-row--processing,
+.swc-ResponseStatus-row--button.swc-ResponseStatus-row--processing {
```
rubencarvalho: why do we need both selectors here? aren't they equivalent?

2nd-gen/packages/swc/patterns/conversational-ai/response-status/response-status.css
```diff
@@ -85,27 +171,226 @@
...
+  padding-inline: token("spacing-100");
+  border-radius: token("corner-radius-medium-default");
+}
+
+.swc-ResponseStatus-row--complete.swc-ResponseStatus-row--button,
+.swc-ResponseStatus-row--stopped.swc-ResponseStatus-row--button {
```
rubencarvalho: Same question here about potential selector redundacy

2nd-gen/packages/swc/patterns/conversational-ai/response-status/response-status.mdx
```diff
@@ -11,45 +11,100 @@ import * as ResponseStatusStories from './stories/response-status.stories';
...
+```js
+status.addEventListener('swc-response-status-toggle', (event) => {
+  const { open } = event.detail;
+  status.open = open;
+});
+```
+
+### Label update cadence
+
+Header-label cadence is consumer-managed. If your app streams rapid status updates, debounce, throttle, or queue those updates in application code before setting `slot="label"`.
```
rubencarvalho: Could we tighten this guidance? Also, I'm not sure how prescriptive we want to be with the code snippet below. Maybe it's enough to justify it under the A11Y umbrella and add something along the lines of "ensure each label stays visible for at least 1s"

2nd-gen/packages/swc/patterns/conversational-ai/response-status/response-status.mdx
```diff
@@ -11,45 +11,100 @@ import * as ResponseStatusStories from './stories/response-status.stories';
...
+- The step timeline uses `role="group"` with an accessible name from the `accessible-label` attribute (defaults to “Execution steps”)
+- The ordered step list uses `role="list"` and each rendered step uses `role="listitem"`
+
+### Best practices
 
-- The loading row carries `role="status"`, so the status label is announced for screen readers
-- When reasoning content exists, the reasoning toggle uses `aria-expanded` and `aria-controls` to communicate panel state
-- The reasoning panel uses `role="group"` with `aria-label="Reasoning"`
+- Keep step labels concise and action-oriented.
+- Use `slot="description"` for supplemental detail, not required state text.
```
rubencarvalho: Every time I read "use..., not...", it just screams AI to me. Could we tighten this up?

2nd-gen/packages/swc/patterns/conversational-ai/response-status/ResponseStatus.ts
```diff
@@ -18,209 +18,673 @@ import { Chevron75Icon } from '@adobe/spectrum-wc/icon/elements/index.js';
...
+} from '../utils/icons/index.js';
+import {
+  ResponseStatusStep,
+  type ResponseStatusStepStatus,
+} from './response-status-step/ResponseStatusStep.js';
 
 import styles from './response-status.css';
 
+export type ResponseStatusStatus =
+  | 'pending'
```
rubencarvalho: Interesting how pending still exists here but not on the step. Response status can be pending but that means at least one status step will be active

2nd-gen/packages/swc/patterns/conversational-ai/response-status/ResponseStatus.ts
```diff
@@ -18,209 +18,673 @@ import { Chevron75Icon } from '@adobe/spectrum-wc/icon/elements/index.js';
...
+    // Engage the transition on the next frame so it animates from the settled
+    // position instead of jumping straight to the rolled state.
+    this._labelRollRaf = window.requestAnimationFrame(() => {
+      this._labelRollRaf = window.requestAnimationFrame(() => {
+        this._labelRollRaf = null;
+        this._rollEngaged = true;
+      });
+    });
+
+    this._labelRollTimer = window.setTimeout(() => {
```
rubencarvalho: I thought we wanted to remove this from the component responsibility

2nd-gen/packages/swc/patterns/conversational-ai/response-status/response-status-step/ResponseStatusStep.ts
TarunAdobe: hmmm handleStatusChange is better !!

2nd-gen/packages/swc/patterns/conversational-ai/response-status/response-status.css
TarunAdobe: not equivalent bt we dont need both right... the last one is good for all cases keeping this :  .swc-ResponseStatus-row--processing

2nd-gen/packages/swc/patterns/conversational-ai/response-status/response-status.mdx
TarunAdobe: tighten as in reduce and demote or?

2nd-gen/packages/swc/patterns/conversational-ai/response-status/ResponseStatus.ts
TarunAdobe: Yus exactly that.

2nd-gen/packages/swc/patterns/conversational-ai/response-status/ResponseStatus.ts
TarunAdobe: oops i thought i removed it. yes we dont want this to be a component respo at it now

2nd-gen/packages/swc/patterns/conversational-ai/response-status/ResponseStatus.ts
```diff
@@ -18,209 +18,692 @@ import { Chevron75Icon } from '@adobe/spectrum-wc/icon/elements/index.js';
...
- * While **`loading`** is `true`, reasoning is not shown.
- *
  * @element swc-response-status
- * @slot - Optional reasoning content. Disclosure UI is shown only when slot has content
- * and `loading` is `false`; content is visible when `open` is `true`.
- * If slot reasoning content is removed while `open=true`, the component collapses itself
- * by setting `open=false`.
- * @fires swc-response-status-toggle - Dispatched when the user expands or collapses the reasoning panel.
+ * @slot label - Header row label. Falls back to the active step label.
+ * @slot summary - Optional secondary summary text for lifecycle states.
```
rubencarvalho: When is `summary` actually rendering? I can't rly see it... Do we need it?

2nd-gen/packages/swc/patterns/conversational-ai/response-status/response-status-step/ResponseStatusStep.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  public static override get styles(): CSSResultArray {
+    return [styles];
+  }
+
+  /** @internal */
+  private _notifyStepChange = (): void => {
+    this.dispatchEvent(
+      new CustomEvent('swc-response-status-step-change', {
+        bubbles: true,
```
rubencarvalho: Iiiiinteresting... a bit of a nit, but if you see, this bubbles but isn't composed, so it climbs past the host and out into whatever the consumer wraps us in. we're using this just as internal parent/child communication (and the name looks public), so we could add an `event.stopPropagation()` in the host's handler so it stops at `<swc-response-status>`.

2nd-gen/packages/swc/patterns/conversational-ai/response-status/response-status.mdx
```diff
@@ -11,45 +11,81 @@ import * as ResponseStatusStories from './stories/response-status.stories';
...
+## Behaviors
+
+### Handling events
+
+`<swc-response-status>` emits **`swc-response-status-toggle`** when users open or close the step timeline. The event **bubbles** and is **composed**.
+
+`event.detail.open` contains the next disclosure state.
+
+```js
+status.addEventListener('swc-response-status-toggle', (event) => {
```
rubencarvalho, line 58: eeerm... the `status.open = open` in this example reads like the component is controlled, but `_handleToggle` already flips `this.open` before firing. Might be worth either removing it or saying smt like "the event is notify-only, component manages its own `open`" so no1 wires up a controlled pattern that isn't there.

2nd-gen/packages/swc/patterns/conversational-ai/response-status/ResponseStatus.ts
```diff
@@ -18,209 +18,692 @@ import { Chevron75Icon } from '@adobe/spectrum-wc/icon/elements/index.js';
...
+      element instanceof ResponseStatusStep ||
+      element.localName === 'swc-response-status-step'
+    );
+  }
+
+  private _readStepElement(element: Element): ResponseStatusStepData {
+    const step = element as ResponseStatusStep;
+    const label = this._readLightDomNamedSlotText(element, 'label');
+    const rawStatus =
+      step.status ||
```
rubencarvalho: Nit: the `getAttribute('status')` fallback is not needed? `step.status` is a reflected prop that defaults to `'active'`, so it should never be empty. prob safe to delete, but nit nonetheless

rubencarvalho (commented):
Image

Can we also make sure to clean up the SB controls?

2nd-gen/packages/swc/patterns/conversational-ai/response-status/ResponseStatus.ts
```diff
@@ -18,209 +18,692 @@ import { Chevron75Icon } from '@adobe/spectrum-wc/icon/elements/index.js';
...
+  private _summarySlotText = '';
 
-  /**
-   * Status row label text shown while `loading=false`.
-   */
+  /** @internal */
+  @state()
+  private _displayedLabel = '';
+
+  /** @internal */
```
rubencarvalho: do we need to mark states as internal?

2nd-gen/packages/swc/patterns/conversational-ai/response-status/response-status.css
```diff
@@ -35,36 +35,143 @@
...
+}
+
+.swc-ResponseStatus-headerTrailStrip {
+  display: flex;
+  flex-direction: column;
+  inline-size: fit-content;
+  max-inline-size: 100%;
+  transform: translateY(0);
+  transition:
+    transform var(--swc-response-status-roll-duration, 650ms) token("animation-ease-in-out"),
```
rubencarvalho: var(--swc-response-status-roll-duration, 650ms)` looks like it is customizable for customers, but if anyone overrides it the JS timer is hard coded so they'll go out of sync which prob breaks the whole animation. if we really want to keep it public, we can read the computed value in JS

2nd-gen/packages/swc/patterns/conversational-ai/response-status/ResponseStatus.ts
```diff
@@ -18,209 +18,692 @@ import { Chevron75Icon } from '@adobe/spectrum-wc/icon/elements/index.js';
...
+          </span>
+        </span>
+      `;
+    }
+
+    const stripClass = this._rollEngaged
+      ? 'swc-ResponseStatus-headerTrailStrip swc-ResponseStatus-headerTrailStrip--rolling'
+      : 'swc-ResponseStatus-headerTrailStrip';
+
     return html`
```
rubencarvalho, line 419: can you verify what happens? when there are no steps + status is `pending`/`active`, the header row is a `role="status"` live region (on the `_renderHeader`). During a roll we render both the from-line and the to-line inside it, then collapse to one, so I _think_ a screen reader announces "old-label new-label" mid-roll and then "new-label" again on settle. can you confirm?? we may need to add aria-hidden to remediate IDK

2nd-gen/packages/swc/patterns/conversational-ai/response-status/ResponseStatus.ts
```diff
@@ -18,209 +18,692 @@ import { Chevron75Icon } from '@adobe/spectrum-wc/icon/elements/index.js';
...
+    }
+
+    this._rollFromLabel = this._displayedLabel;
+    this._rollToLabel = target;
+    this._rollActive = true;
+    this._rollEngaged = false;
+
+    // Engage the transition on the next frame so it animates from the settled
+    // position instead of jumping straight to the rolled state.
+    this._labelRollRaf = window.requestAnimationFrame(() => {
```
rubencarvalho, line 376: Nit : the 650ms snap timer starts when the roll is scheduled, but the CSS transition only engages after this double-rAF (`_rollEngaged`), so it kicks off ~2 frames later 😅

2nd-gen/packages/swc/patterns/conversational-ai/response-status/ResponseStatus.ts
```diff
@@ -18,209 +18,692 @@ import { Chevron75Icon } from '@adobe/spectrum-wc/icon/elements/index.js';
...
+  public override connectedCallback(): void {
+    super.connectedCallback();
+    this.addEventListener(
+      'swc-response-status-step-change',
+      this._handleStepChildChange
+    );
+
+    // Slotchange only fires when assigned nodes are added or removed, not when
+    // their text mutates. Observe the light DOM so slotted text updates render.
+    this._contentObserver = new MutationObserver(() => {
```
rubencarvalho: soooo I was looking at our `ObserveSlotText` mixin in core, but it only exposes a `slotHasContent` boolean and is single-slot, so it doesn't really fit 1:1, I know we need the actual text across the named slots *and* every step child.

BUT! the reusable primitive it's built on does fit though: `MutationController` from `@lit-labs/observers/mutation-controller.js` (already a dep of core). It would manage the observer lifecycle for us (u know, observes on `hostConnected`, `disconnect()`s on `hostDisconnected`) so it'd delete the `_contentObserver` field, the manual `.observe()` here, and the teardown in `disconnectedCallback`:

```ts
  config: { characterData: true, childList: true, subtree: true },
  callback: () => this._syncSlotContent(),
});
```

Two caveats: (1) `swc` would need to declare `@lit-labs/observers` as a direct dep (it's only in `core`), and (2) the controller re-fires its callback on every host update, so `_syncSlotContent` runs after each render instead of only on real mutations, but I think that's harmless here since it's guarded by `_stepsEqual`/string compares, just noting it. Non-blocking, but it does feels cleaner than hand-rolling the observer to me

2nd-gen/packages/swc/patterns/conversational-ai/response-status/ResponseStatus.ts
```diff
@@ -18,209 +18,692 @@ import { Chevron75Icon } from '@adobe/spectrum-wc/icon/elements/index.js';
...
+      'swc-response-status-step-change',
+      this._handleStepChildChange
+    );
+    this._contentObserver?.disconnect();
+    this._contentObserver = undefined;
+    this._clearLabelRollTimers();
+    super.disconnectedCallback();
+  }
+
+  private _isValidStatus(status: string): status is ResponseStatusStatus {
```
rubencarvalho, line 164: nit, I think this can be cleaner here: make the runtime list the source of truth and derive the type from it, so the union, this guard, and any arg options can't drift apart:

```ts
const RESPONSE_STATUSES = ['pending', 'active', 'complete', 'stopped'] as const;
export type ResponseStatusStatus = (typeof RESPONSE_STATUSES)[number];

private _isValidStatus(status: string): status is ResponseStatusStatus {
  return (RESPONSE_STATUSES as readonly string[]).includes(status);
}
```

2nd-gen/packages/swc/patterns/conversational-ai/response-status/ResponseStatus.ts
```diff
@@ -18,209 +18,692 @@ import { Chevron75Icon } from '@adobe/spectrum-wc/icon/elements/index.js';
...
+        (node) =>
+          node.nodeType === Node.TEXT_NODE ||
+          (node instanceof Element && !node.getAttribute('slot'))
+      )
+      .map((node) => node.textContent?.trim() ?? '')
+      .filter(Boolean)
+      .join(' ');
+  }
+
+  private _readShadowSlotText(slotName: string): string {
```
rubencarvalho: Can we use Lit `@queryAssigned*` decorators? This `shadowRoot.querySelector('slot[name=...]')` + `.assignedNodes()` (and the `.assignedElements()` in `_readSteps`, plus the `.swc-ResponseStatus-content-slot` querySelector) all reimplement what the decorators give directly.

Steps would collapse to roughly:
```ts
@queryAssignedElements({ selector: 'swc-response-status-step', flatten: true })
private _stepEls!: ResponseStatusStep[];
```

2nd-gen/packages/swc/patterns/conversational-ai/response-status/ResponseStatus.ts
```diff
@@ -18,209 +18,692 @@ import { Chevron75Icon } from '@adobe/spectrum-wc/icon/elements/index.js';
...
+
   protected override firstUpdated(): void {
-    this._syncReasoningContent();
+    this._displayedLabel = this._getHeaderLabel();
   }
 
-  private _handleToggle(): void {
-    if (this.loading || !this._hasReasoningContent) {
-      return;
+  protected override updated(): void {
```
rubencarvalho: `updated()` calls `_applyLabelRoll()`, which sets `@state` (`_displayedLabel`, `_rollActive`, …) *after* render... so every label change forces a second render pass. can we use `willUpdate()`?
The rAF-to-engage-the-CSS-transition step prob still has to stay post-render, but the "decision" (did the target change? snap vs. start a roll?) can be in `willUpdate` so we're not double-rendering on every update.

2nd-gen/packages/swc/patterns/conversational-ai/response-status/ResponseStatus.ts
```diff
@@ -18,209 +18,692 @@ import { Chevron75Icon } from '@adobe/spectrum-wc/icon/elements/index.js';
...
 
+  /** Accessible name for the step list panel. */
+  @property({ type: String, attribute: 'accessible-label' })
+  public accessibleLabel = '';
+
+  /** @internal */
+  private _contentObserver?: MutationObserver;
+
+  /** @internal */
+  private _rollToLabel = '';
```
rubencarvalho: `_rollToLabel` is a plain field but it's read in `render()` (`_renderLabel` / `_currentVisibleLabel`). It only re-renders because it's always set alongside `_rollFromLabel`/`_rollActive`, which *are* `@state` — so the paint relies on that co-mutation coincidence rather than its own reactivity. Anything read in `render()` should be reactive; suggest making this `@state` too for consistency with its siblings a few lines up.

2nd-gen/packages/swc/patterns/conversational-ai/response-status/response-status-step/ResponseStatusStep.ts
```diff
@@ -0,0 +1,66 @@
...
+    if (changed.has('status')) {
+      this._notifyStepChange();
+    }
+  }
+
+  protected override render(): TemplateResult {
+    return html`
+      <slot name="label" hidden @slotchange=${this._notifyStepChange}></slot>
+      <slot
+        name="description"
```
rubencarvalho: if i understand this correctly, these three `<slot>`s never render anything `:host` is `display: none`, so the step is a slot-based API marker (nice) and the parent draws the actual timeline from its light-DOM text. `display: none` on the host is correct, but what seems redundant to me is these slots exist only to fire `swc-response-status-step-change` via `@slotchange`. But the parent already runs a MutationObserver over its whole light-DOM subtree (`characterData`/`childList`/`subtree`), which catches the same add/remove/text changes... so this is a second change-detection path for what the parent component already sees/tracks. Can we use one or the other??

2nd-gen/packages/swc/patterns/conversational-ai/response-status/response-status.css
```diff
@@ -85,27 +192,211 @@
...
+    color token("animation-duration-200") token("animation-ease-in-out"),
+    opacity token("animation-duration-200") token("animation-ease-in-out");
+}
+
+.swc-ResponseStatus-step-line {
+  flex: 1 1 auto;
+  inline-size: 1px;
+  min-block-size: token("spacing-100");
+  margin-block-start: token("spacing-50");
+  background: token("gray-300");
```
rubencarvalho, line 291: No `@media (forced-colors: active)` block anywhere in this file. in forced-colors the step connector line or the dots likely disappear / or become less easy to see? Can you double check?

2nd-gen/packages/swc/patterns/conversational-ai/response-status/response-status.css
```diff
@@ -35,36 +35,143 @@
...
+  min-inline-size: 0;
+  overflow: hidden;
+  text-overflow: ellipsis;
+}
+
+.swc-ResponseStatus-headerTrailStrip--rolling .swc-ResponseStatus-headerTrailLine:first-child {
+  visibility: hidden;
+}
+
+@media (prefers-reduced-motion: reduce) {
```
rubencarvalho: Can we move this `prefers-reduced-motion` block to the bottom of the file? There's already a second one there, so maybe we can consolidate both in the same bottom blcok?

2nd-gen/packages/swc/patterns/conversational-ai/response-status/response-status.css
```diff
@@ -35,36 +35,143 @@
...
   align-items: center;
   block-size: var(--swc-response-status-row-block-size, 32px);
+  padding-inline: token("spacing-100");
 }
 
 /* When the row is a button (reasoning toggle) */
 .swc-ResponseStatus-row--button {
   inline-size: fit-content;
+  max-inline-size: 580px;
   padding: var(--swc-sources-toggle-padding, 3px) token("spacing-100");
```
rubencarvalho: Are we sure we are following our CSS guide for CSS custom private/public prop naming? Also, `--swc-sources-toggle-padding`... `message-sources.css` owns the identical `padding: var(--swc-sources-toggle-padding, 3px) token("spacing-100")`. Should we even keep these as public typed props? consumer overriding it to tune this header would move padding on Message Sources too. Should be a response-status-owned name (`--_swc-response-status-toggle-padding`) or just a token. The `3px` fallback is also a raw magic number? 😅

2nd-gen/packages/swc/patterns/conversational-ai/response-status/response-status.css
```diff
@@ -35,36 +35,143 @@
   gap: token("spacing-100");
   align-items: center;
   block-size: var(--swc-response-status-row-block-size, 32px);
```
rubencarvalho: to summarise: `--swc-*` knobs (`--swc-response-status-row-block-size` here, plus `--swc-response-status-label-line-size` and `--swc-response-status-roll-duration` below) read as internal constants with hardcoded fallbacks (`32px`, `1.33em`, `650ms`), but per `02_custom-properties.md` the `--swc-*` prefix means "exposed for consumer overrides" (and none are documented as API!). Either rename to `--_swc-*` if internal, or document them as real API

2nd-gen/packages/swc/patterns/conversational-ai/response-status/response-status-step/ResponseStatusStep.ts
TarunAdobe: hmmm good point !!

2nd-gen/packages/swc/patterns/conversational-ai/response-status/ResponseStatus.ts
TarunAdobe: nea its ai hallucinated

2nd-gen/packages/swc/patterns/conversational-ai/response-status/response-status.mdx
TarunAdobe: i can do a bit of work on the docs here.!

2nd-gen/packages/swc/patterns/conversational-ai/response-status/ResponseStatus.ts
TarunAdobe: NO

2nd-gen/packages/swc/patterns/conversational-ai/response-status/response-status.css
TarunAdobe: hmmm so i can hardcode css value instead of a var if thats okay

2nd-gen/packages/swc/patterns/conversational-ai/response-status/ResponseStatus.ts
TarunAdobe: so an aria-hidden on the old (from-line) fixes thissss

2nd-gen/packages/swc/patterns/conversational-ai/response-status/ResponseStatus.ts
TarunAdobe: lol okay i moved the timer to start after _rollEngaged = true

2nd-gen/packages/swc/patterns/conversational-ai/response-status/ResponseStatus.ts
TarunAdobe: hmmmm okay i added @lit-labs/observers: 2.0.2 as a direct @adobe/spectrum-wc dependency and replaced the custom observer with mutation controller

2nd-gen/packages/swc/patterns/conversational-ai/response-status/ResponseStatus.ts
TarunAdobe: this information needs to be in the skills file in our repo lol. i have had this issue where cursor almost never uses queryAssignedElements until you review lol.

2nd-gen/packages/swc/patterns/conversational-ai/response-status/ResponseStatus.ts
TarunAdobe: made it a @state and now this is fixed.

2nd-gen/packages/swc/patterns/conversational-ai/response-status/response-status-step/ResponseStatusStep.ts
```diff
@@ -31,36 +38,9 @@ export type ResponseStatusStepStatus = 'active' | 'complete' | 'stopped';
 export class ResponseStatusStep extends SpectrumElement {
   /** Timeline state for connector icons. */
   @property({ type: String, reflect: true })
-  public status: 'active' | 'complete' | 'stopped' = 'active';
+  public status: ResponseStatusStepStatus = 'active';
```
rubencarvalho: `ResponseStatusStepStatus` reads funny 😆 keep it inline maybe just so it also shows up expanded on the API table? I know it is suboptimal, but better than this?

2nd-gen/packages/swc/patterns/conversational-ai/response-status/response-status.css
```diff
@@ -103,10 +103,10 @@
   gap: token("spacing-100");
   align-items: center;
   max-inline-size: 100%;
-  block-size: var(--swc-response-status-label-line-size, 1.33em);
+  block-size: var(--_swc-response-status-label-line-size, 1.33em);
   white-space: nowrap;
   opacity: 1;
-  transition: opacity calc(var(--swc-response-status-roll-duration, 650ms) / 2) token("animation-ease-in-out");
+  transition: opacity 325ms token("animation-ease-in-out");
```
rubencarvalho, line 109: do we have tokens for these?

TarunAdobe: noooooooooo

2nd-gen/packages/swc/package.json
```diff
@@ -47,9 +47,9 @@
     "./link.css": {
       "default": "./dist/link.css"
     },
-    "./patterns/*": {
-      "types": "./dist/patterns/*/index.d.ts",
-      "import": "./dist/patterns/*/index.js"
+    "./patterns/*/*": {
+      "types": "./dist/patterns/*/*/index.d.ts",
```
rubencarvalho: Are we sure about this? won't it break?

2nd-gen/packages/swc/patterns/conversational-ai/utils/icons/index.ts
```diff
@@ -69,6 +69,36 @@ export const ThreeDotsIcon = (): TemplateResult => html`
...
+  <svg
+    xmlns="http://www.w3.org/2000/svg"
+    width="16"
+    height="16"
+    viewBox="0 0 16 16"
+    fill="none"
+  >
+    <path
+      d="M8.8484 7.99996L10.6242 6.22417C10.8586 5.9898 10.8586 5.61011 10.6242 5.37573C10.3898 5.14136 10.0101 5.14136 9.77574 5.37573L7.99996 7.15152L6.22417 5.37573C5.9898 5.14136 5.61011 5.14136 5.37573 5.37573C5.14136 5.61011 5.14136 5.9898 5.37573 6.22417L7.15152 7.99996L5.37573 9.77574C5.14136 10.0101 5.14136 10.3898 5.37573 10.6242C5.49292 10.7414 5.64605 10.8 5.79995 10.8C5.95385 10.8 6.10698 10.7414 6.22417 10.6242L7.99995 8.8484L9.77573 10.6242C9.89292 10.7414 10.046 10.8 10.1999 10.8C10.3539 10.8 10.507 10.7414 10.6242 10.6242C10.8585 10.3898 10.8585 10.0101 10.6242 9.77574L8.8484 7.99996Z"
+      fill="#292929"
```
rubencarvalho: Won't this `fill` break in HCM?

2nd-gen/packages/swc/patterns/conversational-ai/response-status/ResponseStatus.ts
```diff
@@ -95,132 +446,210 @@ export class ResponseStatus extends SpectrumElement {
...
-            aria-hidden="true"
-          ></swc-progress-circle>
-        </span>
-        <span class="swc-ResponseStatus-label">${label}</span>
-      </div>
-    `;
-  }
+  private _renderHeader(showDisclosure: boolean): TemplateResult {
+    const label = this._currentVisibleLabel();
+    const statusRole =
```
rubencarvalho, line 507: a bit of a nit, but could you check this? invalid statuses already fall back visually to active through the default label and dots, but this check uses the raw value and therefore omits role="status". could we derive a single validated status, falling back to active, and use it consistently for the label, icon, classes, and role? a bit of an edge case as I can only see this happening if a consumer passes an unvalidated runtime value (idk like from an API). Alternatively, if invalid values intentionally have undefined behavior, the current implementation is fine.

2nd-gen/packages/swc/package.json
TarunAdobe: No! this is gonna break. i dont think two * in one export key is valid? anyhow i am aligning this with other existing patterns (utils) ./dist/patterns/*/index.js

rubencarvalho (commented):
I approve 👍 , thanks @TarunAdobe

TarunAdobe (approved):

aramos-adobe (approved):

PR #6466, opened 2026-07-02
fix(tabs): correct selection indicator position in RTL

Rajdeepc:
## Description

`<swc-tabs>`'s animated selection indicator was rendering in the wrong horizontal position when the tab list is rendered in a right-to-left (RTL) context (`dir="rtl"` or an RTL-inheriting ancestor).

**Root cause:** the indicator's resting position is set in CSS via the logical property `inset-inline-start: 0`, which resolves to `right: 0` in RTL. However, `updateSelectionIndicator()` in `Tabs.base.ts` always computed the `translateX` offset as the distance from the container's **left** edge (`tabRect.left - listRect.left`), regardless of direction. In RTL, the indicator's anchor point (right edge) and the JS offset (measured from the left) disagreed, so the indicator was translated to the wrong spot instead of aligning under the selected tab.

**Fix:**
- `Tabs.base.ts`: detect the current writing direction via `getComputedStyle(this).direction === 'rtl'` (the same pattern already used by `placement-controller.ts`, `Tooltip.base.ts`, and `focusgroup-navigation-controller.ts`), and compute the offset from the container's right edge (`tabRect.right - listRect.right`) when RTL.
- `tabs.css`: added a `:dir(rtl)` override so `transform-origin` flips to `right center`, matching the indicator's RTL-anchored resting position (mirrors the existing `:dir(rtl)` pattern in `accordion-item.css`).

Vertical tabs are unaffected, since block-axis layout doesn't change with `dir`.

## Motivation and context

The horizontal selection indicator misalignment is visible on the `components-tabs-vrt--permutations` Storybook story and breaks the visual affordance for which tab is selected in any RTL locale.

## Related issue(s)

No related issue filed.

N/A — see the `components-tabs-vrt--permutations` Storybook story (RTL permutation) before/after this change for a visual comparison.

## Author's checklist

## Reviewer's checklist

### Manual review test cases

- [ ] _Selection indicator aligns with the selected tab in RTL_
  1. Open https://swcpreviews.z13.web.core.windows.net/pr-6466/docs/second-gen-storybook/?path=/story/components-tabs--playground&globals=textDirection:rtl
  2. Select each tab in turn
  3. Expect the indicator bar to sit directly under/beside the currently selected tab, with no gap or overlap, matching its LTR counterpart mirrored

- [ ] _No regression in LTR or vertical orientation_
  1. Open the default (LTR) https://swcpreviews.z13.web.core.windows.net/pr-6466/docs/second-gen-storybook/?path=/story/components-tabs--playground&globals=textDirection:ltr
  2. Select each tab in turn
  3. Expect the indicator to continue tracking the selected tab exactly as before this change

### Device review

- [ ] Did it pass in Desktop?

  The selection indicator is a non-interactive, presentational element (`role="presentation"`, `pointer-events: none`) that is not part of the focus order. This change only affects its visual transform in RTL; it does not alter tab list keyboard navigation (Arrow keys, Home/End, Enter/Space).
  1. Go to any `<swc-tabs>` story with `dir="rtl"` applied
  2. Tab into the tab list and use Arrow Left/Right (already RTL-aware via `getNavigationDelta`) to move between tabs
  3. Expect focus and selection behavior to be unchanged; only the indicator's visual position should differ from before this fix

  The selection indicator carries no accessible name, role, or state (`role="presentation"`) and is not announced. This is a purely visual fix with no screen reader-observable behavior.
  1. Go to any `<swc-tabs>` story with `dir="rtl"` applied
  2. Navigate through tabs with a screen reader active (e.g. VoiceOver)
  3. Expect tab name, selected state (`aria-selected`), and panel announcements to be unchanged before and after this fix

### DIFF
--- a/2nd-gen/packages/core/components/tabs/Tabs.base.ts
+++ b/2nd-gen/packages/core/components/tabs/Tabs.base.ts
@@ -562,9 +562,12 @@ export abstract class TabsBase extends SpectrumElement {
     const listRect = tablist.getBoundingClientRect();
 
     if (this._direction === 'horizontal') {
-      const left = tabRect.left - listRect.left;
+      const isRtl = getComputedStyle(this).direction === 'rtl';
+      const offset = isRtl
+        ? tabRect.right - listRect.right
+        : tabRect.left - listRect.left;
       const scale = tabRect.width / TabsBase.INDICATOR_BASE_SIZE;
-      this.selectionIndicatorStyle = `transform: translateX(${left}px) scaleX(${scale})`;
+      this.selectionIndicatorStyle = `transform: translateX(${offset}px) scaleX(${scale})`;
     } else {
       const top = tabRect.top - listRect.top;
       const scale = tabRect.height / TabsBase.INDICATOR_BASE_SIZE;
--- a/2nd-gen/packages/core/components/tabs/Tabs.base.ts
+++ b/2nd-gen/packages/core/components/tabs/Tabs.base.ts
@@ -562,7 +562,7 @@ export abstract class TabsBase extends SpectrumElement {
     const listRect = tablist.getBoundingClientRect();
 
     if (this._direction === 'horizontal') {
-      const isRtl = this.dir === 'rtl';
+      const isRtl = getComputedStyle(this).direction === 'rtl';
       const offset = isRtl
         ? tabRect.right - listRect.right
         : tabRect.left - listRect.left;
@@ -582,6 +582,19 @@ export abstract class TabsBase extends SpectrumElement {
 
   /** @internal */
   private _resizeObserver?: ResizeObserver;
+
+  /**
+   * @internal
+   *
+   * Watches for `dir` attribute changes so the indicator recalculates
+   * when writing direction flips at runtime. `getComputedStyle` in
+   * `updateSelectionIndicator` resolves inherited direction correctly,
+   * but nothing else observes it: a `dir` flip on the document root or
+   * an ancestor doesn't resize this element, so the `ResizeObserver`
+   * below won't catch it. Mirrors the `dir`-watching pattern in
+   * `placement-controller.ts`.
+   */
+  private _directionObserver?: MutationObserver;
 
   // ───────────────────────────────────
   //     LIFECYCLE
@@ -649,6 +662,20 @@ export abstract class TabsBase extends SpectrumElement {
       this.updateSelectionIndicator();
     });
     this._resizeObserver.observe(this);
+
+    this._directionObserver = new MutationObserver(() => {
+      this.updateSelectionIndicator();
+    });
+    const observeDir = (node: Element | null | undefined) => {
+      if (node) {
+        this._directionObserver?.observe(node, {
+          attributes: true,
+          attributeFilter: ['dir'],
+        });
+      }
+    };
+    observeDir(this.ownerDocument.documentElement);
+    observeDir(this.closest('[dir]'));
   }
 
   public override disconnectedCallback(): void {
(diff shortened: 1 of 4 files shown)


2nd-gen/packages/core/components/tabs/Tabs.base.ts
```diff
@@ -562,9 +562,12 @@ export abstract class TabsBase extends SpectrumElement {
     const listRect = tablist.getBoundingClientRect();
 
     if (this._direction === 'horizontal') {
-      const left = tabRect.left - listRect.left;
+      const isRtl = getComputedStyle(this).direction === 'rtl';
```
rubencarvalho, line 565: you can access the `this.dir === 'rtl'` getter instead 😄

rubencarvalho (approved):
leggo!

2nd-gen/packages/core/components/tabs/Tabs.base.ts
```diff
@@ -562,9 +562,12 @@ export abstract class TabsBase extends SpectrumElement {
     const listRect = tablist.getBoundingClientRect();
 
     if (this._direction === 'horizontal') {
-      const left = tabRect.left - listRect.left;
+      const isRtl = this.dir === 'rtl';
```
blunteshwar: I think this still needs to handle direction changes after initial render. The new offset is computed only when `updateSelectionIndicator()` runs, but `dir`/inherited direction is not reactive here. If a consumer flips from LTR to RTL after the tabs have rendered, the CSS `:dir(rtl)` rule updates the indicator anchor immediately, while the inline `translateX(...)` value remains based on the old LTR geometry until some unrelated event triggers a recalculation.

Could we add an explicit recalculation path for `dir` changes, plus a regression test that toggles `dir` from `ltr` to `rtl` with a non-first selected tab and verifies the indicator position updates?

2nd-gen/packages/core/components/tabs/Tabs.base.ts
Rajdeepc: Switched to `this.dir` per your suggestion, but while addressing Piyush's follow-up comment below I found it doesn't account for direction inherited from an ancestor or `<html dir="rtl">`, which is how most consumers actually set RTL (`this.dir` only reflects an attribute set on the tabs element itself). Reverted to `getComputedStyle(this).direction`, which matches the documented rationale in `focusgroup-navigation-controller.ts` for the same reason.

Thanks for flagging the consistency angle though; the existing `this.dir` check in `getNavigationDelta` looks like it has this same inherited-direction gap. That's outside this PR's scope, but worth a follow-up.

2nd-gen/packages/core/components/tabs/Tabs.base.ts
Rajdeepc: Good catch, thanks. Two changes to address this:

- Added a `MutationObserver` that watches the `dir` attribute on `document.documentElement` and the nearest `[dir]` ancestor (mirroring the existing pattern in `placement-controller.ts`) and recalculates the indicator on change, so a runtime direction flip no longer waits on an unrelated resize or font-load event.
- Reverted the `isRtl` check back to `getComputedStyle(this).direction` (see reply to Ruben above) since `this.dir` doesn't resolve inherited direction either, which your scenario would also hit if the flip happened on an ancestor rather than the tabs element itself.

Also added `SelectionIndicatorDirectionChangeTest` in tabs.test.ts: toggles an ancestor's `dir` from `ltr` to `rtl` with a non-first tab (`tab-id="2"`) selected and asserts the indicator's style changes.

blunteshwar (approved):

PR #6477, opened 2026-07-07
chore(workflow): make changeset grep tolerant of no matches

rubencarvalho:
## Description

The `publish.yml` workflow runs under `bash -e`. In the **Check for changesets** step, when a changeset file contains no `@spectrum-web-components/` mention, `grep -oP` exits with status 1, which under `-e` aborts the step and fails the job. Logs show `Found N changesets` immediately followed by a non-zero exit.

This guards the `grep` calls with `|| true` so a no-match result yields a zero exit and the loop continues, correctly detecting whether any non-core (1st-gen) package is referenced.

## Motivation and context

The publish job was failing on changesets that don't list package names.

https://github.com/adobe/spectrum-web-components/actions/runs/28891039549/job/85703536379

## Related issue(s)

- fixes [N/A]

## Author's checklist

- [ ] I have reviewed at the Accessibility Practices for this feature.

> CI-only change (GitHub Actions workflow); no package code affected, so no changeset. Verified locally by running the loop under `bash -e` with a package-free changeset — the step no longer aborts.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -49,8 +49,8 @@ jobs:
             HAS_1ST_GEN=false
             for file in .changeset/*.md; do
               if [ "$(basename "$file")" != "README.md" ]; then
-                PACKAGES=$(grep -oP "@spectrum-web-components/[a-z-]+" "$file")
-                NON_CORE=$(echo "$PACKAGES" | grep -v "^@spectrum-web-components/core$")
+                PACKAGES=$(grep -oP "@spectrum-web-components/[a-z-]+" "$file" || true)
+                NON_CORE=$(echo "$PACKAGES" | grep -v "^@spectrum-web-components/core$" || true)
                 if [ -n "$NON_CORE" ]; then
                   HAS_1ST_GEN=true
                   break


cdransf (approved):

.github/workflows/publish.yml
```diff
@@ -49,8 +49,8 @@ jobs:
             HAS_1ST_GEN=false
             for file in .changeset/*.md; do
               if [ "$(basename "$file")" != "README.md" ]; then
-                PACKAGES=$(grep -oP "@spectrum-web-components/[a-z-]+" "$file")
-                NON_CORE=$(echo "$PACKAGES" | grep -v "^@spectrum-web-components/core$")
+                PACKAGES=$(grep -oP "@spectrum-web-components/[a-z-]+" "$file" || true)
+                NON_CORE=$(echo "$PACKAGES" | grep -v "^@spectrum-web-components/core$" || true)
```
caseyisonit, line 53: wouldnt this make NON_CORE true meaning that the coniditional after this will always trigger?

rubencarvalho: It does look confusing, hence why I dislike lots of bashing in scripts. `|| true` only changes grep's exit code (so set -e doesn't kill the step on a no-match), it doesn't add any output. `NON_CORE` still captures only grep's matching lines, so it's empty unless a non-core package is actually present.

rubencarvalho: For example:
```sh
NON_CORE=$(echo "$PACKAGES" | grep -v "^@spectrum-web-components/core$" || true)
```
behaves like this:

* If grep -v finds matching lines:
    * NON_CORE contains those lines.
    * Exit status is 0.
* If grep -v finds no matching lines:
    * NON_CORE is an empty string.
    * grep exits with 1.
    * true runs, making the overall command succeed.
    * NON_CORE is still empty.

caseyisonit: ew i dont love it but i get it

caseyisonit (approved):

.github/workflows/publish.yml
Rajdeepc: @rubencarvalho A single `grep -qP` with a negative lookahead should have been the cleanest solution here.

PR #6484, opened 2026-07-08
feat(actiongroup): full fidelity styles

rise-erpelding:
## Description

Phase 5 (styling / full fidelity) of the `swc-action-group` migration, per SWC-2216.

- **Stories:** Extended `action-group.stories.ts` to full Phase 5 scope: `Anatomy`, `Sizes`, `Orientations`, `Static colors`, `Disabled`, `Compact` (horizontal and vertical shown side by side), `Quiet` (demonstrates the compact-join being disabled when both are set), `Justified`.
- **`action-group.css`:** Built out from the Phase 5 stub — flex layout with size-based gap (`spacing-100` default, `spacing-75` at `xs`/`s`, translated from `spectrum-two`'s `actiongroup/index.css`); `orientation="vertical"` column stacking; `justified` equal-fill via `flex: 1 1 0%`; `compact` border-join implemented via a CSS custom-property cascade onto `--swc-action-button-border-*-radius` (interior corners zeroed per child position, outer corners left at the button's natural radius); compact has no effect when `quiet` is also set, matching 1st-gen. Focus-ring stacking in compact mode uses `:focus-within` rather than `:focus-visible` — `delegatesFocus` retargets `:focus`/`:focus-within` onto the custom-element host in Chromium but not `:focus-visible`, so the original `:focus-visible` selector never matched and the z-index escalation was dead, letting a focused button's ring get covered by an adjacent button's border (resolves SWC-1342).
- **`ActionGroup.ts`:** Wrapped the slot in an internal `.swc-ActionGroup` element per the migration plan's Phase 5 checklist; added the `@cssprop --swc-action-group-gap` doc tag.
- **`action-button.css` / `ActionButton.ts` (prerequisite, per migration plan):** Exposed four logical corner-radius custom properties (`--swc-action-button-border-{start-start,start-end,end-start,end-end}-radius`), each falling back to the existing `--swc-action-button-border-radius`, so action-group's compact mode can override individual corners without hardcoding a size-specific token. Also added `inline-size: 100%` to `.swc-ActionButton` — the shadow-DOM `<button>` was shrink-wrapping to its own content regardless of how wide an ancestor flex layout stretched the custom-element host, so `vertical` and `vertical` + `compact` action groups rendered with visibly unequal button widths even though the (invisible) host boxes were already stretched correctly.

## Motivation and context

Implements SWC-2216 (Phase 5: styling) of the Action Group 2nd-gen migration epic (SWC-2212), translating the `spectrum-two` S2 source (`components/actiongroup/index.css`) into the 2nd-gen token-based CSS structure per the approved migration plan. This is the last functional gap before Phase 6 (testing) and Phase 7 (documentation) — API and accessibility semantics were completed in SWC-2215.

## Related issue(s)

- SWC-2212: Gen2 migration: Action Group (epic)
- SWC-2216: Migrate rendering and styles (Phase 5)
- SWC-1342: `sp-action-group` sets z-index on focused button, hiding focus indicator (resolved)

- Storybook — `Compact` story: horizontal and vertical variants side by side, showing joined borders with outer-corner-only rounding.
- Storybook — `Orientations`, `Sizes`, `Static colors`, `Justified` stories.

---

## Author's checklist

---

## Reviewer's checklist

### Manual review test cases

- [ ] Compact join renders with correct corner radii
  1. Open the Action Group `Compact` story in the 2nd-gen Storybook (`yarn storybook` from `2nd-gen/`)
  2. Inspect both the horizontal and vertical groups
  3. Expect: shared borders between adjacent buttons collapse to a single 1px line; only the outer corners of the strip are rounded; interior corners are square

- [ ] Compact focus ring is never clipped by a neighbor
  1. In the `Compact` story, Tab into the horizontal group, then press <kbd>→</kbd> to move through each button
  2. Repeat for the vertical group with <kbd>↓</kbd>
  3. Expect: at every position (first, middle, last), the focused button's complete rectangular focus ring is visible and unobstructed by the adjacent button's border

- [ ] Vertical and vertical + compact groups have equal-width buttons
  1. In the `Orientations` story and the vertical group in the `Compact` story, compare button widths visually and via DevTools (`getBoundingClientRect()` on each `swc-action-button`'s inner `<button>`)
  2. Expect: all buttons in a given vertical group are the same width, matching the widest label

- [ ] Quiet disables the compact join
  1. Open the `Quiet` story
  2. Compare the quiet-only group to the quiet + compact group
  3. Expect: both render identically — no border join, no corner-radius reset, regardless of `compact`

- [ ] Justified fills the available width
  1. Open the `Justified` story
  2. Expect: all three buttons expand equally to fill the constrained container width, with even gaps

- [ ] Standalone and horizontal action-button usage is unaffected
  1. Open the Action Button stories (all variants) and the Action Group `Playground`/`Overview` stories in default (horizontal) orientation
  2. Expect: no width or layout regression from the `inline-size: 100%` change on `.swc-ActionButton` — standalone buttons and horizontal groups still size to content, not to 100% of some ancestor

### Device review

- [ ] Did it pass in Desktop?

- [x] **Keyboard** (required — document steps below) — _What to test for:_ Focus order is logical; <kbd>Tab</kbd> reaches the component and all interactive descendants; <kbd>Enter</kbd>/<kbd>Space</kbd> activate where appropriate; arrow keys work for tabs, menus, sliders, etc.; no focus traps; <kbd>Escape</kbd> dismisses when applicable; focus indicator is visible.
  1. Open the Action Group `Compact` story in 2nd-gen Storybook
  2. Tab into the horizontal group — expect: focus lands on the first button with a fully visible focus ring
  3. Press <kbd>→</kbd> to move to the second, then third button — expect: at each step, the focused button's ring is complete and not covered by either neighbor
  4. Repeat steps 2–3 in the vertical group using <kbd>↓</kbd>
  5. No regression to Tab/Arrow behavior established in SWC-2215 (one Tab stop into the strip; arrow keys move among items; wraps at the ends)

  1. Open the Action Group `Compact` and `Justified` stories with VoiceOver (macOS) or NVDA (Windows)
  2. Navigate through the group — expect: no change in announcements from the non-styled group (styling-only phase); group and button roles/names announce exactly as established in SWC-2215


2nd-gen/packages/swc/components/action-group/stories/action-group.stories.ts
```diff
@@ -176,27 +176,41 @@ export const Anatomy: Story = {
 // ──────────────────────────
 
 export const Sizes: Story = {
-  render: () => html`
+  render: (args) => html`
```
rise-erpelding, line 185: The common convention in most of our other migrated components is to thread the args.

2nd-gen/packages/core/components/action-group/ActionGroup.base.ts
```diff
@@ -73,9 +73,15 @@ export abstract class ActionGroupBase extends SizedMixin(SpectrumElement, {
...
-   * implicit default for `role="group"`.
+   * `FocusgroupNavigationController` moves focus with the Up/Down arrow
+   * keys instead of Left/Right.
+   *
+   * Note: this property does NOT set `aria-orientation` on the host.
+   * `aria-orientation` is only a supported ARIA attribute on roles that
+   * expose it (`toolbar`, `listbox`, `menu`, etc.) — `role="group"` does
+   * not, regardless of the roving-tabindex keyboard model implemented in
+   * JS. Setting it fails axe's `aria-allowed-attr` rule. Matches the same
+   * decision already made for `swc-button-group`.
```
rise-erpelding, line 84: This is an a11y change after it surfaced that button group also does not set `aria-orientation`, @nikkimk would you be able to confirm that this change makes sense?

2nd-gen/packages/swc/components/action-button/action-button.css
```diff
@@ -33,6 +33,7 @@
   gap: var(--swc-action-button-gap, token("base-gap-medium"));
   align-items: center;
   justify-content: center;
+  inline-size: 100%;
```
rise-erpelding, line 36: I would love opinions about this change. It stretches the action group buttons when the action group is vertical so that all the buttons are the same width. An `align-items: stretch` on the `.swc-ActionGroup` doesn't work without it (or, doesn't have a visible effect) because it can't reach into the inner shadow DOM of the action button.

The `inline-size: 100%` isn't changing any stories on action button as far as I'm aware (but I will run VRTs to confirm), it only has an effect here in a vertical action group where the `align-items: stretch` on action group (no `align-items: stretch` is explicitly set, but that is the default behavior when we set `.swc-ActionGroup` to `display: flex`).

Or, TL;DR, in pictures, here's the vertical action group with `inline-size: 100%`:

Image

And without `inline-size: 100%`:

Image

And even more notable, without `inline-size: 100%` and with the `compact` variant set:

Image

I would think we'd prefer the aligned version with `inline-size: 100`, but also want to note that [React's](https://react-spectrum.adobe.com/ActionButtonGroup) doesn't line up nicely either:

Image

2nd-gen/packages/swc/components/action-group/action-group.css
```diff
@@ -10,4 +10,120 @@
...
+
+:host([justified]) ::slotted(*) {
+  flex: 1 1 0%;
+}
+
+/* ── Compact ──────────────────────────────────────── */
+
+:host([compact]:not([quiet])) .swc-ActionGroup {
+  flex-wrap: nowrap;
+  gap: 0;
```
aramos-adobe: XS, S, M, L, XL should all have this spacing here.

```suggestion
  gap: token("spacing-50");
```

2nd-gen/packages/swc/components/action-button/action-button.css
aramos-adobe: The `outline` in transition-property shows the black outline for a quick glance here. Removing it doesn't show this.

2nd-gen/packages/swc/components/action-group/stories/action-group.stories.ts
```diff
@@ -142,3 +161,215 @@ export const Overview: Story = {
...
+      <swc-action-button>Bold</swc-action-button>
+      <swc-action-button>Italic</swc-action-button>
+      <swc-action-button>Underline</swc-action-button>
+    </swc-action-group>
+  `,
+  tags: ['behaviors'],
+};
+Quiet.storyName = 'Quiet (compact join disabled)';
+
+export const Justified: Story = {
```
aramos-adobe, line 350: When adding extra text in one of the buttons it breaks the justified layout here. 

ActionButton group in RSP shows in truncation in the horizontal layout (and it also show in vertical layout but doesn't), but we don't have truncation support for swc-action-button even though swc-button has an opt-in adding truncate property.

I think maaayyybee truncate prop should be automatically for swc-action-group, because adding it for every button manually would be alot for a consumer.

2nd-gen/packages/swc/components/action-button/action-button.css
aramos-adobe: I don't have any objections on that. I think setting `inline-size: 100%` makes sense to fill the space in a flex box for each button. Playing with parent container width so far in the storybook container doesn't change the action button visually

aramos-adobe (changes requested):

2nd-gen/packages/swc/components/action-button/action-button.css
aramos-adobe, line 108:
```css
:host([truncate]) .swc-ActionButton-label {
  white-space: nowrap;
  overflow: hidden;
  text-overflow: ellipsis;
}
```

You also probably have to add a truncate boolean property to ActionButton class as well

rise-erpelding: Hmmm, we don't have `truncate` available on action button, do we? I think I'd want to do some more research on this or possibly defer an API change to action button for now.

rise-erpelding: I added a commit to remove it, but this prevents it from transitioning so there's no animation on it. No animation feels like maybe it's slightly better than flash of black outline 🤷‍♀️ 

The black outline is a known issue that needs more digging, since action button has it but other button components don't and it's unclear why, it's documented in SWC-2308

2nd-gen/packages/swc/components/action-group/stories/action-group.stories.ts
rise-erpelding: It does seem like addressing this might be dependent on adding truncated behavior to action button? So that the buttons are able to shrink as needed.

Although, FWIW, I see Gen 1's [justified action group](https://opensource.adobe.com/spectrum-web-components/storybook/index.html?path=/story/action-group--justified) also breaks like this when the buttons aren't all able to fit on one line, although the container has to be really small before it breaks.
image

aramos-adobe (approved):

2nd-gen/packages/swc/components/action-group/stories/action-group.stories.ts
rise-erpelding: We'll address this in SWC-2464 and SWC-2465, the first one to add `truncate` to the action button and the next one to apply it to action group when it's justified.

cdransf (approved):
This looks _awesome_! ✨

PR #6489, opened 2026-07-09
chore(core): rename @spectrum-web-components/core to @adobe/spectrum-wc-core

blunteshwar:
## Description

Renames the 2nd-gen core package from `@spectrum-web-components/core` to `@adobe/spectrum-wc-core` and aligns its version to `0.3.0` to match `@adobe/spectrum-wc` under the shared `@adobe` scope.

- Updates every import specifier across core's self-imports and swc's (components/patterns/controllers/tests/stories) consumption.
- Bumps the core package `version` to `0.3.0` and updates swc's `@adobe/spectrum-wc-core` dependency range to match, keeping the workspace resolvable.
- Updates build/test tooling: vite, vitest, tsconfig path mappings, Storybook config.
- Updates `.changeset/config.json` (`linked`/`fixed` entries keep their structure, just the renamed string) and `.github/workflows/publish.yml`. The publish workflow now reads the core package name dynamically (`core_name`) so the release summary reflects the renamed package instead of the hardcoded `@spectrum-web-components/core`.
- Updates docs: READMEs, `AGENTS.md`, CONTRIBUTOR-DOCS references.
- Removes now-dead core references from 1st-gen: unused resolution aliases in `tsconfig.json`, `web-test-runner.config.js`, and two `rollup.config.js` files, plus the `dev:core`/`prebuild` workspace-build wiring in `1st-gen/package.json`. The Storybook alias in `storybook/main.js` is renamed to `@adobe/spectrum-wc-core` instead of removed, matching the sibling `@adobe/spectrum-wc` alias. None of the removed ones had an active consumer — confirmed by grep before removal.

## Motivation and context

Core is moving under the same `@adobe` npm scope as `@adobe/spectrum-wc` ahead of the next release. The version is aligned to `0.3.0` so the renamed core and `@adobe/spectrum-wc` share the same version, matching their `linked` relationship in the changeset config.

## Related issue(s)

- SWC-2356

N/A — no UI change.

## Author's checklist

- [x] I have read the [CONTRIBUTING](https://github.com/adobe/spectrum-web-components/blob/main/CONTRIBUTING.md) and [PULL_REQUESTS](https://github.com/adobe/spectrum-web-components/blob/main/PULL_REQUESTS.md) documents.

## Reviewer's checklist

### Manual review test cases

- [ ] _Confirm the renamed workspace resolves and the old name doesn't_
  1. Run `yarn install` at repo root.
  2. Run `yarn workspace @adobe/spectrum-wc-core exec pwd` — should resolve.
  3. Run `yarn workspace @spectrum-web-components/core exec pwd` — should fail with "workspace not found".

- [ ] _Confirm 1st-gen build/start/test scripts are unaffected by the cleanup_
  1. Run `yarn build`, `yarn start`, and `yarn test:watch:focus` inside `1st-gen/`.
  2. Expect no errors referencing `@spectrum-web-components/core` or `dev:core`.

### Device review

- [ ] Did it pass in Desktop?

No UI, component, or behavioral change in this PR — package rename and build-tooling cleanup only.

- [x] **Keyboard**: N/A — no focusable parts touched; no component behavior changed.
- [x] **Screen reader**: N/A — no DOM, ARIA, or component semantics changed.

## Open follow-ups (not blocking this PR)

- The old npm package `@spectrum-web-components/core@0.3.0` is already published and stays live/stale on the registry — no deprecation or forwarding done here.
- `1st-gen/scripts/update-global-changelog.js` still hardcodes the old package name for git-tag lookups when generating core's changelog. Left untouched since it's inside `1st-gen/`, but will need updating before core's next release for changelog generation to find the right tags.
- Whether this rename ships with a changeset (likely `major` for core, given the `linked` tie to swc) is still open.


1st-gen/package.json
```diff
@@ -49,15 +48,14 @@
...
-    "prebuild": "wireit",
     "preeleventy": "yarn docs:analyze",
     "prepack": "pinst --disable",
     "prestorybook": "wireit",
     "prestorybook:build": "cem analyze --outdir storybook/",
     "pretest:bench": "yarn build:tests && test -f test/benchmark/cli.js ||:",
     "pretest:visual": "yarn build && yarn build",
     "process-icons": "wireit",
-    "start": "run-p dev:core storybook",
+    "start": "storybook",
```
blunteshwar, line 58: Removed `dev:core` from `start` (was `run-p dev:core storybook`). `dev:core` ran `yarn workspace @spectrum-web-components/core dev`, which watch-built the old core package for the tsconfig/Storybook aliases removed elsewhere in this PR. Neither alias had an actual consumer in 1st-gen source, so `dev:core` and the `prebuild` wireit task (further down in this file) are removed rather than renamed to the new package name.

1st-gen/projects/documentation/rollup.config.js
```diff
@@ -246,10 +246,6 @@ export default async () => {
           find: '@swc-packages-internal',
           replacement: path.resolve('../../packages'),
```
blunteshwar, line 247: Removed the `@spectrum-web-components/core/(.*)` alias entry that resolved to `2nd-gen/packages/core/dist`. Nothing in the docs-site build actually imports that specifier (confirmed via grep across 1st-gen before removing), so it's dropped instead of renamed to `@adobe/spectrum-wc-core`. The sibling `@swc-packages-internal` alias is untouched.

1st-gen/storybook/main.js
```diff
@@ -53,9 +53,6 @@ export default {
           '../2nd-gen/packages',
         ],
         alias: {
-          '@spectrum-web-components/core': resolve(
-            '../2nd-gen/packages/core/dist'
-          ),
           '@adobe/spectrum-wc': resolve('../2nd-gen/packages/swc/dist'),
```
blunteshwar, line 57: Removed the `@spectrum-web-components/core` alias, kept `@adobe/spectrum-wc`. No gen1 story imports the core specifier — this was dead resolution config, not a live dependency.

1st-gen/test/visual/rollup.config.js
```diff
@@ -12,7 +12,6 @@
 import { rollupPluginHTML as html } from '@web/rollup-plugin-html';
 import { nodeResolve } from '@rollup/plugin-node-resolve';
 import { copy } from '@web/rollup-plugin-copy';
-import alias from '@rollup/plugin-alias';
 import replace from '@rollup/plugin-replace';
```
blunteshwar, line 15: Removed the `@rollup/plugin-alias` import along with the alias plugin block below (its only usage in this file). It existed solely to resolve `@spectrum-web-components/core/*` for VRT builds, and nothing in this bundle imports that specifier.

1st-gen/tools/base/src/version.ts
```diff
@@ -22,4 +22,4 @@ export const version = '1.12.2';
 /**
  * The version of the core base package.
  */
-export const coreVersion = '0.3.0';
+export const coreVersion = '0.0.1';
```
blunteshwar: Auto-regenerated by the `generate-versions.js` pre-commit hook to reflect core's new version (`0.3.0` → `0.0.1`) after the rename in `2nd-gen/packages/core/package.json`. Not a hand edit.

1st-gen/tsconfig.json
```diff
@@ -5,11 +5,6 @@
     "emitDeclarationOnly": true,
     "importHelpers": true,
     "inlineSources": true,
-
-    // Cross-generation import mapping
-    "paths": {
-      "@spectrum-web-components/core/*": ["../2nd-gen/packages/core/*/dist"]
-    },
     "rootDir": ".",
```
blunteshwar, line 8: Removed the `@spectrum-web-components/core/*` path mapping into `2nd-gen/packages/core/*/dist`. No `.ts` file in 1st-gen imports that specifier, so it's dropped rather than repointed at the renamed package.

1st-gen/web-test-runner.config.js
```diff
@@ -40,7 +39,6 @@ const __dirname = path.dirname(fileURLToPath(import.meta.url));
 
 const commonjs = fromRollup(rollupCommonjs);
 const json = fromRollup(rollupJson);
-const alias = fromRollup(rollupAlias);
 const replace = fromRollup(rollupReplace);
```
blunteshwar, line 42: Removed the `rollupAlias`/`fromRollup` setup and its single usage below — same dead `@spectrum-web-components/core` → dist alias pattern as the other 1st-gen build configs, with no actual test-file consumer.

blunteshwar (commented):
Per-file notes on every 1st-gen change in this PR — all of it is either dead-alias removal (confirmed unused via grep before deleting) or an auto-regenerated version constant, not a live dependency change.

blunteshwar (commented):
.

1st-gen/storybook/main.js
```diff
@@ -53,9 +53,7 @@ export default {
           '../2nd-gen/packages',
         ],
         alias: {
-          '@spectrum-web-components/core': resolve(
-            '../2nd-gen/packages/core/dist'
-          ),
+          '@adobe/spectrum-wc-core': resolve('../2nd-gen/packages/core/dist'),
```
blunteshwar, line 56: Update: this alias is renamed to `@adobe/spectrum-wc-core` (not removed). It matches the sibling `@adobe/spectrum-wc` alias directly below it, so gen1 Storybook keeps a working resolution path to core's dist if anything ever imports the specifier.

pfulton:
FYI: I added the `run_vrt` label here to be sure we're covering all of our bases.

blunteshwar:
> FYI: I added the `run_vrt` label here to be sure we're covering all of our bases.

Thanks for doing this. I was planning to add it after receiving at least one approval.

Rajdeepc (changes requested):
Mostly fine but need some clarifications on the below.
1. Please confirm that there is no changes in publish workflow and the new `@adobe/spectrum-wc-core` package is published with `next`.
2. `core` and @adobe/spectrum-wc were both `0.3.0` (linked = same version, not just same bump type). `core` resets to `0.0.1` but swc stays `0.3.0`. Do you need some suggestions?
3. I am not sure why you dropped `dev:core / prebuild` from 1st-gen start and `test:watch:focus`. This is a not a dead code cleanup. Please get buying in before dropping.

blunteshwar:
> Mostly fine but need some clarifications on the below.
> 
> 1. Please confirm that there is no changes in publish workflow and the new `@adobe/spectrum-wc-core` package is published with `next`.
> 2. `core` and @adobe/spectrum-wc were both `0.3.0` (linked = same version, not just same bump type). `core` resets to `0.0.1` but swc stays `0.3.0`. Do you need some suggestions?
> 3. I am not sure why you dropped `dev:core / prebuild` from 1st-gen start and `test:watch:focus`. This is a not a dead code cleanup. Please get buying in before dropping.

▎ 1. Confirm no changes in publish workflow and that @adobe/spectrum-wc-core publishes with next.

  Confirmed — no change to the publish mechanism is required.

  - Publishing runs through yarn changeset publish --no-git-tag --tag $TAG (publish.yml L176–190). Changesets publishes every non-private workspace package by its own name, so the rename is picked up automatically. Core is "private": undefined with "publishConfig": { "access":  "public" }, so it publishes as @adobe/spectrum-wc-core with no workflow edit.
  - Tag: for push runs TAG defaults to next (L107), passed straight to changeset publish. So the renamed package lands on next. ✓
  - The only publish.yml diff in this PR is cosmetic: the "Versions released" summary label now reads the core package name dynamically (CORE_NAME) instead of the hardcoded @spectrum-web-components/core string, so the summary reflects the actual published name. No effect on what/how/where anything publishes.
  - Side note (non-blocking): the grep -v "^@spectrum-web-components/core$" guard in check-changesets (L53) that excluded old-core from triggering 1st-gen React-wrapper builds is now redundant — the renamed package is outside the @spectrum-web-components/* namespace, so it's excluded by construction. Harmless; can be cleaned up in a follow-up.

  ▎ 2. core and swc were both 0.3.0 (linked). core reset to 0.0.1 while swc stayed 0.3.0.

  Agreed, and fixed. Core is now bumped straight to 0.3.0 to match @adobe/spectrum-wc, preserving the linked (same-version) relationship rather than diverging. Also updated swc's @adobe/spectrum-wc-core dependency to 0.3.0 so the workspace resolves. Pushed in d62ca9fd76.

  ▎ 3. Why drop dev:core / prebuild from 1st-gen start and test:watch:focus?

  This is a deliberate decoupling, and I'd argue it belongs in this PR. gen1 and core are unrelated — nothing in 1st-gen source imports the core specifier (the removed aliases were dead resolution config, confirmed by grep). Wiring a core watch-build into gen1's start/prebuild/test:watch:focus meant gen1 rebuilt an unrelated package on every run for no consumer. Removing it decouples the two generations rather than renaming a build step that shouldn't exist in gen1 at all. Happy to split it into a separate PR if you'd prefer the rename to stay mechanical, but the intent is to stop gen1 from building core.

.github/workflows/publish.yml
```diff
@@ -282,7 +285,7 @@ jobs:
             echo "" >> $GITHUB_STEP_SUMMARY
             echo "| Generation | Package | Version |" >> $GITHUB_STEP_SUMMARY
             echo "|------------|---------|---------|" >> $GITHUB_STEP_SUMMARY
-            [ -n "$CORE_VERSION" ]   && echo "| Core    | \`@spectrum-web-components/core\`   | \`${CORE_VERSION}\`   |" >> $GITHUB_STEP_SUMMARY
+            [ -n "$CORE_VERSION" ]   && echo "| Core    | \`${CORE_NAME}\`   | \`${CORE_VERSION}\`   |" >> $GITHUB_STEP_SUMMARY
```
caseyisonit: nit: you can just manually update this so we dont have to track more variables in the output

blunteshwar: Makes sense, Done!

.changeset/config.json
```diff
@@ -13,7 +13,7 @@
   "fixed": [
     [
       "@spectrum-web-components/*",
-      "!@spectrum-web-components/core",
+      "!@adobe/spectrum-wc-core",
```
rubencarvalho: nit: this is not needed now. `@adobe` doesn't need to be negated now because it does not match the `@spectrum-web...` above 😄

.github/workflows/publish.yml
rubencarvalho, line 1: I believe you missed https://github.com/adobe/spectrum-web-components/blob/f72211acf7ab2cb8d367477c82d60f7f0fd5d503/.github/workflows/publish.yml#L53

rubencarvalho (changes requested):
Left 1 blocking, 1 nit! Other than that, looks good 😄

rubencarvalho (approved):

Rajdeepc (approved):

PR #6490, opened 2026-07-09
chore(SlotAttributePropagationController): add tests + docs

cdransf:
## Description

Adds a `getValue`/attribute-removal capability to `SlotAttributePropagationController`, plus unit tests and a Storybook controller docs page. The controller is already used by `ButtonGroupBase` (default slot) and `IllustratedMessageBase` (named `actions` slot), but had no dedicated test coverage or documentation of its own, and could not represent an attribute that is only sometimes present on the host.

Changes:

- **API change / fix** (`core/controllers/slot-attribute-propagation-controller/src/slot-attribute-propagation-controller.ts`): `getValue` may now return `string | null`. Returning `null` removes the propagated attribute from assigned elements instead of setting it to an empty string. Also fixes a bug where `_previousValue` (the `hostUpdated()` no-op guard) was set unconditionally before the slot was resolved; if the slot wasn't resolvable yet, that value was silently marked as "applied" and a later `hostUpdated()` call with the same value would never re-propagate once the slot did resolve. `_previousValue` is now only set inside `_propagateToSlot()`, after a successful slot resolution.
- **Tests** (`core/controllers/slot-attribute-propagation-controller/test/slot-attribute-propagation-controller.test.ts`): Storybook play-function tests covering first-render propagation, `hostUpdated()` re-propagation on value change, the `previousValue` no-op guard (including the slot-not-yet-resolved case), `propagate()` via `slotchange`, the `hostDisconnected()` reset on reconnect, `slotName` targeting a named slot (leaving the default slot untouched), the `selector` option filtering assigned elements, and the new nullable `getValue` / attribute-removal behavior.
- **Storybook docs** (`core/controllers/slot-attribute-propagation-controller/slot-attribute-propagation-controller.mdx` + `stories/`): a full controller docs page (Usage, Options, Behaviors, Accessibility, hand-authored API section) plus demo hosts mirroring the two real production consumers, the `selector` option, and the nullable/optional-attribute case.
- **Public export**: `SlotAttributePropagationController` is now exported from the `core` package's controllers barrel and package.json subpath export (`@spectrum-web-components/core/controllers/slot-attribute-propagation-controller.js`), matching the flat-file controller convention already used by `hover-controller.js`, `color-controller.js`, and `focus-group-navigation-controller.js`.
- **Build fix** (`core/vite.config.js`): `getEntries()` now also scans `controllers/*/index.ts` (via `scanSubdirIndexes`, the same helper already used for `components/`), not just flat files directly in `controllers/`. Without this, Rollup's `preserveModules` output could inline a controller's re-export barrel into its parent chunk and drop the standalone file, silently breaking that controller's package.json subpath export. This was caught while adding the new export above: it turned out to already be broken (pre-existing, unnoticed) for `live-selection-controller` too.
- **Rename**: the controller's folder and files were renamed from `slot-attribute-propagation` to `slot-attribute-propagation-controller` for consistency with sibling controller directories; all internal imports (`ButtonGroup.base.ts`, `IllustratedMessage.base.ts`) were updated accordingly.

## Motivation and context

SWC-2311: the controller shipped as part of the illustrated-message actions slot work with no dedicated tests or docs, as a known follow-up, and could only propagate string values. A second production consumer (`ButtonGroupBase`) has since adopted it, and a future `DropzoneBase` migration is expected to need to propagate an attribute that isn't always present on the host, which the original `getValue: () => string` signature couldn't express.

## Related issue(s)

- fixes SWC-2311

N/A — no visual/UI changes. New Storybook docs page: **Controllers → Slot attribute propagation controller**.

## Author's checklist

- [ ] I have included a well-written changeset if my change needs to be published. (Reviewer is adding this as a `patch` — fix + new public export.)

## Reviewer's checklist

### Manual review test cases

- [ ] _Controller docs page renders correctly_
  1. Go to Storybook → **Controllers → Slot attribute propagation controller**
  2. Confirm the Overview, Options (Default slot / Named slot / Selector filtering / Optional attribute), Behaviors, Accessibility, and API sections all render with no console errors
  3. Expect: each `<Canvas>` shows the described propagation behavior (e.g. "Named slot" shows the `actions`-slot button receiving `size`, while the default-slot paragraph does not; the optional-attribute demo shows the attribute being removed, not set to `""`, when `getValue` returns `null`)

- [ ] _New controller tests pass_
  1. Run `yarn workspace @spectrum-web-components/2nd-gen-swc vitest --run --project storybook -- slot-attribute-propagation-controller`
  2. Expect: all tests in `slot-attribute-propagation-controller.test.ts` pass

- [ ] _No regression in existing consumers_
  1. Run the same command scoped to `button-group` and `illustrated-message`
  2. Expect: all existing size-propagation tests in both components still pass unchanged

- [ ] _New public export resolves_
  1. From `2nd-gen/packages/core`, run `yarn build`
  2. Confirm `dist/controllers/slot-attribute-propagation-controller/index.js` (and the `hover-controller` / `focus-group-navigation-controller` / `live-selection-controller` equivalents) exist and export their controller class

### Device review

- [ ] Did it pass in Desktop?

  1. Go to Storybook → **Controllers → Slot attribute propagation controller** → Options → "Named slot"
  2. Tab through the rendered content: the default-slot paragraph (non-interactive) and the `actions`-slot button
  3. Expect: normal Tab order through the native `<button>` element; the propagated `size` attribute is a plain DOM attribute and does not affect focusability, tab order, or introduce any new tab stops

  1. Go to Storybook → **Controllers → Slot attribute propagation controller** → Options → "Selector filtering"
  2. With VoiceOver (or NVDA) running, navigate to both buttons
  3. Expect: both are announced as "button" with their visible text label; the propagated `variant` attribute carries no ARIA semantics, so no additional or altered announcements occur. The controller itself never sets `role`, `aria-*`, or any accessibility-relevant attribute — it only calls `setAttribute`/`removeAttribute` with whatever attribute name the consuming component configures.


2nd-gen/packages/core/controllers/slot-attribute-propagation/slot-attribute-propagation.mdx
```diff
@@ -0,0 +1,117 @@
...
+
+### Named slot
+
+Set `slotName` to target a specific named slot. Elements assigned to other slots on the same host, including the default slot, are left untouched.
+
+<Canvas of={Stories.NamedSlot} />
+
+### Selector filtering
+
+Set `selector` to a CSS selector to propagate the attribute only to assigned elements that match it. Elements in the same slot that don't match are left untouched. There's no existing production consumer of this option yet; it's demonstrated here for the option's own sake.
```
rubencarvalho: nit: I think we should not keep these temporal references in the docs, I always find tricky to keep them up to date with actual implementations 😛 
```suggestion
Set `selector` to a CSS selector to propagate the attribute only to assigned elements that match it. Elements in the same slot that don't match are left untouched.
```

2nd-gen/packages/core/controllers/slot-attribute-propagation/slot-attribute-propagation.mdx
```diff
@@ -0,0 +1,117 @@
+import { Canvas, Meta } from '@storybook/addon-docs/blocks';
+import { DocsFooter, DocsHeader } from '../../../swc/.storybook/blocks';
+
+import * as Stories from './stories/slot-attribute-propagation.stories';
+
+<Meta of={Stories} />
+
+<DocsHeader />
```
rubencarvalho, line 8: I know this is not necessarily this PR's scope, but this DocsHeader injects the components' GetStarted which renders a wrong path for the core controllers:

```ts
import { SlotAttributePropagationController } from '@adobe/spectrum-wc/components/core/controllers/slot-attribute-propagation-controller.js';
```

(notice it is not importing from the core package)

2nd-gen/packages/core/controllers/slot-attribute-propagation-controller/slot-attribute-propagation-controller.mdx
```diff
@@ -0,0 +1,117 @@
...
+
+### Best practices
+
+- Use this controller for presentation-oriented attributes that slotted children are expected to read declaratively, such as `size` or a visual `variant`.
+- Prefer explicit ARIA wiring (for example `aria-describedby` set directly on the relevant elements) over propagating ARIA attributes through this controller, since assigned elements may have their own ARIA semantics that a blanket propagation could override unexpectedly.
+- Always wire the targeted slot's `slotchange` event to `propagate()` so elements added after first render aren't missed.
+
+<Canvas of={Stories.Accessibility} />
+
+## API
```
rubencarvalho, line 108: I was reviewing the existing API contract and then i checked the controller implementation and noticed one thing, let me know what you think about both `getValue` supporting `string | null`, whereas `null` removes the attribute? Since this is a general "agnostic controller", we may need to propagate optional attributes.

2nd-gen/packages/core/controllers/slot-attribute-propagation/slot-attribute-propagation.mdx
```diff
@@ -0,0 +1,117 @@
...
+
+<Canvas of={Stories.Accessibility} />
+
+## API
+
+### Methods
+
+| Member        | Description                                                                                                                                                           |
+| ------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `propagate()` | Propagates the current value (from `getValue()`) to all matching assigned elements immediately, bypassing the no-op guard. Call from the host's `slotchange` handler. |
```
rubencarvalho: also, on the `propagate()` implementation side, a nit (my intuition tells me it is a prob a micro-optimisation), but could propagate() update `_previousValue` after applying the value? Otherwise, a pending `hostUpdated()` may repeat the same `setAttribute()` calls, potentially (although I think a bit unlikely) triggering duplicate attribute reactions and child updates.

rubencarvalho (commented):
Looking really good! Nothing really blocking, but left a couple of thoughts 😄

2nd-gen/packages/core/controllers/slot-attribute-propagation/slot-attribute-propagation.mdx
cdransf: Gotcha! Updated https://github.com/adobe/spectrum-web-components/pull/6490/commits/acf140c609b1d47d4d5115e53d516d8d7da076c1 ✨

2nd-gen/packages/core/controllers/slot-attribute-propagation-controller/slot-attribute-propagation-controller.mdx
cdransf: I took a look at this in https://github.com/adobe/spectrum-web-components/pull/6490/commits/b109cca378a3c6818c46894dc5a047c1448baac4. Let me know what you think ✨

2nd-gen/packages/core/controllers/slot-attribute-propagation/slot-attribute-propagation.mdx
cdransf: Updated here as well https://github.com/adobe/spectrum-web-components/pull/6490/commits/b109cca378a3c6818c46894dc5a047c1448baac4 ✨

rubencarvalho (approved):
🚢 it!! nice :)

2nd-gen/packages/swc/.storybook/blocks/GettingStarted.tsx
```diff
@@ -30,7 +37,7 @@ Controllers are not published as packages. Instead, they are imported directly f
 Import the controller directly from the core package:
 
 \`\`\`typescript
-import { ${baseClassName} } from '@adobe/spectrum-wc/components/core/controllers/${packageName}.js';
+import { ${baseClassName} } from '@spectrum-web-components/core/controllers/${packageName}.js';
```
caseyisonit:
```suggestion
import { ${baseClassName} } from '@adobe/spectrum-wc-core/controllers/${packageName}.js';
```

We just transitioned core package to this

2nd-gen/packages/swc/.storybook/blocks/DocsHeader.tsx
```diff
@@ -46,7 +51,7 @@ export const DocsHeader = () => {
       <Subtitle />
       <Description />
       <OverviewStory />
-      <GettingStarted tags={tags} />
+      <GettingStarted tags={tags} packageName={packageName} />
```
caseyisonit: I dont think this is needed, the file name includes controller and this is effectively removing it. this is used to generate the import path so controller would need to be included since thats that file name

cdransf: Gotcha! I went down the path of renaming the file to match the exported class name to match other controllers in https://github.com/adobe/spectrum-web-components/pull/6490/commits/e265de78d7169fb21ee3c79e3480f7eb6fdeb507. ✨

Rajdeepc (commented):
@cdransf I will be adding a changeset in this PR as I can see this is fix we are shipping and a new public export also a bug fix. This should be a minimum patch. Also you PR description is not matching the code.
No other concerns.

cdransf:
> @cdransf I will be adding a changeset in this PR as I can see this is fix we are shipping and a new public export also a bug fix. This should be a minimum patch. Also you PR description is not matching the code. No other concerns.

Updated the description and added a changeset. ✨

2nd-gen/packages/swc/.storybook/blocks/GettingStarted.tsx
```diff
@@ -30,7 +37,7 @@ Controllers are not published as packages. Instead, they are imported directly f
 Import the controller directly from the core package:
 
 \`\`\`typescript
-import { ${baseClassName} } from '@adobe/spectrum-wc/components/core/controllers/${packageName}.js';
+import { ${baseClassName} } from '@adobe/spectrum-wc-core/controllers/${packageName}.js';
```
caseyisonit, line 30: this path needs to be more dynamic otherwise all components will have their paths updated to controllers.

caseyisonit: ignore me i needed to expand the code above to see its int he controller if statement

2nd-gen/packages/swc/.storybook/blocks/DocsHeader.tsx
```diff
@@ -34,8 +34,12 @@ import { StatusBadge } from './StatusBadge';
  *   <DocsHeader />
  *   ...component-specific sections...
  *   <DocsFooter />
+ *
+ * @param packageName - Optional override forwarded to `GettingStarted` for
+ * controllers whose `core` package export name doesn't match the title-derived
+ * kebab-case.
  */
-export const DocsHeader = () => {
+export const DocsHeader = ({ packageName }: { packageName?: string } = {}) => {
```
caseyisonit: are these overrides still needed?

2nd-gen/packages/swc/.storybook/blocks/GettingStarted.tsx
```diff
@@ -7,20 +7,27 @@ import { formatTitle } from '../helpers/index.js';
...
- * @param packageName - Optional override for the package name (defaults to derived kebab-case from title)
- * @param componentName - Optional override for the component class name (defaults to derived PascalCase from title)
- * @param tagName - Optional override for the custom element tag name (defaults to swc-{packageName})
+ * @param packageName - Optional override for the controller/package export name (defaults to derived kebab-case from title; needed when a controller's `core` package export key doesn't match the title-derived kebab-case)
  */
-export const GettingStarted = ({ of, tags }: { of?: any; tags?: string[] }) => {
+export const GettingStarted = ({
+  of,
+  tags,
+  packageName: packageNameOverride,
```
caseyisonit: are these overrides still needed?

2nd-gen/packages/core/components/button-group/ButtonGroup.base.ts
caseyisonit, line 111: should we not add this in to the sizeMixin?

caseyisonit (approved):

PR #6496, opened 2026-07-10
fix(combobox, breadcrumbs): propagate lang/dir for a single item's language without breaking layout

majornista:
## Description

Adds "language of parts" support to `sp-combobox` and `sp-breadcrumbs`: consumers can set `lang`/`dir` on an individual slotted item (e.g. a language name rendered in its own script) and have it render correctly, without breaking the surrounding component's layout.

- **Combobox**: a slotted `<sp-menu-item lang="he" dir="rtl">` (or a `ComboboxOption` with `lang`/`dir`) now propagates those attributes to its rendered counterpart in the popover. The text input itself also adopts the `lang` of the currently-selected option, so screen readers pronounce the displayed value correctly.
- **Breadcrumbs**: the same propagation applies to items collapsed into the "More items" overflow menu.
- **BreadcrumbItem**: `lang`/`dir` now apply only to `#item-link` (the text), not the host — so one item's language doesn't flip its own layout or its separator's chevron relative to its siblings. The chevron now tracks the *ambient* direction (nearest ancestor's `dir`, or the document default) instead of the item's own `dir` attribute, and stays correct if an ancestor's `dir` changes after the item has already mounted, or for the overflow-menu wrapper item (which is rendered inside `Breadcrumbs`' own shadow root rather than as a light-DOM child).

Two non-obvious base-class behaviors caused regressions mid-implementation and are now specifically guarded against (with tests that fail without the fix):
- `SpectrumElement` overrides the native `dir` *getter* to return computed CSS direction rather than the attribute, so authored `dir` values must be read via `getAttribute('dir')`.
- CSS `:dir()` resolves directionality via the DOM's `dir` *attribute* chain, entirely independent of the `direction` *property* — so `direction: inherit` alone doesn't stop a mismatched attribute from being picked up by descendant `:dir()` selectors.

## Motivation and context

Combobox and Breadcrumbs synthesize new elements from extracted data (rather than reusing the original slotted elements) when rendering their popover/overflow menu, so any `lang`/`dir` set on the original item was silently dropped — breaking pronunciation and bidi text shaping for mixed-language lists (e.g. a language picker).

## Related issue(s)

- fixes SWC-2359

N/A — no visual/layout changes to default (single-language) usage. New Storybook stories (`languageOfParts`, `LanguageOfParts`) demonstrate the fix; a screenshot/VRT baseline update may be worth attaching before merge.

## Author's checklist

- [x] I have included a well-written changeset if my change needs to be published. *(not yet added — flagging before merge)*
- [ ] I have included updated documentation if my change required it. *(no MDX/README changes made; consider a note in each component's docs)*

---

## Reviewer's checklist

- [x] Includes thoughtfully written changeset if changes suggested include `patch`, `minor`, or `major` features
- [x] Validated on all supported browsers
- [x] All VRTs are approved before the author can update Golden Hash

### Manual review test cases

- [x] _Combobox popover items and input keep an item's own language_
  1. Go to Storybook → Combobox → `languageOfParts`
  2. Open the combobox and inspect each option in devtools
  3. Expect each `<sp-menu-item>` in the popover to carry the same `lang`/`dir` as its source; select the Hebrew or Arabic option and expect the input's own `lang` attribute to update to match

- [x] _Breadcrumbs overflow menu keeps an item's own language_
  1. Go to Storybook → Breadcrumbs → `LanguageOfParts` (forces the overflow menu via `max-visible-items`)
  2. Open the "More items" menu and inspect the items in devtools
  3. Expect each `<sp-menu-item>` to carry the same `lang`/`dir` as the source `<sp-breadcrumb-item>`

- [x] _A single item's language doesn't break the breadcrumb trail's layout_
  1. In the `LanguageOfParts` story, locate the Hebrew/Arabic breadcrumb items amid the (LTR) trail
  2. Expect their separators (chevrons) to point the same direction as their LTR siblings, not mirrored
  3. Toggle `dir="rtl"` on `<sp-breadcrumbs>` via devtools after the story has rendered
  4. Expect all separators (including the "More items" wrapper's, if overflowed) to mirror correctly

### Device review

- [x] Did it pass in Desktop?

  1. Go to Storybook → Combobox → `languageOfParts`
  2. Tab to the combobox, press <kbd>ArrowDown</kbd> to open, arrow through options, press <kbd>Enter</kbd> to select a non-Latin-script option (e.g. עברית)
  3. Expect focus to return to the input with the selected value displayed; no change in tab order or focus trapping from the `lang`/`dir` changes

  1. With a screen reader active, open Storybook → Combobox → `languageOfParts` and arrow through the popover options
  2. Expect each option (including עברית, العربية) to be announced using that language's pronunciation rules, not the page's default language
  3. Select a non-Latin option and expect the combobox's announced value to use the same correct pronunciation
  4. Repeat for Storybook → Breadcrumbs → `LanguageOfParts`, opening the "More items" overflow menu


1st-gen/packages/breadcrumbs/src/BreadcrumbItem.ts
```diff
@@ -49,12 +73,32 @@ export class BreadcrumbItem extends LikeAnchor(Focusable) {
...
     if (!this.hasAttribute('role')) {
       this.setAttribute('role', 'listitem');
     }
+
+    for (const ancestor of ancestorElements(this)) {
+      this.ancestorDirObserver.observe(ancestor, {
+        attributes: true,
+        attributeFilter: ['dir'],
+      });
+    }
```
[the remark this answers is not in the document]
majornista: Fixed and added test: de8e8a4505adb2028dc59a3adb9dafe599f015bf

AliMahmoudDev:
This looks great! Using CSS custom properties for the spacing and color tokens would make it easier to theme and maintain consistency across the design system.

AliMahmoudDev:
Good fix for language/direction propagation! This is the kind of internationalization issue that's easy to miss. Propagating lang/dir correctly ensures screen readers use the right pronunciation for each item.

majornista:
> This looks great! Using CSS custom properties for the spacing and color tokens would make it easier to theme and maintain consistency across the design system.

This PR doesn't use spacing or color tokens. It only concerns propagating `lang` and `dir` attributes to the MenuItem elements in the shadowDOM.

caseyisonit:
@majornista can you revert the golden_hash token so that we can review the VRT failures as part of our review? that is usually the last commit after getting two approvals.

1st-gen/packages/breadcrumbs/src/BreadcrumbItem.ts
```diff
@@ -30,6 +31,30 @@ export interface BreadcrumbSelectDetail {
   value: string;
 }
 
+// Walks up through `parentElement`, crossing shadow-root boundaries via
+// `getRootNode().host` — needed because `Breadcrumbs.renderMenu()` places
+// an "is-menu" `BreadcrumbItem` inside its own shadow root rather than as
+// a light-DOM child, so a plain `parentElement` walk would stop there
+// without ever reaching `<sp-breadcrumbs>` itself.
+function* ancestorElements(start: Element): Generator<Element> {
```
Rajdeepc, line 40: This walks every ancestor (`parentElement`/shadow-host chain) up to the document root and gives each `BreadcrumbItem` its own `MutationObserver` that calls `.observe()` on every one of those ancestors, unconditionally, even for items with no `lang`/`dir` override.

We already have the right pattern for this exact problem: [`LanguageResolution.ts`](../../../tools/reactive-controllers/src/LanguageResolution.ts) (`addLangListener`/`sharedObserver`) uses a single shared singleton `MutationObserver` on `document.documentElement` that fans out to a listener `Set`, instead of one observer per component instance.

As written, an 8-item breadcrumb trail nested 15 levels deep in a real app shell creates 8 separate observers each re-watching the same shared ancestors — O(items × ancestor depth) registrations instead of O(depth). It also sets a precedent problem: the next component that synthesizes elements from source data (Picker, Tabs, ActionBar) now has two incompatible idioms to choose from for "react to an ambient attribute change."

Please refactor this to extend/mirror `LanguageResolutionController`'s shared-observer pattern rather than introducing a second one?

1st-gen/packages/breadcrumbs/src/BreadcrumbItem.ts
```diff
@@ -49,12 +74,40 @@ export class BreadcrumbItem extends LikeAnchor(Focusable) {
...
   }
 
+  // `renderLink()` and `renderSeparator()` bake this host's own `lang`/`dir`
+  // and the ambient `direction` into explicit attributes at render time.
+  // Neither `lang` nor `dir` is a reactive Lit property here, so — unlike
+  // plain CSS inheritance — none of that updates on its own if this host's
+  // own `lang`/`dir` change, or an ancestor's `dir` changes, after mount.
+  // Watching this host (for `lang`/`dir`) and every ancestor (for `dir`)
+  // keeps both cases live.
+  private readonly ancestorDirObserver = new MutationObserver(() =>
```
Rajdeepc: Watching the host's *own* `lang`/`dir` attributes reuses this same MutationObserver. For a custom element's own attribute changes, we already have a lighter native idiom in this codebase — see `Tooltip.ts`'s `static get observedAttributes()` + `attributeChangedCallback`. No extra observer object is needed just to react to your own attributes; that part of this mechanism could be simplified/removed independent of the ancestor-walking discussion above.

1st-gen/packages/combobox/src/Combobox.ts
```diff
@@ -330,6 +346,20 @@ export class Combobox extends Textfield {
     this.inputElement.focus();
   }
 
+  // The input displays the committed option's own text, so it should be
+  // pronounced in that option's language rather than the combobox's
+  // ambient `lang`. `undefined` while typing/filtering (before `itemValue`
+  // resolves to a match), so the input falls back to the ambient `lang`.
+  private get selectedItemLang(): string | undefined {
```
Rajdeepc, line 353: This syncs the input's `lang` to the committed option so screen readers pronounce it correctly, but `dir` isn't synced alongside it. Selecting the Hebrew/Arabic option puts genuinely RTL text in a directionally-LTR input — pronunciation is now correct, but caret placement/text alignment for that value may still be off.

Given this PR's whole purpose is bidi/language-of-parts correctness, was leaving `dir` unsynced here a deliberate scope decision, or should it track alongside `lang`?

Rajdeepc (dismissed):
Reviewed from an architecture/ownership lens (public API, consistency with existing patterns, a11y, release risk). Left 3 inline comments — two are blocking asks around reusing an existing pattern instead of introducing a new one, one is a question about scope. Core diagnosis of the underlying bug (`SpectrumElement`'s `dir` getter returning computed style, `:dir()` resolving via the attribute chain) is correct and well tested; not re-litigating that part.

1st-gen/packages/combobox/src/Combobox.ts
majornista: My thinking is that changing the `dir` for the ComboBox `input` to be different than what it was originally could be confusing to a user, but we can experiment with syncing the `dir` as well.

majornista: See 40c70c8fe4bcb965b6350ef9f0e8d9b3acd77658, where I've added `dir="auto"` to the `combobox#input` and corrected the input padding.

1st-gen/packages/combobox/src/combobox.css
```diff
@@ -61,6 +61,84 @@ sp-popover:not(sp-overlay sp-popover) {
   border-color: var(--mod-combobox-border-color-default, var(--spectrum-combobox-border-color-focus-hover));
 }
 
+/* `#input` has `dir="auto"` (see Combobox.ts), so its own computed `direction`
```
Rajdeepc: This block re-declares six state combinations' worth of `calc()` expressions (default, `[invalid]`/`[pending]`, `[quiet]`, `[quiet][invalid]`/`[quiet][pending]`) that already exist in `spectrum-combobox.css`, converted from logical (`padding-inline-start`/`-end`) to physical (`padding-left`/`padding-right`) properties, then wins the cascade over the original via `:dir()` selectors chosen specifically to "raise specificity just enough... regardless of source order."

Two compounding risks with this as the long-term fix:

1. **Two sources of truth for the same values.** If `spectrum-combobox.css` ever changes (a spacing token rename, a new state, an added selector) nothing keeps this copy in sync — the failure mode is a subtle RTL-only, state-combination-specific layout bug, possibly not even caught by VRT depending on story coverage.
2. **Specificity-race as a strategy is inherently fragile.** "Just enough specificity to win" can flip the moment one more attribute selector is added to the base rule later, silently.

Before accepting this duplication, could you spike whether `unicode-bidi: plaintext` on `#input` (keeping `direction` anchored to the ambient host direction the whole time, i.e. never giving `#input` its own diverging logical-property context) gets the same visual/caret correctness for RTL content as `dir="auto"` does here? If so, `.button`'s clearance never goes out of sync in the first place and this entire override block isn't needed — `direction` stays fixed, only the bidi text-rendering algorithm engages for the RTL content itself.

Rajdeepc (commented):
Re-reviewed after the latest commits — the shared-observer refactor, the native `attributeChangedCallback` switch, and the input `dir` fix are all in and look good (the `dir="auto"` approach for uncommitted RTL search text is better than what I'd suggested). One new thing surfaced by the button-clearance follow-up fix, left inline below.

1st-gen/packages/combobox/src/combobox.css
majornista: I'll give that a try.

majornista: That worked f55a88497a02d02c11bb80aa6594f8c12b14eb5b! Thank you for the recommendation.

majornista:
@caseyisonit, @Rajdeepc, @rise-erpelding this PR still needs review or approval.

1st-gen/packages/breadcrumbs/src/BreadcrumbItem.ts
```diff
@@ -105,11 +177,20 @@ export class BreadcrumbItem extends LikeAnchor(Focusable) {
...
 
   private renderSeparator(): TemplateResult {
+    // `:dir()` resolves directionality by walking up the `dir` *attribute*
+    // chain, independent of the CSS `direction` property — so it still
+    // picks up this host's own `dir` (set for `#item-link`'s content, per
+    // `renderLink()`) even though `:host([dir]) { direction: inherit; }` keeps the
+    // host's own *computed* direction tied to the ambient context. Read that
+    // computed value and set it explicitly here so `:dir()` on `#separator`
+    // resolves from its own accurate attribute instead of the host's.
+    const ambientDir = getComputedStyle(this).direction as 'ltr' | 'rtl';
```
Rajdeepc, line 187: Blocking: 
This will run unconditionally on every render including calls triggered by ResizeObserver. You will layout thrashing for breadcrumb trails. Please cache or memoize this only recompute in `updated()` when `dir` changed on this host or a watched ancestor.

1st-gen/tools/reactive-controllers/src/AttributeObserver.ts
```diff
@@ -0,0 +1,101 @@
...
+  if (registry.size === 0) {
+    sharedObserver = undefined;
+    return;
+  }
+
+  if (!sharedObserver) {
+    sharedObserver = new MutationObserver(processRecords);
+  }
+
+  for (const [target, attributes] of registry) {
```
Rajdeepc, line 57: Need optimisation here. Every call to `observeAttribute()` calls `reconnect()` which disconnects and re-observes every target currently in registry not just the one being added or removed.  Only newly added target actually needs a fresh `observe()` call.

.circleci/config.yml
```diff
@@ -22,7 +22,7 @@ parameters:
   # 3. Commit this change to the PR branch where the changes exist.
   current_golden_images_hash:
     type: string
-    default: 75b8c3cdc71a773889e3bdd1032f89ff73c12475
+    default: 269f7b7535187364b8879232b07e2d7bfba7e02f
```
Rajdeepc, line 25: Can you please remove this hash. Once you get two approvals we generally encourage authors to update this hash

1st-gen/packages/combobox/src/spectrum-combobox.css
```diff
@@ -286,6 +286,7 @@
   font-size: var(--mod-combobox-font-size, var(--spectrum-combobox-font-size));
   font-style: var(--mod-combobox-font-style, var(--spectrum-combobox-font-style));
   line-height: var(--mod-combobox-line-height, var(--spectrum-combobox-line-height));
+  unicode-bidi: plaintext;
```
Rajdeepc, line 289: Verify if `yarn lint:css` passes. Might double check the ordering here.

Rajdeepc (changes requested):
@majornista Nearly there. Once you are ready for a final pass @ me.

1st-gen/tools/reactive-controllers/src/AttributeObserver.ts
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  return () => {
+    listeners.delete(listener);
+    if (listeners.size === 0) {
+      attributes.delete(attribute);
+      if (attributes.size === 0) {
+        registry.delete(target);
+      }
+      reconnectAll();
```
Rajdeepc: Nice consolidation into a shared observer, this is a good pattern to have available. One thing I noticed while poking around: `reconnectAll()` fires here any time the *last* listener for a single `(target, attribute)` pair goes away — even if that `target` still has other attributes being watched (e.g. its `dir` listener unsubscribes but a `lang` listener is still active on the same element). Since `registry`/`sharedObserver` are module-level singletons also used by `LanguageResolutionController` (color-area, meter, slider, etc. all go through it), a routine breadcrumb re-render can end up forcing a full disconnect + re-observe of *every* watched element on the page, not just the one actually changing.

Would it make sense to mirror the `observeTarget()` optimization you already added for the "add" path here too — only calling `reconnectAll()` when the target is fully dropped (`attributes.size === 0`), and otherwise just re-`observeTarget()`-ing with the shrunk filter? Happy to be wrong if I'm missing something about why the full reconnect is needed here specifically.

1st-gen/tools/reactive-controllers/src/AttributeObserver.ts
```diff
@@ -0,0 +1,127 @@
...
+// left. Only removal needs this; adding a target/attribute can reuse the
+// existing observer via `observeTarget()` above.
+function reconnectAll(): void {
+  if (sharedObserver) {
+    // `disconnect()` discards any records already queued but not yet
+    // delivered to the callback, so a mutation that lands right before an
+    // unrelated component's unmount triggers `reconnectAll()` would
+    // otherwise be silently dropped. Drain and process them first.
+    processRecords(sharedObserver.takeRecords());
+    sharedObserver.disconnect();
```
Rajdeepc, line 76: Small thing to flag here, not urgent: `processRecords(sharedObserver.takeRecords())` runs listener callbacks synchronously right before `disconnect()` on the next line. If any listener ever synchronously mutates another attribute that's also being observed (not the case for anything in this PR today — both the Breadcrumbs and BreadcrumbItem callbacks just flag state and call `requestUpdate()`), the resulting new mutation record would land in the observer's queue and then get silently discarded by that immediate `disconnect()`.

Probably totally fine to leave as-is given how it's used right now, but since this is becoming a shared utility other components might build on, might be worth a comment noting that assumption so a future caller doesn't get bitten by it.

1st-gen/packages/breadcrumbs/src/breadcrumb-item.css
```diff
@@ -19,3 +19,10 @@
 :host([disabled]) {
   pointer-events: none;
 }
+
+/* A `dir` set on this host (e.g. to localize one item's text) should not
+   flip this item's own layout or its separator's mirrored chevron; only
+   `#item-link` (see BreadcrumbItem.ts) picks up that direction. */
+:host([dir]) {
```
Rajdeepc, line 26: Clever fix for keeping a single item's language from flipping the whole trail's layout — nice catch on that interaction. One side effect worth calling out somewhere (changeset or a code comment): after this rule, `breadcrumbItemEl.dir` — the JS property, via `SpectrumElement`'s computed-direction getter — now returns the *ambient* direction rather than that item's own authored `dir` attribute. Your code inside `BreadcrumbItem.ts` correctly works around this by reading `getAttribute('dir')` everywhere, so nothing here breaks internally, but any external consumer code checking `.dir` on an `sp-breadcrumb-item` would now get a different answer than it did before this PR.

Not asking for a change, just think it's worth a heads-up in the changeset so nobody's surprised by it down the line.

1st-gen/packages/breadcrumbs/src/BreadcrumbItem.ts
```diff
@@ -30,6 +32,30 @@ export interface BreadcrumbSelectDetail {
   value: string;
 }
 
+// Walks up through `parentElement`, crossing shadow-root boundaries via
+// `getRootNode().host` — needed because `Breadcrumbs.renderMenu()` places
+// an "is-menu" `BreadcrumbItem` inside its own shadow root rather than as
+// a light-DOM child, so a plain `parentElement` walk would stop there
+// without ever reaching `<sp-breadcrumbs>` itself.
+function* ancestorElements(start: Element): Generator<Element> {
```
Rajdeepc: Just flagging for awareness, not blocking: this ancestor-walking generator is pretty similar in spirit to `getAncestors()` inside `Base.ts`'s `hasVisibleFocusInTree()`, and to the `parentElement || root.host` walk in `Tooltip.ts`'s trigger-resolution code. Base's version also checks `assignedSlot`, which this one doesn't — probably doesn't matter for the breadcrumbs case today, but if this pattern is going to keep showing up, it might be worth eventually pulling into a shared "walk ancestors across shadow boundaries" helper so a fix to one doesn't miss the others.

No action needed for this PR necessarily, just leaving it here as a thought for later.

1st-gen/packages/breadcrumbs/src/BreadcrumbItem.ts
```diff
@@ -49,12 +75,72 @@ export class BreadcrumbItem extends LikeAnchor(Focusable) {
...
+  // `dir` actually changed on this host or a watched ancestor, not on every
+  // render an unrelated property change (e.g. a `ResizeObserver`-driven
+  // layout pass) triggers.
+  private ambientDir: 'ltr' | 'rtl' = 'ltr';
+  private dirNeedsResolve = true;
+
+  protected override willUpdate(): void {
+    if (this.dirNeedsResolve) {
+      this.dirNeedsResolve = false;
+      this.ambientDir = getComputedStyle(this).direction as 'ltr' | 'rtl';
```
Rajdeepc: Minor nit, feel free to ignore: this recomputes exactly what the inherited `SpectrumElement.dir` getter already returns (`getComputedStyle(this).direction ?? 'ltr'` in `Base.ts`). Could probably simplify to `this.ambientDir = this.dir as 'ltr' | 'rtl';` and drop the duplicate `getComputedStyle` call — purely cosmetic, no behavior change either way.

1st-gen/packages/breadcrumbs/src/Breadcrumbs.ts
```diff
@@ -190,10 +204,40 @@ export class Breadcrumbs extends SpectrumElement {
         value: el.value || index.toString(),
         offsetWidth: width,
         isVisible: true,
+        lang: el.lang || undefined,
```
Rajdeepc: Small spec nuance, probably not worth changing but wanted to mention it: `el.lang || undefined` treats an explicit `lang=""` the same as having no `lang` attribute at all. Per the HTML spec, `lang=""` actually means "declared unknown language" and shouldn't inherit the ambient `lang`, while omitting the attribute should. This same pattern shows up in `BreadcrumbItem.ts` and `Combobox.ts` too.

Realistically, deliberately authoring `lang=""` is pretty rare, so this is more of a "just so you know" than something I'd block on.

1st-gen/packages/combobox/src/Combobox.ts
```diff
@@ -47,8 +49,22 @@ export type ComboboxOption = {
...
+// `SpectrumElement` (the base class behind `MenuItem`) overrides `dir` to
+// return the *computed* CSS direction rather than the attribute, so a
+// slotted `MenuItem`'s authored `dir` must be read via `getAttribute`
+// instead of the `dir` property. Plain `ComboboxOption` data has no such
+// override.
+const getOptionDir = (
+  option: ComboboxOption | MenuItem
+): Directionality | undefined =>
+  normalizeDir(
+    option instanceof MenuItem ? option.getAttribute('dir') : option.dir
```
Rajdeepc: Heads-up on a pretty narrow edge case here: the `instanceof MenuItem` check would fail if a consuming app somehow ends up with two separately-bundled copies of `@spectrum-web-components/menu` (can happen with yarn/webpack dedup issues in monorepos or micro-frontends). In that situation, a genuinely slotted `<sp-menu-item>` from the "other" copy would fall into the `option.dir` branch and read the *computed* direction instead of the authored attribute — which would basically reproduce the exact bug this PR is fixing, just in that specific environment.

Likely not worth solving now given how unusual the trigger condition is, just wanted to leave a note in case it ever comes up in the wild.

Rajdeepc (commented):
Nearly there. Few architectural and edge case conditions that we would need to cover before we can land this. 
Thanks for working through my requests.

1st-gen/packages/breadcrumbs/src/BreadcrumbItem.ts
majornista: I pulled the walk into a shared `walkAncestors()` helper (with the `assignedSlot` fix)
for this file now, and filing a follow-up to migrate `Base.ts`/`Tooltip.ts` onto it separately.

majornista: See: 9567253dfd181c0a768e7ee773fe888953d1839a

1st-gen/packages/breadcrumbs/src/breadcrumb-item.css
majornista: See 534455fca6550e3312fb12326752f355b6e695e5

1st-gen/packages/breadcrumbs/src/BreadcrumbItem.ts
majornista: See 79f9642ef34bf4f65629eb2697a12fe59ea45b25.

1st-gen/packages/breadcrumbs/src/Breadcrumbs.ts
majornista: See 2c8b9a32a0090665c55aa9435c5b1b7ddbae39cb

1st-gen/packages/combobox/src/Combobox.ts
majornista: See d2c1b847b66e20138fa1cdc4c6ae2175d29e5665

1st-gen/packages/combobox/src/Combobox.ts
```diff
@@ -194,7 +210,7 @@ export class Combobox extends Textfield {
     this.optionEls.map((item) => {
       this.itemObserver.observe(item, {
         attributes: true,
-        attributeFilter: ['id'],
+        attributeFilter: ['id', 'lang', 'dir'],
```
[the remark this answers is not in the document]
majornista: See 46b268254b92329a186046f0b93f343eab8ae4c3

1st-gen/tools/reactive-controllers/src/AttributeObserver.ts
majornista: See 5f32f037eee6e7644198f513777a912d8fd239ba

1st-gen/tools/reactive-controllers/src/AttributeObserver.ts
majornista: See 63dba452922977eaaea0e15cc8e7b25dc95d67f8

majornista:
> Nearly there. Few architectural and edge case conditions that we would need to cover before we can land this. Thanks for working through my requests.

I think I've addressed your concerns.

pfulton:
@Rajdeepc would you mind taking a final look at this?

Rajdeepc (approved):
@majornista Looks ready now. Please update the golden hash and once CI is green, its ready to merge. Thanks for your patience and working through the review.

majornista:
> @majornista Looks ready now. Please update the golden hash and once CI is green, its ready to merge. Thanks for your patience and working through the review.

golden hash has been updated. I think I require one more review approval and for the UI Tests to complete.

.changeset/combobox-breadcrumbs-lang-of-parts.md
```diff
@@ -0,0 +1,13 @@
+---
+'@spectrum-web-components/combobox': patch
+'@spectrum-web-components/breadcrumbs': patch
+'@spectrum-web-components/base': patch
+'@spectrum-web-components/reactive-controllers': patch
+---
```
cdransf, line 7: Can we add `@spectrum-web-components/shared` here to account for `walk-ancestors.ts`? ✨

cdransf (changes requested):
We just need an update to the changeset. `breadcrumbs`' `package.json` has the `shared` dependency pinned, so it'll need to be bumped to pick up the new `walk-ancestors.ts` file. ✨

.changeset/combobox-breadcrumbs-lang-of-parts.md
cdransf: Added in 75ed8372a9a9671f438613b0acd06247b84eb011.

cdransf (approved):

PR #6508, opened 2026-07-14
feat(dev-validation): add reusable dev-mode validation helpers and docs

rubencarvalho:
## Description

Adds the shared foundation for dev-mode validation warnings in 2nd-gen components: reusable helpers, one gating/dispatch path, a dedup fix, and docs. **Foundation only — no component behavior changes.** Applying the helpers to components is a follow-up branch.

**New helpers** in `@spectrum-web-components/core/utils`, replacing the hand-rolled `includes()` + `window.__swc.warn()` checks each component wrote on its own:

- `validateEnum`, `validateRequiredSlot`, `validateAllowedChildren`
- `warnIf` — a general predicate that also covers required and conditionally-required properties and mutually-exclusive combinations

All warn in dev and no-op in production (via the `process.env.NODE_ENV` gate). They surface an invalid value with a console warning; they do not change it.

**One dispatch path:** every helper routes through a single private `emitWarning`, so there is one source of truth for how a warning is gated and emitted.

**Dedup fix:** the `window.__swc.warn` key now includes the message (`localName:type:level` → `…:message`), so two different warnings on the same component no longer suppress each other. Known limitation: identical messages across multiple instances of one component still collapse to one (escape hatch: `window.__swc.verbose = true`).

**`isDebug()` helper:** gates expensive call-site checks (e.g. DOM traversals) so they do not run in production.

**Docs and process:** a [Debug and validation](https://github.com/adobe/spectrum-web-components/blob/ruben/feat-dev-warning-validation-foundation/CONTRIBUTOR-DOCS/02_style-guide/02_typescript/17_debug-validation.md) style-guide page (all validation categories, a lifecycle-hook quick-reference, and a "Known limitations" section); a consumer-facing "Dev mode warnings" Storybook guide; and validation checklist items added to the migration workflow and the `code-conformance` rule.

## Out of scope (tracked as follow-ups)

- Applying the helpers across components (the next branch; includes retrofitting Badge / Tabs / ProgressCircle / IllustratedMessage's ad hoc `warn()` calls).
- A real dev/prod dual-build split so warnings are guaranteed stripped from production bundles. This package currently ships the `process.env.NODE_ENV` gate unreplaced.

## Motivation and context

Dev-warning validation exists but is inconsistently applied (~4 components, each with hand-rolled checks), and several requested categories (conditionally-required properties, mutually-exclusive combinations, required slots, allowed children) had no shared helper at all. This PR builds the reusable, documented foundation so the next phase — applying validation to every component — is mechanical and consistent instead of another one-off per component.

## Related issue(s)

- Fixes SWC-2384

N/A — no visual changes. New Storybook guide page: "Guides/Dev mode warnings".

## Author's checklist

## Reviewer's checklist

### Manual review test cases

- [ ] _New helpers behave correctly_
  1. Open `2nd-gen/packages/core/utils/test/dev-validation.test.ts` in Storybook (or run `yarn vitest --run --project storybook ../core/utils/test/dev-validation.test.ts` from `2nd-gen/packages/swc`)
  2. Confirm all 8 validation steps pass (enum / warnIf / required-slot / allowed-children, fires and does-not-fire cases)
  3. Expect all green

- [ ] _Dedup fix doesn't break existing warning tests_
  1. Run `yarn vitest --run --project storybook components/badge components/tabs components/progress-circle components/illustrated-message` from `2nd-gen/packages/swc`
  2. Expect all 119 tests to still pass unchanged

- [ ] _New Storybook guide renders_
  1. Run Storybook locally (`yarn storybook` from `2nd-gen/packages/swc`)
  2. Navigate to Guides → Dev mode warnings
  3. Expect the page to render with no console errors

### Device review

- [ ] Did it pass in Desktop?

- [ ] **Keyboard**: No focusable parts introduced by this change (utility functions and documentation only); confirm no regressions in the new Storybook guide page's own navigation (standard Storybook Docs page keyboard behavior, unchanged).
- [ ] **Screen reader**: No new DOM/ARIA surface introduced; confirm the new "Dev mode warnings" Storybook Docs page has correct heading structure (headings read in order: h1 "Dev mode warnings", h2 "What triggers a warning" / "Only in development" / "Configuring which warnings you see") when navigated with a screen reader's heading navigation (e.g. VoiceOver `VO+Cmd+H`, NVDA `H`).


1st-gen/tools/base/src/version.ts
```diff
@@ -22,4 +22,4 @@ export const version = '1.12.2';
 /**
  * The version of the core base package.
  */
-export const coreVersion = '0.3.0';
+export const coreVersion = '2.0.0-beta.2';
```
miwha-adobe: I am not sure where this came from. Is this the right call to update this version to `2.0.0-beta.2`? Our latest release was `Since 2.0.0-beta.1`

2nd-gen/packages/core/element/spectrum-element.ts
```diff
@@ -68,6 +68,22 @@ export class SpectrumElement extends SpectrumMixin(LitElement) {
...
 }
 
+/**
+ * Builds the deduplication key for a dev-mode warning. The `message` is part of
+ * the key so two distinct problems that share a `type`/`level` (the common
+ * case) do not suppress each other; only a verbatim repeat of the same warning
+ * is deduplicated. Exported so the dedup key can be unit tested without relying
+ * on the `NODE_ENV`-gated `window.__swc` setup below.
+ */
+export function warningId(
```
miwha-adobe, line 36: Extracted warningId so the dedup-key fix could be unit tested

2nd-gen/packages/swc/.storybook/guides/dev-mode/dev-mode-warnings.mdx
```diff
@@ -0,0 +1,68 @@
...
+
+If your production bundle still contains the dev-mode strings (search your built output for `"Spectrum Web Components is in dev mode"` to check), nothing is broken, but you're shipping validation code, message strings, and the warning engine to users for no benefit. Fixing that is a build configuration change on your end, not something this package can guarantee on its own. See [Known limitations](https://github.com/adobe/spectrum-web-components/blob/main/CONTRIBUTOR-DOCS/02_style-guide/02_typescript/17_debug-validation.md#known-limitations) for why.
+
+## Configuring which warnings you see
+
+Each distinct warning message fires once per element type per page load, not once per problem occurrence; this keeps a single repeated mistake from flooding your console. Two different problems on the same component (even ones that share a warning type and severity) each get their own warning; the one case that's still collapsed is **multiple instances of the same component with the identical problem**: ten badges all set to the same invalid `variant` produce one console warning, not ten, from whichever badge updates first.
```
miwha-adobe, line 38: Trade off: any mid-iteration or a re-introduced mistake goes silent after the first time. Is it worth the trade off?

2nd-gen/packages/core/element/spectrum-element.ts
```diff
@@ -68,6 +68,22 @@ export class SpectrumElement extends SpectrumMixin(LitElement) {
...
+  localName: string,
+  type: WarningType,
+  level: WarningLevel,
+  message: string
+): BrandedSWCWarningID {
+  return `${localName}:${type}:${level}:${message}` as BrandedSWCWarningID;
+}
+
 if (process.env.NODE_ENV === 'development') {
```
miwha-adobe: While "development" seems like standard practice, if a consumer stores env as 'dev', 'prod', 'staging', or NODE_ENV unset, this will fail and falls into a gap: no warnings appear, and the code isn't stripped either (it runs and no-ops).

1st-gen/tools/base/src/version.ts
caseyisonit: we actually should make a ticket to remove this and disable the script thats updating it. its no longer coupled.

2nd-gen/packages/swc/.storybook/guides/dev-mode/dev-mode-warnings.mdx
caseyisonit: i would argue no its not worth the trade off. a warning should fire always and if its making the console noisy, thats the point, fix this code. it also can help us debug rendering issues if something is firing too much or not enough. all mistakes should be documented and communicated.

> the one case that's still collapsed is **multiple instances of the same component with the identical problem**: ten badges all set to the same invalid `variant` produce one console warning, not ten, from whichever badge updates first.

The message in our warnings includes where this error occurs. i would be irrated if i was only warned once, fix it in the spot the message told me and then get the same message again and have to do that 10 more times to clear all warnings. I would rather have a list of all issues, fix it and then rerun my server to check all warnings are clear.

we shouldnt be making the dev experience easier to work with warnings, dev mode is when everything should be firing and getting fixed before prod. by hiding any message we are doing a disservice in my opinion.

2nd-gen/packages/core/element/spectrum-element.ts
caseyisonit: i would be down to extend this to cover potential gaps, great call out about staging too. prod and production should turn everything off.

2nd-gen/packages/swc/.storybook/guides/dev-mode/dev-mode-warnings.mdx
miwha-adobe: Following up on my earlier comment: I misread the doc, so please disregard my "goes silent after the first time" worry. I verified the actual behavior which is:

1. Errors surface on every update as you develop, not just once on initial render. 

2. A page reload also resets the dedup, so nothing goes quiet while you iterate. This was my main concern about things being missed or going silent.

3. Four components with the same invalid value produce 1 warning, not 4 (deduped by message). I am alright with this pattern personally, though I can see the argument in wanting to intentionally make it noisy so the developer fixes the issues. As a thought, perhaps there is something we could do to say "You have one or more  instances of error xyz.....".

My biggest concern was mid-iteration or reintroduced issues being un-captured. But I did some manual testing and that doesn't seem to be the case.

Rajdeepc: @caseyisonit @miwha-adobe I feel you both are right about a different problem.
**Only warn once means** you have 10 broken badges, we point you at one, you fix it, reload and we point you at the next one. Ten times. @caseyisonit your complaint is right and its fair here.

**Now warning every time** means a table with 500 rows bound to one bad valye gives you 500 console messages each holding onto a live DOM element. The devtools will lock up here for just one bug.

I would recommend we warn once per broken element but group them into a single console line. One message with all ten elements listed inside it and a count. This way we fix one root cause and we already knew there were ten.

2nd-gen/packages/core/element/spectrum-element.ts
Rajdeepc: @miwha-adobe @caseyisonit Only the exact word `production` can actually be deleted from our bundle. That's the specific string every build tool will look for. 
`prod` can be made to behave correctly with no warnings but the code will still ship. This is something we cannot fix from inside the package, we need to amplify this saying in the consumer docs so nobody assumed they are equivalent.

1st-gen/tools/base/src/version.ts
Rajdeepc: Fixed here: https://github.com/adobe/spectrum-web-components/pull/6510

2nd-gen/packages/core/element/spectrum-element.ts
miwha-adobe: @Rajdeepc  curious for your take.

We enable warnings on `NODE_ENV === 'development'` but strip on `=== 'production`'. Anything in between (staging, test, unset, non-standard dev) gets no warnings and no stripping.

Do consumers realistically use values like dev, and do we want warnings in staging?

Rajdeepc: Does anyone really use dev? `test` should be hit automatically by vitest and Jest, I am not sure about unset thought.
Also do we want warnings in staging, we cant answer this from inside the package. Its a deploy target not a build type. What we can fix is that there is currently no switch. window.__swc.warn only exists inside the development branch, so DEBUG = true does nothing anywhere else. This is a bug regardless of which strings we enumerate.
I would recommend we flip the default to be on unless production.

2nd-gen/packages/swc/.storybook/guides/dev-mode/dev-mode-warnings.mdx
miwha-adobe: This is what I am proposing (example just for viewing. Not commiting the storybook instance)

Screenshot 2026-07-29 at 12 35 49 PM

2nd-gen/packages/core/element/spectrum-element.ts
miwha-adobe: I agree with that assessment, and good catch about the bug! That is something I did not consider.

Rajdeepc:
This is a kind of infra work I am pleased to see. Thanks

Rajdeepc (approved):

2nd-gen/packages/core/utils/dev-validation.ts
```diff
@@ -0,0 +1,211 @@
...
+    return;
+  }
+  if (!slot) {
+    return;
+  }
+  const allowed = allowedTagNames.map((tag) => tag.toUpperCase());
+  const allowedList = allowedTagNames.map((tag) => `<${tag}>`).join(', ');
+  for (const el of slot.assignedElements()) {
+    if (!allowed.includes(el.tagName)) {
+      emitWarning(
```
rubencarvalho, line 200: small one: since each disallowed child calls `emitWarning` with the host `element`, if a single host has several bad children of the same tag they'll share a message (and id), and the grouped counter ends up reading "N elements affected" when it's really one host with N children. could you check?

miwha-adobe: Yes definitely can check

miwha-adobe: Good catch, this was definitely a bug! I pushed up a change to address that concern and keep a clean count

rubencarvalho (commented):
I approve this PR, but I can't officially approve because I'm the author 😅 So if you want to, feel free to approve it in my behalf and merge @miwha-adobe 😄

miwha-adobe (approved):
Approving on behalf of Ruben

PR #6510, opened 2026-07-15
feat(dev-validation): apply shared validation helpers across all Gen2 components

rubencarvalho:
## Description

Applies the shared dev-validation helpers introduced in #6508 (the foundation branch this PR targets) across the Gen2 component set — core base classes, shared mixins, and swc subclasses.

**What this does:**

- **Retrofits ad hoc checks onto the shared helpers.** Replaces hand-rolled `window.__swc.warn` + `includes()` call sites with `validateEnum` / `warnIf` / `validateRequiredSlot` / `validateAllowedChildren`. This reaches beyond the core base classes into the shared mixins (`PendingMixin`, `LinearProgressMixin`) and the concrete swc subclasses (`Button`, `CloseButton`, `ActionButton`), which each carried their own inline checks.

- **Adds validation where there was none** — Accordion (`density`), Popover (`size`), Avatar (numeric `size`), plus new enum checks on Tooltip (`variant`/`placement`), ActionButton (`staticColor`), and ProgressCircle / LinearProgressMixin (`staticColor` / `labelPosition`).

- **Guards expensive checks with `isDebug()`.** Where a condition is a DOM traversal (ProgressCircle's light-DOM check, Card's `actions`-slot check), the call site short-circuits when validation is off, so the traversal isn't paid on every update.

- **Fixes doc URLs** on every touched file (and the style-guide examples from the foundation branch): stale `opensource.adobe.com/spectrum-web-components/components/*` → the Gen2 docs site `spectrum-web-components.adobe.com/?path=/docs/components-*--docs`.

- **Fixes 7 missing helper imports** (Accordion, AlertBanner, Asset, Avatar, Badge, ButtonGroup, Card): used but not imported, so they transpiled under Vite but threw `ReferenceError` at runtime — only `tsc --noEmit` caught them.

<details>
<summary><strong>Full per-component breakdown (all 20)</strong></summary>

| Component | What changed |
| --- | --- |
| Badge, Tabs, IllustratedMessage | enum/required/mutually-exclusive/allowed-children checks retrofitted onto the shared helpers |
| ProgressCircle, LinearProgressMixin (Meter) | deprecation/required/value-range checks retrofitted; new `staticColor`/`labelPosition` enum checks added |
| Button, Card, Tooltip, ActionButton | combo/required checks retrofitted; new `staticColor` (ActionButton), `variant`/`placement` (Tooltip) enum checks added; Card `actions`-slot check retrofitted |
| Asset, AlertBanner, StatusLight, Popover | enum/deprecation/required checks retrofitted; new `size` enum check added (Popover) |
| Divider, Avatar, ButtonGroup | enum checks retrofitted; new numeric `size` check added (Avatar) |
| PendingMixin, CloseButton | hand-rolled `window.__swc.warn` call sites retrofitted onto `warnIf`/`validateEnum` |
| Accordion | new `density` enum check added (previously undeclared) |
| ColorLoupe, ColorHandle, Icon | reviewed, confirmed no changes needed (booleans/free-text only, per the "don't validate" list) |

</details>

## Motivation and context

Extends the helper-based pattern from the foundation branch across the entire component set — including the mixins and subclasses that validated inline — completing the "retrofit existing ad hoc call sites" goal and closing the gap on components that had no dev-mode validation at all.

## Related issue(s)

- Depends on #6508 (dev-validation foundation: helpers, docs, skills)

N/A — no visual changes; only dev-mode console warning behavior.

## Author's checklist

## Reviewer's checklist

### Manual review test cases

- [ ] _Existing warning behavior is unchanged across the whole package_
  1. Run `yarn vitest --run --project storybook` from `2nd-gen/packages/swc`
  2. Expect all tests to pass, matching pre-retrofit behavior (no runtime behavior change)

- [ ] _Console warnings still fire correctly in Storybook_
  1. Run Storybook locally (`yarn storybook` from `2nd-gen/packages/swc`)
  2. Open any touched component's playground and set an invalid enum value (e.g. `variant`, `size`, `density`, `placement`, `orientation`) via Controls
  3. Expect a console warning referencing the correct doc URL (`spectrum-web-components.adobe.com/?path=/docs/components-*--docs`)

### Device review

- [ ] Did it pass in Desktop?

- [ ] **Keyboard**: No behavioral changes; confirm no regressions in keyboard interaction across touched components (unchanged by this PR).
- [ ] **Screen reader**: No ARIA/DOM changes; confirm accessible-name-related warnings (Tabs, Avatar, Button, Popover) still fire correctly for their respective missing-label cases (unchanged checks, moved to `warnIf`).


2nd-gen/packages/core/components/meter/Meter.base.ts
```diff
@@ -68,16 +69,12 @@ export abstract class MeterBase extends LinearProgressMixin(
   protected override willUpdate(changes: PropertyValues): void {
```
miwha-adobe, line 69: In my third PR, I want to refactor this to an `update` or `updated` 

I think we need more conversations in regards to our fallback behaviors.

2nd-gen/packages/core/components/progress-circle/ProgressCircle.base.ts
```diff
@@ -230,11 +206,32 @@ export abstract class ProgressCircleBase extends SizedMixin(SpectrumElement, {
     // Apply default accessible name fallback after handling explicit label changes.
     if (changes.has('label') && !this.hasAccessibleName()) {
       this.setAttribute('aria-label', ProgressCircleBase.DEFAULT_LABEL);
-      this.warnMissingAccessibleName();
+      warnIf(
+        this,
+        true,
+        `<${this.localName}> requires an accessible name. A default label of "${ProgressCircleBase.DEFAULT_LABEL}" has been applied, but a more specific label should be provided via:`,
```
miwha-adobe, line 217: Should we be coercing warnings and mutating the consumers input? This is another candidate for a refactor PR that removes these coercions and puts the responsibility of correctly setting up the components on the consumer.

2nd-gen/packages/core/components/tabs/Tabs.base.ts
```diff
@@ -673,14 +666,13 @@ export abstract class TabsBase extends SpectrumElement {
...
-      window.__swc.warn(
-        this,
-        `<${this.localName}> requires an "accessible-label" attribute to provide an accessible name for the tablist.`,
-        'https://opensource.adobe.com/spectrum-web-components/components/tabs/',
-        { type: 'accessibility', level: 'high' }
-      );
-    }
+    warnIf(
+      this,
+      !this.accessibleLabel,
```
miwha-adobe, line 671: warnIf(this, !this.accessibleLabel, …) only reads accessibleLabel. If a consumer names the tablist via aria-labelledby (or aria-label) on the host instead, this still warns even though the element has an accessible name,  a false positive. Do we support those as valid naming paths for Tabs? If so, the condition should treat them as satisfied, example:

```
!this.accessibleLabel &&
  !this.hasAttribute('aria-label') &&
  !this.hasAttribute('aria-labelledby')
```

2nd-gen/packages/swc/components/action-button/ActionButton.ts
```diff
@@ -143,16 +147,38 @@ export class ActionButton extends PendingMixin(ButtonBase) {
...
+    | 'true'
+    | 'menu'
+    | 'listbox'
+    | 'tree'
+    | 'grid'
+    | 'dialog';
 
   @state()
-  private _ariaExpanded?: string;
+  private _ariaExpanded?: 'true' | 'false';
```
miwha-adobe, line 160: Increase type safety

2nd-gen/packages/swc/components/dropzone/Dropzone.ts
```diff
@@ -154,26 +153,13 @@ export class Dropzone extends DropzoneBase {
...
-    ) {
-      this._hasWarnedNoAccessibleName = false;
-      return;
-    }
-    if (this._hasWarnedNoAccessibleName) {
-      return;
-    }
-    this._hasWarnedNoAccessibleName = true;
-    window.__swc?.warn(
+    warnIf(
```
cdransf, line 177: Dropped `_hasWarnedNoAccessibleName` here. `warnIf`'s dedup is permanent for the session (keyed by localName:type:level), so fix-then-re-break within a session no longer re-warns. `LinearProgressMixin` kept its `_hasWarned*` and the same fix should apply here. ✨

2nd-gen/packages/core/components/progress-circle/ProgressCircle.base.ts
```diff
@@ -150,39 +151,15 @@ export abstract class ProgressCircleBase extends SizedMixin(SpectrumElement, {
...
 
   private warnDeprecatedLightDomChildren(): void {
-    if (!window.__swc?.DEBUG) {
-      return;
-    }
-    if (!ProgressCircleBase.hasMeaningfulLightDomChildren(this)) {
-      return;
-    }
-    window.__swc.warn(
+    warnIf(
```
cdransf, line 158: `hasMeaningfulLightDomChildren(this)` is evaluated as an arg, so it runs unconditionally every `updated()` — in prod too. Lost the old `isDebug()` early-return. ✨

cdransf (changes requested):

2nd-gen/packages/swc/components/dropzone/Dropzone.ts
miwha-adobe: Great catch!

2nd-gen/packages/core/components/progress-circle/ProgressCircle.base.ts
miwha-adobe: this too, great catch, ty!

2nd-gen/packages/swc/components/dropzone/Dropzone.ts
miwha-adobe: Alrighttt, so I dug into this a bit more, and both `_hasWarned` checks (LinearProgressMixin and Dropzone) don't actually seem to do anything 🤔. 

The logic reads like it should recall the warning, but the dedup underneath swallows any identical message that already fired this session. That's 2nd gen wide, not specific to these two, so the flag never gets to re-surface anything.

For Dropzone there's a second reason on top of that: the accessible name lives in `aria-label` / `aria labelledby`, which aren't reactive, so changing them after connect doesn't trigger a re-render and the check never re-runs. The `changes.has('dragged')` bit looks like a re-check but it's really just a side effect of dragging, not a direct check that the name changed.

(For LinearProgressMixin the inputs *are* reactive, so it does re-render, but the dedup still swallows the repeat. The one warning that reappears there is value-out-of-range, and only because the value is in the message text, so it's technically a different message. That's not the flag doing anything either.)

For PR 3 I'd like to add a controller that watches aria changes so we surface the errors we expect when the name actually changes. Caveat: same message still won't re-surface because of dedup. Open to discussing whether re-surfacing is a pattern we want to support, but these one-offs feel like attempts to work around this that don't really succeed.

miwha-adobe: Added Dropzone and LinearProgressMixin refactor here: https://github.com/adobe/spectrum-web-components/pull/6561

cdransf (approved):

2nd-gen/packages/core/components/button/Button.base.ts
```diff
@@ -140,15 +141,12 @@ export abstract class ButtonBase extends SizedMixin(SpectrumElement, {
 
   protected override update(changedProperties: PropertyValues): void {
     super.update(changedProperties);
```
rubencarvalho, line 143: I'm still trying to make sure I follow the docs correctly, but `super.update(changedProperties)` is called first, then the check. Per the docs the check should run before `super.update()`?

2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
```diff
@@ -475,6 +475,24 @@ export abstract class TooltipBase
 
   protected override updated(changedProperties: PropertyValues): void {
     super.updated(changedProperties);
```
rubencarvalho, line 477: Actually... I'm not sure what to think... Shouldn't we want to let any super class finish first before validation?

2nd-gen/packages/core/components/button/Button.base.ts
miwha-adobe: My understanding is that it depends on when we want to call the warning. This one is on hasIcon, which are slot derived and so they aren't reliable until the slot is rendered, which is why it is after super.update()

2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
miwha-adobe: I think that is a discussion worth having. Certainly the ones that rely on a render should (slots, etc). Do we lose anything from moving all warnings to after the render?  It would keep a consistent pattern. The benefits to a pre-render seem somewhat marginal, though we would need to update our docs to support this pattern.

rubencarvalho: My thought process here is if there is something happening in this super call, won't we want it to happen before we validate?

miwha-adobe (approved):
Approved on behalf of ruben

Issue #3, opened 2019-01-08
Make repo public


GarthDB:
What is missing to make this public?

There has been some requests on the Adobe Devs slack about it.

Westbrook:
Public, but “privately” so for just a little longer. Follow #77 for a more “official” publication.

filmaj:
Thank you all for making this happen, really appreciate it 🙏

PR #6537, opened 2026-07-22
ci(release): split gen1/gen2 changesets and publish workflows

blunteshwar:
Closed in favor of a cleaner PR from the same branch (`blunteshwar/ci-multi-stream-release-split-swc-2315`) against `main`, with a description written against the final Version PR architecture instead of patched over the original direct-push/branch-lock draft. See the new PR for the current state of this work.


Rajdeepc (commented):
The architecture is right — two genuinely independent changesets instances and no cross-publish; nice catch on the dedicated status-check endpoint in commit 4. Inline comments below. Three items that can't be anchored to a diff line:

1. **PR body + manual test case #3 are stale.** They say a `snapshot-release` PR makes `publish.yml` snapshot *both* gens; the final code has `publish.yml` doing 1st-gen only, with gen2 snapshots handled by `publish-2nd-gen.yml`'s own `pull_request` trigger. Please update the description and test #3.
2. **Contributor docs still reference the deleted root `.changeset/`:** `CONTRIBUTOR-DOCS/01_contributor-guides/06_releasing-swc.md` (lines 54, 56, 70) and `CONTRIBUTOR-DOCS/01_contributor-guides/15_changelog-strategy.md` (lines 119, 172). `.ai/skills/documentation/SKILL.md` was updated but these were missed.
3. **`research.md` is gitignored**, so the architecture writeup referenced for review context isn't visible on the PR — please paste it into a comment.

Overall: design is sound; blockers are operational safety (stuck lock) not architecture.

.github/workflows/publish-2nd-gen.yml
```diff
@@ -0,0 +1,209 @@
...
+          if [ "$DRY_RUN" == "true" ]; then
+            echo "🔎 Dry run only — nothing published or pushed" >> $GITHUB_STEP_SUMMARY
+          elif [ "$AUTH_OUTCOME" == "success" ]; then
+            echo "✅ Publish completed with tag \`${TAG}\`" >> $GITHUB_STEP_SUMMARY
+          else
+            echo "❌ Publish failed — NPM authentication did not succeed" >> $GITHUB_STEP_SUMMARY
+          fi
+
+      - name: Unlock gen2-beta after release
+        if: always() && github.event_name != 'pull_request' && env.DRY_RUN != 'true'
```
Rajdeepc: **blocking — stuck-lock recovery.** `if: always()` covers a *step* failure but not job cancellation or runner death. If this job is cancelled/superseded or the runner dies after the lock step, `release/in-progress` stays on `gen2-beta` permanently and blocks all PR merges until someone deletes it by hand. Needs a recovery path before merge: a manually-dispatchable force-unlock job or a scheduled sweeper that clears a stale `release/in-progress`.

.github/workflows/publish-2nd-gen.yml
```diff
@@ -0,0 +1,209 @@
...
+            # PRs with the snapshot-release label always use snapshot-test, same as
+            # 1st-gen's snapshot-release path - a separate, throwaway run either way.
+            WORKFLOW_TAG="snapshot-test"
+          else
+            WORKFLOW_TAG="${INPUT_TAG:-beta}"
+          fi
+          echo "tag=$WORKFLOW_TAG" >> $GITHUB_OUTPUT
+          echo "Using npm tag: $WORKFLOW_TAG"
+
+      - name: Lock gen2-beta during release
```
Rajdeepc: **should-fix — race window not closed.** The lock runs only after `checkout` + `setup-job` (dependency install), which is minutes after checkout. The window this is meant to close (checkout → push-back) is narrowed, not closed — a merge can still land before the lock applies. To fully close it, apply the lock as the first step via a direct `gh api` call (no working tree needed) before checkout.

.github/workflows/publish-2nd-gen.yml
```diff
@@ -0,0 +1,209 @@
...
+          token: ${{ secrets.RELEASE_BOT_TOKEN }}
+
+      - name: Enter changesets pre-release mode
+        if: github.event_name != 'pull_request' && !hashFiles('2nd-gen/.changeset/pre.json')
+        env:
+          TAG: ${{ steps.extract-tag.outputs.tag }}
+        working-directory: 2nd-gen
+        run: yarn changeset pre enter $TAG
+
+      - name: Verify NPM authentication
```
Rajdeepc, line 135: **should-fix — verifies nothing.** This step writes `~/.npmrc` and echoes success, so `steps.npm-auth.outcome == 'success'` is always true and gates nothing downstream (1st-gen actually checks for the OIDC token). Either add a real check (e.g. `npm whoami --registry=https://registry.npmjs.org`) or rename the step so it doesn't imply a guarantee it doesn't provide.

.github/workflows/publish-2nd-gen.yml
```diff
@@ -0,0 +1,209 @@
...
+    permissions:
+      contents: write # Required for git push (via RELEASE_BOT_TOKEN below)
+    env:
+      YARN_ENABLE_IMMUTABLE_INSTALLS: false
+      DRY_RUN: ${{ github.event.inputs.dry_run }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.ref || 'gen2-beta' }}
```
Rajdeepc: nit: `head.ref` with no `repository:` resolves the branch name against the *base* repo — works for same-repo `snapshot-release` PRs, breaks for forks. Fine if snapshot-release is internal-only; worth a one-line comment saying so.

.github/workflows/publish-2nd-gen.yml
```diff
@@ -0,0 +1,209 @@
...
+  publish:
+    needs: check-changesets
+    if: needs.check-changesets.outputs.has_changesets == 'true'
+    runs-on: ubuntu-latest
+    environment: npm-publish
+    permissions:
+      contents: write # Required for git push (via RELEASE_BOT_TOKEN below)
+    env:
+      YARN_ENABLE_IMMUTABLE_INSTALLS: false
+      DRY_RUN: ${{ github.event.inputs.dry_run }}
```
Rajdeepc: nit: `dry_run` only populates on `workflow_dispatch`; on `push`/`pull_request` this is `''`, so the Publish summary prints `Dry run | ` blank. Consider defaulting to `'false'`.

.github/workflows/publish-2nd-gen.yml
```diff
@@ -0,0 +1,209 @@
...
+          TAG: ${{ steps.extract-tag.outputs.tag }}
+        working-directory: 2nd-gen
+        run: yarn changeset publish --no-git-tag --tag $TAG
+
+      - name: Commit and push changes
+        if: github.event_name != 'pull_request' && env.DRY_RUN != 'true'
+        env:
+          RELEASE_BOT_TOKEN: ${{ secrets.RELEASE_BOT_TOKEN }}
+        run: |
+          git add .
```
Rajdeepc: nit: `git add .` is broad for a push to a protected branch — scope to the changeset/version/changelog paths so an unexpected untracked artifact can't be committed to `gen2-beta`.

.github/workflows/publish.yml
```diff
@@ -118,6 +105,14 @@ jobs:
             exit 1
           fi
 
+      - name: Lock main during release
```
Rajdeepc: **should-fix.** Same lock-after-checkout window as `publish-2nd-gen.yml` — lock is applied only after checkout + install, so the checkout → push-back race is narrowed, not closed. See the comment on the 2nd-gen workflow's lock step.

.github/workflows/publish.yml
```diff
@@ -294,12 +279,23 @@ jobs:
...
+          git remote set-url origin "https://x-access-token:${RELEASE_BOT_TOKEN}@github.com/${{ github.repository }}.git"
           git pull --rebase origin main
           git push
 
       - name: Create git tag
         if: steps.extract-tag.outputs.tag == 'latest'
         run: node ./1st-gen/scripts/create-git-tag.js
+
+      - name: Unlock main after release
+        if: always() && steps.extract-tag.outputs.tag == 'latest'
```
Rajdeepc, line 303: **blocking.** Same stuck-lock risk as `publish-2nd-gen.yml`: `if: always()` doesn't fire on job cancellation or runner death, which would leave `main` locked and block all PR merges. Needs the same recovery path (force-unlock dispatch or scheduled sweeper).

.github/workflows/publish.yml
```diff
@@ -294,12 +279,23 @@ jobs:
 
       - name: Commit and push changes
         if: steps.extract-tag.outputs.tag == 'latest'
+        env:
+          RELEASE_BOT_TOKEN: ${{ secrets.RELEASE_BOT_TOKEN }}
         run: |
           git add .
```
Rajdeepc: nit: `git add .` scope — same note as the 2nd-gen workflow; prefer scoping to changeset/version/changelog paths for a push to `main`.

.github/actions/release-branch-lock/action.yml
```diff
@@ -0,0 +1,25 @@
+name: Release branch lock
+description: >-
+  Temporarily blocks PR merges to a branch for the duration of a release run, via a
+  synthetic required status check. No-ops with a warning if the branch has no protection
```
Rajdeepc, line 4: should-fix: this description implies full coverage "for the duration of a release run," but both callers apply the lock only after checkout + dependency install, so a residual race window remains before the lock is active. Either document that caveat here or have callers lock pre-checkout via a direct `gh api` call.

.github/scripts/toggle-release-lock.mjs
```diff
@@ -0,0 +1,75 @@
...
+ */
+
+/**
+ * Blocks (or unblocks) PR merges into a release branch for the duration of a publish run,
+ * by adding (or removing) a synthetic required status check that never reports success.
+ * Required status checks only gate the GitHub merge button/API - they do not block a
+ * direct `git push` from an actor with push access, so this does not interfere with the
+ * release job's own commit-and-push step.
+ *
+ * ponytail: classic branch protection only; repos on the newer rulesets API instead of
```
Rajdeepc: nit: `ponytail:` reads like a personal TODO marker — use a conventional `NOTE:`/`TODO:` so the caveat is greppable by others.

Rajdeepc (changes requested):

.github/workflows/publish-2nd-gen.yml
blunteshwar: Added `.github/workflows/release-branch-unlock.yml` — manual `workflow_dispatch`, `branch` choice input (`main`/`gen2-beta`), gated behind the same `npm-publish` environment, calls the composite action's unlock mode directly.
  Went with manual over a scheduled sweeper since a stuck lock should be a rare edge case; easy to add a sweeper later if that assumption turns out wrong

blunteshwar:
> The architecture is right — two genuinely independent changesets instances and no cross-publish; nice catch on the dedicated status-check endpoint in commit 4. Inline comments below. Three items that can't be anchored to a diff line:
> 
> 1. **PR body + manual test case [Make repo public #3](https://github.com/adobe/spectrum-web-components/issues/3) are stale.** They say a `snapshot-release` PR makes `publish.yml` snapshot _both_ gens; the final code has `publish.yml` doing 1st-gen only, with gen2 snapshots handled by `publish-2nd-gen.yml`'s own `pull_request` trigger. Please update the description and test [Make repo public #3](https://github.com/adobe/spectrum-web-components/issues/3).
> 2. **Contributor docs still reference the deleted root `.changeset/`:** `CONTRIBUTOR-DOCS/01_contributor-guides/06_releasing-swc.md` (lines 54, 56, 70) and `CONTRIBUTOR-DOCS/01_contributor-guides/15_changelog-strategy.md` (lines 119, 172). `.ai/skills/documentation/SKILL.md` was updated but these were missed.
> 3. **`research.md` is gitignored**, so the architecture writeup referenced for review context isn't visible on the PR — please paste it into a comment.
> 
> Overall: design is sound; blockers are operational safety (stuck lock) not architecture.

Addressed all three:
  
  1. **PR body / test  3 stale** — confirmed. Will push a corrected description once these fixes are committed.
  2. **Contributor docs** — `15_changelog-strategy.md` fixed (both flagged lines, plus one more I found while in
  there: its "run `yarn changeset`" instruction would now error outright with no root `.changeset/` to resolve —
  changed to `yarn changeset:2nd-gen`). `06_releasing-swc.md` is more broadly stale than the 3 lines suggested —
  it describes a single workflow (`publish-1st-gen.yml`, which never existed) publishing
  1st-gen+core+2nd-gen+React together. Patching just those 3 lines would leave a doc that's locally correct but
  structurally still wrong, so I added an "out of date, pending rewrite" banner with a correct summary table
  instead, and fixed the 3 lines under it. Full rewrite is real work — tracking as a separate docs-phase follow-up
  rather than folding it in here.
  3. **`research.md`** — pasting in a separate comment since it's gitignored and not visible on the diff.

.github/workflows/publish-2nd-gen.yml
blunteshwar: Moved the lock to the first step, before checkout. Referencing the composite action via its full remote path (`adobe/spectrum-web-components/.github/actions/release-branch-lock@main`) instead of `./local-path` — a local-path reference needs the repo already checked out to resolve, a remote owner/repo/path one doesn't. Window's genuinely closed now, not just narrowed.

.github/workflows/publish-2nd-gen.yml
blunteshwar: Added `npm whoami --registry=https://registry.npmjs.org` right after writing `.npmrc`. Step now actually fails on a bad/missing token instead of always reporting success.

.github/workflows/publish-2nd-gen.yml
blunteshwar: Added a one-line comment above the checkout step documenting that this is same-repo-only. Not building fork support for this — `snapshot-release` is internal-only.

.github/workflows/publish-2nd-gen.yml
blunteshwar: Fixed: `DRY_RUN: ${{ github.event.inputs.dry_run || 'false' }}`

.github/workflows/publish-2nd-gen.yml
blunteshwar: Scoped to `git add 2nd-gen yarn.lock` — covers everything `changeset version` and the changelog step actually touch on this branch.

.github/workflows/publish.yml
blunteshwar: lock moved to the first step, before checkout, via the remote action reference.

.github/workflows/publish.yml
blunteshwar: Covered by the new `.github/workflows/release-branch-unlock.yml` — takes `main` or `gen2-beta` as input, so it handles both workflows' stuck-lock recovery.

.github/workflows/publish.yml
blunteshwar: Scoped to `git add 1st-gen yarn.lock`.

.github/actions/release-branch-lock/action.yml
blunteshwar: Resolved as a side effect of closing the lock-timing gap in both callers — the description's "for the duration of a release run" claim is now actually accurate, so no separate documented exception needed.

.github/scripts/toggle-release-lock.mjs
blunteshwar: Done!

PR #6552, opened 2026-07-24
feat(icon): S2 UI icon POC — generator, swc-ui-icon element, and IconBase unification

caseyisonit:
## Description

Proof-of-concept for the 2nd-gen (Spectrum 2) icon strategy from the icon RFC (`CONTRIBUTOR-DOCS/03_project-planning/05_strategies/icon-rfc.md`), implementing **Phase 0: UI icons** end to end and demonstrating consumption in one component.

- **Generator.** `yarn generate:ui-icons` converts downloaded A4U source SVGs into per-icon Lit template bundles under `components/ui-icons/icon-set/`, keyed by optical step. It runs SVGO and rewrites the A4U fill to `var(--swc-icon-color, currentColor)`. Ships the first 10 UI icons, each with five optical steps.
- **`<swc-ui-icon>` (internal).** `icon` selects the logical icon, `size` selects the matching optical step (with a nearest-step fallback), and `accessible-label` drives host-owned accessibility. It renders the template directly, so components no longer need `unsafeSVG`.
- **`IconBase` unification.** `core` holds a behavior-only base (size, `accessible-label`, and host a11y; no render or CSS). The `<swc-icon>` frame and `<swc-ui-icon>` both extend it. Sizing resolves per family: UI icons use the `ui-icon-*` token scale, the frame uses `workflow-icon-*`.
- **Consumer POC (accordion).** The accordion disclosure chevron now renders via a single `<swc-ui-icon icon="chevron" size=…>`, replacing four per-step imports, a size switch, and per-size CSS.
- **Icon source layout.** Source SVGs live under `icon-source/<family>/` (`ui/`, `workflow/`). Raw SVGs are transient build inputs and are git-ignored; only the generated art and a per-family README are tracked. The RFC is the single source of truth (the standalone strategy notes were retired).

## Motivation and context

Migrated 2nd-gen components need UI icons now, and the 1st-gen icon packages are being replaced. This establishes the RFC's Phase 0 foundation: a repeatable generator, an element that centralizes size-to-step selection (removing scattered `unsafeSVG`), and one shared `IconBase` so the frame, UI icons, and future workflow icons stay consistent.

## Related issue(s)

- Jira: SWC-2437 (icon strategy epic).
- Design defined in `icon-rfc.md`. Opened as a draft for review.

Storybook (dev): **Components → UI icons** shows the generated set and optical sizes.

---

## Author's checklist

> **Draft notes for reviewers:** no changeset yet (2nd-gen packages, POC); the `<swc-icon>` frame and `<swc-ui-icon>` are both `@status internal`.

## Reviewer's checklist

### Manual review test cases

- [ ] _UI-icon gallery renders and is size-accurate_
  1. Run `yarn workspace @adobe/spectrum-wc storybook`
  2. Go to **Components → UI icons → Available icons**
  3. Expect all 10 icons to render in `currentColor`
  4. Open **Optical sizes**; expect the chevron to change optical step across `xs`–`xl` (not just scale)

- [ ] _Playground controls_
  1. Go to **Components → UI icons → Playground**
  2. Change `icon`, `size`, and `accessible-label`
  3. Expect the rendered icon and its accessible name to update

- [ ] _Accordion chevron (consumer POC)_
  1. Go to **Components → Accordion**
  2. Expand and collapse an item; expect the chevron to rotate
  3. Change the accordion size; expect the chevron's optical step to match
  4. Switch to RTL; expect the chevron to mirror

- [ ] _Frame regression (`<swc-icon>`)_
  1. Go to **Components → Icon** stories
  2. Confirm labeled icons still expose an accessible name and decorative icons are hidden
  3. Confirm conversational-ai patterns that embed `<swc-icon>` are unaffected

### Device review

- [ ] Did it pass in Desktop?

- [ ] **Keyboard**
  
  1. Tab through a page containing `<swc-ui-icon>` / `<swc-icon>`
  2. Expect neither element to receive focus (icons are presentational; interactive usage must wrap them in a `<button>`/`<a>`)
  3. Confirm no focus-order regressions around icons

- [ ] **Screen reader**
  
  1. Navigate to a labeled icon (`accessible-label` set)
  2. Expect it announced as an image with that label (host `role="img"` + `aria-label`)
  3. Navigate to a decorative icon (no `accessible-label`)
  4. Expect it skipped (host `aria-hidden="true"`; the SVG carries no role)


2nd-gen/packages/swc/components/ui-icons/UiIcon.ts
```diff
@@ -0,0 +1,62 @@
...
+  public static override get styles(): CSSResultArray {
+    return [styles];
+  }
+
+  protected override render(): TemplateResult {
+    const iconSet = UI_ICONS[this.icon];
+    // Prefer the step for the current size; fall back to the nearest available.
+    const art =
+      iconSet?.[uiStepFor(this.size)] ??
+      (iconSet ? Object.values(iconSet)[0] : undefined);
```
Rajdeepc: So the 10 generated icons currently ships all 5 optical steps so the fallback path is never actually hit and it not tested also. Can you help me with a test where if someone ships an icon that doesnt cover all 5 steps?

caseyisonit: i would actually communcate with the icons team that full five sets are required for UI icons and a fallback isnt even necessary because thats part of the contract. since this is a manual update we would be able to push back and request the right sizes or else the update is blocked.

2nd-gen/packages/swc/stylesheets/_lit-styles/icon-base.css
```diff
@@ -9,8 +9,14 @@
  * OF ANY KIND, either express or implied. See the License for the specific language
  * governing permissions and limitations under the License.
  */
+
+/*
+ * Shared box styling for icon elements (the swc-icon frame and swc-ui-icon). Sizes a
+ * square box from the workflow-icon tokens and lets the SVG fill it, whether the SVG
+ * is slotted (frame) or rendered directly (ui-icon). Dedicated UI-icon size tokens
```
5t3ph: I know this says sizing is differed, but since this first batch is geared toward ui-icons then perhaps go ahead and swap to the correct `ui-icon-[SIZE]` tokens? They are available already in our token data.

caseyisonit: Done in 6882a15cec. `<swc-ui-icon>` now sizes from the `ui-icon-*` token scale via a new `ui-icon-sizes.css` fragment. Since this stylesheet is shared with the `<swc-icon>` frame and the two families use different size scales (RFC §7), I split the size rules out of `icon-base.css` into per-family fragments — the frame keeps `workflow-icon-*` and is otherwise unchanged.

.changeset/swc-ui-icon.md
```diff
@@ -0,0 +1,10 @@
+---
+'@adobe/spectrum-wc': minor
+'@adobe/spectrum-wc-core': minor
+---
+
+Add the internal 2nd-gen UI icon system: the `<swc-ui-icon>` element plus a build-time generator for the S2 UI icon set.
+
+- **`<swc-ui-icon>`** (internal): `icon` selects the icon-set, `size` selects the matching optical step, and `accessible-label` drives host-owned accessibility; renders generated Lit templates with no `unsafeSVG` in consuming components.
+- **Generator** (`yarn generate:ui-icons`): converts committed A4U source SVGs into per-logical-icon template bundles under `components/ui-icons/icon-set/`, rewriting fills to `var(--swc-icon-color, currentColor)`.
+- **`IconBase`**: refactored to a behavior-only base (`size` + `accessible-label` + host-owned accessibility, with no render or styles); the `<swc-icon>` frame and `<swc-ui-icon>` both extend it and share `stylesheets/_lit-styles/icon-base.css`. The frame's `label` is renamed to `accessible-label`, and its accessibility moves from the slotted SVG to the host.
```
Rajdeepc: Should we rename `label` to `accessible-label` to breaking so we have the right record when we do have customers?

2nd-gen/packages/swc/components/ui-icons/test/ui-icons.test.ts
```diff
@@ -0,0 +1,266 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+import { html } from 'lit';
+import { expect } from '@storybook/test';
+import type { Meta, StoryObj as Story } from '@storybook/web-components';
+
+import { UiIcon } from '@adobe/spectrum-wc/ui-icons';
```
Rajdeepc, line 16: You have marked `swc-ui-icon` as internal but we have `"./*"` and `"./components/*"` wildcard exports plus the vite entry globs (`components/*/index.ts`, `components/*/swc-*.ts`) is making `@adobe/spectrum-wc/ui-icons` a real entry point and not internal anymore. 
We need to decide now whether its internal units need a build-level exclusion or marking it is enough.

2nd-gen/packages/swc/scripts/generate-ui-icons.mjs
```diff
@@ -0,0 +1,151 @@
...
+  }
+  icons.get(name).set(step, svg);
+}
+
+const names = [...icons.keys()].sort();
+
+// Ensure the output directory exists, then remove previously generated modules so
+// deletions in source propagate.
+mkdirSync(outDir, { recursive: true });
+for (const stale of globSync('*.ts', { cwd: outDir })) {
```
Rajdeepc, line 86: Can you run `yarn generate:ui-icons` on a clean checkout and see if this destroys the icon set or not. Following the `.gitignore` where it marks `icon-source/{ui,workflow}/*.svg` means no source SVGs exist in the repo. So here in this line globSync will return `[]` and rmSync will write `UI_ICONS = {}`. It might break the accordion build too. 
Please verify this. You might need to add if `files.length` guard first to this generator.

Rajdeepc: Also please remove `committed A4U source SVGs` from the PR body, you have `.gitignore` which is saying the opposite.

2nd-gen/packages/swc/components/ui-icons/UiIcon.ts
```diff
@@ -0,0 +1,61 @@
...
+ * @cssprop --swc-icon-inline-size - Inline size of the icon box.
+ * @cssprop --swc-icon-block-size - Block size of the icon box.
+ */
+export class UiIcon extends IconBase {
+  /**
+   * The logical UI icon to render, matching a key in the icon-set registry (for
+   * example `chevron` or `corner-triangle`).
+   */
+  @property({ type: String })
+  public icon!: UiIconName;
```
Rajdeepc, line 47: You can add a default here or a dev-mode warning. coz any undefined will render an empty box with no warning.

2nd-gen/packages/swc/scripts/generate-ui-icons.mjs
```diff
@@ -0,0 +1,151 @@
...
+  if (stale !== 'ui-icons.types.ts') {
```
Rajdeepc: If this statement can never be false this is a dead guard.

2nd-gen/packages/swc/components/accordion/accordion-item.css
```diff
@@ -110,39 +110,21 @@
...
   rotate: 0deg;
   transition: rotate token("animation-duration-100") token("animation-ease-in-out");
 }
 
-/* Mirror for RTL so the chevron points left */
+/* Mirror for RTL so the chevron points left. The UI icon element does not yet mirror
+ * directional icons itself, so the accordion handles RTL mirroring for now. */
 :dir(rtl) .swc-AccordionItem-indicator {
   scale: -1 1;
 }
```
Rajdeepc, line 126: `direction` is not a property of `swc-icon`. Fine here but this will hit again in next migration.

2nd-gen/packages/swc/components/ui-icons/ui-icons.types.ts
```diff
@@ -0,0 +1,73 @@
...
+ * smaller step). Returns `undefined` when the bundle is missing or ships no steps.
+ */
+export function resolveUiIconArt(
+  bundle: UiIconArt | undefined,
+  size: IconSize
+): TemplateResult | undefined {
+  if (!bundle) {
+    return undefined;
+  }
+  const target = uiStepFor(size);
```
Rajdeepc, line 64: This will return undefined for anything outside `IconSize`. See below comment where it leads if it returns `undefined`

2nd-gen/packages/swc/components/ui-icons/ui-icons.types.ts
```diff
@@ -0,0 +1,73 @@
...
+  const available = UI_STEPS.filter((step) => bundle[step] !== undefined);
+  if (available.length === 0) {
+    return undefined;
+  }
+  const nearest = available.reduce((best, step) =>
+    Math.abs(step - target) < Math.abs(best - target) ? step : best
```
Rajdeepc, line 70: So every `math.abs(step - undefined)`is `NaN < NaN` is false and `reduce` will return the smallest step rather than the `m` default. 
I would recommend to use `SIZE_TO_STEP[size] ?? SIZE_TO_STEP.m` for the function which will be called from a lot of places once this scales.

CONTRIBUTOR-DOCS/03_project-planning/05_strategies/icon-rfc.md
```diff
@@ -0,0 +1,583 @@
+# RFC: 2nd-gen icon strategy (Spectrum 2)
```
Rajdeepc, line 7: This file should not be at the root. Please move this under `CONTRIBUTOR-DOCS/03_project-planning/`. You might need to update in-code section references too.

2nd-gen/packages/swc/components/accordion/AccordionItem.ts
```diff
@@ -17,12 +17,7 @@ import { html as staticHtml, unsafeStatic } from 'lit/static-html.js';
 
 import { AccordionItemBase } from '@adobe/spectrum-wc-core/components/accordion';
 
-import '../icon/swc-icon.js';
-
-import { Chevron75Icon } from '../icon/elements/Chevron75Icon.js';
-import { Chevron100Icon } from '../icon/elements/Chevron100Icon.js';
-import { Chevron200Icon } from '../icon/elements/Chevron200Icon.js';
-import { Chevron300Icon } from '../icon/elements/Chevron300Icon.js';
```
Rajdeepc, line 25: Are we expecting a golden here?

2nd-gen/packages/swc/components/ui-icons/ui-icon-sizes.css
```diff
@@ -0,0 +1,48 @@
+/**
```
Rajdeepc, line 1: This file should not be under `_lit-styles/`. This is a component CSS

2nd-gen/packages/swc/stylesheets/_lit-styles/workflow-icon-sizes.css
```diff
@@ -9,15 +9,16 @@
  * OF ANY KIND, either express or implied. See the License for the specific language
```
Rajdeepc: This file should not be under `_lit-styles/`. This is a component CSS

Rajdeepc (dismissed):
My biggest concern here is after this lands, the repo will have two parallel UI-icon systems* `components/icon/elements/` still holds 22 hand-authored icon functions and six conversational-AI units still import from it. 
Nothing here deprecates the old path or links a tracking issue for converting the rest. Without that, "two icon systems" becomes the steady state.

I see only Accordion is converted other convo AI units still import the old path. and I don't see a reason to carry both for other components and since there is no external cost to converging now. 
I would request you to convert the remaining call sites in this PR or land a deprecation marker on `icon/elements` so the duplication is explicitly temporary till we convert all.

2nd-gen/packages/swc/components/ui-icons/test/ui-icons.test.ts
caseyisonit: how do the other internal components currently do it?

Rajdeepc: Asset and Icon does the same. That's why I want us to take a decision either we cover this pre-existing gap or we set this convention. Lets not block this in this PR.

2nd-gen/packages/swc/scripts/generate-ui-icons.mjs
caseyisonit: Fixed. The generator now aborts with a clear error when no source SVGs are present, before deleting anything, so a clean checkout can't wipe the icon-set. Verified on a clean tree: the guard throws and the committed `icon-set/` is untouched.

2nd-gen/packages/swc/scripts/generate-ui-icons.mjs
caseyisonit: Removed. That guard checked for `ui-icons.types.ts`, which lives one directory above the output, so it never matched. `icon-set/` holds only generated modules, so the cleanup now clears all `*.ts`.

2nd-gen/packages/swc/components/ui-icons/ui-icons.types.ts
caseyisonit: Fixed. `uiStepFor` now defaults to the medium step for sizes outside the t-shirt scale.

2nd-gen/packages/swc/components/ui-icons/ui-icons.types.ts
caseyisonit: Fixed as suggested: `uiStepFor` now returns `SIZE_TO_STEP[size] ?? SIZE_TO_STEP.m`, avoiding the `NaN` comparison that previously selected the smallest step.

2nd-gen/packages/swc/components/ui-icons/UiIcon.ts
caseyisonit: Added a console warning when `<swc-ui-icon>` is given an `icon` outside the registry (it previously rendered an empty box with no feedback).

2nd-gen/packages/swc/components/ui-icons/ui-icon-sizes.css
caseyisonit: Moved `ui-icon-sizes.css` into `components/ui-icons/` since it's component-specific.

2nd-gen/packages/swc/stylesheets/_lit-styles/workflow-icon-sizes.css
caseyisonit: Removed `workflow-icon-sizes.css` from this PR. The `<swc-icon>` frame now takes its default size scale from the shared `icon-base.css`, and the dedicated workflow sizing stylesheet will land with the workflow icon family.

2nd-gen/packages/swc/scripts/generate-ui-icons.mjs
caseyisonit: Done. The changeset, PR body, and RFC now describe the source SVGs as downloaded, git-ignored inputs; only the generated art and metadata are committed.

.changeset/swc-ui-icon.md
caseyisonit: These are net-new 2nd-gen components; the `<swc-icon>` frame and `<swc-ui-icon>` aren't released to consumers yet, so there's no shipped API to break. The changeset stays `minor` as a net-new addition rather than `major`. The `accessible-label` naming is documented in the changeset as part of the new API.

2nd-gen/packages/swc/components/ui-icons/test/ui-icons.test.ts
caseyisonit: Agreed the wildcard exports mean `internal` isn't enforced today. Since it's a package-wide gap, the fix is a build-level exclusion or a documented convention for internal units rather than an icon-only change; tracking it as a follow-up.

caseyisonit: Agreed, and thanks for confirming it's non-blocking. Captured as a follow-up to set the convention (or a build-level exclusion) for internal units package-wide, covering Asset and Icon too.

CONTRIBUTOR-DOCS/03_project-planning/05_strategies/icon-rfc.md
caseyisonit: Done. Moved the RFC to `CONTRIBUTOR-DOCS/03_project-planning/05_strategies/icon-rfc.md` (alongside the other strategy RFCs), updated the in-repo references, and ran the contributor-docs nav script to generate its breadcrumbs and TOC.

2nd-gen/packages/swc/components/accordion/accordion-item.css
caseyisonit: Acknowledged. The durable fix is RTL mirroring inside `<swc-ui-icon>` itself, tracked for the UI-icon RTL work. For now the accordion keeps the `:dir(rtl)` transform.

2nd-gen/packages/swc/components/accordion/AccordionItem.ts
caseyisonit: The accordion's rendered output is unchanged (same chevron and box sizes), so the VRT run on this PR will confirm. This is also 2nd gen and uses chromatic not the old vrt with golden hashes

caseyisonit:
On the two-parallel-systems concern: handled via deprecation rather than a full conversion in this POC. `components/icon/elements` now carries an `@deprecated` marker pointing to `<swc-ui-icon>`, so the duplication is explicitly temporary. Accordion is converted here as the reference; converting the remaining conversational-AI call sites is tracked under the icon epic (SWC-2438).

2nd-gen/packages/core/components/icon/Icon.base.ts
```diff
@@ -10,17 +10,24 @@
...
+ * host-owned accessibility. Concrete elements own their own `render()` and styles;
+ * this base carries no template and no CSS.
  *
- * @attribute {string} label - Accessible label for the icon.
+ * The host owns semantics: when `accessibleLabel` is set the host is exposed as an
+ * image (`role="img"` with that label); when empty (the default) the host is marked
+ * `aria-hidden` and the icon is decorative. The rendered or slotted SVG is never
+ * given its own role, avoiding double-announcement.
+ *
+ * @attribute {string} accessible-label - Accessible label for the icon.
```
rise-erpelding, line 30: Could be worth mentioning in the changeset that `label` changed to `accessibleLabel`

cdransf (approved):
Love it! ✨

2nd-gen/packages/swc/components/ui-icons/README.md
```diff
@@ -0,0 +1,29 @@
+# UI icons (internal)
+
+The internal UI icon set: chevrons, checkmarks, arrows, and other control
+internals. Not published for consumers; consumed by other swc components through the
```
miwha-adobe: I've been noticing that recently claude is wrapping our docs at like 80 characters. It's more of a nuisance than anything, but wanted to point it out.

CONTRIBUTOR-DOCS/03_project-planning/05_strategies/icon-rfc.md
miwha-adobe: How long do RFC's get committed / saved in our project? Is that the right place for them, or is the hub our RFC location?

miwha-adobe (approved):
lgtm

2nd-gen/packages/swc/components/accordion/AccordionItem.ts
```diff
@@ -124,9 +105,12 @@ export class AccordionItem extends AccordionItemBase {
...
         @keydown=${this.handleHeaderKeydown}
       >
-        <swc-icon class="swc-AccordionItem-indicator" aria-hidden="true">
-          ${this.chevronForSize()}
-        </swc-icon>
+        <swc-ui-icon
+          class="swc-AccordionItem-indicator"
+          icon="chevron"
+          size=${this.size ?? 'm'}
+          aria-hidden="true"
```
rise-erpelding: Are we already setting `aria-hidden` on the icon component? Is it possible that adding it here is redundant?

rise-erpelding (commented):
Making my way through but I see that you have 2 approvals so I'm dropping my comments now instead of waiting until I've gotten through the whole PR!

2nd-gen/packages/swc/scripts/generate-ui-icons.mjs
rise-erpelding, line 1: There are a few mentions in the RFC of the generator being "family-agnostic," but the script here pretty specifically caters to UI icons, which maybe is okay given that there are some fundamental differences between the two. It might be worthwhile to adjust the language in the RFC though so that it doesn't sound like a single generator can take care of both types of icons?

2nd-gen/packages/swc/components/ui-icons/README.md
caseyisonit: yeah i started to clock that too. we can look in to it but feel free to also make a backlog ticket for future planning?

caseyisonit:
> How long do RFC's get committed / saved in our project? Is that the right place for them, or is the hub our RFC location?

@miwha-adobe great question. the RFC hub is not in a place where its easy to keep RFC up to date and accruate with decisions made through implementation if a pivot or anything needs to happen. this way too the agents can reference the RFC decisions since the hub is behind VPN. for timing we should discuss as a team?

2nd-gen/packages/swc/scripts/generate-ui-icons.mjs
caseyisonit: yeah im going to update the RFC so that there are two seperate generators for each family

2nd-gen/packages/swc/components/accordion/AccordionItem.ts
caseyisonit: yes good catch!

2nd-gen/packages/core/components/icon/Icon.base.ts
caseyisonit: i believe ive already made that update? let me know if the wording is still off?

rise-erpelding: Looking again, I think it's ok. It does mention `accessibleLabel`, just not that it changed from `label`. This was internal before, right? So probably doesn't really need mentioning.

CONTRIBUTOR-DOCS/03_project-planning/05_strategies/icon-rfc.md
```diff
@@ -0,0 +1,624 @@
...
+    - [Still open](#still-open)
+    - [Deferred to implementation (no design decision)](#deferred-to-implementation-no-design-decision)
+
+</details>
+
+<!-- Document content (editable) -->
+
+|                |                                                                   |
+| -------------- | ----------------------------------------------------------------- |
+| **Status**     | Draft for review                                                  |
```
rise-erpelding: Do we want to update the status here?

rise-erpelding (approved):
Looks good, I would say that my previous feedback is pretty minor and non-blocking! Will just need `run_vrt` added to the PR before merging 🙂

2nd-gen/packages/swc/components/accordion/AccordionItem.ts
caseyisonit: Removed. `IconBase` already sets `aria-hidden` on the host when there's no `accessible-label`, so the accordion's `<swc-ui-icon>` no longer sets it explicitly; the host owns that.

2nd-gen/packages/swc/scripts/generate-ui-icons.mjs
caseyisonit: Extracted the family-agnostic helpers (SVG cleanup, fill rewrite, kebab-casing, license/banner) into `icon-source/utils/` so a future workflow generator reuses them, and reworded the RFC to describe per-family generators sharing those utilities. This generator stays UI-specific.

2nd-gen/packages/swc/components/ui-icons/README.md
caseyisonit: Unwrapped the icon READMEs. The repo's Prettier config leaves markdown prose as authored (`proseWrap` defaults to `preserve`), so the 80-column wrapping was manual noise; I'll stop hard-wrapping docs going forward. - zoltar

CONTRIBUTOR-DOCS/03_project-planning/05_strategies/icon-rfc.md
caseyisonit: Moved here at another reviewer's request (next to `focus-management-strategy-rfc.md`) and ran the nav script for breadcrumbs/TOC. If the canonical RFC home is the wiki hub instead, happy to relocate; flagging the two views so the team can settle on one. No in-repo expiry.

2nd-gen/packages/core/components/icon/Icon.base.ts
caseyisonit: On reflection, dropped this: since the `<swc-icon>` frame was internal (net-new, unreleased), there's nothing consumer-facing to record, so the changeset stays minimal.

caseyisonit: Agreed, dropped the note. The frame was internal, so the `label` to `accessible-label` change isn't consumer-facing.

CONTRIBUTOR-DOCS/03_project-planning/05_strategies/icon-rfc.md
caseyisonit: Updated the status to "Accepted: Phase 0 (UI icons) in progress" and renumbered this PR's foundational work to Phase 0, so the RFC lines up with the epic (whose tickets start at Phase 1, repoint).

2nd-gen/packages/swc/components/accordion/AccordionItem.ts
Rajdeepc: Accepting in chromatic means updating the golden :)

PR #6561, opened 2026-07-29
feat(validation): lifecycle placement of dev warnings

miwha-adobe:
## Description

Moves pure property-value dev-warnings (`validateEnum`, plus StatusLight's deprecation warnings) from `updated()` (post-render) to the pre-render phase, so a warning about an invalid value fires before first paint instead of after. For seven components: Asset, Card, IllustratedMessage, ProgressCircle, StatusLight, Tooltip, ActionButton.

**No behavior change.** Only the timing of dev-mode warnings changes; rendered output, ARIA, and focus are untouched, and production is unaffected (dev warnings are stripped there). DOM-dependent checks stay in `updated()` by design (Card `title-as-link` / `checkActionsSupport`, ProgressCircle aria-label default).

The separate, contested fallback/coercion change is intentionally excluded and lives as a do-not-merge POC in adobe/spectrum-web-components#6564.

## Motivation and context

A property-value check in `updated()` runs after the component has already rendered once with the invalid value. Running it pre-render surfaces the warning before paint and matches the style guide (`validateEnum` in `update()` before `super.update()`).

## Related issue(s)

- fixes SWC-2384

N/A, no visual change.

---

## Author's checklist

- [x] I have added automated tests to cover my changes. _(Existing suites cover it: 894 tests pass, typecheck clean.)_
- [ ] I have included a well-written changeset if my change needs to be published. _(Not needed: dev-only, no published change.)_
- [x] I have included updated documentation if my change required it. _(Not required.)_

---

## Reviewer's checklist

### Manual review test cases

- [ ] _Invalid enum warns, with no visual change_
  1. Run Storybook in dev mode.
  2. On any affected component (e.g. [Card](?path=/docs/components-card--docs)), set an invalid enum value such as `variant="banana"`.
  3. Expect the console warning to fire and the rendering to match `main`.

### Device review

- [ ] Did it pass in Desktop?

Timing-only change; no roles, names, states, focus, or keyboard behavior are affected. Testing confirms no regression.

- [ ] **Keyboard**: Tab to [Action Button](?path=/docs/components-action-button--docs) and [Tooltip](?path=/docs/components-tooltip--docs); confirm activation and tooltip show/hide behave as on `main`.
- [ ] **Screen reader**: With VoiceOver, confirm [Status Light](?path=/docs/components-status-light--docs) and [Progress Circle](?path=/docs/components-progress-circle--docs) announce the same role and name as on `main` (including Progress Circle's default "Loading").


cdransf (approved):
Looks great! ✨

2nd-gen/packages/core/components/progress-circle/ProgressCircle.base.ts
```diff
@@ -171,6 +171,17 @@ export abstract class ProgressCircleBase extends SizedMixin(SpectrumElement, {
         this.progress = clamped;
       }
     }
+
+    if (changes.has('staticColor') && this.staticColor !== undefined) {
```
rubencarvalho, line 175: I believe this is duplicating what is below (lines 241-248)

https://github.com/adobe/spectrum-web-components/pull/6561/changes#diff-2194b94b4d869a61170a29cb27f1ce8c70cad533d6e1f21ea90d8be4687cb2b9R241-R248

rubencarvalho (dismissed):
Looks good to me! Just one little change request 😄

2nd-gen/packages/core/controllers/attribute-observer-controller/src/attribute-observer-controller.ts
```diff
@@ -0,0 +1,102 @@
...
+ * re-runs and re-reads the attributes. The controller intentionally holds no
+ * state and makes no decisions about the attributes it watches.
+ *
+ * @example
+ * ```typescript
+ * // Re-run a dev-only accessible-name warning when the aria attributes change.
+ * class MyComponent extends SpectrumElement {
+ *   private nameObserver = new AttributeObserverController(
+ *     this,
+ *     ['aria-label', 'aria-labelledby'],
```
nikkimk: Because of cross root ARIA issues, hosts shouldn't be using the `aria-label` and `aria-labelledby` attributes directly. Instead we typically have a `label` attribute that sets this. I'm wondering if, rather than creating a controller here, we align the API to have `label` and `labelledby` properties that set  `aria-label` and `aria-labelledby`, even in components where the host has the role.

2nd-gen/packages/core/controllers/attribute-observer-controller/stories/demo-hosts.ts
```diff
@@ -0,0 +1,212 @@
+/**
```
nikkimk: Thank you for using a demo component to illustrate this instead of an SWC component. It provides a simple component for us to understand how to implement and it helps us isolate issues to the controller or component. <3

nikkimk (approved):
I left a comment for discussion that i'll drop into the thread. If we decide to change course, we can always merge this and change direction in a follow up PR.

2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
```diff
@@ -471,10 +471,7 @@ export abstract class TooltipBase
     if (this.disabled && this.open) {
       this.open = false;
     }
-  }
 
-  protected override updated(changedProperties: PropertyValues): void {
-    super.updated(changedProperties);
     if (changedProperties.has('variant')) {
       const constructor = this.constructor as typeof TooltipBase;
       validateEnum(this, {
```
Rajdeepc, line 477: Not a bug: I see Tooltip and ProgressCircle fold the validateEnum check in `willUpdate` while others do in `update`. Again, functionally both is running pre-render but its an inconsistency where we are trying to create a pattern.

2nd-gen/packages/core/controllers/attribute-observer-controller/src/attribute-observer-controller.ts
rubencarvalho, line 1: Do we _need_ this controller? It feels like a big overhead for what could (maybe) be simpler... e.g., the platform already gives us `attributeChangedCallback`/`observedAttributes`, which is what [Lit uses internally](https://github.com/lit/lit/blob/c42ee1e96b8fd61f7256f61d715daef572e76e52/packages/reactive-element/src/reactive-element.ts#L1175C3-L1175C27) to sync attributes to properties, and would get us attribute-change detection without a MutationObserver instance per host (even though I only see it being used on Dropzone for now? - which may also hint at some premature optimization).

But even setting that aside (and following up our Slack thread), we already have a first-party pattern for this, using `accessible-label` / `accessibleLabel`. If we lean on that convention instead of watching `aria-label` directly, we also won't need the controller for this case.

Aaaand as a bit of a tangent, but what I think we haven't decided / established a pattern is for `labelledby`. We've been treating `aria-labelledby` as the attribute... Should we extend the `accessible-label` pattern with an `accessible-labelledby` counterpart too?

rubencarvalho: (also, if this comment is too overloaded, I'm happy to sync on this!)

2nd-gen/packages/core/components/tooltip/Tooltip.base.ts
miwha-adobe: I didn't change it because it was running pre-render, and I didn't know if willUpdate was chosen for a specific reason, but it could be changed for consistency. Additionally Ruben brought up a good point about all warnings happening post render, which depending on what we decide, would change this too.

https://github.com/adobe/spectrum-web-components/pull/6510#discussion_r3738429418

2nd-gen/packages/core/controllers/attribute-observer-controller/src/attribute-observer-controller.ts
miwha-adobe: I appreciate the feedback and it is a good question! I am in agreement that it would be better suited for an API refactor to use accessibleLabel, and then this whole controller becomes unnecessary. 

For this question: 

> Aaaand as a bit of a tangent, but what I think we haven't decided / established a pattern is for labelledby. We've been treating aria-labelledby as the attribute... Should we extend the accessible-label pattern with an accessible-labelledby counterpart too?

I think it's a good question to chat about with the team. Would we be setting up a pattern to do this for all accessible attributes?

2nd-gen/packages/core/controllers/attribute-observer-controller/src/attribute-observer-controller.ts
miwha-adobe: I think aligning the API is a great idea, it creates more consistency within the components.

miwha-adobe:
Thank you for all the feedback everyone. I removed the controller work around, and reinstated what was there, with the understanding that we will refactor the API to include the aria attribute.

2nd-gen/packages/core/components/progress-circle/ProgressCircle.base.ts
Rajdeepc: @miwha-adobe Why are we only validating against static-color and not other properties?

rubencarvalho: (still duplicating! 😛)

miwha-adobe: Oops! Will remove

rubencarvalho (approved):
🚢

Rajdeepc (approved):

2nd-gen/packages/core/components/progress-circle/ProgressCircle.base.ts
miwha-adobe: @Rajdeepc, we are validating other properties but they are handled a bit differently in each case:

- size: the SizedMixin setter already falls back to the default when it gets an invalid value
- progress: gets clamped to 0–100 in willUpdate (non-finite becomes 0).
- label / accessible name: warns and applies the "Loading" fallback in updated.
- light-DOM children: deprecation warning in updated.
- static-color is the only one that gets an explicit validateEnum, since its valid values differ per subclass with no automatic fallback.

PR #6563, opened 2026-07-30
feat(infield-button): 2nd-gen swc-infield-button migration (Phases 2–8)

Rajdeepc:
## Summary

- Adds `<swc-infield-button>`, the Spectrum 2 net-new icon button for form fields (Phases 2–8 of the washing-machine migration workflow)
- New component in `2nd-gen/packages/core/components/infield-button/` (base class) and `2nd-gen/packages/swc/components/infield-button/` (SWC layer)
- Full test suite: 13 Vitest unit tests + Playwright ARIA snapshot and a11y tests
- Storybook docs page (`infield-button.mdx`) and consumer migration guide (`migration-guide.mdx`)
- Changeset included for `@adobe/spectrum-wc` and `@adobe/spectrum-wc-core` (minor)

## Related issue(s)

- Epic SWC-2105

### API

| Attribute | Type | Notes |
| --- | --- | --- |
| `accessible-label` | `string` (required) | Maps to `aria-label` on the inner `<button>`; dev warning fires if absent on an icon-only button |
| `size` | `'s' \| 'm' \| 'l' \| 'xl'` | No default (`noDefaultSize: true`); parent field sets this |
| `quiet` | `boolean` | Removes background; leaves only the icon |
| `disabled` | `boolean` | Forwards to native `disabled` on inner `<button>` |

Content is accepted via the named `icon` slot only.

### Breaking changes from `sp-infield-button`

- Element tag: `sp-infield-button` → `swc-infield-button`
- `label` → `accessible-label`
- Default slot → named `icon` slot (`slot="icon"`)
- `block` and `inline` attributes removed (parent field manages layout)
- Link attributes removed (`href`, `target`, `download`, `rel`, `referrerpolicy`)
- CSS custom properties: `--mod-infield-button-*` → `--swc-infield-button-*`

### Deferred

- VRT stories (sizes × states matrix)
- Form-associated `type="submit"` / `type="reset"` (tracked in the migration plan as A3)

## Author's checklist

- [x] **Keyboard** (required)
  1. Open Storybook → Infield Button → any story with a field wrapper
  2. Press <kbd>Tab</kbd> to navigate through the field
  3. Expected: the `swc-infield-button` is skipped by the tab order (`tabindex="-1"`); focus goes to the text input, not the button. This is by design — keyboard behavior belongs to the parent field.

- [x] **Screen reader** (required)
  1. Open Storybook → Infield Button → Accessibility story
  2. With VoiceOver (macOS) or NVDA (Windows), navigate the three field examples (picker, search, stepper)
  3. Expected: buttons are announced with their `accessible-label` value (e.g. "Open country picker, button"), role is "button", disabled buttons announce as dimmed/unavailable
  4. Expected: inputs are announced with their label ("Country, text field", "Keyword, search field", "Quantity, spin button")


2nd-gen/packages/swc/components/infield-button/infield-button.css
```diff
@@ -0,0 +1,119 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+/* ── Host (outer container) ───────────────────────── */
+
+/* No fixed sizing here — the button self-sizes from padding + icon. The
+   consuming field controls the outer dimensions and any additional spacing. */
```
TarunAdobe: nit... maybe ask ur ai to read the repo rules and remove em dashes and fix positioning of css based on rules

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
TarunAdobe: current code lets size silently default to 'm' like a normal button, which defeats the whole point of the parent-field-driven sizing

- File: 2nd-gen/packages/core/components/infield-button/InfieldButton.base.ts:37 — class only does extends ButtonBase, no SizedMixin(ButtonBase, { noDefaultSize: true }) wrapper.
- Test that locks in the wrong behavior: 2nd-gen/packages/swc/components/infield-button/test/infield-button.test.ts:63 — expect(button.size, 'size is set to m from Overview story').toBe('m').

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
```diff
@@ -425,99 +425,100 @@ html`
...
-- [ ] Pass stylelint (property order, `no-descending-specificity`, token validation)
+- [x] Focus model: the inner `<button>` sets `outline: none` (matches S2 source). Infield buttons are not independently focusable and not in the tab order — the parent field owns the visible focus ring and all keyboard behavior. Do not add a `:focus-visible` ring on the button. Verify via field-level stories that the parent field's focus ring is visible and meets WCAG 2.4.7.
+- [x] Forced colors: `@media (forced-colors: active)` block added at bottom of file with ButtonFace/ButtonText/ButtonBorder/Highlight/GrayText system colors for each state.
+- [x] Verify no `:lang(ja)`, `:lang(ko)`, `:lang(zh)` size modifiers needed (not present in S2 baseline)
+- [x] Add a `@cssprop` JSDoc tag to the primary SWC component class for every exposed `--swc-infield-button-*` property. Exposed: `--swc-infield-button-height`, `--swc-infield-button-icon-size`, `--swc-infield-button-border-radius`, four individual corner-radius properties, `--swc-infield-button-border-color`.
+- [x] Pass stylelint (property order, `no-descending-specificity`, token validation)
```
TarunAdobe: consumers can't override height/radius/border via CSS custom properties as the plan promised; only --swc-infield-button-padding currently exists

- File: 2nd-gen/packages/swc/components/infield-button/infield-button.css — no --swc-infield-button-height, --swc-infield-button-border-radius, no per-corner radius vars, no --swc-infield-button-border-color. Border-radius is hardcoded at whatever line has border-radius: token("corner-radius-100") and border: none (search that file for border-radius/border: near the base .swc-InfieldButton rule, ~line 60ish).

CONTRIBUTOR-DOCS/03_project-planning/03_components/infield-button/migration-plan.md
```diff
@@ -425,99 +425,100 @@ html`
...
+  - Size S: `:host([size="s"])` → `token("component-height-75")`, icon `token("workflow-icon-small")`
+  - Size L: `:host([size="l"])` → `token("component-height-200")`, icon `token("workflow-icon-large")`
+  - Size XL: `:host([size="xl"])` → `token("component-height-300")`, icon `token("workflow-icon-extra-large")`
+- [x] Map visual states to `token()`:
   - Default: background `token("gray-100")`, icon `token("neutral-content-color-default")`
   - Hover: background `token("gray-200")`, icon `token("neutral-content-color-hover")`
-  - Active/down: background `token("gray-200")`, icon `token("neutral-content-color-down")`; `transform: perspective(...) translateZ(...)` on the `<button>` using the S2 downstate tokens
+  - Active/down: background `token("gray-200")`, icon `token("neutral-content-color-down")`. **No active transform** — `spectrum-css` S2 source has no transform on the active state; the plan entry was copied from button and does not apply.
```
TarunAdobe: the plan explicitly flags this as a copy-paste artifact from Button that shouldn't apply here, yet the transform is still implemented (with JSDoc describing it as an intentional "tactile press effect").

need clarification if plan is correct or if implementation is correct

TarunAdobe (dismissed):
implementation has some deviation from the migration plan... need confirmation and then either of those need to be updated

2nd-gen/packages/swc/components/infield-button/InfieldButton.ts
```diff
@@ -0,0 +1,99 @@
...
+
+import { CSSResultArray, html, TemplateResult } from 'lit';
+import { classMap } from 'lit/directives/class-map.js';
+import { ifDefined } from 'lit/directives/if-defined.js';
+
+import {
+  INFIELD_BUTTON_VALID_SIZES,
+  InfieldButtonBase,
+  type InfieldButtonSize,
+} from '@adobe/spectrum-wc-core/components/infield-button';
```
blunteshwar, line 21: This imports `@adobe/spectrum-wc-core/components/infield-button`, but that subpath is missing from the core package’s exports map in its`package.json`. Consumers—and potentially the SWC build—will be unable to resolve the new component. Add the infield-button export to `2nd-gen/packages/core/package.json`.

2nd-gen/packages/core/components/infield-button/InfieldButton.base.ts
```diff
@@ -0,0 +1,63 @@
...
+ * clickable via pointer only; they are not in the tab order and not independently focusable.
+ * All keyboard behavior is owned by the parent field, which also provides the visible
+ * focus ring. The parent field is expected to set the `size` attribute.
+ *
+ * @slot icon - Icon to display inside the button.
+ *
+ * @attribute {ElementSize} size - Size delegated from the parent field. Defaults to
+ *   medium when no parent sets it, though standalone usage is not supported.
+ */
+export abstract class InfieldButtonBase extends ButtonBase {
```
blunteshwar, line 37: Extending `ButtonBase` inherits its `SizedMixin` configuration, which defaults to m and reflects `size="m"` during update. Overriding `VALID_SIZES` does not enable `noDefaultSize`. This contradicts the documented contract that the parent supplies the size. Configure this component with noDefaultSize: true and add a test for an unset size.

2nd-gen/packages/swc/components/infield-button/infield-button.css
```diff
@@ -0,0 +1,112 @@
...
+}
+
+:host([size="xl"]) {
+  --swc-infield-button-padding: token("accessory-item-padding-extra-large");
+  --swc-infield-button-icon-size: token("workflow-icon-extra-small");
+}
+
+:host([quiet]) {
+  --swc-infield-button-background-color: transparent;
+  --swc-infield-button-background-color-disabled: transparent;
```
blunteshwar, line 41: The quiet override only makes the default and disabled backgrounds transparent. Hover, active, and focus use separate custom properties and therefore fall back to gray backgrounds. Override the hover, active, and focus background properties here as transparent too.
```suggestion
:host([quiet]) {
  --swc-infield-button-background-color: transparent;
  --swc-infield-button-background-color-hover: transparent;
  --swc-infield-button-background-color-active: transparent;
  --swc-infield-button-background-color-focus: transparent;
  --swc-infield-button-background-color-disabled: transparent;
}
```

blunteshwar (changes requested):
This PR correctly updates Infield Button, but it also removes previously completed stages for Close Button, Popover, and Progress Bar from `CONTRIBUTOR-DOCS/03_project-planning/02_workstreams/02_2nd-gen-component-migration/01_status.md:39`.If this is not intentional ,restore those three rows to their values from main.

blunteshwar (changes requested):
I don’t think we should implement noDefaultSize by intercepting update() and attributeChangedCallback() in InfieldButtonBase.
The current implementation allows ButtonBase’s SizedMixin to add size="m", suppresses Lit’s handling of that mutation, and then removes the attribute afterward. This works for the initial-render case covered by the new test, but it works around the mixin rather than configuring its behavior.
There are several problems with this approach:
Removing a parent-provided size restores size="m"
Consider the normal parent lifecycle:
```ts
button.setAttribute('size', 'l');
button.removeAttribute('size');
```
When the parent removes the attribute, `_suppressSizeAutoSet` is false. Lit forwards the removal through the inherited size setter. Because ButtonBase’s SizedMixin was created with noDefaultSize=false, the setter selects m as its fallback and writes size="m" back to the host.
The component therefore returns to:
```html
<swc-infield-button size="m">
```
rather than the intended no-attribute state.

Every update performs unnecessary attribute mutations
Whenever the component has no explicit size, super.update() adds size="m" and the override removes it again. Although the component suppresses its own callback processing, external MutationObservers still see those mutations, and the browser may perform unnecessary selector/style work.

The suppression flag is fragile
The flag depends on manually pairing assignments around super.update(), setAttribute(), and removeAttribute(). If super.update() throws, the flag can remain enabled and later legitimate size attribute changes may bypass super.attributeChangedCallback().

It duplicates behavior that belongs in SizedMixin
SizedMixin already owns the noDefaultSize concept. The limitation is only that it captures the option in a closure, preventing a subclass such as InfieldButtonBase from overriding the decision inherited through ButtonBase.

A smaller root-cause fix would make that existing decision available as an overridable static:
```ts
export interface SizedElementConstructor {
  readonly VALID_SIZES: readonly ElementSize[];
  readonly NO_DEFAULT_SIZE: boolean;
}

class SizedElement extends constructor {
  static readonly VALID_SIZES = validSizes;
  static readonly NO_DEFAULT_SIZE: boolean = noDefaultSize ?? false;

  public set size(value: ElementSize) {
    const sizedConstructor =
      this.constructor as unknown as SizedElementConstructor;
    const fallbackSize = sizedConstructor.NO_DEFAULT_SIZE
      ? null
      : defaultSize;

    // Existing validation and update behavior…
  }

    const sizedConstructor =
      this.constructor as unknown as SizedElementConstructor;

    if (
      !this.hasAttribute('size') &&
      !sizedConstructor.NO_DEFAULT_SIZE
    ) {
      this.setAttribute('size', this.size);
    }

    super.update(changes);
  }
}
```
InfieldButtonBase would then only need:
```ts
static override readonly NO_DEFAULT_SIZE = true;
```
Existing components remain unchanged because the static defaults to the current noDefaultSize option. This also preserves direct uses such as SizedMixin(..., { noDefaultSize: true }), while allowing subclasses of an already-sized base class to opt out correctly.
The new test is useful as an initial-render assertion, but it is insufficient because it only creates a bare component and checks that size is absent after the first completed update. It does not exercise the failure path introduced by this implementation.
At minimum, please also test removal of a parent-provided size:
```ts
bare.setAttribute('size', 'l');
await bare.updateComplete;
expect(bare.getAttribute('size')).toBe('l');

bare.removeAttribute('size');
await bare.updateComplete;
expect(
  bare.hasAttribute('size'),
  'removing a parent-provided size does not restore size="m"'
).toBe(false);
```
That test should fail with the current callback-suppression implementation and pass with the static SizedMixin approach. The static approach prevents the incorrect write at its source, avoids lifecycle interception and repeated DOM mutations, and correctly handles both initial rendering and later parent-driven size removal.

blunteshwar (approved):
LGTM!

2nd-gen/packages/swc/components/infield-button/infield-button.css
```diff
@@ -0,0 +1,115 @@
...
+
+* {
+  box-sizing: border-box;
+}
+
+/* ── Fill (button) ────────────────────────────────── */
+
+.swc-InfieldButton {
+  --_swc-infield-button-padding: var(--swc-infield-button-padding, token("accessory-item-padding-medium"));
+  --_swc-infield-button-icon-size: var(--swc-infield-button-icon-size, token("workflow-icon-3x-small"));
```
aramos-adobe, line 52: Let's add a perspective token to show active press animation on the button like action-button and button. Since infield button isn't using fixed sizing I used the padding and icon size tokens

```css
  --_swc-infield-button-perspective-size: calc(var(--_swc-infield-button-padding) * 2 + var(--_swc-infield-button-icon-size));
```

2nd-gen/packages/swc/components/infield-button/infield-button.css
```diff
@@ -0,0 +1,115 @@
...
+  margin: 0;
+  color: var(--swc-infield-button-icon-color, token("neutral-content-color-default"));
+  background-color: var(--swc-infield-button-background-color, token("gray-100"));
+  border: none;
+  border-radius: token("corner-radius-100");
+  appearance: none;
+  cursor: default;
+  transition-timing-function: token("animation-ease-in-out");
+  transition-duration: token("animation-duration-100");
+  transition-property: background-color, color;
```
aramos-adobe:
```css
  transition-property: background-color, color, transform;
```

2nd-gen/packages/swc/components/infield-button/infield-button.css
```diff
@@ -0,0 +1,115 @@
...
+
+.swc-InfieldButton:hover {
+  color: var(--swc-infield-button-icon-color-hover, token("neutral-content-color-hover"));
+  background-color: var(--swc-infield-button-background-color-hover, token("gray-200"));
+}
+
+.swc-InfieldButton:active {
+  color: var(--swc-infield-button-icon-color-active, token("neutral-content-color-down"));
+  background-color: var(--swc-infield-button-background-color-active, token("gray-200"));
+}
```
aramos-adobe:
```css
transform: var(--swc-infield-button-down-state-transform, perspective(var(--_swc-infield-button-perspective-size)) translate3d(0, 0, token("component-size-difference-down")));
  will-change: transform;
```

aramos-adobe: Here's where the animation on `:active` class would take place like [RSP's](https://react-spectrum.adobe.com/ComboBox) infield button in number field, combobox, et

2nd-gen/packages/swc/components/infield-button/infield-button.css
```diff
@@ -0,0 +1,115 @@
...
+
+/* ── States ───────────────────────────────────────── */
+
+.swc-InfieldButton:hover {
+  color: var(--swc-infield-button-icon-color-hover, token("neutral-content-color-hover"));
+  background-color: var(--swc-infield-button-background-color-hover, token("gray-200"));
+}
+
+.swc-InfieldButton:active {
```
aramos-adobe: Be sure to add `:not(:disabled)` as well so it doesn't show the active/pressed state while disabled.

2nd-gen/packages/swc/components/infield-button/migration-guide.mdx
```diff
@@ -0,0 +1,236 @@
...
+removed. Replace `<sp-infield-button href="…">` with a native link:
+
+```html
+<!-- Before -->
+<sp-infield-button href="/destination" label="Go to destination">
+  <sp-icon-arrow></sp-icon-arrow>
+</sp-infield-button>
+
+<!-- After -->
+<a href="/destination">Go to destination</a>
```
aramos-adobe: Was this supposed to be something like this to show the removal of the link?

```js
<swc-infield-button label="Go to destination">
  <svg slot="icon">
</swc-infield-button>
```

2nd-gen/packages/swc/components/infield-button/migration-guide.mdx
```diff
@@ -0,0 +1,236 @@
...
+  <sp-icon-chevron-down></sp-icon-chevron-down>
+</sp-infield-button>
+
+<!-- After: place the button directly in the field's flex container -->
+<swc-infield-button accessible-label="Open picker" size="m">
+  <svg slot="icon" aria-hidden="true" focusable="false">…</svg>
+</swc-infield-button>
+```
+
+### 5. Remove any link-style usage
```
aramos-adobe: Based on S1's version, I could be wrong, but there's no documentation of `href` being used. So maybe we don't need this step at all?

aramos-adobe (approved):
Really great work on this! The only other thing that jumped out was the overflow of code strings in the migration guide tables. That's a global Storybook CSS thing. 

Image

PR #6569, opened 2026-07-31
chore(action-button): add vrt coverage

aramos-adobe:
## Description

Adds Storybook VRT coverage `swc-action-button` which follows the same VRT model as `swc-button`

Files added:
- `action-button.vrt.ts`: 
- `action-button-custom-properties.vrt.ts`
- `action-button-global-styles.vrt.ts`

## Related issue(s)

- fixes SWC-2398

## Screenshots (if appropriate)

---

## Author's checklist

- [ ] I have read the **[CONTRIBUTING](<(https://github.com/adobe/spectrum-web-components/blob/main/CONTRIBUTING.md)>)** and **[PULL_REQUESTS](<(https://github.com/adobe/spectrum-web-components/blob/main/PULL_REQUESTS.md)>)** documents.
- [ ] I have reviewed at the Accessibility Practices for this feature, see: [Aria Practices](https://www.w3.org/TR/wai-aria-practices/)
- [ ] I have added automated tests to cover my changes.
- [ ] I have included a well-written changeset if my change needs to be published.
- [ ] I have included updated documentation if my change required it.

---

## Reviewer's checklist

- [ ] Includes a Github Issue with appropriate flag or Jira ticket number without a link
- [ ] Includes thoughtfully written changeset if changes suggested include `patch`, `minor`, or `major` features
- [ ] Automated tests cover all use cases and follow best practices for writing
- [ ] Validated on all supported browsers
- [ ] All VRTs are approved before the author can update Golden Hash

### Manual review test cases

This PR only adds `.vrt.ts` test stories; it does not change the `swc-action-button` component, so no dependent components are affected. Review focuses on the new stories rendering as intended.

- [ ] _Permutations story renders every size, treatment, anatomy, and state_
  1. Run `yarn storybook` and open **Components → Action button → Action button VRT → Permutations** (or view the Chromatic snapshot)
  2. Confirm the grid shows every size (`xs`–`xl`) in both the default and quiet treatments, in label-only and icon+label anatomy, plus the icon-only anatomy row
  3. Confirm the disabled and pending rows render, and the forced hover / focus-visible / active rows each show the corresponding interaction state
  4. Confirm the full matrix is repeated in light/LTR and dark/RTL, the static-color white/black groups sit on their dark/light contrast backgrounds, and the CJK language row renders correctly

- [ ] _Forced colors story renders under high-contrast emulation_
  1. Open **Action button VRT → Forced colors**
[...]

PR #6463, opened 2026-07-01
test(swc): add dedicated Chromatic VRT stories

rubencarvalho:
## Summary

- Add dedicated `.vrt.ts` Storybook mode for Chromatic so only intentional VRT stories are snapshotted.
- Add shared VRT helpers for permutations, themes, static colors, pseudo-states, and custom-property coverage checks.
- Apply the pattern to Button with split permutation, global-style, and custom-property VRT stories.
- Document the dedicated VRT authoring pattern for migrations and contributors.

## Test plan

- Storybook Vitest for Button VRT stories.
- Prettier and ESLint on touched VRT/helper files.
- `yarn lint:ai` for AI/docs updates.
- Contributor docs nav script run; it reports existing unrelated broken links in sidenav/toast analysis docs.

PR #6576, opened 2026-08-04
chore(meter): add vrt coverage

Rajdeepc:
## Description

Adds dedicated Storybook VRT coverage for `swc-meter`, following the foundation established in the button/action-button VRT work:

- `2nd-gen/packages/swc/components/meter/test/vrt/meter.vrt.ts` — `Permutations` and `ForcedColors` stories covering every size, every variant, both label positions, value states (0/25/50/75/100%, plus an over-100% case with a clamped bar and an overridden `value-label`), full anatomy (label-only, label + description, accessible-label-only fallback, custom value text), behaviors (custom range, format options), text wrapping, CJK language rendering, and static colors on their contrast backgrounds.
- `2nd-gen/packages/swc/components/meter/test/vrt/meter-custom-properties.vrt.ts` — one reference/override row per each of the 6 documented `--swc-linear-progress-*` custom properties, with coverage verified against the generated custom-elements manifest.

No `meter-global-styles.vrt.ts` was added: unlike button, there is no generated global stylesheet for meter/linear-progress to cover. Meter is also non-focusable and read-only, so unlike button/action-button/accordion there are no `:hover`/`:focus-visible`/`:active` rules to force in a VRT snapshot.

## Motivation and context

Meter currently has no dedicated visual regression coverage. Chromatic VRT catches rendering regressions (color, layout, forced-colors/high-contrast) that unit tests can't, and this closes that gap for meter the same way #6463 and #6569 did for button and action-button.

## Related issue(s)

- SWC-2411

N/A — new Chromatic snapshots will be generated on this PR's first VRT run for review.

## Author's checklist

## Reviewer's checklist

### Manual review test cases

- [ ] _Meter VRT stories render as expected_
  1. Run Storybook (`yarn storybook`) in `2nd-gen/packages/swc`
  2. Go to **Meter → Meter VRT → Permutations**, **Forced colors**, and **Custom properties**
  3. Confirm each row renders the labeled size/variant/state/anatomy/behavior/language/static-color case correctly, the forced-colors story shows `ButtonFace`/`ButtonText` track and fill overrides, and each custom-property override cell visibly differs from its reference cell

- [ ] _Chromatic snapshots are accurate_
  1. Run the project VRT command scoped to meter
  2. Review the generated snapshots for each option/state/variant combination
  3. Confirm no unintended diffs against existing shipped meter visuals


cdransf (approved):
Looks great! ✨

2nd-gen/packages/swc/components/meter/test/vrt/meter.vrt.ts
```diff
@@ -0,0 +1,273 @@
...
+        formatOptions: { style: 'currency', currency: 'USD' },
+      }),
+    ],
+    'Behaviors'
+  )}
+  ${row(
+    [
+      renderMeter({
+        label:
+          'A label long enough to wrap onto multiple lines within the available inline space',
```
pfulton, line 224: It would be nice if we could show the wrapping. There's plenty of text here, but the Meter's width is wide enough to accommodate all of the text without actually wrapping when the snapshot is taken.

Image

2nd-gen/packages/swc/components/meter/test/vrt/meter.vrt.ts
```diff
@@ -0,0 +1,273 @@
...
+  `,
+  parameters: vrtParameters,
+};
+
+// `forced-colors` replaces the whole page palette, so it can't be scoped to
+// a subtree the way theme()'s light/dark split is, and needs its own
+// snapshot rather than folding into Permutations. Confirms the
+// `@media (forced-colors: active)` override in linear-progress-base.css
+// (ButtonFace track, ButtonText fill) applies regardless of variant or
+// static-color.
```
pfulton: I don't think this comment is adding much value in its current state. How would you feel about removing it?

2nd-gen/packages/swc/components/meter/test/vrt/meter.vrt.ts
```diff
@@ -0,0 +1,273 @@
...
+      ),
+      staticColor
+    )
+  )}
+`;
+
+// VRT stories
+
+// Rendered once in light/ltr and once in dark/rtl (that combination covers
+// both axes), all still in a single story so it costs one snapshot.
```
pfulton: I don't know that this comment is adding much value in its current state. How would you feel about removing it?

pfulton (commented):
I left a few comments here, @Rajdeepc. I'll leave it to your judgement as to whether you keep or get rid of the AI-generated comments that I left feedback on. For the wrapping label example, if you're able to get that one to actually wrap in the snapshot, that would be valuable.

miwha-adobe:
Can we adjust the color from magenta, to something easier to view? The contrast fails:https://webaim.org/resources/contrastchecker/

Screenshot 2026-08-05 at 1 33 09 PM

Screenshot 2026-08-05 at 1 33 53 PM

miwha-adobe (changes requested):
Please see comment about color contrast above

2nd-gen/packages/swc/components/meter/test/vrt/meter.vrt.ts
Rajdeepc: Fixed in ba1f188a76 — constrained the meter to `max-inline-size: 200px` for this case so the label actually wraps onto multiple lines in the snapshot now.

2nd-gen/packages/swc/components/meter/test/vrt/meter.vrt.ts
Rajdeepc: Removed in ba1f188a76.

2nd-gen/packages/swc/components/meter/test/vrt/meter.vrt.ts
Rajdeepc: Removed in ba1f188a76.

Rajdeepc:
@miwha-adobe Thanks for catching that! Fixed in ba1f188a76 — swapped \`--swc-linear-progress-text-color\`'s override from \`magenta\` to \`darkmagenta\` (~7.5:1 contrast on the default light background vs. magenta's ~2.3:1), so it now passes WCAG AA while still rendering obviously different from the default. Left \`fill-color\`/\`track-color\` as \`magenta\` since those aren't text colors and aren't subject to the same contrast check.

2nd-gen/packages/swc/components/meter/test/vrt/meter-custom-properties.vrt.ts
```diff
@@ -0,0 +1,90 @@
...
+// `<swc-meter>` (see Meter.ts's `@cssprop` list) is a public contract that
+// `<swc-meter>` shares with the not-yet-migrated `<swc-progress-bar>`. One
+// row per property: a reference meter next to the same meter with that one
+// property overridden to an obviously different value, so a real difference
+// confirms the override still works.
+type MeterPropertyCase = CustomPropertyCase<`--swc-linear-progress-${string}`>;
+
+const MOD_PROPERTY_CASES: readonly MeterPropertyCase[] = [
+  { property: '--swc-linear-progress-fill-color', value: 'magenta' },
+  { property: '--swc-linear-progress-track-color', value: 'magenta' },
```
miwha-adobe, line 50: What if we go, deep-plum instead for all three of these? I checked out deep-magenta, and it still fails for the fill (though the text is great and has the 7:5:1 like you said against white). 

Also I don't know that the comment is necessarily needed and or can be condensed. I feel like claude tends to over explain in it's comments.

Image

Image

miwha-adobe (changes requested):

2nd-gen/packages/swc/components/meter/test/vrt/meter-custom-properties.vrt.ts
Rajdeepc: Good catch, and that explains it: I dug into the actual token values and the fill-vs-track boundary is the real failure mode, not text-vs-page. `track-color`'s default is gray `rgb(218,218,218)`, but the default `fill-color` is blue `rgb(59,99,251)` — when `track-color` gets overridden, it's the *blue fill* it needs contrast against, not the page, and any magenta-family dark color is too close in luminance to that blue (~1.77:1, needs 3:1). That's the case that was still failing.

Fixed in 39ceff18fe: switched all three color properties to a single `#1a001a` deep plum, which clears 3:1 against both the default blue fill and the default gray track, and 4.5:1 as text on the page background. Also condensed the comment down to two lines per your note — appreciate the callout, will keep an eye on that.

miwha-adobe (approved):
Lgtm!

PR #6579, opened 2026-08-04
chore(icons): centralize rtl handling for icons in swc-ui-icon

cdransf:
## Description

Moves chevron/arrow RTL mirroring into `<swc-ui-icon>` itself (new `ui-icon-direction.css`, curated `chevron`/`arrow` list, `icon` now `reflect: true`). Removes `accordion-item.css`'s now-redundant mirror rule (kept its RTL+open rotation compensation, which is genuinely Accordion-specific). Adds 5 tests covering direct + inherited `dir`, both curated icons, and non-directional exclusion. Note: `:host(:dir(rtl))[icon="x"]` silently fails to match in Chromium; both conditions must be inside `:host()`'s own argument (`:host(:dir(rtl)[icon="x"])`), documented in the CSS.

## Motivation and context

RFC icon-rfc.md section 11 (resolved): mirror directional UI icons via `:dir(rtl)` in the UI icon stylesheet, owned by `<swc-ui-icon>`, not components.

## Related issue(s)

- Jira: SWC-2440

None. Accordion's VRT snapshot (covers RTL) is byte-identical before/after.

## Author's checklist

## Reviewer's checklist

### Manual review test cases

- [ ] Storybook, any component with a chevron/arrow `swc-ui-icon` (e.g. Accordion): flip page `dir` to `rtl`, confirm it mirrors; flip back, confirm it un-mirrors.
- [ ] Confirm a non-directional icon (e.g. checkmark) never mirrors under RTL.

### Device review

- [ ] Did it pass in Desktop?

- [ ] **Keyboard**: N/A, visual-only change, no new interactive elements.
- [ ] **Screen reader**: N/A. `IconBase`'s existing host-owned a11y (decorative by default, `role="img"` when labeled) is untouched; the icon's own inner SVG was already never announced.

### DIFF
--- a/2nd-gen/packages/swc/components/accordion/accordion-item.css
+++ b/2nd-gen/packages/swc/components/accordion/accordion-item.css
@@ -111,25 +111,22 @@
 /* ── Disclosure indicator (chevron) ─────────────────────────────────── */
 
 /* swc-ui-icon maps the item size to the correct optical step and sizes its own box
- * from the ui-icon-* token scale, so the accordion only passes size. Rotation and RTL
- * mirroring remain here as accordion-specific behavior. */
+ * from the ui-icon-* token scale, so the accordion only passes size. swc-ui-icon also
+ * owns RTL mirroring for the chevron itself; rotation on open remains here as
+ * accordion-specific behavior. */
 .swc-AccordionItem-indicator {
   flex-shrink: 0;
   rotate: 0deg;
   transition: rotate token("animation-duration-100") token("animation-ease-in-out");
 }
 
-/* Mirror for RTL so the chevron points left. The UI icon element does not yet mirror
- * directional icons itself, so the accordion handles RTL mirroring for now. */
-:dir(rtl) .swc-AccordionItem-indicator {
-  scale: -1 1;
-}
-
 :host([open]) .swc-AccordionItem-indicator {
   rotate: 90deg;
 }
 
-/* RTL + open: rotate the opposite direction */
+/* RTL + open: swc-ui-icon has already mirrored the chevron horizontally, so rotating
+ * the same direction as LTR would point the wrong way; rotate the opposite direction
+ * to still land on "pointing down" once open. */
 :host([open]):dir(rtl) .swc-AccordionItem-indicator {
   rotate: -90deg;
 }
--- a/2nd-gen/packages/swc/components/ui-icons/UiIcon.ts
+++ b/2nd-gen/packages/swc/components/ui-icons/UiIcon.ts
@@ -18,14 +18,17 @@ import { UI_ICONS, UiIconName } from './icon-set/index.js';
 import { resolveUiIconArt } from './ui-icons.types.js';
 
 import iconBaseStyles from '../../stylesheets/_lit-styles/icon-base.css';
+import uiIconDirectionStyles from './ui-icon-direction.css';
 import uiIconSizeStyles from './ui-icon-sizes.css';
 
 /**
  * An internal icon renderer for Spectrum UI icons (chevrons, checkmarks, arrows, and
  * other control internals). The `icon` attribute selects the icon-set, and the
  * element renders the optically-tuned step that matches its `size`. Extends
- * `IconBase` for size and host-owned accessibility. Not published for consumers;
- * used by other swc components.
+ * `IconBase` for size and host-owned accessibility. A curated set of directional
+ * icons (chevrons, arrows) mirror automatically in RTL; consuming components do not
+ * need their own mirror rule for these icons. Not published for consumers; used by
+ * other swc components.
  *
  * @element swc-ui-icon
  * @status internal
@@ -41,13 +44,15 @@ import uiIconSizeStyles from './ui-icon-sizes.css';
 export class UiIcon extends IconBase {
   /**
    * The logical UI icon to render, matching a key in the icon-set registry (for
-   * example `chevron` or `corner-triangle`).
+   * example `chevron` or `corner-triangle`). Reflected so RTL mirroring can key off
+   * the host attribute regardless of whether a consumer sets `icon` via markup or a
+   * property binding.
    */
-  @property({ type: String })
+  @property({ type: String, reflect: true })
   public icon!: UiIconName;
 
   public static override get styles(): CSSResultArray {
-    return [iconBaseStyles, uiIconSizeStyles];
+    return [iconBaseStyles, uiIconSizeStyles, uiIconDirectionStyles];
   }
 
   protected override willUpdate(changed: PropertyValues<this>): void {
--- a/2nd-gen/packages/swc/components/ui-icons/UiIcon.ts
+++ b/2nd-gen/packages/swc/components/ui-icons/UiIcon.ts
@@ -26,7 +26,7 @@ import uiIconSizeStyles from './ui-icon-sizes.css';
  * other control internals). The `icon` attribute selects the icon-set, and the
  * element renders the optically-tuned step that matches its `size`. Extends
  * `IconBase` for size and host-owned accessibility. A curated set of directional
- * icons (chevrons, arrows) mirror automatically in RTL; consuming components do not
+ * icons (chevrons, arrows,corner-triangle) mirror automatically in RTL; consuming components do not
  * need their own mirror rule for these icons. Not published for consumers; used by
  * other swc components.
  *
--- a/2nd-gen/packages/swc/components/ui-icons/test/ui-icons.test.ts
+++ b/2nd-gen/packages/swc/components/ui-icons/test/ui-icons.test.ts
@@ -192,18 +192,14 @@ export const LabelTogglingTest: Story = {
 // TEST: RTL mirroring
 // ──────────────────────────────────────────────────────────────
 
-// Iterates every entry in UI_ICONS against DIRECTIONAL_UI_ICONS rather than
-// hand-picking one directional and one non-directional icon: adding an icon to
-// UI_ICONS without deciding whether it belongs in DIRECTIONAL_UI_ICONS now fails
-// this test immediately instead of silently shipping an icon nobody made a
-// mirroring decision for.
+// Adding an icon to UI_ICONS without deciding whether it belongs in DIRECTIONAL_UI_ICONS now fails this test immediately instead of silently shipping an icon nobody made a mirroring decision for.
 export const DirectionalMirroringTest: Story = {
   render: () => html`
     <div id="rtl-wrapper" dir="rtl">
       ${Object.keys(UI_ICONS).map(
         (name) => html`
           <swc-ui-icon
-            icon=${name}
+            icon=${name as UiIconName}
             data-icon=${name}
             accessible-label=${name}
           ></swc-ui-icon>
@@ -214,7 +210,7 @@ export const DirectionalMirroringTest: Story = {
       ${Object.keys(UI_ICONS).map(
         (name) => html`
           <swc-ui-icon
-            icon=${name}
+            icon=${name as UiIconName}
             data-icon=${name}
             accessible-label=${name}
           ></swc-ui-icon>
@@ -223,6 +219,8 @@ export const DirectionalMirroringTest: Story = {
     </div>
   `,
   play: async ({ canvasElement, step }) => {
+    await getComponent<UiIcon>(canvasElement, 'swc-ui-icon');
+
     const rtlWrapper = canvasElement.querySelector(
       '#rtl-wrapper'
     ) as HTMLElement;
@@ -231,7 +229,7 @@ export const DirectionalMirroringTest: Story = {
     ) as HTMLElement;
 
     await step(
-      'every icon mirrors under RTL iff it is in the directional set',
+      'every icon mirrors under RTL if it is in the directional set',
       async () => {
         for (const name of Object.keys(UI_ICONS) as UiIconName[]) {
           const icon = rtlWrapper.querySelector(`[data-icon="${name}"]`)!;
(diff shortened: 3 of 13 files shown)


miwha-adobe:
Nice work! Just wondering, do the public workflow icons need an equivalent? They live in the separate icons package and are named by direction `(chevron-left vs chevron-right as distinct icons)`, so auto-mirroring might not be wanted there the way it is for the semantic UI chevron. Totally fine if it's intentionally scoped to UI icons, just curious.

cdransf:
> Nice work! Just wondering, do the public workflow icons need an equivalent? They live in the separate icons package and are named by direction `(chevron-left vs chevron-right as distinct icons)`, so auto-mirroring might not be wanted there the way it is for the semantic UI chevron. Totally fine if it's intentionally scoped to UI icons, just curious.

Thanks! I don't think we need to build that since the semantics are baked into the name like you mentioned. UI icons are picked by meaning, so we can override there. ✨

miwha-adobe (approved):
lgtm!

2nd-gen/packages/swc/components/accordion/accordion-item.css
```diff
@@ -111,25 +111,22 @@
...
-:dir(rtl) .swc-AccordionItem-indicator {
-  scale: -1 1;
-}
-
 :host([open]) .swc-AccordionItem-indicator {
   rotate: 90deg;
 }
 
-/* RTL + open: rotate the opposite direction */
+/* RTL + open: swc-ui-icon has already mirrored the chevron horizontally, so rotating
```
Rajdeepc, line 127: Why accordion only? [MessagesSources](https://github.com/adobe/spectrum-web-components/blob/d941f7f9dade68a56f6d05608e1a36397dd71d87/2nd-gen/packages/swc/patterns/conversational-ai/message-sources/message-sources.css#L60) and [ResponseStatus](https://github.com/adobe/spectrum-web-components/blob/d941f7f9dade68a56f6d05608e1a36397dd71d87/2nd-gen/packages/swc/patterns/conversational-ai/response-status/response-status.css#L184) uses different patterns. This should also be fixed to align with the new scale behaviour.
Please verify this change in VRT also.

2nd-gen/packages/swc/components/ui-icons/test/ui-icons.test.ts
```diff
@@ -185,6 +185,137 @@ export const LabelTogglingTest: Story = {
...
+        getComputedStyle(icon).scale,
+        'chevron is horizontally mirrored in RTL'
+      ).toBe('-1 1');
+    });
+  },
+};
+
+// chevron and arrow are two independent :host() selector clauses in
+// ui-icon-direction.css, not one dynamically-generated rule, so each is
+// tested explicitly; a typo in one clause wouldn't be caught by testing
```
Rajdeepc: Yup this is the problem we are trying to solve. if a typo in one clause can only be caught by testing that clause specifically, the list needs to be data rather than hand-written stories.

Once `DIRECTIONAL_UI_ICONS` is exported (see my comment on `ui-icon-direction.css`), the five story tests collapse into one that can't drift:

```ts
export const DirectionalMirroringTest: Story = {
  render: () => html`
    <div id="direction-wrapper" dir="rtl">
      ${Object.keys(UI_ICONS).map(
        (name) => html`
          <swc-ui-icon
            icon=${name}
            data-icon=${name}
            accessible-label=${name}
          ></swc-ui-icon>
        `
      )}
    </div>
  `,
    await step('every icon mirrors iff it is in the directional set', async () => {
      for (const name of Object.keys(UI_ICONS) as UiIconName[]) {
        const icon = canvasElement.querySelector(`[data-icon="${name}"]`)!;
        const expected = DIRECTIONAL_UI_ICONS.has(name) ? '-1 1' : 'none';
        expect(getComputedStyle(icon).scale, `${name} in RTL`).toBe(expected);
      }
    });
  },
};
```


Two things this buys beyond covering every icon is adding an icon to `UI_ICONS` without deciding whether it's directional now fails loudly, and it kills `SecondDirectionalIconMirrorsInRtlTest` as a name.

You can keep `DirectionalIconMirrorsWithInheritedRtlTest` separately though. Inherited vs. direct `dir` is a genuinely different question from which icons are in the list, and your reasoning for testing it on its own is right.

2nd-gen/packages/swc/components/ui-icons/ui-icon-direction.css
```diff
@@ -0,0 +1,31 @@
...
+ * (:host(:dir(rtl)[icon="..."])). Chaining an attribute selector after a closing
+ * :host(:dir(...)) instead (:host(:dir(rtl))[icon="..."]) parses without error but
+ * never matches, verified against Chromium 138.0.7204.23 (this repo's pinned
+ * Playwright 1.53.1); keep new entries in this same form rather than "simplifying"
+ * back to the chained one.
+ */
+:host(:dir(rtl)[icon="chevron"]),
+:host(:dir(rtl)[icon="arrow"]) {
+  scale: -1 1;
+}
```
Rajdeepc, line 25: These two names are plain strings in CSS with nothing tying them back to `UI_ICONS`.

Could we make the list the source of truth in TypeScript and keep the CSS as the mechanism only?

In `ui-icons.types.ts` (a new export):

```ts
/**
 * Logical icons that mirror horizontally in RTL. Keep in sync with
 * `ui-icon-direction.css`; the test below fails if they drift.
 */
export const DIRECTIONAL_UI_ICONS = new Set<UiIconName>([
  'chevron',
  'arrow',
]) satisfies ReadonlySet<UiIconName>;
```

That gets you a compile error on a typo or a rename, and it gives the test something to iterate. It doesn't remove the CSS duplication on its own, but the test in my other comment closes that loop.

2nd-gen/packages/swc/components/ui-icons/ui-icon-direction.css
```diff
@@ -0,0 +1,31 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+/*
+ * A4U exposes no per-icon mirror flag, so directional icons are mirrored in RTL via
+ * this curated, hand-maintained list rather than per-icon metadata (icon-rfc.md,
+ * section 11). First pass: chevrons and arrows. Icons not listed here (checkmarks,
```
Rajdeepc: `corner-triangle` is also in the icon set

Rajdeepc (changes requested):

2nd-gen/packages/swc/components/accordion/accordion-item.css
cdransf: Updated in https://github.com/adobe/spectrum-web-components/pull/6579/commits/2455c799a188b1904408cc42bdac2a6b08c63e66 ✨

2nd-gen/packages/swc/components/ui-icons/test/ui-icons.test.ts
cdransf: Updated in https://github.com/adobe/spectrum-web-components/pull/6579/commits/d26850740fb6e5c7e7b8c555b8c6387ae2694710 ✨

2nd-gen/packages/swc/components/ui-icons/ui-icon-direction.css
cdransf: Updated in https://github.com/adobe/spectrum-web-components/pull/6579/commits/d26850740fb6e5c7e7b8c555b8c6387ae2694710 ✨

2nd-gen/packages/swc/components/ui-icons/ui-icon-direction.css
cdransf: Updated in https://github.com/adobe/spectrum-web-components/pull/6579/commits/d26850740fb6e5c7e7b8c555b8c6387ae2694710 ✨

2nd-gen/packages/swc/components/ui-icons/ui-icon-direction.css
```diff
@@ -0,0 +1,37 @@
...
+ * crosses, the corner-triangle hold-affordance marker, kept physically anchored to
+ * its corner in both directions to match 1st-gen, and similar non-directional or
```
Rajdeepc: `DirectionalMirroringTest` now asserts `corner-triangle` computes `scale: none` under RTL. The wrong behavior is locked in by a passing test. This is blocking as it would effect future developers.
This is wrong assertion. If design has decided that S2 should differ from 1st-gen this comment needs to say "S2 decision differs from 1st-gen". In first gen we flip the corner triangle in RTL.

2nd-gen/packages/swc/components/ui-icons/ui-icon-direction.ts
```diff
@@ -0,0 +1,24 @@
...
+ * Logical UI icons that mirror horizontally in RTL. Keep in sync with the
+ * `:host()` selector list in `ui-icon-direction.css`; `ui-icons.test.ts` iterates
+ * every entry in `UI_ICONS` against this set and fails on drift between the two.
+ * Icons not in this set (checkmarks, crosses, the corner-triangle hold-affordance
+ * marker, and other non-directional or physically-anchored art) never flip.
+ */
+export const DIRECTIONAL_UI_ICONS = new Set<UiIconName>([
+  'chevron',
+  'arrow',
+]) satisfies ReadonlySet<UiIconName>;
```
Rajdeepc: You should add `corner-triangle` to this set and to the selector list. And since you used `satisfies` this will only type-checks the expression, it does not narrow the type. Any importer can mutate the curated list process-wide. Please use annotation instead of `satisfies`
```ts
export const DIRECTIONAL_UI_ICONS: ReadonlySet<UiIconName> = new Set([
  'chevron',
  'arrow',
]);
```

2nd-gen/packages/swc/components/ui-icons/ui-icon-direction.css
```diff
@@ -0,0 +1,37 @@
...
+ * decision.
+ *
+ * The selector list below must stay in sync with `DIRECTIONAL_UI_ICONS` in
+ * ui-icon-direction.ts, the source of truth `ui-icons.test.ts` checks every
+ * `UI_ICONS` entry against.
+ *
+ * Both conditions must live inside :host()'s own argument as one compound selector
+ * (:host(:dir(rtl)[icon="..."])). Chaining an attribute selector after a closing
+ * :host(:dir(...)) instead (:host(:dir(rtl))[icon="..."]) parses without error but
+ * never matches, verified against Chromium 138.0.7204.23 (this repo's pinned
```
Rajdeepc: This is not browser specific issue. This is reading as a bug. Per CSS Scoping 1 the shadow host is *featureless*: it matches nothing except `:host`, `:host()`, and `:host-context()`. So the `[icon="x"]` in `:host(:dir(rtl))[icon="x"]` can never match it, in any engine, permanently.

2nd-gen/packages/swc/components/ui-icons/test/ui-icons.test.ts
```diff
@@ -185,6 +188,112 @@ export const LabelTogglingTest: Story = {
...
+      )}
+    </div>
+  `,
+  play: async ({ canvasElement, step }) => {
+    const rtlWrapper = canvasElement.querySelector(
+      '#rtl-wrapper'
+    ) as HTMLElement;
+    const ltrWrapper = canvasElement.querySelector(
+      '#ltr-wrapper'
+    ) as HTMLElement;
```
Rajdeepc, line 229: We should wait for the element to be defined and upgraded first not going straight to check for the elements. This can cause a race condition failure. 
Please add `await getComponent<UiIcon>(canvasElement, 'swc-ui-icon');` before the loops. It costs nothing and removes the ordering dependency.

2nd-gen/packages/swc/components/ui-icons/UiIcon.ts
```diff
@@ -18,14 +18,17 @@ import { UI_ICONS, UiIconName } from './icon-set/index.js';
...
 import uiIconSizeStyles from './ui-icon-sizes.css';
 
 /**
  * An internal icon renderer for Spectrum UI icons (chevrons, checkmarks, arrows, and
  * other control internals). The `icon` attribute selects the icon-set, and the
  * element renders the optically-tuned step that matches its `size`. Extends
- * `IconBase` for size and host-owned accessibility. Not published for consumers;
- * used by other swc components.
+ * `IconBase` for size and host-owned accessibility. A curated set of directional
+ * icons (chevrons, arrows) mirror automatically in RTL; consuming components do not
```
Rajdeepc:
```suggestion
 * icons (chevrons, arrows,corner-triangle) mirror automatically in RTL; consuming components do not
```

2nd-gen/packages/swc/patterns/ai-toolkit/message-sources/message-sources.css
```diff
@@ -56,11 +56,6 @@
   transition: transform token("animation-duration-100") ease;
 }
 
-/* swc-ui-icon mirrors the chevron itself in RTL via its own `scale`, which composes
- * after this `transform`. Rotating first lands on a down-pointing chevron, and
- * mirroring a down-pointing chevron left-to-right is a no-op (the shape is
- * symmetric on that axis), so no RTL-specific override is needed here, unlike
- * accordion's rotate-based indicator. */
```
Rajdeepc: @caseyisonit I am not sure why you deleted these comments but these explains why these two components need no RTL override while accordion needs. Same as in `response-status.css`

caseyisonit: I felt like the code was clear and the comment was redundant. Trying to reduce fatigue

Rajdeepc (approved):
Good work!

PR #6587, opened 2026-08-06
fix(action-button): show swc-avatar inside swc-action-button

rubencarvalho:
## Description

Fixes `<swc-avatar>` so it automatically scales to match `<swc-action-button>`'s icon size when slotted into its `icon` slot. Also updates Avatar's "In Action button" story to demonstrate the fix, including a new `xl` action-button example.

## Motivation and context

This started as a docs-only change: replacing the placeholder `<button>` (with a `TODO: Replace <button> with <swc-action-button> once that component is migrated to 2nd-gen` comment) in Avatar's "In Action button" story with the real `<swc-action-button>`.

That surfaced a real bug. Putting `<swc-avatar>` in ActionButton's `slot="icon"` didn't work: Avatar's own `size` attribute controlled its rendered size regardless of the action button's size, so an avatar sized for one button size looked wrong (overflowed or clipped) in another. Avatar's size scale (16–104px) also doesn't line up with ActionButton's icon-size tokens, so there was no `size` value that coincidentally fixed it — consumers would have been on the hook for manually pairing an avatar `size` with the action-button `size` (e.g. `xl` action button ↔ avatar `size="1000"`).

**The fix**: ActionButton's icon slot now also sets `--swc-avatar-size` directly, the same way it already sets its own internal `--_swc-action-button-icon-size` custom properties for the slotted element. This wins over Avatar's own `:host([size])` rule for a slotted avatar. Confirmed empirically via computed styles: the same `<swc-avatar size="500">` now resolves `--swc-avatar-size` to `18px` inside an `m` action button and `22px` inside an `xl` action button, matching each button's own icon-size token exactly, with **no changes to Avatar's own CSS architecture** and no size-pairing required from consumers. Avatar's `size` attribute simply becomes a no-op in this context (documented in `avatar.mdx`).

The story now shows four examples side by side, none of which write a `size` attribute on the avatar at all (it would be misleading to suggest it matters, since it's a no-op here):
- `m` action button, avatar + visible label (avatar `slot="icon"`, `decorative`)
- `m` action button, avatar-only (icon-only), using `accessible-label` for the accessible name
- `xl` action button, avatar + visible label
- `xl` action button, avatar-only

(Earlier revisions of this PR had all four avatars share an explicit `size="500"` to demonstrate the fix; that was removed once confirmed, since authoring a no-op attribute in the example would suggest it does something.)

The avatar is marked `decorative` (`alt=""`) in all four since ActionButton doesn't automatically mark icon-slot content `aria-hidden`, and the avatar's own `alt` text would otherwise duplicate the button's visible/accessible name — matching Avatar's own documented guidance for `decorative`.

## Related issue(s)

- fixes N/A (found via story/docs follow-up)

## Author's checklist

Existing test suites pass unchanged: avatar (14/14), action-button (19/19), including custom-property/global-style VRT coverage checks (37/37 total across `components/avatar` and `components/action-button`).

## Reviewer's checklist

### Manual review test cases

- [ ] _Avatar scales to match action-button icon size, no manual size pairing needed_
  1. Open Storybook, `Avatar > Docs`, scroll to **Behaviors > In Action button**
  2. Expect four pill-shaped action buttons (two `m`, two `xl`), each avatar rendered as a full circle with no clipping/overflow, correctly sized to each button's own icon-size token — even though every avatar uses the same `size="500"` attribute
  3. Expect each avatar vertically and horizontally centered relative to its button

- [ ] _Standalone avatar sizing is unaffected_
  1. Check `Avatar > Docs > Options > Sizes` and `Outline` stories
  2. Expect all 17 sizes and both outline examples to render identically to before this change

- [ ] _Icon-only avatar action buttons have a correct accessible name_
  1. Inspect the icon-only (avatar-only) buttons in the "In Action button" story
  2. Expect the internal `<button>` to carry `aria-label="Jane Doe"` forwarded from the host's `accessible-label` attribute

### Device review

- [ ] Did it pass in Desktop?

  1. Tab through all four action buttons in the "In Action button" story.
  2. Confirm each receives visible focus and is individually reachable in tab order.
  3. Expect no change in focus behavior versus any other `<swc-action-button>` usage — the avatar itself is non-interactive and not part of the tab order.

  1. With VoiceOver/NVDA running, focus each of the four buttons in turn.
  2. Expect each to announce once as "Jane Doe, button" — not twice — since the avatar is `decorative`/`alt=""` in every case.
  3. For the two labeled buttons, the name comes from the visible text; for the two icon-only buttons, it comes from `accessible-label`. Both should sound identical to a screen reader user.


2nd-gen/packages/swc/components/avatar/stories/avatar.stories.ts
```diff
@@ -231,26 +232,48 @@ export const Disabled: Story = {
...
   render: (args) => html`
-    <button
-      type="button"
-      style="display:inline-flex;align-items:center;gap:8px;padding:4px 12px;cursor:pointer;"
-    >
+    <swc-action-button>
+      <span
+        style="display: inline-flex; align-items: center; gap: 8px; vertical-align: middle;"
+      >
+        <swc-avatar
```
cdransf: Should these be placed in the icon slot instead? ✨

cdransf (approved):

2nd-gen/packages/swc/stylesheets/global/global-action-button.css
```diff
@@ -105,6 +105,7 @@
 .swc-ActionButton-icon {
   color: inherit;
   fill: currentcolor;
+  --swc-avatar-size: var(--_swc-action-button-icon-size);
```
Rajdeepc, line 108: Did this come from the generator or hand generated?

.changeset/fix-action-button-avatar-icon-size.md
```diff
@@ -0,0 +1,9 @@
+---
+'@adobe/spectrum-wc': patch
+---
+
+**fix(action-button):** A `<swc-avatar>` slotted into `<swc-action-button>`'s `icon` slot now automatically scales to match the button's icon size.
+
+Previously, Avatar's own `size` attribute controlled its rendered size regardless of the action button's size, since Avatar's `:host([size])` sizing rule won the specificity contest with the icon slot's generic sizing rule for the avatar's _host box_ only — the visible image inside Avatar's shadow root ignored that squeeze entirely and rendered at its own `size`, causing it to overflow. Consumers previously had to manually pair an avatar `size` with the action-button `size` (e.g. `xl` action button with avatar `size="1000"`) to avoid a visual mismatch.
```
Rajdeepc, line 7: This needs a VRT run.

rubencarvalho: Great call! I even added this to the vrt permutations :)

2nd-gen/packages/swc/stylesheets/global/global-action-button.css
rubencarvalho: global-component styles are auto-generated!

2nd-gen/packages/swc/components/action-button/action-button.css
```diff
@@ -114,6 +114,10 @@ slot[name="icon"]::slotted(*),
 slot[name="icon"]::slotted(*) {
   color: inherit;
   fill: currentcolor;
+
+  /* Lets a slotted <swc-avatar> track the icon slot's own size instead of
+     requiring consumers to pair its size attribute with the button's size. */
+  --swc-avatar-size: var(--_swc-action-button-icon-size);
```
5t3ph, line 122: Perhaps make this more explicitly scoped to `::slotted(swc-avatar)`

.changeset/fix-action-button-avatar-icon-size.md
rubencarvalho: I'm glad that I did because I found another bug!!!!

Lit still leaves its child-part boundary marker, a comment node `<!--?lit$…$-->` with non-empty data as a direct light-DOM child of `<swc-action-button>`. With no slot attribute, it lands in the default slot.

Then in `slot-text-controller.ts`, `checkInitialContent()` classifies nodes as element-or-text:
```ts
if (el.tagName) { … }              // elements
if (this.slotName) return false;   // named slots
return node.textContent ? node.textContent.trim().length > 0 : false;  // ← comment falls here
```

A comment node has no `tagName`, so it hits the last branch, and `node.textContent` returns the comment's data (?lit$…$) → trims to non-empty → hasLabel = true → iconOnly = false.

5t3ph (approved):
Looks good, nice updates to VRTs.

One nit on placement of the avatar size cp.

Rajdeepc (approved):

2nd-gen/packages/swc/components/action-button/action-button.css
rubencarvalho: applied the recommendation!

PR #6588, opened 2026-08-06
chore(avatar): add vrt support

cdransf:
## Description

Adds `avatar.vrt.ts` and `avatar-custom-properties.vrt.ts` under `avatar/test/vrt/`: sizes, outline, disabled, forced-colors, and all 4 custom properties.

## Motivation and context

Extends the VRT foundation (button/link/divider) to Avatar.

## Related issue(s)

- fixes SWC-2417

## Author's checklist

- [ ] **Keyboard**: N/A — Avatar is non-interactive and not focusable; no keyboard behavior to test.
- [ ] **Screen reader**: 1. Open Storybook → Avatar → Avatar VRT. 2. Confirm no console errors/warnings on load. 3. Spot-check one avatar per row with a screen reader; expect `alt` text announced, no extra noise.

### DIFF
--- a/2nd-gen/packages/swc/components/avatar/test/vrt/avatar-custom-properties.vrt.ts
+++ b/2nd-gen/packages/swc/components/avatar/test/vrt/avatar-custom-properties.vrt.ts
@@ -45,12 +45,22 @@ const PLACEHOLDER_SRC = 'https://picsum.photos/id/64/500/500';
 type AvatarPropertyCase = CustomPropertyCase<`--swc-avatar-${string}`> & {
   outline?: boolean;
   disabled?: boolean;
+  // Fixed style applied to both the default and modified renders. Used by
+  // --swc-avatar-outline-width to pin a visible outline color, since the
+  // default token color is too pale against the light background to reveal
+  // a width change on its own.
+  baseStyle?: string;
 };
 
 const AVATAR_PROPERTY_CASES: readonly AvatarPropertyCase[] = [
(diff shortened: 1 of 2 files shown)


rubencarvalho:
I think this outline needs color too, so we are able to see its width 😃 
Screenshot 2026-08-07 at 11 11 02

cdransf:
> I think this outline needs color too, so we are able to see its width 😃

Ah! Good catch. Updated.

5t3ph (approved):
These look good!

Can you add a ticket to check with Design if the outline color should actually flip for light/dark mode or if it should use a static value? The intent of outline is to make the boundary more clear and if the background color/asset doesn't actually change, then the flip to a dark outline may not be meeting the intent.

Image

pfulton:
> These look good!
> 
> Can you add a ticket to check with Design if the outline color should actually flip for light/dark mode or if it should use a static value? The intent of outline is to make the boundary more clear and if the background color/asset doesn't actually change, then the flip to a dark outline may not be meeting the intent.
> Image

+1 yes please to this request.

pfulton (approved):

cdransf:
> These look good!
> 
> Can you add a ticket to check with Design if the outline color should actually flip for light/dark mode or if it should use a static value? The intent of outline is to make the boundary more clear and if the background color/asset doesn't actually change, then the flip to a dark outline may not be meeting the intent.
> Image

Added as SWC-2483 cc @pfulton

PR #6592, opened 2026-08-07
fix(divider): fix zero-height vertical divider in flex/grid containers

rubencarvalho:
## Description

`:host([vertical])` set `block-size: 100%`, which disables flexbox's `align-self: stretch` fallback and resolves to `0` when the parent's height is indefinite (e.g. an auto-height flex toolbar). Replaced with explicit `align-self: stretch`, which works in flex rows and `grid-auto-flow: column` grids. Also removed the inert `align-self: flex-start` on the inner shadow div (not a flex/grid item, had no effect).

Follow-up commit updates the divider docs examples and VRTs to drop the explicit `block-size` workaround that vertical divider examples previously needed, validating that `align-self: stretch` renders correctly on its own in a flex row without a manually set height.

## Motivation and context

A consumer reported `swc-divider[vertical]` rendering at 0 height inside an `inline-flex` toolbar.

- [x] **Keyboard**: no focusable parts; no keyboard behavior affected by this change.
- [x] **Screen reader**: no role/ARIA changes; `role="separator"` and `aria-orientation` are unaffected.

## Author's checklist

- [x] I have read the CONTRIBUTING and PULL_REQUESTS documents.


2nd-gen/packages/swc/components/divider/divider.css
```diff
@@ -42,11 +42,10 @@
 }
 
 :host([vertical]) {
-  block-size: 100%;
+  align-self: stretch;
```
Rajdeepc, line 45: This will only have effect on flex items and grid items. Other layout context will ignore this. `block-size` which you removed will work on any context. 
This will render a 0 height invisible divider. 
Can we keep `block-size: 100%` as a fallback alongside `align-self: stretch` e.g. `block-size: 100%; align-self: stretch;` while the latter simply wins when the host is a flex/grid item, and the percentage still resolves for plain block parents with definite height.

Rajdeepc (commented):
- If the direction is that the parent must also be a flex or a grid container this needs to be updated in `divider.mdx` which is missing from this PR. Kindly add if its not added.
- Would appreciate if you can add a test for this change too.

2nd-gen/packages/swc/components/divider/divider.css
5t3ph: In practice this unfortunately doesn't work and also requires an explicit height on the parent (our previous solution shown in docs). However, the `:host` style is `display: block` which means without a flex or grid context it's visually broken anyway vs. expectation. So either way, the consumer has to own the layout mode for the expected orientation.

5t3ph (changes requested):
Let's update our docs examples and VRTs to remove the previous explicit height to help validate this update.

rubencarvalho:
> Let's update our docs examples and VRTs to remove the previous explicit height to help validate this update.

@5t3ph, I updated it, do you think that guidance is enough?

5t3ph (approved):
Guidance is fine, VRTs look good too.

cdransf (approved):

PR #6601, opened 2026-08-11
ci(release): split gen1/gen2 release pipelines with a Version PR flow

blunteshwar:
Closed — retargeting to the `release-strategy` base branch instead of `main`. See the new PR (same head branch, same commits) for the current state of this work.


.github/workflows/publish.yml
```diff
@@ -292,14 +216,136 @@ jobs:
...
+      - name: Build all packages
+        run: yarn build
+
+      - name: Generate custom elements manifests
+        run: yarn workspace @spectrum-web-components/1st-gen custom-element-json
+
+      - name: Confirm build artifacts
+        run: yarn workspace @spectrum-web-components/1st-gen build:confirm
+
+      - name: Version 1st-gen packages
```
cdransf, line 291: Do we need `env: GITHUB_TOKEN` here? ✨

.github/workflows/publish-2nd-gen.yml
```diff
@@ -0,0 +1,243 @@
...
+      - name: Verify NPM authentication
+        id: npm-auth
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.ADOBE_BOT_NPM_TOKEN }}
+        run: npm whoami --registry=https://registry.npmjs.org
+
+      - name: Build all packages
+        run: yarn build
+
+      - name: Version 2nd-gen packages
```
cdransf, line 190: Do we need `env: GITHUB_TOKEN` here? ✨

.github/workflows/publish.yml
cdransf: Is it an issue if this and the 2nd gen workflow use the same `github.ref`? ✨

.github/workflows/publish.yml
```diff
@@ -21,57 +21,45 @@ jobs:
...
-            if [ "$HAS_1ST_GEN" = "true" ]; then
-              echo "Found 1st-gen changesets - React wrappers will be built and published"
-            else
-              echo "No 1st-gen changesets found - React wrappers will be skipped"
-            fi
+            echo "Found $GEN1_COUNT 1st-gen changeset(s)"
           fi
 
-  publish:
+  # Throwaway snapshot publish: every push to main (continuous `next`), manual
```
cdransf: We should revise this as it's not every push if we have a condition checking `has_changesets`. ✨

.github/workflows/publish.yml
```diff
@@ -1,10 +1,10 @@
...
+name: Publish Packages (1st-gen)
 
 on:
   workflow_dispatch:
     inputs:
       tag:
-        description: 'NPM dist-tag (e.g., latest, beta, snapshot)'
+        description: 'NPM dist-tag for a throwaway snapshot (e.g., next, snapshot-test)'
         required: false
         default: 'next'
```
cdransf, line 9:
```suggestion
        required: false
        default: 'next'
        type: choice
        options:
          - next
          - snapshot-test
```

Can we scope this to a fixed set of options to avoid using `latest` here? ✨

.github/workflows/publish.yml
```diff
@@ -292,14 +216,136 @@ jobs:
...
-          git pull --rebase origin main
-          git push
-
-      - name: Create git tag
-        if: steps.extract-tag.outputs.tag == 'latest'
-        run: node ./1st-gen/scripts/create-git-tag.js
+          yarn workspace @spectrum-web-components/1st-gen changelog:global
+          (cd 1st-gen && yarn changeset version)
+
+      - name: Open or update the 1st-gen Version PR
```
cdransf, line 298: Do we need to refresh the lockfile before this step? ✨

.github/workflows/publish-2nd-gen.yml
```diff
@@ -0,0 +1,243 @@
...
+        if: needs.check-changesets.outputs.has_changesets == 'true'
+        run: |
+          if [ ! -f 2nd-gen/.changeset/pre.json ]; then
+            (cd 2nd-gen && yarn changeset pre enter beta)
+          fi
+          (cd 2nd-gen && yarn changeset version)
+
+      - name: Open or update the 2nd-gen Version PR
```
cdransf, line 199: Do we need to refresh the lockfile before this step? ✨

cdransf (changes requested):

.github/workflows/publish.yml
```diff
@@ -292,14 +230,136 @@ jobs:
...
+
+      - name: Confirm build artifacts
+        run: yarn workspace @spectrum-web-components/1st-gen build:confirm
+
+      - name: Version 1st-gen packages
+        if: needs.check-changesets.outputs.has_changesets == 'true'
+        run: |
+          yarn workspace @spectrum-web-components/1st-gen changelog:global
+          (cd 1st-gen && yarn changeset version)
+          yarn install --refresh-lockfile
```
blunteshwar, line 296: @cdransf Good catch — fixed in e3bcc08162. Added `yarn install --refresh-lockfile` right after `yarn changeset version`, before the Version PR step, so the PR's `yarn.lock` diff actually reflects the bumped internal dependency pins instead of going stale. Applied the same fix to publish-2nd-gen.yml.

blunteshwar (commented):
Addressed the lockfile-refresh and workflow_dispatch-input feedback — see inline replies. (Also pushed fixes for the shared-concurrency-group and redundant-token points from the same review round; those didn't have a specific line to reply to.)

.github/workflows/publish-2nd-gen.yml
```diff
@@ -0,0 +1,249 @@
...
+        run: yarn build
+
+      - name: Version 2nd-gen packages
+        if: needs.check-changesets.outputs.has_changesets == 'true'
+        run: |
+          if [ ! -f 2nd-gen/.changeset/pre.json ]; then
+            (cd 2nd-gen && yarn changeset pre enter beta)
+          fi
+          (cd 2nd-gen && yarn changeset version)
+          yarn install --refresh-lockfile
```
blunteshwar, line 197: @cdransf Same fix as publish.yml (e3bcc08162) — `yarn install --refresh-lockfile` added right after `yarn changeset version`, before the Version PR step.

.github/workflows/publish.yml
```diff
@@ -1,77 +1,79 @@
...
 
 on:
   workflow_dispatch:
     inputs:
       tag:
-        description: 'NPM dist-tag (e.g., latest, beta, snapshot)'
+        description: 'NPM dist-tag for a throwaway snapshot'
         required: false
         default: 'next'
+        type: choice
```
blunteshwar, line 10: @cdransf Done in e3bcc08162 — `tag` is now `type: choice` with only `next`/`snapshot-test` as options (applied to both publish.yml and publish-2nd-gen.yml). Also closes a real gap: without this, a manual dispatch with `tag: latest` could publish to the real `latest` dist-tag outside the reviewed Version PR flow.

PR #6605, opened 2026-08-11
feat(pixel-loader): productionize swc-pixel-loader

aramos-adobe:
## Reviewer context: productionization summary

This PR stems from`jnjosh/pixel-loader` (the prototype) and hardens `swc-pixel-loader` for production. Changes on top of that base:

- **Removed prototype framing** across the component JSDoc, `data.ts`, `animation.ts`, `geometry.ts`, the stories meta, and the MDX (no more "prototype shell" language or internal design-spec section references).
- **`size` is no longer a public attribute.** Sizing is consumer-controlled via the `--swc-pixel-loader-size` CSS custom property (default 56px). Design has no size model yet; a typed `s`/`m`/`l` enum can be layered on non-breakingly later on top of the same property.
- **Fixed a paused / reduced-motion bug:** the preset ticker kept cycling icons while `paused` or under reduced motion. It now freezes on one icon (`_isStatic` gate + re-sync on `paused`).
- **Reduced-motion is now live:** a `matchMedia` change listener re-evaluates when the OS setting toggles while the loader is mounted.
- **Documented both CSS custom properties** (`--swc-pixel-loader-size`, `--swc-pixel-loader-color`) via `@cssprop` so they surface in the API table, plus a Sizing note in the MDX.
- **Accessibility:** replaced the placeholder a11y spec with a real `toMatchAriaSnapshot` for the indeterminate progressbar.
- **Tests:** added unit coverage for defaults, `icon` reflection, `preset` overriding `icon`, the progressbar/`aria-label` wiring, the paused static render (ticker regression guard), and the reduced-motion live toggle. 5/5 passing.

Still open (not in this PR): dedicated behavior/preset stories, and design sign-off on the final icon set / preset names / eventual size scale.

---

## Description

Productionizes the prototype `swc-pixel-loader` from the `jnjosh/pixel-loader` base branch. This PR stacks on that branch and covers the engineering hardening needed before the Prompt field generating-state branding can depend on the loader.

Changes:

- **Removed all "prototype" framing.** Reworded the component JSDoc, `data.ts`, `animation.ts`, `geometry.ts`, the stories meta, and the per-unit MDX so nothing reads as a "prototype shell" or cites internal design-spec section numbers a reader cannot access.
- **Made `size` non-public.** Dropped the freeform `size="56px"` attribute. Sizing is now driven entirely by the `--swc-pixel-loader-size` custom property, which consumers override in their own CSS (for example, response-status will set it in a later project). Design does not yet have a size model, so a typed `s`/`m`/`l` enum would mean inventing values; keeping `size` off the public attribute surface avoids shipping an attribute that a future `size` enum would collide with. The enum can be layered on non-breakingly later, on top of the same custom property.
- **Fixed the paused / reduced-motion ticker bug.** The preset ticker (`setInterval`) previously kept running while `paused` or under `prefers-reduced-motion`, so a "frozen" preset still swapped icons on a timer. Added a single `_isStatic` source of truth, gated `_syncTicker()` on it, and re-synced the ticker when `paused` changes. A paused or reduced-motion loader now holds one icon with no running animations.
- **Replaced the placeholder a11y spec** with a real `toMatchAriaSnapshot` assertion for the indeterminate `progressbar` and its accessible name.
- **Added unit tests** covering defaults, `icon` reflection, `preset` overriding `icon`, the progressbar/`aria-label` wiring, and the paused static render (including a regression guard that a paused preset runs no animations). All four test stories pass.

## Motivation and context

The pixel loader shipped on `jnjosh/pixel-loader` as an explicitly-marked prototype shell. It needs to be productionized before the Prompt field "generating" branding can consume it as the leading AI icon. This is part of the Conversational AI release and is tracked as a P0.

### References

[RSP Code](https://github.com/adobe/react-spectrum/blob/main/packages/%40react-spectrum/ai/src/loader/react.tsx)
[Demo](https://reactspectrum.blob.core.windows.net/reactspectrum/1062d083ad8cd43df2ae0e72bba67413cd6801cb/storybook-s2/index.html?path=/story/ai-promptfield--everything&args=pixelLoader:hourglass;isGenerating:!true;variant:balanced)

## Related issue(s)

- Productionizes the prototype from the `jnjosh/pixel-loader` branch.
- fixes SWC-2477

https://github.com/user-attachments/assets/6fecfa2a-5803-45d0-abab-3639b387dfff

Screenshot 2026-08-14 at 6 02 41 PM

---

## Author's checklist

- [ ] I have included a well-written changeset if my change needs to be published. _(Deferred: this PR targets the `jnjosh/pixel-loader` feature branch, not `main`; a changeset will be added when the stack targets `main`.)_

---

## Reviewer's checklist

### Manual review test cases

- [ ] _Icons and presets render and animate_
  1. In Storybook, open **Conversational AI / Pixel loader**.
  2. In the Playground, cycle the `icon` control through several values and set `preset` to `cc`, `dc`, `exp`, `analyze`, and `mega`.
  3. Expect each icon/preset to assemble, hold, and disassemble; presets cycle one icon per loop.

- [ ] _Paused and reduced-motion render statically_
  1. Open the **States** story (or set `paused` in the Playground).
  2. Expect the paused loader to hold the fully-settled icon with no animation.
  3. With a `preset` set and `paused` on, confirm the icon does not change over time (the ticker is suppressed).
  4. Enable OS "reduce motion" and reload; expect the loader to render the settled icon with no animation.

- [ ] _Size override via CSS custom property_
  1. Wrap a loader with a style such as `swc-pixel-loader { --swc-pixel-loader-size: 24px; }`.
  2. Expect the loader to render at the overridden size (there is no `size` attribute).

### Device review

- [ ] Did it pass in Desktop?

  1. In Storybook, open **Conversational AI / Pixel loader** and <kbd>Tab</kbd> through the page.
  2. Confirm the loader itself is not a tab stop and introduces no focus trap.
  3. Expect focus order in surrounding examples to be unaffected (no regressions).

  1. With VoiceOver (or NVDA) active, navigate to the loader.
  2. Expect it to be announced as a progress indicator using its `label` as the accessible name (default "Loading"; set a specific label such as "Generating response").
  3. Confirm no `aria-valuenow`/percentage is announced (progress is indeterminate) and there are no duplicate or noisy announcements from the animation.


2nd-gen/packages/swc/patterns/conversational-ai/pixel-loader/animation.ts
```diff
@@ -17,17 +17,17 @@ import { TOTAL_FRAMES } from './data.js';
  * Per-cell entry/exit choreography, driven via the Web Animations API rather
```
5t3ph: The animation of all but the `aiLogo` is entering/existing incorrectly.

Expectation: the icon pixels enter from the top and build bottom-up, and then exit from the bottom of the frame and from icon bottom-up. Icon entry/exit besides `aiLogo` is reversed. This should hold steady whether there is one icon selected or it is cycling multiple.

2nd-gen/packages/swc/patterns/conversational-ai/pixel-loader/PixelLoader.ts
```diff
@@ -84,28 +82,62 @@ export class PixelLoader extends SpectrumElement {
...
+  };
+
   protected override updated(changed: PropertyValues<this>): void {
     super.updated(changed);
 
     if (changed.has('preset')) {
       this._presetIndex = 0;
+    }
+
+    // Resync on `paused` too: pausing (or reduced motion) must stop the preset
```
5t3ph: I notice that toggling pause leaves artifacts from the previous render like it's still holding in memory the uncompleted cycle and isn't picking back up correctly.

https://github.com/user-attachments/assets/962fd7f1-f023-4947-9118-1da9a28ca6a3

2nd-gen/packages/swc/patterns/conversational-ai/pixel-loader/PixelLoader.ts
```diff
@@ -120,12 +152,25 @@ export class PixelLoader extends SpectrumElement {
...
     );
   }
 
+  /**
+   * Whether the loader should render its frozen, non-animating appearance:
+   * when explicitly `paused`, or when the user has requested reduced motion.
+   * Both suppress the cell animation and the preset ticker.
+   */
+  private get _isStatic(): boolean {
+    return this.paused || this._prefersReducedMotion();
```
5t3ph: I would love to see this behavior refined so that the icon can complete its current build if it's midway, maybe just by fading in the rest of the pixels to still soften the transition but prevent just a snap between states?

5t3ph: For prefers reduced motion, this has a similar purpose as something like progress circle where the animation is actually communicating an important state of loading.

I would suggest not static but perhaps just a slowed (aka "reduced") animation speed? Potentially also altered to load like a row of pixels at a time to reduce the total duration needed? Could also experiment with fading in/out row by row vs. falling, which would also help meet the intent of "reducing motion" while again preserving the meaning of "something is happening/loading".

5t3ph (changes requested):
Great work on this!

Two bugs, and some suggestions for further polish.

2nd-gen/packages/swc/patterns/conversational-ai/pixel-loader/animation.ts
5t3ph: Closer now, but see this slowed down version from React and see how the pixels also cascade in from the top on entrance, which is missing now. They also should fall all the way down and through on exit, and they seem to only be falling a set distance and then fading out in the current iteration.

https://github.com/user-attachments/assets/0c9fdae8-e92c-492a-b3b1-cb02fef2ea44

2nd-gen/packages/swc/patterns/conversational-ai/pixel-loader/PixelLoader.ts
5t3ph: At least in Storybook, I can't get the paused toggle to even work unless I hit it at the short pause between animation cycles. Can you check if that's a storybook quirk or a real bug where something is preventing pause from being set if the animation is midway?

2nd-gen/packages/swc/patterns/conversational-ai/pixel-loader/PixelLoader.ts
5t3ph: Good improvement here, I like your solution! Maybe also slow it down a bit, try and see what seems to be slower but still smooth without giving the appearance of lagging.

2nd-gen/packages/swc/patterns/conversational-ai/pixel-loader/PixelLoader.ts
5t3ph: I can use devtools to apply the `paused` attribute and it seems to be working well, so just the Storybook toggle is kind of buggy. You could make a separate ticket to look into it, it doesn't need to hold up this PR.

2nd-gen/packages/swc/patterns/conversational-ai/pixel-loader/PixelLoader.ts
5t3ph: The latest update reverted the changes here and lost any motion.

2nd-gen/packages/swc/patterns/conversational-ai/pixel-loader/stories/pixel-loader.stories.ts
5t3ph, line 119: Maybe show only one representative one here with a note to say go to the Playground to preview other options, as this is pretty overwhelming to see altogether, especially when different icons are on different duration timelines.

5t3ph, line 111: Can these be static for purposes of just showing some options here? You can say to go to the Playground to see them in motion.

5t3ph (changes requested):
Latest animation looks nice!

See updated comment reply re: reduced motion.

Also, the icons are squared off again, but the demo you linked still shows the rounded appearance.

2nd-gen/packages/swc/patterns/conversational-ai/pixel-loader/PixelLoader.ts
aramos-adobe: @5t3ph RSP freezes it entirely. It was a design decision

aramos-adobe: I could bring it back. I like the way it moved in reduced motion then it freezing all the way.

5t3ph: As I noted initially, it's actually communicating so to my understanding it's best to not take it away entirely. I suggest pushing back on it, maybe showing a video of your slowed animation.

5t3ph (approved):
Nice work on this! :shipit:

2nd-gen/packages/swc/patterns/conversational-ai/pixel-loader/animation.ts
```diff
@@ -0,0 +1,260 @@
...
+ * two Web Animations (a `translate` drop and a `scale` pop), and the whole grid
+ * runs one `opacity` envelope. `translate` and `scale` are separate animations
+ * because each has its own per-segment easing, which a single merged keyframe
+ * list could not express (a keyframe's easing applies to every property in it).
+ */
+
+import type { Cell } from './data.js';
+
+const FPS = 30;
+const ms2f = (ms: number): number => (ms * FPS) / 1000;
```
miwha-adobe: Nit: This variable name was not super clear to me. Maybe something like `msToFrames`

miwha-adobe (changes requested):
Can we include dev warnings + tests for the icon / preset fallback? Right now an invalid value silently renders aiLogo / drops the preset with no warning. We have shared helpers in core/utils/dev-validation.ts (validateEnum / warnIf) that should make this pretty straightforward.

aramos-adobe:
@miwha-adobe How's this?

`preset` is optional. If someone enters an invalid preset name it'll just render the `aiLogo` as a single icon and no preset

Screenshot 2026-08-17 at 5 16 04 PM
Screenshot 2026-08-17 at 5 15 18 PM

2nd-gen/packages/swc/patterns/conversational-ai/pixel-loader/PixelLoader.ts
```diff
@@ -0,0 +1,559 @@
...
+
+  @state()
+  private _presetIndex = 0;
+
+  private _animations: Animation[] = [];
+
+  // Cached shadow-root lookups. The container and cell elements are stable
+  // between renders, so query them once and reuse across `_playCells` /
+  // `_finishThenSwap` calls (including the render-less reduced-motion and pause
+  // paths). Invalidated in `updated` since a new icon/preset re-renders cells.
```
rubencarvalho, line 120: i could confirm there is a small bug here: on reconnect this only re-arms the ticker, but `disconnectedCallback` cancelled the cell animations via `_cancelAnimations()`. Since no property changes, Lit doesn't re-render and `_playCells()` never runs again. if we have a single-icon loader (no preset) that gets moved in the DOM (e.g. if we use something like appendChild, tab switch, maybe virtualization) it comes back frozen (i created a reproducible and confirmed it). if we add a `this._playCells()` here to mirror `_cancelAnimations()` on disconnect, it should fix it.

2nd-gen/packages/swc/patterns/conversational-ai/pixel-loader/PixelLoader.ts
```diff
@@ -0,0 +1,559 @@
...
+
+  /**
+   * Icon currently on screen. Decoupled from the public `icon` so a mid-build
+   * icon change can finish the current build before swapping (see
+   * `_finishThenSwap`).
+   */
+  @state()
+  private _displayedIcon: PixelLoaderIconName = 'aiLogo';
+
+  /** Guards stale finish-then-swap completions against a newer change. */
```
rubencarvalho, line 136: nit: this `_syncTicker()` runs on mount, then the first `updated()` sees `preset` in changedProperties and immediately clears + recreates the interval. So we tear down/rebuild the timer twice on every mount.

2nd-gen/packages/swc/patterns/conversational-ai/pixel-loader/PixelLoader.ts
```diff
@@ -0,0 +1,559 @@
...
+   * `PixelLoaderPresetName` (e.g. an unset Storybook control arg serialized
+   * as the literal string `"undefined"` rather than the attribute being
+   * absent).
+   */
+  private get _resolvedPreset(): PixelLoaderPresetName | undefined {
+    return this.preset && this._isValidPreset(this.preset)
+      ? this.preset
+      : undefined;
+  }
+
```
rubencarvalho, line 330: hmmm, is this JSDoc here correct? shouldn't it be on `_finishThenSwap()` (below?)

rubencarvalho (commented):
Great work! I think it would make sense making this internal, though. We don't expect (nor want!) consumers to reach out to the pixel loader directly. We will compose it in our Prompt field and eventually Response status.

rubencarvalho (approved):
🚢

2nd-gen/packages/swc/patterns/conversational-ai/pixel-loader/PixelLoader.ts
```diff
@@ -0,0 +1,570 @@
...
+      changed.has('_displayedIcon') ||
+      changed.has('preset') ||
+      changed.has('_presetIndex')
+    ) {
+      this._containerCache = null;
+      this._cellElsCache = null;
+    }
+
+    if (changed.has('preset')) {
+      this._presetIndex = 0;
```
rubencarvalho: a bit of an edge/corner case but this reset runs in `updated()`, i.e. *after* `render()` has already drawn cells using the old `_presetIndex`. If you swap from one preset to another after it's auto-cycled (say preset A is at `_presetIndex=3`, then `preset='B'` is set), `render()` draws `B[3]`'s cells, then this resets to 0 and the synchronous `_playCells()` below queries the `B[3]` DOM cells but reads `B[0]`'s cell list. in practice, it's just that the wrong glyph animates for a frame, and if the counts differ the surplus cells flash opaque.
tl;dr is that moving this reset up into `willUpdate` (before render) should fix it

2nd-gen/packages/swc/patterns/conversational-ai/pixel-loader/PixelLoader.ts
```diff
@@ -0,0 +1,570 @@
...
+
+    const cellEls = this._cellEls();
+    const container = this._container();
+    const { cells, isPreset } = this._activeCells;
+    const mode = this._animationMode;
+
+    // Clear any inline values a previous render left behind so a stale value
+    // never becomes the Web Animations implicit offset-0 keyframe on the next
+    // play (the "artifact from the previous cycle" when toggling `paused`).
+    cellEls.forEach((cellEl) => {
```
rubencarvalho: related to the `_presetIndex` reset above: this loop strips the hidden inline state (translate/scale/opacity) off *every* DOM cell, but further down `if (!cells[index]) return` skips animating cells that have no entry in the active `cells` list. So whenever `cellEls.length > cells.length`, those extra cells are left at the CSS defaults (opacity 1, scale 1) and render as stray solid pixels. Only reachable via the preset-swap transient today, but it's a latent trap — worth either not stripping cells you won't re-animate, or hiding the unused ones explicitly. Fixing the reset above closes the current path.

2nd-gen/packages/swc/patterns/conversational-ai/pixel-loader/animation.ts
```diff
@@ -0,0 +1,260 @@
...
+      opacity: 0,
+      easing: 'linear',
+    },
+    { offset: 1, opacity: 0 },
+  ];
+  return dedupe(frames);
+}
+
+/** Fully settled appearance: full size, full opacity, no vertical offset. */
+export const SETTLED_TRANSLATE = '0 0px';
```
rubencarvalho: tiny consistency nit: this is `'0 0px'` but the drop keyframes settle at `'0 0%'`, so `_finishThenSwap` interpolates a percentage translate to a px translate. Both endpoints are zero so it's visually harmless right now, but it's a latent hazard if a non-zero settled offset is ever introduced. Probably want `'0 0%'` to match the rest of the module.

miwha-adobe (approved):
lgtm!

PR #6606, opened 2026-08-12
docs(menu): add 2nd-gen migration plan

Rajdeepc:
## Description

Adds the Phase 1 (Preparation) migration plan for `swc-menu` at `CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md`.

This is a documentation-only change; no runtime code is touched.

## Scope of this ticket (read this first)

**The first pass is a deliberately simple menu:**

- Trigger + `PlacementController`-anchored surface + shadow-internal `role="menu"` list (**not** `<swc-popover>` — that's a `role="dialog"` component; menu/listbox/combobox build on the shared `PlacementController` directly, per the popover a11y doc's own design update)
- Plain `swc-menu-item` children only — a label plus **one** leading visual, icon *or* image (not both, not combined with description/value/badge/unavailable content)
- Keyboard navigation between items via roving tabindex (`FocusgroupNavigationController`, already shipped and used by `Tabs`): vertical arrow keys, Home/End, wrap-around, disabled items stay focusable
- Disabled rows and link items (`href`) are in scope

**Explicitly not in this pass** (real, documented, just not now):
- No submenus, no `swc-menu-group`, no divider/separator
- No selection (single/multiple mode) at all
- Richer item content — `Unavailable` state, highlight badge, external-link indicator — deferred alongside the above

See [Implementation phasing](https://github.com/adobe/spectrum-web-components/blob/rajdeepchandra/docs-menu-migration-plan-swc-1980/CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md#implementation-phasing) for the full three-phase breakdown (A: this pass, B: submenus/groups/dividers, C: selection) and [Changes overview](https://github.com/adobe/spectrum-web-components/blob/rajdeepchandra/docs-menu-migration-plan-swc-1980/CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md#changes-overview) for the three short decision tables (Must ship / Additive / Deferred).

## Plan summary (for reviewers)

**Breaking changes, this pass only (B2, B4–B10):**
- Link items drop the proxy double-`change` pattern that caused double activation (B2, fixes SWC-923)
- Keyboard controller swap: `RovingTabindexController` → `FocusgroupNavigationController` (B4, already shipped)
- `--mod-*` CSS surface removed (B5)
- Internal `role="menu"` moves into shadow DOM, off the custom element host (B6)
- Disabled-row ARIA standardized — stays focusable/in the roving set, which is the controller's own default (B7)
- No interactive descendants inside menu item content (B8)
- `ignore` property dropped (B9); `getNeighboringFocusableElement()` dropped as public API (B10)

**Additive, this pass:** `xl` size (A1), image as an alternative to icon (A4).

**Deferred to later phases (documented, not scheduled):** `swc-menu-group` breaking its 1st-gen inheritance (B1), submenus (B3), selection at both menu and group level (A2), `Unavailable` state (A3), highlight badge (A5), external-link indicator (A6), printable-character navigation (A7), mobile drilldown — dropped with no replacement (A8). `swc-menu-separator` is not being built at all; `swc-divider` is composed directly instead.

**Open questions:** down to one real blocker — **Q4**, whether disabled rows should stay focusable (WAI-ARIA APG, and the `FocusgroupNavigationController` default) or match React Spectrum's `disabledKeys` (removes them from focus entirely). Nine other questions are resolved and recorded with their reasoning (architecture model, `xl` size, mobile drilldown scope, trigger API, etc.); three are cross-component follow-ups that don't block this plan.

**Where to start reading:** [Scope of this ticket](#scope-of-this-ticket-read-this-first) above → [TL;DR](https://github.com/adobe/spectrum-web-components/blob/rajdeepchandra/docs-menu-migration-plan-swc-1980/CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md#tldr) → [Blockers and open questions](https://github.com/adobe/spectrum-web-components/blob/rajdeepchandra/docs-menu-migration-plan-swc-1980/CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md#blockers-and-open-questions).

## Motivation and context

Required planning output ahead of `swc-menu` implementation work, per the "Analyze component and create migration plan" ticket for the Menu epic.

## Related issue(s)

- Epic: SWC-1980 (Menu migration epic)
- Prerequisite ticket: SWC-1981 (Menu accessibility recommendations, already incorporated into this plan)

N/A — documentation only.

## Author's checklist

- [x] I have reviewed the Accessibility Practices for this feature (this PR documents a11y expectations sourced from the existing Menu accessibility migration analysis and WAI-ARIA APG; it does not implement runtime behavior).
- [ ] I have added automated tests to cover my changes. — N/A, documentation only.
- [ ] I have included a well-written changeset if my change needs to be published. — N/A, no publishable package changes.

## Reviewer's checklist

### Manual review test cases

- [ ] _Scope matches what's actually buildable this pass_
  1. Read [Scope of this ticket](#scope-of-this-ticket-read-this-first) above
  2. Open the [Changes overview](https://github.com/adobe/spectrum-web-components/blob/rajdeepchandra/docs-menu-migration-plan-swc-1980/CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md#changes-overview) tables
  3. Expect "Must ship" and "Additive" to contain only things a simple label+icon/image menu with roving-tabindex navigation needs — nothing about submenus, groups, dividers, or selection

- [ ] _Plan matches source material_
  1. Open `CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md`
  2. Spot-check the 1st-gen API surface tables against `1st-gen/packages/menu/src/Menu.ts`, `MenuItem.ts`, `MenuGroup.ts`, `MenuDivider.ts`
  3. Expect property/method/event/slot names to match source exactly

- [ ] _Remaining open question is the right one to block on_
  1. Read [Blockers and open questions](https://github.com/adobe/spectrum-web-components/blob/rajdeepchandra/docs-menu-migration-plan-swc-1980/CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md#blockers-and-open-questions)
  2. Expect Q4 (disabled-row focus behavior) to read as genuinely needing a named decision, and the 9 resolved items to read as settled with clear reasoning, not hand-waved

### Device review

- [ ] Did it pass in Desktop?

- [ ] **Keyboard** — N/A, documentation only; no interactive component ships in this PR. The plan documents the intended keyboard model for this pass concretely: `FocusgroupNavigationController` with `direction: 'vertical'`, `wrap: true`, `skipDisabled: false` (the controller's own default), Home/End built in, Escape to close.
- [ ] **Screen reader** — N/A, documentation only. The plan documents the intended `role="menu"`/`menuitem` semantics and shadow-DOM placement for the future implementation to satisfy.


CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
```diff
@@ -0,0 +1,741 @@
...
+| `open` | `boolean` | `false` | `open` | **Inferred**, matching the menu-button pattern's open/close state, parallel to `swc-action-menu`. |
+| `disabled` | `boolean` | `false` | `disabled` | **Inferred** carry-forward for the trigger. |
+| `trigger` | `'press' \| 'longPress' \| 'contextMenu'` | `'press'` | `trigger` | **Open question** ([Q6](#blocking-for-swc-menu)) — React Spectrum S2's `MenuTrigger.trigger` has no 1st-gen equivalent (1st-gen composes the trigger externally). Depends on whether `swc-menu` owns trigger-invocation semantics itself or defers entirely to how the consumer wires the trigger slot. |
+| `shouldCloseOnSelect` | `boolean` | `true` (**inferred**, matches 1st-gen's default close-on-select for non-`selects` items) | `should-close-on-select` or similar | **Open question**, tied to [Q2](#blocking-for-swc-menu)/[Q5](#blocking-for-swc-menu) — only meaningful once selection ships, since 1st-gen's unconditional `close` event today only fires for non-selectable items. React Spectrum S2 exposes this at both `Menu` and `MenuItem` level. |
```
rubencarvalho: Do we need to offer this API?

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
```diff
@@ -0,0 +1,741 @@
...
+Additional Figma-confirmed presentation modes for `swc-menu-item` (documented here for `swc-menu`'s composition awareness; the item-level plan owns the authoritative API):
+
+- Default / hover / disabled states
+- Selection: none / single / multi-select (checkbox) / multi-select (switch) — **open question**, see [Q2](#blocking-for-swc-menu)
+- Show leading icon, show drill-in (submenu chevron), show description, show value — **Confirmed**, map to existing 1st-gen slots/behavior
+- Show thumbnail, show highlight badge, show external link, unavailable — **Inferred** additive, no 1st-gen equivalent ([A3](#additive--ships-when-ready-zero-breakage-for-consumers-already-on-2nd-gen)–[A6](#additive--ships-when-ready-zero-breakage-for-consumers-already-on-2nd-gen))
+
+#### Slots (2nd-gen)
+
+| Slot | Content | Notes |
```
rubencarvalho, line 497: So not slot for the actual label? Then the "label" description on the properties above is incorrect. How do we put text on the trigger button?

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
```diff
@@ -0,0 +1,741 @@
...
+| --- | --- | --- | --- | --- |
+| **B1** | `swc-menu-group` no longer extends the full menu host | `MenuGroup extends Menu` — inherits selection engine, roving tabindex, `value`, etc., and overrides `ownRole`/`controlsRovingTabindex` to suppress most of it | `swc-menu-group` is a plain grouping/labeling primitive (per a11y analysis); no selection or roving-tabindex API of its own | Consumers using `sp-menu-group` only for grouping + `header` slot see no functional change; any (unsupported) reliance on inherited `Menu` API on a group instance breaks. Source: [Menu group a11y doc](../menu-group/accessibility-migration-analysis.md#what-swc-menu-group-is-2nd-gen). |
+| **B2** | Link items drop the proxy double-activation pattern | `href` items fire `change` via `Menu`'s click handling in addition to native anchor navigation ([SWC-923](https://jira.corp.adobe.com/browse/SWC-923)) | Real `<a href>` descendant is the only activation path; no proxy `change` dispatch for link rows | Consumers listening for `change` on link-style items must switch to listening on the anchor itself or the native `click`/navigation. |
```
rubencarvalho: Link styling may become difficult, if we say <a> tags will just be slotted.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
```diff
@@ -0,0 +1,741 @@
...
+- **Inferred**: recommended based on multiple signals, but not explicitly specified in one authoritative source
+- **Open question**: unresolved and needs review or more input
+
+### Public API
+
+#### Properties / attributes (2nd-gen)
+
+| Property | Type | Default | Attribute | Notes |
+| --- | --- | --- | --- | --- |
+| `label` | `string` | `''` | `label` | **Inferred** carry-forward from 1st-gen; accessible name for the trigger. |
```
rubencarvalho: Is this the trigger button text?

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
```diff
@@ -0,0 +1,741 @@
...
+- Default / hover / disabled states
+- Selection: none / single / multi-select (checkbox) / multi-select (switch) — **open question**, see [Q2](#blocking-for-swc-menu)
+- Show leading icon, show drill-in (submenu chevron), show description, show value — **Confirmed**, map to existing 1st-gen slots/behavior
+- Show thumbnail, show highlight badge, show external link, unavailable — **Inferred** additive, no 1st-gen equivalent ([A3](#additive--ships-when-ready-zero-breakage-for-consumers-already-on-2nd-gen)–[A6](#additive--ships-when-ready-zero-breakage-for-consumers-already-on-2nd-gen))
+
+#### Slots (2nd-gen)
+
+| Slot | Content | Notes |
+| --- | --- | --- |
+| default | `swc-menu-item`, `swc-menu-group`, `swc-menu-separator` only | **Confirmed** per a11y analysis; verify enforcement in 2nd-gen source once implemented. |
```
rubencarvalho: Gen1 ActionMenu had a "label" slot, we could do something like that? Alternatively calling it "trigger" slot.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
Rajdeepc: Confirmed we don't need it. Leaving it out of Phase A, can add it later without breaking anything if a real need comes up.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
Rajdeepc: Yeah that was a real gap. Added a `label` slot for the visible trigger content, matching sp-action-menu. The `label` property sticks around too, just doing double duty now.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
Rajdeepc: Good point, that's a real risk. Not really this plan's call though, that's menu-item's own territory. Filed it as a follow-up (Q17) pointing at whoever ends up owning that plan.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
Rajdeepc: It was only meant to be the accessible name, but yeah, that leaves nothing for the visible text. Added a `label` slot for that.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
Rajdeepc: Went with your first suggestion, a `label` slot matching sp-action-menu. Keeping it to just the one slot for now instead of also copying over the separate icon/label-only split. Thanks for catching this, it was a real hole.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
5t3ph: We have dropped the "accessible name fallback" chain in other instances and I vote we do so here as well, it's complication we don't need to own when our new standard is use of `accessible-label` if it needs an `aria-label` for some reason (which should be lower probability for this type of menu).

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
5t3ph: Just a note: We can probably use the CSS link expansion trick (also used for Card) to increase the slotted links surface to cover the menu item without it actually needing to wrap the other elements, but we may need to test that out.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
5t3ph: I misinterpreted this and didn't realize it was the _trigger_ label.

**New question/concern**: do we have to nest the trigger in shadow dom, or can we either accept a trigger into a slot so it keeps it's own API and any applicable dev warnings, or even borrow Tooltip's `for` pattern (or similar) to associate a menu with a trigger? If we don't allow it to be separate, then we either need to take on the full API for ActionButton or we consequently exclude features as basic as icon-only for the trigger. If that exclusion and limitation is intentional then we need to be clear about it.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
```diff
@@ -0,0 +1,736 @@
...
+- **Open question**: unresolved and needs review or more input
+
+### Public API
+
+#### Properties / attributes (2nd-gen)
+
+| Property | Type | Default | Attribute | Notes |
+| --- | --- | --- | --- | --- |
+| `label` | `string` | `''` | `label` | **Decided** ([Q16](#blockers-and-open-questions)) — carried forward from 1st-gen, but its role changes now that `swc-menu` owns a trigger: it's the fallback text rendered inside the new `label` slot (see Slots (2nd-gen), below) when no light-DOM content is slotted, and it contributes to the trigger's accessible name whenever no `aria-label`/`aria-labelledby` is present. Same dual role 1st-gen `sp-action-menu`'s `label` property already plays alongside its own `label` slot ([`ActionMenu.ts`](../../../../1st-gen/packages/action-menu/src/ActionMenu.ts)). |
+| `size` | `'s' \| 'm' \| 'l' \| 'xl'` | `'m'` | `size` | **Open question** on default (1st-gen has no default); `xl` addition is **Inferred** from Figma ([A1](#additive)). |
```
5t3ph: Leaning on React we can confirm size `m` as default, and the existence of `xl` as well.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
```diff
@@ -0,0 +1,736 @@
...
+| **B8** | Item content | Text/icon/image only; no interactive descendants |
+| **B9** | `ignore` property | Dropped — no use case once nested `Menu`-in-`Menu` inheritance is gone |
+| **B10** | `getNeighboringFocusableElement()` | Dropped as public API, superseded by the controller |
+
+### Additive
+
+| # | What's added | Note |
+| --- | --- | --- |
+| **A1** | `xl` size | Figma adds `xl` to 1st-gen's `s`/`m`/`l` |
+| **A4** | Image as an alternative to icon | One or the other, not both; no 1st-gen equivalent |
```
5t3ph: Recommendation: use of a slot called `visual` to support _either_ an icon _or_ a swc-thumbnail.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
```diff
@@ -0,0 +1,736 @@
...
+
+Full property list is intentionally left partial until Phase C (the selection surface, [Q2](#blockers-and-open-questions)) is scheduled and designed — filling in a complete table now would present an undesigned surface as settled.
+
+#### Visual matrix (2nd-gen)
+
+| Visual family | S | M | L | XL |
+| --- | --- | --- | --- | --- |
+| Menu group / menu item | Yes (**Confirmed**, Figma) | Yes (**Confirmed**, Figma; 1st-gen default context) | Yes (**Confirmed**, Figma) | Yes (**Inferred** — new, not in 1st-gen) |
+
+Additional Figma-confirmed presentation modes for `swc-menu-item` (documented here for `swc-menu`'s composition awareness; the item-level plan owns the authoritative API):
```
5t3ph, line 486: Even though you're only planning a basic mode, we'll want to be future-thinking for slot names and grid template setup. I have a prototype for this, reach out during the styling phase to discuss!

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
5t3ph: ^ Actually, nevermind. We likely do need to own the link as a wrapper in order to make sure it's the only focusable element (not a nested focus target _inside_ the menu item) as well as to apply the role to it instead of the non-link element.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
```diff
@@ -0,0 +1,736 @@
...
+- **`swc-popover`** — shipped, but not a dependency of `swc-menu` at all (see [Dependencies](#dependencies)).
+- **`swc-action-menu`** — structurally parallel to `swc-menu` (same trigger/`PlacementController`/`role="menu"` shape, different default trigger chrome). Not yet started. The two migrations should stay in lockstep on shared decisions (in-menu keyboard model, disabled-row ARIA, link-item pattern, `PlacementController`-direct anchoring) to avoid divergence; neither should be treated as strictly upstream of the other, but shared logic (if any is factored out) should land once, not twice. Its a11y doc's anchoring section needs updating to `PlacementController` — see [Q13](#cross-component-follow-ups-not-blocking-swc-menu).
+- **`swc-menu-item`, `swc-menu-group`** — out of scope for this plan (see the epic-level note at the top of this document) but are direct composition dependencies of `swc-menu`. `swc-menu`'s own testing and Storybook work cannot be completed until at least a minimal `swc-menu-item` exists. `swc-menu-item`'s a11y doc's submenu-anchoring section needs the same `PlacementController` update ([Q13](#cross-component-follow-ups-not-blocking-swc-menu)).
```
5t3ph, line 356: Given the need for the minimal `swc-menu-item` are you planning to do that plan next before starting work on this, or do you want to add an API table of what you will support in this first phase, including concerns like applying `role="menuitem"`?

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
```diff
@@ -0,0 +1,736 @@
...
+
+### Blocking for swc-menu
+
+| # | Item | Blocking? | Status | Owner |
+| --- | --- | --- | --- | --- |
+| Q4 | [B7](#must-ship) (disabled rows stay focusable/in the roving set) follows WAI-ARIA APG's permissive guidance, but this plan's own [React Spectrum S2 API surface](#react-spectrum-s2-api-surface) section quotes React Spectrum's `disabledKeys` as making items "cannot be selected, focused, or otherwise interacted with" — the opposite behavior. This plan's source-priority rules rank React implementation #1 for behavior; B7 currently resolves the conflict in APG's favor without naming it. Weighs in further toward APG now that it's confirmed to be `FocusgroupNavigationController`'s own default (`skipDisabled: false`) — following React Spectrum instead would mean actively overriding the controller, not just picking a default. Still needs to be a named decision, not an assumed one, given the source-priority conflict. | Yes | Open — affects B7's final wording and the Accessibility checklist | Accessibility reviewer |
```
5t3ph: Tagging in @nikkimk for a definitive call.

We have precedent for keeping disabled group elements focusable for the [incoming action group](https://swcpreviews.z13.web.core.windows.net/pr-6444/docs/gen2-storybook-prod/iframe.html?id=components-action-group--docs&viewMode=docs#disabled), and it does seem that it usually would be beneficial to keep disabled menu items discoverable. However, a future feature is the "unavailable" option which is usually used to explain why a menu item may not be available (ex. extra permission needed, not available with current plan, etc) so it remains active in the list but effectively disables the menu item to instead show the help message.

5t3ph (commented):
See also the critical decision needed about trigger management [in my comment reply.](https://github.com/adobe/spectrum-web-components/pull/6606#discussion_r3814795028)

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
Rajdeepc: Update: the label-slot fix above is out of date. Turns out neither Tooltip nor Popover render their own trigger either, they both reference an external one via for/triggerElement. sp-action-menu was the wrong thing to copy. Switched swc-menu to do the same: consumer brings their own trigger element, references it via for/triggerElement, and we wire aria-haspopup/aria-expanded onto it. No label slot needed anymore. Written up as Q16.

Rajdeepc: Agreed, and turns out it's moot rather than a tradeoff. Went with for/triggerElement instead of rendering our own trigger, so there's no fallback chain to worry about at all, the name just comes from whatever the consumer's trigger already has.

Rajdeepc: Went with for/triggerElement, borrowing Tooltip's pattern like you suggested. Neither Tooltip nor Popover nest their trigger in shadow DOM, they resolve it externally and it even unwraps into a shadow component's inner button automatically for ARIA. So consumers get a full swc-action-button as the trigger with its whole API, icon-only included, nothing excluded. Also means the co-located trigger + role=menu language in the menu a11y doc is stale now in a few places, filed that as Q18.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
Rajdeepc: Agreed, that's where this landed too, swc-menu-item owning the anchor internally rather than a consumer-slotted one. Updated B2 and added a minimal item table that assumes this. It does contradict the current menu-item a11y doc though, so flagged that as part of Q17 for whoever picks up that plan.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
Rajdeepc: Updated to Confirmed, citing React Spectrum's Menu.size directly (default M). The table already said 'm', the notes column was just stale.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
Rajdeepc: Done, renamed to a visual slot for icon or swc-thumbnail. Thumbnail hasn't started 2nd-gen work yet so Phase A only really exercises the icon case, but the name should hold up once it exists.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
Rajdeepc: Makes sense, added a note to sync with you during the styling phase instead of locking anything in now.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
Rajdeepc: Went with adding an inline table instead of waiting on a separate plan. Added a 'Phase A minimal swc-menu-item surface' section covering role, content, disabled, the link model, and focus. Marked it provisional since menu-item's own plan can override it once it exists.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
Rajdeepc: Good catch on the Unavailable overlap, added that to Q4 along with the Action Group precedent. Leaving the actual call to nikkimk rather than guessing.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
```diff
@@ -352,6 +355,20 @@ Per [Q5](#blockers-and-open-questions), `picker`/`combobox`/`action-menu`'s depe
 - **`swc-divider`** — composed directly as the menu separator (decided, [Q14](#blockers-and-open-questions)); no separate `swc-menu-separator` element.
 - **`picker`, `combobox`, `action-menu`** (as consumers) — their 2nd-gen migrations will depend on `swc-menu`/`swc-menu-item` existing. Decided ([Q5](#blockers-and-open-questions)): not pulling baseline single-select into Phase A for their sake, since none has a scheduled migration yet.
 
+### Phase A minimal `swc-menu-item` surface
+
+`swc-menu-item` has no `migration-plan.md` of its own yet, but `swc-menu`'s own Phase A stories and tests cannot run without something to compose. Rather than pausing this plan to wait for that plan (5t3ph review, [Q15](#cross-component-follow-ups-not-blocking-swc-menu)), this table records the minimal surface Phase A actually exercises. **Provisional**: `swc-menu-item`'s own plan is authoritative once it exists and may change any of this.
+
+| Concern | Phase A minimal behavior |
+| --- | --- |
+| Role | `role="menuitem"` on the host or its shadow-internal row, per the [Menu item a11y doc](../menu-item/accessibility-migration-analysis.md) |
```
5t3ph: I believe what that doc is saying is to use that role on the menu-item host which it has shortened to "row", and the shadow-internal references are in regard to the parent menu wrapper that contains `role="menu"`.

_However_ that breaks down if we nest a true `a` - which we should! - so we will probably have to force `role="presentation"` on the `swc-menu-item` element so it's effectively skipped by the a11y tree and it just sees either the `div` or `a` with the `menuitem` role. Otherwise, the a11y tree won't validate if there isn't direct descendants of role=menu > role=menuitem.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
```diff
@@ -352,6 +355,20 @@ Per [Q5](#blockers-and-open-questions), `picker`/`combobox`/`action-menu`'s depe
...
+| Content | Default slot (label text) plus one leading `visual` slot (icon; `swc-thumbnail` once that component exists — [A4](#additive)), not both content types combined |
```
5t3ph: I know we've been going a bit back and forth on this, but given we'll also have description text as well I would strongly suggest we go ahead and make `label` a named slot, not default, so there's no ambiguity on the intent of the text for a consumer.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
```diff
@@ -352,6 +355,20 @@ Per [Q5](#blockers-and-open-questions), `picker`/`combobox`/`action-menu`'s depe
...
+| Disabled | `disabled` property/attribute → `aria-disabled="true"`; stays in `FocusgroupNavigationController`'s roving set per [B7](#must-ship) |
+| Link rows | Owns its `<a href>` internally as a wrapper (revised, [Q17](#cross-component-follow-ups-not-blocking-swc-menu)) so it's the row's sole focusable element and carries the row's role; not a consumer-slotted anchor |
```
5t3ph: This will at minimum require a `href` attribute and that's probably enough to also signal this is a link row.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
```diff
@@ -674,13 +697,13 @@ Grouped by whether the question is about `swc-menu` itself or about a sibling co
...
...gger-anchored components: `Tooltip.base.ts` and `Popover.base.ts` both resolve their trigger through the shared [`resolveTrigger` utility](../../../../2nd-gen/packages/core/utils/resolve-trigger.ts) (`for` ID reference or direct `triggerElement`) rather than rendering one internally — `resolveTrigger` even unwraps into an open-shadow SWC component's inner `<button>` automatically, so the consumer's trigger keeps its full existing API (icon-only, pending, disabled, whatever `<swc-action-button>` already supports) for free. `sp-action-menu` was the wrong precedent: it's a 1st-gen component that has always owned its own trigger, unlike `swc-tooltip`/`swc-popover` (the actual 2nd-gen siblings) or 1st-gen `sp-menu` itself, which this plan's own TL;DR already notes "was most often just the list under an externally-composed trigger." This removes the `label` slot, the `label` property's fallback-text role (5t3ph's first objection dissolves too: there's no separate accessible-name chain to own since the name comes from whichever trigger the consumer authored), and the deferred `icon`/`label-only` slot split that would have applied to it (there's no trigger content slot to split anymore). |
```
5t3ph: [nit] some of this commentary could move to a "decision log" if you feel it's worth recording (same for a few other places)

5t3ph (commented):
A few last things - glad to see carry through of the `for` / `resolveTrigger` pattern!

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
Rajdeepc: That clarifies it, thanks. Updated the Role row: menuitem goes on whichever element is actually interactive, never both. For link rows specifically, the host needs role=presentation so the a11y tree sees menu > menuitem directly. Filed as Q19.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
Rajdeepc: Agreed, switched to a named label slot instead of default. Makes sense once description is sitting right next to it.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
Rajdeepc: Agreed, updated it so href alone is enough to signal a link row, no separate flag needed.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
nikkimk: Given our one role per component. Perhaps we should create a separate `swc-menu-link` with presentation role and slotted anchor. That removes the extra conditional logic and complications.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
nikkimk: See my not about doing a dedicated `swc-menu-link` for menu items that have links.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
5t3ph: Maybe but that's a lot of internal logic/render repeats for just swapping div for `a`. Again, I don't think it can truly be slotted from the consumer due to styling limitations of `::slotted()` plus the need to apply the role to the `a` directly.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
```diff
@@ -0,0 +1,783 @@
...
+
+These are derived from the 1st-gen implementation, the accessibility migration analysis, the Figma source reviewed for this plan, and the React Spectrum S2 implementation. Confirmed items are marked; open items are tracked in [Blockers and open questions](#blockers-and-open-questions).
+
+- **Confirmed**: directly supported by source material
+- **Inferred**: recommended based on multiple signals, but not explicitly specified in one authoritative source
+- **Open question**: unresolved and needs review or more input
+
+### Public API
+
+#### Properties / attributes (2nd-gen)
```
aramos-adobe, line 460: Do we need to include `placement | offset | should-flip` like Tooltip/Popover? I'm assuming that is private/internal for menu.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
```diff
@@ -0,0 +1,783 @@
...
+| Q15 | Migration order among `swc-menu`, `swc-menu-item`, and `swc-menu-group` is not yet decided (`swc-divider` already exists, and `swc-menu-separator` isn't being built — [Q14](#blockers-and-open-questions)). `swc-popover` is **not a sequencing variable** — it shipped in [PR #6356](https://github.com/adobe/spectrum-web-components/pull/6356) (merged 2026-07-14, before this plan was drafted) and `swc-menu` doesn't depend on it anyway. Each menu-family component either has or will have its own plan; sequencing among them is what remains open. **Decided not to block on this** (5t3ph review): rather than pausing `swc-menu`'s Phase A to wait for a `swc-menu-item` plan, this plan now documents the minimal item surface Phase A needs directly — see [Phase A minimal `swc-menu-item` surface](#phase-a-minimal-swc-menu-item-surface). The relative-priority question for a full `swc-menu-item` plan is still open. | Open | Ticket owner |
```
rise-erpelding: I noticed that we don't have separate tickets (at least as far as I could find) for migrating other menu family components, for instance I didn't see a migration epic for for menu-item, or menu-group.

We probably wouldn't forget to migrate any of these but it would be a good idea to have something ticketed soon!

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
```diff
@@ -0,0 +1,783 @@
...
+- `sp-menu`'s selection engine (`selects`, `value`, `selectedItems`, `selectOrToggleItem`) is consumed indirectly by `picker`, `combobox`, and `action-menu` today (all import or extend `sp-menu`). Decided ([Q5](#blockers-and-open-questions)): selection stays Additive/Phase B rather than pulled into Phase A — none of those three consumers has a concrete migration scheduled yet, so there's no live deadline forcing it earlier.
+- **Recommended implementation phasing:** ship `swc-menu` in two stages rather than one — Phase A covers the full menu-button host without selection (trigger, `PlacementController` anchoring, items, submenus, all Must-ship breaking changes), Phase B adds the selection engine. This reduces what reviewers validate per PR and lets Phase A land without waiting on [Q2](#blockers-and-open-questions) to resolve. See [Implementation phasing](#implementation-phasing).
```
rise-erpelding: Do we want to mention Phase C here?

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
```diff
@@ -0,0 +1,783 @@
...
+- **Selection is real, not hypothetical — the a11y doc is stale here.** Beyond the Menu group/Menu item property tables (`Selection: None | Single | Multi-select (checkbox) | Multi-select (switch)`, `Unavailable`, `Show thumbnail`, `Show highlight badge`), a dedicated "Changes in S2 Menu" frame in the same Figma file states design will "continue to support checkboxes in menu items" and is "evaluating optimal use cases for multi-select checkboxes in picker menus," and a live "Examples" frame shows a working multi-select checkbox menu. This directly contradicts the [a11y analysis's Migration scope](./accessibility-migration-analysis.md#migration-scope-current), which defers all selection pending a product decision that, per this evidence, appears to have already been made. See [Q2](#blockers-and-open-questions).
```
rise-erpelding, line 84: I don't know that we have a team norm for this at all, and we never did make a firm decision on what to do with other docs in this directory (the a11y migration analysis and the rendering and styling migration analysis). I am all for folding them into a single planning doc during the course of the migration to give us less documentation to deal with.

That said, if we have no immediate plans to remove the sibling docs, we might as well keep them up to date. Instead of calling out things in the a11y doc as being stale or incorrect, we could update the a11y doc in this PR. By doing so, our a11y doc stays up to date, and we don't need to say anything here about how it's not up to date.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
Rajdeepc: Just want to clarify few things. @nikkimk do you mean "slotted anchor" as in "the anchor is where content gets slotted" i.e., `swc-menu-link` still owns and renders its own internal `<a>`, just as a permanently-link-shaped component rather than a conditional one, then @5t3ph  `::slotted()` engineering cost objection will not apply. 
Let's discuss what "slotted anchor" means.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
Rajdeepc: Ah nice catch. I think this was a miss and there are genuinely three plausible answers, not one obvious one: 
full Tooltip/Popover-style parity, a narrower React Spectrum-style align/direction API translated internally, or fully opinionated/private with no public knobs at all. This needs a decision.

Rajdeepc: Not quite private, but not full Tooltip/Popover parity either. Neither of them actually expose PlacementController's raw Placement union, Tooltip narrows to 6 cardinals and Popover to 12 RTL-aware values (calls it 'the proxy pattern' internally), so matching them fully was never really the right comparison.

Went with direction/align instead, matching React Spectrum S2's MenuTrigger props rather than inventing a third vocabulary. Menu.base.ts translates that into one internal Placement value. shouldFlip carries straight over since everyone already agrees on it. offset/cross-offset stay unexposed though, MenuTrigger has no equivalent and our gap to the trigger is a fixed spec value, not something Tooltip/Popover-style variable content needs.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
```diff
@@ -0,0 +1,790 @@
...
+- **Recommended implementation phasing:** ship `swc-menu` in two stages rather than one — Phase A covers the full menu-button host without selection (trigger, `PlacementController` anchoring, items, submenus, all Must-ship breaking changes), Phase B adds the selection engine. This reduces what reviewers validate per PR and lets Phase A land without waiting on [Q2](#blockers-and-open-questions) to resolve. See [Implementation phasing](#implementation-phasing).
+- In-menu keyboard movement moves from `RovingTabindexController` to `FocusgroupNavigationController` ([PR #6129](https://github.com/adobe/spectrum-web-components/pull/6129), already shipped and in production use by `Tabs.base.ts`), matching the approach already recommended for `swc-action-menu`.
+- The Figma source (`Mngz9H7WZLbrCvGQf3GnsY`) matches the ticket's `S2/Web (Desktop scale)` file by key. Its title now reads **"🚫 S2 / Web (Deprecated)"** (Beta, v1.27.1) — confirmed with Design ([Q1](#blockers-and-open-questions)) that this is the only file; there is no newer replacement. Content pulled from it is final.
```
rise-erpelding, line 89: There is a page for "Menu" in the "S2 / Web" file (`xHBWBBIe2eo5vwoCeNrC4Q`, not the same as the deprecated one), is there a reason that one won't work?

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
```diff
@@ -0,0 +1,790 @@
...
+| Q15 | Migration order among `swc-menu`, `swc-menu-item`, and `swc-menu-group` is not yet decided (`swc-divider` already exists, and `swc-menu-separator` isn't being built — [Q14](#blockers-and-open-questions)). `swc-popover` is **not a sequencing variable** — it shipped in [PR #6356](https://github.com/adobe/spectrum-web-components/pull/6356) (merged 2026-07-14, before this plan was drafted) and `swc-menu` doesn't depend on it anyway. Each menu-family component either has or will have its own plan; sequencing among them is what remains open. **Decided not to block on this** (5t3ph review): rather than pausing `swc-menu`'s Phase A to wait for a `swc-menu-item` plan, this plan now documents the minimal item surface Phase A needs directly — see [Phase A minimal `swc-menu-item` surface](#phase-a-minimal-swc-menu-item-surface). The relative-priority question for a full `swc-menu-item` plan is still open. | Open | Ticket owner |
```
nikkimk: Using `swc-divider` is acceptable. I would note in the accessibility section of the divder docs that it can be used in menus but not in pickers and comboboxes

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
```diff
@@ -0,0 +1,790 @@
...
... state, full-row hit target, focus ring, and icon alignment is materially harder than 1st-gen's shadow-internal `<a class="anchor">` render. A CSS link-expansion trick (as used for Card) was considered and set aside: it doesn't solve the deeper problem, which is that a consumer-slotted `<a>` risks becoming a second focusable element nested inside the row, and the wrong node ends up carrying the row's role. **Leaning conclusion: `swc-menu-item` should own the `<a>` internally as a wrapper**, matching 1st-gen's shadow-internal render, so it's provably the row's sole focusable element and carries the role directly. This contradicts the current [menu-item a11y doc](../menu-item/accessibility-migration-analysis.md), which says "consumers should slot a real `<a href>`" — that doc needs correcting to match, the same pattern as [Q13](#cross-component-follow-ups-not-blocking-swc-menu). This plan doesn't implement `swc-menu-item`, but the conclusion is settled enough to use in [Phase A minimal `swc-menu-item` surface](#phase-a-minimal-swc-menu-item-surface) above. | Open — direction settled, needs `swc-menu-item` plan owner to formalize and correct the a11y doc | `swc-menu-item` plan owner |
```
nikkimk: I would recommend a dedicated `swc-menu-link` and separate concerns rather than allow menuitems to be links. The two behave differently and have different roles. Let's keep these separate and let the consumers consciously choose a link over a regular menu item.

We could fast follow and I could create an a11y doc for this.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
```diff
@@ -0,0 +1,790 @@
...
+| Q19 | (5t3ph review) Once [Q17](#cross-component-follow-ups-not-blocking-swc-menu)'s link-wrapper conclusion ships, `swc-menu-item` cannot put `role="menuitem"` on both its host and the internal `<a>` — the `menu`/`menuitem` required-owned-elements relationship needs `menuitem` on exactly one node, with nothing else carrying a competing implicit or explicit role in between. For link rows, that means `role="presentation"` on the host so the accessibility tree skips over it and sees `menu` > `a[role=menuitem]` directly, not `menu` > (unstyled semantic host) > `a`. Not this plan's implementation to make, but the constraint is concrete enough to hand to whoever builds `swc-menu-item` rather than leaving it to be discovered during an a11y audit. | Open | `swc-menu-item` plan owner |
```
nikkimk: Again I advise an `swc-menu-link` element rather than overloading `swc-menu-item` as we had in the past.

nikkimk (changes requested):
This is quite the component to migrate and plan for. Thanks for pulling this all together. I have noted some changes around how to handle links.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
```diff
@@ -0,0 +1,790 @@
...
+
+### Events
+
+| Event | Bubbles / composed | Cancelable | Fired by | Notes |
+| --- | --- | --- | --- | --- |
+| `change` | yes / yes | yes (from `selectOrToggleItem`); no (from href-link path) | `Menu` | Fires on selection change, and unconditionally (non-cancelable) when an href item is activated — the href path is the known double-activation source ([SWC-923](https://jira.corp.adobe.com/browse/SWC-923)). |
+| `close` | yes / no | no | `Menu` | Fires when a non-submenu, non-selects item is activated, signaling ancestors (overlay/picker) to close. |
+| `sp-menu-submenu-opened` | yes / yes | no | overlay target (via `Menu.handleSubmenuOpened`) | Internal to the 1st-gen overlay-based submenu wiring. |
+| `sp-menu-submenu-closed` | yes / yes | no | overlay target (via `Menu.handleSubmenuClosed`) | Internal to the 1st-gen overlay-based submenu wiring. |
+| `sp-menu-item-added` (`MenuItemAddedOrUpdatedEvent`) | — | — | `MenuItem` | Announces item registration to the owning menu. |
```
rise-erpelding: [nit] I think the JSDoc in 1st-gen is not up to date, I think this is supposed to be `sp-menu-item-added-or-updated`

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
Rajdeepc: Thanks I was also thinking the same.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
Rajdeepc: I dont have a strong opinion to oppose this pattern. let's try this

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
Rajdeepc: Yes that would be nice! If you can create a ticket I would add it here so that AI can create context of a follow up work.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
```diff
@@ -0,0 +1,790 @@
...
+  <!-- (mobile-view only) back row, drilldown containers -->
+
+<!-- sp-menu-item -->
+#shadow-root
+  <a class="anchor" href="..."> <!-- only when href is set -->
+    <slot name="icon"></slot>
+    <slot></slot>
+    <slot name="description"></slot>
+    <slot name="value"></slot>
+  </a>
```
rise-erpelding: [nit] I know this is meant to be a simplified version of the DOM output and I'm not sure if this matters re: the other conversation threads about the link, but in practice the link doesn't wrap if `href` is set:

```ts
      ${this.selected
        ? html`
            <sp-icon-checkmark100
              id="selected"
              class="spectrum-UIIcon-Checkmark100
                            icon
                            checkmark
                            ${this.hasIcon
                ? 'checkmark--withAdjacentIcon'
                : ''}"
            ></sp-icon-checkmark100>
          `
        : nothing}
      <slot name="icon"></slot>
      <div id="label">
        <slot id="slot"></slot>
      </div>
      <slot name="description"></slot>
      <slot name="value"></slot>
      ${this.href && this.href.length > 0
        ? super.renderAnchor({
            id: 'button',
            ariaHidden: true,
            className: 'button anchor hidden',
          })
        : nothing}
      ${this.renderSubmenu()}
```

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
```diff
@@ -0,0 +1,792 @@
...
+| `swc-popover` (2nd-gen, shipped in [PR #6356](https://github.com/adobe/spectrum-web-components/pull/6356)) | `2nd-gen/packages/swc/components/popover` | **Not a dependency of `swc-menu`.** It's a self-contained `role="dialog"` component; menu/listbox/combobox content is explicitly excluded from its intended use per the [popover a11y doc's 2nd-gen design update](../popover/accessibility-migration-analysis.md#2nd-gen-design-update-amends-this-analysis--q4). |
+
+---
+
+## Open gen1 issues
+
+<!-- No Jira query access was available while drafting this plan, and this is not tracked as a blocker for the plan (decided). If useful later, the query is `project = SWC AND component = "Menu" AND labels NOT IN (a11y, gen2) AND issuetype NOT IN (Epic, Initiative) AND status != Done` (adjust field names to match the actual Jira schema) — do not reuse the "Related 1st-gen accessibility (Jira)" table from accessibility-migration-analysis.md, since those rows carry the `a11y` label and are excluded from this table by definition. -->
+
+| Jira | Type | Status (snapshot) | Summary |
+| --- | --- | --- | --- |
```
rise-erpelding: Do we want to manually query for issues? And if there aren't any, note that outside of a comment?

I also didn't realize that we weren't supposed to add a11y issues here (per the skill), that seems like it would make it easier to miss fixing them in the migration?

rise-erpelding (commented):
Menu is a huge component and there's a lot to consider here, so I didn't get all the way through the plan but will continue looking through it!

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
Rajdeepc: Thanks. Should be good now!

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
Rajdeepc: You're right, just checked MenuItem.render() and it doesn't wrap. The content renders as plain siblings; the anchor is a separate, empty, hidden element alongside them just for click/keyboard activation. Fixed the snippet, and adjusted Q17's wording too since it had assumed the wrong mechanism.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
Rajdeepc: Yeah, and it wasn't just incomplete, it was actually wrong: said two stages with Phase B as selection, but everywhere else in this doc uses three phases with C as selection. Fixed it to match.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
Rajdeepc: Agreed this is a real gap, but I don't have Jira access to create the epics myself (same limitation noted in Open gen1 issues below). Can you or whoever owns SWC-1980 file tickets for swc-menu-item and swc-menu-group specifically? Once they exist I'll wire the references into this plan (Q15 already tracks the sequencing question but references a generic 'Ticket owner' rather than a real ticket).

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
Rajdeepc: I like it, updating the a11y doc directly is better than just flagging it stale if we're going to do it anyway. I held back mainly because it widens this PR into files it didn't originally touch. We can touchbase on this in one of our calls.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
Rajdeepc: I don't have a reason it wouldn't work, I don't have that file's contents to compare against, only what was confirmed for the deprecated file (Q1). This reopens Q1: 'confirmed with Design' was based on the file the ticket originally linked, and if xHBWBBIe2eo5vwoCeNrC4Q is actually current, everything pulled from the deprecated file (sizes, variants, the selection/Unavailable/thumbnail/badge properties referenced throughout) needs re-verifying against it before this plan can be called final. Can you confirm with Design which file is authoritative? I'd rather get a real answer than guess and silently swap the citation.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
Rajdeepc: That a11y-issue exclusion isn't something I decided, it's just how the migration-prep template is set up, keep a11y issues in the a11y doc's table, not duplicated here. But your concern's fair, splitting them by label does make it easier to lose track of the a11y ones during implementation. Might be worth raising with whoever owns that template. Want me to flag it, or would you rather since you're the one who caught it?

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
Rajdeepc: Agreed, real gap. I'll wire the reference in once it exists. Will check with Nikki.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
rise-erpelding: Right, I did read the skill and I see that this is not meant to be a table for a11y tickets, it's just how the template is set up. I am happy to raise that concern in a sync later.

As for this table though, I did run a manual check in Jira, it looks like there are about 10 tickets it surfaced; we should add them here rather than leaving this table empty.

nikkimk (approved):
lgtm. thanks for incorporating my suggestions.

5t3ph (approved):
Let's goooo :shipit:

PR #6585, opened 2026-08-06
feat(prompt-field): add collapsed layout with animated transition

rubencarvalho:
## Description

Adds a `collapsed` boolean property to `swc-prompt-field`, matching the expanded/collapsed variant naming used in the Figma AI Prompt Field spec.

- **Expanded (default)**: taller card with a separate action bar below the textarea (upload button + send/stop).
- **`collapsed`**: single-line row; send/stop sits inline with the textarea, no upload button; text scrolls horizontally instead of wrapping.
- Toggling `collapsed` animates.
- Textarea growth (and the card with it) is capped by `--swc-prompt-field-max-block-size` (default `40vh`, or `40cqb` inside a sized container). Optional `min-rows`/`max-rows` add a row-count cap on top.
- Border radius is a constant 24px in both layouts (no exact token match), matching the reference implementation.

Branding colors (the `--brand` OKLCH gradient system) are intentionally out of scope for this PR.

https://github.com/user-attachments/assets/722e14ed-0416-48e5-8de1-ba89a88d08c6

## Motivation and context

Bringing the AI Prompt Field branding/behavior over to SWC; this PR is the first piece — the expanded/collapsed layout toggle — split out from the branding-colors work.

## Related issue(s)

- N/A (design reference: [Figma AI Prompt Field](https://www.figma.com/design/t4tOkjfwUW7zxoPAcQKmwU/AI-Prompt-Field?node-id=21562-23150))

## Screenshots (if appropriate)

Expanded vs. collapsed (see `Layout` story under Options):

| Expanded (default) | Collapsed |
|---|---|
| Card with label, textarea, and action bar (upload + send) | Single-line row, send button inline |

## Author's checklist

- [x] I have read the **CONTRIBUTING** and **PULL_REQUESTS** documents.
- [x] I have reviewed the Accessibility Practices for this feature.
- [x] I have added automated tests to cover my changes.
- [ ] I have included a well-written changeset if my change needs to be published.
- [x] I have included updated documentation if my change required it.

## Reviewer's checklist

- [ ] Includes a Github Issue with appropriate flag or Jira ticket number without a link
- [ ] Includes thoughtfully written changeset if changes suggested include `patch`, `minor`, or `major` features
- [ ] Automated tests cover all use cases and follow best practices for writing
- [ ] Validated on all supported browsers
- [ ] All VRTs are approved before the author can update Golden Hash

### Manual review test cases

- [ ] _Toggle collapsed layout_
[...]

PR #6452, opened 2026-06-25
feat(conversational-ai): improve artifact previews and scrolling

TarunAdobe:
## Description

This PR improves upload artifacts and their integration with Prompt Field:

- Aligns card and media artifact layouts with the approved Figma designs.
- Adds an optional `badge` slot for media previews such as document-type labels.
- Improves media preview sizing, cropping, and Storybook thumbnails.
- Adds a horizontal multi-artifact carousel to Prompt Field.
- Adds overflow chevrons, edge fades, set-based paging, and final-set alignment.
- Supports trackpad, wheel, touch, and draggable overlay-scrollbar interaction.
- Reserves the scrollbar lane to prevent the composer height from shifting.
- Warns developers when card and media layouts are mixed in one artifact strip.
- Warns when required Legal-approved disclaimer content is missing.
- Updates Storybook examples to use the design-approved AI disclaimer and official [AI User Guidelines](https://www.adobe.com/legal/licenses-terms/adobe-gen-ai-user-guidelines.html).
- Reworks the artifact strip's keyboard model: Tab reaches the first tile directly with no separate entry step, a second Tab reveals that tile's Close button, Arrow keys rove between tiles, and activating a chevron keeps focus on the chevron while paging updates which tile Tab or Shift+Tab lands on next.
- Fixes a single artifact's Close button being unreachable by Tab.
- Centers each artifact's dismiss badge on the tile's visible rounded corner, matching the approved corner-badge treatment for both card and media layouts.
- Moves Prompt Field's keyboard focus ring from the whole composer card to the textarea itself, so the indicator matches the actual focused element.
- Corrects the spacing between the artifact strip and the textarea to the specified 8px.
- Expands tests and documentation for artifact layouts, scrolling, badges, events, warnings, and keyboard behavior.

## Motivation and context

Prompt Field previously had limited support for multiple attachments, while Upload Artifact did not fully match the approved card and media treatments. This made larger upload sets difficult to navigate and encouraged mixed card/media layouts that did not produce a consistent attachment strip. Media artifacts also needed a way to display document-type context without using the metadata-first card layout. The artifact strip's keyboard model also required an explicit "enter" step before Tab or Arrow keys did anything, which added friction and didn't match how a roving-tabindex composite should behave. The updated implementation provides a consistent carousel for multiple artifacts, supports media badges, follows a conventional roving-tabindex keyboard model, and documents how applications should normalize mixed file types into one visual layout.

## Acceptance criteria

[...]

PR #6610, opened 2026-08-12
feat(prompt-field): AI brand treatment

rubencarvalho:
## Summary

Restart of the AI brand treatment for `swc-prompt-field`, built on top of #6585.

- Adds the outer ring, hue-sweep, gloss, and inset-shadow decorative layers behind a new `variant` property (`subtle` | `balanced` | `prominent`), following the reference implementation's DOM/layering structure.
- Wires real brand-derived OKLCH colors for all three variants, across every state combination (idle, hover, generating, focus, and their combinations), all derived from a single exposed `--swc-prompt-field-brand-color` custom property.
- Fixes several real bugs found and verified against the reference implementation's own live-rendered computed styles along the way (`@property` initial-value scoping, a Chromium `:host():not()` matching bug, shadow-layer misalignment, an opacity token mismatch, and a trailing-comma bug that collapsed `box-shadow` to `none`).
- Internal-only custom properties (derived color stops, opacities, shadows/glows, hue-sweep rotation/translation) follow the `--_swc-*` private property convention; only the brand-color anchor and JS-set passthrough sizing properties stay exposed.
- Keeps the branch in sync with #6585 as it continues to evolve (artifact scroll rework, JS textarea auto-grow fallback, etc.), with minor spacing fixes for the artifact strip along the way.

Working notes and the full bug-by-bug history are in `research.md` at the worktree root (gitignored, not part of this PR).

### Storybook links (variant × state)

Playground with `variant` and `generating` set via URL args:

| Variant | Idle | Generating |
| --- | --- | --- |
| Subtle | [link](https://swcpreviews.z13.web.core.windows.net/pr-6610/docs/gen2-storybook-dev/?path=/story/patterns-conversational-ai-prompt-field--playground&args=variant:subtle) | [link](https://swcpreviews.z13.web.core.windows.net/pr-6610/docs/gen2-storybook-dev/?path=/story/patterns-conversational-ai-prompt-field--playground&args=variant:subtle;generating:!true) |
| Balanced | [link](https://swcpreviews.z13.web.core.windows.net/pr-6610/docs/gen2-storybook-dev/?path=/story/patterns-conversational-ai-prompt-field--playground&args=variant:balanced) | [link](https://swcpreviews.z13.web.core.windows.net/pr-6610/docs/gen2-storybook-dev/?path=/story/patterns-conversational-ai-prompt-field--playground&args=variant:balanced;generating:!true) |
| Prominent | [link](https://swcpreviews.z13.web.core.windows.net/pr-6610/docs/gen2-storybook-dev/?path=/story/patterns-conversational-ai-prompt-field--playground&args=variant:prominent) | [link](https://swcpreviews.z13.web.core.windows.net/pr-6610/docs/gen2-storybook-dev/?path=/story/patterns-conversational-ai-prompt-field--playground&args=variant:prominent;generating:!true) |

## Test plan

- [ ] Verify all three variants (`subtle`, `balanced`, `prominent`) render correctly across idle, hover, generating, and focus states in Storybook
- [ ] Verify dark mode rendering for all three variants
- [ ] Verify `prefers-reduced-motion: reduce` disables the hue-sweep animation and transitions
- [ ] Cross-check against the reference implementation (react-spectrum's `@react-spectrum/ai` `PromptField`) side by side for each variant
- [ ] Accessibility pass (no regressions from the new decorative layers, which are all `aria-hidden`)


5t3ph:
1. Focus state - either move focus ring to around the element, or potentially the cursor change/blinking is sufficient by itself. React seems to only rely on the cursor change when focused, same for mouse or keyboard.
image

2. Hover state background of action button especially when field is not focused - maybe a semi-transparent darker color to mesh better with the gradient? React does a calculation with CSS relative color to basically end up with a that kind of value.
image
React's calcluated background:
image

Similar calc for the disabled state background of the prompt button:
image

3. Do you want to include refining the styles for the `legal` slot as part of this work? It needs to be a different color, including the link, which is a little tricky since it may need a global style or forced inline-style from the JS.
image

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/PromptField.ts
```diff
@@ -1022,65 +1028,75 @@ export class PromptField extends SpectrumElement {
...
-        <div class="swc-PromptField-box">
-          <div
-            class="swc-PromptField-input-area${hasArtifacts
-              ? ' has-artifact'
-              : ''}"
-          >
-            ${this._renderArtifact()}
+        <div class="swc-PromptField-outer-border">
+          <div class="swc-PromptField-box">
+            <span class="swc-PromptField-hue-sweep" aria-hidden="true"></span>
```
5t3ph: `aria-hidden="true"` is excessive here since these are just plain spans containing no content and they don't have a role that would be picked up in the a11y tree anyway.

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/PromptField.ts
```diff
@@ -1022,65 +1028,75 @@ export class PromptField extends SpectrumElement {
...
+            <span class="swc-PromptField-gloss" aria-hidden="true"></span>
             <span
-              id=${this.labelId}
-              class="swc-PromptField-label swc-VisuallyHidden"
+              class="swc-PromptField-inset-shadow"
```
5t3ph: Curious why this has to be a separate element instead of included in the list for `swc-PromptField-box`?

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/prompt-field.css
```diff
@@ -10,9 +10,31 @@
...
+  Every color above derives from --swc-prompt-field-brand-color (one OKLCH
+  hue) via oklch(from var(--swc-prompt-field-brand-color) ...); retheming
+  means overriding that single property.
+*/
+
 :host {
   display: block;
   inline-size: 100%;
+
+  /* @property can't yet be scoped to a shadow root; see https://shadow-dom-css.adobe.com/features/at-property/. */
```
5t3ph: If you need `@property` (I assume to animate colors?) you could add that typing to the global `swc.css`

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/stories/prompt-field.stories.ts
5t3ph, line 49: The value for the brand color isn't getting passed in via Storybook for some reason

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/prompt-field.css
```diff
@@ -10,9 +10,31 @@
  * governing permissions and limitations under the License.
  */
 
+/*
+  .swc-PromptField-outer-border  (padded gradient ring)
```
5t3ph, line 14: React seems to be accomplishing this with just one extra wrapper rather than the nested extra elements. Could we look into simplifying as well?

5t3ph:
One more ui/ux thing: we should pass on clicks on the prompt surface that aren't on other actions to activating the textfield, especially since we have significant surface space and no visible textarea boundary, so it's not otherwise clear how precise you have to be to enter the textfield (a potential barrier to someone with motor impairments). React supports this behavior, as do other prompt surfaces like for Firefly, ChatGPT, and Claude. They hint this is possible by changing to the `text` cursor in eligible surface areas.

Example from React:
https://github.com/user-attachments/assets/418688fd-7729-4b9f-967e-3f6e98e5cb18

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/stories/prompt-field.stories.ts
rubencarvalho: Ye - variants/generating is also not working... 🤔 I am looking into SB thingies!

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/PromptField.ts
rubencarvalho: Yes, good catch! This needs cleanup, it underwent lots of iterations 😅

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/PromptField.ts
rubencarvalho: Let me look into it! I was trying to rebuild Design's Figma "layer decomposition" example, but maybe we can collapse some of these.

rubencarvalho:
@5t3ph , thank you so much for the review!!

> 1. Focus state - either move focus ring to around the element, or potentially the cursor change/blinking is sufficient by itself. React seems to only rely on the cursor change when focused, same for mouse or keyboard.

This got introduced by https://github.com/adobe/spectrum-web-components/pull/6452, which introduced it after Design asked for it. Similarly to your suggestion, I also pushed back but did not want to block the merge, therefore I consulted with Design so we could get a final say here, because I am not on board with the current UI at all.

> 2. Hover state background of action button especially when field is not focused - maybe a semi-transparent darker color to mesh better with the gradient? React does a calculation with CSS relative color to basically end up with a that kind of value.

Yop! I have started a separate (stacked) PR because all of these buttons are also not using our own `swc-button`/`swc-action-button`, therefore I will do this update there too!

> 3. Do you want to include refining the styles for the `legal` slot as part of this work? It needs to be a different color, including the link, which is a little tricky since it may need a global style or forced inline-style from the JS.

Good call - I will create (another) stacked PR for this!

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/PromptField.ts
rubencarvalho: sooooo, I played around with this and here are my findings: the hue sweep needs its own animation (it rotates/translates/scales) and slightly extends past the card, so it can't really be folded into other elements... inset-shadow could _technically_ become part of `.box`'s own box-shadow, but that flipped its stacking order to behind the content instead of in front. I created pseudo-elements instead (and needed to adapt with some z-indexes). LMK if that looks better or if I should revert it. I did add a VRT so I could capture the before/after!

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/test/vrt/prompt-field.vrt.ts
```diff
@@ -0,0 +1,83 @@
...
+  <swc-prompt-field
+    label="Prompt"
+    variant=${variant}
+    ?generating=${generating}
+    value="Summarize the API changes in this branch."
+  >
+    ${legalDisclaimerSlot}
+  </swc-prompt-field>
+`;
+
```
5t3ph, line 277: Since you went ahead and updated the button colors, maybe add hover/focus states so you can compare better in your work to update to use actual swc button components?

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/stories/prompt-field.stories.ts
5t3ph: I believe it's because you're not using the `template` solution and handling the element and it's properties manually, so you'd have to pass in that custom property here as well if there's a reason you can't switch to using `template`.

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/prompt-field.css
```diff
@@ -38,19 +55,434 @@
...
   border-radius: 24px;
-  box-shadow: token("drop-shadow-emphasized-default-x") token("drop-shadow-emphasized-default-y") token("drop-shadow-emphasized-default-blur") token("drop-shadow-emphasized-default-color");
-  overflow: hidden;
+  box-shadow:
+    0 20px 20px -24px var(--_swc-prompt-field-prominent-outer-glow-color, transparent),
+    0 -3px 10px 1px var(--_swc-prompt-field-drop-shadow-color);
+  overflow: clip;
+  cursor: text;
+  transition:
+    --_swc-prompt-field-con-hue-opacity 700ms cubic-bezier(0.32, 0.72, 0, 1),
```
5t3ph: [nit] hold these duplicate timing and timing functions in a private cp

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/prompt-field.css
```diff
@@ -38,19 +55,434 @@
...
+    --_swc-prompt-field-bg-stop-1 700ms cubic-bezier(0.32, 0.72, 0, 1),
+    --_swc-prompt-field-bg-stop-2 700ms cubic-bezier(0.32, 0.72, 0, 1),
+    --_swc-prompt-field-bg-stop-3 700ms cubic-bezier(0.32, 0.72, 0, 1),
+    --_swc-prompt-field-bg-stop-4 700ms cubic-bezier(0.32, 0.72, 0, 1),
+    box-shadow 700ms cubic-bezier(0.32, 0.72, 0, 1);
+
+  --_swc-prompt-field-border-color: light-dark(oklch(from var(--swc-prompt-field-brand-color) 73.58% 0.2376 h / 10%), rgb(255 255 255 / 5%));
```
5t3ph: [nit] move these cp definitions to the top of the rule

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/stories/prompt-field.stories.ts
5t3ph, line 48: [nit] Can we use perhaps the `flexLayout: 'column-center'`  for purposes of at least the playground story? Full-width and at the top makes the perception at odds against "real-world" and I keep shrinking my viewport to compensate 😆

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/prompt-field.css
5t3ph: Updates here seem successful!

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/prompt-field.css
```diff
@@ -38,19 +55,434 @@
...
 
 .swc-PromptField-box:has(.swc-PromptField-input-area.has-artifact) {
   padding-block-start: token("spacing-100");
 }
 
+/* ─────────────────────────────────────
+   AI brand treatment: variables
+   ───────────────────────────────────── */
+
+/* --swc-prompt-field-brand-color's @property is registered in swc.css, not here -- see that file for why. */
```
5t3ph: Per my understanding of current support, these need to move to swc.css as well to have an effect.

- [my tests](https://codepen.io/spectrum-css/pen/emYRBzG)
- [WPT](https://wpt.fyi/results/css/css-properties-values-api/at-property-shadow.html?label=master&label=stable&aligned)

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/prompt-field.css
```diff
@@ -38,19 +55,434 @@
...
+    --_swc-prompt-field-outer-hue: light-dark(oklch(from oklch(from var(--swc-prompt-field-brand-color) 67.5% 0.2134 calc(h - 292.3) / 50%) l c h / 40%), oklch(from oklch(from var(--swc-prompt-field-brand-color) 23.87% 0.1407 calc(h - 38.8)) l c h / 40%));
+    --_swc-prompt-field-outer-stop-1: light-dark(oklch(from oklch(from var(--swc-prompt-field-brand-color) 73.58% 0.2376 h / 60%) l c h / 4%), oklch(from oklch(from var(--swc-prompt-field-brand-color) 23.18% 0.1116 calc(h - 3.1)) l c h / 40%));
+    --_swc-prompt-field-outer-stop-2: light-dark(oklch(from oklch(from var(--swc-prompt-field-brand-color) 67.5% 0.2134 calc(h - 292.3) / 50%) l c h / 4%), oklch(from oklch(from var(--swc-prompt-field-brand-color) 23.87% 0.1407 calc(h - 38.8)) l c h / 40%));
+    --_swc-prompt-field-outer-stop-3: light-dark(oklch(from oklch(from var(--swc-prompt-field-brand-color) 78.78% 0.1692 calc(h - 254) / 40%) l c h / 4%), oklch(from oklch(from var(--swc-prompt-field-brand-color) 23.87% 0.1407 calc(h - 38.8)) l c h / 40%));
+  }
+
+  /* :not() inside :host(), :has() outside it -- both fail the other way in Chromium. */
+  :host(:not([generating])) &:has(.swc-PromptField-input-area:focus-within) {
```
5t3ph: I think this selector is failing in Firefox (on the right) and possibly something else as well. It might be an issue with nesting support.

https://github.com/user-attachments/assets/863804d2-195e-4d70-9f3d-6eb97e241cf6

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/prompt-field.css
```diff
@@ -38,19 +55,434 @@
...
+  }
+
+  100% {
+    scale: 1 1;
+  }
+}
+
+:host([variant="subtle"]:not([generating])) .swc-PromptField-box {
+  --_swc-prompt-field-con-hue-opacity: 0%;
+  --_swc-prompt-field-bg-stop-1: light-dark(white, token("gray-75"));
```
5t3ph: Can replace these custom light-dark with `token("background-layer-2-color")` (and hold in a private property since duplicated, including in later rules too)

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/prompt-field.css
```diff
@@ -38,19 +55,434 @@
...
+  transition:
+    --_swc-prompt-field-outer-stop-1 700ms cubic-bezier(0.32, 0.72, 0, 1),
+    --_swc-prompt-field-outer-stop-2 700ms cubic-bezier(0.32, 0.72, 0, 1),
+    --_swc-prompt-field-outer-stop-3 700ms cubic-bezier(0.32, 0.72, 0, 1),
+    box-shadow 700ms cubic-bezier(0.32, 0.72, 0, 1);
+
+  :host([variant="subtle"]) & {
+    --_swc-prompt-field-outer-shadow: rgb(0 0 0 / 5%);
+    --_swc-prompt-field-outer-hue: transparent;
+    --_swc-prompt-field-outer-stop-1: light-dark(rgb(255 255 255), rgb(255 255 255 / 2%));
```
5t3ph: Hold these duplicates in private cp, especially since this same light-dark value is used again in other rules.

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/prompt-field.css
```diff
@@ -608,3 +1034,10 @@
     scroll-behavior: auto;
   }
 }
+
+/* Idle focus otherwise relies on the textarea's own cursor as the indicator (no ring); prefers-contrast: more gets an explicit ring around the card instead. */
+@media (prefers-contrast: more) {
+  :host .swc-PromptField-box:has(.swc-PromptField-input-area:focus-within) {
+    outline: token("focus-indicator-thickness") solid token("gray-900");
+  }
+}
```
5t3ph: Add to resolve forced-colors losing the control boundary, also benefiting those wanting more contrast:

Image

```suggestion
@media (prefers-contrast: more) or (forced-colors: active) {
  :host .swc-PromptField-box:has(.swc-PromptField-input-area:focus-within) {
    outline: token("focus-indicator-thickness") solid token("gray-900");
  }

  /* Will become opaque border for forced-colors */
  .swc-PromptField-outer-border {
    border: 2px solid color-mix(in oklch, var(--swc-prompt-field-brand-color), transparent 65%);
  }
}
```

5t3ph:
I know you're going to be swapping to SWC buttons but keep an eye on this - the focus ring is currently clipped in Firefox and Safari, which is technically a WCAG failure. Presently fine in Chromium. We may just need a little container padding to accommodate, TBD after you swap.

image

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/prompt-field.css
```diff
@@ -37,20 +56,388 @@
...
-  border: 1px solid transparent;
+  background: radial-gradient(circle at right bottom in oklch, var(--_swc-prompt-field-bg-stop-1) 0%, var(--_swc-prompt-field-bg-stop-2) 35%, var(--_swc-prompt-field-bg-stop-3) 82%, var(--_swc-prompt-field-bg-stop-4) 100%);
   border-radius: 24px;
-  box-shadow: token("drop-shadow-emphasized-default-x") token("drop-shadow-emphasized-default-y") token("drop-shadow-emphasized-default-blur") token("drop-shadow-emphasized-default-color");
-  overflow: hidden;
+  box-shadow:
+    0 20px 20px -24px var(--_swc-prompt-field-prominent-outer-glow-color, transparent),
+    0 -3px 10px 1px var(--_swc-prompt-field-drop-shadow-color);
+  overflow: clip;
+  cursor: text;
```
5t3ph, line 77: When disabled, cursor needs updated to `default` (matches native disabled input behavior)

5t3ph (commented):
One little thing plus the still open comment from before on considering adding states into the VRT.

rubencarvalho:
> I know you're going to be swapping to SWC buttons but keep an eye on this - the focus ring is currently clipped in Firefox and Safari, which is technically a WCAG failure. Presently fine in Chromium. We may just need a little container padding to accommodate, TBD after you swap.

@5t3ph Fixed!

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/PromptField.ts
```diff
@@ -279,6 +285,28 @@ export class PromptField extends SpectrumElement {
...
+      this._textarea?.focus();
+    }
+  }
+
+  private _isFocusableOrSlotted(element: Element): boolean {
+    return element.matches(
+      'button, a[href], input, textarea, select, [tabindex], slot'
+    );
+  }
+
```
TarunAdobe, line 308: why not use focusable selector from utils?

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/prompt-field.mdx
```diff
@@ -47,7 +47,7 @@ The textarea grows with content up to `--swc-prompt-field-max-block-size` (defau
 
 Every prompt field in production must include Legal-approved disclaimer copy below the composer. Slot your product's approved text into **`legal`**: wording, links, and locale are consumer-owned, not baked into the component.
 
-Apply the `.swc-PromptField-legal-disclaimer` class for default footer typography. `swc-prompt-field` logs a development warning when the `legal` slot is empty.
+Slotted `legal` content inherits the footer typography automatically. When it contains links, add the `swc-Typography--links` class to the wrapping element so the links inherit that typography and render in the subdued neutral color instead of the default accent blue. `swc-prompt-field` logs a development warning when the `legal` slot is empty.
 
 ### Artifact
```
TarunAdobe, line 52: why no mention of variant in doc anywhere?

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/PromptField.ts
rubencarvalho: done!

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/prompt-field.mdx
rubencarvalho: Done!

5t3ph (approved):
LGTM!

TarunAdobe (approved):
GG

PR #6620, opened 2026-08-13
chore(status-light): add VRT coverage

cdransf:
### Description

Adds dedicated `test/vrt/` coverage for `status-light`, following the `button`/`badge` VRT pattern: `status-light.vrt.ts` covers every variant (5 semantic + 14 non-semantic) × size, grouped into one row per variant, plus a `Wrapping` row and a `CJK language` row (status-light.css sets a distinct line-height for `:lang(ja/zh/ko)`); `ForcedColors` reuses the full permutation set since `forced-color-adjust: none` is set on the whole component, so every variant's color survives forced-colors mode rather than collapsing to a system color. `status-light-custom-properties.vrt.ts` covers reference/override rows for all six documented custom properties (`dot-size`, `dot-color`, `font-size`, `line-height`, `text-to-visual`, `content-color`), with `verifyCustomPropertyCoverage()` checked against the manifest. No global-styles VRT file is included since `status-light` has no plain-class global stylesheet.

### Motivation and context

Closes the VRT coverage gap for `status-light` called out in SWC-2410.

### Related issue(s)

- Jira: SWC-2410

### Author's checklist

### Accessibility testing checklist

- [ ] **Keyboard**: N/A. Visual-only test coverage; status-light is non-interactive with no focusable parts and no interactive changes.
- [ ] **Screen reader**: No accessibility-relevant changes. The component's existing semantics are untouched by this test-only addition.

### DIFF
--- a/2nd-gen/packages/swc/components/status-light/status-light.css
+++ b/2nd-gen/packages/swc/components/status-light/status-light.css
@@ -155,12 +155,6 @@
 @media (forced-colors: active) {
   .swc-StatusLight {
     --swc-status-light-content-color: CanvasText;
-
-    forced-color-adjust: none;
-
-    &::before {
-      background-color: CanvasText;
-      border: token("border-width-100") solid;
-    }
+    --swc-status-light-dot-color: CanvasText;
   }
 }
--- a/2nd-gen/packages/swc/components/status-light/status-light.css
+++ b/2nd-gen/packages/swc/components/status-light/status-light.css
@@ -159,6 +159,7 @@
     forced-color-adjust: none;
 
     &::before {
+      background-color: CanvasText;
       border: token("border-width-100") solid;
     }
   }
--- a/2nd-gen/packages/swc/components/status-light/test/vrt/status-light-custom-properties.vrt.ts
+++ b/2nd-gen/packages/swc/components/status-light/test/vrt/status-light-custom-properties.vrt.ts
@@ -1,0 +1,94 @@
+/**
+ * Copyright 2026 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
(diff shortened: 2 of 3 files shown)


rise-erpelding (approved):
LGTM

5t3ph (commented):
For the same reason we don't keep button or badge colors for forced-colors, I don't think we should be keeping them for status light. Reason being that relying on color alone to communicate is a WCAG failure, and we also can't know the intent of those colors out of context.

Can you update the dots to use CanvasText as the `background-color` for forced-colors mode?

Maybe we also need a ticket to investigate some type of like "retain color" opt-in switch so that contextually colors for status light and maybe badge can be kept for forced-colors, or if we just need to add a demo in our customization docs for how to handle that perhaps.

cdransf:
> Can you update the dots to use CanvasText as the `background-color` for forced-colors mode?

That makes sense (and was something I hadn't considered). I revised the comments and added the `background-color` change in 134f728e9e64e0c24b8c905e8b4620b64621e251. Added SWC-2508 for the retain color research too. ✨

2nd-gen/packages/swc/components/status-light/status-light.css
```diff
@@ -159,6 +159,7 @@
     forced-color-adjust: none;
```
5t3ph: Also remove this `forced-color-adjust: none` since we don't want that behavior.

2nd-gen/packages/swc/components/status-light/status-light.css
```diff
@@ -159,6 +159,7 @@
...
 
     &::before {
+      background-color: CanvasText;
```
5t3ph: Can be applied in the existing `.swc-StatusLight` selector using the `--swc-status-light-dot-color`.

Also, since we're filling in the background, we don't need the border anymore.

5t3ph (commented):
A little more simplification can be done for the forced-colors.

2nd-gen/packages/swc/components/status-light/status-light.css
cdransf: Gotcha! Updated in 4c1e3b8814f9fd60c2e9244fba0b4f7ac3208e88 as well. ✨

2nd-gen/packages/swc/components/status-light/status-light.css
cdransf: Ah, thank you! Updated. 4c1e3b8814f9fd60c2e9244fba0b4f7ac3208e88 ✨

5t3ph (approved):
:shipit:

PR #6518, opened 2026-07-16
docs(2nd-gen): fix inaccurate API tables for several components

rubencarvalho:
## Description

Adds missing `@cssprop` JSDoc entries and property descriptions so the generated Custom Elements Manifest (and the Storybook API tables rendered from it) accurately reflect each component's public CSS custom property and property surface. Also fixes the Tabs docs page to render API tables for all three of its elements, not just the root one.

- `tabs`/`tab`/`tab-panel`: 6 missing `--swc-tab*` custom properties (the gap that started this pass)
- `action-button`, `button`: `--swc-*-down-state-transform` (button also gets `--swc-button-max-inline-size`)
- `close-button`: `--swc-close-button-focus-indicator-thickness` / `-gap`
- `illustrated-message`: 10 missing custom properties (illustration sizing/color, heading/description typography) — had zero documented before
- `button`, `action-button`, `close-button` (via shared `ButtonBase`), `status-light`, `badge`: missing `size`/`variant` property descriptions
- `tabs`: the docs page only ever rendered `swc-tabs`'s API table via `<DocsFooter />`; `swc-tab` and `swc-tab-panel` (the component's other two elements — including the CSS custom properties added here) never appeared. Wired up `additionalApiTables`, the same mechanism Accordion already uses for its sibling element.

## Motivation and context

Noticed `swc-tabs` wasn't documenting its CSS custom properties in the Storybook API table. Audited all 23 2nd-gen `swc/components/*` against their actual CSS (including shared `_lit-styles` includes) to find every component with the same gap, plus cross-checked the full generated manifest for any other missing/blank properties, slots, or events. While verifying the Tabs fix in a live Storybook, found the `swc-tab`/`swc-tab-panel` API tables were missing entirely from the docs page, not just their CSS props — fixed that too.

A few related findings were deliberately **not** changed here since they need a naming-convention decision, not a doc fix (see the style guide's [exposed vs. internal](CONTRIBUTOR-DOCS/02_style-guide/01_css/02_custom-properties.md) rules): `--swc-color-handle-picked-color`, `--swc-color-loupe-picked-color`, and `--swc-placement-available-width/-height` (popover/tooltip) are JS-computed passthroughs with no fallback that look like they should be `--_swc-*` (internal) rather than exposed — happy to open a follow-up if wanted.

## Related issue(s)

- fixes N/A (documentation accuracy pass, no tracked issue)

## Author's checklist

- [x] I have read the **CONTRIBUTING** and **PULL_REQUESTS** documents.
[...]

PR #6621, opened 2026-08-14
chore(tabs): add VRT coverage

cdransf:
## Description

Adds `2nd-gen/packages/swc/components/tabs/test/vrt/tabs.vrt.ts` and `tabs-custom-properties.vrt.ts`, providing dedicated Chromatic VRT coverage for the `tabs` three-element component (`swc-tabs`, `swc-tab`, `swc-tab-panel`):

- `Permutations`: direction (horizontal/vertical) x density (regular/compact), individual tab states (default/selected/disabled), a fully disabled container, anatomy (text-only/icon+text/icon-only), forced hover/focus-visible/active on individual tabs, and CJK tab labels (ja/ko/zh).
- `ForcedColors`: the same permutation set under forced-colors emulation, covering the forced-colors overrides defined on all three parts of the architecture.
- `CustomProperties`: one reference/override row per documented custom property, split into two groups (Tabs, Tab) and verified against the manifest for each of the two declarations independently. `TabPanel` has no documented `@cssprop` (the `--swc-tab-panel-padding-top` its CSS uses isn't part of its public API contract), so it has no custom-properties row here - that's expected, not a gap.

This PR touches only files under `components/tabs/test/vrt/` - no component source and no shared `.storybook/helpers` changes. (An earlier version of this PR also documented a missing `@cssprop` on `TabPanel.ts` and re-added a shared VRT caption helper; both were reverted to keep this change scoped purely to test coverage for `tabs`, per review discussion.)

## Motivation and context

The `tabs` component had no dedicated VRT coverage. This follows the VRT foundation established for `button` (PR #6463), extending it to `tabs` per the linked ticket.

The ticket's acceptance criteria asked for a "quiet variant" snapshot, but `swc-tabs` has no `quiet` property - it was explicitly removed in the 2nd-gen migration (`migration-guide.mdx`: "No replacement; not part of Spectrum 2 surface"). Substituted `density` (regular/compact) instead, since that's the actual replacement axis for the removed quiet treatment.

## Related issue(s)

- fixes SWC-2414

N/A - test-only change, no visual/behavioral change to the components themselves. Manually verified via Playwright screenshots of the built Storybook (including real `forced-colors` media emulation, not just the inert Chromatic parameter) during development; a nesting bug in the custom-properties layout was caught and fixed this way before landing.

## Author's checklist

## Reviewer's checklist

### Manual review test cases

- [ ] _Tabs VRT stories render correctly_
  1. Go to Storybook > Components > Tabs > Tabs VRT
  2. Review the `Permutations`, `ForcedColors`, and `CustomProperties` stories
  3. Expect direction/density/state/anatomy/forced-state/CJK combinations to render correctly in both light/ltr and dark/rtl, with the selection indicator correctly positioned under the selected tab in every group

### Device review

- [ ] Did it pass in Desktop?

- [ ] **Keyboard**
  These are non-interactive VRT snapshot fixtures (tagged `dev`, excluded from the interactive Storybook sidebar in normal use); no new keyboard behavior is introduced. Confirm no regressions in the existing `tabs.stories.ts` docs examples' keyboard navigation (arrow keys, Home/End, Enter/Space activation).

- [ ] **Screen reader**
  1. Go to Storybook > Components > Tabs > Tabs VRT > Permutations
  2. Inspect the icon-only anatomy group and the individual tab-states group with a screen reader or the accessibility tree
  3. Expect icon-only tabs to announce their `aria-label` (e.g. "Dashboard"), the disabled tab to announce as unavailable, and the selected tab to announce `aria-selected="true"`

### DIFF
--- a/2nd-gen/packages/swc/components/tabs/tab.css
+++ b/2nd-gen/packages/swc/components/tabs/tab.css
@@ -45,6 +45,12 @@
   font-size: token("font-size-100");
   font-weight: token("regular-font-weight");
   line-height: token("line-height-100");
+
+  &:lang(ja),
+  &:lang(zh),
+  &:lang(ko) {
+    line-height: token("cjk-line-height-100");
+  }
 }
 
 /* ── Icon slot ─────────────────────────────────────────── */
--- a/2nd-gen/packages/swc/components/tabs/test/vrt/tabs-custom-properties.vrt.ts
+++ b/2nd-gen/packages/swc/components/tabs/test/vrt/tabs-custom-properties.vrt.ts
@@ -45,8 +45,8 @@ export default meta;
 // into two groups (Tabs, Tab) since the three-element architecture documents
 // custom properties on separate classes (Tabs.ts, Tab.ts), each verified
 // against its own manifest entry below. TabPanel.ts has no documented
-// `@cssprop` (the `--swc-tab-panel-padding-top` its CSS uses isn't part of
-// its public API contract), so there's nothing to verify coverage for there.
+// `@cssprop` (its CSS only uses the private `--_swc-tab-panel-padding-top`),
+// so there's nothing to verify coverage for there.
 
 const TABS_CUSTOM_PROPERTY_CASES: readonly CustomPropertyCase<`--swc-tabs-${string}`>[] =
   [{ property: '--swc-tabs-indicator-color', value: 'magenta' }];
--- a/2nd-gen/packages/swc/components/tabs/test/vrt/tabs-custom-properties.vrt.ts
+++ b/2nd-gen/packages/swc/components/tabs/test/vrt/tabs-custom-properties.vrt.ts
@@ -1,0 +1,134 @@
+/**
+ * Copyright 2026 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import { html } from 'lit';
+import type { Meta, StoryObj as Story } from '@storybook/web-components';
+
+import '@adobe/spectrum-wc/components/tabs/swc-tabs.js';
+import '@adobe/spectrum-wc/components/tabs/swc-tab.js';
+import '@adobe/spectrum-wc/components/tabs/swc-tab-panel.js';
+
+import type { CustomPropertyCase } from '../../../../.storybook/helpers/index.js';
+import {
+  coveredCustomProperties,
+  customPropertyRows,
+  theme,
+  verifyCustomPropertyCoverage,
+  vrtParameters,
+} from '../../../../.storybook/helpers/index.js';
+import customElementsManifest from '../../../../dist/custom-elements.json';
+
+// Metadata
+
+const meta: Meta = {
+  title: 'Tabs/Tabs VRT',
+  component: 'swc-tabs',
+  tags: ['dev'],
+};
+
+export default meta;
+
+// Helpers
+
+// One row per public custom property: a reference element beside the same
+// element with that one property overridden to an obviously different
+// value, so a real difference confirms the override still applies. Split
+// into two groups (Tabs, Tab) since the three-element architecture documents
+// custom properties on separate classes (Tabs.ts, Tab.ts), each verified
+// against its own manifest entry below. TabPanel.ts has no documented
+// `@cssprop` (the `--swc-tab-panel-padding-top` its CSS uses isn't part of
+// its public API contract), so there's nothing to verify coverage for there.
+
+const TABS_CUSTOM_PROPERTY_CASES: readonly CustomPropertyCase<`--swc-tabs-${string}`>[] =
+  [{ property: '--swc-tabs-indicator-color', value: 'magenta' }];
+
+const renderTabsCase = (_testCase: CustomPropertyCase, style?: string) => html`
+  <swc-tabs
+    selected="1"
+    accessible-label="Product details"
+    style=${style ?? ''}
+  >
+    <swc-tab tab-id="1">Overview</swc-tab>
+    <swc-tab tab-id="2">Specifications</swc-tab>
+    <swc-tab-panel tab-id="1">
+      <p>Overview content.</p>
+    </swc-tab-panel>
+    <swc-tab-panel tab-id="2">
+      <p>Specifications content.</p>
+    </swc-tab-panel>
+  </swc-tabs>
+`;
+
+// Tab renders meaningfully as a standalone custom element (its template has
+// no dependency on a parent `swc-tabs`), so these properties are tested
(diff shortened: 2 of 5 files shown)


aramos-adobe (approved):
Super cool!

aramos-adobe:
My only comment is the forced states. Realistically, shouldn't the focus-visible state be `selected` when tabbed and the underline is visible?

Screenshot 2026-08-17 at 4 12 40 PM

2nd-gen/packages/swc/components/tabs/test/vrt/tabs-custom-properties.vrt.ts
```diff
@@ -0,0 +1,134 @@
...
+// Helpers
+
+// One row per public custom property: a reference element beside the same
+// element with that one property overridden to an obviously different
+// value, so a real difference confirms the override still applies. Split
+// into two groups (Tabs, Tab) since the three-element architecture documents
+// custom properties on separate classes (Tabs.ts, Tab.ts), each verified
+// against its own manifest entry below. TabPanel.ts has no documented
+// `@cssprop` (the `--swc-tab-panel-padding-top` its CSS uses isn't part of
+// its public API contract), so there's nothing to verify coverage for there.
```
rise-erpelding: 🤔 Why do we write it as `--swc-tab-panel-padding-top` if it's not part of the public API contract though? (I did check, and I see that [it does seem like exposing it was a mistake](https://github.com/adobe/spectrum-web-components/pull/6518#discussion_r3598550367).) Shouldn't we be writing it as `--_swc-tab-panel-padding-top` instead? (And using a token for that value instead of `12px`?)

2nd-gen/packages/swc/components/tabs/test/vrt/tabs.vrt.ts
```diff
@@ -0,0 +1,320 @@
...
+          ${tab({ id: '3', label: 'Active', forceState: 'active' })}
+        `,
+        panels: html`
+          ${panel('1', 'Hover content.')} ${panel('2', 'Focus content.')}
+          ${panel('3', 'Active content.')}
+        `,
+      }),
+    ],
+    'Forced states'
+  )}
```
rise-erpelding, line 266: What do you think about adding a first tab here, before the hovered one, that's just selected? I found this a little confusing at first comparing the hovered tab to the others, and thought that the selected underline was part of the hover state, but I see that text color is the only thing that indicates hover state.

Image

2nd-gen/packages/swc/components/tabs/test/vrt/tabs.vrt.ts
```diff
@@ -0,0 +1,320 @@
...
+          ${tab({ id: '1', label: '概览' })} ${tab({ id: '2', label: '规格' })}
+          ${tab({ id: '3', label: '准则' })}
+        `,
+        panels: html`
+          ${panel('1', '概览内容。')} ${panel('2', '规格内容。')}
+          ${panel('3', '准则内容。')}
+        `,
+      }),
+    ],
+    'CJK language'
```
rise-erpelding, line 303: I don't see any CJK selectors in the CSS, do we still need to test CJK?

(But I'm also wondering, usually are lang selectors are a simple line-height modification for CJK. Since we're setting `line-height-100` on `.swc-Tab-label` but we never set `cjk-line-height-100`, is that a bug or a deliberate design decision to not adjust line height for CJK? If it's a bug then perhaps it needs fixing, and then we would have a reason to keep the CJK VRTs.)

rise-erpelding (commented):
A few tiny questions/suggestions, but overall looks good!

2nd-gen/packages/swc/components/tabs/test/vrt/tabs.vrt.ts
cdransf: Ah, yeah! That totally makes sense. Added here :https://github.com/adobe/spectrum-web-components/pull/6621/commits/285921fa08d10c830f00c2e53d2fe96aa3deb003 ✨

cdransf:
> My only comment is the forced states. Realistically, shouldn't the focus-visible state be `selected` when tabbed and the underline is visible?

Ended up dropping in a change here https://github.com/adobe/spectrum-web-components/pull/6621/commits/285921fa08d10c830f00c2e53d2fe96aa3deb003 based on @rise-erpelding's [feedback that touched on the same spot in the code](https://github.com/adobe/spectrum-web-components/pull/6621#discussion_r3799313449). ✨

2nd-gen/packages/swc/components/tabs/test/vrt/tabs-custom-properties.vrt.ts
cdransf: Ohh good catch! Updated to `--_swc-tab-panel-padding-top` in https://github.com/adobe/spectrum-web-components/pull/6621/commits/d786a96e88486ca9667de1916017f3eb7d3dc636 ✨

2nd-gen/packages/swc/components/tabs/test/vrt/tabs.vrt.ts
cdransf: Good catch! I went ahead and added the `CJK` line-height modification here so we can keep the VRTs. https://github.com/adobe/spectrum-web-components/pull/6621/commits/9076e602de68124ef12bdf6d30e15ec14d60de83 ✨

2nd-gen/packages/swc/components/tabs/tab.css
rise-erpelding, line 1: These CSS changes look ok to me, do you want to add a changeset as well?

rise-erpelding:
> My only comment is the forced states. Realistically, shouldn't the focus-visible state be `selected` when tabbed and the underline is visible?
> 
> Screenshot 2026-08-17 at 4 12 40 PM

@aramos-adobe I think this is a valid concern, in a real use case focus-visible would always happen when a tab is also selected... @cdransf Would we want to change the forced states so that the first selected tab is actually a focus visible instead?

Claude _did_ bring up the concern that we're not testing selected + hover, but I think that may be overkill for this component, because the only hover effect is a text color change, and it's the same color change that selected gets. But could be a valid thing to look out for in other components!

rise-erpelding (commented):
Just found this comment that never got submitted as part of a review 🫤 

I also have a comment above continuing the conversation about the forced states that I don't think has been addressed yet.

2nd-gen/packages/swc/components/tabs/tab.css
cdransf: Good call! Just added that https://github.com/adobe/spectrum-web-components/pull/6621/commits/587fb7705c51dd3b9f7d40aa297e495255878b39 ✨

cdransf:
> > My only comment is the forced states. Realistically, shouldn't the focus-visible state be `selected` when tabbed and the underline is visible?
> > Screenshot 2026-08-17 at 4 12 40 PM
> 
> @aramos-adobe I think this is a valid concern, in a real use case focus-visible would always happen when a tab is also selected... @cdransf Would we want to change the forced states so that the first selected tab is actually a focus visible instead?
> 
> Claude _did_ bring up the concern that we're not testing selected + hover, but I think that may be overkill for this component, because the only hover effect is a text color change, and it's the same color change that selected gets. But could be a valid thing to look out for in other components!

Focus visible without select is a valid state before a user selects/hits enter or space, so I went ahead and left that. I _did_ add a focus visible + selected case so we can cover the overlap! (Selected + hover use the same token so I think we're good there.) https://github.com/adobe/spectrum-web-components/pull/6621/commits/a68cad0215809aff87062ef4d8c3b49d459660fe ✨

rise-erpelding (approved):
I like what you ended up adding (the focus-visible + selected) even better! Nice work!✨

PR #6622, opened 2026-08-14
feat(response-status): adding disclosure and content height constraints to agentic steps

aramos-adobe:
## Description

Adds two capabilities to the Response status pattern's agentic step timeline:

**Per-step disclosure.** Each `<swc-response-status-step>` that has a `slot="description"` now renders its own disclosure control (chevron + `aria-expanded`/`aria-controls`), nested inside the existing header-level panel disclosure. Steps expand and collapse independently. The active step auto-expands and auto-collapses again as it completes; a user can override any step by toggling it directly.

- New `open` attribute on `<swc-response-status-step>` sets a step's initial expanded state.
- New event **`swc-response-status-step-toggle`** (`detail: { open, index }`, `bubbles`, `composed`) fires when a user expands/collapses a step.
- The shared chevron rendering and collapse-panel mechanics are DRY'd between the header disclosure and the per-step disclosures.

**Content height constraint.** A step's description caps at a max height (`--swc-response-status-step-detail-max-block-size`, default `120px`). Content shorter than the cap keeps its natural height; taller content scrolls within the step and shows a bottom fade while there is more content below and you are not scrolled to the end.

- The fade is a CSS scroll-driven mask (`@supports (animation-timeline: scroll())`), so it is overflow-aware without JS: short descriptions never fade. It is disabled in forced-colors mode so every line stays legible.

Also adds `Step disclosure` and `Long description` stories, updates the pattern MDX (documents the new attribute, event, and custom property), and adds unit + accessibility test coverage.

> Note: the step `open` attribute sets the *initial* disclosure state only; the live per-step state is owned by the parent and reported via `swc-response-status-step-toggle` rather than reflected back onto the child.

## Motivation and context

Agentic responses can produce many execution steps, each with lengthy reasoning text. Users need to expand and collapse individual steps to focus on what matters, and a single long step description should not blow out the surrounding layout. Per-step disclosure plus a capped, scrollable description with a fade affordance addresses both.

## Related issue(s)

- fixes SWC-2479

[Figma Design](https://www.figma.com/design/t4tOkjfwUW7zxoPAcQKmwU/AI-Prompt-Field?m=dev&node-id=25783-37532)

## Author's checklist

## Reviewer's checklist

### Manual review test cases

- [ ] _Per-step disclosure expands and collapses independently_
  1. Go to Conversational AI/Response status → **Step disclosure** story.
  2. Click a collapsed step's title row; expect only that step's description to expand (chevron rotates, others unchanged).
  3. Click it again; expect it to collapse. Confirm a `swc-response-status-step-toggle` event fires with `{ open, index }`.

- [ ] _Active step auto-expands, then auto-collapses on completion_
  1. Go to the **Steps** story (has an `active` step).
  2. Expect the `active` step to render expanded and the completed steps collapsed by default.

- [ ] _Long description caps height and scrolls with a fade_
  1. Go to the **Long description** story (Chromium).
  2. Expect the long step's description to cap at ~120px, scroll internally, and show a bottom fade that disappears when scrolled to the end.
  3. Confirm the short step in the same story keeps its natural height with no fade.

### Device review

- [ ] Did it pass in Desktop?

- [ ] **Keyboard**
  1. Go to Conversational AI/Response status → **Steps** story.
  2. <kbd>Tab</kbd> to the header disclosure button; press <kbd>Enter</kbd>/<kbd>Space</kbd> and confirm the timeline panel toggles.
  3. <kbd>Tab</kbd> through the step disclosure buttons; press <kbd>Enter</kbd>/<kbd>Space</kbd> on a step and confirm only that step's description toggles.
  4. Confirm focus order is logical, the focus indicator is visible on every disclosure button, and there are no focus traps.
  5. In the **Long description** story, confirm the capped/scrolling region does not trap focus and remains keyboard reachable.

- [ ] **Screen reader**
  1. In the **Steps** story, confirm the header disclosure button announces its label and expanded/collapsed state (`aria-expanded`), and that `aria-controls` points at the panel.
  2. Confirm each step disclosure button announces its step title as its accessible name plus expanded/collapsed state, and toggling updates the announced state.
  3. Confirm the timeline exposes `role="group"` with the `accessible-label` name, and the step list uses `role="list"`/`role="listitem"`.
  4. Confirm no duplicate or spurious announcements when a step auto-expands/auto-collapses as its status changes.


miwha-adobe:
Should the content under the description also be tabable? 

https://github.com/user-attachments/assets/20bef185-b657-46ce-8f70-b7ab86e7508c

miwha-adobe (approved):
lgtm! Nice work!!

2nd-gen/packages/swc/patterns/conversational-ai/response-status/response-status.css
```diff
@@ -383,6 +384,53 @@
...
+  padding: 0;
+  color: inherit;
+  text-align: start;
+  background: transparent;
+  border: none;
+}
+
+/* Description caps at a max height; shorter content stays auto. Override the
+   custom property to tune the cap. */
+.swc-ResponseStatus-step-detailScroll {
```
5t3ph: Right now this visually is wider than the content area which has a capped width, causing any visible scrollbars to be distanced from the content.

Image

One way to resolve is `inline-size: fit-content` however that may cause the virtual scrollbar to appear over the interior content, so you might instead want to calc it with the value provided to step-detail plus 20px to account for typical scrollbar width cross-platform.

2nd-gen/packages/swc/patterns/conversational-ai/response-status/ResponseStatus.ts
```diff
@@ -149,15 +170,40 @@ export class ResponseStatus extends SpectrumElement {
...
 
   protected override willUpdate(_changed: PropertyValues<this>): void {
     this._applyLabelRoll();
   }
 
+  protected override updated(changed: PropertyValues<this>): void {
+    super.updated(changed);
+
+    // Re-measure when what is shown (or its open state) changes. Label-roll
+    // updates do not affect overflow, so they are intentionally excluded.
```
5t3ph: Maybe a bug to handle in another ticket, but these _should_ support wrapping because it is reasonable that they could overflow in a sidebar context, especially given the need to support translations:

Image

2nd-gen/packages/swc/patterns/conversational-ai/response-status/ResponseStatus.ts
```diff
@@ -436,19 +484,89 @@ export class ResponseStatus extends SpectrumElement {
...
+      if (!right.has(value)) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  // Measures which open step descriptions overflow their capped height so only
+  // those expose a focusable scroll region (see `_overflowingSteps`).
+  private _syncDetailOverflow(): void {
```
5t3ph: I understand the "why" but is there any way to do a perf analysis on this? My understanding is that measuring elements and scroll tracking can get intensive (which involves more than just this block). Maybe those measures could only happen if the status is also actually visible? You might be able to gate other things on visibility too (if not already) just to prevent some JS churn.

2nd-gen/packages/swc/patterns/conversational-ai/response-status/ResponseStatus.ts
```diff
@@ -584,6 +703,60 @@ export class ResponseStatus extends SpectrumElement {
...
+        <div class="swc-ResponseStatus-step-body">
+          <p class="swc-ResponseStatus-step-title swc-Detail swc-Detail--sizeS">
+            ${step.label}
+          </p>
+        </div>
+      `;
+    }
+
+    const open = this._isStepOpen(step, index);
+    const detailId = `${this.panelId}-detail-${index}`;
```
5t3ph: [nit] I don't believe this strictly needs a unique ID since each response status is in its own contained shadow DOM, unless I'm missing something on how this comes together.

5t3ph (commented):
Overall, looks good!

My Slack comment re: sharing some functionality with accordion can be addressed separately.

2nd-gen/packages/swc/patterns/conversational-ai/response-status/ResponseStatus.ts
aramos-adobe: @5t3ph I am going to save this for another ticket and get more clarity with design team if that's OK. You do raise a good point on this. Since this maybe an action button after migrating `swc-action-button` into this space.

5t3ph: Just to clarify I'm talking about the highlighted element of the "response status label" which is a span and not a button, and it's the one overflowing due to not currently wrapping.

2nd-gen/packages/swc/patterns/conversational-ai/response-status/ResponseStatus.ts
```diff
@@ -584,6 +773,62 @@ export class ResponseStatus extends SpectrumElement {
...
+        </button>
+        <div
+          id=${detailId}
+          class="swc-ResponseStatus-step-detailPanel ${open
+            ? 'swc-ResponseStatus-step-detailPanel--open'
+            : ''}"
+        >
+          <div class="swc-ResponseStatus-step-detailClip">
+            <div
+              class="swc-ResponseStatus-step-detailScroll"
```
rubencarvalho: When this becomes focusable (`tabindex="0"`) it's just a generic scroll region, so a screen reader announces nothing meaningful for the stop. Maybe we need `role="group"` + an accessible name (e.g. `aria-label` or `aria-describedby` back to the step toggle?).

2nd-gen/packages/swc/patterns/conversational-ai/response-status/ResponseStatus.ts
```diff
@@ -584,6 +773,62 @@ export class ResponseStatus extends SpectrumElement {
...
+            class="swc-ResponseStatus-step-title swc-Detail swc-Detail--sizeS"
+          >
+            ${step.label}
+          </span>
+          ${this._renderChevron(open, 'swc-ResponseStatus-step-chevron')}
+        </button>
+        <div
+          id=${detailId}
+          class="swc-ResponseStatus-step-detailPanel ${open
+            ? 'swc-ResponseStatus-step-detailPanel--open'
```
rubencarvalho: not blocking, super nit: this open/closed class could use `classMap({ 'swc-ResponseStatus-step-detailPanel--open': open })` instead of string interpolation.

2nd-gen/packages/swc/patterns/conversational-ai/response-status/ResponseStatus.ts
```diff
@@ -400,10 +500,19 @@ export class ResponseStatus extends SpectrumElement {
     const chevron = showDisclosure ? this._renderChevron(open) : '';
```
rubencarvalho: also on the same super nitpicky note, the empty-render fallbacks here and in the chevron/loader paths return `''` rather than Lit's `nothing`. just for idiomacy :)

rubencarvalho:
Tthe `StatusOnly` story shows as its own entry in the sidebar (it's on `main` already 😛 )

On `preview.ts` we're setting the `tags: ['!autodocs', '!dev']` globally, so a pattern is meant to opt exactly **one** story into the sidebar with `tags: ['dev']`, the Playground. `response-status` has two `['dev']` stories: `Playground` (correct) and `StatusOnly` (stray). `StatusOnly` also isn't referenced in `response-status.mdx`, so it's leaking into the sidenav.

2nd-gen/packages/swc/patterns/conversational-ai/response-status/ResponseStatus.ts
```diff
@@ -110,6 +149,13 @@ export class ResponseStatus extends SpectrumElement {
   @property({ type: String, attribute: 'accessible-label' })
   public accessibleLabel = '';
 
+  /**
+   * Pixel-loader icon cycle while `status="active"`. Invalid values fall back
+   * to `mega`.
+   */
+  @property({ type: String, reflect: true })
+  public preset: PixelLoaderPresetName = 'mega';
```
rubencarvalho: Could we align this with `swc-prompt-field`, which exposes the same pixel-loader knob as `loader` (`ruben/feat-prompt-field-pixel-loader`, about to be merged on main)? Both patterns surface the same underlying control, and right now they'd ship two different names (`preset` here vs `loader` there).

2nd-gen/packages/swc/patterns/conversational-ai/response-status/response-status.css
```diff
@@ -49,16 +49,17 @@
...
   flex: 1;
   min-inline-size: 0;
+  color: token("neutral-subdued-content-color-default");
   text-align: start;
 }
 
+.swc-ResponseStatus-label--emphasized {
+  --swc-body-font-color: token("neutral-content-color-default");
+}
+
```
TarunAdobe: `.swc-ResponseStatus-row--button .swc-ResponseStatus-label` sets a higher-specificity color, so `.swc-ResponseStatus-label--emphasized` cannot override it through `--swc-body-font-color`.

aramos-adobe: @TarunAdobe The design of this has changed so I will be reverting back to the OG state of the main disclosure button

2nd-gen/packages/swc/patterns/ai-toolkit/response-status/ResponseStatus.ts
```diff
@@ -132,32 +218,90 @@ export class ResponseStatus extends SpectrumElement {
 
     // Slotchange only fires when assigned nodes are added or removed, not when
     // their text mutates. Observe the light DOM so slotted text updates render.
+    // `subtree: true` on the host also matches the host's own `status`/`open`
+    // attribute changes (e.g. every header disclosure click); those are
+    // already handled by normal Lit reactivity, so records targeting the
+    // host itself are filtered out to avoid a redundant full re-scan.
     new MutationController(this, {
```
rubencarvalho, line 186: This is the big one for me. The subtree observer and the `_syncSlotContent` re-scan it drives only exist because the parent reads the steps' text and re-renders them itself. If each step rendered its own slotted content, native slotting would handle streaming updates for free and this observer wouldn't be needed at all. Worth noting too that the callback isn't coalesced, so during token streaming this re-scans every step on every mutation.

2nd-gen/packages/swc/patterns/conversational-ai/response-status/ResponseStatus.ts
```diff
@@ -70,16 +101,53 @@ export class ResponseStatus extends SpectrumElement {
...
 
+  /**
+   * Step elements whose `open` attribute `_handleStepToggle` just reflected
+   * itself. The `MutationController` below still observes that same
+   * attribute (to catch external writes, per the completion-contract note
+   * on step disclosure), so without this it would re-scan every step a
+   * second time for a change already patched into `_steps` synchronously.
+   * Consumed (deleted) the first time its own mutation record is seen.
+   */
+  private _pendingOwnOpenToggles = new Set<ResponseStatusStep>();
```
rubencarvalho: This guard only exists because the parent both writes `open` and then observes its own write through the MutationController above. If the step owned its own `open` state, there'd be no self-write to filter out and this whole thing would disappear.

2nd-gen/packages/swc/patterns/conversational-ai/response-status/ResponseStatus.ts
```diff
@@ -280,7 +451,8 @@ export class ResponseStatus extends SpectrumElement {
   private _syncSlotContent(): void {
```
rubencarvalho: Same root cause as the MutationController note — this is the scraping/mirror layer. With slotted, self-rendering steps there's nothing to mirror, so `_readStepElement`, `_stepsEqual`, and this all come out together.

2nd-gen/packages/swc/patterns/conversational-ai/response-status/ResponseStatus.ts
```diff
@@ -132,32 +218,90 @@ export class ResponseStatus extends SpectrumElement {
...
+        if (isStepMutation) {
+          this._syncSlotContent();
+        }
+      },
+    });
+
+    // Width changes reflow the description text, which can start or stop
+    // overflow without any reactive state changing; re-measure on resize.
+    // Defaults to observing the host and self-manages connect/disconnect.
+    new ResizeController(this, {
```
rubencarvalho: I’ve been looking at these two observers (this one and the IntersectionController just below), and I don’t think we actually need either of them. As far as I can tell, the ResizeController, IntersectionController, _isVisible, and the rAF logic around them all exist to keep the overflow measurement up to date so the scroll region gets tabindex="0" when it overflows, and loses it when it doesn’t. The scrolling itself is already handled by CSS... max-height with overflow-y: auto does that without any of this.

So I think the real question is just how we decide when the region should be focusable. There are two options that would let us remove the observers.

The simplest is to always make it focusable — tabindex="0" and role="group" whenever there’s a description, with no measurement at all. It’s accessible and very little code. The downside is that short descriptions that don’t scroll would still end up in the tab order, so we’d probably have more tab stops than we want.

The other option, which I prefer, is to measure once when a step is expanded instead of watching it continuously. A simple scrollHeight > clientHeight check in _handleStepToggle (plus once for any step that starts open) gives us the same conditional tabindex we have today. We could then remove both observers, _isVisible, and the continuous _scheduleDetailOverflowSync / _detailOverflowRaf logic, and just keep _overflowingSteps.

The tradeoff is that it wouldn’t automatically update if something changes while the step is open. For example, resizing the window could cause the text to wrap differently, or more text could stream in, and the tabindex could be wrong until the next toggle. For agentic steps, though, the description will usually be finished before someone expands it, so that feels like a reasonable tradeoff. And if it does become a problem, we can always add a small re-measure for that specific case rather than keeping all these observers running.

So I think either option lets us remove the observers. It mostly comes down to a few extra tab stops vs. the chance of a stale tabindex after a resize/content change... Performance wise it's gonna make a big difference if we have tens or hundreds of these reponse status steps...

rubencarvalho:
I added a PR for a performance improvement suggestion, after having seen what Devon did with React's implementation:
https://github.com/adobe/spectrum-web-components/pull/6680

Basically, the tldr is that animating the background-position isn't GPU-accelerated so the browser has to repaint the text every single frame (which is expensive for something running continuously like we have here). Animating translate runs on the compositor instead, so the browser doesn't need to repaint the text underneath.

rubencarvalho (approved):

PR #6625, opened 2026-08-17
fix(overlay): don't close auto overlay on click inside content in a focusable ancestor

### LINKED ISSUE #5731: [Bug]: Overlay in focus region dismissed when click into its content

Rajdeepc:
## Description

Fixes `[type="auto"]` `sp-overlay` incorrectly closing when clicking non-focusable content inside its own popover, when the trigger and the overlay share a focusable ancestor (e.g. a `<div tabindex="0">` wrapping both).

`closeOnFocusOut` decided whether focus left the overlay by checking if the newly-focused element (`event.relatedTarget`) was a descendant of the overlay. Clicking non-focusable overlay content (plain text, padding, etc.) causes the browser to resolve focus onto the nearest focusable ancestor instead of the click target; when that ancestor also wraps the overlay itself, it sits above the overlay in the DOM, so the check always concluded focus had left and closed the overlay even though the click landed inside it.

`closeOnFocusOut` now also tracks whether the `pointerdown` causing the current focus change originated inside the overlay's own composed subtree (via a new capture-phase `pointerdown`/`pointerup` pair mirroring the existing `focusout` listener lifecycle), and treats focus as remaining within the overlay in that case regardless of where it was ultimately resolved to.

## Motivation and context

Reported in the linked issue: any consumer that needs a focusable wrapper around a trigger + `sp-overlay` pair (for example, a keyboard-navigable card or list row) sees the overlay dismiss itself on the very first click inside its own content, making `type="auto"` unusable in that composition pattern.

## Related issue(s)

- fixes #5731
- fixes SWC-1336

## Author's checklist

## Reviewer's checklist

### Manual review test cases

- [ ] _Auto overlay stays open when clicking its own content inside a focusable wrapper_
  1. Go to the `Overlay Element` story `focusableAncestorWrapper` (`dev` tag) in Storybook
  2. Click "Button #1" (or "Button #2") to open its popover, then click the popover's text content
  3. Expect the popover to remain open, matching "Button #3" (the unwrapped control)

- [ ] _No regression to existing focus-out / outside-click dismissal_
  1. Open any `type="auto"` overlay elsewhere in the library (e.g. Picker, Action Menu)
  2. Click completely outside the trigger and its overlay
  3. Expect the overlay to still close as before

### Device review

- [ ] Did it pass in Desktop?

  1. Open a `type="auto"` overlay via its trigger with <kbd>Enter</kbd>/<kbd>Space</kbd>
  2. Press <kbd>Tab</kbd> to move focus through the overlay's content and then past it
  3. Expect the overlay to still close when focus moves past its last focusable element (unchanged by this fix — covered by the existing "closes an inline overlay when tabbing past the content" test)

  1. Open a `type="auto"` overlay wrapped in a focusable ancestor, with a screen reader active
  2. Click inside the popover's non-focusable content
  3. Expect the popover's role/state announcements to remain unchanged; no unexpected dismissal or duplicate "closed" announcement

## Verification performed

- Added a regression test (`overlay-element.test.ts`) reproducing the reported scenario exactly (trigger + overlay wrapped in `tabindex="0"`, click on non-focusable content) — passes.
- Full `overlay` yarn test group: all tests pass.
- `overlay-api` yarn test group (action-menu, dialog, menu, overlay, picker, tooltip — the real `type="auto"` consumers): all tests pass, no regressions.
- Added a `dev`-tagged Storybook story (`focusableAncestorWrapper`) reproducing the exact three-button scenario from the issue for manual verification.
- `eslint` / `prettier`: clean.


1st-gen/packages/overlay/src/Overlay.ts
```diff
@@ -735,6 +759,18 @@ export class Overlay extends ComputedOverlayBase {
...
   private closeOnFocusOut = (event: FocusEvent): void => {
+    // Consume the flag immediately so a stale value from an earlier click cannot
+    // affect a later, unrelated focus change (e.g. a keyboard Tab away).
+    const pointerdownWasWithinOverlay = this.lastPointerdownWasWithinOverlay;
+    this.lastPointerdownWasWithinOverlay = false;
+
+    // The interaction that caused this focus change started within the overlay,
+    // even though focus may have resolved onto an ancestor outside of it. Treat
+    // focus as still within the overlay.
+    if (pointerdownWasWithinOverlay) {
```
cdransf: Is this guard too broad? It's placed before we check `event.relatedTarget` (we could check if it's an ancestor of the overlay in the composed tree rather than tracking the pointer gesture origin). E.g. if we have a button or interaction inside the overlay that sends focus elsewhere, the overlay would stay open. ✨

1st-gen/packages/overlay/src/Overlay.ts
```diff
@@ -829,10 +865,32 @@ export class Overlay extends ComputedOverlayBase {
...
+          this.clearPointerdownOrigin,
+          {
+            capture: true,
+          }
+        );
         listenerRoot.addEventListener('focusout', this.closeOnFocusOut, {
           capture: true,
         });
       } else {
+        listenerRoot.removeEventListener(
```
cdransf, line 911: Should we reset `lastPointerdownWasWithinOverlay` here?

`this.lastPointerdownWasWithinOverlay = false;`

cdransf (changes requested):

1st-gen/packages/overlay/src/Overlay.ts
Rajdeepc: If we check `relatedTarget` is an ancestor of the overlay alone isn't a safe replacement though, since `document.body` is trivially an ancestor of every overlay. I think that'd stop the overlay from ever closing on outside click.

I would track the actual `pointerdown` target, and only suppress the close when `relatedTarget` is an ancestor of (or equal to) that specific element.

1st-gen/packages/overlay/src/Overlay.ts
Rajdeepc: Will reset it there so a pending flag can't leak into the next open cycle if the overlay tears down before its matching `pointerup` fires. Good catch.

Rajdeepc:
Pushed in db1873d: both addressed.

- `closeOnFocusOut` now tracks the actual `pointerdown` target instead of a boolean, and only suppresses the close when `relatedTarget` is an ancestor of (or equal to) that specific target — an explicit focus redirect to something unrelated now falls through to the normal close path.
- The pending record is also cleared on `pointercancel` (interrupted gesture) and reset when the auto-type listeners tear down, so nothing leaks into a later open cycle.

Added a regression test for the redirect-to-unrelated-element scenario; confirmed it fails against the previous implementation and passes against this one. Full `overlay` and `overlay-api` suites still green.

1st-gen/packages/overlay/src/Overlay.ts
cdransf: Ahhh yeah, that makes sense!

cdransf (approved):
Looks good! ✨

.changeset/fix-overlay-focusout-tabindex-ancestor.md
rise-erpelding, line 1: I really like the breakdown of this changeset 😍

rise-erpelding (approved):

PR #6626, opened 2026-08-17
fix(*): apply box-sizing to sized slotted elements

5t3ph:
## Description

Adds an explicit `box-sizing: border-box` declaration to every `::slotted()` rule across 2nd-gen components and patterns that sets a real dimensional constraint — `inline-size`, `block-size`, `aspect-ratio`, or a `max-inline-size`/`max-block-size` cap — and didn't already declare it.

Affected files:
- `card.css`, `card-template.css` (collection/preview slots)
- `upload-artifact.css`, `user-message.css` (media thumbnails)
- `prompt-field.css` (artifact tiles)
- `tab.css`, `action-button.css`, `button.css`, `infield-button.css` (icon slots)
- `illustrated-message.css` (slotted SVG)
- `icon-base.css` (shared icon frame, also fixes the non-slotted `svg` selectors sharing the rule)
- `global-action-button.css`, `global-button.css` (generated global-style mirrors of the icon slot fix)

Also updates the style guide and migration skill docs so this isn't lost on future components:
- New anti-pattern entry (`05_anti-patterns.md`) explaining why the reset doesn't reach slotted content
- Note added to the box-sizing rule in `01_component-css.md`
- New checklist item in `03_component-css-pr-checklist.md`
- New rule added to the migration-styling skill's TL;DR reference

`min-*`-only constraints and keyword sizes like `fit-content` are intentionally excluded — they don't produce a rendered box that can overflow.

## Motivation and context

Every 2nd-gen component resets box-sizing with `* { box-sizing: border-box; }` inside its shadow-scoped stylesheet. That universal selector only matches elements in the shadow tree — it never reaches `::slotted()` content, which lives in the consumer's light DOM. Any `::slotted()` rule that constrains size (e.g. `inline-size: 100%` + `aspect-ratio`) is therefore sized against whatever box-sizing the *consumer's* element happens to have (`content-box` by default). If that element also carries its own `padding` or `border` — from consumer styles, a UA stylesheet, or another library — the rendered box grows past the intended constraint, breaking aspect ratio, causing overflow, or misaligning grid layouts. This was caught while resolving the same issue on Card's collection/preview slots and generalized across the rest of the 2nd-gen component set.

## Manual review test cases

- [ ] _Box-sizing does not change any current visual output_
  1. Run Storybook and open each affected component's Overview/Options docs page: Card, Action Button, Button, Infield Button, Tabs, Illustrated Message, Icon, Asset, and the Prompt Field / Upload Artifact / User Message conversational-ai patterns.
  2. Compare rendering against `main` and VRTs.
  3. Expect pixel-identical output — `box-sizing` only changes behavior when the slotted element also has its own padding/border, which none of the current stories set.

- [ ] _Slotted content with its own border/padding no longer overflows its constrained box_
  1. In the Card `Preview`/`Collection` story (or Prompt Field's artifact tiles), temporarily add `border: 4px solid red; padding: 8px` to the slotted element via devtools.
  2. Confirm the element stays within its constrained `inline-size`/`aspect-ratio` box instead of growing past it.
  3. Remove the temporary override — this is a manual verification step, not a permanent story change.

### DIFF
--- a/CONTRIBUTOR-DOCS/02_style-guide/01_css/05_anti-patterns.md
+++ b/CONTRIBUTOR-DOCS/02_style-guide/01_css/05_anti-patterns.md
@@ -77,6 +77,11 @@
     - [Why This Happens](#why-this-happens)
     - [Why This Is a Problem](#why-this-is-a-problem)
     - [✅ Correct Approach](#-correct-approach)
+- [13. Missing `box-sizing` on Sized `::slotted()` Rules](#13-missing-box-sizing-on-sized-slotted-rules)
+    - [❌ Anti-Pattern](#-anti-pattern)
+    - [Why This Happens](#why-this-happens)
+    - [Why This Is a Problem](#why-this-is-a-problem)
+    - [✅ Correct Approach](#-correct-approach)
 - [Final Reminder](#final-reminder)
 
 </details>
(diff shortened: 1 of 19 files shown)


cdransf (approved):

CONTRIBUTOR-DOCS/02_style-guide/01_css/05_anti-patterns.md
```diff
@@ -627,6 +632,45 @@ During migration, this removal may affect alignment of accessories (ex. status l
 🔎 **Status light reference:**
 [status-light.css](../../../2nd-gen/packages/swc/components/status-light/status-light.css) sets only `font-size` and `line-height` per size; `min-block-size` and `padding-block` are absent.
 
+## 13. Missing `box-sizing` on Sized `::slotted()` Rules
```
miwha-adobe, line 635: Ty for adding!

miwha-adobe (approved):
lgtm

PR #6630, opened 2026-08-17
refactor(prompt-field): use swc-action-button for composer and artifact buttons

rubencarvalho:
## Description

Replaces the five hand-rolled `<button>`s in the prompt-field pattern with `swc-action-button`: **upload**, **send**, **stop**, the two **artifact scroll chevrons**, and the **upload-artifact close button**. Icons move to the `icon` slot, `aria-label` → `accessible-label`, and the bespoke button CSS is dropped in favor of the component's, re-themed only where the AI treatment needs it.

Based on `main` (the AI brand treatment has since merged); net −81 lines.

### Button appearance
- **send / stop**: circle (`border-radius: full`), neutral / `gray-900` fill, `gray-25` icon
- **upload**: quiet; static content/hover colors via `light-dark()` (black on the light card, white on the dark card) so the icon matches the themed surface, including the generating status indicator
- **send disabled**: auto-contrast tint instead of the flat Spectrum disabled fill
- **chevrons**: default action-button circle; roving focus managed via `aria-disabled` + `tabindex`
- **dismiss**: 20px round circle, self-sized from a single `--swc-upload-artifact-dismiss-visual-size` (icon-only, so `edge-to-visual-only` is pinned so width tracks height and it stays square)

### Focus
- All composer focus rings (buttons, artifact tiles, and dismiss buttons) are theme-adaptive **black on light / white on dark** via `light-dark()`, instead of the default accent, scoped to the prompt field; standalone `upload-artifact` keeps its own default ring
- When keyboard paging reaches an end and the focused chevron becomes `aria-disabled` (and hidden), focus **moves to the strip's active tile** rather than being stranded on the now-hidden control
- Restored the strip's Tab / Shift+Tab navigation for the new delegated-focus buttons (the deep active element is the action-button's internal `<button>`, so it is mapped back to the host before the Tab comparisons)

### Layout and scrolling
- Button-adjacent layout (text-group padding, collapsed row height, artifact fade plateau, scroll-padding) reads `component-height-100` directly instead of the old `--swc-prompt-field-*-size` custom properties, which the buttons stopped honoring once they became fixed-size action-buttons
- Artifact chevron paging scrolls the snap-aligned page-start tile into view (edge-clamped) with `scroll-snap-type: inline proximity`, incorporating the fix from the closed #6616

## Verification

Rendered and measured every button in light + dark mode (fills, icon colors, shapes, dismiss sizing on card + media); the static colors and focus rings flip with the card. Chevrons coexist with the strip's edge-fade mask. All prompt-field and upload-artifact play-function tests pass; VRT forced-state coverage now targets the action-button's internal element so hover/focus rows are meaningful again.

- [ ] **Keyboard**: Tab through upload → textarea → send/stop; with artifacts, Tab into the strip (chevrons, thumbnails, close buttons), Arrow-key the roving thumbnail; scroll to an end with a chevron and confirm focus lands on a visible tile (not the hidden chevron); black/white focus rings visible throughout in both themes.
- [ ] **Screen reader**: each button announces its `accessible-label` (Add attachment, Send, Stop generating, Show previous/more attachments, Remove attachment); no duplicate or missing names.


TarunAdobe (approved):

aramos-adobe (approved):
This looks great!!

PR #6631, opened 2026-08-17
style(prompt-field): mirror the AI brand gradients in RTL

rubencarvalho:
## Description

The AI brand treatment's ambient gradients (card wash, outer ring, hue-sweep) are anchored to the physical bottom-right. In RTL the composer layout flips (send button moves to the bottom-left) but the glow stays pinned to the right, so the two disagree. This mirrors the gradients to follow the layout.

Based on `ruben/feat-prompt-field-ai-branding-v3`.

## What changed

- Hoisted the horizontal anchor of the card wash and outer ring into custom properties (`--_swc-prompt-field-wash-x`, `--_swc-prompt-field-outer-dir`).
- Three `:host(:dir(rtl))` rules flip them (`right`→`left`, `to right`→`to left`) plus `scaleX(-1)` on the decorative `::before` hue-sweep (its animations use the independent `rotate`/`translate`/`scale` properties, so `transform` is free).

Net: two small base edits + three RTL rules.

## Motivation and context

> ⚠️ **Note:** the react reference does **not** mirror these gradients in RTL. This is a deliberate divergence for visual consistency with the mirrored composer layout, and should get a design nod before merge.

## Verification

Rendered prominent LTR vs RTL in Storybook: LTR glow sits bottom-right (send button), RTL glow mirrors to bottom-left (mirrored send button). Only the composer/gloss (vertical) is unaffected, as intended.

- [ ] **Keyboard**: no interactive changes; decorative-only. Confirm no focus or tab-order regressions in RTL.
- [ ] **Screen reader**: no semantic changes; gradients are decorative (`pointer-events: none`, not in the a11y tree).

### DIFF
--- a/2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/prompt-field.css
+++ b/2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/prompt-field.css
@@ -68,7 +68,7 @@
   flex-direction: column;
   min-inline-size: 0;
   padding: token("spacing-300");
-  background: radial-gradient(circle at right bottom in oklch, var(--_swc-prompt-field-bg-stop-1) 0%, var(--_swc-prompt-field-bg-stop-2) 35%, var(--_swc-prompt-field-bg-stop-3) 82%, var(--_swc-prompt-field-bg-stop-4) 100%);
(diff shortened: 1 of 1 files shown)


rubencarvalho:
Folded directly into the base branch `ruben/feat-prompt-field-ai-branding-v3` (cherry-picked) since the change is small enough not to need its own review cycle. Closing in favor of the base branch.

PR #6632, opened 2026-08-18
style(prompt-field): fade the artifact strip with a mask instead of a solid overlay

rubencarvalho:
## Description

The multi-artifact strip's edge fades painted a `background-layer-2-color` gradient *over* the tiles. On the flat card that read as "tiles fading into the background"; on the AI brand treatment's gradient wash it reads as a grey smudge that doesn't match the branding.

This masks the scroll container so the tiles themselves fade to transparent at each scrollable edge and the branded wash shows through — no color to match, so it adapts to the gradient (and any theme) automatically.

Based on `ruben/feat-prompt-field-ai-branding-v3`.

## What changed

- Added a `mask-image` fade on `.swc-PromptField-artifacts-scroll`, its edges driven by the existing `has-scroll-prev` / `has-scroll-next` classes; RTL flips the physical mask direction.
- Removed the two overlay fade `<div>`s (from the render), the `fade-solid` var, and their LTR/RTL background rules.
- Removed the chevrons' opaque `::before` mask — tiles now fade before reaching a chevron, so it's redundant (and was the same grey-on-gradient issue).

Net −66 lines.

## Verification

Rendered the multi-artifact strip (prominent) in Storybook: at rest only the scrollable edge fades; scrolled to the middle both edges dissolve into the wash; RTL fades the correct edges. Chevrons sit cleanly over the gradient with no grey rectangle behind them.

- [ ] **Keyboard**: no interaction changes; chevron roving/paging and scroll-padding clear-zone are unchanged. Confirm no regression tabbing through the strip.
- [ ] **Screen reader**: no semantic changes; the fade is a decorative `mask-image` (the removed `<div>`s were already `aria-hidden`).

### DIFF
--- a/2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/PromptField.ts
+++ b/2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/PromptField.ts
@@ -972,22 +972,6 @@ export class PromptField extends SpectrumElement {
                 ></slot>
               </div>
             </div>
-            ${this._artifactCanScrollPrev
-              ? html`
-                  <div
-                    class="swc-PromptField-artifacts-fade swc-PromptField-artifacts-fade--start"
-                    aria-hidden="true"
-                  ></div>
-                `
-              : nothing}
-            ${this._artifactCanScrollNext
-              ? html`
-                  <div
-                    class="swc-PromptField-artifacts-fade swc-PromptField-artifacts-fade--end"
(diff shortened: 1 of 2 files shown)


rubencarvalho:
Folded directly into the base branch `ruben/feat-prompt-field-ai-branding-v3` (cherry-picked) since the change is small enough not to need its own review cycle. Closing in favor of the base branch.

PR #6607, opened 2026-08-12
ci(release): split gen1/gen2 release pipelines with a Version PR flow

blunteshwar:
> **Update (2026-08-18):** addressed the latest review round from @Rajdeepc:
> - `publish.yml`'s `release` job dropped the git-identity step when the direct-push-to-`main` flow was replaced, but `create-git-tag.js` still runs `git tag -a`, which needs it. Restored `git config user.name`/`user.email` right before that call, so the tag actually lands instead of silently failing and forcing every subsequent push to re-run the non-idempotent React-wrapper publish loop.
> - Removed the orphaned root `scripts/publish.js` (and its `publish`/`publish:snapshot` yarn scripts). It still ran `yarn changeset version`/`publish` from the repo root, but the root `.changeset/config.json` no longer exists now that 1st-gen and 2nd-gen have fully independent changesets instances, and nothing in CI or docs referenced it.

## Description

Splits the single shared release pipeline into two **fully independent** streams, one per generation, and replaces direct-push-to-`main` versioning with a reviewable **Version PR**.

- `1st-gen/.changeset/` and `2nd-gen/.changeset/` are two genuinely independent `@changesets/cli` instances (own `config.json`, own `README.md`, own pending changeset files) — not a shared root folder. `@changesets/cli` has no `--cwd` flag; each instance is invoked *from inside* its own directory (`working-directory:` in CI, `cd` locally). `yarn changeset:1st-gen` / `yarn changeset:2nd-gen` are thin wrappers for that.
- `publish.yml` is 1st-gen only in every trigger, including its own `snapshot-release` PR path — it never reads `2nd-gen/.changeset/`. `publish-2nd-gen.yml` is the fully independent counterpart.
- Each workflow has two jobs: `snapshot` (throwaway publishing, gated on pending changesets — `snapshot-release`-labeled PRs and pushes to `main` that happen to have a pending changeset publish under `next`/`snapshot-test`) and `release` (the real release path, described below).
- **No more direct push to `main`.** The `release` job versions locally with `yarn changeset version` and opens/updates a dedicated **Version PR** via [`peter-evans/create-pull-request`](https://github.com/peter-evans/create-pull-request), on a distinct branch per generation (`changeset-release/1st-gen` / `changeset-release/2nd-gen`). Merging that PR is itself a push with no pending changesets left, which the same workflow reads as "publish now": builds, `yarn changeset publish`, and (1st-gen only) builds/publishes the React wrappers and creates a git tag.
[...]

PR #6637, opened 2026-08-19
ci(docs): stage 2nd-gen docs on merge, gate prod publish behind #gen2-publish

blunteshwar:
## Description

Splits the 2nd-gen documentation deploy into two targets so the public site can stay in sync with releases instead of updating on every merge:

- **Staging** — every push to `main` (non-1st-gen changes), every manual run, and every successful, push-sourced conclusion of **Publish Packages (2nd-gen)** syncs to `$web/docs-staging/` (`https://swcpreviews.z13.web.core.windows.net/docs-staging/`).
- **Production** — `$web/docs/` (mapped to `spectrum-web-components.adobe.com`) deploys only on a manual `workflow_dispatch`, or once **Publish Packages (2nd-gen)** concludes successfully (from a push, not a fork PR) on a run whose commit message contains `#gen2-publish`.

The keyword is `#gen2-publish` rather than `#publish` on purpose: the 1st-gen `publish-docs-site.yml` matches on the `#publish` substring, so a shared keyword would couple the two sites. It's carried on both the Version PR's `title` and `commit-message`, since GitHub's `squash_merge_commit_title` setting can use either depending on configuration for a single-commit PR.

Fixes found and addressed during review (thanks @Rajdeepc):

- **Critical — fork PR could get code execution against the Azure deploy credentials.** `Publish Packages (2nd-gen)` also runs on `pull_request` (labeled/synchronize, for its snapshot-release path), so a fork PR can produce a `workflow_run` completion too, not just a push to `main`. `workflow_run`-triggered jobs get this repo's full secrets regardless of which run triggered them — checking out and building that fork's `head_sha` (as this job did) would have handed a fork PR arbitrary code execution against `AZCOPY_*`/Azure storage credentials during `yarn install`/`yarn build`. Added a job-level guard so `workflow_run` only proceeds when the upstream run was itself push-triggered and concluded successfully; fork-originated `pull_request` runs never reach this job now. Also moved the `AZCOPY_*` secrets out of the workflow-level `env:` down into just the two deploy steps, so install/build steps never have them in their process environment even on a legitimate run.
- **Production deploy could fire before the actual release shipped.** Originally gated on the push that merges the Version PR — but that push only *starts* `publish-2nd-gen.yml`, whose real npm-publish step can sit behind a required manual environment approval. Switched the production trigger to `workflow_run` on that workflow's completion (now also gated push-sourced-only per the fix above). Checkout pins to the `workflow_run`'s `head_sha` so the build always reflects the commit that was actually released.
- **`cancel-in-progress: true` could corrupt a live deploy.** `azcopy sync --delete-destination` deletes and uploads in the same run; cancelling one mid-flight could leave the site half-deleted. Flipped to `false`.
- **`check-changesets` failing on the common case.** In both `publish.yml` and `publish-2nd-gen.yml`, the changeset count was computed with `ls .../*.md 2>/dev/null | grep -v README | wc -l | tr -d ' '`, which exits nonzero under `set -eo pipefail` whenever zero changesets are pending — exactly the state right after a Version PR merges, so the real publish job was being skipped by GitHub Actions' default failed-dependency behavior. Switched both to `find ... -maxdepth 1 -name '*.md' ! -name 'README.md' | wc -l`, which exits 0 regardless of match count.
- **1st-gen's private root package could get version-bumped.** `1st-gen/.changeset/config.json`'s `fixed: [["@spectrum-web-components/*"]]` wildcard-matches the private `@spectrum-web-components/1st-gen` root itself. Added `"privatePackages": false` — a structural exclusion of anything private, rather than a name-based one that would silently stop protecting the package if it's ever renamed.

## Motivation and context

We're moving into pre-release/beta publishing and want the public 2nd-gen docs site to update in sync with announced releases, not on every merge. Staging keeps tracking `main` for day-to-day validation.

Rebased onto the current `release-strategy` base (post gen1/gen2 changesets split in #6607) — the docs-deploy workflow itself is orthogonal to package publishing, but its production trigger needed to be wired to the actual release event under the new Version PR flow, which surfaced the pipeline and security issues above.

## Related issue(s)

- SWC-2487

N/A — CI/documentation-deploy change only; no component or runtime UI is modified.

## Author's checklist

- [ ] I have reviewed at the Accessibility Practices for this feature — N/A, no UI surface.
- [ ] I have added automated tests to cover my changes — N/A, workflow YAML isn't unit-testable in this repo; traced the job-level `if:` against every `event_name`/`workflow_run.event`/`conclusion` combination by hand, validated the `check-changesets` fix locally, confirmed `privatePackages: false` against `@changesets/config`'s own source.
- [ ] I have included a well-written changeset if my change needs to be published — N/A, CI/docs-only, no published package changed.
- [x] I have included updated documentation if my change required it (`CONTRIBUTOR-DOCS/01_contributor-guides/06_releasing-swc.md`).

## Reviewer's checklist

- [ ] Includes a GitHub Issue with appropriate flag or Jira ticket number without a link

### Manual review test cases

- [ ] _A fork PR's snapshot-release run cannot trigger this workflow_
  1. Label a fork-originated PR `snapshot-release` (or push to it while labeled) so `publish-2nd-gen.yml` runs via `pull_request`.
  2. Confirm that run's completion does **not** start `build-and-deploy` in this workflow — no checkout, no build, no secrets exposure.

- [ ] _Staging deploys on merge_
  1. Merge a non-1st-gen change to `main`.
  2. Confirm **Publish 2nd-Gen Documentation** runs "Deploy to staging" and skips "Deploy to production".
  3. Verify `https://swcpreviews.z13.web.core.windows.net/docs-staging/` reflects the change.

- [ ] _Production deploys only after the release actually concludes_
  1. Merge the 2nd-gen Version PR (commit message and title both carry `#gen2-publish`) opened by `publish-2nd-gen.yml`.
  2. If the `npm-publish` environment requires approval, confirm production docs do **not** deploy while that approval is pending.
  3. Once `Publish Packages (2nd-gen)` concludes with `success`, confirm this workflow runs via `workflow_run` and deploys both staging and production.
  4. Verify `spectrum-web-components.adobe.com` reflects the change.

- [ ] _A Version-PR-opening run doesn't trigger a production deploy_
  1. Push a change with a pending 2nd-gen changeset so `publish-2nd-gen.yml` only opens/updates the Version PR (no publish).
  2. Confirm that run's conclusion does **not** trigger a production docs deploy.

- [ ] _Production deploys on manual dispatch_
  1. Run **Publish 2nd-Gen Documentation** via `workflow_dispatch`.
  2. Confirm both staging and production deploy regardless of any commit message.

- [ ] _No cross-trigger with 1st-gen_
  1. Confirm a `#gen2-publish` commit does not trigger **Publish Documentation Site** (1st-gen).

- [ ] _check-changesets no longer fails on zero pending changesets_
  1. Push to `main` (or run `workflow_dispatch`) with no pending changesets in `1st-gen/.changeset/` or `2nd-gen/.changeset/`.
  2. Confirm both workflows' `check-changesets` job succeeds (not a red X) and reports `has_changesets=false`.
  3. Confirm the downstream publish job actually runs (not skipped) when that's the state right after a Version PR merge.

- [ ] _1st-gen's private root package is never versioned_
  1. Add a changeset touching any real 1st-gen package (e.g. `button`) and let the Version PR generate.
  2. Confirm the resulting version bumps only include real published packages, not `@spectrum-web-components/1st-gen`.

- [ ] _A cancelled/interrupted docs deploy can't happen mid-sync_
  1. Push twice in quick succession.
  2. Confirm the second run queues behind the first (`cancel-in-progress: false`) instead of cancelling it mid-`azcopy sync`.

### Device review

- [ ] Did it pass in Desktop? (N/A — no UI)
- [ ] Did it pass in (emulated) Mobile? (N/A — no UI)
- [ ] Did it pass in (emulated) iPad? (N/A — no UI)

N/A — this PR only changes CI/documentation-deploy configuration; no component, page, or interactive surface is affected. No keyboard or screen reader testing applies.

  1. N/A — no focusable UI is added or changed by this workflow/docs update.

  1. N/A — no roles, names, or announcements are affected by this workflow/docs update.


.github/workflows/publish-2nd-gen.yml
```diff
@@ -195,7 +195,7 @@ jobs:
         uses: peter-evans/create-pull-request@v6
         with:
           commit-message: 'chore: release 2nd-gen packages'
```
Rajdeepc:
```suggestion
          commit-message: 'chore: release 2nd-gen packages #gen2-publish'
```
GitHub's default `squash_merge_commit_title` uses commit message for single commit PRs not the PR title. Please put the keyword in commit-message too, so it survives every merge strategy

.github/workflows/publish-2ndgen-docs.yml
```diff
@@ -73,11 +73,28 @@ jobs:
...
+            --delete-destination=true \
+            --from-to LocalBlob
+          echo "Deployed to https://swcpreviews.z13.web.core.windows.net/docs-staging/"
+
+      - name: Deploy to production
         # spectrum-web-components.adobe.com maps to the /docs path in the storage
         # container, so deploying to $web/docs/ IS the domain root. The generated
         # .well-known/agent-skills/ directory inside storybook-static/ is therefore
         # served at the expected discovery URL without a separate upload step.
+        if: github.event_name == 'workflow_dispatch' || contains(github.event.head_commit.message, '#gen2-publish')
```
Rajdeepc: So once you merge the version PR it will start two independent workflows. the npm release job has required reviewers so it will be waiting. Here this workflow will deploy to production immediately.
My recommendation would be to trigger the docs deploy from the release rather than the same push means a `workflow_run `trigger on `Publish Packages (2nd-gen)` gated on `conclusion == 'success',` or a `repository_dispatch` fired by the publish job after changeset publish returns.

1st-gen/.changeset/config.json
```diff
@@ -12,6 +12,7 @@
   "commit": false,
   "fixed": [["@spectrum-web-components/*"]],
   "ignore": [],
+  "privatePackages": false,
```
Rajdeepc, line 15: You can drop `privatePackages` and restore the targeted exclusion instead

.github/workflows/publish-2ndgen-docs.yml
```diff
@@ -73,11 +73,28 @@ jobs:
       - name: Setup AzCopy
         uses: ./.github/actions/setup-azcopy
 
-      - name: Deploy to Azure Blob Storage
+      # Every merge to main deploys to the staging path (docs-staging/). The
```
Rajdeepc: On line 19 `cancel-in-progress: true` in `publish-2ndgen-docs.ym` in a group with a ref is risky. Any push to main can kill an inflight production sync after deletes but before uploads. 
Why dont you set `false` for this workflow. Docs deploys are cheap to queue and expensive to interrupt.

.github/workflows/publish-2ndgen-docs.yml
```diff
@@ -73,11 +73,28 @@ jobs:
...
+      # production path (docs/) maps to spectrum-web-components.adobe.com and is
+      # published only on a manual run or a commit whose message contains
+      # `#gen2-publish`, so the public site stays in sync with releases. The
+      # keyword deliberately avoids the substring `#publish`, which the Gen1
+      # publish-docs-site workflow matches on.
+      - name: Deploy to staging
```
Rajdeepc, line 108: This has to no `if` so here `worlflow_dispatch` will hit both target. It contradicts with the docs table you added. 
Please add if: `github.event_name == 'push'` on the staging step, or amend the docs table to say a manual run refreshes both.

Rajdeepc (changes requested):
Its a minor but check this:
`paths-ignore: - '1st-gen/**'` is isolated but tge 1st gen version PR also commits yarn.lock at the repo root out this glob. If we run a pure gen1 release the whole gen2 job will run to produce a staging deploy identical to the last one. Can you invert this to an allowlist?
```ts
paths:
  - '2nd-gen/**'
  - '.github/workflows/publish-2ndgen-docs.yml'
```

blunteshwar:
> Its a minor but check this: `paths-ignore: - '1st-gen/**'` is isolated but tge 1st gen version PR also commits yarn.lock at the repo root out this glob. If we run a pure gen1 release the whole gen2 job will run to produce a staging deploy identical to the last one. Can you invert this to an allowlist?
> 
> ```ts
> paths:
>   - '2nd-gen/**'
>   - '.github/workflows/publish-2ndgen-docs.yml'
> ```

Fair catch on the yarn.lock gap, but I'd like to keep paths-ignore over an allowlist here. yarn.lock is a single shared lockfile for the whole monorepo — a path filter genuinely can't tell "this lockfile change is gen1-only" from "this lockfile change bumped a dependency gen2 also uses." An allowlist (2nd-gen/** + the workflow file) fixes the false-positive (redundant staging sync on a pure-gen1 release) but trades it for a false-negative: a real root-level change that does affect the gen2 build (shared config, root package.json, a genuine cross-cutting yarn.lock bump) would silently stop triggering a docs rebuild, and staging would go stale with no visible signal.

A wasted, idempotent azcopy sync that re-uploads identical content is cheap and self-correcting on the next real push. A missed trigger justsits there quietly wrong until someone notices the docs are out of date. Given that asymmetry, I'd rather keep the current over-inclusive paths-ignore and eat the occasional redundant run. Happy to revisit if this turns out to run more often than expected in practice.

1st-gen/.changeset/config.json
blunteshwar: I'd push back on reverting to the named-exclusion approach. The original ignore: [...] list excluded the 1st-gen root by string-matching its package name — which means if that package is ever renamed (or a new private workspace-root package is added later, e.g. for 2nd-gen), it
silently stops being protected and nobody finds out until it gets versioned by accident again. That's exactly the failure mode that got us here: the original ignore entry was dropped in the gen1/gen2 split and nothing caught it.

`privatePackages: false` is structural instead of name-based: "never version anything marked private," full stop. I checked every package in both changesets scopes — the two workspace-root packages are the only private ones either config will ever see — so this has zero effect on anything that's actually meant to be published, and it can't rot the way a name-based list can. I'd rather keep this and let it defend itself against future renames/additions.

.github/workflows/publish-2ndgen-docs.yml
blunteshwar: Fixed by reconciling the docs, not the code. Went with your second option: rather than adding if: github.event_name == 'push' to the stagingstep, I updated the Gen2 docs table to say staging refreshes on every trigger — push, manual dispatch, and now also the workflow_run that follows a real release. Restricting staging to push-only would have been the wrong direction anyway once production started depending on workflow_run: that trigger now also needs staging to stay unconditional, since it's harmless to re-sync and there's no reason a manual or release-triggered run should skip refreshing it.

Table now reads "Every push to main that touches non-Gen1 files, every manual run, and every conclusion of Publish Packages (2nd-gen)" for staging.

.github/workflows/publish-2ndgen-docs.yml
```diff
@@ -10,13 +10,26 @@ on:
...
       - '1st-gen/**'
+  # Production deploy is driven by this, not by the push above - the push that
+  # merges the 2nd-gen Version PR only starts Publish Packages (2nd-gen), whose
+  # real npm-publish step can sit behind a required manual approval for a long
+  # time. Waiting for that workflow to conclude means production docs can't go
+  # live before the release they're documenting actually ships (or ships stale
+  # docs for a release that never got approved).
+  workflow_run:
+    workflows: ['Publish Packages (2nd-gen)']
+    types: [completed]
```
Rajdeepc, line 21: This will open a fork-PR code execution path. This fires on every conclusion of that workflow, including skipped. Anyone opening or pushing to a fork PR gets arbitrary code execution with your Azure storage deploy credentials coz this compromise happens during the build.
Please gate the job so workflow_run only reacts to push-triggered upstream runs, and scope the secrets down:
```bash
if: >-
  github.event_name != 'workflow_run' ||
  (github.event.workflow_run.event == 'push' &&
   github.event.workflow_run.conclusion == 'success')
```

then move the three `AZCOPY_*` secrets out of workflow-level `env:` into the two deploy steps only. Longer term, an explicit repository_dispatch from the publish job is safer than ambient `workflow_run`and it also resolves the keyword-collision finding.

This is critical. Kindly fix it

.github/workflows/publish-2ndgen-docs.yml
Rajdeepc: This staging step has no if: at all, so `types: [completed]` fires it for `failure`, `cancelled`, and `skipped` too.

Rajdeepc (changes requested):

.github/workflows/publish-2ndgen-docs.yml
blunteshwar: took your exact suggestion. Added the job-level guard:
```
  if: >-
    github.event_name != 'workflow_run' ||
    (github.event.workflow_run.event == 'push' &&
    github.event.workflow_run.conclusion == 'success')
```
  so workflow_run only proceeds when the upstream run was itself push-triggered and succeeded — a fork-originated pull_request run (the snapshot-release label path) never reaches this job at all now, regardless of its conclusion.

  Also moved the AZCOPY_* secrets out of the workflow-level env: down into just the "Deploy to staging" and "Deploy to production" steps, so install/build steps never have them in their process environment even on a legitimate push-triggered run — belt
  and suspenders in case a dependency script goes looking.

  This also incidentally fixes your other comment about the staging step running on failure/cancelled/skipped: since the job itself won't start unless conclusion == 'success', that step can't run in those cases either. Didn't do the repository_dispatch
  rework — the push-only gate closes the actual hole, so I'm treating that as a separate hardening follow-up rather than blocking this on it. Let me know if you'd rather have it now.

Rajdeepc (approved):
Nothing looks blocking to me. Few nits are still there but lets catch up in the final boss. Good work!

PR #6509, opened 2026-07-14
chore(2nd-gen): prototype automatic @since version badge

miwha-adobe:
## Description

Prototype + proposal (kept in **draft** for feedback) to make the docs "Since X" version badge automatic instead of the hand-typed `@since` that drifts every beta.

> **⚠️ Precondition:** this only works once the Changesets **prerelease command runs on `main`** (the planned move off the `gen2-beta` side branch). The stamp step reads the version from the 2nd-gen `package.json`, and only that move puts the real `2.0.0-beta.N` there. Today `main`'s `package.json` is still the stable line (`0.3.0`). So this is intentionally **not wired into the release flow** yet; it's the proposal + working example, not an active change.

What's here:

- `scripts/stamp-since.js`: at release, stamps every `@since UNRELEASED` with the version from the 2nd-gen `package.json` and freezes it; `--check` flags elements missing `@since`.
- `2nd-gen/packages/swc/.storybook/blocks/StatusBadge.tsx`: renders the `UNRELEASED` sentinel as an "Unreleased" badge.
- `research-versioning-automation.md` + `versioning-badge-proposal.md`: analysis and shareable proposal.

## Motivation and context

`@since` is hand-typed and re-mirrored onto `main` every beta, so it drifts (we just hand-edited 20 components `2.0.0` → `2.0.0-beta.1`, and a related mirror left the core/swc CHANGELOG disagreeing). The proposal is stamp-on-first-release: freeze each component's version once, at release, reading it from `main`'s `package.json`. That dependency is the whole point of the precondition above. The automation lands **with or after** the pre-mode-on-`main` change, not before. See `versioning-badge-proposal.md` for the open questions.

## Related issue(s)

- fixes [SWC-2377]

## Screenshots (if appropriate)

Example of unreleased badge. This would appear on staging if it is not also in production / released.

Screenshot 2026-07-14 at 5 22 53 PM

---

## Author's checklist

- [x] I have read the **[CONTRIBUTING](https://github.com/adobe/spectrum-web-components/blob/main/CONTRIBUTING.md)** and **[PULL_REQUESTS](https://github.com/adobe/spectrum-web-components/blob/main/PULL_REQUESTS.md)** documents.
- [x] I have reviewed at the Accessibility Practices for this feature, see: [Aria Practices](https://www.w3.org/TR/wai-aria-practices/)
- [ ] I have added automated tests to cover my changes.
- [ ] I have included a well-written changeset if my change needs to be published.
- [x] I have included updated documentation if my change required it.

[...]

PR #6638, opened 2026-08-19
feat(release): fully independent pipelines release pattern

blunteshwar:
## Description

Redesigns npm publish/release automation into two fully independent pipelines: `publish.yml` (gen1) and `publish-gen2.yml` (gen2, renamed from `publish-2nd-gen.yml`), each with its own concurrency group, changeset folder, Version PR branch, git tag prefix, and npm dist-tag. A gen1 release can never race or collide with a gen2 release.

Key behavior:
- **Real release**: triggered on push to `main`, scoped by a `paths` filter per generation. Versions via changesets, opens a Version PR (`changeset-release/1st-gen` / `changeset-release/2nd-gen`), publishes once merged, gated by the `npm-publish` environment's required approval.
- **Tagging**: both generations tag every release (`gen1-<version>` / `gen2-<version>`) and check that tag before publishing, so a re-run never republishes. Releases from before this migration (tagged `v<version>`) still resolve via a fallback in the changelog generator.
- **Throwaway snapshot**: gen1 only, via the `snapshot` job (push or manual `workflow_dispatch`), publishing under `next`/`snapshot-test`. Gen2 has no equivalent - once it enters changesets' persistent pre-release mode, `--snapshot` is disallowed and a plain version bump would collide with the real release's version; testing pending gen2 changes means waiting for the Version PR to merge.
- **Branch guard**: `workflow_dispatch` has no branch filter of its own, so both workflows' shared `check-changesets` job now short-circuits (skips) any manual dispatch fired from a branch other than `main` - without it, a manual run from any branch would execute the same version/tag/publish logic against non-main state.
- **Auth**: gen1 publishes purely via OIDC trusted publishing (no npm token needed). Gen2 still uses `ADOBE_BOT_NPM_TOKEN`, required for its `@adobe/*` npm scope.
- **Gen2 npm tag**: real gen2 releases now publish under `latest` instead of `beta`. Versioning is unaffected - gen2 stays in changesets' persistent pre-release mode, so version numbers keep the `2.0.0-beta.N` format; only the npm dist-tag changed.
- Bumped `peter-evans/create-pull-request` from v6 to v8.
- Renamed `publish-2nd-gen.yml` to `publish-gen2.yml` and `publish-2ndgen-docs.yml` to `publish-gen2-docs.yml` for consistent naming.
- `CONTRIBUTOR-DOCS/01_contributor-guides/06_releasing-swc.md` updated to match: dropped the obsolete PR-labeled snapshot-release section, scoped the throwaway-snapshot section to gen1 only.

**Temporary, to remove before merging to `main`** (all bundled in one commit, revert it when ready to merge): `release-strategy` in both workflows' `push` trigger, each Version PR's `base:` pointed at the triggering branch instead of `main`, and two throwaway testing changesets each for gen1/gen2.

## Motivation and context

The previous release workflow shared a single changesets/action call and concurrency group across both generations, so a gen1 release and a gen2 release could overwrite each other's Version PR branch or serialize unnecessarily. This redesign gives each generation its own release lifecycle end-to-end.

Validated with multiple full end-to-end cycles directly on this branch, across several rounds of review from Rajdeepc, caseyisonit, and miwha-adobe. Real bugs found and fixed along the way: an expired npm token, a missing `paths` filter letting a gen1-only push trigger gen2's workflow, a stray git tag from a revert that would've silently blocked the next idempotency check, a version-collision bug in gen2's snapshot job once persistent pre-release mode is active, an OIDC-verify step that never actually gated npm auth for gen1, and `workflow_dispatch` being triggerable from any branch with no guard against running the version/publish logic off non-main state.

## Related issue(s)

- fixes [Issue Number]

---

## Author's checklist

---

## Reviewer's checklist

### Manual review test cases

- [ ] _Descriptive Test Statement_
  1. Go [here](url)
  2. Do this action
  3. Expect this result

- [ ] _Descriptive Test Statement_
  1. Go [here](url)
  2. Do this action
  3. Expect this result

### Device review

- [ ] Did it pass in Desktop?

  1. Go [here](url)
  2. Do this action
  3. Expect this result

  1. Go [here](url)
  2. Do this action
  3. Expect this result


.github/workflows/publish.yml
```diff
@@ -292,14 +208,137 @@ jobs:
...
-          git add .
-          git commit --no-verify -m "chore: release packages #publish" || echo "No changes to commit"
-          git pull --rebase origin main
-          git push
-
-      - name: Create git tag
-        if: steps.extract-tag.outputs.tag == 'latest'
-        run: node ./1st-gen/scripts/create-git-tag.js
+          yarn workspace @spectrum-web-components/1st-gen changelog:1st-gen
+          (cd 1st-gen && yarn changeset version --snapshot test)
```
Rajdeepc: Is this temp for testing? PR is targeting main so need to be diligent here.

.github/workflows/publish-2nd-gen.yml
```diff
@@ -0,0 +1,244 @@
...
+
+      - name: Version gen2 packages
+        if: needs.check-changesets.outputs.has_changesets == 'true'
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          if [ ! -f 2nd-gen/.changeset/pre.json ]; then
+            (cd 2nd-gen && yarn changeset pre enter beta)
+          fi
+          (cd 2nd-gen && yarn changeset version --snapshot test)
```
Rajdeepc: Is this temp for testing? PR is targeting main so need to be diligent here.

.github/workflows/publish.yml
```diff
@@ -1,77 +1,79 @@
-name: Publish Packages
+name: Publish Packages (gen1)
 
 on:
   workflow_dispatch:
     inputs:
       tag:
-        description: 'NPM dist-tag (e.g., latest, beta, snapshot)'
```
Rajdeepc, line 7: Update the description too.

.github/workflows/publish.yml
```diff
@@ -1,77 +1,79 @@
...
-  pull_request:
-    types: [labeled, synchronize]
+        type: choice
+        options:
+          - next
+          - snapshot-test
   push:
     branches:
       - main
+      - release-strategy # TODO: remove before merging - testing only
```
Rajdeepc: Noting it here. Please remove once testing is done

.github/workflows/publish.yml
```diff
@@ -292,14 +208,137 @@ jobs:
...
+          yarn install --refresh-lockfile
+
+      - name: Open or update the gen1 Version PR
+        if: needs.check-changesets.outputs.has_changesets == 'true'
+        uses: peter-evans/create-pull-request@v6
+        with:
+          commit-message: 'chore: release gen1 packages'
+          title: 'chore: release gen1 packages'
+          branch: changeset-release/1st-gen
+          base: ${{ github.ref_name }}
```
Rajdeepc: Pin `main` here once testing is done.

.github/workflows/publish.yml
```diff
@@ -1,77 +1,79 @@
...
 
 jobs:
   check-changesets:
     runs-on: ubuntu-latest
-    if: >-
-      github.event_name == 'workflow_dispatch' ||
-      github.event_name == 'push' ||
-      (github.event_name == 'pull_request' &&
-      contains(github.event.pull_request.labels.*.name, 'snapshot-release'))
+    if: github.event_name == 'push' || github.event_name == 'workflow_dispatch'
```
Rajdeepc: This is dead now.

Rajdeepc (changes requested):
1. New tagging writes `gen1-<version>` but you are still referring to as below. Kindly fix this.
```ts
const currentTag = `v${currentVersion}`;
```
2. Kindly clean up the orphaned changesets before merging.
3. In `CONTRIBUTOR-DOCS/01_contributor-guides/06_releasing-swc.md` remove `### Snapshot release (for testing a PR)` section. This is obsolete now.

.github/workflows/publish-gen2.yml
caseyisonit, line 1: file name should be gen2 not 2nd-gen

caseyisonit: check throughout the workflows as well, theres several lines to clean up there as well

.github/workflows/publish-2nd-gen.yml
```diff
@@ -0,0 +1,242 @@
...
+  # publish once that PR merges. Every push to main either versions and opens/updates a
+  # dedicated Version PR (changesets pending) or publishes (no changesets pending,
+  # because that PR was just merged). 2nd-gen has no `latest` channel yet - `beta` is the
+  # only real release.
+  #
+  # This does NOT use changesets/action's own built-in PR management - see the identical
+  # comment above the `release:` job in publish.yml for why: changesets/action's PR
+  # branch name is derived only from the base branch, and both this workflow and
+  # publish.yml target `main`, so two changesets/action calls would collide on the same
+  # branch name and overwrite each other's Version PR content. peter-evans/create-pull-request@v6's
```
caseyisonit: why are we locking in at v6 of this action when its presently release a v8? Also should we just fork this so we own it, otherwise i think this might get flagged by security for using a third party action that has permissions we likely dont want to be giving.

.github/workflows/publish-2nd-gen.yml
```diff
@@ -0,0 +1,242 @@
...
+            echo "❌ Publish failed — NPM authentication did not succeed" >> $GITHUB_STEP_SUMMARY
+          fi
+
+  # Real release: version-then-PR for the persistent `beta` pre-release channel, then
+  # publish once that PR merges. Every push to main either versions and opens/updates a
+  # dedicated Version PR (changesets pending) or publishes (no changesets pending,
+  # because that PR was just merged). 2nd-gen has no `latest` channel yet - `beta` is the
+  # only real release.
+  #
+  # This does NOT use changesets/action's own built-in PR management - see the identical
```
caseyisonit: can we not have comments referencing other comments in other files? either remove or copy the comment over

.github/workflows/publish-gen2.yml
```diff
@@ -0,0 +1,242 @@
...
+  # because that PR was just merged). 2nd-gen has no `latest` channel yet - `beta` is the
+  # only real release.
+  #
+  # This does NOT use changesets/action's own built-in PR management - see the identical
+  # comment above the `release:` job in publish.yml for why: changesets/action's PR
+  # branch name is derived only from the base branch, and both this workflow and
+  # publish.yml target `main`, so two changesets/action calls would collide on the same
+  # branch name and overwrite each other's Version PR content. peter-evans/create-pull-request@v6's
+  # explicit `branch:` input avoids that.
+  release:
```
caseyisonit: should this be named beta-release since that is the tag we are releasing, that would make this more scalable as we introduce other tags we may want to use

.github/workflows/publish-gen2-docs.yml
```diff
@@ -10,28 +10,53 @@ on:
...
-env:
-  AZCOPY_AUTO_LOGIN_TYPE: SPN
-  AZCOPY_SPA_APPLICATION_ID: ${{ secrets.AZURE_CLIENT_ID }}
-  AZCOPY_SPA_CLIENT_SECRET: ${{ secrets.AZURE_CLIENT_SECRET }}
-  AZCOPY_TENANT_ID: ${{ secrets.AZURE_TENANT_ID }}
+  # Docs deploys are cheap to queue and expensive to interrupt: azcopy sync
+  # --delete-destination deletes stale files and uploads new ones in the same
+  # run, so cancelling one mid-flight can leave the public site (staging or
+  # prod) with files deleted but not yet replaced.
+  cancel-in-progress: false
```
caseyisonit, line 32: can you explain this change more? What are the potential side effects of this never being canceled?

.github/workflows/publish-gen2-docs.yml
```diff
@@ -73,11 +98,47 @@ jobs:
...
 
-      - name: Deploy to Azure Blob Storage
+      # Every trigger of this workflow refreshes the staging path (docs-staging/) -
+      # push, workflow_dispatch, and the workflow_run that follows a real 2nd-gen
+      # release. It's cheap to re-sync and always safe to keep it tracking main.
+      #
+      # AZCOPY_* creds are scoped to just this step and "Deploy to production"
+      # below, not the whole job, so earlier steps (install, build) never have
+      # them in their process environment.
+      - name: Deploy to staging
```
caseyisonit, line 104: Does this need a conditional check for when to run? Currently, this will always run, including on a production release.

.github/workflows/publish.yml
```diff
@@ -1,77 +1,77 @@
...
+  # Throwaway snapshot publish: gated on has_changesets, same as check-changesets - so
+  # this only runs on a push to main that happens to have 1st-gen changesets pending at
+  # that moment, not on every push. `next` is therefore not a true continuous build of
+  # every main commit; it only refreshes when a push coincides with a pending changeset.
+  # Also runs for manual workflow_dispatch. Never touches main, never commits, never
+  # opens a PR. Runs alongside (not instead of) the
+  # `release` job below - both are triggered by the same push event but never conflict
+  # (separate jobs/runners, and the run-level concurrency group only dedupes overlapping
+  # *workflow runs*, not jobs within one run).
+  snapshot:
```
caseyisonit, line 77: make this more semantic so when this is in the checks its clear which gen is snapshotting

.github/workflows/publish-gen2.yml
blunteshwar: Noted

.github/workflows/publish-2nd-gen.yml
blunteshwar: Makes sense

.github/workflows/publish-gen2.yml
blunteshwar: There are 2 jobs here, one is the snapshot - released on every push to main or by manual dispatch - we can provide any release tag for this job (default: `next`), other is release - the actual release which happens when we merge the version PR. When we want to release other tags like `latest`, we have to remove the `pre-release mode` from this job and use the snapshot job for beta releases at that time.

.github/workflows/publish-gen2-docs.yml
blunteshwar: The `workflow_run` trigger was added so the **production** docs deploy is driven by the real gen2 release actually finishing, not by the push that merely opens/updates the Version PR. That push can sit behind a required manual approval on `npm-publish` for a while, so without this, production docs could go live (or claim to) before the release they document has actually shipped.

On the "never canceled" point: `cancel-in-progress: false` is set at the job's `concurrency.group: publish-2ndgen-docs`, which is shared across every trigger (push, `workflow_dispatch`, and `workflow_run`). The reason is `azcopy sync --delete-destination=true`: it deletes stale files and uploads new ones in the same run. If GitHub canceled a run mid-sync, the live site (staging or prod) could be left with files deleted but the new ones never uploaded - a broken site. `cancel-in-progress: false` trades that risk for queuing: overlapping triggers run sequentially instead of interrupting each other.

Side effect to be aware of: if multiple triggers land close together (e.g. a push and a `workflow_run` conclusion around the same time), they queue up and each runs to completion in order, so a later, more-relevant deploy can be delayed behind an earlier, now-superseded one. That's a "runs slower than optimal" cost, not a correctness risk - the alternative (cancel-in-progress) is what risks corrupting the live site.

.github/workflows/publish-gen2-docs.yml
blunteshwar: "Deploy to staging" is intentionally unconditional - it's meant to track `main` continuously, so it runs the same way on every trigger: push, manual dispatch, and every `workflow_run` conclusion (including production-release runs). That's by design, not an oversight.

"Deploy to production" (the step just below it) already has its own gate:
```yaml
if: >-
  github.event_name == 'workflow_dispatch' ||
  contains(github.event.workflow_run.head_commit.message, '#gen2-publish')
```
So production only deploys on a manual run or a `workflow_run` whose upstream commit message contains `#gen2-publish` - i.e. an actual real release, not a run that only opened/updated the Version PR. Staging always refreshes; production only refreshes on a real release. If the concern was that staging shouldn't redeploy on a production-triggering run too, that's actually intended - staging is meant to always mirror `main`, regardless of what kind of run triggered the workflow.

.github/workflows/publish.yml
blunteshwar: The name which will be shown in the checks is `<workflow name> / <job id>` that is `Publish Packages (gen1) / snapshot`. Also it makes less sense to name workflows like `snapshot-gen1 or snapshot-gen2` when these jobs are in their own perfectly distinguishable workflows

.github/workflows/publish-2nd-gen.yml
blunteshwar: Updated the version to v8

.github/workflows/publish.yml
blunteshwar: yes, this was for testing

.github/workflows/publish-2nd-gen.yml
blunteshwar: for testing only
Updated now

.github/workflows/publish-gen2.yml
```diff
@@ -0,0 +1,254 @@
...
+        id: npm-auth
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.ADOBE_BOT_NPM_TOKEN }}
+        run: npm whoami --registry=https://registry.npmjs.org
+
+      - name: Version packages
+        working-directory: 2nd-gen
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: yarn changeset version
```
Rajdeepc: Blocker: This will cut and exit `beta` and move us to `2.0.0`. You can skip the gen2 snapshot job entirely while pre mode is active. `beta` already is in our prerelease channel so a parallel `next` snapshot adds little.

Rajdeepc (changes requested):
You are still reading `v${nextVersion}` in updated-changelog. Please verify

.github/workflows/publish-2nd-gen.yml
Rajdeepc: This is ok but I am not worried about we take in latest when available.

blunteshwar:
> `v${nextVersion}`

Updated

.github/workflows/publish-gen2.yml
blunteshwar: I have removed the `snapshot` job entirely from gen2 release workflow

.github/workflows/publish-2nd-gen.yml
caseyisonit: we should take the latest when implementing. im not concerned with adding it to an update cadence until we bump node versions across our CI which is something coming up.

.github/workflows/publish-gen2.yml
caseyisonit: can we talk this out based on the slack thread regarding npm tags? i thought we were moving away from `beta` to only have `latest` / `next` / `snapshot`. i might run this through claude to create some diagrams of the release process and the different tags.

miwha-adobe:
Wondering what you think about fitting in the `@since` badge automation from this prototype I built last month: [#6509](https://github.com/adobe/spectrum-web-components/pull/6509). Right now the `@since tag` is hand authored and easily missed. At the time we discussed rolling it into the release strategy. I feel like it could tie in nicely here

miwha-adobe:
Nit: The PR description seems a bit out of sync. `publish-2nd-gen.yml` should be  `publish-gen2.yml`. Maybe worth re-review with claude as some context has changed.

.github/workflows/publish.yml
```diff
@@ -1,77 +1,77 @@
...
+# gen1-scoped group: this and publish-gen2.yml used to share the literal string
+# `publish-${{ github.ref }}`, which is a repo-wide concurrency key, not scoped to the
+# workflow file it's declared in - so the two workflows' entire runs (not just their
+# release jobs) serialized against each other on every push to main. That made sense
+# under the old shared-changeset/direct-push model (both workflows wrote to the same
+# main branch and had to avoid racing), but it directly works against this redesign's
+# goal of fully independent gen1/gen2 release streams: neither writes to anything the
+# other touches anymore (separate Version PR branches, separate dist-tags, separate npm
+# packages), so there's nothing left to protect by blocking one generation's run behind
+# the other's.
```
miwha-adobe: This feels like too much context. Can we ask claude to be more concise here.

.github/workflows/publish.yml
```diff
@@ -83,41 +83,15 @@ jobs:
...
-        env:
-          TAG: ${{ steps.extract-tag.outputs.tag }}
-          GIT_REF: ${{ github.ref }}
-        run: |
-          if [ "$TAG" == "latest" ] && [ "$GIT_REF" != "refs/heads/main" ]; then
-            echo "ERROR: Cannot publish 'latest' from non-main branch"
-            exit 1
-          fi
-
       - name: Verify OIDC token and configure NPM authentication
```
miwha-adobe: This step verifies an OIDC token for "trusted publishing," but the publish then authenticates with `ADOBE_BOT_NPM_TOKEN (_authToken npmrc + NODE_AUTH_TOKEN)`. Nothing actually consumes the OIDC token. This is effectively a no-op that reads like a security feature.

Is trusted publishing the goal, or can this come out?

blunteshwar: Trusted publishing is the goal here (it's configured on npmjs.com for filename: `publish.yml / environment: npm-publish)`, so the token should come out entirely rather than the OIDC check. Gen1 doesn't need `ADOBE_BOT_NPM_TOKEN` at all — that token is only actually required for gen2, since `@adobe`-scoped packages need it to publish under that npm org namespace.

Fixed by dropping `ADOBE_BOT_NPM_TOKEN`, gen1 now authenticates purely via OIDC, matching what's actually configured.

.github/workflows/publish.yml
blunteshwar: You are right! Trimmed the comment to only what is needed

Rajdeepc (approved):
Good design.

.github/workflows/publish.yml
```diff
@@ -292,14 +187,123 @@ jobs:
...
+              wait_time=$((2 ** attempt))
+              echo "✗ Publish failed for $dir, retrying in ${wait_time}s..."
+              sleep $wait_time
+              attempt=$((attempt + 1))
+            done
+          done
+          cd ../..
+          git config user.name "github-actions-bot"
+          git config user.email "support+actions@github.com"
+          node ./1st-gen/scripts/create-git-tag.js
```
rubencarvalho: i think the tag only gets created after the whole react loop finishes. so if one wrapper fail (after its 3 retries), it exits with no tag, and some wrappers are already on npm. so on the next run: changeset publish is fine, but the loop hits `npm publish` on an already-published version -> errors -> we're stuck. wonder if we should skip-if-already-published inside the loop instead of relying on the tag.

.github/workflows/publish.yml
```diff
@@ -292,14 +187,123 @@ jobs:
...
-        if: steps.extract-tag.outputs.tag == 'latest'
+  # Real release: version-then-PR, then publish once that PR merges. Uses
+  # peter-evans/create-pull-request instead of changesets/action's built-in PR
+  # management, since that derives its branch name only from the base branch
+  # (`changeset-release/main`) - gen1 and gen2 would collide on the same branch.
+  release:
+    needs: check-changesets
+    if: github.event_name == 'push'
+    runs-on: ubuntu-latest
+    environment: npm-publish
```
rubencarvalho: do i see this correctly? the whole `release` job sits behind `environment: npm-publish`, but most of the time it's only opening/updating the Version PR (nothing published). so every changeset that lands needs someone to click approve just to get the PR opened? might be nicer to split the version-PR bit (no environment) from the actual publish (environment)?

.github/workflows/publish.yml
```diff
@@ -292,14 +187,123 @@ jobs:
...
 
-      - name: Commit and push changes
-        if: steps.extract-tag.outputs.tag == 'latest'
+  # Real release: version-then-PR, then publish once that PR merges. Uses
+  # peter-evans/create-pull-request instead of changesets/action's built-in PR
+  # management, since that derives its branch name only from the base branch
+  # (`changeset-release/main`) - gen1 and gen2 would collide on the same branch.
+  release:
+    needs: check-changesets
+    if: github.event_name == 'push'
```
rubencarvalho: gen1 release only runs on push, but gen2 got workflow_dispatch added. so if a gen1 publish fails after the Version PR merges, there's no button to re-run it, you'd have to push an empty commit. maybe we can give gen1 the same manual trigger?

.github/workflows/publish-gen2.yml
```diff
@@ -0,0 +1,140 @@
...
+            echo "gen2-$RELEASED_VERSION already released - nothing to do."
+            echo "published=false" >> $GITHUB_OUTPUT
+            exit 0
+          fi
+          yarn install --refresh-lockfile
+          yarn build
+          (cd 2nd-gen && yarn changeset publish --no-git-tag --tag beta)
+          git config user.name "github-actions-bot"
+          git config user.email "support+actions@github.com"
+          node ./2nd-gen/scripts/create-git-tag.js
```
rubencarvalho: nit: this script is basically a copy of 1st-gen's create-git-tag.js, only diff is the path + gen1/gen2 prefix. could be one script taking the gen as an arg if we feel like it. no biggie.

Rajdeepc: Though its a nit but I would upgrade it.

.github/workflows/publish.yml
blunteshwar: Great catch. Added a guard to skip already published packages

.github/workflows/publish.yml
blunteshwar: I agree. Your suggested approach makes more sense. Decoupling the release and versioning jobs to implement the same

.github/workflows/publish.yml
blunteshwar: done!!

.github/workflows/publish-gen2.yml
blunteshwar: Done!

blunteshwar:
> Wondering what you think about fitting in the `@since` badge automation from this prototype I built last month: [#6509](https://github.com/adobe/spectrum-web-components/pull/6509). Right now the `@since tag` is hand authored and easily missed. At the time we discussed rolling it into the release strategy. I feel like it could tie in nicely here

Great timing on this — #6638 actually delivers the exact precondition your PR called out: `main's 2nd gen/packages/swc/package.json` now holds the real 2.0.0-beta.N version directly (via publish-gen2.yml's version-pr job), no more gen2-beta side branch or hand-mirroring. That's the gate that kept #6509 as a draft, so this really does tie in nicely now.

 That said, this PR has already grown pretty large across several rounds of review, so I'd rather not fold another feature in on top of it.
Proposal: let's merge this as-is, get the new release pipeline running and verified on main first, then open a focused follow-up PR to wire `stamp-since.js` in — right after yarn changeset version in the version-pr job seems like the natural spot, so the stamped `@since` diffs ride along in the same Version PR commit for review.
Once that follow-up is up I'll loop you in for a look, especially on the open question.

.github/workflows/publish.yml
```diff
@@ -292,14 +194,151 @@ jobs:
...
-          git push
-
-      - name: Create git tag
-        if: steps.extract-tag.outputs.tag == 'latest'
-        run: node ./1st-gen/scripts/create-git-tag.js
+          yarn workspace @spectrum-web-components/1st-gen changelog:1st-gen
+          (cd 1st-gen && yarn changeset version)
+          yarn install --refresh-lockfile
+
+      - name: Open or update the gen1 Version PR
```
miwha-adobe, line 252: Trying to understand where the Version PR token landed. In [#6607](https://github.com/adobe/spectrum-web-components/issues/6607), @Rajdeepc  asked for token: ${{ secrets.RELEASE_BOT_TOKEN }} on create-pull-request (since GITHUB_TOKEN won't trigger the PR's checks); it was added, then removed as "we can trigger the workflow other ways."

Two questions:

- Why did it come back out? 
- What's the "other way"? I couldn't find it in the workflows.  `lint.yml/type-check.yml` only fire on the pull_request event that GITHUB_TOKEN suppresses, and both are required + strict: true. So the bot's Version PR looks like it'd be stuck on "waiting for status" and unmergeable without a manual commit each release.

.github/workflows/publish.yml
```diff
@@ -292,14 +194,151 @@ jobs:
...
+      - name: Version PR summary
+        run: |
+          echo "## Publish summary (gen1)" >> $GITHUB_STEP_SUMMARY
+          echo "📝 Version PR opened/updated at \`changeset-release/1st-gen\` - nothing published yet" >> $GITHUB_STEP_SUMMARY
+
+  # Real publish, once the Version PR above has merged (no pending changesets left).
+  # Only this job sits behind the npm-publish environment's required-reviewer approval,
+  # since it's the only one that actually publishes to npm. Accepts workflow_dispatch
+  # too, so a failed publish can be retried by hand instead of needing an empty commit.
+  publish:
```
miwha-adobe, line 279: Question on the publish job for both gens. It runs whenever `has_changesets == 'false'` and carries environment: `npm-publish` with required reviewers. Since GitHub parks a job for approval before its first step, the reviewer ping seems to fire before the git-tag guard inside gets a chance to no-op. Am I reading that ordering right?

If so, wouldn't any changeset-less merge to 1st-gen/** (a refactor, a chore, a test-only change, a dep bump) trigger the publish job and ask Rajdeepc or Ruben to approve a "release" that then just exits early once the tag check sees the version is already published? A real Version PR merge and one of those unrelated pushes look identical at the if: (both push + has_changesets == 'false'), so I wasn't sure how the approval prompt distinguishes them. My small worry is that frequent phantom approvals could make the gate easy to click through on autopilot.

Would it work to move the tag check up into `check-changesets` as a `needs_publish` output, then gate publish on `has_changesets == 'false' && needs_publish == 'true'`? Same idempotency, but the prompt would only show when there's genuinely something to ship. I think that job's checkout would need fetch-depth: 0 for the tags to be visible.

blunteshwar: You’re spot on! Any commit without a changeset would definitely need approval. Your suggested fix is a great idea and would help keep the pipeline tidy.

.github/workflows/publish.yml
blunteshwar: You're right that the bot's Version PR doesn't get its checks automatically — create-pull-request runs as GITHUB_TOKEN, so the pull_request event that lint.yml/type-check.yml depend on is suppressed, and the required checks sit unreported.

One correction on the remedy: it doesn't need a manual commit. GitHub shows an "Approve and run workflows" banner on the PR, and as soon as a reviewer clicks it, all the checks kick off and report normally and the PR becomes mergeable. I verified this end-to-end by clicking on the button.You can see this #6707 PR ceated by the actions bot during testing. All the tests are running here.

On why the token came back out: going with `secrets.RELEASE_BOT_TOKEN` means creating and owning a PAT/App token and managing it — rotation, expiry, scope, the risk of it going stale. Rather than take on another long-lived secret to maintain, this approach keeps the cost at a single reviewer click on the banner per release — no extra secret, nothing to rotate. Given that release Version PRs are already reviewed by a human before merge, that click folds naturally into the same step.

So: you're correct about the behavior, just that the per-release action is one "Approve and run workflows" click rather than a manual commit.

miwha-adobe: Makes sense to me! Thank you for that explanation.

miwha-adobe (approved):
great work!

caseyisonit (approved):

.github/workflows/publish.yml
rubencarvalho, line 1: Hmmm I noticed `publish-docs-site.yml` only gets triggered on `contains(head_commit.message, '#publish') || contains(..., 'docs:') || contains(..., 'docs(')`
That worked before because gen1's publish job directly committed "chore: release packages #publish", but if I understand correctly, we're not doing that anymore. I _think_ as it stands, a gen1 release will not trigger the gen1 docs deploy automatically.

.github/workflows/publish-gen2.yml
rubencarvalho, line 1: Are we losing the ability to have a `--snapshot` release for Gen2? This would be not ideal... It already helped us a lot during the AI Toolkit development. Can we think of an alternative? Even if it ends up just being a documented prompt to still be able to do it on a given branch?

blunteshwar: Yes, we are losing the ability for a `--snapshot` publish. The primary reason for that is - A snapshot publish can not happen in pre-release mode. We tried it and encountered the following error
`🦋  error Snapshot release is not allowed in pre mode`
`🦋  To resolve this exit the pre mode by running 'changeset pre exit'`
Since we are on Beta in gen2 we can increase the release frequency. Once we decide to exit the pre-release mode latter, we can then introduce the snapshot release pipeline

.github/workflows/publish.yml
blunteshwar: Done!

rubencarvalho (approved):
Approving, with snapshot releases as follow-up 😄

PR #6639, opened 2026-08-19
feat(prompt-field): integrate pixel loader and expose its preset/icon API

rubencarvalho:
## Description

Integrates the pixel loader (#6605) into `<swc-prompt-field>` as its status indicator, and exposes the loader's `preset` / `icon` API through the field. **Stacked on #6630** (`ruben/refactor-prompt-field-swc-buttons`); review/merge that first. The diff here is only the loader integration.

## What changed

- `<swc-prompt-field>` renders `<swc-pixel-loader>` in the status slot at the start of the input row, replacing the placeholder icon. It rests on a settled frame while idle and animates while **`generating`** (`?paused=${!generating}`).
- Two new attributes forward straight to the loader:
  - **`icon`** (default `aiLogo`) — the single icon the loader builds.
  - **`preset`** (`cc`, `dc`, `exp`, `analyze`, `mega`) — a themed icon sequence cycled one per loop; overrides `icon` while set.
- The loader fills the 24px status slot (`--swc-pixel-loader-size`) and inherits the status color (its `currentcolor` default), so it stays the themed black/white.

## Review fixes

- `icon` no longer goes `null` when `preset` is set (removing the attribute tripped the loader's dev validation); it now keeps a valid default and lets `preset` override it, as the loader already documents.
- VRT: the status loader is now paused before every snapshot, so Chromatic doesn't capture it mid-animation and diff against itself on a re-run.
- The loader now stays aligned with the textarea's first line as it grows, instead of centering against the whole multi-line block.

## Accessibility

The status indicator is decorative (`aria-hidden`), so the idle/paused frame is never announced as "Loading"; the generating state is conveyed by the stop button that replaces send (`accessible-label="Stop generating"`).

## Verification

Added `StatusLoaderTest`: `preset` passes through and `generating` un-pauses the loader; clearing `preset` falls back to `icon`, and idle re-pauses. All prompt-field and upload-artifact tests pass (35). Docs page validates.

- [ ] **Keyboard**: no new interactive elements; confirm no focus-order regressions around the input row while a `preset`/`generating` field is present.
- [ ] **Screen reader**: with `generating` toggled, confirm the loader is not announced (decorative) and the stop button's name ("Stop generating") conveys the state; no duplicate "Loading" announcements.


2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/PromptField.ts
```diff
@@ -105,6 +110,45 @@ export class PromptField extends SpectrumElement {
   @property({ type: String, reflect: true })
   public variant: 'subtle' | 'balanced' | 'prominent' = 'balanced';
 
+  /** Status loader artwork: a single icon name (static), or a preset name (`cc`, `dc`, `exp`, `analyze`, `mega`) that cycles a themed sequence. Unions inlined so the API table lists the values; routed to the loader's icon/preset in `_renderStatusIcon`. */
+  @property({ type: String, reflect: true })
```
miwha-adobe, line 117: **Claude Suggestion:** The guard only runs one way right now. `const icon: PixelLoaderIconName = this.icon` catches values the loader doesn't have, but not the reverse. So if someone adds a new icon to the loader's catalog later, this union quietly falls out of sync with no build error, and there'd be no way to set that icon through `<swc-prompt-field>`.

What if we add a quick bidirectional check? That keeps `PixelLoaderIconName` as the source of truth while still listing the values in docs:

```
    type Exact<A, B> = [A] extends [B] ? ([B] extends [A] ? true : never) : never;
    const _iconsExhaustive: Exact<PromptField['icon'], PixelLoaderIconName> = true;
    const _presetsExhaustive: Exact<NonNullable<PromptField['preset']>, PixelLoaderPresetName> = true;
```

2nd-gen/packages/swc/patterns/conversational-ai/upload-artifact/upload-artifact.css
miwha-adobe, line 46: Nit: This renames a consumer-facing custom prop  `--swc-upload-artifact-dismiss-inline-size/-block-size` are collapsed into a single `--swc-upload-artifact-dismiss-visual-size` and also removes the `--swc-prompt-field-send-*/-upload-*/-stop-*/-artifact-scroll-button-*` size overrides.

Is anyone consuming these yet? If there are consumers, might be worth a changeset note so they're not caught off guard.

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/PromptField.ts
rubencarvalho: Ah, thank you for catching this. So I was going back and forth debugging + finding the best solution and left this committed. I've reverted this because it will need to be tackled project-wide: https://github.com/adobe/spectrum-web-components/pull/6643

2nd-gen/packages/swc/patterns/conversational-ai/upload-artifact/upload-artifact.css
rubencarvalho: Great call! We still have no consumers, and we’re iterating so much on these components that we’re holding off on changesets (and therefore changelog entries) for now, as they might introduce more noise than useful information at this stage. :)

TarunAdobe (approved):
all looks good!!

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/PromptField.ts
```diff
@@ -1072,11 +1081,25 @@ export class PromptField extends SpectrumElement {
...
+    );
+    const preset = isPreset
+      ? (this.loader as PixelLoaderPresetName)
+      : undefined;
+    const icon = isPreset ? undefined : (this.loader as PixelLoaderIconName);
     return html`
       <span class="swc-PromptField-status-icon" aria-hidden="true">
-        <swc-ui-icon icon="asterisk" size="xl"></swc-ui-icon>
+        <swc-pixel-loader
+          icon=${ifDefined(icon)}
```
Rajdeepc: icon goes null when a preset is selected.
Please stop letting the attribute disappear by handing the loader a valid icon at all times and let preset dow the overriding which is what the loader already documents.
```ts
const DEFAULT_LOADER_ICON: PixelLoaderIconName = 'aiLogo';
// …
const preset = isPreset ? (this.loader as PixelLoaderPresetName) : undefined;
// `preset` overrides `icon` while set, so keep a valid icon on the element
// rather than removing the attribute — a removed String attribute lands as
// `null` on the property and trips the loader's dev validation.
const icon = isPreset
  ? DEFAULT_LOADER_ICON
  : (this.loader as PixelLoaderIconName);
```
then icon=${icon}. Dont keep ifDefined, keep it on preset and reuse `DEFAULT_LOADER_ICON` as the loader property default so the two can't drift.

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/PromptField.ts
```diff
@@ -1072,11 +1081,25 @@ export class PromptField extends SpectrumElement {
...
+          preset=${ifDefined(preset)}
+          ?paused=${!this.generating}
```
Rajdeepc, line 1107: This will put an infinite WAAPI loop into the Chromatic snapshot. Can you run Chromatic on this branch twice with no code change. If the permutations story diffs against itself, you need to fix this in the VRT story as below and not the component.
```ts
// in test/vrt/prompt-field.vrt.ts, alongside forceButtonStates
const pauseLoaders = async ({ canvasElement }: { canvasElement: HTMLElement }) => {
  canvasElement.querySelectorAll('swc-prompt-field').forEach((field) => {
    const loader = field.shadowRoot?.querySelector('swc-pixel-loader');
    if (loader) (loader as PixelLoader).paused = true;
  });
};
```
You can compose it into the existing play: forceButtonStates. This is safe here.

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/prompt-field.css
```diff
@@ -500,30 +500,37 @@
...
+  padding-inline-start: token("spacing-75");
+  padding-inline-end: calc(token("component-height-100") + token("spacing-300"));
 }
 
 .swc-PromptField-status-icon {
   display: flex;
   flex: 0 0 auto;
   align-items: center;
-  align-self: start;
+  align-self: center;
```
Rajdeepc: This might put the loader float to the vertical middle of a grown text-area instead of the beside the first line. This should be on the group and not here.

Rajdeepc (changes requested):
VRT flake is somewhat blocking if it comes in double Chromatic run. Once you verify it @ me for a final pass.

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/prompt-field.css
rubencarvalho: Great catch!! Updated it.

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/PromptField.ts
rubencarvalho: Applied the suggestion! :) I hoped Chromatic was good with their anti-flake measures, but this is safer!

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/PromptField.ts
rubencarvalho: oooh crazy, nice catch. updated 👍

Rajdeepc (approved):

PR #6643, opened 2026-08-24
build(cem): expand type aliases in the manifest via type-parser

rubencarvalho:
## Description

Expands referenced TypeScript aliases in the Custom Elements Manifest (CEM), so Storybook generates literal API types and `select` controls without duplicate `argTypes` wiring.

### Root cause

The CEM stored alias names instead of their literal values, so Storybook could not infer options.

## What changed

- Added `@wc-toolkit/type-parser` to expand aliases into `parsedType`.
- Preserved core-package modules by matching resolved absolute paths.
- Narrowed Card’s CEM size type to its supported values.
- Removed redundant story `argTypes`; retained only custom behavior.
- Removed the ineffective CEM watcher from the default Storybook start command.
- Added story guidance and a changeset.

## Intentionally left as manual wiring

Manual `argTypes` remain only for empty/undefined sentinels, custom ranges or ordering, and internal attributes that Storybook must not round-trip.

## Why drop the CEM analyze watch from `start`

Storybook reads the manifest only at startup. The watcher reanalyzed files but could not update the running preview.

## Performance note

CEM analysis increases from about 1.1s to 3.5s, once per Storybook start.

## Verification

- CEM generation and direct `parsedType`, inherited-member, default, and sentinel assertions pass.
- 1,141 Storybook tests pass.
- Gen 2 type-check, lint, formatting, and documentation checks pass.

## Notes for reviewers

Restart Storybook, then confirm affected API tables show literal values and inferred controls remain correct.


2nd-gen/package.json
```diff
@@ -21,7 +21,7 @@
     "dev:analyze": "yarn workspace @adobe/spectrum-wc analyze:watch",
     "dev:core": "yarn workspace @adobe/spectrum-wc-core dev",
     "lint": "cd .. && eslint 2nd-gen --ext .ts,.js,.json",
-    "start": "run-p dev:core dev:analyze storybook",
+    "start": "run-p dev:core storybook",
```
rubencarvalho, line 24: I was very surprised by this because I thought CEM "HMR" was working, but apparently Storybook only reads the manifest once, at preview boot (preview.ts → setCustomElementsManifest). So the watcher wasn't giving live updates, it was just re-running a full TypeScript program pass (now heavier with type-parser wired in, ~3.5s vs ~1.1s before) on every save for no visible benefit.

nikkimk:
Should we include updates to the following in this PR or as a fast follow?
- badge.stories.ts:43-46 (variant, backed by BadgeVariant = (typeof BADGE_VARIANTS)[number])
- status-light.stories.ts:39-42 (variant)
- meter.stories.ts:39-42 (variant)
- button-group.stories.ts:42-45 and :52-55 (orientation, align)
- accordion.stories.ts:64-67 (density) — this one's notable because it's the exact same property (density) on the exact same shape as card.density, which this PR did clean up.

.changeset/workflow-icons-package.md
```diff
@@ -7,5 +7,5 @@ Add `@adobe/spectrum-wc-icons`, a new public package delivering the Spectrum 2 *
...
 - **API**: every element extends `IconBase`, so it carries `size` (`xs`–`xl`) and host-owned accessibility (`accessible-label` sets `role="img"` + `aria-label`; empty is decorative `aria-hidden`). Color follows CSS `color` with a `--swc-icon-color` override.
 - **Tree-shaking**: per-icon subpath exports (`@adobe/spectrum-wc-icons/swc-icon-star.js`, `@adobe/spectrum-wc-icons/Star.js`) mean an app ships only the icons it imports; a register-all `elements.js` and a `manifest.js` (name + tag list) are provided for galleries and pickers.
-- **Naming**: for an A4U logical name `<Name>`, the function is `Icon_<Name>()`, the element class `Icon<Name>`, and the tag `swc-icon-<kebab>` (`AddCircle` → `<swc-icon-add-circle>`, `3DAsset` → `<swc-icon-3d-asset>`).
+- **Naming**: for an A4U logical name `<Name>`, the function is `Icon_`<Name>`()`, the element class `Icon`<Name>`, and the tag `swc-icon-`<kebab>` (`AddCircle` → `<swc-icon-add-circle>`, `3DAsset` → `<swc-icon-3d-asset>`).
```
caseyisonit, line 10:
```suggestion
- **Naming**: for an A4U logical name `<Name>`, the function is `Icon_`<Name>`()`, the element class `Icon`<Name>``, and the tag `swc-icon-`<kebab>`` (`AddCircle` → `<swc-icon-add-circle>`, `3DAsset` → `<swc-icon-3d-asset>`).
```

I dont think this change is needed but if it is there were missing back ticks

2nd-gen/packages/core/components/card/Card.base.ts
```diff
@@ -69,6 +69,13 @@ import {
 export abstract class CardBase extends SizedMixin(SpectrumElement, {
   validSizes: CARD_VALID_SIZES,
 }) {
+  /**
+   * The size of the card.
```
caseyisonit, line 73: future clean up: why do we have to declare this every time we use the sizedMixin? That should be coming from the mixin as well as the jsdoc comment. Also we should be using the default size type that is present in the mixin when that aligns for the component so we start to learn which component deviate from the default size. we are currently writing size types for every component instead of reusing.

2nd-gen/packages/swc/components/button-group/stories/button-group.stories.ts
```diff
@@ -31,34 +31,12 @@ const { args, argTypes } = getStorybookHelpers('swc-button-group');
 
 argTypes.size = {
   ...argTypes.size,
-  control: { type: 'select' },
-  options: [...BUTTON_GROUP_SIZES],
   table: {
     category: 'attributes',
     defaultValue: { summary: 'm' },
   },
 };
```
caseyisonit, line 38: For throughout the diff: Why cant we remove this entire argTypes definition since size is defo a union type and this PR is attempting to resolve the need for this pattern?

2nd-gen/packages/swc/components/popover/stories/popover.stories.ts
caseyisonit, line 38: missed removal?

2nd-gen/packages/swc/patterns/ai-toolkit/pixel-loader/stories/pixel-loader.internal.stories.ts
caseyisonit, line 31: should this not be removed as well?

caseyisonit (commented):
Gave a quick breeze through this and think theres still more clean that can happen. i loveeeee the intention of this and super excited. I also beg the question of why any argTypes should be declared in stories, storybook can infer a lot of controls based on types i.e. booleans etc. can we clean those up too in this PR? if not can we make a ticket to come back and do that?

nikkimk (commented):
Left a comment here: https://github.com/adobe/spectrum-web-components/pull/6643#issuecomment-5456870968

rubencarvalho:
@nikkimk Thanks for the follow-up list. The final branch removes the redundant overrides for badge/status-light/meter variants, button-group orientation and alignment, and accordion density! 😄

.changeset/workflow-icons-package.md
rubencarvalho: oops, reverted

gen2/packages/core/components/card/Card.base.ts
```diff
@@ -69,6 +69,13 @@ import {
 export abstract class CardBase extends SizedMixin(SpectrumElement, {
   validSizes: CARD_VALID_SIZES,
 }) {
+  /**
+   * The size of the card.
```
rubencarvalho: in this case we needed to redeclare it because it's narrowing the sizes. I do think we can simplify the sizedmixin in a dedicated PR though!

2nd-gen/packages/swc/components/button-group/stories/button-group.stories.ts
rubencarvalho: done

rubencarvalho:
> Gave a quick breeze through this and think theres still more clean that can happen. i loveeeee the intention of this and super excited. I also beg the question of why any argTypes should be declared in stories, storybook can infer a lot of controls based on types i.e. booleans etc. can we clean those up too in this PR? if not can we make a ticket to come back and do that?

Tried to do all I could find and were possible to change!

PR #6644, opened 2026-08-24
feat(prompt-input): add drag and drop functionality

TarunAdobe:
## Description

  Prompt field can now accept a dragged file anywhere over the composer, not just through the `+` button's picker flow. Dragging over the field reuses the textarea's own focus-ring look rather than introducing a separate drop-target treatment, and dropping fires `swc-prompt-field-drop` with the raw files so consumesize/filled/dropEffect API. `DropzoneBase` now delegates to it internally; its public behavior is unchanged.

  Wired the new event into the pattern's full end-to-end example too.

  ## Motivation and context

  The `+` button's external picker flow was the only way to attach a file to `swc-prompt-field`, which is a heavier interaction than the drag-and-drop most conversational AI composers support today. This adds that path without duplicating the drag timing logic Dropzone already had battle-tested, by extracting it into a shared controller.

  ## Related issue(s)

  - fixes [Issue Number]

  ---

  ## Author's checklist

  ---

  ## Reviewer's checklist

  ### Manual review test cases

  - [ ] _Drag a file over the prompt field shows the focus-ring drop affordance_
    1. Go to Storybook → Conversational AI → Prompt field → Drag and drop
    2. Drag a file from your desktop over the field without dropping it
    3. Expect the field's border to switch to the same focus-indicator color the textarea shows on `:focus-visible`, with no separate drop-target styling

  - [ ] _Dropping a file fires `swc-prompt-field-drop` and slots an artifact_
    1. Go to Storybook → Conversational AI → Prompt field → Drag and drop
    2. Drop a file onto the field
    3. Expect the field to log `swc-prompt-field-drop delivered 1 file...` and slot a media upload artifact, the same as using the `+` button's external picker

  - [ ] _Disabled field rejects the drag_
    1. Go to Storybook → Conversational AI → Prompt field → Playground, set `mode="disabled"`
    2. Drag a file over the field
    3. Expect the OS cursor to show "not allowed" and no focus-ring affordance to appear

  - [ ] _Dropzone's own behavior is unchanged after the controller extraction_
    1. Go to Storybook → Dropzone
    2. Drag a file over the zone, drag it out, and drop it
    3. Expect `swc-dropzone-should-accept`, `swc-dropzone-dragover`, `swc-dropzone-dragleave`, and `swc-dropzone-drop` to fire exactly as before, with the same 100ms dragleave debounce

  - [ ] _Full pattern example wires the new event_
    1. Go to Storybook → Conversational AI → Pattern overview → Full example
    2. Drop a file directly onto the composer instead of using the `+` button
    3. Expect the file to appear as an attached artifact in the conversation, identically to the upload-button path

### Device review

- [ ] Did it pass in Desktop?

  1. Go to Storybook → Conversational AI → Prompt field → Accessibility
  2. Tab through the field with no drag in progress
  3. Expect focus order and the `+` button's operability to be unaffected; drag-and-drop is a progressive enhancement only, the `+` button remains the fully keyboard-operable path for attaching files since native HTML drag-and-drop has no keyboard equivalent

  1. Go to Storybook → Conversational AI → Prompt field → Drag and drop
  2. With a screen reader running, drop a file onto the field
  3. Expect no unexpected or duplicate announcements from the drag affordance itself; the resulting artifact should announce the same way it does when added via the `+` button's picker flow


2nd-gen/packages/core/controllers/drag-and-drop-controller/src/drag-and-drop-controller.ts
```diff
@@ -0,0 +1,176 @@
...
+    this._clearDragLeaveTimer();
+  }
+
+  private readonly _onDragOver = (event: DragEvent): void => {
+    event.preventDefault();
+    if (!event.dataTransfer) {
+      return;
+    }
+
+    if (this._options.shouldAccept?.(event) === false) {
```
rubencarvalho, line 130: When `shouldAccept` flips to false mid-drag (e.g. a consumer's should-accept listener starts vetoing based on changing app state), we only swap the cursor to 'not-allowed' here. `_dragged` never gets reset and `onDragLeave` never fires, so `_onDrop` below (which only checks `isDragged()`, not `shouldAccept`) still lets the drop through. Claude says that PromptField happens to be safe because its own `onDrop` double-checks `this.disabled` before dispatching, but Dropzone doesn't have an equivalent guard, so a consumer who rejects mid-hover via the should-accept event would still get a `drop` fired on them. might be worth resetting dragged state (or calling `onDragLeave`) here too, so the controller is safe by default rather than relying on every consumer to re-check? could be missing something, lmk what you think 🙂

2nd-gen/packages/core/controllers/drag-and-drop-controller/drag-and-drop-controller.mdx
```diff
@@ -0,0 +1,108 @@
...
+### Basic usage
+
+1. Construct the controller in the host's constructor.
+2. Return the host's current dragged state from `isDragged`.
+3. Update that state from `onDragEnter` and `onDragLeave`.
+4. Handle accepted drops in `onDrop`.
+
+```typescript
+import { LitElement } from 'lit';
+import { DragAndDropController } from '@adobe/spectrum-wc-core/controllers/index.js';
```
rubencarvalho: random side thing i noticed while reading the docs... every other controller in this package (color, hover, focus-group-navigation, etc.) gets its own dedicated `./controllers/<name>.js` entry in core's package.json exports/typesVersions. this PR doesn't add one for drag-and-drop-controller, so the only way to import it is through the barrel `controllers/index.js` (which is what this example does). probably just needs the exports map entry added to match the pattern?

rubencarvalho (approved):
Approving! But if you can tackle the little nits I left, it would perfect!

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/prompt-field.css
```diff
@@ -458,6 +463,38 @@
...
+ * generating state. */
+:host .swc-PromptField-outer-border.dragged {
+  --_swc-prompt-field-drag-fill: light-dark(color-mix(in oklch, white, oklch(from var(--swc-prompt-field-brand-color) 91.32% 0.0721 calc(h + 0.1)) 30%), color-mix(in srgb, token("background-layer-2-color") 85%, token("static-fuchsia-600") 15%));
+  --_swc-prompt-field-drag-focus-indicator-color: token("indigo-800");
+  --_swc-prompt-field-outer-stop-1: var(--_swc-prompt-field-outer-recede-stop);
+  --_swc-prompt-field-outer-stop-2: var(--_swc-prompt-field-outer-recede-stop);
+  --_swc-prompt-field-outer-stop-3: var(--_swc-prompt-field-outer-recede-stop);
+
+  box-shadow: none;
+  transition: none;
```
5t3ph: I agree with this, but the transition is still active for "drag off / drag complete" (default state) which is a little jarring to have it "pop" back into place:

https://github.com/user-attachments/assets/086819cb-5b5f-4485-afd7-25679cfa922e

5t3ph (changes requested):
1. Upon drop, if keyboard focus is _not_ in the prompt input, it should probably be placed on the dropped attachment element. Something to add to the demo, and probably a11y guidance too since that will be consumer controlled?
2. Something about the Behaviors stories for prompt field is preventing the Storybook show/hide code functionality.
3. Can drag and drop be responded to for the Playground story?

5t3ph (approved):

PR #6645, opened 2026-08-24
feat(docs): add 2nd-gen shared resources quick reference

5t3ph:
## Description

Adds a new contributor doc, [2nd-gen shared resources quick reference](CONTRIBUTOR-DOCS/01_contributor-guides/16_2nd-gen-shared-resources.md), indexing every shared controller, mixin, utility, and directive in `@adobe/spectrum-wc-core` (plus a couple of sibling-package helpers), what each one does, and which components/patterns currently consume it. Also documents the `swc/patterns/conversational-ai` layer's dependency surface, since it consumes core resources and embeds components directly without its own `*Base` class.

While researching the doc, found and corrected staleness in three existing style guide docs that no longer matched the current codebase:

- `13_mixin-composition.md` documented `ObserveSlotText`/`ObserveSlotPresence` mixins at file paths that no longer exist (they were refactored into `SlotTextController`/`SlotPresenceController`), and its "Available mixins" list was missing `DisabledMixin`, `PendingMixin`, and `LinearProgressMixin` entirely.
- `14_controller-composition.md`'s "Available controllers" table listed only 2 of 11 real controllers; its "Planned controllers" table listed `PlacementController`, `HoverController`, and `PendingStateController` as not yet built, though all three are shipped and in use.
- `15_directive-composition.md` claimed "2nd-gen does not currently use custom directives," but `renderPendingSpinner` is one, used by `button`/`action-button`.

Also updates four AI agent skill files so this doc stays current going forward, particularly during component migrations:

- `migration-prep` (Phase 1): added a check for existing shared resources to the "escalate when" list and final review checklist, at the point where a migration plan's architecture is decided (this is where a near-miss happened during a recent draft plan, almost missing that the `for` attribute + `resolveTrigger()` pattern already existed).
- `migration-api` (Phase 3): one-line reminder to check the doc before implementing new logic.
- `consistency-pass`: added a new **Part 3: Shared resources** check that only fires when changed files compose or define a `core` resource, so it costs nothing on phases that don't touch this.
- `migration-review` (Phase 8): added a blocking (not advisory) final-gate check, since this is the last checkpoint before merge.

## Motivation and context

Shared 2nd-gen patterns (like `resolveTrigger()`, born out of `tooltip`/`popover`) are easy to miss if you weren't on the PR that introduced them — there was a near-miss where a draft migration plan almost reinvented trigger resolution instead of reusing the existing pattern, because the engineer wasn't aware it existed. This doc closes that visibility gap, and the skill updates make checking it (and keeping it current) part of the normal migration workflow rather than something that has to be remembered.

### Manual review test cases

Provide feedback if docs are:
- accurate
- useful
- correct cross-links
- appropriate inclusion in AI skills


CONTRIBUTOR-DOCS/01_contributor-guides/16_2nd-gen-shared-resources.md
rise-erpelding, line 1: This is really helpful! 👏

CONTRIBUTOR-DOCS/01_contributor-guides/16_2nd-gen-shared-resources.md
```diff
@@ -0,0 +1,149 @@
...
+| `SizedMixin` | Reactive `size` property with per-component valid-size enforcement | `button`, `button-group`, `badge`, `card`, `accordion`, `status-light`, `divider`, `dropzone`, `icon`, `infield-button`, `meter`, `progress-bar`, `progress-circle` |
+| `LinearProgressMixin` | Shared value/label/percent-format logic for linear-progress components | `progress-bar`, `meter` |
+| `PendingMixin` | `pending`/`pending-label` properties, click suppression, built on `PendingController` | `button`, `action-button` |
+| `DisabledMixin` | Reactive `disabled` with `aria-disabled` (not native `disabled`), tabindex/blur handling. See [Focus management](14_focus-management.md#disabledmixin) | none (`button` still hand-rolls `disabled`) |
+
+## Utilities
+
+| Utility | What it does | Used by |
+|---|---|---|
+| `resolveTrigger()` | Resolves `for` ID / `triggerElement` to a trigger, and its AT-facing inner `<button>` across shadow boundaries | `tooltip`, `popover` |
```
rise-erpelding, line 66: It looks like tooltip may have its own `resolveTrigger()` method instead of using a shared util:

```ts
  private resolveTrigger(): HTMLElement | null {
    if (this.triggerElement) {
      return this.triggerElement;
    }
    if (this.for) {
      const root = this.getRootNode() as Document | ShadowRoot;
      const trigger = root.getElementById(this.for);
      warnIf(
        this,
        !trigger,
        `<${this.localName}> for="${this.for}" did not resolve to an element in the current tree root. Check that the referenced id exists in the same document tree root.`,
        'https://spectrum-web-components.adobe.com/?path=/docs/components-tooltip--docs',
        { level: 'high' }
      );
      return trigger;
    }
    return null;
  }
```

Might be worth an update in the documentation now, and possibly a refactor to use the utility (maybe there's even already a ticket?) eventually?

cdransf (approved):
Looks great! ✨

CONTRIBUTOR-DOCS/01_contributor-guides/16_2nd-gen-shared-resources.md
```diff
@@ -0,0 +1,149 @@
...
+>
+> ⚠️ **Important:** Composition happens at two layers. Most controllers/mixins are applied on the shared abstract `*Base` class in `core/components/*` and inherited by every concrete element. A few (`PendingMixin`) are applied one layer up, directly on the concrete element in `swc/components/*` (e.g. `Button extends PendingMixin(ButtonBase)`). Check both layers before concluding something isn't in use.
+
+## Controllers
+
+| Controller | What it does | Used by |
+|---|---|---|
+| `PlacementController` | Floating UI-based positioning (flip, shift, arrow, size) for anchored surfaces | `tooltip`, `popover` |
+| `HoverController` | Hover and keyboard-focus wiring for native-Popover-API triggers | `tooltip` |
+| `FocusgroupNavigationController` | Roving-tabindex arrow key navigation for composite widgets ([Open UI `focusgroup`](https://open-ui.org/components/focusgroup.explainer/)-aligned). See [Focus management](14_focus-management.md) | `tabs`, `message-feedback`, `conversation-thread` |
```
rise-erpelding: It looks like `prompt-field` may use `FocusgroupNavigationController` as well as `getActiveElement()` but the component isn't mentioned under either of those utilities:

```ts
import {
  focusgroupNavigationActiveChange,
  type FocusgroupNavigationActiveChangeDetail,
  FocusgroupNavigationController,
} from '@adobe/spectrum-wc-core/controllers/index.js';
import { SpectrumElement } from '@adobe/spectrum-wc-core/element/index.js';
import {
  deepContains,
  focusableSelector,
  getActiveElement,
} from '@adobe/spectrum-wc-core/utils/index.js';
```

CONTRIBUTOR-DOCS/01_contributor-guides/16_2nd-gen-shared-resources.md
```diff
@@ -0,0 +1,149 @@
...
+| `upload-artifact` | `getLabelFromSlot()` | `action-button`, `icon` |
+| `message-feedback` | `FocusgroupNavigationController` | `icon` |
+| `response-status` | `uniqueId()` | `icon` |
+| `message-sources` | `uniqueId()` | `icon` |
+| `suggestion` (`SuggestionGroup`) | `uniqueId()` | — |
+| `suggestion-item` | — | `icon` |
+| `conversation-thread` | `FocusgroupNavigationController`, `getActiveElement()` | — |
+| `pixel-loader` | `validateEnum()` | — |
+
+`prompt-field` has the widest dependency surface. Check its stories when changing `action-button`, `icon`, `FocusgroupNavigationController`, `deepContains()`, or the dev-validation utilities.
```
rise-erpelding: Does maybe `prompt-field`'s entry in the table above here also need some updates? (I see that it has been acknowledged as having a wide dependency surface here)

CONTRIBUTOR-DOCS/01_contributor-guides/16_2nd-gen-shared-resources.md
```diff
@@ -0,0 +1,149 @@
...
+| `getActiveElement()` | Deepest focused element, across shadow boundaries | `popover`, `conversation-thread` |
+| `deepContains()` | Shadow-piercing containment check | `popover`, `prompt-field` |
+| `registerDismissible()` / `unregisterDismissible()` / `isTopDismissible()` (`dismissibleStack`) | LIFO stack coordinating Escape handling across top-layer mechanisms | `popover` |
+| `warnIf()`, `validateEnum()` | Batched, deduped dev-mode warning framework (see below and [Debug and validation](../02_style-guide/02_typescript/17_debug-validation.md)) | 10+ components, `prompt-field`, `pixel-loader` |
+| `focusableSelector` | Spec-based focusable-element selector. See [Focus management](14_focus-management.md#focusableselector-and-tabbableselector) | `prompt-field` |
+| `tabbableSelector` | Spec-based tabbable-element selector | none |
+| `isFocusVisibleInTree()` | `:focus-visible` check across shadow roots | none |
+| `getLabelFromSlot()` | Derives a text label from slotted content | `upload-artifact` |
+| `validateRequiredSlot()`, `validateAllowedChildren()` | Dev-mode structural validation | `validateAllowedChildren`: 1 component |
```
rise-erpelding: I don't know enough about these 2 utilities to say whether or not they should be split, but it's something we could consider.

The "1 component" here though I think is Illustrated message; it might be more helpful to name it here.

CONTRIBUTOR-DOCS/01_contributor-guides/16_2nd-gen-shared-resources.md
```diff
@@ -0,0 +1,149 @@
...
+| `PageScrollLockController` | Reference-counted page scroll lock for stacked blocking surfaces | `popover` |
+| `PendingController` | Delayed busy-state activation, size freeze, pending accessible name | `button`, `action-button` (via `PendingMixin`) |
+| `LanguageResolutionController` | Resolves locale (from `<html lang>`, `navigator.language`, or a language-provider event) for formatting. See [Controller composition](../02_style-guide/02_typescript/14_controller-composition.md#languageresolutioncontroller) | `progress-circle`; also `progress-bar`, `meter` (via `LinearProgressMixin`) |
+| `ColorController` | `colorjs.io` conversion wrapper | none |
+
+## Mixins
+
+| Mixin | What it does | Used by |
+|---|---|---|
+| `SizedMixin` | Reactive `size` property with per-component valid-size enforcement | `button`, `button-group`, `badge`, `card`, `accordion`, `status-light`, `divider`, `dropzone`, `icon`, `infield-button`, `meter`, `progress-bar`, `progress-circle` |
```
rise-erpelding: Do we want to add action-group too now that that's landed?

https://github.com/adobe/spectrum-web-components/blob/main/2nd-gen/packages/core/components/action-group/ActionGroup.base.ts#L17

CONTRIBUTOR-DOCS/01_contributor-guides/16_2nd-gen-shared-resources.md
```diff
@@ -0,0 +1,149 @@
...
+
+## Controllers
+
+| Controller | What it does | Used by |
+|---|---|---|
+| `PlacementController` | Floating UI-based positioning (flip, shift, arrow, size) for anchored surfaces | `tooltip`, `popover` |
+| `HoverController` | Hover and keyboard-focus wiring for native-Popover-API triggers | `tooltip` |
+| `FocusgroupNavigationController` | Roving-tabindex arrow key navigation for composite widgets ([Open UI `focusgroup`](https://open-ui.org/components/focusgroup.explainer/)-aligned). See [Focus management](14_focus-management.md) | `tabs`, `message-feedback`, `conversation-thread` |
+| `SlotAttributePropagationController` | Propagates a host attribute (e.g. `size`) to slotted children | `card`, `dropzone`, `illustrated-message`, `button-group` |
+| `SlotPresenceController` | Observes whether slotted content matching a selector is present | `button`, `action-button`, `badge`, `accordion` |
```
rise-erpelding: I think also `progress-bar` and `meter` (via `LinearProgressMixin`)

CONTRIBUTOR-DOCS/01_contributor-guides/16_2nd-gen-shared-resources.md
```diff
@@ -0,0 +1,149 @@
...
+| `SlotTextController` | Observes whether a slot has meaningful text/element content | `card`, `button`, `action-button`, `illustrated-message`, `badge` |
```
rise-erpelding, line 46: Would use of `SlotTextController` in `button` and `action-button` be via `ButtonBase`? Do we want to mention that in some of these Used by lists? (Similar to how we say "via `LinearProgressMixin`")

.ai/skills/consistency-pass/SKILL.md
```diff
@@ -78,9 +79,22 @@ Common cascading patterns to check:
 
 This check also applies during plan authoring. If you are writing or editing the plan in Phase 1 and make a change, run the same cascading scan before completing the phase.
 
+## Part 3: Shared resources
+
+`CONTRIBUTOR-DOCS/01_contributor-guides/16_2nd-gen-shared-resources.md` is the index of shared controllers, mixins, utilities, and directives in `@adobe/spectrum-wc-core` (and a couple of sibling-package helpers) and who uses each one. It goes stale silently: nothing fails to build or lint if it isn't updated.
+
+Skip this part entirely, with no report line, if the phase touched no code that composes or defines a `core` resource (e.g. a pure styling or test-only phase) — do not pad the report with a no-op check.
```
rise-erpelding: This looks to be in conflict with the lines below that say:

```md
Report all three parts together before ending your response:
```

and also

```md
Shared resources: [not applicable — no core resource touched this phase | updated for [resource] | ⚠️ [resource] newly consumed/introduced, doc not yet updated]
```

CONTRIBUTOR-DOCS/01_contributor-guides/16_2nd-gen-shared-resources.md
```diff
@@ -0,0 +1,149 @@
...
+| `tabbableSelector` | Spec-based tabbable-element selector | none |
+| `isFocusVisibleInTree()` | `:focus-visible` check across shadow roots | none |
+| `getLabelFromSlot()` | Derives a text label from slotted content | `upload-artifact` |
+| `validateRequiredSlot()`, `validateAllowedChildren()` | Dev-mode structural validation | `validateAllowedChildren`: 1 component |
+| `isDebug()` | Whether dev mode is active (the gate `warnIf`/`validateEnum` build on) | `progress-circle`, `card`, `dropzone`; also `progress-bar`, `meter` (via `LinearProgressMixin`) |
+| `capitalize()` | Capitalizes first character of a string | `progress-circle`, `divider` |
+| `physicalSide()` | Drops alignment suffix from a placement (`bottom-start` to `bottom`) | `popover` |
+| `runAfterTransition()` | Runs a callback once an element's CSS transition settles (or synchronously if none will run); built on `hasActiveTransition()`/`maxTransitionDurationMs()` in the same file | `popover` |
+| `uniqueId()` (`swc/utils/id.ts`, not `core`) | Collision-resistant component ID generation | `prompt-field`, `suggestion`, `response-status`, `message-sources` |
```
rise-erpelding, line 80: The skill as we have it now isn't set up to find and document something like `uniqueId()` if we have to add something like that in the future, because it's not in core:

```md
Otherwise, check the files changed in this phase for:

- A new `extends XMixin(...)`, a new `private x = new XController(this)` field, or a new import of a utility/directive from `@adobe/spectrum-wc-core` — an existing resource gained a consumer. Add the component or pattern to its "Used by" entry.
- A new controller, mixin, utility, or directive defined under `2nd-gen/packages/core` (something promoted out of this component's own logic because it turned out to be general) — a new resource exists. Add a row for it.
```

CONTRIBUTOR-DOCS/01_contributor-guides/16_2nd-gen-shared-resources.md
5t3ph: Awww, good call, forgot we didn't already go back for this update - added SWC-2515

CONTRIBUTOR-DOCS/01_contributor-guides/16_2nd-gen-shared-resources.md
5t3ph: A note above this table covers the case of "applied at the base class" so that's meant to be implied, vs the "via" is when it's less directly used like inside a mixin.

CONTRIBUTOR-DOCS/01_contributor-guides/16_2nd-gen-shared-resources.md
5t3ph: Moved the deps inline to the table

CONTRIBUTOR-DOCS/01_contributor-guides/16_2nd-gen-shared-resources.md
5t3ph: listed illustrated-message

CONTRIBUTOR-DOCS/01_contributor-guides/16_2nd-gen-shared-resources.md
5t3ph: Updated from main to address this and couple other items from latest

.ai/skills/consistency-pass/SKILL.md
5t3ph: Good call - broadened to keep a message even if not applicable

CONTRIBUTOR-DOCS/01_contributor-guides/16_2nd-gen-shared-resources.md
5t3ph: True, broadened to look for sibling utilities

rise-erpelding (approved):
Looks great, I'm so excited for future component migrations to be able to use this!

PR #6648, opened 2026-08-25
ci: dynamic config, shard 2nd-gen tests, build once

caseyisonit:
## Description

Restructures the CircleCI pipeline to cut wall-clock time and credit usage, and fixes the `test-2ndgen-browser-webkit` OOM. `.circleci/config.yml` becomes a small **setup config**; all real jobs move to `.circleci/continue_config.yml`.

### 1. Dynamic config — stop booting containers just to halt
Previously every gated job (including each of the `parallelism: 5` containers) booted an **xlarge** container solely to run a git diff and `step halt`. On a scoped PR that's ~20 xlarge boots doing nothing.

Now a single `small` `decide` job diffs the PR against `origin/main` once and *continues* the pipeline with `run_1st_gen` / `run_2nd_gen` pipeline parameters. Unaffected-gen workflows are **never created** → zero credits instead of boot-and-halt. The relevance logic is copied verbatim from the old `halt-if-not-affected` command (a positive-only path-filtering orb can't express its "outside the other gen AND not docs-only" negation).

### 2. Shard 2nd-gen browser tests (fixes the webkit OOM)
`test-2ndgen-browser` ran the entire suite in one container per browser — webkit exhausted memory and was killed mid-run. It's now `parallelism: 5` with vitest `--shard`, so each container loads ~1/5 of the suite. Coverage moved to its own single-container `test-2ndgen-coverage` job (v8 coverage needs one process; still Chromium-only, base=main-only).

### 3. Build once, attach everywhere
- `build-2nd-gen` builds the packages + CEM once (was rebuilt per browser — and would be 15× once sharded); browser/coverage jobs `attach_workspace`.
- `build-2nd-gen-storybook` builds the `ci-a11y` Storybook once (a11y + axe each rebuilt the identical Storybook before).

### 4. Right-size resource classes
New `node-large` executor for the now-sharded test jobs (half the RAM/credits of xlarge, ample per-shard); `decide` on `small`; xlarge kept for builds, coverage, and VRT.

## How has this been tested?
- Both configs pass `circleci config validate`.
- Verified the a11y Playwright server serves the persisted `2nd-gen/packages/swc/storybook-static` path, and vitest 4.1.7 supports `--shard`.
- Full behavior (persist set, per-shard memory) will be confirmed by this PR's own CI run.

## ⚠️ Required before merge — branch protection
Unaffected-gen job checks are now **absent, not green**. Update branch protection to require **only** the always-present `ci/circleci: decide` check, and remove every per-gen job name from the required list. CircleCI still fails the pipeline if any continued job fails.

## Types of changes
- [x] CI / infrastructure
- [x] Breaking change to CI status-check contract (see branch-protection note)


.circleci/config.yml
```diff
@@ -1,758 +1,105 @@
...
-      - test-chromium:
-          requires: [build]
-      - test-chromium-memory:
-          requires: [build]
-      - test-firefox:
-          requires: [build]
-      - test-webkit:
-          requires: [build]
-      - test-chromium-coverage:
-          requires: [build]
```
cdransf, line 720: Do we need to preserve these 1st gen test steps (or run/scope them elsewhere?) ✨

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/test/prompt-field.test.ts
```diff
@@ -442,50 +442,72 @@ export const ArtifactScrollRTLTest: Story = {
...
       settled = true;
+      clearTimeout(timeoutId);
+      cancelAnimationFrame(rafId);
+      scrollEl.removeEventListener('scrollend', finish);
       resolve();
     };
 
+    // Safety net: resolve no matter what, so a scroll that never moves (or
+    // a missing `scrollend`) can't hang the test.
+    const timeoutId = setTimeout(finish, maxWaitMs);
```
cdransf: Will this cause intermittent failures if we call `finish` and the next `getBoundingClientRect` hits an unexpected scroll position? (E.g. if the machine is moving slowly and `finish` is called early). ✨

cdransf (changes requested):

.circleci/continue_config.yml
```diff
@@ -0,0 +1,773 @@
...
+    executor: node
+    steps:
+      - setup-node-dependencies
+      - attach_workspace:
+          at: .
+      - resolve-pr-base
+      - run:
+          name: Run 2nd-gen Chromium coverage (base=main only)
+          command: |
+            if [ "$PR_BASE_BRANCH" = "main" ]; then
```
rubencarvalho: If I see it correctly, gen2 coverage was not really uploading the artifacts as expected running on PRs 🙃 In this PR it also does not but because the job log prints `Skipping coverage (base='<unknown>')` (and goes green).

Asked Claude to help me debug this and basically `resolve-pr-base` depends on `CIRCLE_PULL_REQUEST`, which CircleCI leaves empty on most PR builds, so `PR_BASE_BRANCH` stays empty and this guard falls through to the skip branch.
I think we can fix this by resolving the base in the `decide` setup job (which already fetches `origin/main` and diffs) and passing it as a pipeline parameter - I'll leave a comment below on `config.yml`.

.circleci/continue_config.yml
```diff
@@ -0,0 +1,773 @@
...
+              yarn workspace @spectrum-web-components/2nd-gen run test \
+              --shard=$((CIRCLE_NODE_INDEX + 1))/$CIRCLE_NODE_TOTAL
+      - store_test_results:
+          path: /root/project/2nd-gen/packages/swc/test-results
+
+  # 2nd-gen coverage: Chromium only, and only when the PR targets main (matches
+  # the previous inline guard). Runs unsharded in a single container because v8
+  # coverage needs the whole run in one process. xlarge for headroom.
+  test-2ndgen-coverage:
+    executor: node
```
rubencarvalho, line 405: This is interesting, right? All our test runs are now sharded, but this coverage one is not, and I understand that it prob needs to be unshareded for V8 to collect the coverage, but this means we may still run into the resource exhaustion problem later. I don't think we need to worry now but I do know there is a way of sharding coverage too (`--coverage` per shard → merge reports)

.circleci/continue_config.yml
```diff
@@ -0,0 +1,773 @@
...
+          path: /root/project/results/
+
+  # ---------------------------------------------------------------------------
+  # 2nd-gen integration tests (Vitest browser mode)
+  # ---------------------------------------------------------------------------
+  # Sharded across 5 containers per browser via vitest `--shard`. This is what
+  # fixes the webkit OOM: each container loads ~1/5 of the suite, so peak
+  # memory stays well under the (now `large`) container ceiling. No coverage
+  # here — coverage is its own single-container job below.
+  test-2ndgen-browser:
```
rubencarvalho: I'm just going to leave a general comment here, but there would be multiple instances and maybe you can try to do a search and replace, but can we prefer `gen1` / `gen2` over `1st-gen` / `2nd-gen` for our "human-facing" (🙅 🤖) names? 😄

.circleci/continue_config.yml
```diff
@@ -0,0 +1,773 @@
...
+  # Build the 2nd-gen (ci-a11y) Storybook ONCE and persist the static output.
+  # The a11y and axe jobs both consumed this identical build before, each
+  # rebuilding it on its own xlarge container. Uses `downstream` because the
+  # Storybook build (generate:contributor-docs) still pulls in 1st-gen refs.
+  build-2nd-gen-storybook:
+    executor: node
+    steps:
+      - downstream
+      - run:
+          name: Build 2nd-gen Storybook (ci-a11y mode)
```
rubencarvalho: Can we disambiguate all the storybook builds? 😛 

e.g., `Build gen2 Storybook (a11y/axe — ci-a11y mode)`, for the beta-docs: `Build gen1 Storybook (beta docs)`

and maybe rename the job `build-2nd-gen-storybook` → `build-gen2-storybook` to match.

.circleci/continue_config.yml
```diff
@@ -0,0 +1,773 @@
...
+  second-gen:
+    when: << pipeline.parameters.run_2nd_gen >>
+    jobs:
+      # 2nd-gen package build — consumed by the browser + coverage jobs.
+      - build-2nd-gen
+      # 1st-gen build — needed only by the Storybook build (contributor-docs
+      # generation pulls in 1st-gen refs). Named distinctly so it doesn't
+      # collide with the `first-gen` workflow's `build` status check.
+      - build:
+          name: build-for-storybook
```
rubencarvalho: similarly, the `second-gen` workflow has three build jobs and the names don't make their roles obvious: `build-2nd-gen` (gen2 packages + CEM), this `build` aliased to `build-for-storybook` (actually a gen1 build, needed because contributor-docs generation pulls in gen1 refs??), and `build-2nd-gen-storybook` (the static Storybook).

`build-for-storybook` reads as "builds the storybook" but it builds gen1 packages. Consider `build-gen1-for-storybook-docs` (or similar) so the intent is more legible from the workflow graph? (you know the circleci nice diagrams)

.circleci/continue_config.yml
```diff
@@ -0,0 +1,773 @@
...
+      - downstream
+      - run:
+          name: Run chromium tests
+          command: |
+            TEST=$(circleci tests glob "1st-gen/{packages,tools}/*/test/*.test.js" | grep -v ".*-memory\.test\.js$" | circleci tests split --split-by=timings | sed 's/1st-gen\///')
+            yarn workspace @spectrum-web-components/1st-gen test:start --files $TEST --config ./web-test-runner.config.ci-chromium.js --group unit-ci
+      - store_test_results:
+          path: /root/project/1st-gen/results/
+      - store_artifacts:
+          path: 1st-gen/coverage
```
rubencarvalho: so i think this is what i meant (prob carried from `main`), but `test-chromium` stores `1st-gen/coverage` as an artifact but runs without `--coverage`, so this likely uploads nothing? while `test-chromium-coverage` (the job that _does_ run `--coverage`) has no `store_artifacts`, so the gen1 coverage report it produces is discarded.

Looks like the `store_artifacts` is on the wrong job. leaving it as a nit though, because... 1st-gen! 🤷

.circleci/continue_config.yml
```diff
@@ -0,0 +1,773 @@
...
+    steps:
+      - downstream
+      - run:
+          name: Run memory tests
+          command: |
+            TEST=$(circleci tests glob "1st-gen/{packages,tools}/*/test/*-memory.test.js" | grep -v "packages/color-.*/test/.*-memory\.test\.js" | grep -v "tools/grid/test/.*-memory\.test\.js" | circleci tests split --split-by=timings | sed 's/1st-gen\///')
+            echo $TEST
+            yarn workspace @spectrum-web-components/1st-gen test:start --files $TEST --config web-test-runner.config.ci-chromium-memory.js --group unit-ci
+      - store_test_results:
+          path: /root/project/results/
```
rubencarvalho: So this is prob the other root cause of what I was seeing (no tests reported in the CircleCI tab)... `test-chromium` stores results from `/root/project/1st-gen/results/`, but this job (memory) plus `test-firefox` and `test-webkit` store from `/root/project/results/`. The gen1 runner writes to `1st-gen/results/`, so the `/root/project/results/` jobs may be capturing nothing.

.circleci/config.yml
```diff
@@ -1,758 +1,105 @@
...
-              if [ -n "$PR_NUMBER" ] && [ -n "$GITHUB_TOKEN" ]; then
-                BASE=$(curl -s -H "Authorization: Bearer $GITHUB_TOKEN" \
-                  -H "Accept: application/vnd.github+json" \
-                  "https://api.github.com/repos/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PROJECT_REPONAME/pulls/$PR_NUMBER" \
-                  | python3 -c "import json,sys; print(json.load(sys.stdin).get('base',{}).get('ref',''))" 2>/dev/null || echo "")
+              # Make sure origin/main is fetched so we can diff against it.
+              # No --depth: a stale PR can have a deep merge-base, and the
+              # fetch is fast since CircleCI's checkout already has shared
+              # history.
+              git fetch --quiet origin main 2>/dev/null || true
```
rubencarvalho, line 34: I think this is a good place to resolve the PR base branch and fix the `base='<unknown>'` coverage skip in `continue_config.yml`. `decide` already fetches `origin/main` and has full git + checkout context, so it can determine the base reliably (it doesn't depend on `CIRCLE_PULL_REQUEST`! 😄).

We could just have a `base_branch` (or mb even simpler a `is_base_main` boolean) pipeline parameter here, pass it in the continuation `params`, and gate `test-2ndgen-coverage` on it at the workflow level.

.circleci/continue_config.yml
caseyisonit: Yeah this is an issue throughout the code base, I’ll fix it here but the bug sweep will come with a code quality epic that miwha and I are spear heading. We will share out about that this week.

.circleci/config.yml
caseyisonit: They're preserved — all the gen1 jobs moved into the `gen1` workflow in `continue_config.yml`, gated by the `run_gen1` pipeline parameter so they run exactly when gen1 is affected (and always on `main`). Nothing dropped, just relocated + gated.

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/test/prompt-field.test.ts
caseyisonit: Good catch, but moot now: merging `main` brought in an equivalent `waitForScrollEnd` fix (wait for the scroll to actually move before counting stable frames, plus a frame cap), and I took `main`'s version in the merge. The early-`finish` window you describe is closed.

.circleci/continue_config.yml
caseyisonit: Fixed in 64760df746. The base branch is now resolved in the `decide` setup job (git + the GitHub API by branch, not `CIRCLE_PULL_REQUEST`) and passed as an `is_base_main` pipeline parameter; `test-gen2-coverage` gates on that. No more `base='<unknown>'` skip.

.circleci/continue_config.yml
caseyisonit: Acknowledged — keeping it single-container for now since v8 needs the whole run in one process. Sharded coverage (`--coverage` per shard → merge reports) is a solid follow-up; noting it for later rather than expanding this PR's scope.

.circleci/continue_config.yml
caseyisonit: Done in 64760df746 — renamed human-facing identifiers to gen1/gen2 (pipeline params `run_gen1`/`run_gen2`, workflows, jobs, and run-step names). Left directory paths and the `@spectrum-web-components/1st-gen` workspace names as-is since those are package identifiers rather than display names.

.circleci/continue_config.yml
caseyisonit: Done — steps are now `Build gen2 Storybook (a11y/axe — ci-a11y mode)` and `Build gen1 Storybook (beta docs)`, and the job is `build-gen2-storybook`.

.circleci/continue_config.yml
caseyisonit: Renamed for legibility: `build-gen2` (gen2 packages + CEM), `build-gen1-for-storybook-docs` (the gen1 build the Storybook needs because contributor-docs generation pulls in gen1 refs), and `build-gen2-storybook` (the static Storybook). Should read clearly in the workflow graph now.

.circleci/continue_config.yml
caseyisonit: Fixed in 64760df746 — moved `store_artifacts: 1st-gen/coverage` onto `test-chromium-coverage` (the job that runs `--coverage`) and off `test-chromium`. Not just a gen1 nit; may as well be correct. 🙂

.circleci/continue_config.yml
caseyisonit: Fixed in 64760df746 — the memory/firefox/webkit jobs now `store_test_results` from `/root/project/1st-gen/results/` (where the gen1 runner actually writes), instead of the empty `/root/project/results/`. That should restore the missing results in the CircleCI tab.

.circleci/config.yml
caseyisonit: Done in 64760df746 — `decide` resolves the base and passes an `is_base_main` boolean param, and `test-gen2-coverage` gates on it (went with the boolean, as you suggested). Kept the gate inline in the job rather than a separate workflow, so it can attach the shared `build-gen2` workspace instead of duplicating that build on every PR to main.

cdransf (approved):

miwha-adobe:
Really like where this is going. Just want to flag one thing before it lands.

Right now, if a test goes red, GitHub won't let you merge. After this change, the tests still run and still show red/green on the PR, but nothing stops you from merging a red one anymore. You'd just have to notice it.

Screenshot 2026-08-27 at 2 10 37 PM

Might be a fix though: if the CircleCI GitHub App can post one always-present "whole pipeline" check (it's already posting the gen1/gen2/setup ones), we just require that and we should be good in theory.

Also your note above is correct:

> Unaffected-gen job checks are now absent, not green. Update branch protection to require only the always-present ci/circleci: decide check, and remove every per-gen job name from the required list. CircleCI still fails the pipeline if any continued job fails.

The catch is that "CircleCI fails the pipeline" only makes the red X show up on the PR. It doesn't block the merge button, since GitHub only gates on the required checks (just decide, which is already green before the tests run). So the pipeline can be red and we can still merge.

.circleci/continue_config.yml
```diff
@@ -0,0 +1,749 @@
...
+          path: /root/project/2nd-gen/packages/swc/test-results
+
+  # gen2 coverage: Chromium only, and only when the base is main. Guarded on
+  # the `is_base_main` pipeline parameter (resolved reliably in `decide`, not
+  # the old CIRCLE_PULL_REQUEST-dependent path that left this permanently
+  # skipped). The guard is inline rather than a separate workflow so it can
+  # attach the shared `build-gen2` workspace instead of duplicating that build
+  # on every PR to main. Runs unsharded in a single container because v8
+  # coverage needs the whole run in one process. xlarge for headroom.
+  test-gen2-coverage:
```
miwha-adobe, line 404: **Nit:** Skipping coverage on non-main makes total sense (no baseline to compare against). But the skip happens after the xlarge boots, checks out, and runs yarn --immutable, so we do a few minutes of setup just to echo "skipping."  

**Suggestion(non blocking):** move the is_base_main check to the top and bail early when it's false

.circleci/continue_config.yml
```diff
@@ -0,0 +1,749 @@
...
+      - run:
+          name: Generate Custom Elements Manifest
+          command: yarn workspace @adobe/spectrum-wc analyze
+      - save-node-dependencies-cache
+      - persist_to_workspace:
+          root: .
+          paths:
+            - 2nd-gen/packages/core/dist
+            - 2nd-gen/packages/swc/dist
+            - 2nd-gen/packages/core/custom-elements.json
```
miwha-adobe: Claude is saying that these two lines could be deleted:

```
- 2nd-gen/packages/core/custom-elements.json
- 2nd-gen/packages/swc/custom-elements.json
```

Both of these point at files that don't exist. analyze writes the swc manifest into dist/ (per cem.config.js outdir: 'dist'), so it's already covered by the swc/dist line above. And core has no analyze step at all, so core/custom-elements.json is never generated.

rubencarvalho:
I'm still thinking about the fact we have no required checks now... Before, we were still returning success from skipped tests to circumvent this, but Miwha's comment seems to hint at an alternative?

.circleci/continue_config.yml
caseyisonit: ok so i dug deeper in to this and breaking it out would cost more, not less — i have modified this job though to halt before anything attaches so its the lowest cost. Presently this is an edge case optimization and something we can keep an eye on if we start doing more stacked PRs or feature branching.

caseyisonit:
@miwha-adobe @rubencarvalho 
Heads up I added new required check + branch protection changes

This PR switches us to a dynamic CircleCI config — `decide` diffs the PR up front and only spins up the workflows for the gens that actually changed. The upside is we stop burning credits on jobs that just no-op. The catch is that when a gen isn't touched, its jobs don't run at all, so those checks show up as absent rather than green. GitHub won't let you merge while a required check is missing, so I can't keep the individual gen checks in the required list anymore.

So thanks for the suggestion, I added a single gate called `required-checks` that's always there.

What the gate does

There's a new gate workflow that runs on every pipeline (no `when:` on it), so its `require-checks` check always shows up. That job looks at the `run_gen1` / `run_gen2` params, then hits the CircleCI API to confirm that every gen workflow that should have run actually ran and passed. If one that was supposed to run is missing, I treat that as a failure on purpose — I didn't want a silently dropped workflow to sneak through green.

Quick rundown of how it behaves:
- Normal gen1/gen2 PR → waits on those workflows, passes only if they're green
- Docs-only PR → passes right away, nothing to run
- A gen workflow fails → require-checks fails, merge blocked
- A workflow that should exist never gets created → fails after a short grace period

What needs to change in branch protection

Once this lands, I will update the required checks:
- Add `require-checks` and `decide`
- Remove the per-gen job checks (test-chromium, test-gen2-browser-*, build, etc.) — those go missing on partial PRs and will deadlock merges if they're still required

caseyisonit:
After meeting with @miwha-adobe and @rubencarvalho we have come up with a solid solution that doesnt require polling API

Branch protection needs updating when this lands. I swapped the API-polling gate for two commit statuses that decide and each workflow post directly — no token, no polling.

Update the required checks to:

✅ `gen1-tests`
✅ `gen2-tests`
✅ `decide`
❌ remove the individual gen job names (they're absent on PRs that don't touch that gen)

`gen1-tests` / `gen2-tests` start pending, flip to success once that gen's jobs pass, and stay pending (blocking) if anything fails. Gens that aren't affected get marked success automatically, so partial PRs aren't held up.

miwha-adobe (approved):
lgtm! Thanks for addressing all of that. The checks are hanging for some reason but when I look at the build in CircleCI it seems like it passed? Not sure if this is a result of the config changes.

PR #6653, opened 2026-08-26
feat(tokens): update spectrum-tokens to 15.2.0 and adopt opacity tokens in prompt-field

rubencarvalho:
## Description

Updates `@adobe/spectrum-tokens` from **14.13.1 → 15.2.0** ([release notes](https://github.com/adobe/spectrum-design-data/releases/tag/@adobe/spectrum-tokens@15.2.0)) and adopts the new `opacity-*` tokens in the prompt-field where they map exactly.

### Why this is more than a version bump

15.2.0 restructured the token package: the aggregated `color-component.json` was removed and every component's tokens were split into per-component files (`card.json`, `avatar.json`, … ~83 files). The `swc-tokens` loader hard-coded an 8-file list, so the token build broke (`MODULE_NOT_FOUND` for `color-component`). The typography generator was similarly affected (per-variant tokens moved out of `typography.json`).

### Changes

- **`swc-tokens` loader** (`src/tokens.js`): auto-discover per-component token files instead of a fixed list. Component files are emitted with aliases kept as `var()` references (matching how component color/typography tokens were emitted before the split).
- **Typography generator** (`src/typography.js`): load the per-variant files (`heading.json`, `body.json`, `title.json`, `detail.json`, `code.json`) where those tokens now live. Output is byte-for-byte identical to before.
- **`diff-versions` helper**: updated for the new global/component file split.
- **`diff-versions.js` package.json resolution fix**: fixed `ERR_PACKAGE_PATH_NOT_EXPORTED` when the script resolves the installed package's version. Some `@adobe/spectrum-tokens` releases (including the currently installed 14.13.1) define an `exports` map that doesn't allowlist `./package.json`, so `require.resolve('@adobe/spectrum-tokens/package.json')` fails under Node's ESM resolution. The script now resolves the package root via the already-exported `.` entry and reads `package.json` with a plain filesystem join instead of going through module resolution.
- **`diff-versions.js` `custom/deleted.json` merge fix**: fixed a bug where the script would silently drop previously curated entries that weren't part of the *current* from→to diff (only entries matching the current run were retained; anything curated during an earlier version bump was overwritten out of the file). The script now always starts from the existing file's contents and only appends newly-found deletions, so curation from past upgrades is never lost.
- **Regenerated `tokens.css`**: adds the new `opacity-*` scale (18 custom properties) and refreshes component token values per 15.2.0. Large diff is mostly token reordering (discovery order); the real delta is **+31 tokens / −5 unused / 93 value changes**, and every `var()` reference resolves.
- **prompt-field**: two surface transparencies that map exactly to the scale now use tokens — `--_swc-prompt-field-inset-shadow-color` → `opacity-100` (0.04) and the outer recede stop → `opacity-50` (0.02). Resolved output is identical (`.04` / `.02`) — **no visual change**.
- **`swc-tokens` README**: restructured so the version-bump workflow is impossible to miss — added a top-level "Updating tokens" section immediately after Terminology that splits the two paths (custom tokens only vs. package version bump) and links to each, fixed a stale anchor left over from a heading rename, and removed duplicated command snippets in the "Commands" reference section that risked drifting from the canonical steps.
- **New skill** (`.ai/skills/token-update/SKILL.md`): a thin dispatcher skill that routes any "update/bump/upgrade tokens" request to the correct README section (asking the user which path applies if ambiguous) without duplicating the steps themselves.

### Notes

- Only the `opacity-*` scale is available in the npm package; the `transparent-static-*` / `transparent-neutral-*` tokens from the AI-branding work are **Figma-only** and cannot ship in SWC.
- Off-scale alphas (5%, 10%, 12.5%, 50%, 75%), multi-value `box-shadow` alphas (the `postcss-token` plugin only replaces one `token()` per declaration), and the animation-opacity vars were intentionally left as-is.

## How has this been tested?

- VRTs show no visual changes, as expected
- `swc-tokens` unit tests: 48/48 pass.
- Full `yarn build` of `@adobe/spectrum-wc` succeeds (moved color tokens for popover/avatar/etc. resolve).
- `stylelint` clean; regenerated `typography.css` matches the committed file exactly.
- Verified the built prompt-field CSS resolves to `rgba(255,255,255,.04)` / `.02`.
- Re-ran `node scripts/diff-versions.js --to 15.2.0` after the resolution and merge fixes to confirm it completes successfully and that `custom/deleted.json` retains all previously curated entries while only appending the 5 newly-found deletions.


cdransf (approved):
Looks great! ✨

pfulton (approved):

rubencarvalho:
Lots of VRTs are failing 🤔

pfulton:
> Lots of VRTs are failing 🤔

Looks like (potentially) a spacing change. Not sure if this was anticipated.

PR #6660, opened 2026-08-27
feat(ai): add spectrum-wc agent skill for 2nd-gen components, patterns, controllers

blunteshwar:
## Description

Adds a third generated Agent Skill, `spectrum-wc`, alongside the existing `spectrum-web-components` (Gen 1) and `migrate-swc-gen1-to-gen2` skills produced by `scripts/generate-agent-skills.mjs`.

The new skill covers `@adobe/spectrum-wc` and `@adobe/spectrum-wc-core` (Gen 2): one reference file per component, per pattern, and per core controller, generated from each unit's existing `<name>.mdx` Storybook docs page.

Changes to `scripts/generate-agent-skills.mjs`:

- Added listing helpers for `swc/components`, `swc/patterns` (one level deeper: `patterns/<group>/<unit>/<unit>.mdx`), and `core/controllers`.
- `stripMdx` now also drops bare `<DocsHeader />` / `<DocsFooter />` tags — Storybook chrome with no static Markdown equivalent (title/subtitle/description come from `stories.ts` meta; the API table is rendered live from the Custom Elements Manifest).
- New `buildApiTables()` / `buildApiSection()` read `2nd-gen/packages/swc/dist/custom-elements.json` (written by `yarn analyze`, which runs as part of `@adobe/spectrum-wc`'s `build` script, already produced before `generate:skills` runs in CI) and render Markdown tables for Properties, Slots, Events, CSS Custom Properties, and CSS Parts, appended as a `## API` section — mirroring the categories in the live `<ApiTable />` Storybook block.
- Multi-element units (e.g. `accordion`, `tabs`) declare sibling element tags via `parameters.additionalApiTables` in their `stories.ts`, and `DocsFooter` renders a separate table per tag. `readAdditionalApiTags()` extracts that array so the generator produces one labeled sub-table per tag (e.g. `swc-tabs` also gets `swc-tab` and `swc-tab-panel`; `swc-accordion` also gets `swc-accordion-item`), matching the live docs page exactly.
- Controllers are excluded from API-table generation (they aren't custom elements) since each controller's `.mdx` already hand-authors its own `## API` section.
- New skill source prose: `2nd-gen/packages/ai/skills/spectrum-wc-skill/SKILL.md`.

### Update: rebuild DocsHeader content, preserve code fences, namespace patterns by group

- `buildDocsHeader()` / `buildStoriesMeta()` (via a new `readStoriesMeta()`) rebuild the title/subtitle/description/Getting-started Markdown that `<DocsHeader />` renders in Storybook, so generated reference pages carry that content instead of losing it along with the rest of the stripped JSX.
- `stripMdx` now swaps fenced ` ```code``` ` blocks for placeholders before running its MDX-stripping passes and restores them verbatim afterward, so an example snippet's own `import ...` line (or other MDX-looking text inside a fence) survives intact instead of being mistaken for real MDX syntax.
- Pattern units are now namespaced by group (`references/patterns/<group>/<unit>.md`) via a new `unitSlug()` helper, so two pattern groups can each contain a same-named unit without one's generated reference overwriting the other's.
- `readAdditionalApiTags()` and `readStoriesMeta()` now share a single cached read of each unit's `stories.ts` instead of reading it twice.

## Motivation and context

The Gen 1 and migration skills give agents API references and step-by-step guidance for `sp-*` components and the Gen 1 → Gen 2 migration path, but there was no equivalent skill for building directly with Gen 2 (`swc-*`) components, patterns, or core controllers. This closes that gap using the same generation pipeline, so it stays in sync with the published docs automatically.

## Related issue(s)

No linked issue; standalone tooling addition.

N/A — build script change; output is generated Markdown/JSON under `.well-known/agent-skills/spectrum-wc/`.

## Author's checklist

## Reviewer's checklist

### Manual review test cases

- [ ] _Generated skill output is correct_
  1. Run `yarn workspace @adobe/spectrum-wc build` (or `yarn build`) to produce `2nd-gen/packages/swc/dist/custom-elements.json`.
  2. Run `node scripts/generate-agent-skills.mjs`.
  3. Expect a third skill directory at `2nd-gen/packages/swc/public/.well-known/agent-skills/spectrum-wc/` with `references/components/`, `references/patterns/`, `references/controllers/`, and `references/guides/`, and `index.json` listing all three skills.
  4. Open `references/components/action-button.md` and confirm the `## API` section renders real Properties/Slots/CSS-custom-properties tables (not `{{TOKEN}}` placeholders or empty tables).
  5. Open `references/controllers/hover-controller.md` and confirm its existing hand-authored `## API` section is unchanged (no duplicate CEM-generated table appended).
  6. Open `references/components/tabs.md` and `references/components/accordion.md` and confirm the `## API` section has one labeled sub-table per tag (`swc-tabs`/`swc-tab`/`swc-tab-panel`, and `swc-accordion`/`swc-accordion-item` respectively), matching the sibling tables on the live Storybook docs pages.
  7. Open any component reference page (e.g. `references/components/badge.md`) and confirm it now starts with a `#` title, subtitle, description, and `## Getting started` section mirroring `<DocsHeader />`.
  8. Open a reference page whose source `.mdx` ends in a fenced code block and confirm the code sample is intact (not replaced by a `CODE_BLOCK_N` placeholder).

### Device review

N/A — no rendered UI change; not applicable to a build script.

N/A — this PR only changes a Node build script and adds generated documentation output. It introduces no new UI, component, or interactive behavior, so there is no keyboard or screen reader surface to test.


2nd-gen/packages/ai/skills/spectrum-wc-skill/SKILL.md
```diff
@@ -0,0 +1,94 @@
...
+  author: 'Adobe'
+  website: 'https://spectrum-web-components.adobe.com'
+---
+
+# Spectrum Web Components (Spectrum 2)
+
+Spectrum 2 Web Components (`swc-*`) is Adobe's Spectrum 2 design system implemented
+as framework-agnostic web components. All components ship from a single package,
+`@adobe/spectrum-wc`, with shared primitives (controllers, mixins, base classes)
+in `@adobe/spectrum-wc-core`.
```
rubencarvalho, line 15: I'm not sure this needs to be surfaced 🤔 Given consumers can't shouldn't reach out to `core` yet, this may give the wrong impression to the agent

2nd-gen/packages/ai/skills/spectrum-wc-skill/SKILL.md
```diff
@@ -0,0 +1,94 @@
...
+## Quick start
+
+```bash
+yarn add @adobe/spectrum-wc
+```
+
+```ts
+// App entry: import once
+import '@adobe/spectrum-wc/swc.css';
+import '@adobe/spectrum-wc/components/button/swc-button.js';
```
rubencarvalho, line 29: This may make it seem the button should be imported at the top level app entry, which is not necessarily correct. This only applies to the tokens above

2nd-gen/packages/ai/skills/spectrum-wc-skill/SKILL.md
```diff
@@ -0,0 +1,94 @@
...
+```
+
+```html
+<html class="swc-theme swc-theme--light">
```
rubencarvalho, line 33: Probably worth a super small disclaimer about the theme classes

2nd-gen/packages/ai/skills/spectrum-wc-skill/SKILL.md
```diff
@@ -0,0 +1,94 @@
...
+    <swc-button variant="accent">Save</swc-button>
+  </body>
+</html>
+```
+
+## Key concepts
+
+- **Single package, subpath imports**: `@adobe/spectrum-wc` is a monolithic
+  package; import each component from its own subpath so bundlers only include
+  what is used: `@adobe/spectrum-wc/components/badge/swc-badge.js`.
```
rubencarvalho, line 44:
```suggestion
  what is used, e.g.: `@adobe/spectrum-wc/components/badge/swc-badge.js`.
```

2nd-gen/packages/ai/skills/spectrum-wc-skill/SKILL.md
```diff
@@ -0,0 +1,94 @@
...
+
+- **Theme classes, not a theme element**: unlike Spectrum 1's `<sp-theme>`
+  wrapper, Spectrum 2 applies theme and scale via CSS classes on an ancestor
+  element: `swc-theme--light`, `swc-theme--dark`, `swc-theme--adaptive`.
+
+- **Side-effectful import**: importing the `.js` entry registers the custom
+  element. Import the class directly (e.g. from `Button.js`) for extension or typing.
```
rubencarvalho, line 51: Maybe example of that path? And maybe we can discourage this in favor of the always side effectful import for now

2nd-gen/packages/ai/skills/spectrum-wc-skill/SKILL.md
```diff
@@ -0,0 +1,94 @@
...
+
+- **`@adobe/spectrum-wc-core`**: shared controllers (hover, placement, focus
```
rubencarvalho, line 53: Not a public surface

2nd-gen/packages/ai/skills/spectrum-wc-skill/SKILL.md
```diff
@@ -0,0 +1,94 @@
...
+  Reach for a controller instead of re-implementing common interaction logic.
+
+- **Accessibility**: `swc-*` components expose correct ARIA roles, labels, and
+  keyboard navigation. Each component reference documents its accessibility
+  features and best practices.
+
+## Documentation structure
+
+The `references/` directory contains guides and one Markdown file per component,
+pattern, and controller. Each component/pattern reference ends with an **API**
```
rubencarvalho, line 64: From the customer-facing perspective, core is an internal abstraction. The agent should not reach out to it.

2nd-gen/packages/ai/skills/spectrum-wc-skill/SKILL.md
```diff
@@ -0,0 +1,94 @@
...
+
+### Patterns
+
+Higher-level compositions of multiple components, one file per pattern in `references/patterns/`.
+
+Available patterns: {{GEN2_PATTERN_NAMES}}.
+
+{{GEN2_PATTERN_LIST}}
+
+### Controllers
```
rubencarvalho, line 89: Controllers are not public

2nd-gen/packages/ai/skills/spectrum-wc-skill/SKILL.md
```diff
@@ -0,0 +1,94 @@
...
+
+### Guides
+
+{{GEN2_GUIDE_LIST}}
+
+### Components
+
+One file per component in `references/components/` (e.g. `references/components/badge.md`).
+
+Available components: {{GEN2_COMPONENT_NAMES}}.
```
rubencarvalho, line 77: I think we would benefit from a brief dos/donts of how / when to use the components. I know this is a bit of a stretch, but maybe we can have a follow up action item of surfacing the Usage guidelines in our own docs for each component (e.g. https://s2.spectrum.corp.adobe.com/page/badge/#usage-guidelines) and maybe we can start by reusing the small frontmatter so the agent knows when to reach out to that specific component...

scripts/generate-agent-skills.mjs
```diff
@@ -607,6 +1180,15 @@ const SKILL_CONFIGS = [
     kind: 'migration',
     buildFn: buildMigrationSkill,
   },
+  {
+    name: 'spectrum-wc',
+    description:
+      'Build UIs with Spectrum 2 Web Components (swc-* elements, @adobe/spectrum-wc, ' +
+      '@adobe/spectrum-wc-core). Use when developers are working with @adobe/spectrum-wc ' +
```
rubencarvalho, line 1187: No core

scripts/generate-agent-skills.mjs
```diff
@@ -607,6 +1180,15 @@ const SKILL_CONFIGS = [
...
+      'or @adobe/spectrum-wc-core packages or swc-* custom elements. Includes component, ' +
```
rubencarvalho, line 1188: no core

scripts/generate-agent-skills.mjs
```diff
@@ -607,6 +1180,15 @@ const SKILL_CONFIGS = [
...
+      'pattern, and controller API references and usage guidance.',
```
rubencarvalho, line 1189: no controllers

2nd-gen/packages/ai/skills/spectrum-wc-skill/SKILL.md
blunteshwar: Agreed, removed the core mention here. The skill now only points to `@adobe/spectrum-wc`.

2nd-gen/packages/ai/skills/spectrum-wc-skill/SKILL.md
blunteshwar: Good catch. Split it up: only `swc.css`/tokens at app entry, and the component import is shown separately where it's used.

2nd-gen/packages/ai/skills/spectrum-wc-skill/SKILL.md
blunteshwar: Added a short note: apply the theme/scale classes once on an ancestor (e.g. `<html>`), not per component.

2nd-gen/packages/ai/skills/spectrum-wc-skill/SKILL.md
blunteshwar: Applied, thanks.

2nd-gen/packages/ai/skills/spectrum-wc-skill/SKILL.md
blunteshwar: Added an example path and reworded to prefer the side-effectful import; the direct class import is now flagged as type/extension only.

2nd-gen/packages/ai/skills/spectrum-wc-skill/SKILL.md
blunteshwar: Removed. Core isn't a public surface, so it's out of the skill.

2nd-gen/packages/ai/skills/spectrum-wc-skill/SKILL.md
blunteshwar: Agreed. Dropped core/controllers from the docs structure; the agent only sees components and patterns now.

2nd-gen/packages/ai/skills/spectrum-wc-skill/SKILL.md
blunteshwar: Removed the Controllers section, and the generator no longer emits any controller references.

2nd-gen/packages/ai/skills/spectrum-wc-skill/SKILL.md
blunteshwar: Really like this. Out of scope for this PR, but noting it here: SWC-2594 as a follow-up to pull each component's Usage guidelines frontmatter into the skill.

scripts/generate-agent-skills.mjs
blunteshwar: Removed core from the description.

scripts/generate-agent-skills.mjs
blunteshwar: Done, no core in the description.

scripts/generate-agent-skills.mjs
blunteshwar: Removed controllers too; the description now reads components and patterns only.

scripts/generate-agent-skills.mjs
```diff
@@ -443,6 +484,402 @@ function listMigrationComponents() {
...
+ * `<unitName>.mdx` doc page, tagged `swc-<unitName>` by convention.
+ * Returns [{ dir, mdxPath, tagName }] sorted by dir.
+ */
+function listGen2Units(dir) {
+  return readdirSync(dir, { withFileTypes: true })
+    .filter((entry) => entry.isDirectory())
+    .map((entry) => ({
+      dir: entry.name,
+      mdxPath: join(dir, entry.name, `${entry.name}.mdx`),
+      tagName: `swc-${entry.name}`,
```
Rajdeepc: This discovery rule assumes every public pattern uses `<directory>/<directory>.mdx`, but `patterns/ai-toolkit/suggestion/suggestion-group.mdx` is a valid public pattern doc that does not match it. As a result, `swc-suggestion-group` is omitted from both the generated references and the skill index. Could we discover the actual public MDX file (while still excluding `*.internal.mdx`) rather than deriving its name solely from the directory?

scripts/generate-agent-skills.mjs
```diff
@@ -443,6 +484,402 @@ function listMigrationComponents() {
...
+    dirname(unit.mdxPath),
+    'stories',
+    `${unit.dir}.stories.ts`
+  );
+  const source = readStoriesSource(storiesPath);
+  if (source === null) {
+    return null;
+  }
+
+  const metaStart = source.indexOf('const meta');
```
Rajdeepc: This finds the substring `const meta` after the `export ` prefix for files that use `export const meta`. The JSDoc lookup below then examines text ending in `export `, so it cannot recognize the adjacent documentation block. Confirmed in generated references for Tabs and Illustrated message: their title, subtitle, and getting-started content are present, but their Storybook descriptions are missing. Please support both declaration forms without splitting the JSDoc from its `meta` declaration.

Rajdeepc (dismissed):
Thanks for building this on the existing skill-generation pipeline and deriving the static references from the same Storybook/CEM sources. I found two coverage issues that need to be resolved before this can reliably serve as the Spectrum 2 reference: one valid public pattern is omitted because its MDX filename differs from its directory name, and descriptions are dropped for valid `export const meta` story files. The inline comments include the affected paths and expected behavior.

scripts/generate-agent-skills.mjs
blunteshwar: Good catch, fixed. `listGen2Units` now discovers the actual public .mdx in each unit dir (excluding `*.internal.mdx` and `migration-guide.mdx`) instead of assuming `<dir>/<dir>.mdx`, deriving the name and `swc-` tag from the doc filename. `swc-suggestion-group` is now generated and listed in the index.

scripts/generate-agent-skills.mjs
blunteshwar: Fixed. The meta parser now matches both const meta and export const meta and anchors at the keyword, so the JSDoc block is no longer split off by the export  prefix. Tabs and Illustrated message now include their Storybook descriptions.

scripts/generate-agent-skills.mjs
blunteshwar: Good catch, fixed. `listGen2Units` now discovers the actual public `.mdx` in each unit dir (excluding `*.internal.mdx` and `migration-guide.mdx`) instead of assuming `<dir>/<dir>.mdx`, and derives the name and `swc-` tag from the doc filename. `swc-suggestion-group` is now generated and listed in the index.

scripts/generate-agent-skills.mjs
blunteshwar: Fixed. The meta parser now matches both `const meta` and `export const meta` and anchors at the keyword, so the JSDoc block is no longer split off by the `export ` prefix. Tabs and Illustrated message now include their descriptions.

scripts/generate-agent-skills.mjs
```diff
@@ -585,6 +1057,64 @@ function buildMigrationSkill(skillDir) {
...
+      GEN2_COMPONENT_LIST: buildGen2UnitList(components, 'components'),
+      GEN2_PATTERN_NAMES: patterns.map((c) => `\`${c.tagName}\``).join(', '),
+      GEN2_PATTERN_LIST: buildGen2UnitList(patterns, 'patterns'),
+    })
+  );
+
+  function writeUnitDocs(units, subdir) {
+    for (const unit of units) {
+      const meta = readStoriesMeta(unit);
+      let content = stripMdx(readFileSync(unit.mdxPath, 'utf8'));
```
Rajdeepc, line 1269: This is preserving Storybook relative links such as `../?path=/docs/...`, `?path=/docs/...`, and `/docs/...`. Once we install this skill as loose markdown files these will resolve to non existent local paths rather than Spectrum document site links. 
You should rewrite bundled targets to relative generated `md` files and unbundled targets to absolute 
`https://spectrum-web-components.adobe.com/...` URLS. Also add generated output link validation

scripts/generate-agent-skills.mjs
```diff
@@ -443,6 +484,429 @@ function listMigrationComponents() {
...
+    ],
+    [
+      'CSS custom properties',
+      mdTable(
+        ['Name', 'Default', 'Description'],
+        (component.cssProperties ?? []).map((prop) => [
+          code(prop.name),
+          prop.default != null ? code(prop.default) : '-',
+          prop.description,
+        ])
```
Rajdeepc: You are emitting events and css properties without validation and deduplication. This will generate phantom events, blank event rows and duplicate CSS custom properties. So the agents will generate listeners for non existent events and interpret duplicate entries as separate APIs.
You need to reject entries without names, exclude analyzer artifacts such as pseudo event and deduplicate each API section by name using precedence.
Add some CEM cases with fixture based tests too.

scripts/generate-agent-skills.mjs
blunteshwar: Fixed. Generated references now rewrite Storybook doc links — in-skill targets become relative `.md` links, and everything else an absolute `https://spectrum-web-components.adobe.com/...` link (Storybook adds a per-directory `titlePrefix`, so the doc id is `kebab(<section>/<title>)`). Added generate-time validation that fails on any unresolved `/docs/` link or broken relative `.md` target.

scripts/generate-agent-skills.mjs
blunteshwar: Fixed. API tables now drop nameless CEM entries (the null/pseudo event that caused blank rows) and dedupe each section by name — no more phantom events or duplicate CSS custom properties (e.g. `--swc-button-down-state-transform`); the default slot's empty name is preserved. The same validation pass fails generation on blank/duplicate rows. On fixture tests: the script isn't wired as a testable module yet (CI runs only the web-test-runner suites), so this ships as a generate-time guard — happy to add isolated CEM fixtures as a follow-up if you'd prefer.

scripts/generate-agent-skills.mjs
rubencarvalho: Now we're missing for example the main conversational composition/anatomy guidance which sits at the root of the pattern `patterns/ai-toolkit/pattern-overview.mdx` 😛

gen2/packages/ai/skills/spectrum-wc-skill/SKILL.md
```diff
@@ -0,0 +1,88 @@
...
+  element: `swc-theme--light`, `swc-theme--dark`, `swc-theme--adaptive`.
+
+- **Side-effectful import**: importing a component's `.js` entry registers its
+  custom element (e.g. `@adobe/spectrum-wc/components/button/swc-button.js`).
+  Prefer this registering import; import the class directly (e.g.
+  `@adobe/spectrum-wc/components/button/Button.js`) only for extension or typing.
+
+- **Accessibility**: `swc-*` components expose correct ARIA roles, labels, and
+  keyboard navigation. Each component reference documents its accessibility
+  features and best practices.
```
rubencarvalho:
```suggestion
- **Accessibility**: components handle their own roles, state, and keyboard
  interaction, but some require attributes you must supply. Read the
  accessibility section of each component's reference before using it.
```

rubencarvalho (approved):
Approved and left 1 small suggestion 😄

Rajdeepc (approved):

PR #6658, opened 2026-08-27
chore(text-field): scaffold 2nd-gen structure

miwha-adobe:
## Description

Phase 2 (Setup) of the `text-field` migration: scaffolds the core and swc file structure. Stub files only — they build and register `<swc-text-field>`, but no public API, rendering, or styling is implemented yet. Controller wiring (SWC-2466/2467) is deferred to Phase 3.

## Motivation and context

Establishes a buildable foundation so the API, a11y, and styling phases can proceed on a stable structure. Part of Epic SWC-2323.

## Related issue(s)

Epic SWC-2323 · builds on #6629

## Author's checklist

- [x] Read CONTRIBUTING and PULL_REQUESTS
- [x] Reviewed accessibility practices
- [ ] Automated tests — N/A, scaffold only (Phase 6)
- [ ] Changeset — N/A, package not yet published
- [x] Updated docs (migration plan + status)

### Manual review test cases

- [ ] `yarn workspace @spectrum-web-components/2nd-gen build` succeeds
- [ ] `import '@adobe/spectrum-wc/components/text-field/swc-text-field.js'` registers `<swc-text-field>`; importing the class alone does not

### Device review

- [ ] Did it pass in Desktop?
- [ ] Did it pass in (emulated) Mobile?
- [ ] Did it pass in (emulated) iPad?

## Accessibility testing checklist

Non-functional scaffold — no API, labels, or behavior yet. Keyboard and screen reader verification are scoped to Phase 4.

PR #6669, opened 2026-08-27
feat(text-field): migrate API and partial a11y

miwha-adobe:
## Description

Phase 3 (API) of the text-field migration, plus the one Phase 4 (a11y) piece that's independent of the deferred controllers. 2nd-gen 

`TextField.base.ts` gets the full property surface (`accessible-label`, `value`, `name`, `type`, `placeholder`, `pattern`, `inputmode`, `autocomplete`, `maxlength`/`minlength`, `readonly`, `required`, `invalid`, `valid`, `label-position`, `disabled`) with `validateEnum` for `type` and `label-position`, and `delegatesFocus` so the field is a single tab stop. 

The removed/renamed 1st-gen surface (`label`, `quiet`, `multiline`/`grows`/`rows`, `tooltip-placement`, `autocomplete` list/none) is deprecated with `@deprecated` + `__swc.warn`, scoped to `sp-textfield` so the shared-base siblings are unaffected.

## Motivation and context

Establishes the public API and focus behavior before the controllers (SWC-2466 labelling, SWC-2467 form association) and Phase 5 render land — those will refactor this phase to adopt them via the `@todo` seams. Part of Epic SWC-2323.

## Related issue(s)

Epic SWC-2323 · stacked on #6658

## Author's checklist

- [x] Read CONTRIBUTING and PULL_REQUESTS
- [x] Reviewed accessibility practices
- [ ] Automated tests — existing 1st-gen suites verified green (textfield, number-field, color-field, combobox, search); new 2nd-gen tests land in Phase 6
- [ ] Changeset — N/A, package not yet published
- [x] Updated docs (migration plan: Q27, Phase 3/4 checklists)

### Manual review test cases

- [x] `yarn workspace @spectrum-web-components/2nd-gen build` succeeds
- [ ] Deprecation warnings fire only for `<sp-textfield>`, not for `sp-color-field`/`sp-number-field`/`sp-combobox`/`sp-search`
- [x] `<swc-text-field>` is a single tab stop; focus lands on the inner `<input>`

### Device review

- [ ] Did it pass in Desktop?

`delegatesFocus` is the only a11y behavior in this PR; accessible naming, description/error association, and `aria-invalid`/`aria-errormessage` are deferred to SWC-2466 and Phase 5 render.

- [x] **Keyboard** — <kbd>Tab</kbd> reaches `<swc-text-field>` as one stop and focus lands on the inner `<input>`; <kbd>Shift</kbd> + <kbd>Tab</kbd> leaves cleanly.
- [x] **Screen reader** — the inner `<input>` exposes role `textbox`. (Name/description wiring lands with the LabellingController, SWC-2466.)

### DIFF
--- a/1st-gen/packages/textfield/src/Textfield.ts
+++ b/1st-gen/packages/textfield/src/Textfield.ts
@@ -93,6 +93,8 @@ export class TextfieldBase extends ManageHelpText(
 
   /**
    * A string applied via `aria-label` to the form control when a user visible label is not provided.
+   *
+   * @deprecated Renamed to `accessible-label` in Spectrum 2.
    */
   @property()
   public label = '';
@@ -136,6 +138,8 @@ export class TextfieldBase extends ManageHelpText(
   /**
    * Whether a form control delivered with the `multiline` attribute will change size
    * vertically to accomodate longer input
+   *
+   * @deprecated Multiline moves to `<swc-text-area>` in Spectrum 2.
    */
   @property({ type: Boolean, reflect: true })
   public grows = false;
@@ -154,6 +158,8 @@ export class TextfieldBase extends ManageHelpText(
 
   /**
    * Whether the form control should accept a value longer than one line
+   *
+   * @deprecated Multiline moves to `<swc-text-area>` in Spectrum 2.
    */
   @property({ type: Boolean, reflect: true })
   public multiline = false;
@@ -167,12 +173,16 @@ export class TextfieldBase extends ManageHelpText(
   /**
    * Placement of the tooltip shown when the value is truncated (e.g. 'bottom', 'top').
    * Defaults to 'bottom' per Spectrum design.
+   *
+   * @deprecated The truncated-value tooltip is removed in Spectrum 2.
    */
   @property({ attribute: 'tooltip-placement' })
   public truncatedValueTooltipPlacement: Placement = 'bottom';
 
   /**
    * The specific number of rows the form control should provide in the user interface
+   *
+   * @deprecated Multiline moves to `<swc-text-area>` in Spectrum 2.
    */
   @property({ type: Number })
   public rows = -1;
@@ -204,6 +214,8 @@ export class TextfieldBase extends ManageHelpText(
 
   /**
    * Whether to display the form control with no visible background
+   *
+   * @deprecated The `quiet` style is removed in Spectrum 2.
    */
   @property({ type: Boolean, reflect: true })
   public quiet = false;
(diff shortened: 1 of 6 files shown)


1st-gen/packages/textfield/src/Textfield.ts
```diff
@@ -427,6 +441,78 @@ export class TextfieldBase extends ManageHelpText(
...
+      'https://opensource.adobe.com/spectrum-web-components/components/textfield/';
+    // Scope to sp-textfield: subclasses (number-field, color-field, combobox,
+    // search) use these properties legitimately and migrate on their own schedule.
+    if (window.__swc?.DEBUG && this.localName === 'sp-textfield') {
+      if (changedProperties.has('label') && this.label.length > 0) {
+        window.__swc.warn(
+          this,
+          `The "label" attribute on <${this.localName}> has been deprecated and will be removed in a future release. Set "aria-label" or "aria-labelledby" on the element instead.`,
+          url,
+          { level: 'deprecation' }
```
cdransf, line 456:
```suggestion
          { type: 'label', level: 'deprecation' }
```

Can we give each of these a distinct type? Otherwise they'll all collapse into this first one ✨

cdransf (approved):
Looks good! Just a suggestion for the deprecation warnings. ✨

1st-gen/packages/textfield/src/Textfield.ts
```diff
@@ -427,6 +441,78 @@ export class TextfieldBase extends ManageHelpText(
...
+      }
+      if (changedProperties.has('grows') && this.grows) {
+        window.__swc.warn(
+          this,
+          `The "grows" attribute on <${this.localName}> has been deprecated and will be removed in a future release.`,
+          url,
+          { level: 'deprecation' }
+        );
+      }
+      if (changedProperties.has('rows') && this.rows !== -1) {
```
rise-erpelding, line 483: Can you remind me, did we only address the dedup thing (only showing 1 warning when more than 1 applies) for 2nd-gen? So, for instance if `multiline` and `rows` both apply we would expect to only see `multiline`?

rise-erpelding (approved):
Question about warnings (are you tired of talking about them yet? 😜 ), but that doesn't have to block anything, the 2nd-gen validation steps are all passing! Thanks for adding a story so I could test it!

1st-gen/packages/textfield/src/Textfield.ts
miwha-adobe: Yep! Exactly, we didn't address the dedup issue in first gen so that would be the expected behavior. Not ideal but the refactor was scoped to 2nd gen.

1st-gen/packages/textfield/src/Textfield.ts
miwha-adobe: So the types are a closed union ('default' | 'accessibility' | 'api'), so I wouldn't be able to add a distinct type without expanding the union. Our dev-warning refactor was scoped to 2nd gen, but it brings up a good point. Currently all of our 1st gen dev-warnings would be effected by this collapse as you said. I'm curious if we want to expand the refactor to 1st gen, I can bring it up as a question or include it in the code quality talk.

PR #6671, opened 2026-08-27
fix: stop orange band in prompt field hover wash

rubencarvalho:
## Description

On hover, the prompt field's card wash showed a hard orange diagonal band (looks glitchy). Root cause: the card background gradient interpolated `in oklch`. Its four stops span far-apart hues, and OKLCH interpolates **hue around the color wheel** — the shorter arc between the warm and cool stops passes straight through orange, painting a saturated band where none of the stops are orange.

This is **not** sRGB/GPU banding (that would show as concentric stepped contours following the radial shape, not a single diagonal line of a foreign hue).

### Fix
Interpolate the wash `in oklab` instead. OKLab blends along the straight chord between stop colors (no hue-wheel traversal), so no foreign hue is introduced. Endpoint colors are unchanged.

```diff
- ...bottom in oklch, stop-1 0%, stop-2 35%, stop-3 82%, stop-4 100%)
+ ...bottom in oklab, stop-1 0%, stop-2 35%, stop-3 82%, stop-4 100%)
```

### VRT
Added a **Hover** permutation to the prompt-field VRT (variant × generating), forcing the field's own `:hover` treatment via `forcePseudoState` on `.swc-PromptField-outer-border`. The existing VRT only forced button hover/focus, so the field wash itself was never snapshotted.

## Note
The outer ring gradient (`.swc-PromptField-outer-border`) also interpolates `in oklch` across far-apart hues, so it could show the same sweep in principle. It's left unchanged here: it's ~2% opacity and not part of the report. Happy to fold it in if reviewers want.

## How to test
Storybook → Conversational AI → Prompt field → hover the card; the orange band is gone. Run the VRT to see the new Hover row.


TarunAdobe (approved):
nit: forceFieldHover() duplicates the shared forcePseudoStates() helper. Use data-force-state="hover" with the existing helper and remove ~10 lines

5t3ph (approved):
TIL - nice!

PR #6673, opened 2026-08-27
fix: give prompt field send button an auto static translucent fill

rubencarvalho:
## Description

The send button used a **solid opaque `neutral-background-color` fill**, which clashed with the card's gradient wash. Switch it to an auto-adjusting translucent treatment so the gradient shows through, matching the transparent treatment the upload/add button already uses.

### How
The send button is anchored bottom-right, which is exactly where the wash gradient originates (`circle at right bottom`, i.e. stop-1). So its fill and icon are derived from `--_swc-prompt-field-bg-stop-1`:

- **Fill**: `oklch(from stop-1 calc((0.5 - l) * infinity) 0 0deg)` at 85% (95% hover/down) — a grayscale overlay that snaps **dark-on-light / light-on-dark** by the wash luminance behind it, kept translucent.
- **Icon**: the same trick inverted (`(l - 0.5)`), full opacity, so it always contrasts the fill.
- **Disabled**: same tint at 11% (was already using this pattern).

This reuses the exact auto-contrast relative-color trick the send **disabled** state already used, now applied to the live states.

### Scope
- **Send**: changed (the clashing one).
- **Upload/add**: left as-is; it already uses transparent overlay tokens (`transparent-black/white-*` via `light-dark`), so it already lets the gradient through.
- **Stop**: unchanged.

## How to test
Storybook → Conversational AI → Prompt field. The send button now reads as a translucent circle that tints with the wash instead of a flat neutral fill, in both light and dark. Existing VRT already snapshots send hover/focus.


TarunAdobe (approved):

2nd-gen/packages/swc/patterns/conversational-ai/prompt-field/prompt-field.css
```diff
@@ -878,13 +878,20 @@
...
 .swc-PromptField-send {
-  --swc-action-button-background-color-default: token("neutral-background-color-default");
-  --swc-action-button-background-color-hover: token("neutral-background-color-hover");
-  --swc-action-button-background-color-focus: token("neutral-background-color-default");
-  --swc-action-button-background-color-down: token("neutral-background-color-default");
-
-  /* Disabled falls back to the auto-contrast tint rather than the flat fill. */
-  --swc-action-button-background-color-disabled: oklch(from var(--_swc-prompt-field-bg-stop-1) calc((0.5 - l) * infinity) 0 0deg / 11%);
+  /* Auto static treatment from stop-1 (under this button): translucent fill snaps dark-on-light / light-on-dark, icon takes the opposite, so the gradient shows through. */
+  --_swc-prompt-field-send-fill: oklch(from var(--_swc-prompt-field-bg-stop-1) calc((0.5 - l) * infinity) 0 0deg);
```
5t3ph: This seems complex for effectively getting white or black with an 85-95% alpha as far as final computed values are concerned 😅 To "the naked eye" there's no visible color shift beyond white or black no matter what brand color I put in:

Image

If the effectively white or black is the intended output, can we simplify this to color mix against `gray-25` (essentially `light-dark(black, white)`) to just add the alpha channel?

rubencarvalho: Good catch, yes it was always resolving to pure black/white. Simplified to use the static `black`/`white` tokens directly with `light-dark()`, applying alpha via `color-mix()` in bf52ad041f.

5t3ph (approved):

PR #6680, opened 2026-08-31
perf(response-status): make active-step shimmer GPU-friendly

rubencarvalho:
## What

The "in progress" text shimmer on the active response step used to sweep by animating `background-position` on `background-clip: text`. This PR swaps that for a masked overlay that animates `translate` instead.

## Why

Animating `background-position` isn't GPU-accelerated — the browser has to repaint the text every single frame, which is expensive for something running continuously while a step is active. Animating `translate` runs on the compositor instead, so the browser can slide the highlight around without repainting the text underneath.

## How

- The real label text is left alone (plain color, no animation).
- An invisible clone of the label sits on top of it, masked to its own letter shapes (`-webkit-mask-clip: text`) so only the part of it inside the letters is visible.
- A moving gradient band is animated across that clone using `translate`, which shows through the mask as a highlight sweeping across the text.

## Trade-offs

- Needs a small duplicate of the label text kept in sync with the real one (previously the effect lived entirely in CSS on the real text).
- **Firefox caveat:** the mask only works via `-webkit-mask-clip: text`, which Firefox doesn't support (no unprefixed equivalent exists yet). The whole thing is wrapped in `@supports (-webkit-mask-clip: text)`, so Firefox just skips it — the clone stays fully transparent and invisible, and the step title shows as plain, unanimated text there instead of shimmering.
- Reduced-motion and forced-colors users still just see flat text, same as before.

### DIFF
--- a/2nd-gen/packages/swc/patterns/conversational-ai/response-status/response-status-step/ResponseStatusStep.ts
+++ b/2nd-gen/packages/swc/patterns/conversational-ai/response-status/response-status-step/ResponseStatusStep.ts
@@ -76,13 +76,7 @@ export class ResponseStatusStep extends SpectrumElement {
   @state()
   private _hasDescription = false;
 
-  /**
-   * Perf prototype: mirrors the current label text so an inert clone can be
-   * rendered on top of the real label and masked to its glyph shapes for the
-   * shimmer overlay (see response-status-step.css). Kept separate from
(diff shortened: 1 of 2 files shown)


rubencarvalho:
thanks to @devongovett https://x.com/devongovett/status/2092991811157463500 😄

aramos-adobe (approved):
Very cool!

PR #6681, opened 2026-08-31
feat(prompt-field): add animateLoader property to control status loader animation

rubencarvalho:
## Summary
- The status pixel-loader previously animated automatically whenever `generating` was set. Animation ownership is moving to `swc-response-status`, so `swc-prompt-field`'s status loader now stays static by default.
- Add a new `animateLoader` property (attribute `animate-loader`), boolean, default `false`. The loader only animates when both `generating` and `animate-loader` are set; `?paused=${!(this.generating && this.animateLoader)}` in `_renderStatusIcon()`.
- Add an `Animate loader` behaviors story + MDX section demonstrating static vs. animated.
- Update `StatusLoaderTest` in `prompt-field.test.ts` to cover the new decoupling (generating + animate-loader animates; generating alone stays paused).
- Drop the full-pattern conversation demo's `mode=${...}` attribute on `<swc-prompt-field>` (not a real `PromptField` property, and no longer needed now that `swc-response-status`'s `active` status supplies the animation there).

## Test plan
- [x] `tsc --noEmit` shows no new errors from these changes
- [ ] `yarn test` / `yarn test:a11y` pass for prompt-field and conversation-thread


miwha-adobe (approved):

2nd-gen/packages/swc/patterns/ai-toolkit/prompt-field/PromptField.ts
```diff
@@ -123,6 +123,10 @@ export class PromptField extends SpectrumElement {
   @property({ type: String, reflect: true })
   public loader: PixelLoaderIconName | PixelLoaderPresetName = 'aiLogo';
 
+  /** Animates the status loader while generating; otherwise it stays static. */
```
aramos-adobe: I would suggest adding something along the lines "Has no visible effect unless `generating` is also set to true". I know in the code both `generating && animateLoader = true` to make this happen but others looking at it first time in Storybook may not.

aramos-adobe (approved):

2nd-gen/packages/swc/patterns/ai-toolkit/prompt-field/PromptField.ts
rubencarvalho: Great suggestion - I added this!

PR #6684, opened 2026-08-31
fix(a11y-testing): stabilize flaky gen2 axe smoke tests

caseyisonit:
## Description

Reduces flakiness in the `test-gen2-axe` job (Storybook test-runner + axe-core). The failures were timing- and network-dependent, not real accessibility regressions.

## Changes

- **Remove the docs-view axe scan** (`test-runner.ts`): the scan was scoped to `#storybook-root`, but in docs mode Storybook renders into `#storybook-docs`, so it scanned nothing (a silent no-op). Dropping it also removes the docs page load and its `networkidle` wait, a real flake source. Docs-view accessibility coverage is tracked separately with the vitest-addon migration.
- **Font loader** (`font-loader.ts`): cap the wait for Adobe Fonts (external Typekit CDN) at 3s so a slow or blocked CDN can't hang a story render, and memoize the readiness promise at module scope so that ceiling is paid at most once per page instead of per story.
- **Settle before scanning** (`test-runner.ts`): force `prefers-reduced-motion` and wait for `document.fonts.ready` before the axe scan, so color-contrast is measured on a settled frame rather than mid-animation or fallback glyphs.
- **CI budget** (`continue_config.yml`): add `--maxWorkers=2` (prevents browser over-subscription on the xlarge container) and `--testTimeout=30000` (Jest's 5s default left no room for the font-settle wait).


cdransf (approved):

2nd-gen/packages/swc/.storybook/test-runner.ts
```diff
@@ -25,7 +26,28 @@ type StorybookTestContext = {
   };
 };
 
+// Wait for web fonts to settle so axe measures color-contrast against the final
+// glyphs, not the fallback face. `document.fonts.ready` resolves even when the
+// font CDN fails (unlike the custom `typekit-loaded` event), and the timeout
+// race guarantees the scan never hangs on a slow or blocked CDN.
+const FONT_SETTLE_TIMEOUT_MS = 3000;
```
Rajdeepc: This timeout is speculative.

.circleci/continue_config.yml
```diff
@@ -448,7 +448,7 @@ jobs:
           name: Run axe-core WCAG tests
           command: |
             mkdir -p 2nd-gen/packages/swc/test-results
-            yarn start-server-and-test "npx http-server 2nd-gen/packages/swc/storybook-static -p 6006 --silent" http://127.0.0.1:6006 "JEST_JUNIT_OUTPUT_DIR=./test-results JEST_JUNIT_OUTPUT_NAME=junit.xml yarn workspace @adobe/spectrum-wc test:storybook --url http://127.0.0.1:6006 --index-json --ci --junit"
+            yarn start-server-and-test "npx http-server 2nd-gen/packages/swc/storybook-static -p 6006 --silent" http://127.0.0.1:6006 "JEST_JUNIT_OUTPUT_DIR=./test-results JEST_JUNIT_OUTPUT_NAME=junit.xml yarn workspace @adobe/spectrum-wc test:storybook --url http://127.0.0.1:6006 --index-json --ci --junit --maxWorkers=2"
```
Rajdeepc: Jest normalizes `testTimeout` to 5000 so every story is getting 5 seconds to complete from preVisit to close. The problem with this PR is that it adds more waiting inside that same 5-second window which is up to 3s waiting for fonts before rendering, 3s more before the story scan, 3s more before the docs scan, and one wait set to 15 seconds, which is three times the entire budget. 

So under exactly the slow-network conditions this PR is meant to fix, it is making things worse, not better. The fix should be one flag with giving each test 60 seconds instead of 5.

```suggestion
            yarn start-server-and-test "npx http-server 2nd-gen/packages/swc/storybook-static -p 6006 --silent" http://127.0.0.1:6006 "JEST_JUNIT_OUTPUT_DIR=./test-results JEST_JUNIT_OUTPUT_NAME=junit.xml yarn workspace @adobe/spectrum-wc test:storybook --url http://127.0.0.1:6006 --index-json --ci --junit --maxWorkers=2 --testTimeout=60000"
```

2nd-gen/packages/swc/.storybook/test-runner.ts
```diff
@@ -120,10 +144,22 @@ const config: TestRunnerConfig = {
...
+      // Wait for the element axe scans, not network quiet. `networkidle` is
+      // timing-based and unreliable against the external font CDN; the selector
+      // wait is deterministic, and the catch keeps a missing root from throwing.
+      await docsPage.goto(docsUrl.toString(), {
+        waitUntil: 'domcontentloaded',
+      });
+      await docsPage
+        .waitForSelector('#storybook-root', {
+          state: 'attached',
+          timeout: 15000,
```
Rajdeepc: This is 3x the entire test budget. If that wait ever actually engages, Jest will kill the test at 5 s and the graceful `.catch(() => {})`  will never runs. Even after raising `--testTimeout`, a 15 s per-story wait is unbudgeted. Please cap it at `3-5s` and let a genuine failure fail than swallowing it.

2nd-gen/packages/swc/.storybook/loaders/font-loader.ts
```diff
@@ -22,5 +30,8 @@ export const FontLoader = async () => ({
         resolve();
       }
     });
+
+    // Fallback: never block the render on the external font CDN.
+    setTimeout(resolve, FONT_LOAD_TIMEOUT_MS);
   }),
 });
```
Rajdeepc: I would not recommend this. FontLoader is aregistered as a global loader so it runs on every story render. In the failure mode the Typekit script never loads and window.Typekit stays undefined forever, so all 338 stories pay the full 3 s. That's 17 min of added serialised wait, ~8.5 min at --maxWorkers=2 on an approximate basis. 
You should memoize at module sceope so the ceiling is paid once per page and clean up.
```ts
let fontsReady: Promise<void> | undefined;

const whenFontsReady = (): Promise<void> =>
  (fontsReady ??= new Promise<void>((resolve) => {
    if (typeof window.Typekit !== 'undefined') return resolve();
    const timer = setTimeout(resolve, FONT_LOAD_TIMEOUT_MS);
    document.addEventListener(
      'typekit-loaded',
      () => { clearTimeout(timer); resolve(); },
      { once: true }
    );
  }));

export const FontLoader = async () => ({ fonts: whenFontsReady() });
```

2nd-gen/packages/swc/.storybook/test-runner.ts
```diff
@@ -120,10 +144,22 @@ const config: TestRunnerConfig = {
     // Check docs on a separate page so Storybook test-runner state remains intact.
     const docsPage = await page.context().newPage();
```
Rajdeepc: Loading docs page once per story is costly. There are 338 stories across 40 story files and a components docs entry is indetical for all of its stories. So we are doing redundant full page loads  + axe scans per run. This is a bigger lever and a flake surface thant --maxWorkers.
My recommednation is to do it once per components. because each story file maps to one Jest test file and one docs entry a moduel scoped `Set` will dedupe like this
```ts
const docsScanned = new Set<string>();
// ...
if (!docsScanned.has(context.title)) {
  docsScanned.add(context.title);
  // ...existing docs block...
}
```

Rajdeepc (commented):
Some observations around docs scan and redundant runs on the font-loader. Nothing is blocking but if you can surface up these changes this will help us against something hitting back again in future.

.circleci/continue_config.yml
caseyisonit: I think 60 seconds is super long. What if we compromise on 10s to make up for font-loading delays, which should hopefully be resolved before 3s.

caseyisonit:
Thanks for the review @Rajdeepc — all four points addressed in dab67f77e1. Two of them converged on a deeper finding about the docs scan, so surfacing the reasoning here.

**Docs-view scan removed (not deduped).** Digging into your "load docs once per component" note, the docs scan wasn't just redundant, it was a no-op. The runner derived the docs URL by flipping `viewMode=docs` on the story id and then ran axe scoped to `#storybook-root`. But in docs mode Storybook renders into `#storybook-docs`, with each embedded example in its own nested docs-story canvas; `#storybook-root` is present-but-empty. So the docs pass scanned nothing and has never reported a violation. (Corroborated by timing: if the `#storybook-root` wait had actually been timing out, the 15s swallow across 338 stories would have blown well past the ~12m gen2 workflow.) Rather than dedupe a scan that tests nothing, I dropped it — which also removes the docs page load, the `networkidle`/selector wait, and that whole flake surface. Real docs-view a11y coverage is tracked with the upcoming vitest-addon migration.

**Font loader memoized** (your suggestion): the readiness promise now resolves at most once per page, so the blocked-CDN failure mode no longer costs the full 3s timeout per story.

**testTimeout:** you're right the waits didn't fit Jest's 5s default. Landed on `--testTimeout=30000`. Even with memoization and no docs scan, the heaviest story still pays the font-settle wait plus an axe run over a large permutation, so 10s felt tight and 60s over-pays on a genuine hang. `--maxWorkers=2` retained.

The `#storybook-root` / `domcontentloaded` / 15s-wait code is gone along with the docs block, so the "3x the budget" and "swallowed failure" concerns are moot now.

rise-erpelding (approved):

PR #6686, opened 2026-08-31
docs(toast): migration plan

rise-erpelding:
## Description

Makes some updates to the a11y analysis in order to clarify the direction for the migration plan:
1. **Removed the dead roadmap link.** The doc referenced rendering-and-styling-migration-analysis.md, which doesn't exist. Dropped the "Also read" TOC entry and section, and the matching References entry.

2. **Fixed the `role="status"` vs. `role="alertdialog"` contradiction.** The doc used to argue for `role="status"` in one section while its own Recommendations table already used `role="alertdialog"`. Replaced the status-based reasoning (Pattern in the APG bullets, two Guidelines table rows, Bottom line, References) with the alertdialog + inner alert framing now confirmed correct.

3. **Corrected the focus-on-open claim.** The doc previously said focus moves to the host when a toast opens (citing an RSP-team quote). Verified against the current React Spectrum source that this isn't true, tabIndex: 0 is present but nothing autofocuses a newly opened toast. Corrected in six places: the Recommendations table's "Host role" row, both ASCII accessibility-tree diagrams, the "Keyboard and focus" section's lead paragraph, and two summary-checklist items.

4. Flagged three known **divergences from React Spectrum**, without changing the recommendation:
    * Timeout floor: noted RSP floors at 5000ms, not this doc's 6000ms.
    * Timer pause scope: noted RSP pauses every visible toast in the region together, while this doc still describes per-toast pause.
    * `aria-labelledby`: noted RSP splits this into separate aria-labelledby (title) + aria-describedby (description), while this doc uses one simplified content ID.

5. **Added peek-stack and expand/collapse content**, now that the container is confirmed in scope: two new paragraphs in "Assistive technology, live regions" (decorative role="presentation" peek layers; the expanded list's scrim/Escape/focus-trap behavior, including that every toast becomes its own tabindex="0" alertdialog once expanded), plus two matching summary-checklist items.

6. **Fixed the close-button naming.** Both mentions of the close button's accessible name said **label**, the 1st-gen attribute. Corrected to **accessible-label**, the real **swc-close-button** attribute, verified against its source.

7. Marked the stale `warning` icon-label default as deprecated in the summary checklist (2nd-gen has four variants, no `warning`).

8. **A second pass on the a11y doc**, closing out remaining review feedback and later consistency checks:
    * Unified the focus-on-dismiss framing: closing via the close button, action button, or auto-dismiss now all describe the same container-managed focus behavior (nearest remaining toast, or return to the trigger), matching how React Spectrum's `useToastRegion` actually branches; it doesn't distinguish by removal cause.
    * Fixed one more stray reference to the close button's accessible name via a `label` property (should be `accessible-label`).
    * Expanded the Playwright ARIA-snapshot and manual screen-reader testing coverage to include the peek stack and the expanded list, not just the single-toast states.
    * Removed situational framing ("...for applications showing multiple toasts") now that multiple toasts are the default case, not a special one.
    * Corrected the container-ownership language: previously framed as an open design question to raise at a team sync. That sync happened and settled it, so the doc now states directly that a first-party container ships alongside `swc-toast`, and the matching stale checklist item was removed rather than checked off.

Based on these changes, adds the Phase 1 migration plan (`migration-plan.md`):

- 1st-gen API surface, package and shared-controller dependencies, and known 1st-gen issues carried forward as regression risks.
- Confirmed breaking changes: the variant set shrinks to four (`neutral`/`info`/`positive`/`negative`), S2 token adoption, the `role="alertdialog"` + inner `role="alert"` structure, host naming, `aria-hidden` on close, corrected timer-pause behavior, and `accessible-label` on the close button.
- Toast ships alongside a first-party container/queue in this migration, settled via a team sync. The container's shape and API are still being finalized incrementally across the plan: peek-stack rendering behind the front toast, expanding into a full list, focus management when a toast closes, and a queue-level `clear()`.
- A three-layer architecture split (Core / Queue / SWC): ARIA and property logic, the queue's own countdown state (which has to survive a toast being demoted to a decorative peek layer and promoted back), and the toast's own rendering. The expanded view's focus containment follows `swc-popover`'s existing `modal` mode (native `<dialog>.showModal()`), reusing its native focus trap, background inert, and backdrop rather than a hand-rolled implementation.
- Currently open questions: timeout minimum (Q1), action + auto-dismiss enforcement (Q2), action API shape (Q3), timer pause scope, per-toast vs. region-wide (Q7), and the message-content slot shape (Q9).
- A Decision log recording what's already settled (host `tabindex`/focus-on-open behavior, and the container/queue scope decision) separately from what's still open, so open-questions stays focused on what actually needs a call.
- A migration checklist covering setup, API (including the container/queue), styling (including the new container/queue visuals), accessibility, testing, and documentation.

See the plan for full detail and current status of each open question.

## Motivation and context

Phase 1 of the Toast 1st-gen → 2nd-gen migration. The a11y analysis needed one non-contradictory ARIA pattern (it previously recommended both `role="status"` and `role="alertdialog"`) before the migration plan could be written against it.

## Related issue(s)

- fixes SWC-2257

## Author's checklist

---

## Reviewer's checklist

### Manual review test cases

Docs only, no functional code changes. Nothing to manually verify beyond reading `accessibility-migration-analysis.md` and `migration-plan.md` for accuracy.


CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/accessibility-migration-analysis.md
```diff
@@ -46,10 +45,6 @@
 
 This doc covers how `swc-toast` should work for accessibility. It targets WCAG 2.2 Level AA. Until `swc-toast` exists under `2nd-gen/`, use `1st-gen/packages/toast/src/Toast.ts` (`<sp-toast>`) to validate behavior, and update this spec against the real 2nd-gen source when it ships.
 
-### Also read
-
-[Toast migration roadmap](./rendering-and-styling-migration-analysis.md) for layout, CSS, and DOM changes.
-
```
rise-erpelding, line 52: removed; we didn't do one for this component

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/accessibility-migration-analysis.md
```diff
@@ -71,25 +66,24 @@ Not a replacement for critical error messaging or decisions requiring mandatory
...
+| [Name, Role, Value (WCAG 4.1.2)](https://www.w3.org/TR/WCAG22/#name-role-value) | The close button and any action button must have accessible names. The toast host must expose its role (`alertdialog`) to the accessibility tree. |
 
-**Bottom line:** `swc-toast` must expose its content as a polite live region via `role="status"` on the host element, with `aria-hidden="true"` when closed. The auto-dismiss timer must pause on hover and focus. The close button and any action button must have accessible names. Color alone must not convey variant meaning.
+**Bottom line:** `swc-toast` uses `role="alertdialog"` on the host (`aria-modal="false"`) and `role="alert"` with `aria-atomic="true"` on an inner content wrapper. Opening a toast must not move focus. Name the host with `aria-label` from the default-slot text. Set `aria-hidden="true"` when closed. Pause the auto-dismiss timer on hover and focus-within, preserving remaining time. The close button and any action button must have accessible names. Color alone must not convey variant meaning.
```
rise-erpelding: The changes here and above are mainly an adjustment to remove `role="status"` from the doc. The rest of this doc mostly recommends outer role of `alertdialog` and inner role of `alert`, aligning with RSP.

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/accessibility-migration-analysis.md
```diff
@@ -109,31 +103,31 @@ Not a replacement for critical error messaging or decisions requiring mandatory
...
-| **`aria-labelledby` on host** | The `role="alertdialog"` host must have an accessible name. Align with React Spectrum: the host carries `aria-labelledby` referencing a stable ID on the content element. Because the notification text is in the default slot (light DOM), the ID is resolvable from the host without crossing a shadow boundary. When no explicit content ID is available, the component should fall back to setting `aria-label` dynamically from the resolved slot text content. |
+| **`aria-label` on host** | The `role="alertdialog"` host must have an accessible name. Set `aria-label` from the resolved default-slot text. Do not use `aria-labelledby` on the host pointing at a shadow title ID; that IDREF will not resolve. |
```
rise-erpelding: Unlike the `role` change above, this change is maybe a bit more open for debate, but I needed to pick a direction. I went with `aria-label` even though it's a break from RSP because it seemed like it would be more shadow-DOM friendly:

My understanding is that we can use `aria-labelledby` and put the associated `id` in the slotted light DOM, but that would require the consumer to wrap their message in a `<span>` in order to provide the `id`, which isn't the way Gen1 worked.

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/accessibility-migration-analysis.md
```diff
@@ -109,31 +103,31 @@ Not a replacement for critical error messaging or decisions requiring mandatory
...
-| **Close button** | Must have an accessible name ("Close"). 1st-gen uses `label="Close"` on `sp-close-button`. 2nd-gen should use `swc-close-button` with the same label. The close button's label is on the button element itself; no cross-root ARIA concern. |
-| **Action button slot** | Design allows a maximum of one action button per toast. The `action` slot is light DOM; authors must provide a descriptive label (for example, "Undo file deletion" rather than just "Undo"). Docs must state the one-action limit explicitly. |
+| **Close button** | Must have an accessible name ("Close"). 1st-gen uses `label="Close"` on `sp-close-button`. 2nd-gen should use `swc-close-button` with `accessible-label="Close"`. The name is on the close button; no cross-root ARIA concern. |
```
rise-erpelding: Updated to specify use of `accessible-label`

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/accessibility-migration-analysis.md
```diff
@@ -109,31 +103,31 @@ Not a replacement for critical error messaging or decisions requiring mandatory
...
+| **Action button slot** | Design allows a maximum of one action button per toast. The `action` slot is light DOM; authors must provide a descriptive label (for example, "Undo file deletion" rather than just "Undo"). Docs must state the one-action limit explicitly. Docs must warn against combining `timeout` with the `action` slot. Fire a dev warning if both are set. |
```
rise-erpelding: This pops up on the migration plan also as an open question (Q2), the plan was unsure whether we should hard-disable the auto-disappear timeout when there's an action button, or just warn/document against having both an action and a timeout.

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/accessibility-migration-analysis.md
```diff
@@ -109,31 +103,31 @@ Not a replacement for critical error messaging or decisions requiring mandatory
...
 
 The action slot is light DOM; the author-provided button and its label do not create cross-root ARIA concerns.
 
-The close button (`swc-close-button`) is a web component in shadow DOM; its accessible name is set via its own `label` property and does not require cross-root ARIA.
+The close button (`swc-close-button`) is a web component in shadow DOM; its accessible name is set via `accessible-label` and does not require cross-root ARIA.
 
 ### Accessibility tree expectations
 
-The structure below follows the React Spectrum Toast pattern.
+The structure below follows the React Spectrum Toast pattern. Trees show a **standalone** toast (no `tabindex` on the host). When the toast is inside a container, the host also has `tabindex="0"`.
```
rise-erpelding: This `tabindex` and the idea of a standalone toast are also open questions on the migration plan

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/accessibility-migration-analysis.md
```diff
@@ -145,53 +139,55 @@ host [aria-hidden="true"]
...
-host [role="alertdialog", aria-modal="false", tabindex="0", aria-labelledby="[content-id]"]
+host [role="alertdialog", aria-modal="false", aria-label="[slot text]"]
   └── (shadow) alert [role="alert", aria-atomic="true"]
-        └── [text content from default slot, id="[content-id]"]
+        └── [text content from default slot]
   └── (shadow) close button: "Close"
 ```
 
-Screen reader announces assertively when toast opens: "[message text]". Focus is placed on the host alertdialog element.
+Screen reader announces assertively when toast opens: "[message text]". Focus does not move.
```
rise-erpelding, line 149: This also is RSP's pattern; it's announced but focus doesn't move

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/accessibility-migration-analysis.md
```diff
@@ -109,31 +103,31 @@ Not a replacement for critical error messaging or decisions requiring mandatory
 
 | Topic | What to do |
 |-------|------------|
-| **Host role: `role="alertdialog"`** | Prescribed and fixed on the host element. `role="alertdialog"` conveys that the toast is a non-modal notification that may contain interactive controls (close button, optional action button). Set `aria-modal="false"` so assistive technologies do not restrict browsing outside the toast. Set `tabindex="0"` on the host so focus lands on the dialog element itself rather than its first focusable child. Do not allow authors to override this role. This pattern aligns with the [React Spectrum Toast](https://react-spectrum.adobe.com/react-spectrum/Toast.html) implementation. |
```
rise-erpelding, line 112: This shows up as Q4 on the migration plan: should `tabindex` on host be conditional based on whether the container is present (we wouldn't build the container right now anyway), or should it always be `0`?

I did get a bit confused with these changes, and I definitely want further opinions on the `tabindex="0"` thing, but it ties into the toast container question as well. The way I adjusted it, the `tabindex="0"` is set on the toast container if we build it, which aligns more with how React has it. If we opt for a standalone toast right now, we would not set `tabindex="0"` on the host, is that ok or does this need a revert/adjustment?

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/migration-plan.md
```diff
@@ -0,0 +1,450 @@
...
+| `@spectrum-web-components/icons-workflow` | 1.12.2 | `sp-icon-info`, `sp-icon-alert`, `sp-icon-checkmark-circle` |
+| `@spectrum-web-components/shared` | 1.12.2 | `FocusVisiblePolyfillMixin` |
+
+---
+
+## Open gen1 issues
+
+| Jira | Type | Status (snapshot) | Summary | Notes |
+| ---- | ---- | ----------------- | ------- | ----- |
+| SWC-610 | Bug | To Do | Toast timeout minimum differs from design docs | See Q1 |
```
rise-erpelding, line 169: This is the basis of Q1, whether the timeout should be 5 seconds or 6. The a11y plan opted for 6 seconds based on [this formula](https://sheribyrnehaber.medium.com/designing-toast-messages-for-accessibility-fb610ac364be) from Sheri Byrne-Haber, which feels logical to me, but happy to have folks weigh in on this.

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/migration-plan.md
```diff
@@ -0,0 +1,450 @@
...
+Icons confirmed against Figma. Source from the public `@adobe/spectrum-wc-icons` workflow-icon package (`Icon_InfoCircle()`, `Icon_CheckmarkCircle()`, `Icon_AlertTriangle()`, or the `<swc-icon-*>` elements), not the internal lean icon set in `2nd-gen/packages/swc/components/icon/elements/`. Forward Toast's resolved icon label (default or `icon-label` override) straight into the icon's own `accessible-label`; empty renders it decorative, matching the a11y doc's `icon-label=""` suppression behavior for free.
+
+Action button (when present): secondary, outline, `static-color="white"`. Confirmed by Figma playground and RSP S2 (`variant="secondary" fillStyle="outline" staticColor="white"`).
+
+#### Slots (2nd-gen)
+
+| Slot | Content | Notes |
+| ---- | ------- | ----- |
+| default | Toast message text | **Confirmed.** |
+| `action` | Optional action button | **Open question**, Q3: may become `action-label`/`swc-action` props instead |
```
rise-erpelding: This is the basis for Q3, I _think_ an action slot probably makes more sense for our system

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/migration-plan.md
```diff
@@ -0,0 +1,454 @@
...
+
+## TL;DR
+
+- Toast is a small, self-contained component: one host, one message slot, one optional action, a close button, an auto-dismiss timer.
+- Toast queue/container is currently out of scope. Toast will ship standalone where consumers compose their own container.
+- 1st-gen defines 5 variant values (`negative`, `positive`, `info`, `error`, `warning`); `error` and `warning` are already deprecated aliases of `negative` in 1st-gen and do not carry forward. 2nd-gen has 4 variants: `neutral`, `info`, `positive`, `negative`.
+- Points of disagreement, see [Design](#design) for more detail:
+    - **Timeout minimum** (6000ms vs. 5000ms)
+    - **Action+timeout** (soft warning vs. hard block)
+    - **Action API shape** (light-DOM `action` slot vs. `actionLabel`/`onAction` props
```
aramos-adobe: Missing `)` at the end 😁

aramos-adobe (approved):

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/accessibility-migration-analysis.md
5t3ph: We could also write this ID to keep use of `aria-labelledby`, we have `uniqueId()` utility available

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/migration-plan.md
```diff
@@ -0,0 +1,454 @@
...
+
+Icons confirmed against Figma. Source from the public `@adobe/spectrum-wc-icons` workflow-icon package (`Icon_InfoCircle()`, `Icon_CheckmarkCircle()`, `Icon_AlertTriangle()`, or the `<swc-icon-*>` elements), not the internal lean icon set in `2nd-gen/packages/swc/components/icon/elements/`. Forward Toast's resolved icon label (default or `icon-label` override) straight into the icon's own `accessible-label`; empty renders it decorative, matching the a11y doc's `icon-label=""` suppression behavior for free.
+
+Action button (when present): secondary, outline, `static-color="white"`. Confirmed by Figma playground and RSP S2 (`variant="secondary" fillStyle="outline" staticColor="white"`).
+
+#### Slots (2nd-gen)
+
+| Slot | Content | Notes |
+| ---- | ------- | ----- |
+| default | Toast message text | **Confirmed.** |
```
5t3ph: I'm on the fence of whether this should be default or use a named slot since this isn't likely to be as simple of a label as Button. A named slot would let use apply a dynamic `id` for labeling, per my reply in another comment.

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/migration-plan.md
```diff
@@ -0,0 +1,454 @@
...
+| `action` | Optional action button | **Open question**, Q3: may become `action-label`/`swc-action` props instead |
+
+#### CSS custom properties (2nd-gen)
+
+No `--mod-*` properties will be exposed. New `--swc-*` component-level properties may be introduced where needed — these are additive and not breaking. See [Component Custom Property Exposure](../../../../CONTRIBUTOR-DOCS/02_style-guide/01_css/02_custom-properties.md#component-custom-property-exposure) for what to expose and how.
+
+**Inferred initial set**, based on the [Component Custom Property Exposure decision tree](../../../../CONTRIBUTOR-DOCS/02_style-guide/01_css/02_custom-properties.md#decision-tree-for-exposure) applied to Toast's own 1st-gen `--mod-toast-*` surface (see [CSS custom properties](#css-custom-properties) above) and precedent from already-migrated components (`swc-badge`):
+
+- `--swc-toast-height`: min block size, matches `--swc-badge-height` precedent
```
5t3ph: Design is trying to move away from this and let it be composed of padding + font-size + line-height, which is likely to be sufficient for this use case too.

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/migration-plan.md
```diff
@@ -0,0 +1,454 @@
...
+| Slot | Content | Notes |
+| ---- | ------- | ----- |
+| default | Toast message text | **Confirmed.** |
+| `action` | Optional action button | **Open question**, Q3: may become `action-label`/`swc-action` props instead |
+
+#### CSS custom properties (2nd-gen)
+
+No `--mod-*` properties will be exposed. New `--swc-*` component-level properties may be introduced where needed — these are additive and not breaking. See [Component Custom Property Exposure](../../../../CONTRIBUTOR-DOCS/02_style-guide/01_css/02_custom-properties.md#component-custom-property-exposure) for what to expose and how.
+
+**Inferred initial set**, based on the [Component Custom Property Exposure decision tree](../../../../CONTRIBUTOR-DOCS/02_style-guide/01_css/02_custom-properties.md#decision-tree-for-exposure) applied to Toast's own 1st-gen `--mod-toast-*` surface (see [CSS custom properties](#css-custom-properties) above) and precedent from already-migrated components (`swc-badge`):
```
5t3ph: [nit] These don't need decided ahead of time, and we will continue to ultimately only expose what may be needed for override internally first.

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/accessibility-migration-analysis.md
```diff
@@ -201,25 +197,22 @@ The React Spectrum container structure for reference:
...
 
-**Actionable toasts and timing:** When an action button is present, auto-dismiss should generally be disabled or set to a much longer timeout. A screen reader user who hears the alert announcement and moves focus to the toast (via Tab or the live region) needs sufficient time to activate the action. The docs must warn against combining `timeout` with the `action` slot.
+**Actionable toasts and timing:** When an action button is present, auto-dismiss should generally be disabled or set to a much longer timeout. A screen reader user who hears the alert and then tabs to the action needs enough time to activate it. The docs must warn against combining `timeout` with the `action` slot. Fire a dev warning if both are set.
 
 ### Keyboard and focus
 
-Focus placement: When a toast opens, focus is placed on the `role="alertdialog"` host element itself (`tabindex="0"`), not on the first focusable child. This matches the React Spectrum behavior confirmed by the RS team (Michael Jordan): "We actually want to focus the alert dialog element itself, rather than the first focusable element."
```
5t3ph, line 208: I don't want to lose this context of what's been deleted from this section, including the bullets below, so maybe instead shift it into a subsection for toast container behavior.

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/migration-plan.md
5t3ph: Yes, I think so for shadow DOM reasons of consumers attaching behavior to it. We could propagate size/variant so it is styled in the expected way.

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/migration-plan.md
```diff
@@ -0,0 +1,454 @@
...
+- `--swc-toast-gap`: spacing between icon and message text
+
+**Excluded from the initial set:**
+
+- Background color, text color, icon color, per the style guide's exclusions. Toast's text/icon color is fixed white against a saturated variant background (the same contrast-intent exclusion as static-color components), and background color is set per semantic variant, not consumer-overridable per the badge/status-light precedent for semantic (non-decorative) variant colors.
+- `max-inline-size`. A single constant value regardless of variant, same as `swc-tooltip`/`swc-popover` (also not exposed). The host is directly stylable from outside (`swc-toast { max-inline-size: 400px; }`), no custom property needed.
+
+### Behavioral semantics
+
+- Auto-dismiss timer pauses on `pointerenter` + `focusin`, preserving remaining time; resumes only once both `pointerleave` and `focusout` have fired.
```
5t3ph: If we ship Toast without a container, we're stuck either making every consumer manually re-set the same timeout policy everywhere, or trusting a dev warning to catch the action+timeout combo that RSP just enforces automatically at the queue level. Plus if toasts ever queue behind a max-visible cap, only the container knows what's actually on screen, so it's the only thing that can start the clock at the right time. Basically: timeout looks like a solved, standalone feature, but doing it right still routes through the container. If we're deferring the container, we should defer locking in real timeout behavior too, not just ship it standalone and hope it holds up later.

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/migration-plan.md
```diff
@@ -0,0 +1,454 @@
...
+- `role="alertdialog"` + `aria-modal="false"` on host; opening never moves focus.
+- `tabindex` on host follows the a11y doc: no `tabindex` on a standalone host; `tabindex="0"` once the toast is inside a container. See Q4.
+- Close fires a cancelable `close` event before the component closes itself (matches 1st-gen).
```
5t3ph: We might expand events to the full set similar to our other visibility toggling components:

- `swc-open`: Dispatched when the toast begins to open, before the enter transition plays.
- `swc-after-open`: Dispatched after the enter transition completes.
- `swc-close`: Dispatched when the toast begins to close, before the exit transition plays.
- `swc-after-close`: Dispatched after the exit transition completes.

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/migration-plan.md
```diff
@@ -0,0 +1,454 @@
...
+
+| # | Item | Blocking? | Status | Owner |
+| --- | ---- | --------- | ------ | ----- |
+| Q5 | Should the pause-preserving countdown be extracted to a shared core controller now, or stay inline in `Toast.base.ts` until a second consumer needs it? | No | Open | Architecture reviewer |
+
+### Scope and prerequisites
+
+| # | Item | Blocking? | Status | Owner |
+| --- | ---- | --------- | ------ | ----- |
+| Q6 | Should `swc-toast` later ship alongside a first-party `swc-toast-queue` container, or should authors keep composing their own? Explicitly out of scope for this migration; the a11y doc raises it as a question to bring to a team sync before a container API is added. If built, `placement` (RSP precedent: `top`/`bottom`/`top end`/`bottom end`) belongs on the container, not on `swc-toast` itself. | No | Open (deferred) | Design + accessibility reviewer |
```
5t3ph: As shared in Slack, I strongly suggest we ship the container simultaneously, but we'll discuss more!

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/migration-plan.md
rise-erpelding: Aw what excellent proof that I actually hand-edited some of this! 😆

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/migration-plan.md
rubencarvalho: This is the larger point I would also make in this migration. I think we should decide how opinionated this should be out of the box. If we have all of these constraints, I would be for offering the whole container, timeout, placement, eventually queuing (we know stacked toasts will come too) from the start...
I have the feeling anything else would make it harder for consumers (and for us to document to enforce Design). Happy to discuss this! 😄

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/accessibility-migration-analysis.md
rise-erpelding: Let me know if you think this needs more!

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/migration-plan.md
```diff
@@ -0,0 +1,537 @@
...
+| --- | ------------ | ---------------- | ---------------- | ----------------------- |
+| S1 | Adopt S2 tokens | S1 tokens | S2 tokens from `spectrum-css` `spectrum-two` | Visual update only |
+| S2 | Peek-stack depth styling | None (no queue exists) | Stacked-card depth behind the front toast: positional offset and opacity falloff, `opacity: 0` beyond the third position, matching RSP S2 | None (new functionality) |
+| S3 | Expand/collapse layout | None (no queue exists) | Full-screen expanded-view layout with a dismissible scrim and a collapsed/expanded transition; front-toast expand control laid out via `gridTemplateAreas` alongside the action button | None (new functionality) |
+
+#### Accessibility and behavior
+
+| # | What changes | 1st-gen behavior | 2nd-gen behavior | Consumer migration path |
+| --- | ------------ | ---------------- | ---------------- | ----------------------- |
+| A1 | Host role | `role="alert"` on inner `.body` div only | `role="alertdialog"` + `aria-modal="false"` on host; inner `role="alert"` `aria-atomic="true"` | None |
```
cdransf, line 226: Do this and `:244` need to differ or can `A1` be used in separate contexts? ✨

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/migration-plan.md
```diff
@@ -0,0 +1,537 @@
...
+### Slots
+
+| Slot | Content | Notes |
+| ---- | ------- | ----- |
+| default | Toast message text | |
+| `action` | Optional action button (light DOM) | Max one, per design |
+
+### CSS custom properties
+
+No dedicated rendering-and-styling analysis doc exists for Toast (out of scope this cycle); this is pulled directly from `1st-gen/packages/toast/src/toast.css.js`.
```
cdransf: I don't see this in the file tree, do we need to drop the `.js`? ✨

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/migration-plan.md
```diff
@@ -0,0 +1,537 @@
...
+</sp-toast>
+```
+
+---
+
+## Dependencies
+
+| Package | Version | Role |
+| ------- | ------- | ---- |
+| `@spectrum-web-components/base` | 1.12.2 | Lit base class |
```
cdransf: I _think_ some of these may now be `1.12.3`. ✨

cdransf (approved):

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/migration-plan.md
rise-erpelding: This one stays; the other one changed to P1

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/migration-plan.md
rise-erpelding: Closing this thread out to neaten things up, but Q3 does reflect that we're leaning toward the slot

PR #6688, opened 2026-09-01
fix(response-status): align with Figma for hover/expanded state, complete label, and Storybook leak

rubencarvalho:
## Summary
Fixes several `swc-response-status` issues found during design QA:
- Header darkens on hover and while expanded
- Completed label reads "Response complete"
- Checkmark no longer shows blue on mobile
- Header row is fully clickable, not just the label
- Chevron rotates smoothly again
- `Status only` story no longer leaks into the sidebar/docs/prod build
- Pattern overview demo steps roll at a more readable pace

## Test plan
- [x] Unit tests passing
- [x] Lint clean
- [x] Manually verified in Storybook

### DIFF
--- a/.changeset/response-status-figma-parity.md
+++ b/.changeset/response-status-figma-parity.md
@@ -1,0 +1,10 @@
+---
+'@adobe/spectrum-wc': patch
+---
+
+Fix several `swc-response-status` visual/behavior mismatches against the Figma spec:
+
+- The header label no longer stays in its lighter, unfocused color while the step timeline is expanded — it now darkens to match the hover treatment.
+- The completed-status default label reads "Response complete" instead of "Response generated".
+- The header toggle button no longer relies on the browser's native `<button>` text color, which some mobile browsers rendered as accent blue for the completed checkmark icon.
+- The `Status only` Storybook story no longer leaks into the sidebar/docs/prod build; it was incorrectly tagged `dev`, a tag reserved for the Playground story.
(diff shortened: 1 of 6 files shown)


.changeset/response-status-figma-parity.md
```diff
@@ -0,0 +1,12 @@
...
+'@adobe/spectrum-wc': patch
+---
+
+Fix several `swc-response-status` visual/behavior mismatches against the Figma spec:
+
+- The header label no longer stays in its lighter, unfocused color while the step timeline is expanded — it now darkens to match the hover treatment.
+- The completed-status default label reads "Response complete" instead of "Response generated".
+- The header toggle button no longer relies on the browser's native `<button>` text color, which some mobile browsers rendered as accent blue for the completed checkmark icon.
+- The `Status only` Storybook story no longer leaks into the sidebar/docs/prod build; it was incorrectly tagged `dev`, a tag reserved for the Playground story.
+- The header toggle's clickable area now spans the full row, while the visible label and chevron stay grouped together instead of stretching apart.
```
5t3ph, line 11: I'm not sure it _should_ span the full row since there's no indication that is clickable so a stray click could toggle it unexpectedly.

The mismatch with focus outline vs hover background emphasizes this "mystery" blank space, since this is true for subitems too:

Image

If we were still changing cursor for buttons then that would alleviate this since the cursor change would indicate "you are now over a clickable item" when drifting over the blank space.

aramos-adobe (approved):

5t3ph (approved):
Approving since fixes you intended look good.

Something to address later - this is toggleable but doesn't have the icon or other indication, something to probably take back to design since if nothing else its a cognitive a11y error

Image

miwha-adobe (approved):

.changeset/response-status-figma-parity.md
rubencarvalho: Yes - I agree. This was an initial ask to get us up to parity with React but I think we’ll definitely need to iterate here… we’re following up on this.

PR #6698, opened 2026-09-01
docs(rfcs): introduce an RFC process adapted from React Spectrum

caseyisonit:
## Description

Introduces an **RFC process** for Spectrum Web Components, adapted from [React Spectrum's RFC process](https://github.com/adobe/react-spectrum/blob/main/rfcs/README.md).

The goal: give substantial changes (new components, breaking API changes, cross-cutting architecture, migration policy) a lightweight, PR-based design process that produces a durable record of *why* we decided what we decided — the kind of context that currently lives in Slack threads and PR comments and evaporates.

### What's included

- **`rfcs/README.md`** — the process. When an RFC is required (narrowed to components / public API / `--swc-*` tokens / migration policy), the flow, and explicit decision authority: **2 maintainer approvals or a shepherd's call after a 1-week minimum comment period** (the one place React Spectrum's informality doesn't transfer to us).
- **`rfcs/template.md`** — React Spectrum's template plus two SWC-specific sections (**Design/Figma Alignment**, **Token & Styling-API Impact**) and a living `Status:` header.
- **`rfcs/2026-swc-styling-api-prefix.md`** — a back-filled worked example recording the 2nd-gen `--swc-*` public styling-API decision, to seed the format with a real example.
- **`rfcs/process-analysis.md`** — the analysis of React Spectrum's model, the SWC adoption plan, and risks/mitigations.

### Notes for reviewers

- This is **docs-only** and starts a conversation — the process itself (trigger scope, decision authority, comment period) is the thing to review and adjust.
- The three adaptations beyond a straight copy of RS: (1) a narrow "when required" trigger, (2) explicit decision authority + comment period, (3) `Status:` + required post-merge updates so RFCs don't go stale.

## How and why we got here

We looked at how React Spectrum runs RFCs (deliberately low-volume — ~16 in 7 years, no voting/SLA/labels, consensus-based) and adapted it to SWC's context. See `rfcs/process-analysis.md` for the full comparison and rationale.

## Related issue

_None — process proposal._


Rajdeepc:
Why more processes?

rfcs/README.md
```diff
@@ -0,0 +1,54 @@
+# Spectrum Web Components RFCs
+
+Most changes — bug fixes, refactors, docs, single-component work that follows an agreed playbook — can go through the normal pull-request flow. Some changes are **substantial**, and we ask that these be put through a design process to build consensus and produce a durable record of *why* we decided what we decided.
+
+This process is adapted from [React Spectrum's RFC process](https://github.com/adobe/react-spectrum/blob/main/rfcs/README.md). It is deliberately lightweight: an RFC is a markdown file added to this directory via a pull request.
+
+## When you need an RFC
```
jnjosh, line 7: This changes the process mildly and I want to discuss further. The assumption I'm reading here is that we want people to just know when an RFC is needed or not. Which in my experience unravels. We've treated it as a planning step for that reason. You shouldn't know when you need an RFC because you'll have a ticket to create that RFC. I think the spirit here is good, but we shouldn't take on that scope in this context.

rfcs/README.md
```diff
@@ -0,0 +1,54 @@
...
+6. Incorporate feedback from the community and maintainers. Expect the RFC to evolve; revising the proposal mid-discussion is normal and healthy.
+7. A decision is made after the review period (see below).
+
+## Review, decision, and comment period
+
+- RFCs are discussed publicly on the pull request.
+- An RFC stays open for a **minimum of one week** to give people time to comment. There is no maximum; substantial proposals may take longer.
+- **Decision authority:** an RFC is **Accepted** when at least **two maintainers approve** and no maintainer has an unresolved blocking objection. If consensus cannot be reached, the **designated RFC shepherd** (or, absent one, the lead maintainer) makes the final call after the comment period. We seek consensus, not unanimity.
+- **Accepted** RFCs are merged into `rfcs/` with `Status: Accepted`.
+- **Declined** RFCs have their pull request closed without merging.
```
jnjosh, line 41: This is one reason I don't like PRs as RFCs, it's not always Accepted / Declined. Sometimes you want to have the decision not to do something available without hunting for it. Declined should be merged in this context because we decided not to do a decision.

rfcs/README.md
```diff
@@ -0,0 +1,54 @@
...
+You do **not** need an RFC for:
+
+- Bug fixes and internal refactors with no public-API impact.
+- Documentation.
+- Performance improvements invisible to consumers.
+- A single-component migration that follows the already-agreed migration playbook.
+
+If you're unsure, open an issue or ask a maintainer before investing in a full RFC.
+
+## The process
```
jnjosh, line 25: Wherever we go this process needs to align with our existing process or improve that process. cc @najikahalsema

rfcs/README.md
```diff
@@ -0,0 +1,54 @@
+# Spectrum Web Components RFCs
```
jnjosh, line 1: I think generally I don't have a huge thought about using Github as the mechanism for RFCs, I don't prefer it for many reasons. But minimally I don't want this process to be defined here. We have an existing process and I'd like to make sure we are following that—and if it needs improved we should improve it.

jnjosh (commented):
I don't mind using Github for RFCs as a storage mechanism, it's definitely better than Wiki 🫠. I hoped the internal RFC hub could be the place and would like to understand why improving that isn't an option. 

Lastly, I'd like to not recreate process in this repo that we already have defined. The RFC guide has process for how/when and relative workflows. As for hosting the files, RFC guide specifies wiki, but I think it's acceptable improvement to allow teams to host elsewhere like Github.

jnjosh:
> Why more processes?

@Rajdeepc I left some comments too, but maybe we don't consider this a new process but rather a new storage location. The existing RFC process isn't changing, just where the files are reviewed would be with this.

caseyisonit:
@Rajdeepc @jnjosh This was a request from @najikahalsema and @pfulton as an action item from a team retro on "How might we RFC easier." 

This is just an evaluation, an explanation, and a simple example of how React Spectrum handles its RFCs. This is not flagged as "Status: Ready for review" for that reason. I'll convert this to a draft PR, but this is research only. 

Regarding the hub question, it's been too challenging to collaborate and wait for features at the velocity at which we are moving. 

The pains of RFCs and keeping them as first-class citizens in our review workflow have been a struggle, and things are being forgotten or moving on without acceptance in the current process. This is a piece of the conversation, not the final proposed solution. The benefit of using GitHub is that we can integrate it into our review process and Slack alert system. Also, it keeps records of decisions close to where those decisions are implemented and can be used with our agents/skills in context more easily. We can easily modify the structure to keep rejected RFCs in a separate directory, but another added value of GitHub is that open and closed PRs are searchable. There are options to this approach, and we are just looking at how other teams are handling this successfully.

PR #6699, opened 2026-09-01
feat(text-field): add FieldAssociationController for native form participation

miwha-adobe:
## Description

Unlike a native `<input>`, a custom element like `swc-text-field` doesn't join a `<form>` on its own: its value won't submit, reset, or disable inside a `<fieldset disabled>`. The browser grants that behavior only through the `ElementInternals` (form-associated custom element) API, which is fiddly, easy-to-get-wrong wiring.

This PR adds a small, shared **`FieldAssociationController`** that owns that wiring once, so any form field can plug into it the same way. Concretely, it:

- puts the field's value into the form's `FormData`, and pulls it back out (via `null`) when the field is disabled or has nothing to submit;
- restores the field's starting value when the form is reset;
- reacts to a surrounding `<fieldset disabled>` or disabled form, so the field disables and drops out of submission;
- exposes the **read** side of validity (`validity`, `checkValidity()`, …) so a field can report its state.

`swc-text-field` is the first component to adopt it, which is what proves the controller end-to-end. It also ships with three small, unstyled demo elements (text, radio, and combobox stand-ins) that double as the controller's living documentation, plus Storybook play tests and a docs page.

## Motivation and context

The 2nd-gen field family (text field, checkbox, radio, combobox, and more) all need the *same* form-participation wiring. Without a shared piece, each component re-implements the same `ElementInternals` boilerplate and re-tests it, and mistakes in that wiring are easy to make and hard to notice.

Doing it once, in a controller proven on the simplest field first (`swc-text-field`), keeps later field migrations small: a new field brings only its own inner control and a rule for how that control's state becomes a submitted value.  

## Related issue(s)

- SWC-2467

No visual changes: the controller is non-visual and the `swc-text-field` render remains a stub. Storybook demo harnesses are intentionally unstyled.

---

## Author's checklist

- [ ] I have read the **[CONTRIBUTING](https://github.com/adobe/spectrum-web-components/blob/main/CONTRIBUTING.md)** and **[PULL_REQUESTS](https://github.com/adobe/spectrum-web-components/blob/main/PULL_REQUESTS.md)** documents.
- [x] I have reviewed the Accessibility Practices for this feature. (Controller sets no ARIA; labelling/announcements are a separate ticket, SWC-2466.)
- [x] I have added automated tests to cover my changes. (5 controller play tests plus a `swc-text-field` form-participation smoke test.)
- [x] I have included updated documentation if my change required it. (Controller docs page plus migration-plan and workflow-doc updates.)

---

## Reviewer's checklist

### Manual review test cases

- [ ] **Form value, exclusion, and disabled cascade**
  1. Storybook → **Controllers → Field association controller → Interactive demo**.
  2. Type in **Username**, click **Submit** → readout's `FormData` shows `username=<typed>`.
  3. Click **Toggle fieldset disabled** → all fields drop out of `FormData` and their inner controls disable; toggle again → they return.

- [ ] **Reset restores the default**
  1. Same demo; change **Username**, then click **Reset** → value restores to its initial `Example`; `FormData` reflects the restored value.

- [ ] **Validity pass-throughs**
  1. Same demo; clear the required **Username**, click **Check validity** → readout reports the field invalid with a message; `willValidate=false` for a disabled field.

- [ ] **`swc-text-field` participates in a form**
  1. Storybook → **Text field → Tests → Form participation** (runs automatically) → value included, excluded when disabled, restored on reset.

### Device review

No component-level visual output in this PR; demos are unstyled harnesses.

- [ ] Did it pass in Desktop?

- [ ] **Keyboard**: *Field association controller → Interactive demo.*
  1. <kbd>Tab</kbd> through the form → focus reaches the text input, select, and radio in order.
  2. Edit/toggle each with the keyboard → values update in the readout on submit.
  3. Expect: native focus order and operation; the controller adds no custom keyboard behavior.

- [ ] **Screen reader**: *Same demo.*
  1. Navigate the form fields with a screen reader.
  2. Expect: each inner control announces its `aria-label` and native role; form participation (submit/reset/disabled) behaves as native. Visible labelling and validation announcements are out of scope here (deferred to `LabellingController`, SWC-2466).


2nd-gen/packages/core/components/text-field/TextField.base.ts
```diff
@@ -183,16 +183,74 @@ export abstract class TextFieldBase extends SizedMixin(SpectrumElement, {
...
+    return this.fieldAssoc.form;
+  }
+
+  /** The field's constraint-validation state. */
+  public get validity(): ValidityState {
+    return this.fieldAssoc.validity;
+  }
+
+  /** The localized validation message. */
+  public get validationMessage(): string {
```
miwha-adobe, line 217: I looked into this a little further, and it seems like this validationMessage's purpose, is to suport the native fields `validationerror` behavior, which has a tooltip pop up that declares the message. Is this a behavior we want to suppress?

2nd-gen/packages/core/components/text-field/TextField.base.ts
```diff
@@ -183,16 +183,74 @@ export abstract class TextFieldBase extends SizedMixin(SpectrumElement, {
...
+  }
+
+  /** Runs constraint validation; returns whether the field is valid. */
+  public checkValidity(): boolean {
+    return this.fieldAssoc.checkValidity();
+  }
+
+  /** Runs constraint validation and reports any problem to the user. */
+  public reportValidity(): boolean {
+    return this.fieldAssoc.reportValidity();
```
miwha-adobe, line 233: Additionally, since we're rendering our own error UI via the error-text slot + aria-errormessage, the browser's native bubble (shown on reportValidity() and on blocked form submits, using validationMessage) would show up alongside it

We can suppress it per-field without losing submit-blocking by canceling the invalid event on the host:
```
this.addEventListener('invalid', (event) => event.preventDefault());
```

2nd-gen/packages/core/components/text-field/TextField.base.ts
```diff
@@ -183,16 +183,74 @@ export abstract class TextFieldBase extends SizedMixin(SpectrumElement, {
...
+    return this.disabled || this.fieldAssoc.formDisabled;
+  }
+
+  /** The form the field participates in, or `null`. */
+  public get form(): HTMLFormElement | null {
+    return this.fieldAssoc.form;
+  }
+
+  /** The field's constraint-validation state. */
+  public get validity(): ValidityState {
```
miwha-adobe, line 212: These validity members are plumbed but inert right now. They read straight off ElementInternals, which returns the default valid/empty state until something calls `internals.setValidity(...)`, which should happen when we wire up the render.

2nd-gen/packages/core/controllers/field-association-controller/src/field-association-controller.ts
```diff
@@ -0,0 +1,138 @@
...
+ * `addController`. The host keeps `static formAssociated = true` and its own
+ * `attachInternals()` (neither can be delegated), passes that result in here, and
+ * delegates its `formResetCallback` / `formDisabledCallback`. See the docs page
+ * for a full host example.
+ */
+export class FieldAssociationController {
+  private readonly _internals: ElementInternals;
+  private readonly _onDisabledChange?: (disabled: boolean) => void;
+
+  private _defaultValue: string;
```
miwha-adobe: `FieldFormValue` can be `string | File | FormData | nul`l, but I've set `defaultValue` to string for now until there are other use cases.

2nd-gen/packages/core/components/text-field/TextField.base.ts
miwha-adobe: Video example:

https://github.com/user-attachments/assets/554799e2-e71f-4d84-9900-5df4ee52a0d1

2nd-gen/packages/core/components/text-field/TextField.base.ts
```diff
@@ -183,16 +183,74 @@ export abstract class TextFieldBase extends SizedMixin(SpectrumElement, {
   //     IMPLEMENTATION
   // ──────────────────────
 
-  // @todo (SWC-2466): resolve the accessible-labelledby / accessible-describedby
-  // IDREF stubs to cross-root element references, and add the "unlabeled field"
-  // dev-warning, via the LabellingController.
+  // @todo (SWC-2466): the LabellingController resolves the labelledby/describedby
```
miwha-adobe, line 186: When the `LabellingController` lands here, the rule that keeps the two controllers decoupled: it should read host props only `(accessibleLabel/accessibleLabelledby/accessibleDescribedby, invalid, required, labelPosition, disabled)`, never `fieldAssoc` directly.

The important one is invalid: the label side sets aria-invalid from `host.invalid`, not from `fieldAssoc.validity`. Validity flows `fieldAssoc.validity` → host reflects into `host.invalid` → labelling reads `host.invalid`, so the host stays the only thing touching both controllers. If labelling ever reaches into `fieldAssoc`, that's the coupling we're avoiding.

2nd-gen/packages/core/components/text-field/TextField.base.ts
5t3ph: Yes, we want to suppress it - I had looked into this during the a11y planning for text-field, [here is the relevant line](https://github.com/adobe/spectrum-web-components/blob/main/CONTRIBUTOR-DOCS/03_project-planning/03_components/text-field/accessibility-migration-analysis.md?plain=1#L142):

> **Native validation bubble (deferred decision)** - Reflecting `required` (and other constraints) onto the native `<input>` engages the browser's own [constraint validation](https://developer.mozilla.org/en-US/docs/Web/HTML/Guides/Constraint_validation) on submit. That is desirable for `:invalid` styling and `internals.validity`, but it also surfaces the browser's native error bubble, which visually competes with Spectrum's inline error message. Suppress the native bubble in favor of the inline message rather than showing both. The mechanism depends on whether the migration ultimately provides a form wrapper: setting [`novalidate`](https://developer.mozilla.org/en-US/docs/Learn_web_development/Extensions/Forms/Form_validation#a_more_detailed_example) on the enclosing `<form>` suppresses the bubble while leaving the constraint API intact (inline messaging still reads `internals.validity`), and [`formnovalidate`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/button#formnovalidate) on a submit button does the same per-submit and may be a simpler first stage. Do not reach for `aria-required` in place of native `required` to dodge the bubble: `aria-required` announces the state to AT but creates no native constraint, so it would forfeit `:required`/`:invalid` matching and `internals.validity` participation and force a JS re-implementation of the required check. This decision is deferred to the forms-strategy work; capture it here so it is not lost.

For now, we can demonstrate any of our controls within a plain html form element that uses the `novalidate` attribute and be sure to document that is a requirement. When we extend `submit` behaivor for Button, we can consider adding `formnovalidate` for that context which puts us a bit more in control.

2nd-gen/packages/core/components/text-field/TextField.base.ts
5t3ph: I responded to previous comment before seeing this - does this for sure not block any of the other features we need from `reportValidity`?

2nd-gen/packages/core/controllers/field-association-controller/stories/demo-hosts.ts
```diff
@@ -0,0 +1,467 @@
...
+ * Shared base for the Storybook-only form-field harnesses: intentionally
+ * unstyled elements that exercise {@link FieldAssociationController} in isolation,
+ * so a regression triages as a controller bug rather than inside a full `swc-*`
+ * component. They double as the controller's living documentation.
+ *
+ * It carries the host boilerplate the controller can't own (`formAssociated`,
+ * `attachInternals()`, the form callbacks). Each subclass supplies its inner
+ * control and its `computeFormValue` mapping (`null` excludes it from `FormData`).
+ */
+export abstract class DemoFieldHostBase extends LitElement {
```
5t3ph, line 51: Just flagging that this is basically a prototype of what a mixin would look like to hold this boilerplate that all form field components will need. We could wait until radio goes in just to validate it in one more context, but would you like to make a ticket for that mixin?

2nd-gen/packages/core/controllers/field-association-controller/stories/demo-hosts.ts
```diff
@@ -0,0 +1,467 @@
...
+    return this.checked && !this.effectiveDisabled ? this.value || 'on' : null;
+  }
+
+  private handleChange(event: Event): void {
+    this.checked = (event.target as HTMLInputElement).checked;
+  }
+
+  protected override render(): TemplateResult {
+    return html`
+      <input
```
5t3ph: Bit of a nit since this isn't the main point but can you add a visible label? It snagged me a bit in reviewing the controller docs :)

2nd-gen/packages/core/controllers/field-association-controller/stories/demo-hosts.ts
```diff
@@ -0,0 +1,467 @@
...
+      requestAnimationFrame(() => this.refresh());
+    }
+  }
+
+  // Toggles the field's own `disabled` (the other half of effectiveDisabled).
+  private handleToggleFieldDisabled(): void {
+    const field =
+      this.renderRoot.querySelector<DemoFieldText>('demo-field-text');
+    if (field) {
+      field.disabled = !field.disabled;
```
5t3ph: I noticed that once this is toggled on the `disabled` attribute stays active on the input.

2nd-gen/packages/core/controllers/field-association-controller/stories/field-association-controller.stories.ts
```diff
@@ -0,0 +1,220 @@
...
+      ></demo-field-text>
+    </form>
+  `,
+  tags: ['behaviors'],
+};
+DefaultValueReset.storyName = 'Default value and reset';
+
+export const DisabledCascade: Story = {
+  render: () => html`
+    <form>
```
5t3ph: Missing attribute:

```suggestion
    <form disabled>
```

2nd-gen/packages/core/controllers/field-association-controller/field-association-controller.mdx
```diff
@@ -0,0 +1,157 @@
+import { Canvas, Meta } from '@storybook/addon-docs/blocks';
+import { DocsFooter, DocsHeader } from '../../../swc/.storybook/blocks';
+
+import * as Stories from './stories/field-association-controller.stories';
+
+<Meta of={Stories} />
+
+<DocsHeader />
+
```
5t3ph, line 9: The interactive demo is really nice since it shows output, but the other demos aren't providing much at the moment. Maybe setup similar submit + output? Otherwise they are just field UI and any relevant code is hidden behind the demo components.

5t3ph (commented):
Overall, really sound implementation!

Mostly some comments on the controller docs/demos, and potentially one bug but might just be local to the demo.

2nd-gen/packages/core/controllers/field-association-controller/stories/demo-hosts.ts
```diff
@@ -0,0 +1,467 @@
...
+            name="username"
+            value="Example"
+            required
+            label="Username (required)"
+          ></demo-field-text>
+          <demo-field-combobox
+            name="color"
+            label="Favorite color"
+          ></demo-field-combobox>
+          <demo-field-radio
```
aramos-adobe, line 547: Can we add a second radio option here with the same name? Don't we need at least two options to check for validity in radio?

2nd-gen/packages/core/controllers/field-association-controller/stories/field-association-controller.stories.ts
```diff
@@ -0,0 +1,220 @@
...
+  `,
+  tags: ['behaviors'],
+};
+Validity.storyName = 'Validity pass-throughs';
+
+// ────────────────────────────────
+//    ACCESSIBILITY STORIES
+// ────────────────────────────────
+
+export const Accessibility: Story = {
```
aramos-adobe:
```js
export const Accessibility: Story = { 
...FormValue, tags: ['a11y'] };
Accessibility.storyName = 'Accessibility';
```

Using the same as FormValue

2nd-gen/packages/core/controllers/field-association-controller/stories/demo-hosts.ts
miwha-adobe: I sure can. I thought the same that when radio goes in, it will have more context and better build out the mixin.

2nd-gen/packages/core/controllers/field-association-controller/stories/demo-hosts.ts
```diff
@@ -0,0 +1,467 @@
...
+        <div class="actions">
+          <button type="submit">Submit</button>
+          <button type="reset">Reset</button>
+          <button type="button" @click=${this.handleCheckValidity}>
+            Check validity
+          </button>
+          <button type="button" @click=${this.handleToggleDisabled}>
+            Toggle fieldset disabled
+          </button>
+          <button type="button" @click=${this.handleToggleFieldDisabled}>
```
aramos-adobe: Can we get rid of this and maybe just use the `Disable fieldset` button. The username doesn't toggle back to enabled state after disabling. The fieldset one does

2nd-gen/packages/core/controllers/field-association-controller/stories/demo-hosts.ts
aramos-adobe: You probably have to update the handleChange in the Demo Radio class if you do this?

2nd-gen/packages/core/controllers/field-association-controller/stories/demo-hosts.ts
miwha-adobe: Good catch! Removed the "Toggle Username disabled" button entirely rather than patching it. It wasn't adding much value and we have other instances capturing the behavior.

2nd-gen/packages/core/controllers/field-association-controller/stories/demo-hosts.ts
miwha-adobe: Definitely can do!

2nd-gen/packages/core/controllers/field-association-controller/stories/field-association-controller.stories.ts
miwha-adobe: Done! Reusing FormValue as suggested. Skipped the explicit storyName since the export name already renders as "Accessibility".

2nd-gen/packages/core/controllers/field-association-controller/stories/demo-hosts.ts
miwha-adobe: Added a second subscribe radio ("Do not subscribe"). And you were right about handleChange!  Since each radio is its own element rather than a native group, I added logic to uncheck same-name siblings on selection so exclusivity works. Thanks!

2nd-gen/packages/core/components/text-field/TextField.base.ts
miwha-adobe: As far as I can see, canceling the invalid event only suppresses its default behavior,  the browser's native error bubble ([MDN reportValidity](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/reportValidity): the check runs the same as checkValidity(), and the browser only displays the problem if the invalid event isn't canceled). Nothing else is affected

miwha-adobe:
Thank you for the review feedbck! I addressed all the comments. Additionally, I improved the demo page so the examples behave like a real form: the output now updates on submit rather than on every change, so it reflects what the form actually submits instead of mirroring each keystroke.

2nd-gen/packages/core/controllers/field-association-controller/stories/demo-hosts.ts
```diff
@@ -329,102 +470,70 @@ export class DemoFieldBench extends LitElement {
...
-
-  // Toggles the field's own `disabled` (the other half of effectiveDisabled).
-  private handleToggleFieldDisabled(): void {
-    const field =
-      this.renderRoot.querySelector<DemoFieldText>('demo-field-text');
-    if (field) {
-      field.disabled = !field.disabled;
-      requestAnimationFrame(() => this.refresh());
+      // The disabled cascade runs through async callbacks; read after it settles.
+      requestAnimationFrame(this.refresh);
```
5t3ph, line 528: Is the reason this is needed worth documenting consumer-side for purposes of checking "true" validity or anything like that?

5t3ph (approved):
Looks good!

One question for possible documentation.

aramos-adobe (approved):

2nd-gen/packages/swc/components/text-field/TextField.ts
```diff
@@ -36,14 +36,15 @@ export class TextField extends TextFieldBase {
...
   protected override render(): TemplateResult {
-    // @todo (SWC-2466 / Phase 4–5): render the visible label, required indicator,
-    // validation icon, and description/error container via the LabellingController.
-    // Until then the input takes its accessible name from `accessible-label`.
+    // @todo (SWC-2466): render the visible label, required indicator, validation
+    // icon, and description/error container via the LabellingController. Until
+    // then the input takes its accessible name from `accessible-label`.
     return html`
       <div class="swc-TextField">
         <input
```
Rajdeepc, line 57: You need to have a `.value` binding or `@input`. Users will not be able to see the submitted value.

2nd-gen/packages/core/components/text-field/TextField.base.ts
```diff
@@ -183,16 +183,74 @@ export abstract class TextFieldBase extends SizedMixin(SpectrumElement, {
...
+
+  // @todo setSelectionRange() / select() delegate to the rendered native
+  // <input>; they land with the render implementation.
 
-  // @todo (Phase 5): setSelectionRange() / select() delegate to the rendered
-  // native <input>, so they land with the render implementation.
+  protected override firstUpdated(changedProperties: PropertyValues): void {
+    super.firstUpdated(changedProperties);
+    // The reset target is the initial `value` attribute.
+    this.fieldAssoc.defaultValue = this.value;
```
Rajdeepc: `formResetCallback` can fire before `firstUpdated` and permanently can wipe out the declared default value. You can delete this hook and read `getAttribute('value')` lazily in the `formResetCallback` instead.

2nd-gen/packages/core/components/text-field/TextField.base.ts
```diff
@@ -183,16 +183,74 @@ export abstract class TextFieldBase extends SizedMixin(SpectrumElement, {
...
+
+  private fieldAssoc = new FieldAssociationController(this.internals, {
+    onDisabledChange: () => this.requestUpdate(),
+  });
+
+  /**
+   * The host's own `disabled` OR the cascaded form / `<fieldset disabled>` state.
+   * The subclass render reads it to disable the inner control.
+   */
+  protected get effectiveDisabled(): boolean {
```
Rajdeepc, line 202: This only reaches the inner input. Under `<fieldset disabled>` the host gets no disabled attribute, no aria-disabled, no custom state and  `:host([disabled])` in `text-field.css` never applies and AT announces nothing. `DisabledMixin` does exactly this. You should use `internals.states.add('disabled')` and document `:host(:state(disabled))`.

2nd-gen/packages/core/controllers/field-association-controller/stories/demo-hosts.ts
```diff
@@ -0,0 +1,578 @@
...
+//     RADIO HARNESS
+// ────────────────────────────────────────
+
+/**
+ * @internal
+ *
+ * Single-radio harness. Mirrors the checkbox exclusion pattern: submits its
+ * value only when checked and not effectively disabled, otherwise `null`.
+ */
+@customElement('demo-field-radio')
```
Rajdeepc, line 219: This contradict [forms-strategy-rfc.md](https://github.com/adobe/spectrum-web-components/blob/main/CONTRIBUTOR-DOCS/03_project-planning/05_strategies/forms-strategy-rfc.md#32-where-aria-roles-live). For radio-like controls put role="radio" on the host via ElementInternals, and combobox isn't a <select>. The radio migration will copy the wrong pattern. Can you add a comment scoping them as plumbing-only harnesses.

2nd-gen/packages/core/controllers/field-association-controller/src/field-association-controller.ts
```diff
@@ -0,0 +1,138 @@
...
+    this._internals.setFormValue(value);
+  }
+
+  // ─────────────────────────
+  //     DEFAULT VALUE
+  // ─────────────────────────
+
+  /** The value the field restores to on `formResetCallback`. */
+  public get defaultValue(): string {
+    return this._defaultValue;
```
Rajdeepc: `defaultValue: string` can't express `checkbox/radio`;.`?? ''` is unreachable under the type.

2nd-gen/packages/core/components/text-field/TextField.base.ts
miwha-adobe: Good catch, will fix

2nd-gen/packages/core/components/text-field/TextField.base.ts
```diff
@@ -183,16 +183,68 @@ export abstract class TextFieldBase extends SizedMixin(SpectrumElement, {
...
+  }
+
+  /** Whether the field is a candidate for constraint validation. */
+  public get willValidate(): boolean {
+    return this.fieldAssoc.willValidate;
+  }
+
+  /** Runs constraint validation; returns whether the field is valid. */
+  public checkValidity(): boolean {
+    return this.fieldAssoc.checkValidity();
```
Rajdeepc, line 228: This will return a false pass on a required empty field. If `value:"" required:true` then `checkValidity is true and valueMissing: false, validationMessage: ""` and `form.checkValidity` will return `true.` I see you have deferred this to the labelling and render work but i feel it needs action here. 
Removing the host getters wouldn't help, because `form.checkValidity()` reads internals.validity directly. The false pass exists at the form level regardless of what the element exposes.

2nd-gen/packages/swc/components/text-field/TextField.ts
```diff
@@ -35,15 +35,22 @@ export class TextField extends TextFieldBase {
...
           class="input"
           aria-label=${ifDefined(this.accessibleLabel || undefined)}
+          .value=${this.value}
+          ?disabled=${this.effectiveDisabled}
+          @input=${this.handleInput}
```
Rajdeepc, line 62: no change event here. input will escape the composed shadow root. A consumer writing `<swc-text-field @change=…>` will get nothing. Maybe add something like this
```ts
private handleChange(): void {
  this.dispatchEvent(new Event('change', { bubbles: true, composed: true }));
}
// …and on the input:
@change=${this.handleChange}
```

2nd-gen/packages/core/components/text-field/TextField.base.ts
```diff
@@ -12,6 +12,7 @@
 import { PropertyValues } from 'lit';
 import { property } from 'lit/decorators.js';
 
+import { FieldAssociationController } from '@adobe/spectrum-wc-core/controllers/index.js';
```
Rajdeepc: This barrel will drag every controller into consumer bundles. Please use subpath.

2nd-gen/packages/core/controllers/field-association-controller/src/field-association-controller.ts
```diff
@@ -0,0 +1,123 @@
...
+    this._internals = internals;
+    this._onDisabledChange = options.onDisabledChange;
+  }
+
+  // ─────────────────────────
+  //     FORM VALUE
+  // ─────────────────────────
+
+  /** Sets the form value. Pass `null` to exclude the field from `FormData`. */
+  public setValue(value: FieldFormValue): void {
```
Rajdeepc: you should not foreclose state restore here. You can add an optional param now.
```suggestion
  public setValue(value: FieldFormValue, state?:FieldFormValue): void {
```

2nd-gen/packages/core/components/text-field/TextField.base.ts
miwha-adobe: I'm going to push back on this one here because Nikki is already in progress for the validity behavior / the labeling work includes the full validation behavior, and I don't want to create a conflict here. But I agree this should be addressed.

2nd-gen/packages/core/controllers/field-association-controller/stories/demo-hosts.ts
miwha-adobe: This guy is just demo only. refresh() is a private method on the story host, not controller API, so consumers never hit this. (The async timing is just native formDisabledCallback, not the controller.

PR #6700, opened 2026-09-02
ci(lint): surface ESLint & Stylelint warnings in PR annotations

caseyisonit:
## Summary

CI ran ESLint and Stylelint through reviewdog with `--quiet` and `level: error`, so **warnings were suppressed** — never annotated on PRs, accumulating silently. This surfaces them.

- Drop `--quiet` from the ESLint and Stylelint reviewdog steps.
- Set `level: warning` so warnings annotate changed lines.
- Keep `fail_level: error` and `filter_mode: diff_context` — the existing ~61 2nd-gen warning backlog does **not** block PRs.

Flip `fail_level` → `warning` (one line, noted in-file) once the backlog is cleared to gate newly introduced warnings.

## Scope

One file: `.github/workflows/lint.yml`. Carved out of the larger CI-tooling branch so PRs stay small.

Jira: SWC-2551

PR #6701, opened 2026-09-02
ci(type-check): add TypeScript check workflow, hooks, and unified check commands

caseyisonit:
## Summary

2nd-gen CI had no TypeScript compiler check — the vite build uses esbuild and strips types without type-checking. This adds one, plus local tooling to run the same checks before CI.

- **`type-check:2nd-gen`** — `tsc -b` build mode across the 2nd-gen composite project references; **`type-check`** aliases it. Adds `typescript` as a root devDependency so `tsc` is exposed to root scripts under Yarn Berry.
- **`check` / `check:2nd-gen`** — run all linters plus the type-check without short-circuiting; a single local "will CI pass" command.
- **`.github/workflows/type-check.yml`** — runs `tsc` via reviewdog.
- **pre-push hook** — runs `yarn check:2nd-gen`, non-blocking while the backlog is cleared.

## Open decision (see Jira)

CI currently reports **diff-only** (reviewdog `filter-mode: diff_context`) so it doesn't block on the ~28 pre-existing type errors. The ticket's original AC asks for a **whole-repo gate** (fail on any error anywhere). Two paths — resolve the backlog then gate whole-repo, or keep diff-only. Flagged in SWC-2549 for a decision before this is made blocking.

## Scope

Carved out of the larger CI-tooling branch. Sibling PR #6700 (ESLint warning surfacing, SWC-2551).

Jira: SWC-2549


cdransf (approved):

rubencarvalho:
oooh nice! I was thinking... what about we just fix the existing errors and run tsc as a blocking step (e.g. in the CI build and in the pre-push hook like you have here), instead of the reviewdog? wdyt??

rubencarvalho (approved):

2nd-gen/packages/swc/tsconfig.json
```diff
@@ -5,20 +5,31 @@
...
   "extends": "../../tsconfig.json",
   "include": [
     "**/*.ts",
     "**/*.d.ts",
     "vite-env.d.ts",
-    ".storybook/storybook-env.d.ts"
+    ".storybook/storybook-env.d.ts",
+    ".storybook/helpers/**/*.ts",
+    ".storybook/decorators/**/*.ts",
+    ".storybook/intl/**/*.json",
```
Rajdeepc, line 31: Please validate if TS globs descend into dot-directories or not with `tsc --listFilesOnly`. 
My recommendation would be not to add more patterns. You can replace `.storybook/helpers/**/*.ts` / `.storybook/decorators/**/*.ts` with a single `.storybook/**/*.ts` and `.storybook/**/*.json` if JSON imports are involved so that nothing under the dot-directory is skipped.

package.json
```diff
@@ -48,7 +50,9 @@
     "tokens:diff": "yarn workspace @adobe/swc-tokens diff:versions",
     "tokens:fix": "yarn workspace @adobe/swc-tokens fix:tokens",
     "tokens:test": "yarn workspace @adobe/swc-tokens test && yarn workspace @adobe/postcss-token test && yarn workspace swc-vscode-token test",
-    "tokens:update": "yarn tokens:test && yarn workspace swc-vscode-token tokens && yarn workspace @adobe/spectrum-wc stylesheet:tokens && yarn workspace @adobe/spectrum-wc stylesheet:typography"
+    "tokens:update": "yarn tokens:test && yarn workspace swc-vscode-token tokens && yarn workspace @adobe/spectrum-wc stylesheet:tokens && yarn workspace @adobe/spectrum-wc stylesheet:typography",
+    "type-check": "yarn type-check:2nd-gen",
+    "type-check:2nd-gen": "tsc -b 2nd-gen/packages/core 2nd-gen/packages/icons 2nd-gen/packages/swc --force --pretty false"
```
Rajdeepc: This `--force` will bypass incremental `.tsbuildinfo` cache on every call. Every git push will pay the full `tsc -b` cost across core, icons and swc. You have the same in the `.husky/pre-push` hook also.
Please drop `--force` from here, `tsc -b` is already correctly invalidated by file changes. `--force` is needed for cache corruption recovery not for routine runs.

2nd-gen/packages/swc/tsconfig.json
caseyisonit: Validated with `tsc --listFilesOnly`: bare `**/*.ts` matches 0 files under `.storybook/`, so the dot-directory is skipped and explicit patterns are required.

I tried the single `.storybook/**/*.ts` + `.storybook/**/*.json` but it pulls in Storybook config/tooling that's out of scope for the gate and errors on its own:
- `.storybook/main.ts` — implicit `any` binding elements, `reactDocgen` option
- `.storybook/preview.ts` — cannot find `./DocumentTemplate.mdx`
- `.storybook/blocks/index.ts` — resolves to `.tsx`, needs `--jsx`
- `.storybook/addons/screen-reader-addon/...` — untyped `query-selector-shadow-dom` dep

So I kept the includes scoped to just the `helpers`/`decorators`/`intl` that the stories and VRT actually import.

package.json
caseyisonit: Done — dropped `--force` from `type-check:2nd-gen` (the pre-push hook calls the same script, so it's covered there too). Agreed, `tsc -b` is already correctly invalidated by file changes.

PR #6708, opened 2026-09-03
fix: correct accidentally-released gen1/gen2 versions (1.12.3, 2.0.0-beta.3)

blunteshwar:
## Description

Corrects the version state on `main` so the next real release doesn't collide with versions already published to npm:

- **gen1 `1.12.3`**: this version was released accidentally and has since been deprecated on npm. `main`'s `package.json` files were never bumped to reflect it.
- **gen2 `2.0.0-beta.3`**: this version was published from a separate feature branch (not `main`), so `main`'s `2nd-gen` package.json files never caught up to reflect it either.

This PR:
- Bumps every 1st-gen package's `version` field (fixed group, all 80 `@spectrum-web-components/*` packages) and their internal cross-package dependency pins from `1.12.2` to `1.12.3`.
- Bumps the linked group (`@adobe/spectrum-wc`, `@adobe/spectrum-wc-core`) from `2.0.0-beta.2` to `2.0.0-beta.3`, and updates the exact-pinned dependency on core in both `swc` and `icons`. `@adobe/spectrum-wc-icons`' own version is untouched - `swc` depends on it via `workspace:*`, no coupling required.
- Regenerates `version.ts` for both generations via `node scripts/generate-versions.js` to match.

No functional changes - this is a version-number correction so `main`'s committed state matches what was actually published.

## Motivation and context

Without this, the next `yarn changeset version` run for either generation would compute `1.12.3` / `2.0.0-beta.3` as the next version (since `main`'s `package.json` files hadn't caught up), colliding with versions that were already published: `1.12.3` accidentally and now deprecated, and `2.0.0-beta.3` shipped from a feature branch outside `main`'s normal release flow.


Rajdeepc (approved):
Looks fine! But lets be careful on how we test going forward.

rubencarvalho (approved):

PR #6709, opened 2026-09-03
feat(menu): migrate menu API to 2nd-gen

Rajdeepc:
## Summary

Migrates `swc-menu`'s Phase A API (per the [migration plan](https://github.com/adobe/spectrum-web-components/blob/rajdeep/menu-migration/CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md)) from 1st-gen to 2nd-gen:

- **Properties**: `for` / `triggerElement`, `direction`, `align`, `shouldFlip`, `open`, `size`
- **Trigger wiring**: resolves `for`/`triggerElement` via the shared `resolveTrigger` utility and wires `aria-controls` / `aria-expanded` / `aria-haspopup` plus click-to-toggle, matching `Tooltip`/`Popover`
- **Positioning**: `PlacementController` anchors the surface; `direction`/`align` translate to a `Placement` value internally
- **Events**: `swc-open`, `swc-after-open`, `swc-close`, `swc-after-close`
- **Slot**: default slot restricted to `swc-menu-item` this phase, enforced via `validateAllowedChildren`
- **JSDoc**: added throughout for the new public API

**Not in this PR** (deferred per the plan): `disabled` (open question, tracked as Q22), selection (Phase B/C), keyboard/focus management (Phase 4), and any rendering/styling of the anchored surface (Phase 5).

Also updates the shared-resources doc to list `menu` as a new consumer of `PlacementController`, `SizedMixin`, `resolveTrigger()`, and `validateAllowedChildren()`.

## Test plan

- [x] `tsc --noEmit` clean on both `core` and `swc` packages for all menu files
- [x] ESLint and Prettier clean on all changed files
- [x] Manually cross-checked implementation against the migration plan's Phase A property/event/slot tables


2nd-gen/packages/core/components/menu/Menu.base.ts
```diff
@@ -10,17 +10,348 @@
...
+  @property({ attribute: false })
+  public triggerElement: HTMLElement | null = null;
+
+  /**
+   * Where the anchored surface opens relative to the trigger.
+   *
+   * @default 'bottom'
+   */
+  @property({ type: String, reflect: true })
+  public direction: MenuDirection = 'bottom';
```
5t3ph: Let's continue using `placement` to match Popover and Tooltip.

2nd-gen/packages/core/components/menu/Menu.base.ts
```diff
@@ -10,17 +10,348 @@
...
+
+  /**
+   * Cross-axis alignment of the anchored surface against the trigger.
+   *
+   * @default 'start'
+   */
+  @property({ type: String, reflect: true })
+  public align: MenuAlignment = 'start';
```
5t3ph: I suggest dropping this as the full `placement` values (as used by Popover) include logical alignment.

2nd-gen/packages/core/components/menu/Menu.base.ts
```diff
@@ -10,17 +10,348 @@
...
+  private readonly _handleTriggerClick = (): void => {
+    this.open = !this.open;
+  };
+
+  /**
+   * The element `PlacementController` positions: `swc-menu`'s shadow-internal
+   * `role="menu"` surface. Returns `null` in the base class; the SWC
+   * rendering layer overrides this once that surface exists.
+   */
+  protected get surfaceElement(): HTMLElement | null {
```
5t3ph, line 150: Are any of the following methods already available in Popover and can they be move to a shared location?

5t3ph (commented):
I think there's a bit more sharing we can do from Popover.

2nd-gen/packages/core/components/menu/Menu.base.ts
rubencarvalho: not yet afaik

2nd-gen/packages/core/components/menu/Menu.base.ts
Rajdeepc: Thanks for the feedback. Before I make any change I want to flag a wringkle that Popover's placement values dont include `left/right` at all but React Spectrum S2's own `Menutrigger.direction` foes support `left/right` and that the source we can match for Menu's API. 
We can still consolidate down to one `placement` property but lets give it its own Menu specific value list that keeps all six sides.
WDYT? @5t3ph

5t3ph: The `Placement` type exported from the `PlacementController` types does include `left` and `right`, and the alignment extensions like `left-top`, so it will just be a tweak of what is "valid" from that full type.

Rajdeepc: Done, switched to a single `placement` property.

2nd-gen/packages/core/components/menu/Menu.base.ts
Rajdeepc: Dropped it, folded into placement. One thing to flag though: if I match Popover's exact 12 values 1:1, we lose left/right, since Popover's own placement set only covers its logical start/end sides, not left/right. React Spectrum's MenuTrigger.direction does support left/right, so I kept those in a Menu-specific MenuPlacement union instead of reusing Popover's list outright. Let me know if you'd rather just match Popover exactly and drop left/right for Menu too.

2nd-gen/packages/core/components/menu/Menu.base.ts
Rajdeepc: Yeah, matches Q12 in the plan for now, swc-action-menu doesn't exist yet so there's nothing proven-identical to pull into a shared base. Once that lands I'll diff the two and pull out whatever's actually the same.

2nd-gen/packages/core/components/menu/Menu.base.ts
rubencarvalho, line 1: Should we also support a Virtual Trigger? I know that is a common use case :)

2nd-gen/packages/core/components/menu/Menu.base.ts
```diff
@@ -10,17 +10,310 @@
...
+      changedProperties.has('for') ||
+      changedProperties.has('triggerElement') ||
+      changedProperties.has('open')
+    ) {
+      this.wireTrigger();
+    }
+
+    const openChanged = changedProperties.has('open');
+    if (openChanged) {
+      this.dispatchOpenEvents(this.open);
```
rubencarvalho, line 302: I think this will always fire on the very first update

5t3ph (approved):

2nd-gen/packages/core/components/menu/Menu.base.ts
Rajdeepc: Good catch, that was real. changedProperties.has('open') is true on the very first updated() call too since Lit records the property's own initializer as a change, so every menu was firing a phantom swc-close (or swc-open if it started open) right on mount. Fixed by also checking changedProperties.get('open') !== undefined, that's only true for a real prior value, not the init entry.

2nd-gen/packages/core/components/menu/Menu.base.ts
Rajdeepc: Yeah that's a real use case, right-click context menus especially. Held off for this pass though since it's not a drop-in type widening: resolveTrigger's triggerElement option is HTMLElement-only, and Popover has to bypass it and special-case ARIA wiring and the click-to-toggle listener for its VirtualTrigger branch since neither applies to something with no real DOM presence. Logged it as Q23 in the plan so it's not lost, happy to pick it up as a follow-up if there's a concrete need.

rubencarvalho (approved):

PR #6482, opened 2026-07-08
fix(storybook): disable Vite publicDir to stop static copy race

blunteshwar:
## Description

The 2nd-gen **Build second-gen Storybook** step (`yarn workspace @adobe/spectrum-wc storybook:build`) fails intermittently with:

```
Error: EEXIST: file already exists, mkdir
'./storybook-static/.well-known/agent-skills/spectrum-web-components/references/guides'
  at async mkDirAndCopy (node:internal/fs/cp/cp)
```

Root cause: `storybook build` copies `public/` into `storybook-static/` **twice, concurrently**:

1. Storybook's own `staticDirs: ['../public']` copy (`copyAllStaticFilesRelativeToMain`).
2. Vite's default `publicDir`. `builder-vite` sets Vite `root = resolve(configDir, '..')` (the package dir) and never sets `publicDir`, so Vite defaults to `<root>/public` — the same folder.

Both copies run inside a single `Promise.all([viteBuild, staticDirsCopy])`. Node's `cp` recurses sequentially and issues a **non-recursive** `await mkdir(dest)` per directory, so when the two copiers reach the same nested directory the loser throws `EEXIST`. Timing-dependent, hence flaky. The large generated `.well-known/agent-skills` tree (`yarn generate:skills`, added in #6423) widened the collision window, which is why the failures started appearing.

The fix sets `publicDir: false` in `viteFinal`, so Storybook's `staticDirs` copy is the single writer of `public/` and there is no second, racing copier.

## Motivation and context

Intermittent red builds on the docs-deployment path (`preview-docs.yml` and `publish-2ndgen-docs.yml`, both of which call `storybook:build`) block unrelated PRs and force manual reruns. This removes the race at its source rather than papering over it with a retry.

Reproduction (two concurrent copies of `public/` into one destination) throws `EEXIST` 40/40 on Node v24.11.1; a single copy never does.

## Related issue(s)

- No linked issue; surfaced by a failing `preview-docs` run.

## Screenshots (if appropriate)

n/a — build configuration change, no visual output.

## Author's checklist

- [x] I have read the **[CONTRIBUTING](https://github.com/adobe/spectrum-web-components/blob/main/CONTRIBUTING.md)** and **[PULL_REQUESTS](https://github.com/adobe/spectrum-web-components/blob/main/PULL_REQUESTS.md)** documents.
- [x] I have reviewed the Accessibility Practices for this feature, see: [Aria Practices](https://www.w3.org/TR/wai-aria-practices/)
[...]

PR #6710, opened 2026-09-03
fix(gen2): disable Vite's redundant publicDir copy to fix flaky prod Storybook build

blunteshwar:
## Description

Fixes an intermittent failure in the "Build Gen2 Storybook (prod mode)" CI step (`preview-docs.yml`):

```
Error: EEXIST: file already exists, mkdir
'./storybook-static-prod/.well-known/agent-skills/spectrum-web-components/references/guides'
```

**Root cause:** `2nd-gen/packages/swc/public/` was getting copied into the Storybook build output twice:
1. Storybook's own `staticDirs: ['../public', ...]` (`.storybook/main.ts`) copies it first.
2. Right after, `vite build` runs its own separate built-in `publicDir` → `outDir` copy, since `vite.config.ts` never set `publicDir: false` (it defaults to the same `public/` folder, same output dir). Vite's copy doesn't know Storybook already put those files there, so it tries to `mkdir` a directory that already exists and crashes.

It's sequential, not a timing race - copy #2 always runs into copy #1's leftovers. The "sometimes passes" flakiness comes from `fs.cp`'s inconsistent handling of pre-existing nested directories, not the two copies actually overlapping in time.

**Fix:** set `publicDir: false` in `vite.config.ts` - Storybook's `staticDirs` already does this copy correctly, so Vite's redundant one is disabled. Side benefit: this also stops `public/.well-known/agent-skills/...` (Storybook-only static assets) from leaking into the npm-published `dist/` during the library build, which shares this same config.

## Motivation and context

CI has been intermittently failing on this step, requiring manual re-runs. No known reason to keep the duplicate copy.

## Testing

Verified locally: `yarn storybook:build --output-dir storybook-static-prod-test` in `2nd-gen/packages/swc` completes successfully with this change, and `.well-known/agent-skills/...` still lands correctly in the output (via `staticDirs` alone).

### DIFF
--- a/2nd-gen/packages/swc/vite.config.ts
+++ b/2nd-gen/packages/swc/vite.config.ts
@@ -109,6 +109,13 @@ export default defineConfig({
       },
     }),
   ],
+  // Storybook's `staticDirs: ['../public', ...]` (.storybook/main.ts) already copies
+  // this package's public/ directory into the Storybook build output. Vite's own
+  // built-in publicDir copy is redundant with that and, run right after it against the
+  // same output directory, throws EEXIST on the pre-existing nested directories. It
+  // also has no reason to bundle public/ (Storybook-only static assets) into the
+  // library's published dist/.
+  publicDir: false,
   css: {
     transformer: 'postcss',
     postcss: {


pfulton:
Would we prefer this one over #6482? If yes, can we close out the other?

blunteshwar:
> #6482

The older one got stale so let us prefer this one

cdransf (approved):

aramos-adobe (approved):

PR #6716, opened 2026-09-04
feat(asset): files, API, a11y, and styling

5t3ph:
## Description

Expands `swc-asset` (Spectrum 2's internal image/media primitive) into a general-purpose primitive that consumers can slot directly, per the Asset component plan (`CONTRIBUTOR-DOCS/03_project-planning/03_components/asset/component-plan.md`)

**API changes:**
- Removed `variant` (`file`/`folder` built-in icons) and `label`; slotted content (`` or `<svg>`) is now the only supported input
- Added `aspectRatio` (CSS `<ratio>` syntax, plus a `square` keyword and `:`-to-`/` normalization), `width`, `height` for sizing
- Added `fit` (`cover` default, `contain`) for how slotted content fills the box
- Added `background` (`transparent` default, `solid`, `checkerboard`) for what shows behind/around slotted content
- Added `decorative` and `accessibleLabel` (renamed from `label`, generalized into a fallback accessible name)
- Promoted from internal genre (`@status internal`, `.internal.*` naming) to public

**Accessibility:**
- `decorative` sets `aria-hidden="true"` on the host
- `accessibleLabel` is applied as a fallback accessible name (`alt` on ``, `role="img"` + `aria-label` on `<svg>`) only when the slotted content doesn't already carry its own accessible name
- DEBUG warnings for: invalid `fit`/`background` values, `aspectRatio` combined with both `width` and `height`, more than one slotted child, an unsupported child type, and a missing accessible name

**Styling:**
- `aspect-ratio`/`width`/`height` resolve through a private-then-exposed custom-property chain on the internal wrapper (never a literal `aspect-ratio`/`object-fit` on the host), so an ancestor (e.g. Card) can hand Asset a default via `--swc-asset-aspect-ratio` without breaking an explicit value
- `background="checkerboard"` reuses the shared `opacity-checkerboard.css` fragment; `background="solid"` exposes `--swc-asset-background-color`
- `.swc-Asset` gets `border-radius: inherit` and `overflow: hidden` so a host-level `border-radius` clips slotted content

## Motivation and context

`swc-asset` today is a minimal file/folder-icon-or-slot component with no sizing, fit, loading-state, or background API — not enough to serve as Card's image primitive or a general-purpose media slot. This is not a 1st-gen → 2nd-gen migration (`sp-asset` is unaffected and out of scope); it's an in-place expansion of the existing internal `swc-asset` in preparation for genre promotion and use by Card and other consumers.

## Related issue(s)

- SWC-2317 (Epic)
- SWC-2319

## Screenshots

image

image

### Manual review test cases

- [ ] _Fit and background render correctly at a fixed size_
  1. Go to `?path=/story/components-asset--fit` and `?path=/story/components-asset--background`
  2. Compare `cover` vs `contain`, and `transparent`/`solid`/`checkerboard`
  3. Expect `contain` to letterbox against the chosen `background`; `checkerboard` should render the shared opacity-checkerboard pattern

- [ ] _Sizing via aspect-ratio vs width/height_
  1. Go to `?path=/story/components-asset--sizing`
  2. Confirm the `16/9` aspect-ratio example and the explicit width/height example both size correctly with no collapse to zero size

- [ ] _Dev-mode warnings fire as expected_
  1. Open the `Asset/Tests` story group in dev mode with the console open
  2. Run through `Invalid Fit Warning Test`, `Invalid Background Warning Test`, `Multiple Children Warning Test`, `Unsupported Child Type Warning Test`, and `Missing Accessible Name Warning Test`
  3. Expect a single, clear `console.warn` per scenario, referencing the offending value/attribute

### Device review

- [ ] Did it pass in Desktop?

  - `<swc-asset>` has no interactive parts and is not focusable by design (matches the a11y model used by Thumbnail: the parent embedding Asset owns any interactive/focus state).
  1. Go to `?path=/story/components-asset--overview`
  2. Tab through the page
  3. Expect the asset to be skipped entirely; focus moves directly to the next focusable element with no stop on `<swc-asset>`

  1. Go to `?path=/story/components-asset--accessibility`
  2. With VoiceOver (or NVDA/JAWS) active, navigate to each asset example
  3. Expect the `` example to announce from its own `alt`, and the `<svg>` example to announce from `role="img"` + `aria-label`


2nd-gen/packages/core/components/asset/Asset.base.ts
```diff
@@ -14,54 +14,222 @@ import { PropertyValues } from 'lit';
...
+    });
+    warnIf(
+      this,
+      typeof this.aspectRatio !== 'undefined' &&
+        typeof this.width !== 'undefined' &&
+        typeof this.height !== 'undefined',
+      `<${this.localName}> "aspect-ratio" has no effect when both "width" and "height" are set.`,
+      DOCS_URL
+    );
+    this.#validateSlottedContent();
```
cdransf: Should we gate this behind `isDebug()`? ✨

2nd-gen/packages/core/components/asset/Asset.base.ts
```diff
@@ -14,54 +14,222 @@ import { PropertyValues } from 'lit';
...
+    this.#resolveAccessibleName();
```
cdransf: Should we compute `Array.from(this.children);` once here and pass it to this and `#validateSlottedContent();`? ✨

cdransf:
It looks like we've got some test failures that we need to clean up (looks like string mismatches and some timeouts). https://app.circleci.com/pipelines/github/adobe/spectrum-web-components/43182/workflows/987e895a-1df0-442f-913a-5bf666ee5a05/jobs/1000055 ✨

2nd-gen/packages/core/components/asset/Asset.base.ts
```diff
@@ -14,54 +14,222 @@ import { PropertyValues } from 'lit';
...
+    );
+  }
+
+  /**
+   * Whether `child` (an `<img>` or `<svg>`, per `tagName`) already carries
+   * its own accessible name and should be left untouched.
+   */
+  #hasOwnAccessibleName(child: Element, tagName: 'img' | 'svg'): boolean {
+    if (tagName === 'img') {
+      return child.hasAttribute('alt');
```
cdransf: Should we check `aria-label`/`aria-labelledby` for image too? ✨

cdransf (approved):
Looks great! Just a few questions and some test failures to clean up. ✨

2nd-gen/packages/swc/components/asset/test/asset.a11y.spec.ts
```diff
@@ -46,18 +46,8 @@ test.describe('Asset - ARIA Snapshots', () => {
...
-  test('should handle variants', async ({ page }) => {
-    const root = await gotoStory(
-      page,
-      'components-asset--variants',
-      'swc-asset'
-    );
-    await expect(root).toMatchAriaSnapshot(`
-      - img "README.md"
+      - img "README.md preview"
+      - img "Folder icon"
```
Rajdeepc: Dont see this string under `stories/`. Do you `yarn test:a11y` pass?

2nd-gen/packages/core/components/asset/Asset.base.ts
Rajdeepc: There is no `@slotchange` on the default slot and light DOM mutation doesnt schedule a Lit update. If you swap the `img` after first render the replacement will end up with `alt=null` and without a DEBUG warning as mentioned by @cdransf this will drift.

2nd-gen/packages/core/components/asset/Asset.base.ts
```diff
@@ -14,54 +14,222 @@ import { PropertyValues } from 'lit';
...
+   *    slotted node (`alt` for `<img>`, `aria-label` + `role="img"` for
+   *    `<svg>`, which has no native `alt`).
+   * 4. None of the above → DEBUG warning.
+   */
+  #resolveAccessibleName(): void {
+    if (this.decorative) {
+      this.setAttribute('aria-hidden', 'true');
+      return;
+    }
+    this.removeAttribute('aria-hidden');
```
Rajdeepc: This is an open remove. Can you at-least track whether the component have set it or not.

2nd-gen/packages/swc/components/asset/asset.mdx
```diff
@@ -0,0 +1,65 @@
...
+
+The asset automatically centers its content both horizontally and vertically within the available space.
+
+<Canvas of={AssetStories.Anatomy} />
+
+## Options
+
+### Fit
+
+`fit` controls how slotted content fills the asset's box: `cover` (default) scales content to fill
```
Rajdeepc, line 24: Do fit has any effect on a slotted `svg`? Kindly verify!

2nd-gen/packages/swc/components/asset/Asset.ts
```diff
@@ -11,95 +11,63 @@
...
+ * @slot - a single `<img>` or `<svg>` element to display
  *
- * @status internal
- * @since 2.0.0-beta.1
+ * @cssprop --swc-asset-aspect-ratio - Fallback aspect ratio applied when
+ * `aspectRatio` is unset; an ancestor can set this to hand Asset a default.
+ * @cssprop --swc-asset-background-color - Background color used when
+ * `background="solid"`. Defaults to `token("gray-100")`.
+ *
+ * @since 2.0.0-beta.3
```
Rajdeepc: This should be `2.0.0-beta.1`. Asset existed since beta.1 but please do cross check.

2nd-gen/packages/swc/components/asset/asset.mdx
```diff
@@ -0,0 +1,65 @@
...
+The `<swc-asset>` element implements several accessibility features:
+
+#### ARIA implementation
+
+- **Non-interactive**: assets have no interactive behavior and are not focusable
+
+### Best practices
+
+- Provide a meaningful `alt` attribute on a slotted `<img>`, or `aria-label` (with `role="img"`) on a slotted `<svg>`
+- For decorative images, use an empty `alt=""` attribute on the img tag
```
Rajdeepc: This statement is competing with `@decorative` property.

2nd-gen/packages/swc/components/asset/asset.mdx
```diff
@@ -0,0 +1,65 @@
...
+
+An asset consists of:
+
+- **Default slot**: a single `<img>` or `<svg>` element to display
+
+The asset automatically centers its content both horizontally and vertically within the available space.
+
+<Canvas of={AssetStories.Anatomy} />
+
+## Options
```
Rajdeepc, line 20: You missed adding `decorative` or `accessibleLabel` cssprops here.

2nd-gen/packages/core/components/asset/Asset.base.ts
```diff
@@ -14,54 +14,222 @@ import { PropertyValues } from 'lit';
...
+ * `:`-separated ratios (e.g. `16:9`) to the CSS `<ratio>` separator (`16/9`).
+ */
+function normalizeAspectRatio(value: string | undefined): string | undefined {
+  if (!value) {
+    return value;
+  }
+  if (value === 'square') {
+    return '1/1';
+  }
+  return value.replace(':', '/');
```
Rajdeepc, line 60: This will replace only first occurrence, any malformed ration will get no validation at all though `fit/background`  both go through `validateEnum`

Rajdeepc (commented):
@5t3ph Should we also expose `--swc-asset-width/--swc-asset-height` layer layer to match the `aspect-ratio` chain so an ancestor like Card can also hand down a default-ratio but not a default size?

2nd-gen/packages/swc/components/asset/Asset.ts
5t3ph: I bumped this because it was not public then, but it will be with this release, plus it is essentially entirely rewritten/re-purposed.

2nd-gen/packages/swc/components/asset/test/asset.a11y.spec.ts
5t3ph: I missed a re-test after changing the strings - updated now

2nd-gen/packages/core/components/asset/Asset.base.ts
5t3ph: Resolved and added regression test

2nd-gen/packages/swc/components/asset/asset.mdx
5t3ph: Good call! Confirmed it didn't (`object-fit` only applies to CSS replaced elements). Fixed by driving the SVG's own `preserveAspectRatio` from `fit`, documented in the MDX. It still doesn't extend it beyond the bounds (a faux crop) like an image, it just essentially causes it to fill up as much as it can until it hits either its height or width dimension.

2nd-gen/packages/swc/components/asset/asset.mdx
5t3ph: Updated MDX descriptions

2nd-gen/packages/swc/components/asset/asset.mdx
5t3ph: I think `cssprops` was a typo here, but these are now added under Accessibility. We can evaluate in the docs phase if we want to add explicit Options stories.

2nd-gen/packages/core/components/asset/Asset.base.ts
5t3ph: There is now a `warnIf` against a new regex in case of malformed aspect ratio so it doesn't silently break.

2nd-gen/packages/core/components/asset/Asset.base.ts
```diff
@@ -14,54 +14,269 @@ import { PropertyValues } from 'lit';
...
+      `<${this.localName}> "aspect-ratio" has no effect when both "width" and "height" are set.`,
+      DOCS_URL
+    );
+
+    const children = Array.from(this.children);
+    if (isDebug()) {
+      this.validateSlottedContent(children);
     }
+    this.resolveAccessibleName(children);
+    this.applyFitToSvg(children);
```
Rajdeepc, line 184: This will overwrite a consumer's `preserveAspectRatio`. You can early return when the attribute is already present and this writes on every update even when the value is unchanged.
This is an unnecessary DOM write on a production path. Please guard with a read-compare.

2nd-gen/packages/core/components/asset/Asset.base.ts
```diff
@@ -14,54 +14,269 @@ import { PropertyValues } from 'lit';
...
+  private resolveAccessibleName(children: Element[]): void {
+    if (this.decorative) {
+      this.setAttribute('aria-hidden', 'true');
+      this._appliedAriaHidden = true;
+      return;
+    }
+    if (this._appliedAriaHidden) {
+      this.removeAttribute('aria-hidden');
+      this._appliedAriaHidden = false;
+    }
```
Rajdeepc, line 227: `_appliedAriaHidden` is claiming an attribute it is not setting. 
You should check with `this._appliedAriaHidden = !this.hasAttribute('aria-hidden')` before setting

2nd-gen/packages/core/components/asset/Asset.base.ts
```diff
@@ -14,54 +14,269 @@ import { PropertyValues } from 'lit';
...
-  /**
-   * @internal
-   *
-   * A readonly array of all valid variants for the asset.
-   *
-   * This is an actual internal property, intended not for customer use
-   */
-  static readonly VARIANTS: readonly AssetVariant[] = ASSET_VARIANTS;
+/** A CSS `<ratio>`: one or two positive numbers separated by `/`. */
+const ASPECT_RATIO_PATTERN = /^\d+(\.\d+)?(\s*\/\s*\d+(\.\d+)?)?$/;
```
Rajdeepc: I think this will also warn on valid CSS like `auto`, `auto 16/9`, `.5` these all will warn. You should allow `auto` and leading dot-numbers.

2nd-gen/packages/core/components/asset/Asset.base.ts
```diff
@@ -14,54 +14,269 @@ import { PropertyValues } from 'lit';
...
+      prop: 'background',
+      value: this.background,
+      valid: ASSET_BACKGROUND_VALUES,
+      url: DOCS_URL,
+    });
+    warnIf(
+      this,
+      typeof this.aspectRatio !== 'undefined' &&
+        !ASPECT_RATIO_PATTERN.test(this.aspectRatio),
+      `<${this.localName}> expects "aspect-ratio" to be a CSS <ratio> (e.g. "16/9"), the "square" keyword, or a ":"-separated ratio (e.g. "16:9"). Received "${this.aspectRatio}".`,
```
Rajdeepc: `aspect-ratio="foo:bar"` will warn `Received "foo/bar"` You should keep the raw input for the message.

2nd-gen/packages/swc/components/asset/Asset.ts
Rajdeepc: Your rationale is right but `beta.3` is already published as `latest`, so nothing here can ship in it. This would need an unreleased version then.

2nd-gen/packages/core/components/asset/Asset.base.ts
```diff
@@ -14,54 +14,269 @@ import { PropertyValues } from 'lit';
...
+    tagName: 'img' | 'svg'
+  ): boolean {
+    const hasAriaName =
+      child.hasAttribute('aria-label') || child.hasAttribute('aria-labelledby');
+    if (tagName === 'img') {
+      return hasAriaName || child.hasAttribute('alt');
+    }
+    const hasRoleImg = child.getAttribute('role') === 'img';
+    const hasTitleChild = !!child.querySelector(':scope > title');
+    return hasRoleImg && (hasAriaName || hasTitleChild);
```
Rajdeepc: `hasOwnAccessibleName` requires `role="img"` and a name source. The `` half is fixed but this is the other half. `<svg><title>Chart</title></svg>` which is a widely-used accessible pattern which still warns, and `accessibleLabel` overwrites it.
You should treat a child `title` as a name on its own.

2nd-gen/packages/core/components/asset/Asset.base.ts
```diff
@@ -14,54 +14,269 @@ import { PropertyValues } from 'lit';
...
+      if (tagName === 'img') {
+        (child as HTMLImageElement).alt = this.accessibleLabel;
+      } else {
+        child.setAttribute('role', 'img');
+        child.setAttribute('aria-label', this.accessibleLabel);
+      }
+      return;
+    }
+
+    warnIf(
```
Rajdeepc, line 253: This will read as a bug whose condition is the literal `true` at the call site. You can either restructure so the condition is real like `warnIf(this, !hasName, ....)` with the early returns folded into a boolean or you can add a plain `warn()` primitive to `dev-validation.ts` alongside `warnIf`

2nd-gen/packages/core/components/asset/Asset.base.ts
5t3ph: After re-evaluating and double-checking accessible SVG patterns, I've updated this to "self-heal" the role (add if missing) but not warn just for a missing role, only warn for missing labeling method and accept `title` as priority over `accessible-label`.

2nd-gen/packages/core/components/asset/Asset.base.ts
5t3ph: Chose the restructure option, but added SWC-2564 to separately consider adding the `warn()` primitive to the system

2nd-gen/packages/core/components/asset/Asset.base.ts
5t3ph: I missed that the `auto && <ratio>` option was supported, so TIL! Updated.

2nd-gen/packages/swc/components/asset/Asset.ts
5t3ph: Whoops - changed to `beta.4`

Rajdeepc (approved):
Thanks for working through my feedback!

PR #6717, opened 2026-09-07
fix(ai-toolkit): document @cssprop JSDoc and remove over-exposed custom properties

rubencarvalho:
## Summary

- Documents `@cssprop` JSDoc for every custom property the AI Toolkit patterns actually expose (`swc-pixel-loader`, `swc-upload-attachment`, `swc-prompt-field`, `swc-response-status`, `swc-response-status-step`) — previously undocumented across the board.
- Audited every exposed `--swc-*` custom property against the [Custom properties style guide](2nd-gen/packages/swc/.storybook/contributor-docs/02_style-guide/01_css/02_custom-properties.mdx#component-custom-property-exposure): "Expose only when the component itself overrides the property based on its own variant, state, or size needs. Do not expose properties for consumer convenience alone."
- Removed 29 properties that had a single fixed value with no internal variant/state/size override anywhere in the component. These become plain static token/pixel values. (Two later passes caught inconsistencies from the first: `user-message`'s per-type padding/gap properties, and `message-sources`' toggle gap/padding, which had only been renamed rather than re-evaluated against the same test as their siblings.)
- Made three properties private (`--_swc-*`) instead of removing them outright, since each is either reused multiple times within one rule or is pure internal wiring, not a real consumer override surface: `--swc-prompt-field-textarea-min-rows`/`-max-rows` (JS→CSS plumbing for the `min-rows`/`max-rows` attributes) and `--swc-response-status-step-shimmer-spread` (reused in two `calc()` expressions in the same rule).
- Replaced hardcoded pixel values with the matching design token wherever one exists at that exact value:
  - `border: 1px solid transparent` → `token("border-width-100")` (matches existing convention, e.g. `action-button`), plus several `2px`/`3px`/`6px`/`16px` gaps, margins, padding, and insets → their `spacing-*`/`base-padding-vertical-*` token equivalents.
  - Element dimensions that exactly match a platform-scale token → `component-height-*` / `thumbnail-size-*` / `workflow-icon-*`: message-feedback's 32px button, upload-attachment's 24px file-type badge, user-message's 32px card thumbnail, response-status's 20px loader/leading-icon. These resolve through the same `--swc-theme--sizeM` variable every other shipped component (avatar, button, action-button) already uses, so this isn't new behavior. (Easy to miss on a first pass: these tokens are responsive desktop/mobile *objects* in the token source, not flat px strings, so a naive value-string scan skips them entirely.)
  - Bespoke Figma-spec dimensions with no matching token (attachment tile sizes, badge geometry, box-shadow offsets, scrollbar width, etc.) are left as literals.

## Kept public (with rationale)

- `--swc-pixel-loader-size` / `-color`: required for nested-component relationships (consuming patterns like `prompt-field` and `response-status` re-theme the nested loader from outside its shadow root).
- `--swc-prompt-field-brand-color`, `-max-block-size`, attachment sizing props: tunable content/config knobs consistent with existing precedent already merged in this component.
- `--swc-upload-attachment-*` sizing props: vary per `type` attribute (card vs. media map to genuinely different rendered sub-elements); `-focus-indicator-color` is a documented, intentional non-token color (no matching design token exists yet) shared with a nested `swc-action-button`.
- `--swc-response-status-label-max-lines`, `--swc-response-status-step-detail-max-block-size`: tunable knobs parallel to existing prompt-field precedent (confirmed as the desired approach).

No visual change: every removed/renamed/retokenized/privatized property previously resolved to its existing fallback (or, for the token swaps, the exact same pixel value) in every shipped usage.

## Test plan

- [x] `yarn stylelint` on all touched CSS — clean
- [x] `yarn eslint` on all touched TS — clean
- [x] `yarn vitest run patterns/ai-toolkit --project storybook` — 134/134 tests passing
- [x] Confirmed no remaining references to removed/renamed property names anywhere in the repo (tests, stories, docs) outside the changeset itself
- [ ] Reviewer sign-off on the removal list, since this changes the public custom-property surface of a beta (`2.0.0-beta.3`) package


2nd-gen/packages/swc/patterns/ai-toolkit/message-sources/message-sources.css
```diff
@@ -32,10 +32,10 @@
 
 .swc-MessageSources-toggle {
   display: inline-flex;
-  gap: var(--swc-sources-toggle-gap, token("spacing-100"));
+  gap: var(--swc-message-sources-toggle-gap, token("spacing-100"));
   align-items: center;
   inline-size: fit-content;
-  padding: var(--swc-sources-toggle-padding, 3px) token("spacing-100");
+  padding: var(--swc-message-sources-toggle-padding, 3px) token("spacing-100");
```
5t3ph: Why were these two kept?

2nd-gen/packages/swc/patterns/ai-toolkit/response-status/response-status-step/ResponseStatusStep.ts
```diff
@@ -57,6 +57,9 @@ export type ResponseStatusStepStatus =
  * @fires swc-response-status-step-active-label-change - Dispatched when this
  * step's label text changes while `status="active"`, so the parent can keep
  * its header label in sync with streamed text. Detail: `{ label: string }`
+ *
+ * @cssprop --swc-response-status-step-shimmer-spread - Half-width of the shimmer highlight swept across the active step title. Defaults to `4ch`.
```
5t3ph: Reading this description, it seems like maybe premature to expose?

2nd-gen/packages/swc/patterns/ai-toolkit/suggestion-item/suggestion-item.css
```diff
@@ -61,14 +61,14 @@
...
-  margin-inline-end: var(--swc-suggestion-item-icon-margin-inline-end, 6px);
+  margin-inline-start: 14px;
+  margin-inline-end: token("spacing-85");
 }
 
 .swc-SuggestionItem-label {
   display: inline-flex;
-  padding-block: var(--swc-suggestion-item-label-padding-block, 7px);
-  padding-inline-end: var(--swc-suggestion-item-label-padding-inline-end, token("spacing-300"));
+  padding-block: 7px;
```
5t3ph: Could use:

```suggestion
  padding-block: token("base-padding-vertical-medium");
```

5t3ph (commented):
Thanks for taking on this clean-up!

2nd-gen/packages/swc/patterns/ai-toolkit/suggestion-item/suggestion-item.css
rubencarvalho: Done!

2nd-gen/packages/swc/patterns/ai-toolkit/response-status/response-status-step/ResponseStatusStep.ts
rubencarvalho: done!

2nd-gen/packages/swc/patterns/ai-toolkit/message-sources/message-sources.css
rubencarvalho: removed them.

5t3ph (approved):
:shipit:

blunteshwar (approved):

PR #6719, opened 2026-09-08
feat(core): labelling mixin

nikkimk:
## Description

Replaces the labelling half of the never-built `LabellingController` (planned in the forms-strategy RFC and the text-field migration plan) with `LabellingMixin`, a composable mixin in `2nd-gen/packages/core/mixins` (this is a companion PR to `nikkimk/textfield-help-text-mixin`, which does the same for description/error-text).

`LabellingMixin` owns visible-label rendering and precedence-ordered accessible-name wiring:

- `accessible-labelledby` (highest precedence) resolves space-separated element `id`s against the host's root node and writes them to the role element's `ariaLabelledByElements`.
- `accessible-label` (next) sets `aria-label` directly on the role element.
- A slotted `label` (lowest precedence) renders as a real, same-root `<label for>`, giving native click-to-focus.

Each source is wired independently; when more than one is set, the browser's [accessible-name computation](https://www.w3.org/TR/accname-1.2/) decides which is announced (accessible-labelledby > accessible-label > slotted label). A slotted visible label always renders as a real <label for>, so it stays clickable even when a higher-precedence programmatic name is announced. The mixin owns three dev-mode warnings: no accessible name; an accessible-labelledby idref that resolves to nothing; and a conflicting accessible-labelledby + accessible-label.

Rendering is delegated to a new stateless directive, `renderFieldLabel` (`2nd-gen/packages/core/directives/render-label`), mirroring the existing `PendingController` → `renderPendingSpinner` split. The mixin exposes an overridable `roleElement` getter (defaulting to `null`) so it can be applied at the core-base-class layer before a render layer exists, with the concrete render class supplying the real element (same pattern as `TooltipBase.tipElement`).

`swc-text-field` is refactored to consume `LabellingMixin` as the proving-ground component: `TextField.base.ts` now composes `SizedMixin(LabellingMixin(SpectrumElement))` (dropping the old `accessible-label`/`accessible-labelledby` property stubs and their `@todo (SWC-2466)` comments), and `TextField.ts` renders the label around a functional `<input>`, with a `roleElement` override pointing at that input. `accessible-describedby` stays as the original inert stub property; description/error-text rendering is out of scope here (see the companion PR).

**New:**
- `core/mixins/labelling-mixin.ts` (+ unit tests, Storybook demo stories, and a per-mixin `.mdx` doc — a first for mixins in this repo, following the controller-doc conventions)
- `core/directives/render-label/`
- `swc/components/text-field/text-field.mdx`, `swc/components/text-field/test/text-field.test.ts`

**Changed:**
- `core/components/text-field/TextField.base.ts`, `swc/components/text-field/TextField.ts`, `text-field.css`, `stories/text-field.stories.ts` (added Anatomy/Options → Labelling/States/Accessibility stories)
- `core/mixins/index.ts`, `core/directives/index.ts`, `core/package.json` (export entries for the new directive)

**Explicitly out of scope** (separate work, tracked elsewhere): description/error-text association (`nikkimk/textfield-help-text-mixin`), `FieldAssociationController` (SWC-2467, form participation), and the shared `.swc-FormFieldTemplate` CSS grid / `label-position` layout (Q22).

## Motivation and context

The migration plan called for a single `LabellingController` to own in-shadow label/description/error rendering for form fields, following the `LinearProgressMixin` precedent of "prove it on the first consumer, generalize later." Splitting the controller's two genuinely independent responsibilities (naming vs. describing) into two mixins, delivered as two separate PRs, keeps each change small and independently reviewable, and lets future field components (`text-area`, `number-field`, `color-field`) adopt only the one they need.

## Related issue(s)

- Addresses the labelling half of SWC-2466 (`LabellingController`) — see epic SWC-2323. Companion PR: `nikkimk/textfield-help-text-mixin`.

N/A — see the new Storybook stories under **Text field** (Anatomy, Options → Labelling, States, Accessibility) and **Mixins → Labelling mixin** for visual coverage.

## Author's checklist

## Reviewer's checklist

### Manual review test cases

- [ ] _Label precedence resolves correctly_
  1. Go to **Text field → Options → Labelling** in Storybook.
  2. Inspect the three fields: slotted-label-only, `accessible-label`-only, and `accessible-labelledby` (pointing at the row/column header `div`s).
  3. Expect: field 1 renders a real `<label for>` in its shadow root; field 2 has `aria-label` on the input and no `<label>`; field 3 has no `aria-label`/`<label for>` and `input.ariaLabelledByElements` resolves to the two header elements.

- [ ] _Missing accessible name warns in dev mode_
  1. Open the browser console with Storybook running in dev mode.
  2. Navigate to **Text field → Tests → MissingAccessibleNameTest**.
  3. Expect: a console warning mentioning "accessible name."

- [ ] _Mixin works standalone, outside text-field_
  1. Go to **Mixins → Labelling mixin**.
  2. Review the "Name source precedence" behavior story.
  3. Expect: behavior matches the description in the mixin's `.mdx` page.

### Device review

- [ ] Did it pass in Desktop?

  1. Go to **Text field → Options → Labelling** with a screen reader off, keyboard only.
  2. Tab to the slotted-label field, then click its visible label text.
  3. Expect: clicking the label moves focus into the associated `<input>` (native `<label for>` behavior — no custom click handler involved).
  4. Tab through all fields in **States**; confirm the disabled field is skipped in tab order and the read-only field remains focusable.

  1. With VoiceOver (or NVDA) running, focus each field in **Text field → Options → Labelling**.
  2. Expect: field 1 announces "Slotted visible label, text field, edit text"; field 2 announces "Accessible-label only (no visible label), text field"; field 3 announces the concatenated row + column header text as the name.


2nd-gen/packages/core/mixins/labelling-mixin.ts
```diff
@@ -0,0 +1,323 @@
...
+     * @internal
+     */
+    public get hasLabelSlotContent(): boolean {
+      return this._labelSlotPresence.isPresent;
+    }
+
+    /**
+     * @internal
+     */
+    public get roleElement(): Element | null {
```
nikkimk, line 105: If `roleElement` were protected, `TextField.ts` (over in the swc package) wouldn't be allowed to override it anymore. So it stays public, just documented as internal-only. `LinearProgressMixin` does the same thing for its own internal getters, for the same reason — this isn't a one-off.

2nd-gen/packages/core/mixins/labelling-mixin.ts
```diff
@@ -0,0 +1,323 @@
...
+        present.push('labelledby');
+      }
+      if (this.accessibleLabel) {
+        present.push('label');
+      }
+      if (this.hasLabelSlotContent) {
+        present.push('slot');
+      }
+      return present;
+    }
```
nikkimk: `_presentNameSources` pushes things in precedence order (`labelledby` first, then `label`, then slot) so `_activeNameSource` can just grab [0] instead of re-checking the precedence logic a second time.

2nd-gen/packages/core/mixins/labelling-mixin.ts
```diff
@@ -0,0 +1,323 @@
...
+     * Warns when more than one accessible-name source is set at once (some
+     * combination of a resolved `accessibleLabelledby`, `accessibleLabel`,
+     * and a slotted `label`). Only the highest-precedence source present is
+     * ever read as the accessible name; the rest are silently ignored, which
+     * risks a WCAG 2.5.3 (Label in Name) mismatch between what's shown and
+     * what's announced if their text differs. The message names which
+     * source wins and which will not be read, so the warning is actionable
+     * regardless of which combination triggered it.
+     */
+    private _warnLabelConflict(): void {
```
nikkimk, line 310: Heads up: this warns any time two or more label sources are set, even if their text happens to be identical. This won't compare the actual strings because it would mean dealing with whitespace/casing normalization, and honestly, setting two sources that happen to agree today is still fragile (one could change later and silently drift). But wanted to call it out in case anyone expects the warning to only fire on a real mismatch.

2nd-gen/packages/core/mixins/stories/labelling-mixin.stories.ts
```diff
@@ -0,0 +1,171 @@
...
+      // (`ariaLabelledByElements`), which is how `LabellingMixin` resolves
+      // `accessible-labelledby`. It only inspects attributes, so it reports a
+      // false "Form element does not have a label" ("label" rule) violation
+      // on this host even though the name resolves correctly in real
+      // browsers and assistive technology. See the forms-strategy RFC's
+      // axe-core policy (CONTRIBUTOR-DOCS, "3.4 axe-core policy") for the
+      // documented false-positive list. Remove once axe-core adds ARIAMixin
+      // element-reference support (review quarterly).
+      exclude: {
+        label: ['#labelling-mixin-labelledby-host'],
```
nikkimk, line 111: Because it's a cross root ARIA solution, `accessible-labelledby` doesn't set the plain `aria-labelledby` attribute. It sets `ariaLabelledByElements`, which is a newer JS-only ARIA property. `axe-core` can't see it, so it thinks the field has no label at all. 

That's a false alarm, not a real bug, but it's why every story using `accessible-labelledby` needed a `parameters.a11y.exclude` added. This is also the first time we've used that exclude mechanism anywhere in the repo, so I'd love a second opinion that this is the right way to handle it going forward, not just a rubber stamp because it's already there.

2nd-gen/packages/core/mixins/labelling-mixin.ts
```diff
@@ -0,0 +1,323 @@
...
+ * writes the resolved accessible name onto (for example, the `<input>` inside
+ * a text field's own shadow root).
+ */
+type LabelledByTarget = Element & {
+  ariaLabelledByElements: Element[] | null;
+};
+
+/** The API {@link LabellingMixin} adds to its host. */
+export interface LabellingInterface {
+  accessibleLabel?: string;
```
nikkimk, line 37: Small inconsistency worth a look: the interface says `accessibleLabel?: string` (optional), but the actual mixin always sets it to '' by default — it's never really undefined. I made it optional mainly to match its neighbor, `accessibleLabelledby?: string`, for anyone implementing the interface from scratch. Just flagging that the type is a bit looser than what the real implementation guarantees.

2nd-gen/packages/swc/components/text-field/TextField.ts
```diff
@@ -17,6 +17,18 @@ import { TextFieldBase } from '@adobe/spectrum-wc-core/components/text-field';
...
+ * The `<input>`'s `id`, referenced by the rendered `<label for>`. A fixed
+ * string (not a per-instance counter) is safe here: the `id` only has to be
+ * unique within this component's own shadow root, where exactly one `<input>`
+ * ever exists, and shadow roots are their own `id` scope (same-named `id`s in
+ * other instances' shadow roots never collide). A per-instance counter would
+ * instead depend on module-level mutable state and instantiation order,
+ * which a server render and a client hydration pass are not guaranteed to
+ * agree on.
+ */
+const INPUT_ID = 'input';
```
nikkimk, line 25: This is different from how we do it elsewhere (LinearProgressMixin uses an incrementing counter for its ids), so it might look like an oversight rather than a choice. 

It's intentional: this id only ever needs to be unique inside the field's own shadow root, and there's only ever one `<input>` there. Using a fixed string instead of a counter also means the id would stay identical if we ever add server rendering later. (A counter could produce different ids on the server vs. the client and break hydration.) 

Calling it out since it doesn't match the existing pattern, and I want that to read as "on purpose," not "missed it."

nikkimk (commented):
Just so it's clear while reviewing: description/error text (HelpTextMixin) isn't in here at all. It's a separate PR so the two can be reviewed on their own. If something about labels-only seems incomplete, that's why.

cdransf (approved):
Looks great! ✨

2nd-gen/packages/core/mixins/labelling-mixin.ts
```diff
@@ -0,0 +1,323 @@
...
+  /**
+   * Renders the visible label for the given role-element `id`. Pass `null`
+   * when the role element has no stable `id` yet.
+   */
+  renderLabel(forId: string | null): RenderFieldLabelResult;
+}
+
+/**
+ * A mixin that adds visible-label rendering and accessible-name wiring to a
+ * host: the reactive `accessible-label` / `accessible-labelledby` properties,
```
miwha-adobe: I'm surprised it didn't catch the em dash, I thought we included in a skill to remove those. Additionally this comment is very verbose. Can we make it a bit more concise?

2nd-gen/packages/core/mixins/labelling-mixin.ts
```diff
@@ -0,0 +1,323 @@
...
+     * The currently active accessible-name source, in precedence order.
+     */
+    private get _activeNameSource(): 'labelledby' | 'label' | 'slot' | 'none' {
+      return this._presentNameSources[0] ?? 'none';
+    }
+
+    public renderLabel(forId: string | null): RenderFieldLabelResult {
+      return renderFieldLabel({
+        hasLabelSlotContent: this.hasLabelSlotContent,
+        forId: this._activeNameSource === 'slot' ? forId : null,
```
Rajdeepc: **Design question:**
I think the `<span>` fallback can be deleted. It removes click-to-focus, and the accessible-name protection it's meant to provide already happens automatically in the browser. I want to check whether there's a reason I'm not seeing before we bake this into the labelling contract every future field component inherits.

The stated reason in `render-label.ts:23` is that the `<span>` keeps the text visible while excluding it from the accessible-name computation, so it can't create a conflicting second name.

I want to push on this part: Consider a field with a visible label reading "Query" and `accessible-label=
"Search Products", it will announce as Search Products but the visible label is Query.

This violation lives in the mismatch between the two strings not in which HTML wraps the visible one and we pay for two visually identical fields behaving differently on click with no clue. Also lit tears down and rebuilds subtree when crossing the boundary and voice control users will loose a `label for` association.

My proposal would be to render `label for` unconditionally whenever there is  a slotted content or I will change my mind if a screen reader call that concatenates rather than override or a rule that an overriden label must not be clickable.

2nd-gen/packages/core/mixins/labelling-mixin.ts
```diff
@@ -0,0 +1,323 @@
...
+      if (!this.accessibleLabelledby) {
+        return [];
+      }
+      const root = this.getRootNode() as Document | ShadowRoot;
+      return this.accessibleLabelledby
+        .split(/\s+/)
+        .filter(Boolean)
+        .map((id) => root.getElementById(id))
+        .filter((element): element is HTMLElement => element !== null);
+    }
```
Rajdeepc, line 134: This is a crash path. This cast assets something `getRootNode()` doesn't guarantee. On a detached element it returns the topmost ancestor node which for a lone removed element is the element itself, with no `getElementById`. 
Lit's performUpdate has no isConnected guard so once updating is enabled, a pending update flushes after disconnect. f`ield.accessibleLabelledby = 'x'; field.remove(); in one tick reaches `root.getElementById is not a function`.
It is narrow but very cheap to close. You can try this 
```ts
const root = this.getRootNode();
if (!(root instanceof Document || root instanceof ShadowRoot)) return [];
```

2nd-gen/packages/swc/components/text-field/TextField.ts
```diff
@@ -35,15 +47,35 @@ export class TextField extends TextFieldBase {
...
+    // icon, and description/error container via a future help-text mixin.
     return html`
       <div class="swc-TextField">
+        ${this.renderLabel(INPUT_ID)}
         <input
+          id=${INPUT_ID}
           class="input"
-          aria-label=${ifDefined(this.accessibleLabel || undefined)}
+          type=${this.type}
+          .value=${this.value}
```
Rajdeepc, line 76: I dont see any `@input/@change` handler here so `field.value` will read `' '` after the user types. Your docs is telling consumers the input will carry the value. This will drift like this, but please do validate on your end too
set `value="a"`, user types `'b'`, set `value="a"` again with no re-commit input will still show `'b'`.
`@todo (SWC-2467)` covers form participation but I dont see this is convering value sync. You can either add the handler now or drop `.value` from the render and an explicit `@todo` for this.

2nd-gen/packages/core/mixins/labelling-mixin.ts
```diff
@@ -0,0 +1,323 @@
...
+    private get _resolvedLabelledbyElements(): Element[] {
+      if (!this.accessibleLabelledby) {
+        return [];
+      }
+      const root = this.getRootNode() as Document | ShadowRoot;
+      return this.accessibleLabelledby
+        .split(/\s+/)
+        .filter(Boolean)
+        .map((id) => root.getElementById(id))
+        .filter((element): element is HTMLElement => element !== null);
```
Rajdeepc, line 133: This will cause silent IDREF failure. This is a known axe blind spot. Can you warn when `accessibleLabelledby` is non-empty but some `id` resolves to nothing. Also `ids` resolve at `updated()` time now if a label element is appended after the first update it will never resolve and nothing will mark the host dirty.
If this is a pattern I would recommend the ordering requirement in the docs.

2nd-gen/packages/swc/components/text-field/test/text-field.test.ts
```diff
@@ -0,0 +1,227 @@
...
+    const fields = await getComponents<TextField>(
+      canvasElement,
+      'swc-text-field'
+    );
+
+    await step('required is reflected onto the native input', () => {
+      const requiredField = fields[1];
+      const requiredInput = requiredField.shadowRoot?.querySelector('input');
+      expect(requiredInput?.required).toBe(true);
+    });
```
Rajdeepc, line 98: Expecting more tests here on `disabled`, `readonly`, `value` and `aria-invalid` bindings.

Rajdeepc (commented):
I like the shape and the design. Self review comments were really helpful. One design question and one crash path, rest looks nice!

2nd-gen/packages/core/mixins/labelling-mixin.ts
```diff
@@ -0,0 +1,323 @@
...
+     */
+    private get _resolvedLabelledbyElements(): Element[] {
+      if (!this.accessibleLabelledby) {
+        return [];
+      }
+      const root = this.getRootNode() as Document | ShadowRoot;
+      return this.accessibleLabelledby
+        .split(/\s+/)
+        .filter(Boolean)
+        .map((id) => root.getElementById(id))
```
rubencarvalho: Possible SSR concern: this getter runs during `render()` (via `renderLabel` → `_activeNameSource`) whenever `accessible-labelledby` is set, and `root.getElementById(id)` isn't guarded. On the server DOM shim that method may be absent, which would throw during SSR — seemingly the same class of issue the earlier SSR fix commits were addressing for the input `id`. The codebase already uses a guarded form in `resolve-trigger.ts` (`root.getElementById?.(id) ?? null`); mirroring it here would keep this path SSR-safe. Could you confirm whether an SSR render with `accessible-labelledby` set is exercised anywhere?

2nd-gen/packages/core/mixins/labelling-mixin.ts
```diff
@@ -0,0 +1,323 @@
...
+      }
+    }
+
+    private _syncLabelling(): void {
+      const target = this.roleElement as LabelledByTarget | null;
+      if (!target) {
+        return;
+      }
+      const source = this._activeNameSource;
+      target.ariaLabelledByElements =
```
rubencarvalho: `accessible-labelledby` is resolved to concrete element references only at render/`updated()` time, and nothing observes the referenced *external* element. If a target `#id` is inserted after the first render (async or framework-driven), or removed later, there's no property change or host-child mutation to re-trigger `updated()`, so `ariaLabelledByElements` may never sync (or go stale). Native `aria-labelledby` is a live IDREF and wouldn't have this gap. Is post-render appearance of the label target an intended supported case? If so it might be worth a note in the docs or a small re-resolution hook; if not, no action needed.

2nd-gen/packages/core/mixins/labelling-mixin.ts
```diff
@@ -0,0 +1,323 @@
...
+        source === 'labelledby' ? this._resolvedLabelledbyElements : null;
+      if (source === 'label') {
+        target.setAttribute('aria-label', this.accessibleLabel);
+      } else {
+        target.removeAttribute('aria-label');
+      }
+    }
+
+    private _warnMissingAccessibleName(): void {
```
rubencarvalho: Minor, dev-only: this warning evaluates on the first `updated()`, so a valid host that receives its name source imperatively right after append (e.g. `el.accessibleLabel = 'Name'` before the property-set render lands) can emit a spurious “requires an accessible name” once. The dedup prevents repeats but can't retract the first message. Probably fine given most consumers set attributes before insertion — flagging in case the false positive shows up in tests.

2nd-gen/packages/swc/components/text-field/TextField.ts
```diff
@@ -35,6 +47,14 @@ export class TextField extends TextFieldBase {
     return [styles];
   }
 
+  /**
+   * The real role element `LabellingMixin` wires the resolved accessible-name
+   * ARIA relationship onto.
+   */
+  public override get roleElement(): HTMLInputElement | null {
+    return this.renderRoot.querySelector('input.input');
```
rubencarvalho: Small robustness suggestion: `roleElement` queries by the `input` class, while the label wiring (`renderLabel(INPUT_ID)` → `<label for="input">`) anchors on the `id`. If a later styling change renames or drops `class="input"` with the `id` untouched, `roleElement` would return `null` and `_syncLabelling` would silently no-op (aria-label / ariaLabelledByElements stop applying, no error). Querying the same anchor the label uses (`input#input`) would keep the two in lockstep. Not urgent — just removes a quiet failure mode.

2nd-gen/packages/core/mixins/labelling-mixin.ts
rubencarvalho: maybe `root.getElementById?.(id)` so we're ssr safe - I saw your other commits to fix these cases

2nd-gen/packages/core/mixins/labelling-mixin.ts
rubencarvalho: FYI, no real action required here: `accessible-labelledby` is resolved once at render so it won't re-sync if it appears/disappears later. I remembered this because we had the same in the popover `for`. if we ever want to solve for it (mutationobserver), we can tackle it everywhere

2nd-gen/packages/core/mixins/labelling-mixin.ts
rubencarvalho: Dev-only nit: a name set imperatively right after append can emit a spurious "requires an accessible name" once (dedup can't retract it).

2nd-gen/packages/swc/components/text-field/TextField.ts
rubencarvalho: hmm, a small nit, `roleElement` queries by class but the label does it by id... maybe `input#input`?

2nd-gen/packages/core/mixins/labelling-mixin.ts
5t3ph: Is there a reason for an array, or could you simplify to `else if` to ultimately only return one option?

2nd-gen/packages/core/mixins/labelling-mixin.ts
```diff
@@ -0,0 +1,323 @@
...
+
+  /**
+   * Renders the visible label for the given role-element `id`. Pass `null`
+   * when the role element has no stable `id` yet.
+   */
+  renderLabel(forId: string | null): RenderFieldLabelResult;
+}
+
+/**
+ * A mixin that adds visible-label rendering and accessible-name wiring to a
```
5t3ph: [nit] comments can be reduced, anything critical for documentation should move to/live in the MDX

2nd-gen/packages/core/mixins/labelling-mixin.ts
5t3ph: +1 to keeping a visible proper label regardless of other labelling methods. To my understanding/how I've been guided to do this before, this would be on the consumer to have for example visible label "Query" and accessible-label "Query products" which would meet the noted criterion.

2nd-gen/packages/core/mixins/labelling-mixin.ts
5t3ph: This will need revised per earlier comment from @Rajdeepc and myself on ensuring a visible label + other method is allowed.

2nd-gen/packages/core/mixins/stories/labelling-mixin.stories.ts
```diff
@@ -0,0 +1,171 @@
...
+
+import { html } from 'lit';
+import type { Meta, StoryObj } from '@storybook/web-components';
+
+import './demo-hosts.js';
+
+// ────────────────
+//    METADATA
+// ────────────────
+
```
5t3ph, line 21: There's a decent amount of duplication in these docs that could be condensed.

2nd-gen/packages/core/mixins/labelling-mixin.mdx
```diff
@@ -0,0 +1,112 @@
...
+`LabellingMixin` adds visible-label rendering and accessible-name wiring to a host so components do not hand-roll it. It is one of two decoupled primitives for form-field labelling:
+
+- **`LabellingMixin`** (`@adobe/spectrum-wc-core/mixins`) — the `accessible-label` / `accessible-labelledby` properties, `label` slot presence tracking, the "no accessible name" dev warning, and `renderLabel()`.
+- **`renderFieldLabel`** (`@adobe/spectrum-wc-core/directives/render-label`) — the underlying render-only directive. The mixin owns and exposes it; use it directly only for stateless label rendering without the mixin.
+
+`HelpTextMixin` (`@adobe/spectrum-wc-core/mixins`) is the companion mixin for description/error-text association; `swc-text-field` composes both.
+
+### What it does
+
+- **Precedence-ordered accessible name** — only the highest-precedence source that is actually set is wired, in this order:
```
5t3ph: I wasn't sure where this precedence was coming from, but here is the W3 source of "[Accessible Name and Description Computation 1.2](https://www.w3.org/TR/accname-1.2/)" - can we include this source in this documentation?

I also learned that AT depend on the browser pre-computing this in the case multiple sources are set, so even if we didn't "force" the order, browsers would also use this precedence to determine the final used `accname`.

2nd-gen/packages/swc/components/text-field/text-field.css
```diff
@@ -11,9 +11,16 @@
...
-   .swc-TextField, not :host. Style disabled via `:host(:state(disabled))`
-   (covers own + cascaded disabled), not `[disabled]`. */
+   consume the shared form-fields stylesheet fragment (including the
+   label-position="side" grid variant, Q22). Keep styling on .swc-TextField,
+   not :host. Style disabled via `:host(:state(disabled))` (covers own +
+   cascaded disabled), not `[disabled]`. */
 .swc-TextField {
   display: inline-flex;
+  flex-direction: column;
+  gap: var(--swc-spacing-75, 4px);
```
5t3ph: [nit] since this is going to move to the shared styles and be refined but reminder to use `token()`

```suggestion
  gap: token("spacing-75");
```

2nd-gen/packages/swc/components/text-field/text-field.mdx
```diff
@@ -0,0 +1,67 @@
...
+- **One accessible-name writer** — `accessible-labelledby` &gt; `accessible-label` &gt; slotted label, in that order; exactly one source is wired at a time.
+- **Dev-mode warnings** — a field with no visible label, no `accessible-label`, and no `accessible-labelledby` triggers a console warning; so does a field that sets more than one of `accessible-labelledby`, `accessible-label`, and a slotted label, naming which source wins and which will not be read.
+
+> **Known axe-core false positive.** A field named only via `accessible-labelledby` fails automated axe-core scans with a "Form element does not have a label" (`label` rule) violation, even though the name resolves correctly in real browsers and assistive technology — axe-core inspects DOM attributes and has no standardized way to read `ariaLabelledByElements`, the ARIA element-reflection API `LabellingMixin` uses. The affected stories in this file (`Labelling`, `ConflictingLabelSources`, `Accessibility`) exclude the specific host from the `label` rule with a written `// reason:` comment, per the forms-strategy RFC's axe-core policy (CONTRIBUTOR-DOCS "3.4 axe-core policy"), rather than disabling the rule globally.
```
5t3ph: Since this is public facing, can remove the parts that are specific to our implementation:

```suggestion
> **Known axe-core false positive.** A field named only via `accessible-labelledby` fails automated axe-core scans with a "Form element does not have a label" (`label` rule) violation, even though the name resolves correctly in real browsers and assistive technology. axe-core inspects DOM attributes and has no standardized way to read `ariaLabelledByElements`, the ARIA element-reflection API in use.
```

5t3ph (changes requested):
⚠️  **Near-term conflict:** `radio-group`'s migration plan is built around `LabellingController` being capable of host-level (`ElementInternals`) ARIA wiring, since the group has no inner control to point at. `LabellingMixin` as written only wires onto a child element via `roleElement` (defaults to `null` → no-op), so it can't cover that case as-is. Other potential consumers needing host attachment are checkbox-group and the possible field-group. Not a blocker for text-field, but we should either add a host-attachment mode before radio-group starts implementation or update its plan to keep hand-rolling this.

2nd-gen/packages/core/mixins/labelling-mixin.ts
miwha-adobe: Agreed, and fixed in [3eaa783](https://github.com/adobe/spectrum-web-components/pull/6719/changes/3eaa783c39241da6ab91661585c032f5e9ba2887). I removed the hidden-<span> workaround. The visible label is now always a real, clickable label, so clicking it still focuses the field for everyone, including voice-control users. When someone also sets a screen-reader name, the browser decides which name to announce, and the visible label stays clickable either way. So "Query" as the visible label plus "Query products" as the screen-reader name works exactly like @5t3ph described. Thanks for pushing on this.

2nd-gen/packages/core/mixins/labelling-mixin.ts
miwha-adobe: Reworked in [3eaa783.](https://github.com/adobe/spectrum-web-components/pull/6719/changes/3eaa783c39241da6ab91661585c032f5e9ba2887) The old warning was a bit too noisy and fired any time two label sources were set at all, which meant it complained even when you did the right thing: a visible label plus a screen-reader name (for example, visible "Query" with accessible-label="Query products"). That combination is actually the recommended pattern.

Now it only warns in the one case that's genuinely a mistake: setting both accessible-label and accessible-labelledby at the same time. Those are two different ways to give the field a screen-reader name, so one of them always loses. When that happens, the warning tells you accessible-labelledby is the one that wins and accessible-label is ignored.

And to keep @nikkimk's original call: it still doesn't compare the actual text of the two names, since that's the author's responsibility, and the browser already picks the winner the same way every time.

2nd-gen/packages/core/mixins/labelling-mixin.ts
miwha-adobe: Good call, and done in [3eaa783](https://github.com/adobe/spectrum-web-components/pull/6719/changes/3eaa783c39241da6ab91661585c032f5e9ba2887). You were right that the array was doing more than it needed to, and it actually went a step past the else-if. Now that the browser decides which name wins, nothing in the code needs to know the "active" source at all, so I could delete the whole thing.

For context on what was there: it built a list of every label source that was set, in priority order, and two things used it. One picked the winner (the first item), and the other was the old conflict warning, which needed the full list to say "two or more are set" and name the ones being ignored. Once the browser owns the priority and the warning shrank to a simple "are both screen-reader names set?" check, both of those uses went away.

What's left is a single _hasAccessibleName boolean (does the field have any name at all?), used only for the "this field has no name" warning.

miwha-adobe: Good call, and done in [3eaa783](https://github.com/adobe/spectrum-web-components/pull/6719/changes/3eaa783c39241da6ab91661585c032f5e9ba2887). You were right that the array was doing more than it needed to, and it actually went a step past the else-if. Now that the browser decides which name wins, nothing in the code needs to know the "active" source at all, so I could delete the whole thing.

For context on what was there for anyone reading: it built a list of every label source that was set, in priority order, and two things used it. One picked the winner (the first item), and the other was the old conflict warning, which needed the full list to say "two or more are set" and name the ones being ignored. Once the browser owns the priority and the warning shrank to a simple "are both screen-reader names set?" check, both of those uses went away.

What's left is a single _hasAccessibleName boolean (does the field have any name at all?), used only for the "this field has no name" warning.

2nd-gen/packages/core/mixins/labelling-mixin.ts
miwha-adobe: Reworked in [3eaa783.](https://github.com/adobe/spectrum-web-components/pull/6719/changes/3eaa783c39241da6ab91661585c032f5e9ba2887) The old warning was a bit too noisy and fired any time two label sources were set at all, which meant it complained even when you did the right thing: a visible label plus a screen-reader name (for example, visible "Query" with accessible-label="Query products"). That combination is actually the recommended pattern.

Now it only warns in the one case that's genuinely a mistake: setting both accessible-label and accessible-labelledby at the same time. Those are two different ways to give the field a screen-reader name, so one of them always loses. When that happens, the warning tells you accessible-labelledby is the one that wins and accessible-label is ignored.

And to keep @nikkimk's original call: it still doesn't compare the actual text of the two names, since that's the author's responsibility, and the browser already picks the winner the same way every time.

2nd-gen/packages/core/mixins/labelling-mixin.ts
miwha-adobe: Agreed, and fixed in [3eaa783](https://github.com/adobe/spectrum-web-components/pull/6719/changes/3eaa783c39241da6ab91661585c032f5e9ba2887). I removed the hidden- workaround. The visible label is now always a real, clickable label, so clicking it still focuses the field for everyone, including voice-control users. When someone also sets a screen-reader name, the browser decides which name to announce, and the visible label stays clickable either way. So "Query" as the visible label plus "Query products" as the screen-reader name works exactly like @5t3ph described. Thanks for pushing on this.

2nd-gen/packages/core/mixins/labelling-mixin.ts
miwha-adobe: Reduced verbosity in [3eaa783.](https://github.com/adobe/spectrum-web-components/pull/6719/changes/3eaa783c39241da6ab91661585c032f5e9ba2887)

2nd-gen/packages/core/mixins/labelling-mixin.ts
miwha-adobe: Agreed!

2nd-gen/packages/core/mixins/labelling-mixin.mdx
miwha-adobe: Screenshot 2026-09-10 at 1 30 54 PM

2nd-gen/packages/core/mixins/stories/labelling-mixin.stories.ts
miwha-adobe: Condensed in [d2fa10c](https://github.com/adobe/spectrum-web-components/pull/6719/changes/d2fa10c8501111583eba86f7696eda269b37f53f). Tightened the overlapping prose and updated the conflicting-sources demo so it only shows real conflicts. There's still room for a deeper pass, but the worst of the duplication and the now-stale copy is gone.

2nd-gen/packages/swc/components/text-field/text-field.mdx
miwha-adobe: Applied your wording in [d2fa10c](https://github.com/adobe/spectrum-web-components/pull/6719/changes/d2fa10c8501111583eba86f7696eda269b37f53f)

2nd-gen/packages/swc/components/text-field/text-field.css
miwha-adobe: Fixed!

miwha-adobe:
> ⚠️ Near-term conflict: radio-group's migration plan is built around LabellingController being capable of host-level (ElementInternals) ARIA wiring, since the group has no inner control to point at. LabellingMixin as written only wires onto a child element via roleElement (defaults to null → no-op), so it can't cover that case as-is. Other potential consumers needing host attachment are checkbox-group and the possible field-group. Not a blocker for text-field, but we should either add a host-attachment mode before radio-group starts implementation or update its plan to keep hand-rolling this.

Good point @5t3ph! Would you like me to address this as a follow up, or is that something radio-group can expand on?

2nd-gen/packages/core/mixins/labelling-mixin.ts
miwha-adobe: I used the optional-chaining form `(root.getElementById?.(id) ?? null)` to match the existing guard in `utils/resolve-trigger.ts` rather than a separate instanceof check. Returns the same result!

2nd-gen/packages/core/mixins/labelling-mixin.ts
miwha-adobe: Done! https://github.com/adobe/spectrum-web-components/pull/6719/changes/1cf0b56f9698eda6f90a81fbc20fb4490b5d9173

2nd-gen/packages/core/mixins/labelling-mixin.ts
miwha-adobe: https://github.com/adobe/spectrum-web-components/pull/6719/changes/1cf0b56f9698eda6f90a81fbc20fb4490b5d9173

2nd-gen/packages/swc/components/text-field/TextField.ts
miwha-adobe: This was handled in the FormAssociationController, and should now be present after merging in the feature branch.

2nd-gen/packages/swc/components/text-field/test/text-field.test.ts
miwha-adobe: Added. New coverage for readonly reflection, aria-invalid on invalid, and the value round-trip. Kept it to the bindings that weren't tested anywhere: disabled is already exercised by the disabled-state/form-participation tests, and constraint validity (required/pattern/checkValidity) is covered one layer down in the FieldAssociationController tests. While I was in here I also added focus delegation (host focus() → native input) and the type/label-position enum-validation warnings.

https://github.com/adobe/spectrum-web-components/pull/6719/changes/4510bc19b42980b37b2184fe57421bd1781050ed

2nd-gen/packages/core/mixins/labelling-mixin.ts
miwha-adobe: I can definitely add that

miwha-adobe: https://github.com/adobe/spectrum-web-components/pull/6719/changes/29249e9d0077d026f7dea56e774a6351625335cc

2nd-gen/packages/core/mixins/labelling-mixin.ts
miwha-adobe: Oh yes, that would be nice to have.

5t3ph:
> > ⚠️ Near-term conflict: radio-group's migration plan is built around LabellingController being capable of host-level (ElementInternals) ARIA wiring, since the group has no inner control to point at...
> 
> Good point @5t3ph! Would you like me to address this as a follow up, or is that something radio-group can expand on?

Claude assures me it's "mostly an additive `labelInternals` hook alongside `roleElement` and a small branch in `_syncLabelling`" to accomplish but up to you if you want to defer for me to pick up with radio group!

5t3ph (commented):
It occurred to me that perhaps the mixin should also go ahead and handle `placeholder` since it's unfortunately common that folks try to add _only_ a placeholder without a proper label, and placeholder isn't reliably used in the accName computation.

> Cleanest fit is another optional override hook, same pattern as `roleElement`, e.g. `protected get placeholderText(): string | undefined` defaulting to `undefined`, that TextField etc. overrides to expose its own placeholder value.

Along with the dev mode warning, ex. `_warnPlaceholderOnlyName`, when _only_ placeholder is provided.

5t3ph:
Perhaps we should also warn against `aria-label` / `aria-labelledby` on the host since those will not successfully apply?

miwha-adobe:
Thanks Steph! I went ahead and added the labelInternals hook now (defaults to null, and _syncLabelling wires the name onto the host's internals when it's set)m so radio-group just overrides it to return its internals. Left the concrete override + group label rendering for that work since it's easier to shape against the real component.

Also changes for the warnings are included here: https://github.com/adobe/spectrum-web-components/pull/6719/changes/22d9511c20948fd2eeb7980ef31b35b2a6d3fb47

5t3ph (approved):
LGTM, great teamwork! 💜

PR #6722, opened 2026-09-09
feat(menu): implement accessibility for menu

Rajdeepc:
## Description

Implements accessibility for `swc-menu`, on top of the API migration in #6709:

- `role="menu"` on the shadow-internal surface (not on the `<swc-menu>` host)
- `FocusgroupNavigationController` wired for roving-tabindex/arrow-key/Home/End navigation (`direction: 'vertical'`, `wrap: true`)
- Focus moves into the first item on open (`queueMicrotask`-deferred); returns to the trigger on close if it was inside the menu
- Escape closes via `registerDismissible`/`isTopDismissible`, matching `Tooltip.base.ts`/`Popover.base.ts`
- `surfaceElement` now returns the real `.swc-Menu` node with a `?hidden=${!open}` binding, so the surface actually shows/hides — pulled forward from Phase 5 per reviewer feedback (Ruben), since focus-management correctness can't be genuinely verified against a no-op surface
- `menu.css` gets only the unstyled positioning/layout this needs: `position: absolute` for `PlacementController`, and `display: flex` / `flex-direction: column` so slotted rows stack vertically instead of relying on their own (unpredictable) default display. **No visual styling** (background, border, radius, elevation) — `swc-menu-item` doesn't exist yet, so there's nothing real to size or style against; that comes back once it does
- Storybook Playground: a real `swc-button` trigger wired via `for`, and placeholder rows using `swc-menu-item`'s own tag name with `role="menuitem"` (that component doesn't exist yet) so the flow is actually clickable and testable

## Motivation and context

Phase 4 (accessibility) of the `swc-menu` 2nd-gen migration, plus the minimal positioning/layout slice of Phase 5 needed to make it testable.

## Related issue(s)

- fixes SWC-1985

N/A — unstyled positioning only, no visual design. `swc-menu-item` doesn't exist yet, so surface chrome (background/border/radius/shadow) is deferred until there's real content to style against.

---

## Author's checklist

---

## Reviewer's checklist

### Manual review test cases

- [ ] _Trigger click opens/closes the menu, wires ARIA, and toggles visibility_
  1. Storybook → **Menu** → **Playground**, inspect the trigger and `<swc-menu>` in DevTools.
  2. Click the trigger twice.
  3. Expect: on open, an unstyled surface (no background/border/shadow yet) appears under the trigger with rows stacked vertically, `<swc-menu>` gains `open`, and the trigger gains `aria-expanded="true"`/`aria-haspopup="menu"`; on close, it disappears and both attributes revert.

- [ ] _Arrow keys, Home, End, Escape navigate the placeholder rows_
  1. Same story. Open the menu, press <kbd>ArrowDown</kbd> x3, <kbd>Home</kbd>, <kbd>End</kbd>, <kbd>Escape</kbd>.
  2. Expect: focus starts on "Cut", wraps Paste→Cut, Home/End jump to first/last, Escape closes and returns focus to the trigger.

### Device review

- [ ] Did it pass in Desktop?

  1. Storybook → **Menu** → **Playground**. <kbd>Tab</kbd> to the trigger, <kbd>Enter</kbd> to open, <kbd>ArrowDown</kbd>/<kbd>ArrowUp</kbd> through rows, <kbd>Escape</kbd>.
  2. Expect: trigger reachable in normal tab order; open moves focus to "Cut" with a visible ring; arrows move focus with wraparound; Escape closes with no trap and returns focus to the trigger.

  1. Same story with VoiceOver (<kbd>Cmd</kbd> + <kbd>F5</kbd>). Open the menu, arrow through rows, close with Escape.
  2. Expect: trigger announces expanded/collapsed state; opening announces the menu region and "Cut"; arrowing announces each row by name; closing removes the menu from the accessibility tree (rotor/landmark scan should no longer find it).


rise-erpelding (approved):
Not seeing anything blocking here! Did manual keyboard/voiceover checks as well.

Some menu item things to note here that I think are probably out of scope, but I want to make sure we don't lose track of later:
* focusing of disabled menu items, that seemed a little fuzzy in the plans which maybe is expected
* making sure we close the menu when an item is activated
* handling submenus (when you click into a submenu, make sure it's not handled like an outside click)

blunteshwar (commented):
`updated()` 's openChanged guard (L396-398) in Menu.base.ts excludes the very first update:

  `const openChanged = changedProperties.has('open') && changedProperties.get('open') !== undefined;`

 This is meant to stop a phantom `swc-open/swc-close` event when a menu starts open or closed, but it gates the entire branch below it (L399-436) — not just the event
 dispatch. So a menu rendered with open already true also skips `registerDismissible`,
 the `Escape/outside-click listeners`, `startPlacement()`, and moving focus to the first
 item. It ends up ARIA-wired and reflecting open, but unanchored, undismissable, and
 unfocused, until open is toggled again.

`Popover.base.ts/Tooltip.base.ts` — the precedent this file's comments cite — don't do this. Both run the full open branch (dismiss registration, listeners, positioning, swc-open) on the first update whenever open starts true. The only thing either suppresses on initial-open is focus-stealing, and each does it narrowly (Popover's separate _hasCompletedFirstUpdate flag), not by skipping the whole branch.

2nd-gen/packages/core/components/menu/Menu.base.ts
```diff
@@ -151,6 +157,83 @@ export abstract class MenuBase extends SizedMixin(SpectrumElement, {
...
+  private isFocusWithin(): boolean {
+    return deepContains(this, getActiveElement());
+  }
+
+  // Escape closes the menu. Registered on `document` (capture) only while
+  // open (see `updated()`), matching `Tooltip.base.ts`/`Popover.base.ts`:
+  // `swc-menu` has no native top-layer light-dismiss to fall back on.
+  private readonly handleKeyDown = (event: KeyboardEvent): void => {
+    if (event.key !== 'Escape' || !this.open) {
+      return;
```
blunteshwar, line 229: **Tab leaves the menu open** : The only keyboard dismissal implemented is `Escape`. Tabbing from a menu item moves focus outside while the surface remains visible and the trigger retains `aria-expanded="true"`. APG requires `Tab/Shift+Tab` to move focus out and close the menu.[WAI-ARIA Menu pattern](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/)
@nikkimk What would you recommend here?

2nd-gen/packages/core/components/menu/Menu.base.ts
```diff
@@ -301,10 +399,40 @@ export abstract class MenuBase extends SizedMixin(SpectrumElement, {
...
+          capture: true,
+        });
+        document.addEventListener('click', this.handleOutsideClick, {
+          capture: true,
+        });
         this.startPlacement();
+        // Re-checks eligibility (e.g. newly visible rows) before moving focus
+        // in, then delegates "first item" to the controller's own preferred-
+        // item algorithm rather than re-deriving eligibility here.
+        this.focusNavigation.refresh();
```
blunteshwar: **Reopening focuses the last item, not the first**:  With the controller’s default memory: true, refresh() prefers lastFocused. After navigating to Paste, closing, and reopening, focus returns to Paste. Both the PR description and migration plan require every normal open to focus the first item. Explicitly reset the active item or disable memory for this menu.

blunteshwar (changes requested):

2nd-gen/packages/core/components/menu/Menu.base.ts
Rajdeepc: Good catch, fixed. Added a focusout listener that closes the menu when focus lands outside its own subtree (Tab or Shift+Tab out), without redirecting focus anywhere -- it just lets Tab continue wherever it was headed. Verified live: Tab out now flips open back to false and clears aria-expanded on the trigger.

2nd-gen/packages/core/components/menu/Menu.base.ts
Rajdeepc: Fixed, thanks. Every open now explicitly forces the first item active via setActiveItem() before falling back to whatever the controller's own refresh()/getActiveItem() would have preferred, so it no longer matters that memory defaults to true. Verified live: navigate to Paste, close, reopen -- lands on Cut every time.

Rajdeepc: @blunteshwar re your comment on the openChanged guard gating the whole branch: fixed. Replaced the local "not the init entry" checks in willUpdate() and updated() with a _hasCompletedFirstUpdate flag, matching Popover.base.ts's own pattern exactly -- the full open branch (dismiss registration, listeners, startPlacement()) now runs on the first update even when open starts true, and only the phantom event dispatch and the initial focus-steal are narrowly suppressed. Verified live: a menu rendered with open already true is now positioned (has actual-placement), dismissible via Escape, and doesn't steal focus on mount.

@rise-erpelding also picked up the close-on-item-activation item from your approval notes -- clicking a slotted row now closes the menu.

2nd-gen/packages/core/components/menu/Menu.base.ts
Rajdeepc: RSP reference: https://react-spectrum.adobe.com/Menu#content

blunteshwar: Correct me if I am wrong but in RSP, pressing the tab key doesn't shift focus out of the menu-item. So if we are following RSP then we should not allow focus to go out of menu-item with a tab press.

5t3ph: Opening the menu and then trying to use tab it seems tab is actually currently totally blocked? My focus stays on the menu item. I see your commit that reverted, and I see that RSP does _not_ allow Tab, but the [WAI-ARIA pattern does allow it](https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/examples/menu-button-actions-active-descendant/). We should get clarity in the accessibility channel, but we could follow-up before this merges, it doesn't have to hold up this phase.

5t3ph (approved):
I responded re: Tab to exit the menu but as noted we could come back to that later.

blunteshwar (approved):

PR #6723, opened 2026-09-09
feat(asset): add loadState property, load/error events, and phase 6 test coverage

5t3ph:
## Description

- Adds a readable `loadState` property (`'loading' | 'loaded' | 'error'`, reflected as
  `load-state`) plus `swc-asset-load`/`swc-asset-error` events, so a consumer (e.g. Card) can show
  a pending spinner while a slotted `` loads and hide it on completion/failure.
  - `loadState` is `'loaded'` immediately for a slotted `<svg>` or no slotted content — nothing
    async to wait for.
  - _Timing guarantee:_ an already-`complete` `` (e.g. served from cache) still fires
    `swc-asset-load`/`swc-asset-error` via a scheduled microtask, so consumers can always just
    listen for the event without special-casing the cache-hit case.
  - Swapping the slotted `` (via the existing `slotchange` re-resolution) resets `loadState`
    and re-attaches listeners to the new element.
  - `swc-asset-error`'s `detail.src` mirrors the `DropzoneDragLeaveDetail` pattern of capturing
    fields synchronously rather than exposing the raw native event.
- CSS: centers slotted content and hides any slotted element after the first.
- Full decision-log writeup added to `component-plan.md` for Q1–Q3, including why
  `HTMLImageElement.decode()` and `loading="lazy"` don't need special handling.

## Motivation and context

Asset's plan called for a loading-state mechanism so embedding components (Card, primarily) can
represent per-image load progress without Asset imposing its own visual/AT treatment. This was
explicitly left open pending a design decision (Q1–Q3); this PR resolves and implements it. The
remaining Phase 6 testing items (zero-size confirmation, a11y snapshots for the newer
Fit/Background/Sizing stories, VRT coverage) were also outstanding and are closed out here.

## Related issue(s)

- SWC-2317 (Epic), SWC-2320 (Testing)

## Manual review test cases

- [ ] _`loadState` reflects a successful `` load_
  1. Go to `components-asset--fit` in Storybook
  2. Inspect the `load-state` attribute on `<swc-asset>` after the image loads
  3. Expect `load-state="loaded"` and a `swc-asset-load` event fired on the element

- [ ] _`loadState` reflects a failed `` load_
  1. Slot an `` with a broken `src` into `<swc-asset>`
  2. Listen for `swc-asset-error`
  3. Expect `load-state="error"` and the event's `detail.src` to match the broken URL

- [ ] _Cached image still fires the load event_
  1. Load a page with an `` already in the browser cache slotted into `<swc-asset>`
  2. Attach a `swc-asset-load` listener after the element connects
  3. Expect the listener to still fire exactly once (via the microtask-scheduled re-fire)

- [ ] _Slotted SVG VRT rows_
  1. Review `test/vrt/asset.vrt.ts`'s `Permutations`/`ForcedColors` stories in Chromatic
  2. Expect `` and slotted-`<svg>` rows to render equivalently across background/fit,
     border-radius, and sizing/aspect-ratio permutations


cdransf (approved):
This looks awesome! ✨

2nd-gen/packages/core/components/asset/Asset.base.ts
```diff
@@ -14,54 +14,398 @@ import { PropertyValues } from 'lit';
...
+
+  /**
+   * Implements `decorative` and the `accessibleLabel` fallback, in priority
+   * order: `decorative` > the child's own name (self-healing a missing
+   * `role="img"` on a named `<svg>`) > `accessibleLabel` > warn.
+   */
+  private resolveAccessibleName(children: Element[]): void {
+    if (this.decorative) {
+      // Don't claim ownership of a consumer's own `aria-hidden`.
+      this._appliedAriaHidden = !this.hasAttribute('aria-hidden');
```
rubencarvalho: i might be misreading this, but doesn't this get recomputed on every update while `decorative` is true? the loadState loading→loaded change triggers a second update, and by then `aria-hidden` is already set so `_appliedAriaHidden` flips back to false. then toggling `decorative` off won't remove it. wonder if we should only set the flag when we actually add the attr?

2nd-gen/packages/swc/components/asset/test/asset.test.ts
```diff
@@ -58,230 +61,918 @@ export const OverviewTest: Story = {
...
+            warnCalls.length,
+            'no warnings for any of these valid CSS values'
+          ).toBe(0);
+        })
+    );
   },
 };
 
-export const VariantMutationTest: Story = {
+export const AriaHiddenSurvivesDecorativeToggleTest: Story = {
```
rubencarvalho, line 585: could we add a variant of this without the consumer aria-hidden to start? this one passes even if the removal path is broken since the attribute's already there

rubencarvalho (commented):
left a couple small notes, mostly one thing around the decorative aria-hidden handling. rest looks good to me 👍

2nd-gen/packages/swc/components/asset/asset.css
```diff
@@ -37,11 +37,23 @@
 ::slotted(*) {
   box-sizing: border-box;
   display: block;
+  place-content: center;
```
aramos-adobe: Isn't this used with `display: grid | flex`? Maybe we don't need this.

2nd-gen/packages/core/components/asset/Asset.base.ts
5t3ph: Gooooood catch - updated and regression test added 👍

2nd-gen/packages/swc/components/asset/asset.css
5t3ph: I added a comment to clarify - I found this worked for aligning the message block that appears if the image fails to load.

PR #6729, opened 2026-09-10
feat(asset): documentation

5t3ph:
## Description

Phase 7 (Documentation, SWC-2321) for Asset (SWC-2317)

- **`asset.mdx`**: documents the `background`/`fit="contain"` letterboxing-and-loading pairing, the `aspect-ratio` + single-dimension derivation behavior, and adds a new "Loading state" subsection covering `loadState` and the `swc-asset-load`/`swc-asset-error` events, including a consumer spinner example and `loading="lazy"`/`decoding="async"` guidance. The Accessibility section now also documents a child `<title>` element as a valid SVG accessible-name source, alongside `aria-label`/`aria-labelledby`.
- **`asset.stories.ts`**: adds a `LoadingState` behaviors story (success + broken-image cases, wired to the Actions panel), extends the `Accessibility` story with `decorative` and `accessible-label` fallback examples, and wraps the shared Storybook helper's `template()` so the internally-managed `load-state` attribute never appears in the Controls panel or generated code snippets (it would otherwise misleadingly suggest a consumer sets it).
- **`component-plan.md`**: checks off the completed Documentation checklist items.
- **New `migration-guide.mdx`**: the consumer-facing `sp-asset` → `swc-asset` migration guide, covering the tag/import rename, `label` → `accessible-label` (with its scope change from icon-only to general fallback), the removed `variant` attribute and its `--mod-asset-*` custom properties (no replacement), the default `fit` change from always-`contain` to `cover`, and the new sizing/background/loading-state API.

## Motivation and context

Closes out the remaining Phase 7 documentation gaps identified in the plan's Documentation checklist, and gives Asset the same consumer migration guide every other component ships with before promotion to public.

## Related issue(s)

- SWC-2317 (Epic), SWC-2321 (Documentation)

## Manual review test cases

- [ ] _Loading state story and docs match implementation_
  1. Go to `components-asset--loading-state` in Storybook
  2. Open the **Actions** panel
  3. Expect `swc-asset-load` to fire for the successful image and `swc-asset-error` (with `detail.src`) for the broken one

- [ ] _`load-state` hidden from generated code_
  1. Go to any Asset story (e.g. `components-asset--anatomy`) and open the "Show code" panel
  2. Expect `load-state="loading"` to **not** appear in the snippet, even though it's visible as a live attribute on the rendered element in devtools

- [ ] _Decorative/accessible-label examples render correctly_
  1. Go to `components-asset--accessibility` in Storybook
  2. Expect three examples: an image named via `accessible-label`, a decorative image, and a named `<svg>`

- [ ] _Migration guide renders_
  1. Go to `Asset/Migration guide` in Storybook docs
  2. Confirm all sections render, all before/after snippets are correct, and no MDX errors appear in the console

  1. No focusable parts are added by this change — the new Loading state story and migration guide are non-interactive documentation; confirm no regression in keyboard flow through the existing Asset stories

  1. Go to `components-asset--accessibility` in Storybook
  2. Run a screen reader (VoiceOver/NVDA) over the new `decorative` and `accessible-label` examples
  3. Expect the decorative example to be skipped entirely, and the `accessible-label` example to announce as an image named "Named File Asset"-equivalent text; confirm this matches the existing `asset.a11y.spec.ts` ARIA-snapshot coverage


cdransf (approved):
Looks great! ✨

2nd-gen/packages/swc/components/asset/asset.mdx
```diff
@@ -45,10 +50,56 @@ Consuming components are responsible for applying other styles such as borders,
...
+```javascript
+const asset = document.querySelector('swc-asset');
+const spinner = document.querySelector('swc-progress-circle');
+
+spinner.hidden = asset.loadState !== 'loading';
+asset.addEventListener('swc-asset-load', () => (spinner.hidden = true));
+asset.addEventListener('swc-asset-error', () => (spinner.hidden = true));
+```
+
+Open the **Actions** panel in Storybook to observe the events as the images below load.
```
rise-erpelding, line 93: This might very well be my lack of awareness around Storybook controls but I'm not immediately seeing a way to open the **Actions** panel and be able to see this specific story, I can open the Playground, but that's a different story/component instance that I'm not sure is actually illustrating what's meant to be shown here.

2nd-gen/packages/swc/components/asset/asset.mdx
```diff
@@ -36,6 +36,11 @@ attribute instead (`xMidYMid slice` for `cover`, `xMidYMid meet` for `contain`).
 `background` controls what shows behind slotted content and around its edges when `fit="contain"`
 letterboxes it: `transparent` (default), a `solid` color, or an opacity `checkerboard`.
```
rise-erpelding, line 37: The way that `solid` works is you can specify a background color with `--swc-asset-background-color`, otherwise it just defaults to `gray-100`, right? I think I would find that information useful in this section if I were a consumer.

Ditto for the description of `background` on the API table, but I could also see justification not to add it to try to keep the description brief. I know the Custom Properties table also covers this, but that might not be the first place that I would look for information like that.

2nd-gen/packages/swc/components/asset/migration-guide.mdx
```diff
@@ -0,0 +1,219 @@
...
+```html
+<!-- Before -->
+<sp-asset>
+  <img src="demo.jpg" alt="Demo image" />
+</sp-asset>
+
+<!-- After: name the image directly -->
+<swc-asset>
+  <img src="demo.jpg" alt="Demo image" />
+</swc-asset>
```
rise-erpelding, line 118: The before/after look the same to me aside from the name of the tag?

2nd-gen/packages/swc/components/asset/asset.mdx
```diff
@@ -45,10 +50,56 @@ Consuming components are responsible for applying other styles such as borders,
 ### Sizing
 
 Set `aspect-ratio`, or `width`/`height`, to size the asset. `aspect-ratio` accepts CSS `<ratio>`
-syntax (including the `square` keyword) and has no effect when both `width` and `height` are set.
+syntax (including the `square` keyword). Pairing `aspect-ratio` with only one of `width` or
+`height` derives the other dimension from the ratio; setting both `width` and `height` leaves
+`aspect-ratio` with no effect, since both dimensions are already fixed.
```
rise-erpelding, line 56: What happens if you don't specify `width`, `height`, or `aspect-ratio`?

2nd-gen/packages/swc/components/asset/stories/asset.stories.ts
rise-erpelding, line 256: Does this need this `role` added or does `role="img"` get added automatically?

2nd-gen/packages/swc/components/asset/asset.mdx
5t3ph: Ah, true - modified to just note that you can see the load event in the Actions panel.

2nd-gen/packages/swc/components/asset/asset.mdx
5t3ph: Added a note in this section!

2nd-gen/packages/swc/components/asset/asset.mdx
5t3ph: Added a note to clarify.

2nd-gen/packages/swc/components/asset/migration-guide.mdx
5t3ph: Updated this section to be "Ensure an accessible label is provided" and edited the content accordingly.

2nd-gen/packages/swc/components/asset/stories/asset.stories.ts
5t3ph: It technically doesn't need it since we've built-in adding it, so went ahead and removed in the samples since it's not required.

rise-erpelding (approved):

PR #6734, opened 2026-09-11
feat(menu): migrate surface styling and adopt native Popover API

Rajdeepc:
## Description

Phase 5 (styling) follow-up to the `swc-menu` accessibility migration (#6722):

- Migrates `.swc-Menu`'s surface chrome to real S2 tokens
- Adopts the native Popover API (`popover="auto"`) for the surface, fixing real clipping/animation gaps a reviewer flagged

## Motivation and context

- `spectrum-css`'s `.spectrum-Menu` has **zero chrome** — background/border/radius/shadow all live on `.spectrum-Menu-item`, which is `swc-menu-item`'s scope and doesn't exist yet. The floating-popup look comes from a wrapping `.spectrum-Popover` in real Spectrum, so `.swc-Menu` mirrors `swc-popover`'s own `--swc-popover-*` chrome tokens instead. Row-level styling stays deferred until `swc-menu-item` ships.
- `swc-menu` had no native top-layer rendering, so the surface got clipped by overflow ancestors (confirmed live, not a Storybook quirk) and had no entrance/exit animation. Fixed with `popover="auto"`, matching `swc-popover`'s default mode (not `swc-tooltip`'s `manual` mode, which is specific to coexisting with an already-open popover).
- Deleted the hand-rolled outside-click listener and the closing half of the Escape handler in favor of native light-dismiss. `registerDismissible`/`unregisterDismissible` stay for cross-component coordination, matching `Popover.base.ts`.
- `_onBeforeToggle` is now the single reconciliation point for any close (native or programmatic), mirroring `Popover.base.ts`'s `_syncOpen`/`_onBeforeToggle` split to avoid double-teardown.
- Tab-trap, Enter/click-to-close, and focus management are unaffected — verified live via Playwright.
- Also fixed an unrelated `menu.stories.ts` syntax error that was breaking Storybook's entire story index, and added a `select` control for `placement`.

## Related issue(s)

- Follow-up to #6722

## Author's checklist

- [x] I have reviewed the Accessibility Practices for this feature: [ARIA Practices](https://www.w3.org/TR/wai-aria-practices/)

- [x] **Keyboard**
  1. Open `Components/Menu/Playground` in Storybook, click the trigger.
  2. <kbd>Tab</kbd> — focus stays trapped on the active row, menu stays open.
  3. <kbd>Enter</kbd> on the focused row — menu closes, focus returns to the trigger.
  4. <kbd>Escape</kbd> — menu closes via native light-dismiss, focus returns to the trigger.
  5. Click outside — menu closes via native light-dismiss.

- [x] **Screen reader**
  1. No role/name/state changes — `role="menu"`, `aria-expanded`, `aria-haspopup` unchanged from #6722.
  2. Surface chrome and top-layer promotion carry no semantic meaning, no new announcements.


Rajdeepc:
@rubencarvalho We're planning to move menu over to the native Popover API with top-layer rendering + native light-dismiss, instead of the outside-click/Escape logic it hand-rolls today. My plan was `popover="auto"` on the menu's surface, matching what `swc-popover` already does in its default mode — not manual, since that's more of swc-tooltip's special case for staying open over an already-open popover. 
After discussing this with @5t3ph I wanted to check with you before building it, does anything about how menu behaves (submenus later on, coexisting with other open popovers/tooltips, etc.) push toward manual instead, or is auto the right call?

Also figured that menu should plug into the same dismissible-stack mechanism you built for popover, so everything using the Popover API coordinates visibility the same way rather than menu rolling its own version. Does that sound right to you, or did you have a different shape in mind for how these should coordinate?

I want to split this into its own PR (separate from the menu styling one) so we can get this right without having to redo it later.

2nd-gen/packages/swc/components/menu/menu.css
```diff
@@ -14,19 +14,43 @@
...
+ * tokens exists anywhere in that repo, so the values are shared rather than
+ * guessed. `swc-menu` does not wrap `<swc-popover>` as a component (see the
+ * migration plan's Architecture section for why), but nothing about that
+ * decision applies to reusing the same design tokens for the same visual
+ * role. Row-level styling (colors, icons, hover/selection states, item
+ * padding) is `swc-menu-item`'s own scope, not yet implemented. */
 .swc-Menu {
+  --swc-menu-background-color: token("background-layer-2-color");
+  --swc-menu-border-color: token("popover-border-color");
+  --swc-menu-corner-radius: token("corner-radius-700");
```
rubencarvalho: I'd default to having "private" properties, unless we really know we need this customization

2nd-gen/packages/swc/components/menu/menu.css
```diff
@@ -14,19 +14,43 @@
...
+ * Background, border, radius, and shadow mirror `swc-popover`'s own tokens,
+ * not `spectrum-css`'s `.spectrum-Menu` (which carries none of this chrome
+ * itself): in real Spectrum this chrome comes from a wrapping
+ * `.spectrum-Popover`, and no menu-specific override of those popover
+ * tokens exists anywhere in that repo, so the values are shared rather than
+ * guessed. `swc-menu` does not wrap `<swc-popover>` as a component (see the
+ * migration plan's Architecture section for why), but nothing about that
+ * decision applies to reusing the same design tokens for the same visual
+ * role. Row-level styling (colors, icons, hover/selection states, item
+ * padding) is `swc-menu-item`'s own scope, not yet implemented. */
```
rubencarvalho: We decided to reduce comment length - let's try to summarise this in 1-2 sentences if possible or reconsider the temporal usefulness of this

Rajdeepc: Yes , this was agent creating loops to gather context.

2nd-gen/packages/swc/components/menu/menu.css
5t3ph: or just keep the token() inline with those properties 👍

2nd-gen/packages/swc/components/menu/stories/menu.stories.ts
```diff
@@ -73,6 +76,12 @@ export const Playground: Story = {
   args: {
     open: false,
     for: 'playground-trigger',
+    'actual-placement': null,
+    placement: 'bottom-start',
+    'should-flip': true,
+    '--swc-menu-background-color': '#8d1919',
+    '--swc-menu-border-color': '#deb0b0',
+    'default-slot': 'Hello I am a button',
```
5t3ph: Suggest removing all of these, especially if custom properties become private.
```suggestion
```

2nd-gen/packages/swc/components/menu/menu.css
```diff
@@ -14,33 +14,65 @@
...
+ * the discrete transition and would snap to its default mid-fade otherwise. */
+.swc-Menu:popover-open {
+  display: flex;
+  opacity: 1;
 }
 
-/* Minimal padding so slotted rows are visually distinct and roving focus is
- * easy to see moving between them; not a real row treatment. `swc-menu-item`
- * owns its own padding once it exists, superseding this. */
+@starting-style {
```
5t3ph, line 96: As can be [verified against React](https://react-spectrum.adobe.com/Menu#menu-trigger), Menu has the same entrance/exit animations as popover which includes a bit of a translate that adjusts based on side, as well as the fade in/out. Please copy what you can from Popover's CSS.

Note that while there isn't a "tip", there is supposed to be a bit of a gap between the trigger and the menu, as confirmed from React and also latest Figma specs.

2nd-gen/packages/swc/components/menu/menu.css
```diff
@@ -14,33 +14,65 @@
...
 .swc-Menu {
-  display: flex;
+  --swc-menu-background-color: token("background-layer-2-color");
+  --swc-menu-border-color: token("popover-border-color");
+  --swc-menu-corner-radius: token("corner-radius-700");
+
   position: absolute;
+  inset: auto;
   flex-direction: column;
+  max-block-size: min(var(--swc-placement-available-height, 90vb), 90vb);
```
5t3ph, line 37: Missing handling inline-size. Borrowing from our rule from Popover and also adding `320px` as used by React.
```suggestion
  inline-size: max-content;
  max-inline-size: min(var(--swc-placement-available-width, 100vi), calc(100vi - 24px), 320px);
  max-block-size: min(var(--swc-placement-available-height, 90vb), 90vb);
```

2nd-gen/packages/swc/components/menu/menu.css
```diff
@@ -14,33 +14,65 @@
...
+  background-color: var(--swc-menu-background-color);
```
5t3ph: Per React and specs, the container has padding in addition to individual menu item padding. Right now, it is inheriting UA padding.
```suggestion
  padding: token("spacing-100");
  background-color: var(--swc-menu-background-color);
```

2nd-gen/packages/swc/components/menu/menu.css
```diff
@@ -14,33 +14,65 @@
...
+  border: token("border-width-100") solid var(--swc-menu-border-color);
+  border-radius: var(--swc-menu-corner-radius);
+  box-shadow: token("drop-shadow-elevated");
+  overflow: auto;
+  opacity: 0;
+  transition:
+    opacity token("animation-duration-200") token("animation-ease-out"),
+    overlay token("animation-duration-200") allow-discrete,
```
5t3ph, line 48: `transform` will need added to accomodate the requested entrance/exit animations.

```suggestion
    transform token("animation-duration-200") token("animation-ease-out"),
    overlay token("animation-duration-200") allow-discrete,
```

[nit] could also save the duration to a private property.

5t3ph (changes requested):
Update to incorporate popover API looks good 👍 

Requesting changes primarily to complete the entrance/exit animations.

2nd-gen/packages/swc/components/menu/menu.css
Rajdeepc: Good call, went with the simpler option and inlined the `token()` calls directly. There's no real consumer override request for these yet, so no reason to carry the extra `--swc-menu-*` indirection until one shows up.

2nd-gen/packages/swc/components/menu/stories/menu.stories.ts
Rajdeepc: I have pulled the custom-property overrides out of the Playground args entirely, matching the CSS change above.

2nd-gen/packages/swc/components/menu/menu.css
Rajdeepc: Thanks for the pointers, that was genuinely missing. I have added the transform-based slide (borrowed straight from Popover's pattern) plus a proper gap between the trigger and the menu, so it now fades and slides in together instead of just popping into place with zero space above the trigger.

2nd-gen/packages/swc/components/menu/menu.css
Rajdeepc: Good catch, this was a real missing gap.

2nd-gen/packages/swc/components/menu/menu.css
Rajdeepc: You're right, it was just inheriting UA padding before. I have added `padding: token("spacing-100")` on the container itself now, alongside the per-item padding.

2nd-gen/packages/swc/components/menu/menu.css
```diff
@@ -21,27 +21,30 @@
 /* `inset: auto` resets the UA popover default of `inset: 0`, which
  * auto-centers the surface and fights the inline top/left/translate
  * PlacementController writes -- same reset `swc-popover` applies.
- * Background/border/radius/shadow mirror `swc-popover`'s own tokens, since
- * `spectrum-css`'s `.spectrum-Menu` carries no chrome of its own.
- * `display` is intentionally absent here (see `:popover-open` below); an
- * unconditional value would override the UA's default-hidden state. */
+ * Background/border/radius/shadow are inline `token()` calls, not
+ * `--swc-menu-*` custom properties: no consumer override request exists yet
+ * (review feedback, PR #6734). `display` is intentionally absent here (see
```
5t3ph, line 26: [nit] the history here doesn't need to be kept

5t3ph (commented):
Previous round of style updates look good.

I wonder if we should set it up so that the minimum width at least matches the trigger? I suggest passing the computed value from JS into the styles as a custom property.

Image

5t3ph (approved):
As discussed in Slack, we can revisit the menu width later, but it seems most real product use cases are attached to icon-only ActionButtons

2nd-gen/packages/core/components/menu/Menu.base.ts
```diff
@@ -534,14 +474,11 @@ export abstract class MenuBase extends SizedMixin(SpectrumElement, {
     super.disconnectedCallback();
     this.placementController.stop();
     this.clearTriggerAria();
-    this.removeTriggerClickListener();
+    this._trigger = null;
```
rubencarvalho: I might be missing something, but I'm not sure the menu recovers when it's removed and re-added to the DOM. `disconnectedCallback` detaches the trigger listeners and clears the ARIA, and `wireTrigger()` only runs from `updated()` when `for`/`triggerElement`/`open` changes, so after a move the trigger may stop toggling.

Also, if the menu is open when it's removed, I believe the browser hides the popover without firing `beforetoggle`, so `open` could stay `true`.

Would it make sense to rewire in `connectedCallback()` (when `hasUpdated`) and reset `open` here? Probably better in `MenuBase` than in the ported controller, so that one stays identical to main.

2nd-gen/packages/core/components/menu/Menu.base.ts
```diff
@@ -423,21 +321,109 @@ export abstract class MenuBase extends SizedMixin(SpectrumElement, {
...
+        shown = true;
+      } catch {
+        shown = false;
+      }
+    }
+    if (!shown) {
+      return;
+    }
+    if (this._hasCompletedFirstUpdate) {
+      this.dispatchOpenEvents(true);
```
rubencarvalho: Now that the surface has a 200ms transition, I think `swc-after-open` / `swc-after-close` fire before it finishes, since they're dispatched right after `swc-open` / `swc-close`. Popover waits for the transition with `runAfterTransition`. Do you think it's worth doing the same here, or would you rather keep that for a follow-up?

2nd-gen/packages/core/controllers/trigger-press-controller/src/trigger-press-controller.ts
```diff
@@ -0,0 +1,273 @@
...
+ *       // Not already closed via our own setter, so this is a genuine
+ *       // native dismissal (Escape, outside click, ...).
+ *       this.pressGuard.noteNativeDismiss();
+ *     }
+ *     this.open = false;
+ *   };
+ * }
+ * ```
+ */
+export class TriggerPressController implements ReactiveController {
```
rubencarvalho: I couldn't find test changes in this PR. Did I miss them? The controller on main has `test/trigger-press-controller.test.ts`, which could be easy to bring over.

For the menu, some cases that seem worth covering:
- Escape and outside-click dismiss
- clicking the trigger while the menu is open
- focus going back to the trigger
- programmatic open/close
- remove and re-add

Totally fine as a follow-up if that's already planned.

2nd-gen/packages/swc/components/menu/menu.css
```diff
@@ -14,33 +14,105 @@
...
- * and this stylesheet's `display: flex` above comes after the UA
- * stylesheet's `[hidden] { display: none }` in the cascade, so it was
- * winning the tie and the surface never actually hid despite the attribute
- * being set correctly. The extra specificity here (0,2,0) restores it. */
-.swc-Menu[hidden] {
-  display: none;
+/* Gated on `actual-placement` (async, set once the first position lands) so
+ * the surface doesn't fade in at its 0,0 origin and jump to the trigger
+ * mid-transition; matches `swc-popover`'s own gating. */
+:host([actual-placement]) .swc-Menu:popover-open {
```
rubencarvalho: Edge case, and I'm not sure how likely it is: `opacity: 1` depends on `actual-placement`, and `startPlacement()` returns early when there's no resolved trigger. So if `open` is set with a missing or wrong `for`, I think the menu would be in the top layer but invisible, and still take focus. Is that worth guarding against?

2nd-gen/packages/swc/components/menu/menu.css
```diff
@@ -14,33 +14,105 @@
...
   flex-direction: column;
+  inline-size: max-content;
+  max-inline-size: min(var(--swc-placement-available-width, 100vi), calc(100vi - 24px), 320px);
+  max-block-size: min(var(--swc-placement-available-height, 90vb), 90vb);
+  padding: token("spacing-100");
+  background-color: token("background-layer-2-color");
+  border: token("border-width-100") solid token("popover-border-color");
+  border-radius: token("corner-radius-700");
+  box-shadow: token("drop-shadow-elevated");
+  overflow: auto;
```
rubencarvalho: Should this also have `overscroll-behavior: contain`, like `swc-popover`? Otherwise I think scrolling past the end of a long menu would scroll the page behind it.

2nd-gen/packages/swc/components/menu/menu.css
```diff
@@ -14,33 +14,105 @@
...
+* {
+  box-sizing: border-box;
+}
+
+/* `inset: auto` resets the UA popover default of `inset: 0`, which
+ * auto-centers the surface and fights the inline top/left/translate
+ * PlacementController writes -- same reset `swc-popover` applies.
+ * Background/border/radius/shadow are inline `token()` calls, not
+ * `--swc-menu-*` custom properties: no consumer override request exists yet
+ * (review feedback, PR #6734). `display` is intentionally absent here (see
```
rubencarvalho: nit: I think the "(review feedback, PR #6734)" part can go, in line with the earlier comment about keeping history out of comments.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
```diff
@@ -614,14 +616,15 @@ Planned rendering shape:
...
+- [x] Surface is a native `popover="auto"` element, not a plain hidden `<div>` — fixes top-layer clipping, adds entrance/exit animation ([Q25](#blockers-and-open-questions)).
 
 #### Visual model and regressions
 
-- [ ] Verify i18n size modifiers (`:lang(ja)`, `:lang(ko)`, `:lang(zh)`) if present in S2 source
-- [ ] Add `@cssprop` JSDoc tag to the primary SWC component class for every exposed `--swc-*` property
-- [ ] Pass stylelint (property order, `no-descending-specificity`, token validation)
+- [x] Verify i18n size modifiers (`:lang(ja)`, `:lang(ko)`, `:lang(zh)`) if present in S2 source — present, but scoped to `--spectrum-menu-item-*` line-height properties, not anything `.swc-Menu` itself owns; deferred with the rest of row-level styling
+- [x] Add `@cssprop` JSDoc tag to the primary SWC component class for every exposed `--swc-*` property — `--swc-menu-background-color`, `--swc-menu-border-color`, `--swc-menu-corner-radius`
```
rubencarvalho: I think this line predates 155f044294. The `--swc-menu-*` properties were removed in that commit, so there's no `@cssprop` to list anymore. Line 620 and the PR description also still mention mirroring the `--swc-popover-*` properties. Could you update those too?

2nd-gen/packages/core/components/menu/Menu.base.ts
Rajdeepc: Leaving this as-is for now. Noted for whenever we do decide it's worth solving (probably alongside whatever fix Popover eventually gets for the same thing).

2nd-gen/packages/core/controllers/trigger-press-controller/src/trigger-press-controller.ts
Rajdeepc: Checked the test PR (#6763). Status on each:

- **Escape and outside-click dismiss**: covered with real trusted input in `menu.a11y.spec.ts` ("Escape closes the menu and returns focus to the trigger", "an outside click closes the menu"). The Vitest play functions also cover the synthetic-input side of the same reconciliation path.
- **Focus going back to the trigger**: covered both ways, synthetic (`FocusNotStolenOnCloseTest` plus a positive case that asserts focus restores when it was inside the menu) and trusted (the Escape spec above).
- **Programmatic open/close**: covered throughout, most tests toggle `menu.open` directly rather than via click.
- **Clicking the trigger while the menu is open**: partially covered. `ClickToToggleTest` asserts a second click closes the menu, but it uses synthetic `userEvent.click`, which never triggers native light-dismiss, so it isn't actually exercising the reopen-guard race `TriggerPressController` protects against. There's no trusted-click equivalent yet, and popover.a11y.spec.ts already has one for `swc-popover` ("clicking the trigger again while open closes it instead of reopening it"). I can add the same test to `menu.a11y.spec.ts` if you'd like, that's the one real gap I found.

2nd-gen/packages/swc/components/menu/menu.css
Rajdeepc: I confirmed, this was real. Fixed by guarding the focus-first-item logic in `_show()` on `this._trigger` being resolved, so it only steals focus when there's actually somewhere visible to land. Verified again after the fix: focus stays on `body` in the broken-trigger case, and a normal open with a valid trigger still focuses the first row as before.

2nd-gen/packages/swc/components/menu/menu.css
Rajdeepc: Added `overscroll-behavior: contain`, matching `swc-popover`.

2nd-gen/packages/swc/components/menu/menu.css
Rajdeepc: Removed.

CONTRIBUTOR-DOCS/03_project-planning/03_components/menu/migration-plan.md
Rajdeepc: You're right on both counts

2nd-gen/packages/core/controllers/trigger-press-controller/src/trigger-press-controller.ts
Rajdeepc: Added: `clicking the trigger again while open closes it instead of reopening it`, using a real trusted click against the published Open and close story. Also found the test PR (#6763) was 6 commits behind its own base (docs-menu-storybook), missing the reopen-guard fix itself among other things, so I merged that in first, otherwise the new test would have failed against stale code. Both pushed to #6763.

rubencarvalho (approved):

PR #6739, opened 2026-09-11
feat(asset): expanded API and making it public

5t3ph:
## Description

Final review pass for Asset (Epic SWC-2317) on `seckles/asset-component`. The bulk of the work (API, accessibility, styling, testing, and documentation) already shipped and was reviewed across three separate PRs: #6716, #6723, and #6729 — all merged to `main`. This PR contains only what's new since then: closing out the plan's last open item and a compliance/cleanup pass, plus the changeset those PRs were missing.

- **Code-conformance pass**: added the required class-level JSDoc to `AssetBase` (was missing entirely; `@slot` was only declared on the concrete class), fixed a section-separator width off-by-one, reordered `asset.test.ts`'s `Dev mode warnings` section to the end of the file (matching `Card`/`ProgressCircle` convention, verified against both), added two missing descriptive `expect()` messages, and renamed a test section header that referenced "PR feedback" instead of describing its content.
- **Comment cleanup**: trimmed a few comments that referenced past review/bug history instead of just explaining current behavior (e.g. a test section titled "PR feedback regressions," a comment describing what "previously" broke).
- **Docs consistency**: checked off Asset's row in the 2nd-gen migration status table (all 7 steps), added `asset` to `isDebug()`'s "used by" list in the shared-resources reference doc.
- **Changeset added**: `2nd-gen/.changeset/asset-expanded-public-api.md`

## Motivation and context

Closes out Asset's Phase 8 (Review) checklist so the component plan has no remaining open items, aside from Card's own integration work, which is intentionally tracked separately in SWC-2566.

## Related issue(s)

- SWC-2317 (Epic), SWC-2322 (Review)

## Author's checklist

## Reviewer's checklist

### Manual review test cases

- [ ] _Sizing behaves correctly across all three input modes_
  1. Go to `Asset` > `Sizing` in Storybook
  2. Expect the `16/9 aspect ratio` example to render at that ratio, and the `Explicit width and height` example to render at exactly 100×160px, both without collapsing to zero size
  3. In the Playground, clear `width`/`height`/`aspect-ratio` entirely with an image slotted
  4. Expect the asset to still render at a sensible non-zero size, sized to the image's natural dimensions

- [ ] _`fit` renders correctly for both `` and `<svg>` content_
  1. Go to `Asset` > `Fit` in Storybook
  2. Expect `cover` to crop the portrait image to fill its box, and `contain` to letterbox the landscape image inside its box without cropping
  3. Go to `Asset` > `Anatomy` and toggle `fit` in the Playground with the slotted SVG example
  4. Expect the SVG to resize the same way an `` would (`preserveAspectRatio` substituting for `object-fit`, which doesn't apply to SVG)

- [ ] _`background` shows through correctly in every mode_
  1. Go to `Asset` > `Background` in Storybook
  2. Expect `transparent` to show the page behind it, `solid` to show a filled `gray-100` background, and `checkerboard` to show the opacity-checkerboard pattern
  3. Confirm the background is visible around the edges when paired with `fit="contain"` on a mismatched-aspect-ratio image (see the Playground)

- [ ] _Loading state reflects real image load/error outcomes_
  1. Go to `Asset` > `Loading state` in Storybook 
  2. Inspect each element in devtools; expect `load-state="loaded"` and `load-state="error"` respectively

### Device review

- [ ] Did it pass in Desktop?

  1. Go to `Asset` > `Overview` in Storybook
  2. Press <kbd>Tab</kbd> from before the component through past it
  3. Expect focus to skip the asset entirely — `<swc-asset>` has no interactive parts and is never a tab stop
  5. Expect no instance to trap or receive focus, and no visible focus ring on any of them

  1. Go to `Asset` > `Overview` in Storybook with VoiceOver/NVDA running
  2. Expect the slotted `` to announce as an image with its `alt` text as the name
  3. Go to `Asset` > `Accessibility`
  4. Expect the first example (named via `accessible-label`) to announce with that label as its accessible name
  5. Expect the second example (`decorative`) to be skipped entirely — not announced at all, not even as an unlabeled image
  6. Expect the third example (a slotted `<svg>` with only `aria-label` set, no explicit `role`) to announce as an image with that label — confirming the self-healed `role="img"` is actually exposed to the AT, not just present in the DOM
  7. Navigate to an asset with a slotted `` that has no `alt`, no `accessible-label`, and isn't `decorative` (e.g. temporarily clear `alt` in the Playground)
  8. Expect DEBUG-mode console warning (not an AT-facing signal) — confirm the screen reader still announces *something* reasonable (e.g. falls back to the `src` filename per browser default) rather than silence, since this is an authoring-time warning, not a runtime accessibility fix


2nd-gen/packages/core/components/asset/Asset.base.ts
```diff
@@ -14,54 +14,402 @@ import { PropertyValues } from 'lit';
...
+   * in flight, `'loaded'` on success, `'error'` on failure. Immediately
+   * `'loaded'` for a slotted `<svg>` or no slotted content, since there's
+   * nothing asynchronous to wait for. Set internally by Asset; a consumer
+   * influences it only by changing what's slotted, not by setting it
+   * directly.
    */
-  @property({ type: String })
-  public label = '';
+  @property({ type: String, reflect: true, attribute: 'load-state' })
+  public loadState: AssetLoadState = 'loading';
```
cdransf, line 157: Should this not be public if consumers aren't intended to influence it? ✨

2nd-gen/packages/core/components/asset/Asset.base.ts
```diff
@@ -14,54 +14,402 @@ import { PropertyValues } from 'lit';
...
   protected override update(changes: PropertyValues): void {
-    if (typeof this.variant !== 'undefined') {
-      const constructor = this.constructor as typeof AssetBase;
-      validateEnum(this, {
-        prop: 'variant',
-        value: this.variant,
-        valid: constructor.VARIANTS,
-        url: 'https://spectrum-web-components.adobe.com/?path=/docs/components-asset--docs',
-      });
+    validateEnum(this, {
```
cdransf: Should we be returning all of these `validateEnum` calls on every update or should some be gated behind a `changedProperties.has`check? ✨

2nd-gen/packages/core/components/asset/Asset.base.ts
```diff
@@ -14,54 +14,402 @@ import { PropertyValues } from 'lit';
...
+  @property({ type: String, reflect: true, attribute: 'load-state' })
+  public loadState: AssetLoadState = 'loading';
 
   // ──────────────────────
   //     IMPLEMENTATION
   // ──────────────────────
 
+  // Only clear `aria-hidden` if this instance set it, so a consumer's own
+  // attribute survives.
+  private _appliedAriaHidden = false;
```
cdransf, line 165: Should we apply this pattern to avatar too? 🤔

2nd-gen/packages/core/components/asset/Asset.base.ts
```diff
@@ -14,54 +14,402 @@ import { PropertyValues } from 'lit';
...
+      }
+      return;
+    }
+    if (this._appliedAriaHidden) {
+      this.removeAttribute('aria-hidden');
+      this._appliedAriaHidden = false;
+    }
+
+    const [child] = children;
+    if (!child) {
```
cdransf, line 280: Would we ever want a non-decorative, empty `<swc-asset>` to hit the `warnIf` below? ✨

cdransf (approved):
Looks awesome! Just had some non-blocking questions ✨

2nd-gen/packages/core/components/asset/Asset.base.ts
5t3ph: Kind of a weird case, yes, but it's modeled off of Dropzones `dragged` attribute as the precedent for being kept public

2nd-gen/packages/core/components/asset/Asset.base.ts
5t3ph: Added SWC-2568

2nd-gen/packages/core/components/asset/Asset.base.ts
```diff
@@ -14,54 +14,420 @@ import { PropertyValues } from 'lit';
...
+    } else if (this.accessibleLabel) {
+      if (tagName === 'img') {
+        (child as HTMLImageElement).alt = this.accessibleLabel;
+      } else {
+        child.setAttribute('role', 'img');
+        child.setAttribute('aria-label', this.accessibleLabel);
+      }
+    }
+
+    warnIf(
```
aramos-adobe, line 307: Can we use this warning check for both cases if child exists or not? Only because I thought there was duplicate check at L284

2nd-gen/packages/core/components/asset/Asset.base.ts
```diff
@@ -14,54 +14,420 @@ import { PropertyValues } from 'lit';
...
+        this.setAttribute('aria-hidden', 'true');
+      }
+      return;
+    }
+    if (this._appliedAriaHidden) {
+      this.removeAttribute('aria-hidden');
+      this._appliedAriaHidden = false;
+    }
+
+    const missingNameWarning = `<${this.localName}> requires an accessible name: set "alt" on the slotted <img> (or role="img" plus "aria-label"/"aria-labelledby"/a child <title> on the slotted <svg>), set "accessible-label" on <${this.localName}>, or set "decorative".`;
```
aramos-adobe, line 277: `role="img"` is automatically set (?) So maybe we don't need to let devs know they need to set the role.

2nd-gen/packages/swc/components/asset/asset.mdx
```diff
@@ -0,0 +1,129 @@
...
+```javascript
+const asset = document.querySelector('swc-asset');
+const spinner = document.querySelector('swc-progress-circle');
+
+spinner.hidden = asset.loadState !== 'loading';
+asset.addEventListener('swc-asset-load', () => (spinner.hidden = true));
+asset.addEventListener('swc-asset-error', () => (spinner.hidden = true));
+```
+
+Open the **Actions** panel to observe the event as the image completes the load.
```
rubencarvalho: It's not really possible to see Action panel from here, maybe we could specify it is in on the Playground? Eventually linking directly to it?

2nd-gen/packages/core/components/asset/Asset.base.ts
```diff
@@ -14,54 +14,420 @@ import { PropertyValues } from 'lit';
...
+      // Already warned about by validateSlottedContent.
+      return;
+    }
+
+    const hasOwnName = this.hasOwnAccessibleName(child, tagName);
+    if (hasOwnName) {
+      if (tagName === 'svg' && !child.hasAttribute('role')) {
+        child.setAttribute('role', 'img');
+      }
+    } else if (this.accessibleLabel) {
```
rubencarvalho: We might be missing an ownership check here. Once Asset writes `accessibleLabel` into `alt`/`aria-label`, the next update sees that generated attribute as the child’s own name. Changing or clearing `accessibleLabel` then appears to leave the original label behind. I tried to ask Copilot for an example to better illustrate this:

```
<swc-asset accessible-label="First label">
  <img>                              // initially no alt
</swc-asset>

Asset update #1
  └─ sees no image name
     └─ writes alt="First label"

accessibleLabel changes to "Second label"

Asset update #2
  └─ sees alt="First label"
     └─ assumes the image supplied its own name
        └─ leaves it unchanged ❌
```

2nd-gen/packages/core/components/asset/Asset.base.ts
```diff
@@ -14,54 +14,420 @@ import { PropertyValues } from 'lit';
...
+   * incorrectly reset an already-resolved `loadState` back to `'loading'`).
+   */
+  private updateLoadState(children: Element[]): void {
+    const [child] = children;
+    const img =
+      child?.tagName.toLowerCase() === 'img'
+        ? (child as HTMLImageElement)
+        : null;
+
+    if (img === this._trackedImg) {
```
rubencarvalho, line 384: If an existing  gets a new src/srcset/sizes, slotchange won’t fire, so loadState may retain its previous terminal value while the existing listeners emit another event, which also seems a bit contradicting with the docs saying events fire exactly once per image instance. Now, i don't think this will be something that happens very often so we prob fine here, but I thought it was worth calling it out.

2nd-gen/packages/core/components/asset/Asset.base.ts
```diff
@@ -14,54 +14,420 @@ import { PropertyValues } from 'lit';
...
+        valid: ASSET_LOAD_STATE_VALUES,
+        url: DOCS_URL,
       });
     }
+    if (
+      changes.has('aspectRatio') ||
+      changes.has('width') ||
+      changes.has('height')
+    ) {
+      warnIf(
```
rubencarvalho, line 211: nit: the component plan says invalid `width` and `height` values should produce DEBUG warnings, but I don’t see those values validated

CONTRIBUTOR-DOCS/03_project-planning/03_components/asset/component-plan.md
```diff
@@ -460,113 +500,185 @@ plan contract pattern this document follows).
...
+  otherwise-valid pattern like `<svg><title>Chart</title></svg>` would still trigger the
+  "missing accessible name" DEBUG warning purely for lacking a role the consumer had no reason to
+  know to add. Self-healing the role removes that false positive without loosening what counts as
+  a real name source.
+- Self-healing only fires when a name source is already present, so it can't accidentally give a
+  role to genuinely unnamed or intentionally-`decorative`-adjacent SVG content.
+
+**Affects:** [Accessibility semantics notes](#accessibility-semantics-notes), [Changes
+overview](#changes-overview) (B7), [Public API](#public-api), [Implementation
+checklist](#implementation-checklist).
```
rubencarvalho, line 685: can you check? these references look unresolved: both `./research.md` and `./rendering-and-styling-migration-analysis.md` seem to not exist?

2nd-gen/packages/swc/components/asset/asset.mdx
5t3ph: updated the language

2nd-gen/packages/core/components/asset/Asset.base.ts
5t3ph: good call, updated!

2nd-gen/packages/core/components/asset/Asset.base.ts
5t3ph: comment added to docs just in case consumers hit this behavior

2nd-gen/packages/core/components/asset/Asset.base.ts
5t3ph: Updated the plan to clarify - these are passed to CSS custom props to deal with so that means a pretty wide syntax is accepted (including like `calc()`) so not adding a warning at this time. Will fallback to `auto` per the custom property chain if it is in fact an invalid value.

CONTRIBUTOR-DOCS/03_project-planning/03_components/asset/component-plan.md
5t3ph: removed - leftover from early in the plan history

aramos-adobe (approved):

rubencarvalho (approved):

PR #6732, opened 2026-09-10
feat(user-card): files, api, styling extend CardBase

aramos-adobe:
## Description

 **Scope:** This PR covers the file structure, public API, and styling for `swc-user-card` only. Tests (unit + a11y spec + VRT) and finalized documentation are on this branch as a starting point, but will be reviewed and completed in a **follow-up PR**.

Adds the new 2nd-gen `<swc-user-card>` component: a `Card` family member that identifies a person, following the approved [card family plan](https://github.com/adobe/spectrum-web-components/blob/main/CONTRIBUTOR-DOCS/03_project-planning/03_components/card/migration-plan.md). Extends `CardBase` directly — the same sibling-inheritance pattern as `ActionButton` extending `ButtonBase` — rather than a dedicated `UserCardBase`.

**Core** (`UserCard.types.ts`): declares `USER_CARD_VARIANTS` (`primary`/`secondary`/`tertiary` — `quiet` is not supported).

**API/Styling** (`UserCard.ts`, `user-card.css`): 

- Uses a `2px` outline on the avatar for contrast against the banner.
- wires the shared `renderCardTemplate()`'s `renderGlyph` callback to a new `avatar` slot; 
- defaults the `preview` aspect ratio to `3/1`
- positions the avatar in its own row between the media and content regions, pulling it up to overlap the preview banner by half its own size when a preview is present

**Avatar sizing**: a `SlotAttributePropagationController` (the same mechanism `CardBase` already uses for the `actions` slot) keeps the slotted avatar's own `size` attribute in sync with the card's `size` (`xs`: `100`, `s`: `300`, `m`: `500`, `l`: `700`, `xl`: `900`), overwriting any `size`/`outline` the consumer sets directly.

Adds a `UserCard` example to `swc-card`'s own Behaviors section (`card.stories.ts`/`card.mdx`), demonstrating that `title-as-link`/`selectable` are shared `CardBase` behavior across the whole card family.

Nests "User card" under "Card" in the Storybook sidebar for discoverability.

## Motivation and context

`swc-user-card` is the first of two planned `Card` family follow-ups (alongside `swc-product-card`) building on the already-merged `swc-card` scaffold. It's also the first concrete card to exercise `renderCardTemplate()`'s `renderGlyph` callback and the glyph-slot code paths, which were previously untested since `swc-card` never supplies a glyph.

## Related issue(s)

- SWC-2368

## Screenshots (if appropriate)

Screenshot 2026-09-10 at 2 17 08 PM

---

[...]

PR #6743, opened 2026-09-14
test(user-card): add vrt, tests and a11y specs

aramos-adobe:
## Description

Follow-up to #6732 — that PR had the API, styling, and file structure for `swc-user-card` but no tests, since it was already a big review. This one fills that in: unit tests, an a11y spec, and VRT.

What's covered:

- Variant restriction (`quiet` isn't allowed on user card, and now warns if you try it)
- The avatar slot itself — renders correctly, propagates `size`/`outline` onto whatever avatar you slot in (across all 5 card sizes), and now warns in dev mode if you forget to add one at all. I added that last part while writing tests — a user card without an avatar kind of defeats the point, so it seemed worth catching.
- The preview banner's `3/1` ratio and the avatar-overlap-the-banner effect, checked pixel-for-pixel across every size
- A Playwright a11y spec for roles/focus, matching the pattern `swc-card` already uses

One snag: the `title-as-link` keyboard test failed in WebKit. Turns out that's expected — Safari doesn't Tab to plain links unless you've turned on "Full Keyboard Access" in system settings, it's a real default not a bug. Skipped that one case for WebKit specifically rather than weaken the test for Chromium/Firefox, which both pass fine.

## Motivation and context

Just closing the testing gap left by #6732 before calling this component done. Bonus: this is the first card that actually uses the `renderGlyph` callback in the shared template, so it's also the first real test coverage for that code path.

## Related issue(s)

- SWC-2369

N/A, test-only change — the VRT snapshots are the visual coverage.

---

## Author's checklist

---

## Reviewer's checklist

### Manual review test cases

- [ ] _Missing-avatar warning_ — With `window.__swc.DEBUG = true` in the console, go to `Card > User card > Docs`, pull the avatar out of the Playground. You should see a console warning about the `avatar` slot.
- [ ] _Size/outline propagation_ — Same Docs page, **Sizes** story. Check each avatar's `size` attribute in devtools: should read 100/300/500/700/900 for xs through xl, and `outline` should be set on all of them.
- [ ] _Overlap_ — **With preview** story. The avatar should sit right on the seam between the banner and the content, roughly half overlapping the image.
- [ ] _VRT_ — Give the `User card VRT` Permutations and ForcedColors snapshots a look.

### Device review

- [ ] Did it pass in Desktop?

- [ ] **Keyboard** — Go to `Card > User card > Docs`, **Selectable** story. Tab to it — should get a visible focus ring, Enter/Space should fire `swc-card-click`. Then try **Title as link** — Tab should land on the actual link text, not the card itself. (WebKit won't Tab to it at all by default — that's a real Safari setting, not something to flag.)
- [ ] **Screen reader** — Overview story, VoiceOver or NVDA. Card itself shouldn't announce a role. Avatar should read as an image with its alt text, title/description read normally.


5t3ph (changes requested):
I realized we missed showing the actions slot populated in the regular stories. Can you add it maybe to the second card in the Anatomy section? Then also add it as a VRT and adjust any other tests as needed? It _should_ behave the same as Card, but just to make sure.

5t3ph (approved):
:shipit:

2nd-gen/packages/swc/components/user-card/test/user-card.a11y.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+    page,
+    browserName,
+  }) => {
+    // WebKit ships with "Full Keyboard Access" off by default, so Tab only
+    // cycles through form controls, not plain <a> links, unless a user has
+    // explicitly enabled that OS accessibility setting. That's a real
+    // Safari/WebKit default (https://bugs.webkit.org/show_bug.cgi?id=22261),
+    // not a bug in this component or in Playwright; Chromium and Firefox
+    // both correctly Tab to the anchor.
+    test.skip(
```
cdransf, line 75: Should we keep this? I think `card.a11y.spec.ts` has a similar assertion without the skip. ✨

2nd-gen/packages/swc/components/user-card/UserCard.ts
```diff
@@ -105,11 +109,38 @@ export class UserCard extends CardBase {
       ?.addEventListener('slotchange', this._handleAvatarSlotChange);
   }
 
+  protected override updated(changedProperties: PropertyValues): void {
+    super.updated(changedProperties);
+    this._checkAvatarSlot();
+  }
+
```
cdransf:
```suggestion
  protected override updated(changedProperties: PropertyValues): void {
    super.updated(changedProperties);
    this._checkAvatarSlot();
  }
```

I think we can remove this and add `this._checkAvatarSlot();` to the end of `firstUpdated` since updated fires on every property change.

cdransf (approved):
A few minor things, looks good! ✨

2nd-gen/packages/swc/components/user-card/test/user-card.a11y.spec.ts
aramos-adobe: I'm going to keep this one for now. I think card a11y spec has to be tested against webkit. I failed this test specifically a few times on this use case.

PR #6747, opened 2026-09-15
ci(release): add a throwaway snapshot job to the gen2 pipeline

Rajdeepc:
## Description

- gen2 had no snapshot channel; testing pending changes meant cutting a real `beta` release.
- Adds a manual `workflow_dispatch` job that publishes a throwaway build under `next` (or a custom tag). The `beta` pre-release state is untouched.
- The snapshot job deletes `pre.json` on its own ephemeral checkout only, so changesets takes the normal snapshot path (changesets blocks `--snapshot` while `pre.json` is present). Safety is structural: `permissions: contents: read`, no push/commit/tag steps, and `2nd-gen/.changeset/pre.json` on `main` is never read back or modified.
- `version-pr` and `publish` (the real release path) gained `!inputs.branch`, so a branch-scoped dispatch can never drive a release against `main`.

### How to trigger a snapshot release

1. Go to **Actions → Publish Packages (gen2) → Run workflow**.
2. Leave the branch selector on `main`.
3. Set **`branch`** to the branch you want to snapshot (dispatching from `main` means that branch doesn't need to carry this workflow file itself).
4. Set **`tag`** (defaults to `next`).
5. Run. Publishes `<version>-<tag>.<datetime>`; `beta`/`latest` are unaffected. Fork branches aren't reachable (internal-only).

## Motivation and context

Requested so a consumer team can validate in-progress gen2 work from a feature branch without waiting for a real beta release.

## Related issue(s)

- N/A

N/A: CI only, no UI change.

## Author's checklist

- [ ] I have reviewed at the Accessibility Practices for this feature: N/A, no UI surface.
- [ ] I have added automated tests to cover my changes: N/A, validated by parsing the workflow and asserting each job's guard, checkout ref, and permissions.
- [ ] I have included a well-written changeset: N/A, CI-only, no published package changed.
- [x] I have included updated documentation if my change required it: the job carries inline rationale.

### Manual review test cases

- [ ] _Snapshot a feature branch that does not have this workflow file_
  1. Actions → **Publish Packages (gen2)** → Run workflow on `main`, `tag=next`, `branch=<a branch with a pending 2nd-gen changeset>`
  2. Expect: publishes `...-next.<datetime>` built from that branch; `version-pr` and `publish` both skip; `beta` unchanged; no new commit, tag, or PR

- [ ] _Release path still main-only_
  1. Push to `main` under `2nd-gen/**` with pending changesets
  2. Expect: `version-pr` opens/updates as today; `snapshot` does not run

- [ ] _Beta state intact after a snapshot_
  1. Run a snapshot, then diff `2nd-gen/.changeset/pre.json` on `main`
  2. Expect: no change

### Device review

- [ ] Desktop / Mobile / iPad: N/A, no UI

N/A: CI workflow configuration only; no component, page, or interactive surface affected.


.github/workflows/publish-gen2.yml
```diff
@@ -2,6 +2,19 @@ name: Publish Packages (gen2)
...
+      tag:
+        description: 'NPM dist-tag for a throwaway gen2 snapshot'
+        required: false
+        default: 'next'
+        type: choice
+        options:
+          - next
+          - snapshot-test
+      branch:
+        description: 'Branch to snapshot. Leave blank to use the branch you dispatched from. Set it to snapshot a branch that does not carry this workflow file yet.'
```
blunteshwar: Since this is a `workflow dispatch` shouldn't the branch name be a required field.

.github/workflows/publish-gen2.yml
```diff
@@ -10,24 +23,30 @@ on:
 
 # Scoped to this workflow so a gen1 run and a gen2 run never block on each other -
 # both used to share the same `publish-${{ github.ref }}` group string.
+# Keyed on the branch actually being operated on, so snapshots of two different feature
+# branches run in parallel. For a push (no `branch` input) this falls back to github.ref
+# and behaves exactly as before: a main release still serializes against a main snapshot.
 concurrency:
-  group: publish-gen2-${{ github.ref }}
+  group: publish-gen2-${{ github.event.inputs.branch || github.ref }}
```
blunteshwar: Great Design

.github/workflows/publish-gen2.yml
```diff
@@ -60,11 +79,128 @@ jobs:
...
+  # own ephemeral checkout before versioning. That copy dies with the runner: the
+  # committed pre state on `main` is never modified or pushed back, so the beta channel's
+  # version accounting (initialVersions + consumed changesets) is untouched. Snapshot
+  # versions are `<calculated>-<tag>.<datetime>` per this gen's `snapshot` config, so they
+  # can never collide with a real `2.0.0-beta.N` or with the eventual stable `2.0.0`.
+  snapshot:
+    needs: check-changesets
+    if: >-
+      github.event_name == 'workflow_dispatch' &&
+      needs.check-changesets.outputs.has_changesets == 'true'
```
blunteshwar: Can we add another trigger of adding PR label so that this workflow could be triggered when someone adds a label `snapshot-gen2` on their PR.

.github/workflows/publish-gen2.yml
```diff
@@ -60,11 +79,128 @@ jobs:
...
+        uses: ./.github/actions/setup-job
+
+      - name: Determine snapshot tag
+        id: extract-tag
+        env:
+          INPUT_TAG: ${{ github.event.inputs.tag }}
+        run: |
+          WORKFLOW_TAG="${INPUT_TAG:-next}"
+          echo "tag=$WORKFLOW_TAG" >> $GITHUB_OUTPUT
+          echo "Using npm tag: $WORKFLOW_TAG"
```
blunteshwar, line 128: I think we should remove this step and stick to a single npm tag `snapshot`. Releasing under `next` tag with a snapshot release is a bit odd because a `next` tag release should be done from a main branch.

.github/workflows/publish-gen2.yml
blunteshwar: Can we ensure that this `snapshot` job should exit if manually triggered from `main` branch.

.github/workflows/publish-gen2.yml
```diff
@@ -60,11 +79,128 @@ jobs:
...
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.ADOBE_BOT_NPM_TOKEN }}
+        run: npm whoami --registry=https://registry.npmjs.org
+
+      - name: Leave pre mode for this runner only
+        run: |
+          # Ephemeral by construction: this checkout is discarded when the job ends and
+          # nothing below commits or pushes. Removing the file locally is the only thing
+          # that lets `changeset version --snapshot` run at all while gen2 is in pre mode.
+          rm -f 2nd-gen/.changeset/pre.json
```
blunteshwar: Instead of manually deleting the `pre.json` we can run `yarn changeset pre exit` to formally exit the pre-release mode.

.github/workflows/publish-gen2.yml
```diff
@@ -60,11 +79,128 @@ jobs:
...
+
+      - name: Version gen2 packages (snapshot)
+        working-directory: 2nd-gen
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: yarn changeset version --snapshot ${{ steps.extract-tag.outputs.tag }}
```
blunteshwar, line 160: Should we add `cd 2nd-gen ` before we run `yarn changeset version --snapshot ${{ steps.extract-tag.outputs.tag }}`

```suggestion
        run: cd 2nd-gen && yarn changeset version --snapshot ${{ steps.extract-tag.outputs.tag }}
```

.github/workflows/publish-gen2.yml
```diff
@@ -60,11 +79,128 @@ jobs:
...
+        run: |
+          yarn install --refresh-lockfile
+          yarn build
+
+      - name: Publish gen2 packages (snapshot)
+        if: steps.npm-auth.outcome == 'success'
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.ADOBE_BOT_NPM_TOKEN }}
+        working-directory: 2nd-gen
+        run: yarn changeset publish --no-git-tag --tag ${{ steps.extract-tag.outputs.tag }}
```
blunteshwar, line 183: Here also we should add `cd 2nd-gen`.

```suggestion
        run: cd 2nd-gen && yarn changeset publish --no-git-tag --tag ${{ steps.extract-tag.outputs.tag }}
```

blunteshwar (changes requested):
The design is solid. I’ve left a few comments, and I believe we’re good to go after that.

.github/workflows/publish-gen2.yml
Rajdeepc: Good call. `branch` is now required, and snapshots are feature-branch-only. Release path is unaffected (`inputs` is empty on push).

.github/workflows/publish-gen2.yml
Rajdeepc: 🙏

.github/workflows/publish-gen2.yml
Rajdeepc: Love the ergonomics, but I'd like to do it as a separate PR — a `pull_request: [labeled]` trigger that needs the npm token has to be written carefully to avoid fork-PR secret exposure (same class of thing that bit the gen2 docs `workflow_run` earlier). Filing it as a follow-up so this stays a clean unit.

.github/workflows/publish-gen2.yml
Rajdeepc: Agreed. `next` is fed by main, so a branch snapshot under it would move the shared `next` dist-tag onto unmerged code. Defaulting to `snapshot` now, `next` removed.

.github/workflows/publish-gen2.yml
Rajdeepc: It's feature-branch-only now: a guard step rejects `main` (or an empty target) with a clear error before checkout.

.github/workflows/publish-gen2.yml
Rajdeepc: Left it as `rm -f` on purpose. Both clear the guard (changesets only blocks `--snapshot` when `mode === 'pre'`), but `rm -f` is idempotent where `pre exit` throws if a branch isn't in pre mode, and it leaves no leftover pre state. The 'formally exit' benefit only matters for a persisted repo — this checkout is ephemeral with `contents: read` and is never pushed. Happy to switch if you feel strongly.

.github/workflows/publish-gen2.yml
Rajdeepc: This step already sets `working-directory: 2nd-gen`, which is the same as `cd 2nd-gen` — adding the `cd` would run from `2nd-gen/2nd-gen` and fail, so I've left it as-is 🙂

.github/workflows/publish-gen2.yml
Rajdeepc: Same here — `working-directory: 2nd-gen` on this step already handles it.

.github/workflows/publish-gen2.yml
```diff
@@ -60,11 +79,140 @@ jobs:
...
+      - name: Reject main as a snapshot target
+        env:
+          TARGET: ${{ github.event.inputs.branch }}
+        run: |
+          # Snapshots are feature-branch-only. `branch` is a required input, but this also
+          # rejects someone typing `main` into it - main ships through the real beta flow,
+          # never a throwaway snapshot.
+          if [ -z "$TARGET" ] || [ "$TARGET" = "main" ] || [ "$TARGET" = "refs/heads/main" ]; then
+            echo "::error::gen2 snapshots are for feature branches only. Set 'branch' to your feature branch; main releases via the beta flow."
+            exit 1
```
blunteshwar: Instead of creating a separate step for `Reject main as a snapshot target` and exiting with an error `exit 1`, we should incorporate these checks in the if step (line 105) to prevent the job from starting in the first place when the branch is main. This is because there might be instances when we want to trigger the workflow `publish-gen2.yml` from the main branch, and in such cases, the snapshot job will throw an error instead of being skipped.

.github/workflows/publish-gen2.yml
blunteshwar:
```suggestion
    if: >-
    github.event_name == 'workflow_dispatch' &&
    needs.check-changesets.outputs.has_changesets == 'true' &&
    github.event.inputs.branch != 'main' &&
    github.event.inputs.branch != 'refs/heads/main'
```

.github/workflows/publish-gen2.yml
```diff
@@ -123,9 +277,11 @@ jobs:
   # since it's the only one that actually publishes to npm.
   publish:
     needs: check-changesets
-    # No event/branch check needed: push and workflow_dispatch are this workflow's only
-    # triggers, and check-changesets (which this needs) already skips any non-main run.
+    # Main-only, and never on a branch-targeted dispatch - see the matching comment on
+    # version-pr above for why `!inputs.branch` is load-bearing here.
     if: >-
+      github.ref == 'refs/heads/main' &&
+      !github.event.inputs.branch &&
```
blunteshwar: We should remove this as this will not allow us to manually trigger this from main

.github/workflows/publish-gen2.yml
```diff
@@ -73,9 +221,15 @@ jobs:
...
-    # triggers, and check-changesets (which this needs) already skips any non-main run.
-    if: needs.check-changesets.outputs.has_changesets == 'true'
+    # Main-only: check-changesets is no longer branch-gated (so snapshot can target a
+    # feature branch), so the real release path carries that restriction itself.
+    # `!inputs.branch` matters as much as the ref check: when a branch is supplied,
+    # check-changesets counted THAT branch's changesets, so its outputs must not be
+    # allowed to drive a release against main.
+    if: >-
+      github.ref == 'refs/heads/main' &&
+      !github.event.inputs.branch &&
```
blunteshwar: We should remove this as this will not allow us to manually trigger this from main

blunteshwar (changes requested):
I think we are in a good shape here. Just a couple of thing to ponder here, first is the case when we manually want to trigger the workflow from main to do an actual release then the snapshot will throw error and the jobs `version-pr` and `publish` will be skipped. Second when we manually trigger workflow from main the `version-pr` and `publish` job will be skipped bcz of `!github.event.inputs.branch` check.

.github/workflows/publish-gen2.yml
```diff
@@ -60,11 +72,141 @@ jobs:
...
+        run: |
+          # Ephemeral by construction: this checkout is discarded when the job ends and
+          # nothing below commits or pushes. Removing the file locally is the only thing
+          # that lets `changeset version --snapshot` run at all while gen2 is in pre mode.
+          rm -f 2nd-gen/.changeset/pre.json
+
+      - name: Ensure a changeset exists
+        run: |
+          # `changeset version --snapshot` exits without versioning when nothing is
+          # pending. So a branch can be snapshotted regardless of changesets, synthesize a
```
blunteshwar, line 146: nit:
is there a use case when we want to do a throwaway release without a changeset present?

blunteshwar (approved):

.github/workflows/publish-gen2.yml
Rajdeepc: If this a feature branch release, there will be many cases where there will be no changesets present typically while we do internal changes + tooling.

.github/workflows/publish-gen2.yml
```diff
@@ -60,11 +72,141 @@ jobs:
...
+          echo "|-------|-------|" >> $GITHUB_STEP_SUMMARY
+          echo "| **Snapshotted branch** | \`${TARGET_REF}\` |" >> $GITHUB_STEP_SUMMARY
+          echo "| **NPM tag** | \`${TAG}\` |" >> $GITHUB_STEP_SUMMARY
+          [ -n "$SWC_VERSION" ] && echo "| **@adobe/spectrum-wc** | \`${SWC_VERSION}\` |" >> $GITHUB_STEP_SUMMARY
+          [ -n "$CORE_VERSION" ] && echo "| **@adobe/spectrum-wc-core** | \`${CORE_VERSION}\` |" >> $GITHUB_STEP_SUMMARY
+          [ -n "$ICONS_VERSION" ] && echo "| **@adobe/spectrum-wc-icons** | \`${ICONS_VERSION}\` |" >> $GITHUB_STEP_SUMMARY
+          echo "" >> $GITHUB_STEP_SUMMARY
+
+          if [ "$AUTH_OUTCOME" == "success" ]; then
+            echo "✅ Throwaway snapshot published under \`${TAG}\`. The \`beta\` channel and its pre-release state are unchanged." >> $GITHUB_STEP_SUMMARY
```
rubencarvalho: doesn't this always publish "Throwaway snapshot published." even if it fails?

.github/workflows/publish-gen2.yml
```diff
@@ -2,32 +2,44 @@ name: Publish Packages (gen2)
...
 
 jobs:
   check-changesets:
     runs-on: ubuntu-latest
-    # Both triggers reduce to "must be on main": push only fires on main (trigger
-    # above), and a workflow_dispatch is only allowed from main. dispatch from any
-    # other branch skips here, so version-pr/publish (which need this job) skip too.
-    if: github.ref == 'refs/heads/main'
+    # Deliberately NOT branch-gated: the snapshot job runs from a feature branch, so this
+    # job (which it needs) must run there too and count that branch's changesets. The
```
rubencarvalho: I think this comment is not accurate anymore?

.github/workflows/publish-gen2.yml
```diff
@@ -60,11 +72,141 @@ jobs:
...
+          rm -f 2nd-gen/.changeset/pre.json
+
+      - name: Ensure a changeset exists
+        run: |
+          # `changeset version --snapshot` exits without versioning when nothing is
+          # pending. So a branch can be snapshotted regardless of changesets, synthesize a
+          # throwaway one when none exist. It patches the linked lead package, which bumps
+          # the whole `@adobe/spectrum-wc(-core|-icons)` group together. Ephemeral checkout,
+          # never committed.
+          if [ -z "$(find 2nd-gen/.changeset -maxdepth 1 -name '*.md' ! -name 'README.md' -print -quit)" ]; then
```
rubencarvalho:
```suggestion
          if [ -z "$(find gen2/.changeset -maxdepth 1 -name '*.md' ! -name 'README.md' -print -quit)" ]; then
```

Like this one, there's a lot of renames needed. I won't comment on all to not spam, but just FYI!

rubencarvalho (commented):
I'm not gonna block the merge on this BUT ideally if we follow devops/security best practices we should be separating building/packaging from publishing. e.g., the feature branch would build the packages in a job with no credentials, then a trusted job running from `main` could publish those artifacts without executing any branch code. This would keep the npm credentials limited to the publish step. rn, something in a feature branch can access them... this is only ok-ish because we have the `npm-publish` environment approval, but still not idiomatic 😄

.github/workflows/publish-gen2.yml
Rajdeepc: I would have written the snapshot job against `2nd-gen/` before the rename landed on the branch, so it renamed the older jobs but left mine pointing at a directory that no longer exists. It would've failed on the first run. I have fixed all of them to `gen2/`.

.github/workflows/publish-gen2.yml
Rajdeepc: Yes, it keyed off the npm-auth step, so a failed `changeset publish` would still have printed a success line. I've given the publish step an `id` and gated the message on its actual outcome now. While I was in there I also made it point people at the exact version to install rather than the shared `@snapshot` tag.

.github/workflows/publish-gen2.yml
Rajdeepc: You're right, that comment was stale. Thanks for flagging it.

Rajdeepc:
> I'm not gonna block the merge on this BUT ideally if we follow devops/security best practices we should be separating building/packaging from publishing. e.g., the feature branch would build the packages in a job with no credentials, then a trusted job running from `main` could publish those artifacts without executing any branch code. This would keep the npm credentials limited to the publish step. rn, something in a feature branch can access them... this is only ok-ish because we have the `npm-publish` environment approval, but still not idiomatic 😄

I completely agree with you. 

The one wrinkle I keep hitting for the snapshot case specifically is the whole point is to publish a feature branch's code, so I can't fully avoid building branch code somewhere. The cleanest version is probably a no-credential build job that hands a tarball to a trusted publish job, rather than truly never touching branch code. 

I would take it as a follow-up covering the whole gen2 pipeline rather than bolting it onto this PR. 
Ticket to track this: SWC-2597

.github/workflows/publish-gen2.yml
```diff
@@ -60,11 +73,144 @@ jobs:
...
+  # in-repo branches, so fork code can never reach this job or the npm token.
+  #
+  # gen2 lives in changesets' persistent pre mode (`gen2/.changeset/pre.json`, tag
+  # `beta`), and changesets refuses `--snapshot` while that file is present ("Snapshot
+  # release is not allowed in pre mode"). This job therefore deletes pre.json from its
+  # own ephemeral checkout before versioning. That copy dies with the runner: the
+  # committed pre state on `main` is never modified or pushed back, so the beta channel's
+  # version accounting (initialVersions + consumed changesets) is untouched. Snapshot
+  # versions are `<calculated>-<tag>.<datetime>` per this gen's `snapshot` config, so they
+  # can never collide with a real `2.0.0-beta.N` or with the eventual stable `2.0.0`.
```
rubencarvalho: Actually... If someone had `^2.0.0-beta.3` or like `^2.0.0` I think this would actually colide with the resolution... So like `2.0.0-snapshot.20260916091811` may win?
What about we leave snapshot releases at `0.0.0-snapshot.xxxx...`? I think you'd need to unset `snapshot.useCalculatedVersion`.

.github/workflows/publish-gen2.yml
```diff
@@ -28,6 +39,8 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
         with:
+          # The dispatched/pushed ref - main for a release, the feature branch for a
```
rubencarvalho: I think after your last changes this will never run _from_ a feature branch?

.github/workflows/publish-gen2.yml
```diff
@@ -73,9 +219,12 @@ jobs:
   # same branch.
   version-pr:
     needs: check-changesets
-    # No event/branch check needed: push and workflow_dispatch are this workflow's only
-    # triggers, and check-changesets (which this needs) already skips any non-main run.
-    if: needs.check-changesets.outputs.has_changesets == 'true'
+    # Main-only: check-changesets is not branch-gated (so snapshot can run from a feature
```
rubencarvalho: this sounds contradicting?

.github/workflows/publish-gen2.yml
rubencarvalho: is this true? doesn't snapshot now sit behind npm-publish too?

rubencarvalho (commented):
I think we're very close! Can we just check on the version bit?

.github/workflows/publish-gen2.yml
Rajdeepc: You are right. A calculated `2.0.0-snapshot.<datetime>` sorts above `2.0.0-beta.3`, so it could satisfy and win a consumer's `^2.0.0-beta.x` range, which is the opposite of what I want for a throwaway build. I've set `snapshot.useCalculatedVersion: false` so snapshots come out as `0.0.0-snapshot.<datetime>` now, which sits below any real 2.x and can't be resolved into. I also rewrote that comment to explain the actual reasoning (6a7d7da). Thank you for pushing on this one.

.github/workflows/publish-gen2.yml
Rajdeepc: I gated check-changesets to main in a later change and left this comment behind. It never runs from a feature branch anymore, so I've corrected the wording to match. Thanks for catching it.

.github/workflows/publish-gen2.yml
Rajdeepc: Yup! same root cause as the other one. Once I gated check-changesets to main this comment stopped being true. I've reworded it so it no longer claims the job is un-gated.

.github/workflows/publish-gen2.yml
Rajdeepc: Yes really good catch. The snapshot job also sits behind `npm-publish` now, so I've dropped the "only" and noted both jobs go through the approval. Fixed in the same commit.

rubencarvalho (approved):

PR #6749, opened 2026-09-15
fix(picker): sync controller open state on external close

### LINKED ISSUE #6678: [Bug]: Action Menu (mobile Tray) doesn't reopen after being dismissed by an outside tap
### Code of conduct

### Impacted component(s)

Action Menu (`sp-action-menu`)

### Library version

1.12.2

### Expected behavior

On a mobile/touch device, tapping the action button should reopen the menu, even if the previous time it was open, it was dismissed by tapping outside of it rather than by selecting a menu item or moving focus elsewhere.

### Actual behavior

On a mobile device (or Chrome DevTools device emulation), opening the Action Menu renders it in a tray. If the tray is dismissed by tapping outside of it (instead of selecting a menu item), tapping the action button again does **not** reopen the tray. The button becomes unresponsive to taps until focus is moved to some other element and back.

### Screenshots

See linked Slack thread / screen recording (internal): reproduces directly on the SWC docs site action-menu example.

Chrome

1. Go to https://opensource.adobe.com/spectrum-web-components/components/action-menu/
2. Open the page with Chrome DevTools device emulation using an iPhone 16 Pro Max (or any touch/mobile emulation).
3. Tap the action button. The menu opens in a tray.
4. Tap outside the tray to dismiss it (without selecting a menu item).
5. Tap the action button again.
6. **Expected:** the tray reopens. **Actual:** nothing happens; the tray does not reopen until focus is moved elsewhere first.

```html
<sp-action-menu label="More Actions">
  <sp-menu-item>Deselect</sp-menu-item>
  <sp-menu-item>Select Inverse</sp-menu-item>
</sp-action-menu>
```

No special markup is required; this reproduces with the default Action Menu example on the docs site, as long as the mobile/touch `MobileController` code path is active (viewport matches the mobile media query).

### Severity

SEV 4

### Root cause (found during investigation)

In `packages/picker/src/Picker.ts`, `ExpandableElement.handleBeforetoggle` handles the overlay's native `beforetoggle` event when it closes. When the close is triggered externally (e.g. tapping outside the mobile Tray, which closes via the Tray's underlay `close` event rather than a click on the trigger button) while `strategy.preventNextToggle === 'no'`, the handler does:

```ts
} else if (this.strategy?.preventNextToggle === 'no') {
  this.open = false;
}
```

This sets the host's own `open` property (so the Tray visually closes) but never syncs `this.strategy.open`, unlike the sibling branch just above it (`if (!this.open) { ... this.strategy.open = false; }`), which does sync both.

Because of that, `MobileController`'s internally tracked `open` state (`InteractionController.open`, which is separate from `host.open`) stays stale at `true`. On the very next tap, `MobileController.handlePointerdown()` reads that stale value and sets `preventNextToggle = 'yes'` (believing the menu is still open), which causes `MobileController.handleClick()` to skip calling `toggle()` entirely - so the tap does nothing.

This doesn't affect desktop because `DesktopController.handleActivate()` unconditionally calls `host.toggle()`, which recomputes fresh from the host's own `open` property and resyncs `strategy.open` as a side effect (self-healing). `MobileController.handleClick()` instead gates the `toggle()` call itself on the stale controller state, so it never reaches that self-healing path.

**Suggested fix:** also sync `this.strategy.open = false;` in that `else if` branch, mirroring the branch above it.

### Would you like to track this issue in Jira?

- [x] Yes, please tell me the ticket number!

rubencarvalho:
## Description

In `handleBeforetoggle` (`packages/picker/src/Picker.ts`), the branch that handles an external close (e.g. tapping outside the mobile Tray) while `strategy.preventNextToggle === 'no'` set the host's `open` to `false` but never synced the interaction controller's `open` state:

```ts
} else if (this.strategy?.preventNextToggle === 'no') {
    this.open = false;
    this.strategy.open = false; // added
}
```

Because the controller's `open` stayed stale at `true`, the next `MobileController.handlePointerdown()` set `preventNextToggle = 'yes'` and `handleClick()` skipped `toggle()`, so the Tray never reopened. The fix syncs `strategy.open`, matching the sibling `if (!this.open)` branch just above it.

Desktop was unaffected because `DesktopController.handleActivate()` unconditionally calls `host.toggle()`, which self-heals the controller state.

## Motivation and context

On a mobile/touch device, dismissing an `sp-action-menu` Tray by tapping outside left the action button unresponsive to taps until focus moved elsewhere and back. Tapping the action button should always reopen the menu.

## Related issue(s)

- fixes #6678
- SWC-2543

## Testing

New end-to-end test `packages/action-menu/test/action-menu-reopen.test.ts`: opens the Tray by tapping the button, dismisses it by tapping outside (no selection), then taps the button again and asserts the Tray reopens. This exercises the real outside-tap close path (where `open` is still `true` at `beforetoggle`, hitting the patched branch). The test fails without the fix and passes with it.

Note on browser coverage: with synthetic pointer events the reopen failure reproduces on Firefox; Chromium/WebKit take the already-synced branch under synthetic events. The fix targets the exact branch the outside-tap close takes on real touch input.

## Author's checklist

---

### Manual review test cases

- [ ] Mobile Tray reopens after outside-tap dismiss
  1. Open the `main` [Action Menu docs](https://opensource.adobe.com/spectrum-web-components/components/action-menu/) with Chrome DevTools device emulation (e.g. iPhone) and then compare to this fix: https://swcpreviews.z13.web.core.windows.net/pr-6749/docs/gen1-docs/components/action-menu/.
  2. Tap the action button; the menu opens in a Tray.
  3. Tap outside the Tray to dismiss it (without selecting an item).
  4. Tap the action button again.
  5. Expected: the Tray reopens.


blunteshwar (approved):
LGTM!

Rajdeepc (approved):

PR #6750, opened 2026-09-15
fix(menu): reset mobile drill-down when containing tray closes

rubencarvalho:
## Description

Resolves **SWC-2525**.

In `mobile-view` (drill-down navigation), the submenu stack was only cleared on `disconnectedCallback` or when `mobile-view` was turned off. A dismissed `<sp-tray>` stays connected, so drilling into a submenu, dismissing the tray, and reopening it landed the user back in the stale submenu instead of the top-level menu.

## What changed

- **`Menu.ts`** — while drilled into a submenu, `<sp-menu mobile-view>` listens for the containing `<sp-tray>`'s bubbling `close` event and calls `resetMobileSubmenus()` on dismiss (click-outside, Escape, or programmatic close). The listener is attached to the resolved tray element (not `document`), scoped to the drill-down lifecycle, and removed once the stack empties. Desktop flyout submenus are unaffected.
- **`submenu.stories.ts`** — the `mobileView` story no longer force-reopens the tray on `close`; an "Open menu" button reopens it, so the reset is visible in the published docs.
- **`README.md`** — clarifies that `mobile-view` menus are composed inside a customer-owned `<sp-tray>` and that dismissing the tray resets the drill-down.
- Added a changeset.

## Scope note

The reset hooks the `<sp-tray>` `close` event, which is the documented mobile pattern. A bare `<sp-overlay>` wrapper emits `sp-closed` (not `close`) and is intentionally out of scope here.

## How to test

1. Open the Menu **mobile-view (drill-down navigation)** docs/story with touch emulation: https://swcpreviews.z13.web.core.windows.net/pr-6750/docs/gen1-docs/components/menu/#mobile-view-(drill-down-navigation). You can compare it to the `main` one: https://opensource.adobe.com/spectrum-web-components/components/menu/#mobile-view-(drill-down-navigation)
2. Drill into a submenu, dismiss the tray, reopen it.
3. The menu shows the top-level items, not the previous submenu.


pfulton:
Reviewers: can we also verify via Browserstack that this is working in mobile (Android, iOS) browsers? You should be able to hit the deploy URL from the Browserstack instances.

Rajdeepc (approved):

blunteshwar (approved):
Checked and verified in my mobile phone. The bug is absent.

PR #6753, opened 2026-09-15
chore(toast): setup, closing decisions, container planning

rise-erpelding:
## Description

Delivers the Setup phase of the Toast migration: scaffolds the gen2 file structure (`gen2/packages/core/components/toast/`, `gen2/packages/swc/components/toast/`), with no functional implementation yet. Also refines the migration plan merged in #6686 ahead of the API implementation phase:

- Records four resolved API decisions in the Decision log: timeout floors to 6000ms (Q1), `timeout` is hard-disabled when the `action` slot has content, checked once at open time (Q2), the light-DOM `action` slot is kept rather than switching to props (Q3), and timer pause is region-wide rather than per-toast (Q7).
- Reworks the plan's sequencing: resolving the container/queue architecture (Q5, Q8) is now a near-term blocker for most remaining API and Accessibility work, not a deferred item, and scopes a placeholder follow-up ticket for that construct.
- Adds a new accessibility finding (Q10) from manual VoiceOver testing: the inner `role="alert"` live region only announces reliably in Safari and Edge with VoiceOver when its message content is set after the node already exists, empty, in the accessibility tree. Flags a follow-up open question (Q11) to verify that finding against slotted content and a toggled `open` lifecycle.
- Proposes the container's shape as a new `swc-toast-container` element (Q13): its core layer owns the queue's countdown state and a decoupled `add()`/`close()`/`clear()` export, its SWC layer owns the region/peek-stack/expanded-view rendering, with `placement` (default `bottom`) and a one-container-per-view dev warning. How an action button and an `onClose` callback flow through that export is still open.
- Adds Q14, resolved: peek-stack layers render as wholly empty, non-interactive elements, matching React Spectrum S2's actual source, which settles a tab-order concern raised in review (`role="presentation"` alone doesn't reliably exclude real interactive content from Tab).
- Adds Q15, leaning documented: `swc-toast` likely has no timer of its own; `timeout` only takes effect once paired with `swc-toast-container`, since a fully standalone toast was never a completely accessible configuration once more than one toast exists.

## Motivation and context

Establishes the file structure and finalized design decisions Toast's API implementation needs before that work starts, so it isn't blocked mid-flight on open questions.

## Related issue(s)

- Related to SWC-2258. This is one of a series of PRs against that ticket (Setup, API, and Accessibility); this PR covers Setup only.

## Author's checklist

- [x] I have read the [CONTRIBUTING](https://github.com/adobe/spectrum-web-components/blob/main/CONTRIBUTING.md) and [PULL_REQUESTS](https://github.com/adobe/spectrum-web-components/blob/main/PULL_REQUESTS.md) documents.
- [ ] I have reviewed the Accessibility Practices for this feature, see: [Aria Practices](https://www.w3.org/TR/wai-aria-practices/)
- [ ] I have included a well-written changeset if my change needs to be published. *(Skipped: no publishable behavior yet; deferred to the API PR.)*

## Reviewer's checklist

### Manual review test cases

- [ ] _Plan reads consistently_
  1. Open the [Toast migration plan](CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/migration-plan.md)
  2. Confirm Q1/Q2/Q3/Q7/Q10/Q14 read as resolved, and Q5/Q8/Q9/Q11/Q12/Q13/Q15 read as the active blockers
- [ ] _Scaffold builds cleanly_
  1. Run the project build with the new `toast` package files present
  2. Confirm no compile errors from the stub `Toast.ts`/`Toast.base.ts` files


CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/migration-plan.md
```diff
@@ -178,18 +173,22 @@ Already-fixed gen1 bugs whose behavior must not regress in gen2 (not listed as r
...
+
+Internally, though, resolving Q5 and Q8 is the near-term prerequisite for most of the remaining work. A narrow single-toast skeleton (variant/icon-label validation, `action`-slot wiring via `SlotPresenceController`/`SlotAttributePropagationController`, the close button, lifecycle events via `runAfterTransition`, and the A1-A3/A7/`tabindex` accessibility items) is independent and can proceed now. Everything else, timeout/pause behavior (Q1/Q2/Q7), the API checklist's container-and-queue bullets, and nearly all of the Accessibility checklist (focus management across toasts, focus-on-queue-empty, collapse-to-region, landmark navigation), depends on the container/queue construct actually existing, so it waits on Q5/Q8.
```
cdransf, line 180: > ...so it waits on Q5/Q8

Should the table on line ~533 have the blocking column updated to `Yes`? ✨

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/migration-plan.md
```diff
@@ -57,21 +57,16 @@
...
 - 1st-gen defines 5 variant values (`negative`, `positive`, `info`, `error`, `warning`); `error` and `warning` are already deprecated aliases of `negative` in 1st-gen and do not carry forward. gen2 has 4 variants: `neutral`, `info`, `positive`, `negative`.
-- Points of disagreement, see [Design](#design) for more detail:
-    - **Timeout minimum** (6000ms vs. 5000ms)
-    - **Action+timeout** (soft warning vs. hard block)
-    - **Action API shape** (light-DOM `action` slot vs. `actionLabel`/`onAction` props)
-    - **Timer pause scope** (per-toast vs. region-wide)
+- Q1, Q2, Q3, Q7, Q10, and Q14 are resolved; see the [Decision log](#decision-log). Q9, Q12, and Q13 remain open; Q12 and Q13 have a documented leaning, see below.
```
cdransf: Should `Q15` be here since it's listed as blocking on line ~525? ✨

cdransf (approved):
Looks great! Just a few nits ✨

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/migration-plan.md
rise-erpelding, line 107: Further investigation has revealed that the correct component from a visual standpoint to slot in here would be `swc-button`, not action button.

rise-erpelding, line 149: Unsure if we would actually need this, we're not doing sizing for toast and the variant, I believe, would always be the same (the secondary outline static white button)

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/migration-plan.md
```diff
@@ -226,16 +227,16 @@ Whether the queue's countdown construct belongs in a shared, cross-component loc
 | A1 | Host role | `role="alert"` on inner `.body` div only | `role="alertdialog"` + `aria-modal="false"` on host; inner `role="alert"` `aria-atomic="true"` | None |
 | A2 | Host naming | None | `aria-labelledby` referencing a content-element ID, falling back to `aria-label` from default-slot text | None |
```
rise-erpelding, line 228: I don't believe we have a specific checklist item for documenting this, feels like it would be a good idea to have one.

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/migration-plan.md
```diff
@@ -57,21 +57,18 @@
...
-- **Q7** in [Design](#design): timer pause scope. Per-toast (a11y doc's current recommendation) vs. region-wide (RSP S2 reality).
+- **Q5** in [Architecture and behavior](#architecture-and-behavior): queue-level countdown ownership and location. The queue construct must be decided before the dependent container, timeout, and accessibility work can proceed; the current recommendation is to keep it specific to `swc-toast-container` until a second consumer requires extraction.
+- **Q8** in [Architecture and behavior](#architecture-and-behavior): whether repositioning between peek, front, and expanded-list positions uses View Transitions or an instant swap. The choice affects the container's rendering and transition implementation.
 - **Q9** in [Design](#design): message content slot. Default slot (current plan, may lack an ID-bearing wrapper) vs. a named slot (guarantees one).
+- **Q12** in [Design](#design): countdown timing. Leaning toward starting the moment a toast is added to the queue and running continuously regardless of render state, matching RSP S2, but still open for comment.
```
5t3ph, line 69: Testing RSP behavior, if I add multiple of the auto-dismissing (timed) toasts what happens is:
1. timeout only runs if the toast is in view, so if another one gets queued to the front of the stack, the previous timed one will wait to be front of the stack
2. if the stack is expanded and contains timed toasts, given they become visible, then they dismiss in timeout order
3. hover pauses timeout (RSP has a bug where given a stack of timed toasts, then on mouseout from a hover the timeout resumes for _all_, not just the top toast, leading to full dismissal)
4. given a stack of timed toasts, when expanded, after toasts dismiss in timeout order, the "last" one (top of the stack) requires manual dismissal, presumably since the user is currently interacting in some capacity since they opened the stack

I think the logic is pretty good (given the hover timeout-resume is corrected) except maybe not dismiss toasts at all when in the expanded state. That feels like a violation of [Timing Adjustable](https://www.w3.org/WAI/WCAG21/Understanding/timing-adjustable) where their "expand" interaction I think should be taken as a signal they want to read and take action on the toasts (turn off/extend the timeframe), but the timeout likely would be too quick to individually consume them.

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/migration-plan.md
```diff
@@ -57,21 +57,18 @@
...
+- **Q13** in [Architecture and behavior](#architecture-and-behavior): container naming and API shape. Proposed `swc-toast-container` with a decoupled queue export for showing toasts, open for feedback.
+- **Q15** in [Design](#design): standalone auto-dismiss. Leaning toward `swc-toast` having no timer of its own at all, `timeout` only takes effect once paired with `swc-toast-container`, since a standalone toast was never a fully accessible configuration to begin with. Still open for confirmation.
```
5t3ph, line 71: Agree that `timeout` only has an effect as coordinated with the container, per previous comment of RSP observations.

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/migration-plan.md
5t3ph: That seems like enough justification to ensure the expected button variant _if_ we do in fact keep the slot.

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/migration-plan.md
```diff
@@ -296,20 +297,40 @@ No properties are exposed in the initial set. Add a `--swc-toast-*` property onl
...
+
+**Still open as part of Q13:** how an action button is supplied through the imperative queue export, since `action` is a light-DOM slot (Q3), not a prop. One option is accepting a real element/node in the call's options for the container to slot in; the other is treating actionable toasts as the hand-authored-markup case above, and leaving the queue export to cover the plain-message case only. Also open: whether the queue export accepts an `onClose` callback per call, matching RSP S2's `addToast()`, which fires whenever that specific toast closes, for any reason.
```
5t3ph, line 316: I'm not sure we need to support the standalone case? That's putting all of the burden on the consumer to figure out the a11y pieces and screen placement, which is precisely why we decided to go ahead and create the container.

Either way, I don't think we should stand firm on the "action" slot vs. allowing the basic necessary action-related props for the queue, especially since there is a specific variant/size to be locked in for Toast, so consumers don't actually need access to the full button API. Icons can reasonably be disallowed so they don't compete with toast UI. I think I'm reversing my own earlier statement on the original plan, but the container + queue concept changes it.

Image

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/migration-plan.md
```diff
@@ -296,20 +297,40 @@ No properties are exposed in the initial set. Add a `--swc-toast-*` property onl
...
 - Every queued toast stays tracked in the queue, with no cap; only a few are ever rendered as full content. RSP S2's own `ToastQueue` is explicitly constructed with `maxVisibleToasts: Infinity` (the primitive's own default is 1, a single visible toast; S2 overrides it), and only the front two peek positions plus the front toast itself render as anything visible.
+- Leaning (Q12, open for comment): a toast's countdown starts the moment it's added to the queue and runs continuously afterward, regardless of whether it's currently the front toast, a peek layer, or past the render cutoff; it does not pause on demotion, only the region-wide pause above affects it. This matches RSP S2's own `Timer` implementation (`packages/react-stately/src/toast/useToastState.ts`) and is a direct consequence of there being no real visibility cap, above. Accepted trade-off: a toast can in principle expire without ever having been the front-most toast, if new ones keep arriving quickly.
```
5t3ph, line 325: See earlier comment, I would heavily push back older toasts expiring before being viewed is "an accepted trade-off" as a violation of cognitive accessibility and [timing adjustable](https://www.w3.org/WAI/WCAG21/Understanding/timing-adjustable).

Note that the timing adjustable doc actually gives a toast example _however_ the critical part is that the auto-dismiss is really only acceptable if the information is available in another way (ex. appearance of a message in an inbox, or a more persistent notifications panel). We should cover that condition in our docs.

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/migration-plan.md
```diff
@@ -296,20 +297,40 @@ No properties are exposed in the initial set. Add a `--swc-toast-*` property onl
...
-- An expand control appears on the front toast once two or more toasts are queued, laid out on its own row below the message with the action button (RSP S2: `gridTemplateAreas: ['content content content', 'expand . action']`). Activating it moves focus to that toast's own host (since the control itself disappears once expanded) and opens every toast into a full list, where each becomes its own real alertdialog.
+- Peek stack: collapsed with two or more toasts queued, only the front toast renders as a real, interactive alertdialog; the two behind it are wholly separate, empty decorative layers with no focusable descendants (Q14), not the real toast's markup dimmed or covered. Anything further back stays in the DOM at `opacity: 0` rather than being removed (RSP S2: `opacity: index >= 3 ? 0 : 1`), to support a smooth transition if it's later promoted forward; see Q8.
```
5t3ph, line 328: I'm curious to see this play out in implementation, I'm not sure why it can't stay populated, just with `tabindex` management? Even slowing down RSP's animation to 10% it still seems like content is just fading in/out so removing it/repopulating seems like extra hoops, but we'll probably learn more as we get into using view transitions!

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/migration-plan.md
```diff
@@ -318,7 +339,7 @@ See [Toast accessibility migration analysis](./accessibility-migration-analysis.
...
... expanded view takes over the screen: a dismissible scrim, <kbd>Escape</kbd>, and a dedicated Collapse control all collapse it, and focus is contained within it while open. `swc-popover`'s `modal` property is the gen2 precedent for exactly this: opting into native `<dialog>.showModal()` gives a native focus trap, native background inert, and a native `::backdrop` scrim with no hand-rolled focus-trap implementation needed (RSP S2 does the equivalent with `FocusScope`/`useModalOverlay`, a React-only mechanism with no gen2 counterpart). `Escape` handling should go through the shared `dismissibleStack` (`registerDismissible()`/`isTopDismissible()`) so it coordinates correctly with a popover or tooltip also open at the same time, and the screen-takeover itself should use `PageScrollLockController` for the scroll lock; `swc-popover`'s modal mode already wires up both, all already built, see [Dependencies](#dependencies). Collapsing moves focus to the container region, not to whatever was focused before expanding (RSP S2: `collapse()` calls `regionRef.current?.focus()`). The region carries `tabindex="-1"` so it's programmatically focusable for this without joining the normal Tab sequence.
```
5t3ph, line 342: I'm not sure the expanded stack should coordinate with `dismissableStack` if we are using modal dialog behavior because the inert feature will block any background page popover or tooltip based items anyway. We should perhaps actually force close other popover/tooltip things if toast is expanded because of issue of the inert behavior. We'll need to set up some scenarios and test throughly.

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/migration-plan.md
```diff
@@ -488,18 +522,20 @@ Checklist items sourced from [accessibility-migration-analysis.md](./accessibili
...
+| Q11 | Q10's ordering requirement (the live region's content must be set at least one frame after the node exists, empty, in the accessibility tree) has only been verified against a single toast whose alert node owns its text directly in the shadow tree, freshly created per toast. Still needs verification against: message content delivered through the default slot (light DOM, per Q9) and projected into the alert wrapper; and a persistent host whose `aria-hidden` and content toggle with `open`, rather than the node being created fresh each time. | No | Open | Accessibility reviewer |
```
5t3ph, line 537: Ensuring this will have to be a priority before we get too deep into the implementation. We might have to resort to managing a separate live region as a light DOM sibling if the content population timing is not reliable cross-browser.

5t3ph (changes requested):

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/migration-plan.md
rubencarvalho: This is interesting because it sounds like we'd be changing the popover type on expanding? Also, it'd be interesting to understand how Toast would play along an open modal dialog (current scenario in Express, where e.g. you're performing an action that triggers a success/error notification). I'm not even sure what the desired UX is in this case. But anyway +1 to Steph's comment on the testing - I'm just unsure what we'd exactly want.

CONTRIBUTOR-DOCS/03_project-planning/03_components/toast/migration-plan.md
rubencarvalho: It'd be probably nice to confirm with design, but it feels to me the "auto dismiss" after expand should not be a thing? It feels like there may be too many toasts and therefore we would not know which one is being read?

Also, I noticed there's another bug in React where the toasts leave without animation when they time out in the stack.

rise-erpelding: Interesting. This observation re: the expanded state and auto-dismiss isn't actually what I saw initially when I explored React's auto-dismiss behavior, and I'm wondering if that comes down to a difference in browser behavior:

* Chrome macOS (where I looked initially) appears to pause toast auto-dismiss timers while pointer is in the window... so to me initially it looked like it didn't auto-dismiss at all when expanded
* Safari macOS is similar, it was even stickier with maintaining focus so didn't auto-dismiss at all ever
* Firefox & Windows Edge kept the timers running and auto-dismisses even if expanded and toast focused, which seems like a bug
* Safari iOS also keeps timers running and will auto-dismiss when expanded

I also see the thing you're saying about the dismissal animation being absent, @rubencarvalho.

I'll surface this with implementations to get confirmation with design and chat with RSP, but it seems like ensuring that turning off timers when expanded would be a good thing to build into the plan.

PR #6759, opened 2026-09-16
fix(icons): build declaration output for TS project references

rubencarvalho:
## Summary
- `@adobe/spectrum-wc-icons` declares `composite: true` and a TS project reference to `../core`, but its `build` script only ran `vite build`, so the `dist/src/*.d.ts` declaration output a `tsc --build` project reference expects was never produced.
- Any consumer resolving the package via TS `paths` mapping (e.g. the `swc` package's `@adobe/spectrum-wc-icons/*` → `../icons/src/*`) hits `TS6305: Output file ... has not been built from source file ...` as soon as it imports any icon, since the referenced project had never actually been built.
- Runs `tsc --build` ahead of `vite build` in the package's `build` script, and sets `emitDeclarationOnly` in its `tsconfig.json` so the `tsc --build` step only emits the missing declarations without duplicating the bundled JS output vite already produces.

Found while wiring up the first real runtime import of `@adobe/spectrum-wc-icons` from the `swc` package (in a separate branch); this fix is split out on its own since it's an independent build-config issue.

## Test plan
- [x] `yarn build` in `gen2/packages/icons` completes cleanly (tsc build + vite build + tree-shaking verification)
- [x] Reproduced the original `TS6305` error with a scratch import of `@adobe/spectrum-wc-icons/File.js` from the `swc` package and confirmed `tsc --noEmit -p .` no longer reports it after the fix
- [x] `yarn lint` (eslint/stylelint/prettier/ai checks) passes with no new errors


caseyisonit (approved):
thanks for catching this!

cdransf (approved):

PR #6764, opened 2026-09-17
feat(core): extract TriggerPressGuardController from Popover

Rajdeepc:
## Summary

- Extracts Popover's trigger-press reopen-guard into a new, reusable `TriggerPressGuardController` (`gen2/packages/core/controllers/trigger-press-guard-controller/`).
- **The bug it prevents**: a naive click-to-toggle handler (`open = !open`) breaks once the surface can close itself. Pressing the trigger again while open is an "outside" press from the surface's own perspective, so the native light-dismiss closes it *before* the trailing `click` fires. That click then reads `open` as already `false` and flips it back to `true`, silently reopening the surface on the very click that should have closed it.
- `Popover.base.ts` now composes the controller (`attach()`/`noteNativeDismiss()`/`detach()`) instead of carrying its own copy of the gesture-correlation state and listeners — net -97 lines in that file.
- Ships with unit tests, a Storybook demo (a working fix side-by-side with the naive bug for contrast), a docs page, and a `CONTRIBUTOR-DOCS` shared-resources entry.

## Context

Menu (`rajdeepchandra/feat-menu-styling-migration`, not yet merged) hit the exact same bug and needed the exact same fix. That PR ports the logic manually for now, since Menu's branch stack predates the `2nd-gen` → `gen2` rename and isn't merged yet. Once it rebases past the rename, it can switch to this controller instead of its own copy.

## Test plan

- [x] New controller unit tests: `yarn vitest --run --project storybook gen2/packages/core/controllers/trigger-press-guard-controller/` — 12/12 passing
- [x] Full Popover suite after the refactor: `yarn vitest --run --project storybook components/popover/` — 56/56 passing, no regressions
- [x] `tsc --noEmit` on `gen2/packages/core` — clean
- [x] `yarn eslint` / `yarn prettier --check` / `yarn lint:docs-pages` on all changed files — clean


gen2/packages/core/controllers/trigger-press-guard-controller/src/trigger-press-guard-controller.ts
```diff
@@ -0,0 +1,265 @@
...
+    this.detach();
+  }
+
+  // ─────────────────────────
+  //     IMPLEMENTATION
+  // ─────────────────────────
+
+  // Capture phase, before a native light-dismiss (which runs on the same
+  // press) gets a chance to close the surface first.
+  private readonly onPressStart = (): void => {
```
5t3ph, line 216: Copilot informs me:

> `onPressStart` can run twice for one touchscreen gesture (touchstart and pointerdown). It replaces `pressEndAbort` without aborting the previous controller. Because the document listener uses the same callback and capture flag, the second `addEventListener` is deduplicated; the original listener therefore remains permanently attached. Repeated touch gestures and disconnects can retain the controller through document. Abort the existing controller before creating a new one, and add a regression test covering touchstart followed by pointerdown.

5t3ph (changes requested):

gen2/packages/core/controllers/trigger-press-controller/trigger-press-controller.mdx
```diff
@@ -0,0 +1,99 @@
...
+### Basic usage
+
+1. Construct the controller in the host's constructor.
+2. Call `attach()` every time the resolved trigger element changes (for example from `updated()` reacting to a `for`/`triggerElement` property).
+3. Call `noteNativeDismiss()` from the surface's own native-close reaction, guarded by whatever the host already uses to tell a genuine native dismissal apart from its own programmatic close.
+
+```typescript
+import { LitElement, type PropertyValues } from 'lit';
+import { property } from 'lit/decorators.js';
+import { TriggerPressController } from '@adobe/spectrum-wc-core/controllers/trigger-press-controller.js';
```
rubencarvalho, line 31: aren't we missing the export of this in the core `package.json`

CONTRIBUTOR-DOCS/01_contributor-guides/16_gen2-shared-resources.md
```diff
@@ -46,6 +46,7 @@ Index of shared, reusable resources in gen2: controllers, mixins, utilities, and
 | `SlotTextController` | Observes whether a slot has meaningful text/element content | `card`, `button`, `action-button`, `illustrated-message`, `badge` |
 | `LiveSelectionController` | Coordinates single/multiple selection across event-dispatching children | `accordion` |
 | `PageScrollLockController` | Reference-counted page scroll lock for stacked blocking surfaces | `popover` |
+| `TriggerPressController` | Click-to-toggle for a trigger of a native light-dismissible surface, without the surface reopening on the click that closed it | `popover` |
```
rubencarvalho, line 49: There's a second inventory table that also needs this row: the **Available controllers** table in `CONTRIBUTOR-DOCS/02_style-guide/02_typescript/14_controller-composition.md`.

gen2/packages/core/controllers/trigger-press-controller/src/trigger-press-controller.ts
```diff
@@ -0,0 +1,272 @@
...
+    this.detach();
+  }
+
+  // ─────────────────────────
+  //     IMPLEMENTATION
+  // ─────────────────────────
+
+  // Capture phase, before a native light-dismiss (which runs on the same
+  // press) gets a chance to close the surface first.
+  private readonly onPressStart = (): void => {
```
rubencarvalho: This tracks *any* `pointerdown`, including secondary buttons, and that can wedge the guard:

1. Right-click the trigger while the surface is open.
2. The platform still light-dismisses (light-dismiss keys off pointerdown/pointerup regardless of button), so the host calls `noteNativeDismiss()` and `dismissedByPress` is set.
3. `pointerup` lands on the trigger, so `onPressEnd` takes its early return and deliberately leaves both flags set for the trailing click.
4. No `click` ever comes — a secondary-button press produces `contextmenu`, not `click`.

Both flags stay set, and the next genuine left click on the trigger is consumed instead of reopening the surface.

This is the same "press ends without a click" class that `DismissDuringCancelledPressDoesNotLeakTest` covers and that the changeset claims to fix, so it seems worth closing here too. Simplest fix is to not open the gesture window for non-primary presses — the handler would need the event it currently discards:

```ts
private readonly onPressStart = (event: Event): void => {
  if (event instanceof PointerEvent && event.button !== 0) {
    return;
  }
  // ...
};
```

(`touchstart` has no `button`, so the `instanceof` guard keeps the touch path unchanged.)

gen2/packages/core/controllers/trigger-press-controller/src/trigger-press-controller.ts
```diff
@@ -0,0 +1,272 @@
...
+    // A same-target pointerup still gets its own click, which reads and
+    // clears `dismissedByPress` itself; clearing it here too would race that
+    // read. Any other press end (pointercancel, or a pointerup that lands off
+    // the trigger) means no click is coming for this gesture, so both flags
+    // are cleared here instead — otherwise a dismissal noted during a press
+    // that never resolves to a click would incorrectly consume the next,
+    // unrelated click on the trigger.
+    if (
+      event.type === 'pointerup' &&
+      event.composedPath().includes(this.trigger as EventTarget)
```
rubencarvalho: Nit: the `as EventTarget` cast hides a nullable. It happens to work (`includes(null)` is `false`, which falls through to the reset — the right outcome), but it reads like the null case was overlooked rather than intended. An explicit check says it out loud:

```ts
if (
  event.type === 'pointerup' &&
  this.trigger !== null &&
  event.composedPath().includes(this.trigger)
) {
```

gen2/packages/core/controllers/trigger-press-controller/src/trigger-press-controller.ts
```diff
@@ -0,0 +1,272 @@
...
+    if (this.pointerActive) {
+      this.dismissedByPress = true;
+    }
+  }
+
+  /**
+   * Lit `ReactiveController` hook: tears down listeners when the host
+   * disconnects.
+   */
+  public hostDisconnected(): void {
```
rubencarvalho: There's no matching `hostConnected`, so a disconnect is permanent for the controller: `detach()` nulls `trigger`, and nothing re-attaches on reconnect.

For `PopoverBase` that means moving a `<swc-popover>` in the DOM silently kills click-to-toggle — `_wireTrigger()` only runs from `updated()` when `for`/`triggerElement`/`modal`/`manual` change, and neither `PopoverBase` nor `SpectrumElement` defines a `connectedCallback`. Pre-existing (the old `_removeTriggerListeners()` had the same shape), so not a regression from this PR.

But now that the controller owns disconnect teardown, it's also the natural place to fix it for every future consumer: keep the last `trigger`/`options` in a field and re-`attach()` them from `hostConnected()`. Fine as a follow-up if you'd rather keep this PR a pure extraction.

gen2/packages/core/components/popover/Popover.base.ts
```diff
@@ -332,7 +317,7 @@ export abstract class PopoverBase extends SpectrumElement {
     unregisterDismissible(this);
     this._removeEscapeListener();
     this._scrollLock.unlock();
-    this._removeTriggerListeners();
+    this._pressGuard.detach();
```
rubencarvalho: I think this is now redundant: `super.disconnectedCallback()` above already runs the controller's `hostDisconnected()`, which calls `detach()`. it contradicts the guidance in the new MDX ("Rely on `hostDisconnected()` for cleanup; there is no need to call `detach()` manually when a host is removed from the DOM").

gen2/packages/core/controllers/trigger-press-controller/src/trigger-press-controller.ts
rubencarvalho: I think we can avoid the cast here:
```suggestion
      event.type === 'pointerup' &&
      this.trigger !== null &&
      event.composedPath().includes(this.trigger)
```

gen2/packages/core/controllers/trigger-press-controller/trigger-press-controller.mdx
Rajdeepc: You're right, it was missing. I added both subpath entries (exports map and typesVersions) in core's package.json, matching the pattern the other controllers with a flat, non-index.js-suffixed import path use.

CONTRIBUTOR-DOCS/01_contributor-guides/16_gen2-shared-resources.md
Rajdeepc: Good catch, thank you. I added a row for it to the Available controllers table there as well.

gen2/packages/core/components/popover/Popover.base.ts
Rajdeepc: Confirmed and removed. I checked Lit's ReactiveElement.disconnectedCallback() and it does call hostDisconnected() on every registered controller, so this really was redundant, and it was contradicting the guidance I wrote in the MDX. Thanks for catching that inconsistency.

gen2/packages/core/controllers/trigger-press-controller/src/trigger-press-controller.ts
Rajdeepc: Applied, thank you. The explicit null check reads better than the cast anyway.

rubencarvalho (approved):

5t3ph (approved):

